PR #124, opened 2023-12-14
Add Portkey gateway to backend docs & example repo

vrushankportkey:


ataibarkai:
Thanks @vrushankportkey!

For maintainability's sake, I think it'd be better to add a second api endpoint to the existing example -
rather than a whole new example.

Then users can simply change the api endpoint given to the CopilotProvider to get the PortKey implementation.
I'm ok with adding a single-line comment in the CopilotProvider highlighting this.

If you want to go the extra mile, you can arrange the file structure under /api/ for multiple implementations, e.g.
/api/copilotkit/openai/route.ts
/api/copilotkit/portkey/route.ts

That way users can see all the offerings in an organized fasion.
What do you think?

ataibarkai:
Oh and one more thing - make sure everything compiles and lints locally.
Instructions here: https://docs.copilotkit.ai/contribute/quickstart-contribute#testing-your-solution

ataibarkai:
hi @vrushankportkey we've updated our architecture, you can now easily add support to portkey by creating a new Adapter class.

See the OpenAIAdapter for reference:
https://github.com/CopilotKit/CopilotKit/blob/aa51c5d2ab57fe6eba48e2fd81007bcdad4723ae/CopilotKit/packages/backend/src/lib/openai-adapter.ts#L15-L57

Let me know if you want to port over the work to this

Issue #62, opened 2023-11-02
Support bold and italicized text in CopilotTextarea


ataibarkai:
### Proposal
- Add support for bold and italicized text in CopilotTextarea
- CopilotTextarea uses [slate-js](https://github.com/ianstormtaylor/slate) under the hood. Lots of examples for adding bold/italicized support
- Initially only add programatic support. UI support will be added separately in [TODO add issue]

### Implementation tips
- changes will be made to `render-element.tsx` and `base-copilot-textarea.tsx` 
-  `custom-editor.tsx` structures may also require changes

zsskar:
I would Iike to work on this feature request

ataibarkai:
@zsskar fantastic!
Happy to provide as much support as you’d like.

I’d encourage you to put up a PR early, to make sure we are aligned on the direction

zsskar:
@ataibarkai i need some help like in the CopilotTextarea for bold and italic support will there be buttons for these two or while selecting the text it should show the options i am little bit confuse here.

RohanMishra315:
Hey @ataibarkai If no one is working I'll do it !

shu-toz:
Is anyone tackling this? @ataibarkai, I'm keen to have a crack

ataibarkai:
@shu-toz please do!
Let me know if you need any support

shiven24k:
Is anyone working on it? I would like to contribute

sahilborse:
@ataibarkai I would like to work on this issue

kachurun:
In my opinion, it's a mistake to use slate.js for this component. I don't think it's CopilotKit's responsibility to implement a WYSIWYG editor in such a simple part as a textarea. I believe it would be better to get rid of it and create a simple, core CopilotContentEditable component instead—without any styling, wrappers, WYSIWYG, or other unnecessary stuff—and allow the user to decide how to use it. If the user wants, they can theoretically render any kind of HTML inside this contenteditable area and implement or integrate any WYSIWYG editor they want.

jpr5:
This was a feature request for bold/italic in CopilotTextarea. The underlying Slate.js approach has since been superseded by architectural changes. Closing as stale.

PR #125, opened 2023-12-14
FIXED -> Support bold and italicized text in CopilotTextarea #62

zsskar:
add functionality as per the requirement please check it if you find it working as per the requirement please merge the code.

### DIFF
--- a/CopilotKit/packages/react-textarea/src/components/base-copilot-textarea/base-copilot-textarea.tsx
+++ b/CopilotKit/packages/react-textarea/src/components/base-copilot-textarea/base-copilot-textarea.tsx
@@ -1,6 +1,6 @@
 import React, { useCallback, useEffect, useMemo, useState } from "react";
-import { Descendant, Editor } from "slate";
-import { Editable, Slate } from "slate-react";
+import { Descendant, Editor ,Transforms, Text } from "slate";
+import { Editable, Slate,RenderElementProps } from "slate-react";
 import { twMerge } from "tailwind-merge";
 import { useAutosuggestions } from "../../hooks/base-copilot-textarea-implementation/use-autosuggestions";
 import { useCopilotTextareaEditor } from "../../hooks/base-copilot-textarea-implementation/use-copilot-textarea-editor";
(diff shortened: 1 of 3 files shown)


CopilotKit/packages/react-textarea/src/types/base/custom-editor.tsx
```diff
@@ -16,7 +16,17 @@ export type SuggestionElement = {
   children: CustomText[];
 };
 
-export type CustomElement = ParagraphElement | SuggestionElement;
+export type BoldElement = {
```
ataibarkai, line 19: looks good

CopilotKit/packages/react-textarea/src/components/base-copilot-textarea/base-copilot-textarea.tsx
```diff
@@ -128,8 +128,54 @@ const BaseCopilotTextareaWithHoveringContext = React.forwardRef(
...
+    >
+      {props.children}
+    </em>
+  );
+};
+
+
     const renderElementMemoized = useMemo(() => {
-      return makeRenderElementFunction(suggestionStyleAugmented);
+      return makeRenderElementFunction({
```
ataibarkai: looks like `makeRenderElementFunction` still only takes a single input (`suggestionStyle`).
So passing it `boldElement` and `italicElement` is incorrect.

We have 2 options here:
1. The pre-existing approach uses CSS styling. We can mirror what we have for the suggestion element with bold/italics 

2. I actually like the approach you started here - using <strong> and <em> to demarcate bold/italics, and carrying over the remaining stylings.

Assuming it works, there's no need to pass boldElement / italicsElement to the `makeRenderElementFunction` function.

ataibarkai (commented):
Thank you. @zsskar !

Looks like it's heading in a good direction.

A couple of related notes:
1. looks like there are some build errors. I added a section to the [contributing docs quick-start](https://docs.copilotkit.ai/contribute/quickstart-contribute) with how to test your solution locally

2. Let's get a running example so we can see it working in action.
You can modify the existing /textarea example, to get text editing working.

We also need to create UI hooks for enabling/disabling bolding, but we can address that separately. For the purposes of the running example, it can be hard-coded.

Let me know if you need additional support on this!

CopilotKit/packages/react-textarea/src/components/base-copilot-textarea/base-copilot-textarea.tsx
zsskar: Ya sure ataibarkai I will first setup it locally and test it on my local and try to make changes what you have mentioned. If I need any help definitely I will ask you.

Thanks.

PR #138, opened 2024-01-04
Add Hello World example

mme:
This adds a "Hello World" example to the nextjs demo.

It can be accessed via http://127.0.0.1:3000/helloworld


ataibarkai (approved):
Looks great!

Congrats on 1st PR! 🎉🎉

CopilotKit/examples/next-openai/src/app/helloworld/page.tsx
```diff
@@ -0,0 +1,84 @@
...
+  CopilotProvider,
+  useMakeCopilotActionable,
+  useMakeCopilotReadable,
+} from "@copilotkit/react-core";
+import { CopilotSidebarUIProvider } from "@copilotkit/react-ui";
+import { useState } from "react";
+
+const HelloWorld = () => {
+  return (
+    <CopilotProvider chatApiEndpoint="/api/copilotkit/chat">
```
ataibarkai: Oh actually one problem!

I just refactored this (that was the cause of the error we were seeing yesterday)
It should point to `/api/copilotkit/openai`

ataibarkai (approved):
I did actually find one issue-
Still leaving it as approved, so you can merge right after fixing it

Issue #136, opened 2023-12-28
Change `content-type` header for POST requests from "text/plain;charset=UTF-8" to `application/json`


arjshiv:
**Describe the bug**
If the request content type header is not set to `application/json` , common API frameworks like `express` with `body-parser` cannot parse it

**To Reproduce**
1. Set up `CopilotTextArea`
2. Send a request to an `express` server with `body-parser` middleware enabled [ really common for JSON ]
3. Try and parse the body with `req.bod`
4. `req.body` is empty

**Expected behavior**
I expect 

**Screenshots**

** Browser (please complete the following information):**
 - Universal

**Additional context**
 - Sending JSON as `json` is a standard convention, this could break not just Express but most other frameworks.

iamtouha:
@ataibarkai I would like to work on this issue.

ataibarkai:
Great @iamtouha , let me know if you need any support

iamtouha:
Certainly! Could you please specify which files I should be looking at to apply the changes.

nsvoriginals:
Is this bug done?

PR #149, opened 2024-01-06
FIXED -> Changed content-type header for POST requests from "text/plain;charset=UTF-8" to "application/json" #136

ElvinPero:
If the request content type header is not set to application/json , common API frameworks like express with body-parser cannot parse it.


zsskar:
> @zsskar Kindly review my PR.

@ataibarkai please check this.

ataibarkai:
Thank you @zsskar .

This is a good change so we can merge it, however you made the change against the experimental v2 implementation, which is not the one most users of the library are using right now.

Let me know if you'd like to also fix this in the current implementation in this git branch.

Off the top of my head, I believe you want to make the change in `api/copilotkit/openai/route.ts`. But I am not 100% cetain.

PR #152, opened 2024-01-10
Quick fix context bug

mme:
This fixes a bug where the context is not updated correctly because the `initialMessages` parameter of `useChat` uses these messages only for initialization. Subsequent updates are ignored by useChat.

I verified the bugfix with the WaterBnb and HelloWorld demos- the context was changing as expected.


ataibarkai:
@mme should we close this PR after the last update?

mme:
@ataibarkai closing

Issue #65, opened 2023-11-02
CopilotTextarea: do not display HTTP error message as insertion/editing suggestion


ataibarkai:
Related to yet distinct from #64 

### Proposal
Currently when the underlying LLM function returns an HTTP error, the entire error payload is displayed as the autocompletion suggestion.
Instead the error should be propagated/re-throwed, and no autocompletion text should be displayed.

### Reproduction tips
- Edit `examples/next-openai/app/api/copilotkit/chat/route.ts` to use `gpt-42` instead of `gpt-4`. 
- Run the `next-openai` example (see quickstart in the [documentation](docs.copilotkit.ai))
- Press Command-K in the CopilotTextarea in the homepage, and type an insertion prompt.

### Implementation tips
- See `use-make-standard-insertion-function.tsx`

ataibarkai:
Fantastic @RohanMishra315 , please let me need if you need any help

RohanMishra315:
Hey @ataibarkai Could you check the PR and do suggest !

ataibarkai:
Taking a look, thanks @RohanMishra315 !

PR #154, opened 2024-01-10
Update route.ts

RohanMishra315:
Solved issue #65 

 From these the error should be propagated/re-throwed, and no autocompletion text should be displayed.


arielweinberger:
Closing this PR as we've refactored the Copilot Runtime and it's now up to date.

PR #158, opened 2024-01-12
Fix textarea

mme:
This makes the textarea work with the vercel ai replacement.

- Introduces a new class `ChatCompletionStream` for simple, text only responses in the form of `ReadableStream<string>`
- Rename the low level http parsing class to `ChatCompletionTransport` for clarity
- `useMakeStandardAutosuggestionFunction` and `useMakeStandardInsertionOrEditingFunction` have been adapted to make use of this

@ataibarkai please test, for me the textarea demo was working, but I am not sure how to test  useMakeStandardInsertionOrEditingFunction

### DIFF
--- a/CopilotKit/packages/react-textarea/src/hooks/make-autosuggestions-function/use-make-standard-autosuggestions-function.tsx
+++ b/CopilotKit/packages/react-textarea/src/hooks/make-autosuggestions-function/use-make-standard-autosuggestions-function.tsx
@@ -1,13 +1,11 @@
-import { CopilotContext } from "@copilotkit/react-core";
+import { CopilotContext, Message } from "@copilotkit/react-core";
 import { useCallback, useContext } from "react";
-import {
-  AutosuggestionsBareFunction,
-  ChatlikeApiEndpoint,
-  MinimalChatGPTMessage,
-} from "../../types";
+import { AutosuggestionsBareFunction, MinimalChatGPTMessage } from "../../types";
 import { retry } from "../../lib/retry";
 import { InsertionEditorState } from "../../types/base/autosuggestions-bare-function";
 import { SuggestionsApiConfig } from "../../types/autosuggestions-config/suggestions-api-config";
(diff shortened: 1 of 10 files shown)


CopilotKit/packages/react-textarea/src/hooks/make-autosuggestions-function/use-make-standard-autosuggestions-function.tsx
```diff
@@ -52,14 +50,20 @@ export function useMakeStandardAutosuggestionFunction(
           },
         ];
 
-        const apiEndpoint = ChatlikeApiEndpoint.fromCopilotApiConfig(copilotApiConfig);
```
ataibarkai, line 55: there's an issue with the new implementation -
`fromCopilotApiConfig` integrates properties from the `copilotApiConfig` that we are now ignoring (.body, .headers)

CopilotKit/packages/react-textarea/src/hooks/make-autosuggestions-function/use-make-standard-insertion-function.tsx
```diff
@@ -67,8 +67,14 @@ export function useMakeStandardInsertionOrEditingFunction(
           },
         ];
 
-        const apiEndpoint = ChatlikeApiEndpoint.fromCopilotApiConfig(copilotApiConfig);
```
ataibarkai, line 70: same as above - we're now ignoring some `copilotApiConfig` properties

Also, whatever we implement for the insertion function, we also need to implement for the editing function below (`const editingFunction = useCallback...`)

ataibarkai (commented):
Looks very good overall, great job.
But there are a couple of problems.

I bet the same issue occurs with the previous chatbot implementation so we should take a look there

ataibarkai:
hey @mme , i addressed the concerns above in this PR:
https://github.com/CopilotKit/CopilotKit/pull/159

Looks good to merge?

Also, for future reference, in the textarea you bring up the insertion/edit box with command+k

PR #181, opened 2024-01-24
fix: hot reload when dependencies change in dev

arielweinberger:
# The Problem
The `dev` script was using the same `tssup.config.ts` script as `build`.

Whenever you'd make a change to any dependency, `clean: true` causes the removal of the entire `dist` folder. This results in a few miliseconds where the dependency cannot be found, enough to cause an error.

# The Fix
Removing `clean: true` from `tsup.config.ts` and specifying it as a `--clean` flag specifically for `build`.

### DIFF
--- a/CopilotKit/packages/react-core/tsup.config.ts
+++ b/CopilotKit/packages/react-core/tsup.config.ts
@@ -5,7 +5,6 @@ export default defineConfig((options: Options) => ({
   format: ["esm", "cjs"],
   dts: true,
   minify: false,
-  clean: true,
   external: ["react"],
   sourcemap: true,
   exclude: [
--- a/CopilotKit/packages/backend/package.json
+++ b/CopilotKit/packages/backend/package.json
@@ -8,11 +8,13 @@
   "sideEffects": false,
   "main": "./dist/index.js",
   "module": "./dist/index.mjs",
-  "exports": "./dist/index.mjs",
+  "exports": {
+    ".": "./dist/index.js"
+  },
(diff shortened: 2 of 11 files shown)


CopilotKit/packages/react-core/tsup.config.ts
```diff
@@ -5,7 +5,6 @@ export default defineConfig((options: Options) => ({
   format: ["esm", "cjs"],
   dts: true,
   minify: false,
-  clean: true,
```
ataibarkai, line 8: dang. good find.

ataibarkai: Down the line it might be a good idea to refactor the clean config variable in the `tsup.config.ts` itself, e.g. via an environment variable. 

adding `--clean` everywhere is a bit brittle.

But we can address that later, i will create an issue

ataibarkai (approved):
great find, what a fail

CopilotKit/packages/react-core/tsup.config.ts
arielweinberger: yeah the best way to approach this IMO is to have a base "tsup.config.ts" and "tsup.config.dev.ts" file that sits in a shared config folder, imported by the libs and extended.

ataibarkai: yes, added this issue here, with your approach as an alternative:
https://github.com/CopilotKit/CopilotKit/issues/182

PR #185, opened 2024-01-25
remove the v2 api route and associated code

mme:
This removes the v2 api route and the associated assistant API code.


mme:
@ataibarkai please have a look if this is what we want.

This removes all the code associated with the assistant API too, since it depends on the v2 route.

We also discussed removing `CopilotApiConfig` but I think it is good to have this internally (to keep all configuration in one place)

PR #186, opened 2024-01-25
fix: import issues related to treeshaking

arielweinberger:
Still not entirely sure why this only happens with `<CopilotTextarea>`, but seems that removing the tsup `--treeshake` flag solves the issue. There are still type issues that show up in the IDE, will tackle in a separate PR.

### DIFF
--- a/CopilotKit/packages/react-textarea/package.json
+++ b/CopilotKit/packages/react-textarea/package.json
@@ -17,7 +17,7 @@
   "types": "./dist/index.d.ts",
   "license": "MIT",
   "scripts": {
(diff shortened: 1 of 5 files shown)


CopilotKit/packages/react-textarea/package.json
```diff
@@ -17,7 +17,7 @@
   "types": "./dist/index.d.ts",
   "license": "MIT",
   "scripts": {
-    "build": "tsup --treeshake --clean",
+    "build": "tsup --clean --no-splitting",
```
ataibarkai: --no-splitting
on purpose?

PR #189, opened 2024-01-25
Add CopilotTask class

mme:
This adds `CopilotTask`, a class for running custom tasks (function calling) without chat.

`instructions` tells the GPT what to do.

There is an optional `data` parameter to provide context.

In addition, it can take the current state and functions from an optional `context` object (using data and functions previously defined via `useMakeCopilotReadable` and `useMakeCopilotActionable`.

It can also take an array of `functions` (type: `AnnotatedFunction`) to make an independent, standalone task.

Example:

```ts
const context = useCopilotContext();
const randomSlideTask = new CopilotTask({
  instructions: "Make a random slide",
});
await randomSlideTask.run(context);
```

### DIFF
--- a/CopilotKit/packages/react-core/src/lib/copilot-task.ts
+++ b/CopilotKit/packages/react-core/src/lib/copilot-task.ts
@@ -1,11 +1,5 @@
-import {
-  AnnotatedFunction,
-  FunctionCall,
-  Message,
-  Role,
-  annotatedFunctionToChatCompletionFunction,
-} from "@copilotkit/shared";
-import { CopilotApiConfig, CopilotContextParams } from "../context";
+import { AnnotatedFunction, FunctionCall, Message } from "@copilotkit/shared";
+import { CopilotContextParams } from "../context";
 import { defaultCopilotContextCategories } from "../components";
 import { fetchAndDecodeChatCompletion } from "../utils/fetch-chat-completion";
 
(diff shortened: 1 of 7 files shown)


CopilotKit/packages/react-core/src/lib/copilot-task.ts
```diff
@@ -0,0 +1,136 @@
...
+      entryPoints[fn.name] = fn;
+    }
+
+    let contextString = "";
+
+    if (this.data) {
+      contextString =
+        (typeof this.data === "string" ? this.data : JSON.stringify(this.data)) + "\n\n";
+    }
+
```
ataibarkai: did you have a specific use-case in mind for data?

it feels like something that may be useful, but if we don't have a specific use-case yet, it may be premature.

A few thoughts:
- arguably data should be typed (by making the interface generic)
- perhaps data should be provided upon running the task, rather than upon upon creating the task?
- how does the data relate to all the useMakeCopilotReadable context?
- how is data different from context provided inside the instructions string?

ataibarkai (commented):
Love CopilotTask.

It feels like we still have some abstractions to figure out around the global context, but we will figure those out as we go.

One thing did stick out a bit - the `data` param.
I left some comments about it

CopilotKit/packages/react-core/src/lib/copilot-task.ts
mme: Hey Atai,

Yes, there are two use cases I have in mind for `data`:

1. When `data` comes from another API call, it does not fit into our `useMakeCopilotReadable` paradigm.
2. Sometimes `data` might even come from a user input field, but it's only needed to run a `CopilotTask`. So making it "readable" temporarily seems fragile and unpredictable, although it will work.

- _arguably data should be typed (by making the interface generic)_
In this design, the type would not be used again afterwards. But maybe when data is provided when running the as per your next point there would be a small benefit in terms of type safety/ catching bugs.
- _perhaps data should be provided upon running the task, rather than upon upon creating the task?_
Good Point!
- _how does the data relate to all the useMakeCopilotReadable context?_
It's merged with the context, unless `includeCopilotReadable` is set to false
- _how is data different from context provided inside the instructions string?_
`instructions` is what to do, `data` is the context to use. But again, it makes much more sense to provide the data when running the task, as per your second point!

ataibarkai: Makes sense!

So should we switch it to a typed object provided on run?

mme: 👍  Done

ataibarkai (approved):

PR #192, opened 2024-01-26
Support writing node server responses

mme:
This adds support for streaming back responses using the built in node http server.


CopilotKit/package.json
```diff
@@ -2,7 +2,7 @@
   "private": false,
   "scripts": {
     "build": "turbo build",
-    "dev": "turbo dev",
+    "dev": "turbo dev --concurrency 11",
```
ataibarkai, line 5: note for future: dev build complains that more concurrency is required to run this

PR #199, opened 2024-01-29
Fix: OpenAI gets instantiated before running constructor

### LINKED ISSUE #193: The `OpenAI` argument passed to `OpenAIAdapter` is not being respected
**Describe the bug**
`OpenAIAdapter` allows for passing an `OpenAI` object as a parameter.
```
new OpenAIAdapter({openai})
```

However this parameter does not appear to be respected in runtime.

**To Reproduce**
1. Remove `OPENAI_API_KEY` env var
2. Create OpenAI instance with a manually specified key
3. Pass the OpenAI instance as the openai param to OpenAIAdapter
4. Try to use backend, observe failure

```
import { CopilotBackend, OpenAIAdapter } from "@copilotkit/backend";
import { OpenAI } from "openai";

export const runtime = "edge";


const openai = new OpenAI({
  apiKey: "XXXX",
});
const openaiAdapter = new OpenAIAdapter({openai});

export async function POST(req: Request): Promise<Response> {
  const copilotKit = new CopilotBackend({
    functions: [],
  });

  return copilotKit.response(req, openaiAdapter);
}

```
**Expected behavior**
CopilotBackend should run fine, with the manually specified API key.

**Additional context**
Tried both Node and Next.js runtimes.

mme:
Fixes #193 

The problem was that the default for the `openai` instance var gets instantiated before the constructor is run, thus resulting in an error when the env variable is missing.

### DIFF
--- a/CopilotKit/packages/backend/src/lib/openai-adapter.ts
+++ b/CopilotKit/packages/backend/src/lib/openai-adapter.ts
@@ -13,13 +13,11 @@ export interface OpenAIAdapterParams {
 }
 
 export class OpenAIAdapter implements CopilotKitServiceAdapter {
-  private openai: OpenAI = new OpenAI({});
+  private openai: OpenAI;
   private model: string = DEFAULT_MODEL;
   private debug: boolean = false;
   constructor(params?: OpenAIAdapterParams) {
-    if (params?.openai) {
-      this.openai = params.openai;
-    }
+    this.openai = params?.openai || new OpenAI({});
     if (params?.model) {
       this.model = params.model;
     }


ataibarkai (approved):

ataibarkai:
So strange.

Does OpenAI maintain global state as part of the OpenAI instance initialization?

mme:
No, it's just that `private openai: OpenAI = new OpenAI({});` is executed before the constructor.

mme:
(and it throws if it does not have an apiKey as constructor or the env variable)

PR #211, opened 2024-02-05
Implement function call results and feedback

### LINKED ISSUE #141: Actions should be able to return results
**Is your feature request related to a problem? Please describe.**

Actions should be able to return an optional result when being executed. This result can then be fed back to the LLM.

This enables new functionality:

- LLM can use the function to look something up (i.e. not RAG but a "search" REST call, a call to get the status of something, etc)
- LLM can learn about the result of a function call and respond by chatting with the user.
- LLM can do consecutive function calls based on the result of a previous function

**Describe the solution you'd like**

- Change the function type of `implementation` to `(...args: Inputs) => Promise<any>` so that it can optionally return a result.
- introduce `maxFeedback` parameter to control how often the output of a function is fed back
- See this: https://github.com/mme/beakjs/blob/c837e84da19082ae6ed977381548efb74930d678/packages/core/src/beak.ts#L128

mme:
This adds function call results to the messages stream and feeds back the results to GPT.

With this PR, CopilotKit can now:
- retrieve information from the backend for context
- present this information or use it in another function call
- chain multiple function calls together and use results of previous function calls

For now, this is implemented by extending the OpenAI streaming protocol. The code needs cleanup by properly implementing our own streaming protocol between client and server.

fix #141


CopilotKit/packages/backend/src/utils/openai.ts
```diff
@@ -155,12 +157,32 @@ export function copilotkitStreamInterceptor(
     for (let arg of fn.argumentAnnotations) {
       paramsInCorrectOrder.push(args[arg.name as keyof typeof args]);
     }
-    await fn.implementation(...paramsInCorrectOrder);
+    const result = await fn.implementation(...paramsInCorrectOrder);
+
+    let resultString = "";
+    if (result !== undefined) {
+      resultString = typeof result === "string" ? result : JSON.stringify(result);
```
ataibarkai: nit:
might be worth putting an if else block, getting a bit much for ternary operator

CopilotKit/packages/react-core/src/hooks/use-chat.ts
```diff
@@ -154,15 +175,48 @@ export function useChat(options: UseChatOptionsWithCopilotConfig): UseChatHelper
...
           setMessages([...messages, ...newMessages]);
 
           // Execute the function call
-          await options.onFunctionCall?.(messages, currentMessage.function_call);
+          try {
+            if (options.onFunctionCall && value.scope === "client") {
+              const result = await options.onFunctionCall(messages, currentMessage.function_call);
+              let resultString = "";
+              if (result !== undefined) {
+                resultString = typeof result === "string" ? result : JSON.stringify(result);
```
ataibarkai: nit: might be worth factoring this into a utility function?

ataibarkai (commented):
Hey Markus, looks like great work.

But i don't fully understand the architecture -- why do we need to add a scope field to the function call? the client + server both already know where a function came from right?

also, i left 2 comments prefixed with 'nit'.
'nit' stands for 'nitpick', i.e. minor comment that isn't a big deal.
Feel free to ignore nits or act on them as you see fit.

mme:
Hey @ataibarkai !

We need to add the `scope` field because the client does not know that the server has already executed the function. But we do need the function call and result from the server for the chat history.

ataibarkai (approved):

Issue #64, opened 2023-11-02
CopilotTextarea: do not display HTTP error message as autocompletion suggestion


ataibarkai:
Related to yet distinct from #65 

### Proposal
Currently when the underlying LLM function returns an HTTP error, the entire error payload is displayed as the autocompletion suggestion.
Instead the error should be propagated/re-throwed, and no autocompletion text should be displayed.

### Reproduction tips
- Edit `examples/next-openai/app/api/copilotkit/chat/route.ts` to use `gpt-42` instead of `gpt-4`.  (this will error out, which is what we want).
- Run the `next-openai` example (see quickstart in the [documentation](docs.copilotkit.ai))
**- Start typing in the CopilotTextarea in the homepage, and wait a second for the autocompletion**

### Implementation tips
- See `use-make-standard-autosuggestions-function.tsx`

sakshambp:
I have been trying to run example locally however failing to do so, after lot of npm installation manually i get following on `yarn run`

```
@copilotkit/react-textarea:build: src/lib/debouncer.ts(15,5): error TS2322: Type 'Timeout' is not assignable to type 'number'.
@copilotkit/react-textarea:build: 
@copilotkit/react-textarea:build: RollupError: Failed to compile. Check the logs above.
@copilotkit/react-textarea:build:     at error (/Users/saksham.jain/Documents/ago/copi/CopilotKit/CopilotKit/node_modules/.pnpm/rollup@3.4.0/node_modules/rollup/dist/shared/rollup.js:206:30)
@copilotkit/react-textarea:build:     at Object.error (/Users/saksham.jain/Documents/ago/copi/CopilotKit/CopilotKit/node_modules/.pnpm/rollup@3.4.0/node_modules/rollup/dist/shared/rollup.js:23607:20)
@copilotkit/react-textarea:build:     at Object.error (/Users/saksham.jain/Documents/ago/copi/CopilotKit/CopilotKit/node_modules/.pnpm/rollup@3.4.0/node_modules/rollup/dist/shared/rollup.js:22872:42)
@copilotkit/react-textarea:build:     at Object.transform (/Users/saksham.jain/Documents/ago/copi/CopilotKit/CopilotKit/node_modules/.pnpm/tsup@6.5.0_postcss@8.4.20_typescript@4.9.4/node_modules/tsup/dist/rollup.js:7287:20)
[...]

PR #218, opened 2024-02-10
fix(react-textarea): Correct `timeoutId` type in Debouncer

arnoldmukisa:
Resolved a type mismatch issue in `debouncer.ts` by changing the `timeoutId` type from `number` to `NodeJS.Timeout`, aligning with the correct return type of `setTimeout` in NodeJS environments.

This was causing a build error

### DIFF
--- a/CopilotKit/packages/react-textarea/src/lib/debouncer.ts
+++ b/CopilotKit/packages/react-textarea/src/lib/debouncer.ts
@@ -1,7 +1,7 @@
 export type AsyncFunction<T extends any[]> = (...args: [...T, AbortSignal]) => Promise<void>;
 
 export class Debouncer<T extends any[]> {
-  private timeoutId?: number;
+  private timeoutId?: NodeJS.Timeout;
   private activeAbortController?: AbortController;
 
   constructor(private wait: number) {}
--- a/CopilotKit/packages/react-textarea/src/lib/debouncer.ts
+++ b/CopilotKit/packages/react-textarea/src/lib/debouncer.ts
@@ -1,7 +1,7 @@
 export type AsyncFunction<T extends any[]> = (...args: [...T, AbortSignal]) => Promise<void>;
 
 export class Debouncer<T extends any[]> {
-  private timeoutId?: NodeJS.Timeout;
+  private timeoutId?: ReturnType<typeof setTimeout>;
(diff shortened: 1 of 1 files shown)


CopilotKit/packages/react-textarea/src/lib/debouncer.ts
```diff
@@ -1,7 +1,7 @@
 export type AsyncFunction<T extends any[]> = (...args: [...T, AbortSignal]) => Promise<void>;
 
 export class Debouncer<T extends any[]> {
-  private timeoutId?: number;
+  private timeoutId?: NodeJS.Timeout;
```
ataibarkai: this is really strange.

react-textarea targets frontends, so ` NodeJS.Timeout` should throw a build error, and `number` should not (it doesn't on all our machines 🤔)

need to figure out why this error occured

ataibarkai:
I also see that on our CI build system, the new addition results in a build failure, not the old one.

So the question is -- why is react-textarea building targeting Node.js on your machine @arnoldmukisa?
🤔

CopilotKit/packages/react-textarea/src/lib/debouncer.ts
arnoldmukisa: @ataibarkai I see very strange. i don't see anything odd with my setup so far

could we replace with `private timeoutId?: ReturnType<typeof setTimeout>;` which will correctly represent the return type of setTimeout in both browser and Node.js environments?

arnoldmukisa: @ataibarkai I found that one other contributer run into this issue before: https://github.com/CopilotKit/CopilotKit/issues/64#issuecomment-1820691561

ataibarkai: Very strange. I like your suggested change as a workaround in the meantime- let’s go with that.

I’ll crate a separate ticket to track the underlying issue

arnoldmukisa: Pushed

ataibarkai (approved):

ataibarkai:
Looks good @arnoldmukisa , thank you

PR #234, opened 2024-02-20
Extend action definitions: Support type inference and complex types

mme:
This adds support for

- complex types such as arrays of objects, nested properties etc.
- enums
- type inferences in the handler function

TODO:

- [x] server side support
- [x] deal with prettier error
- [x] testing
- [x] what happens if we pass empty dependencies? Make sure it works as expected
  - [x] https://react.dev/reference/react/useEffect#examples-dependencies
  - [x] `useCallback` to implement?
- [x] add documentation and migration guide
- [x] more comments - why we need to directly update state in useCopilotAction (infinite loop etc., not UI, can't use useCallback)
- [x] migration guide


CopilotKit/.prettierignore
```diff
@@ -2,4 +2,6 @@ README.md
 .next
 .nuxt
 node_modules
-dist
\ No newline at end of file
+dist
+packages/react-core/src/hooks/use-copilot-action.ts
```
ataibarkai, line 6: Great stuff!

Just documenting our discussion comments here for convenience:
- instead of removing both of these files from prettier, let's make the minimal amount of code needed to prettier-ignore, and put it in 2 distinct files
    - `useCopilotAction`: straightforward, just move the hook function
    - `addAction`:  create a base class for the implementation, and extend it with a light subclass that only exposes a different interface (`const ...` constructor)
    - Add bug repot ticket to prettier itself

ataibarkai (commented):
Amazing stuff, really looking forward to merging it!!

PR #270, opened 2024-03-15
Challenge Accepted: Reorganize MappedParameterTypes

pashabou:
Saw the [original tweet](https://twitter.com/ataiiam/status/1765089261374914957) and this [follow-up post](https://www.copilotkit.ai/blog/posts/ceiling-has-been-raised-analyzing) and saw an opportunity to present something a *little* different.

The goal of this PR is not to rearchitect the typing nor to find clever tricks to shrink or minify the code. I quite like the ergonomics of the typing support for parameters of `useCopilotAction` so I tried to keep the functionality as close as possible to the existing implementation (I haven't written a test suite to be super confident, but the common cases work as I expect). I wanted to make these essentially "parser" cases more independent, which makes it not only easier to reason about as a reader but also simpler to bolt-on additional functionality down the line.

For example, let's look at the main type declaration and a specific "parsing" type in the "boring-looking version":
```ts
type EnumParameterType<E, Required> = E extends string
  ? Required extends false
    ? E | undefined
    : E
  : never;

// prettier-ignore
export type MappedParameterTypes<T extends Parameter[]> = {
    [P in T[number] as P["name"]]: P extends { enum: Array<infer E> } ? EnumParameterType<E, P["required"]>
    : P extends { type: "object"; attributes: infer Attributes } ? ObjectTypeParameter<Attributes>
    : P extends { type: "object[]"; attributes?: never } ? any[]
    : P extends { type: "object[]"; attributes: infer Attributes } ? ObjectArrayTypeParameter<Attributes>
    : OtherParameterType<P["type"], P["required"]>;
};
```
The logic here determines what type of parameter we're dealing with through serial type checks and terminates or passes off the parsing to another type to expand as necessary. Deeply nested ternaries and AI authorship aside, the code confounds two parts of the parsing process -- in my opinion: determining the shape and kind of the parameter, and the data within. In addition, the properties are coupled unnecessarily: why should each "parsing" type (like `EnumParameterType`) need to worry about `Required` (which is extracted from the parameter at a different location)? And beware of parsing `{ type: "object[]"; attributes: infer Attributes }` before `{ type: "object[]"; attributes?: never }`.

In my update:
```ts
type OptionalParameterType<P extends AbstractParameter> = P["required"] extends false
  ? undefined
  : never;

export type MappedParameterTypes<T extends Parameter[]> = {
  [P in T[number] as P["name"]]:
    | OptionalParameterType<P>
    | StringParameterType<P>
    | ObjectParameterType<P>
    | ObjectArrayParameterType<P>
    | BaseParameterType<P>;
};
```
You'll notice that, most importantly, the `prettier-ignore` directive is removed so the new code is objectively more beautiful. On top of that, we've moved away from ternary expressions, and leveraged discriminated unions to our benefit. Not only is it clearer and easier to reason about for sore eyes, but it's also trivial to add additional types without worrying about parsing order.

Let's turn our attention to the existing parameter definitions:
```ts
export type BaseParameter = {
  name: string;
  type?: Exclude<keyof TypeMap, "string" | "object" | "object[]">; // Exclude object types for BaseParameter
  description?: string;
  required?: boolean;
};
export type ObjectParameter = {
  name: string;
  type: "object" | "object[]";
  description?: string;
  required?: boolean;
  attributes?: Parameter[];
};
export type Parameter = BaseParameter | StringParameter | ObjectParameter;
```
Repetitive, but gets the job done. With the existing implementation you would have to keep in mind that that the parameter type strings are duplicated back in `MappedParameterTypes` so you'd have to keep them in sync as well. Oh, and that `BaseParameter` needs to have exclusions for types of other parameters which need to be added manually. If only we could solve that too...
```ts
type SpecialParameters = StringParameter | ObjectParameter | ObjectArrayParameter;
interface BaseParameter extends AbstractParameter {
  type?: Exclude<AbstractParameter["type"], SpecialParameters["type"]>;
}
export type Parameter = BaseParameter | SpecialParameters;
```
Now what about the "parsing" type expressions themselves? 
```ts
type EnumParameterType<E, Required> = E extends string
  ? Required extends false
    ? E | undefined
    : E
  : never;
type ObjectTypeParameter<Attributes> = Attributes extends Parameter[]
  ? MappedParameterTypes<Attributes>
  : never;
```
Maybe this doesn't look like parsing to you, and I'd remind you that some of the important work was actually done in the `MappedParameterTypes` expression itself. Once we return that to the specialized expressions, we find something more regular.
```ts
type StringParameterType<P> = P extends StringParameter
  ? P extends { enum?: Array<infer E> }
    ? E
    : string
  : never;
type ObjectParameterType<P> = P extends ObjectParameter
  ? P extends { attributes?: infer Attributes extends Parameter[] }
    ? MappedParameterTypes<Attributes>
    : object
  : never;
```
Definitely not shorter, but I would absolutely argue for its clarity. Consistent inputs, explicit type checks with the parameter types we constructed above, and decoupled logic. 

I can't say it's perfect but certainly a step in the right direction. Thanks for the challenge!


pashabou:
> https://docs.copilotkit.ai/contribute/quickstart-contribute

Thanks for this, I didn't find a contribution page at first. Tests all pass locally now.

mme:
Hi @pashabou, this is just to let you know that we are adding some test cases now specifically for the type inference in useCopilotAction.

Will let you know about the outcome!

mme:
@pashabou @ataibarkai looks good to me! 

I tested with some use cases which are now defined in QA scripts (see here: https://github.com/CopilotKit/CopilotKit/pull/283)

ataibarkai (approved):
Incredible work @pashabou  - and congratulations!

We'll make an announcement post and coordinate with you on the prize winnings by the end of this month

PR #274, opened 2024-03-17
Feat: When there is an error response from OpenAI, display it in the chat window #267

### LINKED ISSUE #267: When there is an error response from OpenAI, display it in the chat window
**Is your feature request related to a problem? Please describe.**

Right now, when the API returns an error message, for example "model not found", the Copilot fails quietly and shows an empty response. 

**Describe the solution you'd like**
Instead of showing an empty response, show an error message as response in the chat window.

Currently there is an error message displayed in the larger window at bottom left side. Do not remove it, add the in-chat error message in addition.

**Reproduce**:
To reproduce the issue, provide an invalid OpenAI api token in your environment, and follow the OpenAI-based CopilotKit backend.

VaibhavKVerma:
/claim #267
/fixes #267

Screenshot 2024-03-17 at 10 53 05 AM

- Handled API error
- Added the error statuxText as response in the message


VaibhavKVerma:
Please review @mme / @ataibarkai

CopilotKit/packages/backend/src/lib/openai-adapter.ts
```diff
@@ -40,15 +40,19 @@ export class OpenAIAdapter implements CopilotKitServiceAdapter {
...
         messages: messages as any,
-      })
-      .toReadableStream();
-
-    return { stream };
+      });
+      stream.on("error", (error) => {
+        reject(error); // Reject the promise with the error
+      });
+      stream.on("chatCompletion", () => {
```
ataibarkai: using `.on("chatCompletion"` here breaks the backend, because this trigger it's called when the stream finishes - by which point it's too late to stream

But if you replace this with `.on("connect")` it should work 
https://github.com/openai/openai-node/blob/master/helpers.md#onconnect---

ataibarkai (changes requested):
Hi @VaibhavKVerma thanks for the PR.

A small change is needed. See comment.

CopilotKit/packages/backend/src/lib/openai-adapter.ts
VaibhavKVerma: Ok will do the changes

VaibhavKVerma:
Please review @ataibarkai

ataibarkai (approved):
looks great, thanks!

PR #280, opened 2024-03-21
remove setting NODE_OPTIONS env inline

mme:
Setting `NODE_OPTIONS='--inspect'` is invalid on a windows command line and causes the build to fail.


mme:
Alternative: https://stackoverflow.com/questions/67993622/node-options-is-not-recognized-as-an-internal-or-external-command-still-an-i

ataibarkai:
For context, seems this is for enabling the debugger

PR #282, opened 2024-03-21
Fix ESM build errors when not using "use client" in the host app

mme:
This adds `"use client"` in the `index.tsx` files of `react-core`, `react-ui` and `react-textarea`.

This way, the host app does not need to worry about placing "use client" when using CopilotKit.


ataibarkai (approved):
Sounds good - but let's do a full QA run against a beta build

PR #309, opened 2024-04-15
function calling works with new GPT-4-Turbo (April 2024)

ataibarkai:
1. Bump OpenAI SDK dependency to latest version
2. Clarity improvements / simplifications 

3. **Do not forward function_call.scope to OpenAI**

@mme can you please review this, with a focus on the 3rd point?
I am not 100% certain it's ok to remove `function_call.scope` here, but I think it should be fine.

I also added GitHub comments on the code to help with the code review.

### DIFF
--- a/CopilotKit/examples/next-openai/src/app/api/copilotkit_v2_work-in-progress/assistant/route.ts
+++ b/CopilotKit/examples/next-openai/src/app/api/copilotkit_v2_work-in-progress/assistant/route.ts
@@ -1,6 +1,6 @@
 import { experimental_AssistantResponse } from "@copilotkit/backend";
 import OpenAI from "openai";
-import { MessageContentText } from "openai/resources/beta/threads/messages/messages";
+import { MessageContent } from "openai/resources/beta/threads/messages/messages";
 
 // Create an OpenAI API client (that's edge friendly!)
 const openai = new OpenAI({
@@ -124,9 +124,10 @@ export async function POST(req: Request) {
         sendMessage({
           id: message.id,
           role: "assistant",
-          content: message.content.filter(
-            (content) => content.type === "text",
-          ) as Array<MessageContentText>,
+          content: message.content.filter((content) => content.type === "text") as {
+            type: "text";
+            text: { value: string };
+          }[],
         });
       }
     },
--- a/CopilotKit/packages/backend/src/lib/copilotkit-backend-implementation.ts
+++ b/CopilotKit/packages/backend/src/lib/copilotkit-backend-implementation.ts
@@ -89,25 +89,18 @@ export class CopilotBackendImplementation {
       }
     }
 
-    // merge server side functions with langserve functions
-    let mergedTools = mergeServerSideTools(
-      this.actions.map(actionToChatCompletionFunction),
-      langserveFunctions.map(actionToChatCompletionFunction),
-    );
-
-    // merge with client side functions
-    mergedTools = mergeServerSideTools(mergedTools, forwardedProps.tools);
+    const serversideTools: Action<any>[] = [...this.actions, ...langserveFunctions];
+    const mergedTools = flattenToolCallsNoDuplicates([
+      ...serversideTools.map(actionToChatCompletionFunction),
+      ...forwardedProps.tools,
+    ]);
 
     try {
       const result = await serviceAdapter.getResponse({
         ...forwardedProps,
         tools: mergedTools,
       });
-      const stream = copilotkitStreamInterceptor(
-        result.stream,
-        [...this.actions, ...langserveFunctions],
-        this.debug,
-      );
+      const stream = copilotkitStreamInterceptor(result.stream, serversideTools, this.debug);
       return { stream, headers: result.headers };
     } catch (error) {
       console.error("Error getting response:", error);
@@ -164,17 +157,14 @@ export class CopilotBackendImplementation {
   }
 }
 
-export function mergeServerSideTools(
-  serverTools: ToolDefinition[],
-  clientTools?: ToolDefinition[],
-) {
-  let allTools: ToolDefinition[] = serverTools.slice();
-  const serverToolsNames = serverTools.map((tool) => tool.function.name);
-  if (clientTools) {
-    allTools = allTools.concat(
-      // filter out any client functions that are already defined on the server
-      clientTools.filter((tool: ToolDefinition) => !serverToolsNames.includes(tool.function.name)),
-    );
+export function flattenToolCallsNoDuplicates(toolsByPriority: ToolDefinition[]): ToolDefinition[] {
+  let allTools: ToolDefinition[] = [];
+  const allToolNames: string[] = [];
+  for (const tool of toolsByPriority) {
+    if (!allToolNames.includes(tool.function.name)) {
+      allTools.push(tool);
+      allToolNames.push(tool.function.name);
+    }
   }
   return allTools;
 }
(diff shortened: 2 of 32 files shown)


CopilotKit/examples/next-openai/src/app/api/copilotkit_v2_work-in-progress/assistant/route.ts
```diff
@@ -124,9 +124,10 @@ export async function POST(req: Request) {
         sendMessage({
           id: message.id,
           role: "assistant",
-          content: message.content.filter(
-            (content) => content.type === "text",
-          ) as Array<MessageContentText>,
+          content: message.content.filter((content) => content.type === "text") as {
```
ataibarkai, line 127: accommodate change to the openai sdk.
This isn't used anyways (dead code, kept around for now because it's architecturally useful for our coming protocol refactor)

CopilotKit/packages/backend/src/lib/copilotkit-backend-implementation.ts
```diff
@@ -89,25 +89,18 @@ export class CopilotBackendImplementation {
...
 
-    // merge server side functions with langserve functions
-    let mergedTools = mergeServerSideTools(
-      this.actions.map(actionToChatCompletionFunction),
-      langserveFunctions.map(actionToChatCompletionFunction),
-    );
-
-    // merge with client side functions
-    mergedTools = mergeServerSideTools(mergedTools, forwardedProps.tools);
+    const serversideTools: Action<any>[] = [...this.actions, ...langserveFunctions];
```
ataibarkai, line 92: we need to pass `serversideTools` to `copilotkitStreamInterceptor` below.
So we are defining it as a single unit once explicitly as `serversideTools`, instead of twice implicitly (Before, we were calling `[...this.actions, ...langserveFunctions]` again)

CopilotKit/packages/backend/src/lib/copilotkit-backend-implementation.ts
```diff
@@ -89,25 +89,18 @@ export class CopilotBackendImplementation {
...
+    const mergedTools = flattenToolCallsNoDuplicates([
```
ataibarkai, line 93: Previously we were calling the old `mergeServerSideTools` twice, once with client tools.

We were simply using it as a de-duplication step by precedence as prioirty.

So i re-implemented `flattenToolCallsNoDuplicates` to take an array of tools and de-dupe it (earlier = higher priority). See below in this file

CopilotKit/packages/backend/src/lib/copilotkit-backend-implementation.ts
```diff
@@ -164,17 +157,14 @@ export class CopilotBackendImplementation {
...
-  clientTools?: ToolDefinition[],
-) {
-  let allTools: ToolDefinition[] = serverTools.slice();
-  const serverToolsNames = serverTools.map((tool) => tool.function.name);
-  if (clientTools) {
-    allTools = allTools.concat(
-      // filter out any client functions that are already defined on the server
-      clientTools.filter((tool: ToolDefinition) => !serverToolsNames.includes(tool.function.name)),
-    );
+export function flattenToolCallsNoDuplicates(toolsByPriority: ToolDefinition[]): ToolDefinition[] {
```
ataibarkai, line 160: see previous comment

CopilotKit/packages/backend/src/lib/copilotkit-backend-implementation.ts
```diff
@@ -164,17 +157,14 @@ export class CopilotBackendImplementation {
...
+  let allTools: ToolDefinition[] = [];
+  const allToolNames: string[] = [];
+  for (const tool of toolsByPriority) {
+    if (!allToolNames.includes(tool.function.name)) {
+      allTools.push(tool);
+      allToolNames.push(tool.function.name);
+    }
```
ataibarkai, line 167: see previous comment

CopilotKit/packages/backend/src/lib/openai-adapter.ts
```diff
@@ -40,6 +40,14 @@ export class OpenAIAdapter implements CopilotKitServiceAdapter {
...
     );
 
+    // remove message.function_call.scope if it's present.
+    // scope is a field we inject as a temporary workaround (see elsewhere), which openai doesn't understand
+    messages.forEach((message) => {
+      if (message.function_call?.scope) {
+        delete message.function_call.scope;
+      }
+    });
+
```
ataibarkai: the presence of `function_call?.scope` causes an error in openai calls with the latest GPT-4-Turbo model.

So we remove it here before calling openai.

@mme: can you please verify that the order of operations allows for this?
We don't need this information from the return value of this function, right? (because we have it elsewhere)

CopilotKit/packages/backend/src/utils/stream.ts
```diff
@@ -145,55 +149,61 @@ export function copilotkitStreamInterceptor(
...
+
           // if we get a function call, emit it only if we don't execute it server side
-          else if (mode === "function") {
+          else if (mode.type === "function") {
             // Set the function name if present
-            if (value.choices[0].delta.tool_calls?.[0]?.function?.name) {
-              functionCallName = value.choices[0].delta.tool_calls![0].function.name!;
+            const maybeFunctionName = mode.toolCall.function.name;
+            if (maybeFunctionName) {
+              functionCallName = maybeFunctionName;
```
ataibarkai, line 193: here and below, we repeat ourselves less by defining an intermediate variable, checking if it's null/undefined, and then proceeding

CopilotKit/packages/backend/src/utils/stream.ts
```diff
@@ -145,55 +149,61 @@ export function copilotkitStreamInterceptor(
...
+          } // done == true (terminal case)
+
+          if (debug) {
             console.log("data: " + JSON.stringify(value) + "\n\n");
           }
 
-          let mode: "function" | "message" = value.choices[0].delta.tool_calls
-            ? "function"
-            : "message";
+          type Mode = { type: "function"; toolCall: ToolCallPayload } | { type: "message" };
```
ataibarkai, line 158: store mode type + associated value (toolCall) together.
Below, if we know we are in the type function, we also have a non-optional toolCall available, without having to repeat `value.choices[0].delta.tool_calls?.[0]` over and over again (which is esepcially error prone due to `?`, `!`, etc.)

CopilotKit/packages/shared/src/utils/decode-chat-completion.ts
```diff
@@ -35,7 +35,9 @@ export function decodeChatCompletion(
 ): ReadableStream<ChatCompletionEvent> {
   const reader = stream.getReader();
 
-  let mode: "function" | "message" | null = null;
+  type Mode = { type: "function"; function: ToolCallFunctionCall } | { type: "message" };
```
ataibarkai, line 38: same idea as above

CopilotKit/packages/shared/src/utils/parse-chat-completion.ts
```diff
@@ -1,5 +1,23 @@
...
+  // execute the function.
+  scope?: "client" | "server";
+}
+
+export interface ToolCallPayload {
+  index: number;
+  id?: string;
+  function: ToolCallFunctionCall;
+}
+
```
ataibarkai, line 20: these types are used in the associated value types of the 2 `Mode` types.
So we factor them out for convenience

PR #341, opened 2024-05-14
feat(backend): add final chat completion callback

arielweinberger:
This PR adds an optional `options` object to the `getResponse` method of `CopilotRuntime`.

Currently, it holds one optional property - `onFinalChatCompletion`, a callback that will be called at the end of a streaming response.

The `OnFinalChatCompletionCallback` type is generic, so individual adapters such as the `openai-adapter.ts` can define the return type (in this case, OpenAI's `ChatCompletion`).

This change is backwards compatible.

### DIFF
--- a/CopilotKit/packages/backend/src/lib/openai-adapter.ts
+++ b/CopilotKit/packages/backend/src/lib/openai-adapter.ts
@@ -5,7 +5,7 @@ import {
   OnFinalChatCompletionCallback,
 } from "../types/service-adapter";
 import { limitOpenAIMessagesToTokenCount, maxTokensForOpenAIModel } from "../utils/openai";
-import { Chat, ChatCompletion } from "openai/resources";
+import { ChatCompletion } from "openai/resources";
 import { ChatCompletionStream } from "openai/lib/ChatCompletionStream";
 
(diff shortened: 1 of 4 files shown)


CopilotKit/packages/backend/src/lib/openai-adapter.ts
```diff
@@ -49,18 +60,36 @@ export class OpenAIAdapter implements CopilotKitServiceAdapter {
...
 
-      const stream = this.openai.beta.chat.completions.stream({
-        model: this.model,
-        ...forwardedProps,
-        stream: true,
-        messages: messages as any,
-      });
+      let stream: ChatCompletionStream;
+
+      try {
```
arielweinberger, line 65: This was previously not caught and resulted in an unhandled rejection

mme: nice catch

mme:
Might be possible to implement by extending this?

```
export abstract class CopilotCloud {
  abstract checkGuardrailsInput(
    params: CloudCheckGuardrailsInputParams,
  ): Promise<CloudCheckGuardrailsInputResponse>;
}
```

PR #382, opened 2024-06-07
add support for Groq and Ollama models, update dependencies and refac…

calathus:
Groq and Ollama adapters have been incorporated alongside existing adapters such as OpenAPI and Google GenAI adapter. Their usage mirrors that of the existing adapters. Presently, this implementation does not support function calls, though it may be feasible to integrate this capability for Groq, as they appear to have recently introduced support for it.

A notable difference from the Google GenAI adapter is in the conversion of system messages; here, no additional strings are appended to the original message to avoid potentially affecting the language model's behavior. However, this scenario hasn't been thoroughly tested.

Parameters for the language model (LLM), such as temperature, are currently hard-coded. It's advisable to handle these parameters similarly to the Google GenAI adapter, although this hasn't been implemented yet.

Testing across various use cases would be beneficial. Additionally, the `node/index.ts` file has been modified to potentially accommodate other adapters, but since the original code only supported OpenAI, this adaptation might not be necessary.

### DIFF
--- a/CopilotKit/examples/next-openai/.env.local.example
+++ b/CopilotKit/examples/next-openai/.env.local.example
@@ -1,1 +1,5 @@
+LLM_PROVIDER='groq'
+
 OPENAI_API_KEY=xxxxxxx
+GOOGLE_API_KEY=xxxxxxx
+GROQ_API_KEY=xxxxxxx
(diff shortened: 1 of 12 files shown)


ataibarkai:
Great stuff!
Thanks for the contribution @calathus !

ataibarkai:
Hey @calathus just a heads up, we have a small CI error delaying this due to a recent refactor.
We love this contribution and will accept it in the next 1-2 days.
Thanks so much once again!

PR #395, opened 2024-06-17
Port tasks and textarea to the new client

mme:


mme:
@ataibarkai I ported this to the new client.

Textarea was not a straight port because it uses the `name` property which is not available in all adapters. Therefore I wrapped the example one shot messages in tags, for example:

```content: `<TextBeforeCursor>${editorState.textBeforeCursor}</TextBeforeCursor>`,```

In my initial tests this works, but I think this needs some more in depth testing...

PR #441, opened 2024-07-08
updated the unify adapter similar to the groq adapter using the openai client

vedpatwardhan:


vedpatwardhan:
Hey @ataibarkai, just made some changes to the unify adapter which should ideally resolve the issue with the demo. However, I haven't been able to test whether the adapter works myself so I can't say for sure. Is there a way in which we can test this before merging? Thanks 😄

ataibarkai (approved):

PR #442, opened 2024-07-08
replace nanoid

mme:


CopilotKit/packages/shared/src/utils/random-id.ts
```diff
@@ -0,0 +1,13 @@
+export function randomId(length = 16, alphabet = "abcdefghijklmnopqrstuvwxyz0123456789") {
```
arielweinberger: We can instead simply use `globalThis.crypto.randomUUID()`

This works in both web V8 and Node.

This way:

1. No need to maintain our own random ID function
2. No need to have `@copilotkit/shared` as a dependency solely for ID generation

arielweinberger (changes requested):

CopilotKit/packages/shared/src/utils/random-id.ts
mme: Done but we keep the `@copilotkit/shared` dependency

Issue #483, opened 2024-07-17
🐛 Bug: appendMessage in useCopilotChat overwrites history by a single new message item


aldrinjenson:
### ♻️ Reproduction Steps

Try running

```jsx
     appendMessage(
      new TextMessage({
        content: "Go to home page",
        role: Role.User,
      })
    );

    setTimeout(() => {
      appendMessage(
        new TextMessage({
          content: "Go back to where I was",
          role: Role.User,
        })
      );
    }, 1500);
  ```  
  
This will work. but in sidebar there'll be only the latest message. the first action has been performed and now the context seems to have been lost

<img width="479" alt="image" src="https://github.com/user-attachments/assets/910726b1-4586-437f-9e85-8078dd5f15a9">

  

### ✅ Expected Behavior

The appendMesasge function should take in a Textmessage that will append this to the existing conversation history.

Any new appendMessage call should not overwrite whatever message was given before.

### ❌ Actual Behavior

AppendMessage overwrites all previous chat + context info

### 𝌚 CopilotKit Version

```shell
├── @copilotkit/react-core@1.0.7
├── @copilotkit/react-textarea@1.0.7
├── @copilotkit/react-ui@1.0.7
├── @copilotkit/runtime@1.0.7
```

### 📄 Logs (Optional)

```shell
I think it's related but sometimes copilotkit loses out on certain capabilities that it has been given access to when switching routes while using appendMessage

to replicate, 
create 3 routes. say /home, /edit, /about
define some function in home. say toggle_theme
define copilotkit in root layout.

try running from /about route



     appendMessage(
      new TextMessage({
        content: "Go to editor page",
        role: Role.User,
      })
    );

    setTimeout(() => {
      appendMessage(
        new TextMessage({
          content: "Go to home route",
          role: Role.User,
        })
      );
    }, 1500);

    setTimeout(() => {
      appendMessage(
        new TextMessage({
          content: "Toggle the theme to dark mode",
          role: Role.User,
        })
      );
    }, 2500);
    
  
On running the above, copilotkit will reach the Home page and will say it doesn't have the capability to change theme. The history in sidebar will also be having just one message.
However if we're starting from home page(where toggle_theme) action is added, now the flow will work fine but a function defined in /about or /edit route will fail.



----


I think it's related but let me know if you folks want me to make a separate issue for this.
```

BenTaylorDev:
[...]

PR #494, opened 2024-07-22
fix(message-state): use previous message state when appending new messages

ranst91:
This draft adds a suggestion on how to fix the first issue described on https://github.com/CopilotKit/CopilotKit/issues/483

## The problem and solution:
When appending messages, there's an update of state going on. Several times for several reasons.
These updates are counting on messages being passed with prop drilling (sort of?) as the "previous messages to append to". I am not sure what causes the latest state to arrive late/not at all, but it doesn't matter, as a good practice of appending to state would be to use the set state method argument directly. It will hold the latest state each time.
So in order to append to the latest messages, it is better to simple `setState(oldMessages => [...oldMessage, newMessage])`.

## So, why is it a draft?
Because there are more problems :(
Maybe it wasn't visible due to how things got overridden with state, and maybe it's intentional(?), but with the nature of keeping the messages in the context, it causes a rerender of the whole tree. which then means, more calls to `appendMessage`. Let me explain.
In order to test my solution, I've done this:
```
  useEffect(() => {
    appendMessage(
      new TextMessage({
        content: "Message 1.",
        role: Role.User,
      })
    );
    setTimeout(() => {
      appendMessage(
        new TextMessage({
          content: "Message 2.",
          role: Role.User,
        })
      );
    }, 5000);
  }, [])
```
So it's an effect that should only run once, and append 2 messages over time.
The problem is, it runs twice, so we end up with double the messages, like so:
Screenshot 2024-07-23 at 1 35 27

I will be leaving this solution as a draft like this for now.
Feel free to take over it, or otherwise suggest any change in order to block the re-rendering.

## How to Reproduce:
1. Implement the solution in this draft
2. Go to `CopilotKit/packages/react-ui/src/components/chat/Chat.tsx` and add the above `useEffect`.
3. Add a console log inside the `useEffect` log and you'll see it's being called twice

### DIFF
--- a/CopilotKit/packages/react-core/src/hooks/use-chat.ts
+++ b/CopilotKit/packages/react-core/src/hooks/use-chat.ts
@@ -128,7 +128,7 @@ export function useChat(options: UseChatOptions): UseChatHelpers {
     const abortController = new AbortController();
     abortControllerRef.current = abortController;
 
-    setMessages([...previousMessages, ...newMessages]);
+    setMessages(prevMessages => [...prevMessages, ...newMessages])
 
     const systemMessage = makeSystemMessageCallback();
 
@@ -255,7 +255,7 @@ export function useChat(options: UseChatOptions): UseChatHelpers {
         }
 
         if (newMessages.length > 0) {
-          setMessages([...previousMessages, ...newMessages]);
+          setMessages(prevMessages => [...prevMessages, ...newMessages])
         }
       }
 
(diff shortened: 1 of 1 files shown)


arielweinberger:
@ranst91 Could the "it's running twice" issue be related to this?

https://stackoverflow.com/questions/60618844/react-hooks-useeffect-is-called-twice-even-if-an-empty-array-is-used-as-an-ar

ranst91:
You could say that. The accepted answer there a actually defines the problem. "Put your console in a use effect... ...If the log "i fire once" is triggered more than once it means..."

I can already guess this originates from the fact the answers are changed, but more checks are required there

ranst91:
@arielweinberger I can confirm that removing strict mode indeed "resolves" the issue.
It's not nice when developing, but depending on how pressing the issue is, we can push this towards merging or figure out a solution.
Your call ofc :)

arielweinberger:
We've fixed this issue already, and thanks for helping debug this!

PR #503, opened 2024-07-24
centering `CMD + K` popup

codingfarhan:
centered the popup in next-openai example app so that the popup doesn't run out of the screen.


mme:
Awesome work @codingfarhan!

I tested it and looks quite nice.

image

@ataibarkai can you also have a quick look at the changes?

MEGLDON83 (approved):

PR #540, opened 2024-08-20
fix(react-ui): pin headless-ui to v2.1.1

arielweinberger:
We've received reports of failed installations of CopilotKit, which is likely caused by a broken release of `headlessui/react@v2.1.2`. 

This PR temporarily pins the package to `v2.1.1`.

More information:
https://github.com/tailwindlabs/headlessui/issues/3430


Myahr208 (changes requested):
10INHOME[]()"[Screenshot_20240903_062125_Reddit.png](https://github.com/user-attachments/assets/d27dedae-2c91-42e6-89e1-a20d159ad994]"

PR #560, opened 2024-09-06
feat: inject minified css in bundle

arielweinberger:
# Description

- [x] removes the need to import `styles.css` manually
- [x] empty `styles.css` included in the build for backwards compatibility
- [x] uses tsup's `injectStyles` with `postcss` to bundle and minify the CSS, then inject it as a style tag
- [x] currently uses my fork of `tsup` where I added support for async function in `injectStyles` (must-have for postcss), a PR from my fork to the main library will follow shortly
- [x] remove material-ui, and use `react-icons` for icons (same icons as before)
- [x] remove unused `IncludedFilesPreview` component
- [x] updated docs

# How to test this

Check out the preview deployment below.
- To test the changes of `react-ui`, head over to `/presentation`
- To test the changes of `react-textarea`, head over to `/textarea`


mme (commented):
As far as I can tell, this looks good.

Just some pointers:
-  search the code for "@copilotkit/react-ui/styles.css" and the other css files, I'm pretty sure there are more occurences
- We should QA usage in different frameworks e.g. remix

arielweinberger:
> We should QA usage in different frameworks e.g. remix
I don't think there's anything that could be affected by frameworks:

```js
        if (globalThis.hasOwnProperty("document")) {
          const style = document?.createElement("style");
          style.innerHTML = '${resultCSS}';
          document?.head.appendChild(style);
        }
```

PR #572, opened 2024-09-16
Return current thread id from useCopilotChat

apperside:
Sometimes it may be useful to have access to the current thread id, so it can be used for any purpose (eg.: link specific threads to specific users of your platform).

This PR is preparatory for others I want to submit in the future in order to allow chat restoration (at the moment, every time I refresh the page, a new chat starts)


whitew1994WW:
This would be very useful to have - What is missing to get thread_id supported?

tylerslaton:
@whitew1994WW this PR is pretty old but we will have thread_ids working this week from our dev team! Going to close this in favor of that since this has gotten too stale.

Thank you for the contribution even if it didn't make it in this time.

whitew1994WW:
Ah brilliant, I have just been trawling through your latest release and can see the breadcrumbs... Can't wait

PR #710, opened 2024-10-05
fixed_header.css

Adity-av:
Now, no matter how much you scroll the page, the header will stay fixed at the top and will always be visible.


arielweinberger:
@mme is this something we want by default? or enabled by a prop/flag?

mme:
Hey @Adity-av thanks for your contribution! 

Could you add a screenshot of the header before/after the change?

Does this relate to a bug?

PR #717, opened 2024-10-07
docs-dark-mode

Mstr-Creta:
The dark mode styles will activate when the .dark class is applied to the html or body element.


arielweinberger:
The proper fix for this was applied in #726

PR #874, opened 2024-11-02
Opening Popover  on windows when we press ctrl + k in textarea

### LINKED ISSUE #400: CTRL + K does not work on Windows (CopilotTextarae and potentially others)
# Bug Report
**Describe the bug**
In `<CopilotTextarea>`, `cmd + k` works on a Mac to bring up the popover. However, on Windows, `ctrl + k` does not.
This could potentially be the case in other places where we use keyboard shortcuts.

**To Reproduce**
Steps to reproduce the behavior (the easier it is to replicate, the sooner it will get fixed):

1. On a Windows machine, go to https://crm-demo-delta.vercel.app/
2. Focus on the textarea field
3. Type `ctrl + k` to open up the CopilotKit popover

**Expected:** The popover opens as it does on a Mac when using `cmd + k`.
**Actual:** Nothing happens

# Potential Solution
We might want to look into dedicated shortcut handling libraries for JS/React that handle cross-platform key mapping differences. **We need to make sure it's fixed in other places that keyboard shortcuts.**

kilbha:
## What does this PR do?
This PR adds functionality to detect when a user presses Ctrl + K on a Windows machine and displays a popup if this specific key combination is used.

Implementation Details:
Key Press Detection: The code checks each keypress event to determine if the user is on a Windows OS and if the Ctrl + K key combination has been pressed.
Popup Trigger: If both conditions (Windows OS and Ctrl + K) are met, a popup is displayed.

## Related PRs and Issues
Fixes #400 
image

## Checklist

### DIFF
--- a/CopilotKit/packages/react-textarea/src/types/base/base-autosuggestions-config.tsx
+++ b/CopilotKit/packages/react-textarea/src/types/base/base-autosuggestions-config.tsx
@@ -76,10 +76,20 @@ const defaultShouldToggleHoveringEditorOnKeyPress = (
   event: React.KeyboardEvent<HTMLDivElement>,
   shortcut: string,
 ) => {
-  // if command-k, toggle the hovering editor
-  if (event.key === shortcut && event.metaKey) {
-    return true;
+  const isWindowsOs = navigator.userAgent.includes("Windows");
+
+  if (isWindowsOs) {
+    // if ctrl-k, toggle the hovering editor on windows
+    if (event.key === shortcut && event.ctrlKey) {
+      return true;
+    }
+  } else {
+    // if command-k, toggle the hovering editor on mac
+    if (event.key === shortcut && event.metaKey) {
+      return true;
+    }
   }
+
   return false;
 };
 


CopilotKit/packages/react-textarea/src/types/base/base-autosuggestions-config.tsx
```diff
@@ -76,10 +76,20 @@ const defaultShouldToggleHoveringEditorOnKeyPress = (
   event: React.KeyboardEvent<HTMLDivElement>,
   shortcut: string,
 ) => {
-  // if command-k, toggle the hovering editor
-  if (event.key === shortcut && event.metaKey) {
-    return true;
+  const isWindowsOs = navigator.userAgent.includes("Windows");
```
suhasdeshpande, line 79: Recommend refactoring the OS-specific shortcut logic to be more concise:

```javascript
// Check if it's the keyboard combo we want
const isMac = /Mac|iPod|iPhone|iPad/.test(navigator.platform);
const modifierKey = isMac ? event.metaKey : event.ctrlKey;

return event.key.toLowerCase() === shortcut.toLowerCase() && modifierKey;
```

Rationale:
- Eliminates duplicate code paths for Windows/Mac
- More maintainable with a single return statement
- Case-insensitive comparison prevents bugs
- Uses ternary operator for clean modifier key selection
- Still captures the same functionality in fewer lines

PR #899, opened 2024-11-11
fix(langchain-adapter): fix how tool calls are being separated and read

ranst91:
Our LangChain adapter was failing to work with a "one at a time" tools. Tools that should be used multiple times with one argument or so.
This PR introduces some changes to make this available, as well as a change to our demo to integrate such action

## Checklist


ranst91:
In favor of different solutions, closing this, as it became irrelevant

PR #920, opened 2024-11-14
feat: add coagent todo application

tylerslaton:
## What does this PR do?

This is built off the original quickstart's todo application which can be found below.

https://github.com/CopilotKit/example-todos-app

It expands on it by adding in CoAgent support. The agent it uses is able to manage todos and search the internet. As it does this the state of the LangGraph agent is represented in the UI.

This works both ways, editing the todos in the UI will update them in the Agent's state.

Not included with this is a hosted version or shot-out in the README.md, if those are desired they will be added in a follow-up.

## Checklist


tylerslaton:
Closing this for now since its essentially replaced by the #942 example. If we want to reopen this we can do so at a later date.

PR #976, opened 2024-11-26
Support async memory in LangGraph

mme:
Support async memory in LangGraph

## Checklist


sdk-python/pyproject.toml
```diff
@@ -26,4 +26,4 @@ requires = ["poetry-core"]
 build-backend = "poetry.core.masonry.api"
 
 [tool.poetry.scripts]
-demo = "copilotkit.demos.research_canvas.demo:main"
+demo = "copilotkit.demos.multi_agent.demo:main"
```
suhasdeshpande, line 29: this change is intentional?

mme: Thanks for having a look suhas! yes, it is intentional, in general we want this demo for testing purposes

PR #1090, opened 2024-12-12
docs: remove workaround docs for google gemini

ranst91:
Previously it was required to add a workaround to make google gemini work with CoAgents.
This is not needed anymore


Dianne087:
pls. screenshot...

PR #1092, opened 2024-12-12
docs: remove leftover of google gemini workaround

ranst91:
remove leftover of google gemini workaround


Dianne087:
okey sir  but pls give me time to check it, ok i promised i will check it
one by one where is the technical error... but for now, i cNt make, after
the batteryfull then i will do.. that is good for you sir?

PR #1158, opened 2024-12-23
Implementing Empty Service Adapter

MrCsabaToth:
@tylerslaton 

## What does this PR do?

* Implementing Empty Service Adapter (fixes #1157)
* Exporting the adapter for use
* Also exporting the Ollama experimental adapter
* Exporting CopilotRuntimeChatCompletionRequest, CopilotRuntimeChatCompletionResponse to make it easier for someone to implement their own custom adapter

## Related PRs and Issues

- https://github.com/CopilotKit/CopilotKit/issues/1157

## Checklist

- [x] I have read the [Contribution Guide](https://github.com/copilotkit/copilotkit/blob/master/CONTRIBUTING.md)
- [ ] If the PR changes or adds functionality, I have updated the relevant documentation

PR #1164, opened 2024-12-26
trigger e2e flow for empty adapter

ranst91:
As our pipeline is different for community PRs, I'm creating this one to have the right pipeline running for the following community PR:
https://github.com/CopilotKit/CopilotKit/pull/1158

The code is an exact copy and this PR here will be removed once all checks passed


MrCsabaToth:
I think prettier failed because these changes are not the latest. It needs to pull from the PR again possibly?

PR #1165, opened 2024-12-26
fix: use react 19 for all cpk packages

ranst91:
While we support React 19 already, we do not use it as a dev dependency when developing.
This PR makes the change


818Nawaf (approved):

Dianne087:
https://github.com/Dianne087|**|@rootdoc

PR #1177, opened 2025-01-02
fix: async agent state fetching

ranst91:
We're calling some methods when resolving langgraph agent communication. Some of these should be called async. This PR attempts a fix for that


suhasdeshpande:
Closing this PR. after thinking more about it, this should stay synchronous! we have remote actions construct to achieve this!

ranst91:
Closing. Will be available as coagents v0.3

PR #1178, opened 2025-01-02
feat: add temperature parameter support for LLM completions

suhasdeshpande:
## What does this PR do?

Adds temperature parameter forwarding to LLM chat completions API calls. Temperature controls randomness in model outputs, with lower values (0-1) producing more focused results.

### Example
Set the `temperature` field in CopilotKit component. 

```tsx
<CopilotKit forwardedParameters={{ temperature: 0 }}>
  <CopilotSidebar>
    <VacationList />
  </CopilotSidebar>
</CopilotKit>
```

## Checklist


CopilotKit/packages/react-textarea/src/hooks/make-autosuggestions-function/use-make-standard-autosuggestions-function.tsx
```diff
@@ -40,7 +39,7 @@ export function useMakeStandardAutosuggestionFunction(
     ...copilotApiConfig.headers,
     ...(publicApiKey ? { [COPILOT_CLOUD_PUBLIC_API_KEY_HEADER]: publicApiKey } : {}),
   };
-  const { maxTokens, stop } = apiConfig;
+  const { maxTokens, stop, temperature = 0.2 } = apiConfig;
```
suhasdeshpande: @mme tell me if this default value is a good idea if not, I can remove it!

CopilotKit/packages/react-core/src/lib/copilot-task.ts
```diff
@@ -149,6 +155,8 @@ export class CopilotTask<T = any> {
             requestType: CopilotRequestType.Task,
           },
           forwardedParameters: {
+            // if forwardedParameters is provided, use it
+            ...(this.forwardedParameters ? this.forwardedParameters : {}),
```
ranst91:
```suggestion
            ...(this.forwardedParameters ?? {}),
```

CopilotKit/packages/react-core/src/utils/extract.ts
```diff
@@ -122,6 +125,7 @@ export async function extract<const T extends Parameter[]>({
           requestType: requestType,
         },
         forwardedParameters: {
+          ...(forwardedParameters || {}),
```
ranst91:
```suggestion
          ...(forwardedParameters ?? {}),
```

CopilotKit/packages/react-textarea/src/hooks/make-autosuggestions-function/use-make-standard-autosuggestions-function.tsx
ranst91: Since temperature makes the llm "artistic" I'd use 0 as a default. I feel like it's the most expected option

ranst91 (approved):
Let's make sure it works with all llms (in terms of param names etc) 
LGTM

CopilotKit/packages/react-textarea/src/hooks/make-autosuggestions-function/use-make-standard-autosuggestions-function.tsx
suhasdeshpande:
```suggestion
  const { maxTokens, stop, temperature = 0 } = apiConfig;
```

mme (changes requested):
- Please add an example to the PR description how the user would configure temperature for chat, text completion and task. It's not clear to me if this is global or per type, since `SuggestionsApiConfig` is also changed.

- All adapters should set the temperature based on the `temperature` passed by the user except `LangChainAdapter` (see toolChoice as an example)

- I'm in favor of exposing temperature directly instead of exposing the whole forwardedParameters- for example `toolChoice` and `toolChoiceFunctionName` can never really be set by the user, only internally

mme (approved):

PR #1284, opened 2025-01-28
[CPK-1233] pass default headers from properties

suhasdeshpande:
## What does this PR do?

(Describe the changes introduced in this PR)

## Related PRs and Issues

- (Direct link to related PR or issue, if relevant)

## Checklist

- [ ] I have read the [Contribution Guide](https://github.com/copilotkit/copilotkit/blob/master/CONTRIBUTING.md)
- [ ] If the PR changes or adds functionality, I have updated the relevant documentation

PR #1292, opened 2025-01-30
fix: pass additional context to the langgraph state

suhasdeshpande:
## What does this PR do?

Adds the option to pass authorization tokens from the frontend by adding them to the `properties` property of `copilotRuntimeNextJSAppRouterEndpoint` and the `CopilotKit` components in the frontend. The LangGraph Client was previously being constructed without the option to pass auth headers, making it impossible to use the next custom auth workflow with [LangGraph platform](https://langchain-ai.github.io/langgraph/how-tos/auth/custom_auth/)

This is more of a hotfix, since the graphql properties object was the only convenient way to transfer auth headers all the way down the call stack to where the LangGraph Platform Clients were being constructed.

Note: the more consistent way to pass these auth tokens is by accessing the cookies in browser. CopilotKit seems to call `/assistants/search` to check for available agents prior to loading much else of the page. For some reason, this request seems to omit headers included in the NextJS request during initial page load, so retrieving the auth token from cookies is more consistent.

## Related PRs and Issues

- [Original PR from @suhasdeshpande based on our discussion](https://github.com/CopilotKit/CopilotKit/pull/1284)

## Checklist

### DIFF
--- a/CopilotKit/.changeset/silent-terms-float.md
+++ b/CopilotKit/.changeset/silent-terms-float.md
@@ -1,0 +1,5 @@
+---
+"@copilotkit/runtime": patch
+---
+
+- add headers handling to other LangGraphClients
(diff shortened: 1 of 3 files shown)


suhasdeshpande:
Clone of #1291

Thanks @mlsimon734 for contribution

PR #1409, opened 2025-02-21
fix: use memoization in useCoAgent internal functions

ranst91:
This PR internally improves the stability and predictability of functions that depends on state or a bound-to-change variables inside the `useCoAgent` implementation


apoorvkk:
Hey @ranst91 I am not sure if this is related so apologies if I commented in the wrong place but I noticed `useCoAgent` sometimes outputs an empty object while the agent is running:
image

The issue resolves itself once the chat completion is done but was wondering if this is expected behaviour. Any help will be greatly appreciated 🙏

PR #1617, opened 2025-04-02
feat: enable usage of python 3.13 with cpk

ranst91:
Lift off restriction for python 3.12, enabling 3.13


mlkauppila:
Hi, are there any estimates on when we can expect the latest versions of CopilotKit to be compatible with python 3.13?
The last officially compatible version is 0.1.39, but that is no longer compatible with latest versions of langgraph.

tylerslaton:
CopilotKit is now compatible will all versions of LangGraph and python!

PR #1639, opened 2025-04-08
fix(react-core): Enable multiple instances of renderAndWaitForRespons…

suhasdeshpande:
Replace single reference handling in useCopilotAction with a Map-based approach to fix issue where only the most recent instance of an action with renderAndWaitForResponse could be properly resolved through UI interaction.

The fix allows multiple instances of the same action to be triggered in sequence while maintaining their own independent resolve/reject functions, enabling users to interact with any instance regardless of creation order.

This resolves the bug where users requesting multiple operations of the same type
(e.g., "generate 3 images") could only complete or cancel the most recent one.

### DIFF
--- a/CopilotKit/packages/react-core/src/hooks/use-copilot-action.ts
+++ b/CopilotKit/packages/react-core/src/hooks/use-copilot-action.ts
@@ -130,7 +130,7 @@
  *
  * This hooks enables you to dynamically generate UI elements and render them in the copilot chat. For more information, check out the [Generative UI](/guides/generative-ui) page.
  */
-import { Parameter, randomId } from "@copilotkit/shared";
+import { Parameter, randomId, MappedParameterTypes } from "@copilotkit/shared";
 import { createElement, Fragment, useEffect, useRef } from "react";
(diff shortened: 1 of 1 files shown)


CopilotKit/packages/react-core/src/hooks/use-copilot-action.ts
```diff
@@ -158,7 +158,8 @@ export function useCopilotAction<const T extends Parameter[] | [] = []>(
 ): void {
   const { setAction, removeAction, actions, chatComponentsCache } = useCopilotContext();
   const idRef = useRef<string>(randomId());
-  const renderAndWaitRef = useRef<RenderAndWaitForResponse | null>(null);
+  // Replace single ref with Map to track multiple instances
+  const renderAndWaitMapRef = useRef<Map<string, RenderAndWaitForResponse>>(new Map());
```
ranst91: "// Replace single ref with Map to track multiple instances" suggest a diff from previous implementation. Instead, let's just explain the situation without previous implementation context

ranst91 (approved):

PR #1712, opened 2025-04-23
Suhas/interface change

suhasdeshpande:
## What does this PR do?

refactor: rename mcpEndpoints to mcpServers for naming consistency

This commit renames the `mcpEndpoints` property to `mcpServers` across the codebase for improved naming consistency.

Key changes:
- Remove `mcpEndpoints` from CopilotKitProps in favor of using the hook-based approach 
- Add `mcpServers` to CopilotApiConfig interface with proper typing
- Update the `useCopilotChat` hook to expose `mcpServers` and `setMcpServers`
- Update CopilotRuntime to use `mcpServers` instead of `mcpEndpoints`
- Maintain backward compatibility in runtime by checking for both property names
- Update property merging logic in use-chat.ts to use the new property name

This change improves the API consistency while maintaining the same functionality.
MCP servers should now be managed via the hook:
```tsx
const { mcpServers, setMcpServers } = useCopilotChat();
```
instead of through the component props.

## Related PRs and Issues

## Checklist

### DIFF
--- a/CopilotKit/packages/react-core/src/components/copilot-provider/copilotkit-props.tsx
+++ b/CopilotKit/packages/react-core/src/components/copilot-provider/copilotkit-props.tsx
@@ -1,7 +1,6 @@
 import { ForwardedParametersInput } from "@copilotkit/runtime-client-gql";
 import { ReactNode } from "react";
 import { AuthState } from "../../context/copilot-context";
-
 /**
  * Props for CopilotKit.
  */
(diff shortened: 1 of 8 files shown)


examples/coagents-starter/ui/app/api/copilotkit/route.ts
```diff
@@ -1,30 +1,32 @@
...
+        url: config.endpoint,
+        headers: config.apiKey
+          ? { Authorization: `Bearer ${config.apiKey}` }
+          : undefined,
+      },
+    });
+  },
 });
 
 export const POST = async (req: NextRequest) => {
```
tylerslaton: @suhasdeshpande this look like a product of testing out this PR. Is that right? My assumption here is we should remove the changes to this file.

tylerslaton (changes requested):

PR #1716, opened 2025-04-24
fix: parse mcp tool params bug 

v1xingyue:
Sometimes mcp server .tools returnd list does not contains schema field. But can get it from .parameters?.jsonSchema get it.


tylerslaton:
Hi @v1xingyue! Thanks so much for the contribution. We're giving this a look on our side and will follow-up with a review.

PR #1743, opened 2025-05-01
udpates crewai to 0.118.0

suhasdeshpande:
## What does this PR do?

## Related PRs and Issues

## Checklist


examples/coagents-starter-crewai-flows/agent-py/sample_agent/agent.py
```diff
@@ -362,6 +363,38 @@ async def chat(self):
         """
         system_prompt = f"You are a helpful assistant. Talk in {self.state.language}."
 
+        llm = LLM(
```
mme: Should this be removed / converted to something else? (since we are not using their LLM class)

mme (changes requested):
Would be good to see how the current user code will look like (i.e. without the LLM class)

Couldn't find it in the examples

examples/coagents-starter-crewai-flows/agent-py/sample_agent/agent.py
suhasdeshpande: yes, removing it is right thing

PR #1782, opened 2025-05-09
chore: refactor and cleanup pre langgraph stream actions

ranst91:
The underlaying layer of how we stream langgraph events is about to change.
This PR introduces small refactors which will makes the code easier to handle when this change happens


tylerslaton:
Hey @ranst91 if this should be reopened, just feel free. Closing for now.

PR #1794, opened 2025-05-11
react-textarea: Fixed useless div in Body

kachurun:
## What does this PR do?

This PR primarily fixes the issue where the react-textarea component was injecting **an unnecessary div into the body**.

It also addresses several bugs related to window positioning — the position was being calculated before the content inside the window was rendered, which in some cases caused the window to overlap the content.

This PR can be merged, but I don’t consider the work complete — in my opinion, the component needs a full rewrite.

Here are the main issues still in the code and/or logic:

**HoveringToolbar is bundled into react-textarea by default**
It shouldn’t be. I expect to control this behavior myself and insert the HoveringToolbar wherever I need it. For example, I might want to render a button inside the input that triggers the window, or even trigger the UI with a hotkey from anywhere in the app — not just when a textarea is focused.

**No proper positioning logic**
The window lacks flexibility in how it’s positioned. I generally avoid extra dependencies, but in this case, [Floating UI](https://floating-ui.com/docs/useFloating) would be very helpful. For example, Radix UI uses it under the hood.

**Magic numbers in the code**
Take verticalOffsetFromCorner, for example — it’s not configurable and doesn’t really make sense, so I just zeroed it out. You can achieve the same effect using padding on the root element (which already has 0.5em on all sides).

**Breaks in complex layouts**
The window cannot position itself properly if it’s rendered inside an iframe or if the target element lives inside a custom scroll container.

And that’s not even the full list. Fixing all of this will require a serious rewrite, which would be a breaking change.
I’d be happy to take that on if I get the green light.

### DIFF
--- a/CopilotKit/packages/react-textarea/src/components/hovering-toolbar/hovering-toolbar.tsx
+++ b/CopilotKit/packages/react-textarea/src/components/hovering-toolbar/hovering-toolbar.tsx
@@ -1,6 +1,6 @@
-import { useEffect, useRef, useState } from "react";
+import { useEffect, useLayoutEffect, useRef, useState } from "react";
 import { Editor, Location, Transforms } from "slate";
-import { useSlate, useSlateSelection } from "slate-react";
+import { useSlate, useSlateSelection, ReactEditor } from "slate-react";
 import { HoveringInsertionPromptBox } from "./text-insertion-prompt-box";
 import { Menu, Portal } from "./hovering-toolbar-components";
 import { useHoveringEditorContext } from "./hovering-editor-provider";
@@ -24,14 +24,16 @@ export const HoveringToolbar = (props: HoveringToolbarProps) => {
   const editor = useSlate();
   const selection = useSlateSelection();
   const { isDisplayed, setIsDisplayed } = useHoveringEditorContext();
-
+  
   // only render on client
   const [isClient, setIsClient] = useState(false);
   useEffect(() => {
     setIsClient(true);
   }, []);
+  
+  const isShown = isClient && isDisplayed && selection;
 
-  useEffect(() => {
+  useLayoutEffect(() => {
     const el = ref.current;
     const { selection } = editor;
 
@@ -51,7 +53,7 @@ export const HoveringToolbar = (props: HoveringToolbarProps) => {
 
     const domRange = domSelection.getRangeAt(0);
     const rect = domRange.getBoundingClientRect();
-
+    
     // We use window = (0,0,0,0) as a signal that the selection is not in the original copilot-textarea,
     // but inside the hovering window.
     //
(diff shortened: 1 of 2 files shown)


CopilotKit/packages/react-textarea/src/components/hovering-toolbar/hovering-toolbar.tsx
```diff
@@ -51,57 +53,82 @@ export const HoveringToolbar = (props: HoveringToolbarProps) => {
...
     // but inside the hovering window.
     //
     // in such case, we simply do nothing.
     if (rect.top === 0 && rect.left === 0 && rect.width === 0 && rect.height === 0) {
       return;
     }
-    const minGapFromEdge = 60;
-    const verticalOffsetFromCorner = 35;
-    const horizontalOffsetFromCorner = 15;
-    let top = rect.top + window.scrollY - el.offsetHeight + verticalOffsetFromCorner;
```
kachurun, line 65: el.offsetHeight here doesn't make any sense because it renders as an empty div, so its height will be equal to 1em by default (just block paddings). Also, it's weird to position it 35px from the top side of the selection. It must use the bottom if we're trying to position the modal below the selection.

CopilotKit/packages/react-textarea/src/components/hovering-toolbar/hovering-toolbar.tsx
```diff
@@ -51,57 +53,82 @@ export const HoveringToolbar = (props: HoveringToolbarProps) => {
...
-    // make sure top is in the viewport and not too close to the edge
-    if (top < minGapFromEdge) {
-      top = rect.bottom + window.scrollY + minGapFromEdge;
```
kachurun, line 68: This case is possible only in a crazy scenario when the textarea is positioned above the start of the document. I don't think it's worth caring about this case. Fix me if I'm wrong and I rewrite it like `let top = Math.max(verticalOffsetFromCorner, rect.bottom + window.scrollY + verticalOffsetFromCorner);`

CopilotKit/packages/react-textarea/src/components/hovering-toolbar/hovering-toolbar.tsx
```diff
@@ -51,57 +53,82 @@ export const HoveringToolbar = (props: HoveringToolbarProps) => {
...
     // We use window = (0,0,0,0) as a signal that the selection is not in the original copilot-textarea,
     // but inside the hovering window.
     //
     // in such case, we simply do nothing.
     if (rect.top === 0 && rect.left === 0 && rect.width === 0 && rect.height === 0) {
       return;
     }
-    const minGapFromEdge = 60;
-    const verticalOffsetFromCorner = 35;
-    const horizontalOffsetFromCorner = 15;
```
kachurun, line 64: Magic numbers are uncontrolled by the user. And what is "edge" and what is "corner" is pretty unclear. I reset it but left the variables to make future improvements easier.

CopilotKit/packages/react-textarea/src/components/hovering-toolbar/hovering-toolbar.tsx
```diff
@@ -51,57 +53,82 @@ export const HoveringToolbar = (props: HoveringToolbarProps) => {
...
+    else if (left + el.offsetWidth > window.innerWidth - horizontalOffsetFromCorner) {
+      left = window.innerWidth - el.offsetWidth - horizontalOffsetFromCorner;
     }
 
     el.style.opacity = "1";
     el.style.position = "absolute";
 
     el.style.top = `${top}px`;
     el.style.left = `${left}px`;
-  });
```
kachurun, line 87: There's no reason to recalculate it on every selection change (e.g., input in the textarea).

CopilotKit/packages/react-textarea/src/components/hovering-toolbar/text-insertion-prompt-box/hovering-insertion-prompt-box.tsx
```diff
@@ -8,7 +8,6 @@ export interface Props {
   editorState: EditingEditorState;
   apiConfig: InsertionEditorApiConfig;
   performInsertion: (insertedText: string) => void;
-  closeWindow: () => void;
```
kachurun, line 11: This thing isn't used anywhere.

CopilotKit/packages/react-textarea/src/components/hovering-toolbar/hovering-toolbar.tsx
```diff
@@ -94,12 +101,13 @@ export const HoveringToolbar = (props: HoveringToolbarProps) => {
...
     return () => {
       document.removeEventListener("mousedown", handleClickOutside);
     };
-  }, [ref, setIsDisplayed]);
+  }, [ref, isShown, setIsDisplayed]);
 
-  if (!isClient) {
+  if (!isShown) {
     return null;
   }
```
kachurun, line 112: No need to portal an empty `<Menu>` until we need to show the modal.

suhasdeshpande:
@kachurun Thanks for opening this. If you are, let's do all the changes, if you prefer multiple PRs, we can do that as well. We have not seen many instances of textarea component being used and hece it was not touched in long time. We would be more than happy to take this contribution. :)

kachurun:
@suhasdeshpande Cool, I’ll make the next changes, but maybe not within the textarea component.
I think it would be better to create a simpler "contenteditable" basic component and separate the HoveringToolbar from the textarea, untangling it from some dependencies (like its dependency on Slate contexts). This is a pretty big task and should be done in separate PRs and discussed with the team first. After that, the textarea is basically just a styled contenteditable component, maybe with the HoveringToolbar inside for compatibility (though I’d prefer not to do that).

So, for now, I think we should merge this PR to fix current minor issue and not expect all these changes in the upcoming commits to this one.

PR #1861, opened 2025-05-22
fix: synchronously execute renderAndWaitForResponse

tylerslaton:
Previously, it was impossible to execute multiple human-in-the-loop (renderAndWaitForResponse) calls in a row. Ultimately this was due to an issue with how CopilotKit was rendering the updates when multiple renderAndWaitForResponse actions appeared on screen due to a reference based approach.

With this change, actions will be executed in a synchronous way appearing almost queue like. This works with any combination of action given much more freedom when asking for user input.

## Checklist

### DIFF
--- a/CopilotKit/packages/react-core/src/hooks/use-copilot-action.ts
+++ b/CopilotKit/packages/react-core/src/hooks/use-copilot-action.ts
@@ -159,10 +159,13 @@ export function useCopilotAction<const T extends Parameter[] | [] = []>(
   const { setAction, removeAction, actions, chatComponentsCache } = useCopilotContext();
   const idRef = useRef<string>(randomId());
   const renderAndWaitRef = useRef<RenderAndWaitForResponse | null>(null);
+  const activatingMessageIdRef = useRef<string | null>(null);
   const { addToast } = useToast();
 
   // clone the action to avoid mutating the original object
   action = { ...action };
+
+  // const { currentlyActivatingHitlActionMessageIdRef } = useCopilotContext() as any; // <-- REMOVE THIS FOR NOW
 
   // If the developer provides a renderAndWaitForResponse function, we transform the action
   // to use a promise internally, so that we can treat it like a normal action.
@@ -172,12 +175,21 @@ export function useCopilotAction<const T extends Parameter[] | [] = []>(
     // check if renderAndWaitForResponse is set
     (action.renderAndWait || action.renderAndWaitForResponse)
   ) {
+    (action as any)._isRenderAndWait = true; // Internal flag to identify this action type later
     const renderAndWait = action.renderAndWait || action.renderAndWaitForResponse;
     // remove the renderAndWait function from the action
     action.renderAndWait = undefined;
     action.renderAndWaitForResponse = undefined;
+
+    // Add a method for use-chat.ts to set the activating message ID.
+    // This helps correlate the action instance with the message being processed by use-chat.
+    (action as any)._setActivatingMessageId = (id: string | null) => {
+      activatingMessageIdRef.current = id;
+    };
+
     // add a handler that will be called when the action is executed
     action.handler = useAsyncCallback(async () => {
+      const currentActivatingId = activatingMessageIdRef.current;
       // we create a new promise when the handler is called
       let resolve: (result: any) => void;
       let reject: (error: any) => void;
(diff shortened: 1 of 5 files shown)


examples/e2e/tests/multi-action.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+          await page.goto(`${config.url}/multi${variant.queryParams}`);
+          await page.getByRole('button', { name: 'Multiple of the same action' }).click();
+          // It's better to locate by more specific selectors if possible,
+          // but using .first() for "Continue" implies sequence is important.
+          await page.getByRole('button', { name: 'Continue' }).first().click();
+          await page.waitForTimeout(INTERACTION_DELAY); // Consider replacing with waitForResponse or specific UI change
+          await page.getByRole('button', { name: 'Continue' }).first().click();
+          await page.waitForTimeout(INTERACTION_DELAY);
+          await page.getByRole('button', { name: 'Continue' }).first().click();
+          // Asserting on body might be too broad; prefer specific message containers
```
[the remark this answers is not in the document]
ranst91: Let's ignore the suggestion to not target valid HTML features and instead look for synthetic attributes.
Targeting by role is a "free accessibility test". If we have the chance to do it, let's do it.

YassineRjl:
+1, spent hours rebuild my code to account for rerendering. Hope this gets merged soon.

suhasdeshpande:
> +1, spent hours rebuild my code to account for rerendering. Hope this gets merged soon.

Sorry to hear that. We have a [alpha release](https://github.com/CopilotKit/CopilotKit/actions/runs/15196358244/job/42741310727#step:7:31) of this change if you would like to give it a try 

```
npm i @copilotkit/react-ui@0.0.0-tyler-fix-multiple-hitl-calls-20250522204017
```

YassineRjl:
> > +1, spent hours rebuild my code to account for rerendering. Hope this gets merged soon.
> 
> Sorry to hear that. We have a [alpha release](https://github.com/CopilotKit/CopilotKit/actions/runs/15196358244/job/42741310727#step:7:31) of this change if you would like to give it a try
> 
> ```
> npm i @copilotkit/react-ui@0.0.0-tyler-fix-multiple-hitl-calls-20250522204017
> ```

Massive help, thanks a lot. Now the component doesn't unmount and remount when the chat's isLoading goes from true to false as well, not sure if the fix was intended for this behavior too but it fixed it as well which is nice. 
Dealing with this before this update was a nightmare honestly. thank you guys. Looking forward to getting this merged in the next release. for now we will pin our production version to this alpha release. 
Beautiful day.

Edit @suhasdeshpande @tylerslaton :

I retract what i said previously. The PR does solve the tools being umounted and remounted, but not completely, there is 1 last update when the chat goes from isLoading (true) to isLoading(false) and sometimes even randomly before that, the component gets umounted and remounted. I solved it by setting a key hashed from the arguments, hacky solution that works for now:



```ts

const getOrCreateStableKey = (
  componentName: string,
  args: any,
  status?: string
) => {
  if (status === "inProgress") {
    return undefined;
  }

  try {
    // Create a unique identifier for this component instance
    const argsString = JSON.stringify(args, Object.keys(args).sort());
    const hash = stringToHash(argsString);
    const componentId = `${componentName}-${hash}`;

    // If we already have a key for this component, reuse it
    if (componentKeysRef.has(componentId)) {
      return componentKeysRef.get(componentId);
    }

    // Generate new stable key and store it
    const newKey = `${componentName}-${Date.now()}-${Math.random()
      .toString(36)
      .substring(2, 8)}`;
    componentKeysRef.set(componentId, newKey);

    // Clean up old keys periodically to prevent memory leaks
    if (componentKeysRef.size > 200) {
      const keysToDelete = Array.from(componentKeysRef.keys()).slice(0, 50);
      keysToDelete.forEach((key) => componentKeysRef.delete(key));
    }

    return newKey;
  } catch (error) {
    console.error("Error generating stable key:", error);
    // Fallback to a simple key if there's any error
    return `${componentName}-fallback-${Date.now()}-${Math.random()
      .toString(36)
      .substring(2, 8)}`;
  }
};

useCopilotAction({
    name: "full_ebook_writer",
    available: "disabled",
    description:
      "Writes a complete structured ebook",
    parameters: [
     // truncated for brevity
    ],
    renderAndWaitForResponse: (props) => {
      return (
        <FullEbooksWriter
          key={getOrCreateStableKey(
            "full_ebook_writer",
            props.args,
            props.status
          )}
          args={props.args}
          state={state}
          status={props.status}
          result={props.result}
          isLoading={isLoading}
          saveModelFeedbackResolve={props.respond}
        />
      );
    },
  });
  ```

File: CopilotKit/packages/react-core/src/hooks/use-copilot-action.ts [hunk 2]
```diff
@@ -159,11 +159,14 @@ export function useCopilotAction<const T extends Parameter[] | [] = []>(
   const { setAction, removeAction, actions, chatComponentsCache } = useCopilotContext();
   const idRef = useRef<string>(randomId());
   const renderAndWaitRef = useRef<RenderAndWaitForResponse | null>(null);
+  const activatingMessageIdRef = useRef<string | null>(null);
   const { addToast } = useToast();
 
   // clone the action to avoid mutating the original object
   action = { ...action };
 
+  // const { currentlyActivatingHitlActionMessageIdRef } = useCopilotContext() as any; // <-- REMOVE THIS FOR NOW
```
suhasdeshpande, line 168: can we remove it completely and not have commented code?

CopilotKit/packages/react-core/src/hooks/use-copilot-action.ts
```diff
@@ -172,39 +175,77 @@ export function useCopilotAction<const T extends Parameter[] | [] = []>(
     // check if renderAndWaitForResponse is set
     (action.renderAndWait || action.renderAndWaitForResponse)
   ) {
+    (action as any)._isRenderAndWait = true; // Internal flag to identify this action type later
```
suhasdeshpande, line 178: `as any` seems bit dangerous. Do we need any checks before we assign the internal flag so we don't have weird side-effects later?

PR #1913, opened 2025-06-02
Cpk 2443 travel demo

orca-copilotkit:
## What does this PR do?

This PR addresses the demo ticket #2443. This allows the addition of checkboxes and updated dropdown UI


tylerslaton:
Hey @orca-copilotkit I gave this a test. Some initial feedback.

1. If I select a collection of items, I notice that those selections are not respected by the LLM. For example, if I select a subset of suggestions then it will still add all of them.
2. When the agent is editing a trip, we should only include the delta of suggestions to include. I.E if I have a trip with 3 places already selected, then the suggestion from the agent should only include what is new.

Issue #1991, opened 2025-06-16
🐛 Bug: Error Resume LangGraph


the-waste-land:
### ♻️ Reproduction Steps

1. LangGraph Interrupt and CopilotKit show Component by useLangGraphInterrupt hook:
```TSX
useLangGraphInterrupt({
    enabled: ({ eventValue }) => {
      // 返回true表示这个interrupt应该处理这个事件
      return eventValue.type === 'approval' || eventValue.type === 'choice' // 只处理这两种类型
    },
    render: ({ event, resolve }) => {
      // 根据事件类型返回不同的组件
      if (event.value.type === 'approval') {
        return <ApproveComponent content={event.value.content} onAnswer={(answer: any) => resolve(answer)} />
      } else if (event.value.type === 'choice') {
        return (
          <ChoiceComponent
            message={event.value.message}
            options={event.value.options}
            onAnswer={(answer: any) => resolve(answer)}
          />
        )
      }

      // 默认情况下返回一个空的div，避免返回null
      return <div style={{ display: 'none' }}></div>
    },
  })
```

<img width="691" alt="Image" src="https://github.com/user-attachments/assets/123a98a6-b5ec-412b-9f41-5382a0625fd8" />

2. Close the current browse tab and open a new tab
3. Open CopilotKit and ask a new question, langGraph throws an error:
```
Traceback (most recent call last):
  File "/Users/liweixin/code/langGraph/langgraph-server/.venv/lib/python3.12/site-packages/langgraph_api/worker.py", line 148, in worker
    await asyncio.wait_for(consume(stream, run_id), BG_JOB_TIMEOUT_SECS)
  File "/Users/liweixin/.local/share/uv/python/cpython-3.12.9-macos-aarch64-none/lib/python3.12/asyncio/tasks.py", line 520, in wait_for
    return await fut
           ^^^^^^^^^
  File "/Users/liweixin/code/langGraph/langgraph-server/.venv/lib/python3.12/site-packages/langgraph_api/stream.py", line 286, in consume
    raise e from g
  File "/Users/liweixin/code/langGraph/langgraph-server/.venv/lib/python3.12/site-packages/langgraph_api/stream.py", line 275, in consume
    async for mode, payload in stream:
  File "/Users/liweixin/code/langGraph/langgraph-server/.venv/lib/python3.12/site-packages/langgraph_api/stream.py", line 148, in astream_state
    event = await wait_if_not_done(anext(stream, sentinel), done)
[...]

PR #2026, opened 2025-06-20
fix: resolve activeInterruptEvent global state pollution across threads

the-waste-land:
## What does this PR do?

This PR fixes a global state pollution issue in the LangGraph integration where the `activeInterruptEvent` variable was shared across all threads, causing concurrent requests from different browser tabs to interfere with each other.

**Changes:**
- Convert global `activeInterruptEvent` to thread-specific state management using `Map<threadId, boolean>`
- Prevent incorrect `command.resume` being added to unrelated thread requests
- Maintain proper interrupt handling within the same thread while isolating different threads

**Problem solved:**
- When one browser tab triggered a LangGraph interrupt but didn't respond, subsequent requests from other tabs would incorrectly include `command.resume` in the payload
- This caused `'NoneType' object is not iterable` errors on the LangGraph server side
- Multiple concurrent sessions now work independently without state pollution

## Related PRs and Issues

- Fixes issue with concurrent browser sessions causing LangGraph errors due to global state pollution

## Checklist

### DIFF
--- a/CopilotKit/packages/runtime/src/lib/runtime/remote-lg-action.ts
+++ b/CopilotKit/packages/runtime/src/lib/runtime/remote-lg-action.ts
@@ -84,7 +84,8 @@ type SchemaKeys = {
   config: string[] | null;
 } | null;
 
-let activeInterruptEvent = false;
+// use map to store the interrupt state of each thread to avoid global state pollution
+const threadInterruptStates = new Map<string, boolean>();
 
 export async function execute(args: ExecutionArgs): Promise<ReadableStream<Uint8Array>> {
   return new ReadableStream({
@@ -179,6 +180,9 @@ async function streamEvents(controller: ReadableStreamDefaultController, args: E
       `Cannot use the threadId ${threadId} with LangGraph Platform. Must be a valid UUID.`,
     );
   }
+
+  // get or initialize the interrupt state of this thread
+  let activeInterruptEvent = threadInterruptStates.get(threadId) || false;
 
   let wasInitiatedWithExistingThread = true;
   try {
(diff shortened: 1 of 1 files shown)


the-waste-land:
Fix the issue: https://github.com/CopilotKit/CopilotKit/issues/1991

ranst91:
Great contribution! Much appreciated!

the-waste-land:
> Great contribution! Much appreciated!

I found there's still an issue when I have two Node.js instances running the CopilotKit runtime. 
The CopilotKit runtime should be stateless. 
Maybe we should let the browser carry the activeInterruptEvent state, or use Redis to store it. 
What do you think?

the-waste-land:
Maybe we could get the activeInterruptEvent state from the LangGraph thread state.

ranst91:
@the-waste-land I wouldn't worry about this too much.
We are slowly moving towards AG-UI LangGraph agent. You can find the core implementation here:
https://github.com/ag-ui-protocol/ag-ui/blob/main/typescript-sdk/integrations/langgraph/src/index.ts

As you can see there, we are not saving the interrupt anywhere, but rather fetching it before calling the stream. This way we can say "hold on, can we even stream or are there any pending interrupt events".

This approach is not exclusive to AG-UI, so it can also be implemented here. It's indeed what you mentioned as "get the activeInterruptEvent state from the LangGraph thread state"

If you see issues with the approach you have implemented, I can look into syncing the two implementations and bringing this concept to "remote-lg-action"

ranst91:
https://github.com/CopilotKit/CopilotKit/pull/2031 I have created this which follows the AGUI implementation

the-waste-land:
> @the-waste-land I wouldn't worry about this too much. We are slowly moving towards AG-UI LangGraph agent. You can find the core implementation here: https://github.com/ag-ui-protocol/ag-ui/blob/main/typescript-sdk/integrations/langgraph/src/index.ts
> 
> As you can see there, we are not saving the interrupt anywhere, but rather fetching it before calling the stream. This way we can say "hold on, can we even stream or are there any pending interrupt events".
> 
> This approach is not exclusive to AG-UI, so it can also be implemented here. It's indeed what you mentioned as "get the activeInterruptEvent state from the LangGraph thread state"
> 
> If you see issues with the approach you have implemented, I can look into syncing the two implementations and bringing this concept to "remote-lg-action"

Great! Looking forward to the new version.

PR #2031, opened 2025-06-20
fix: use active interrupt from thread instead of saving to global state

ranst91:
See https://github.com/CopilotKit/CopilotKit/pull/2026 for discussions & final resolution


the-waste-land:
@ranst91  Could you please release a version to fix this issue? Thank you!

ranst91:
@the-waste-land There would probably be a release today.
Meanwhile, you can already use the latest release candidate, which includes the fix:
`1.9.2-next.8`

PR #2089, opened 2025-07-02
fix(runtime): address issue with agent discovery being too langgraph …

tylerslaton:
## What does this PR do?
Fixing an issue that broke agent discovery for non-langgraph ag-ui agents.

## Checklist


CopilotKit/packages/runtime/src/lib/runtime/copilot-runtime.ts
```diff
@@ -975,46 +975,57 @@ please use an LLM adapter instead.`,
...
     }
     let state: any = {};
-    try {
-      state = (await client.threads.getState(threadId)).values as any;
-    } catch (error) {
-      // All errors from agent state loading are user configuration issues
-      const errorMessage = error instanceof Error ? error.message : String(error);
+    if (client) {
+      try {
+        state = (await client.threads.getState(threadId)).values as any;
```
ranst91, line 985: why do we need to fetch the state? can't we go for a full ignore solution?

Instead let's just return an empty state

PR #2096, opened 2025-07-03
fix: fix how agui agents are mapped to copilotkit agents

ranst91:
This PR patches the AGUI agents discovery which errors out for different agents (such as llamaindex)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the accuracy of agent mapping, ensuring agents are listed with correct names and IDs in the interface.

### DIFF
--- a/CopilotKit/packages/runtime/src/lib/runtime/copilot-runtime.ts
+++ b/CopilotKit/packages/runtime/src/lib/runtime/copilot-runtime.ts
@@ -884,10 +884,10 @@ please use an LLM adapter instead.`,
   }
 
   discoverAgentsFromAgui(): Agent[] {
-    return Object.values(this.agents ?? []).map((agent: LangGraphAgent) => ({
-      name: agent.agentName,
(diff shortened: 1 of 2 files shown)


CopilotKit/packages/runtime/src/lib/runtime/copilot-runtime.ts
```diff
@@ -884,8 +884,8 @@ please use an LLM adapter instead.`,
   }
 
   discoverAgentsFromAgui(): Agent[] {
-    return Object.values(this.agents ?? []).map((agent: LangGraphAgent) => ({
-      name: agent.agentName,
+    return Object.entries(this.agents ?? []).map(([key, agent]: [string, AbstractAgent]) => ({
```
tylerslaton: Do we also need to patch the thread retrieval mechanism like in https://github.com/CopilotKit/CopilotKit/pull/2089? When I tested I noticed things would break on start-up and execution due to the client not being defined for non-LG systems.

PR #2058, opened 2025-06-26
Fix extract toolParameters in extractParametersFromSchema when using mcp

jarvisaoieong:
## What does this PR do?

Fix the toolParameters extract problem when using mcp

## Related PRs and Issues

- (Direct link to related PR or issue, if relevant)

## Checklist

- [x] I have read the [Contribution Guide](https://github.com/copilotkit/copilotkit/blob/master/CONTRIBUTING.md)
- [x] If the PR changes or adds functionality, I have updated the relevant documentation

PR #2100, opened 2025-07-03
Fix extract toolParameters in extractParametersFromSchema

suhasdeshpande:
## What does this PR do?

Fix the toolParameters extract problem when using mcp

## Related PRs and Issues

## Checklist

## Summary by CodeRabbit

* **Refactor**
  * Improved internal handling of parameter extraction from schemas to enhance data processing accuracy without affecting user experience.

Clone of #2058

### DIFF
--- a/CopilotKit/packages/runtime/src/lib/runtime/mcp-tools-utils.ts
+++ b/CopilotKit/packages/runtime/src/lib/runtime/mcp-tools-utils.ts
@@ -51,7 +51,7 @@ export function extractParametersFromSchema(
       ? (toolOrSchema as MCPTool).schema
       : (toolOrSchema as MCPTool["schema"]);
 
-  const toolParameters = schema?.parameters || schema?.parameters?.jsonSchema;
+  const toolParameters = schema?.parameters?.jsonSchema || schema?.parameters;
   const properties = toolParameters?.properties;
   const requiredParams = new Set(toolParameters?.required || []);
 


CopilotKit/packages/runtime/src/lib/runtime/__tests__/mcp-tools-utils.test.ts
```diff
@@ -41,238 +40,424 @@ describe("MCP Tools Utils", () => {
...
+      expect(result[0]).toEqual({
+        name: "simpleArray",
+        type: "array<string>",
+        description: "Array of strings",
+        required: true,
+      });
+      expect(result[1]).toEqual({
+        name: "objectArray",
+        type: "array",
+        description: "Array of objects Array of objects with properties: name, value",
```
hejtmii, line 114: The information of the shape for the nested properties gets lost. Especially when using general names like `value`

hejtmii: The concatenation with the original description is also somewhat weird, and might cause confusion for the LLM.

Imagine someone wants to narrow down the description to achieve desired behavior

`description: "Array of terms, name should be short and descriptive"`

Ending up with:

`Array of terms. Name should be short and descriptive Array of objects with properties: name, color`

Making the LLM to think `name` subproperty should be an array of objects...

It would deserve some better separator to make it clear those are two independent bits of information

alankpax:
Hi, I think I am still getting issues with params not being populated when LLM is calling tools resulting in validation errors in version 1.9.3.

It can bet reproduced easily running the Connect to MCP guide with a simple Gmail MCP Server running on Composio expecting a parameter

image

I have also reproduced similar behaviour implementing a runtime by providing tools from Mcp Client. I get these errors on every tool needing params when logging client calls server side

mcp failed copilotkit

Or maybe it's another issue ?

Issue #2113, opened 2025-07-06
🐛 Bug: loadAgentState returns all states (including input states and internal states)


angular-moon:
### ♻️ Reproduction Steps

define agent use inputs-outputs schema:
https://docs.copilotkit.ai/coagents/shared-state/state-inputs-outputs

```typescript
 <CopilotKit
      threadId="8c954fea-b40c-468d-8fcb-5fdad9d74920"
      ...
    >
      <CopilotChat/>
   </CopilotKit>
```

Load the specified thread

### ✅ Expected Behavior

don't return input states and internal states

### ❌ Actual Behavior

loadAgentState returns all states, including input states and internal states

### 𝌚 CopilotKit Version

```shell
@copilotkit/react-core@1.9.0 
@copilotkit/react-ui@1.9.0
@copilotkit/runtime-client-gql@1.9.1

python-sdk:
copilotkit=0.1.52
```

### 📄 Logs (Optional)

```shell

```

NathanTarbert:
Hi @angular-moon, just curious if you'v had the chance to try the CopilotKit v1.50 beta?

BenTaylorDev:
Superseded — the `loadAgentState` GraphQL path that returned unfiltered state has been gutted (the resolver returns `{}`, and react-core makes no `loadAgentState` calls); agent state now flows via AG-UI snapshots. Reopen if you still hit this on a current version.

PR #2125, opened 2025-07-08
improve: get_state (filter the output in the state).

angular-moon:
In fact, the get_input_jsonschema and get_output_jsonschema of langgraph do not use the config parameter. 
The get_state has been improved to only return the output data, avoiding returning the input and internal state when initializing the thread.

fixed: #2113

### DIFF
--- a/sdk-python/copilotkit/langgraph_agent.py
+++ b/sdk-python/copilotkit/langgraph_agent.py
@@ -185,6 +185,12 @@ class LangGraphAgent(Agent):
         self.graph = cast(CompiledStateGraph, graph or agent)
         self.active_interrupt_event = False
 
+        # Get the output and input schema keys the user has allowed for this graph
+        input_keys, output_keys, config_keys = self.get_schema_keys()
+        self.output_schema_keys = output_keys
+        self.input_schema_keys = input_keys
+        self.config_schema_keys = config_keys
+
     def execute( # pylint: disable=too-many-arguments
             self,
             *,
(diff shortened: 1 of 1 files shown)


ranst91:
Thanks for the contribution
Due to lack of setup on our side, I will have to recreate the PR in order for our tests and workflow to run. I'll try to get to it soon so this can be added quickly

ehthind:
> Thanks for the contribution Due to lack of setup on our side, I will have to recreate the PR in order for our tests and workflow to run. I'll try to get to it soon so this can be added quickly

Hey @ranst91, was this fixed in a follow up PR for a specific version? I'm running into this issue as well

ranst91:
Hey @angular-moon and @ehthind 
Since then we moved to AGUI implementation (the fastapi part too!) and the version this PR refers to is no longer maintained. I encourage you to move over to the AG-UI version, where these kind of fixes are regular
See code example here: https://github.com/CopilotKit/with-langgraph-fastapi

ehthind:
> Hey @angular-moon and @ehthind Since then we moved to AGUI implementation (the fastapi part too!) and the version this PR refers to is no longer maintained. I encourage you to move over to the AG-UI version, where these kind of fixes are regular See code example here: https://github.com/CopilotKit/with-langgraph-fastapi

@ranst91  Thanks for the info! How do we forward the auth token? [Similar to this question here](https://discord.com/channels/1122926057641742418/1369715020392304732/1437426072999428166) with `LangGraphAgent` I was able to use the langgraph_config property. I understand that we can forward auth information from the frontend using the coagent hook and the config property. However, I don't have access to auth information on the frontend I was doing something like this before:

``` 
const handler = copilotRuntimeNodeExpressEndpoint({
      endpoint: ENDPOINT,
      runtime,
      serviceAdapter,
      properties: {
        authorization: req.headers.authorization || '',
        threadId: req.body?.variables?.data?.threadId || '',
      },
      ```

**Comment by ranst91** [comment]
Thanks for drawing my attention to this.
I have replied there directly: https://discord.com/channels/1122926057641742418/1369715020392304732/1443647245558874243

With the following:
LangGraphAGUIAgent is indeed the AGUI Implementation.

As for the the other questions regarding auth, the "LangGraphAgent" class takes a "propertyHeaders" proprety like so:
```
new LangGraphAgent({
  ...your definition
  propertyHeaders: {...any auth you want to pass to the LG client}
})
```
This is helpful if you're using headers to call the client (so actual LG auth system).
If you just need this in the "config" or "configurable" object, you'd need to use "assistantConfig" instead
```
new LangGraphAgent({
  ...your definition
  assistantConfig: { ...stuff, configurable: {...more stuff} } <--- This is also accessibly to set in the 
})
```

mme:
Closing this as resolved. Thanks again for the PR and for taking the time to contribute!

PR #2154, opened 2025-07-14
fix: do not show error on state loading if thread does not exist

ranst91:
Currently in development we show error when loading langgraph agent state for CPK initialization sake.
The thread ID we supply might not exist on the LG side, therefore a 404 error should not show up as an error, as the thread might not exist which is acceptable

### DIFF
--- a/CopilotKit/packages/runtime/src/lib/runtime/copilot-runtime.ts
+++ b/CopilotKit/packages/runtime/src/lib/runtime/copilot-runtime.ts
@@ -975,16 +975,21 @@ please use an LLM adapter instead.`,
     } catch (error) {
       // All errors from agent state loading are user configuration issues
       const errorMessage = error instanceof Error ? error.message : String(error);
+      const errorStatus = error?.response?.status || error?.status;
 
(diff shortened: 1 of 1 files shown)


CopilotKit/packages/runtime/src/lib/runtime/copilot-runtime.ts
```diff
@@ -975,16 +975,21 @@ please use an LLM adapter instead.`,
...
-      // Log user configuration errors at debug level to reduce noise
-      console.debug(`Agent '${agentName}' configuration issue: ${errorMessage}`);
-
-      // Throw a configuration error - all agent state loading failures are user setup issues
-      throw new ResolvedCopilotKitError({
-        status: 400,
-        message: `Agent '${agentName}' failed to execute: ${errorMessage}`,
-        code: CopilotKitErrorCode.CONFIGURATION_ERROR,
-      });
+      if (errorStatus === 404) {
```
suhasdeshpande, line 980: why do we not want to surface 404?

PR #2213, opened 2025-07-22
fix(sdk-python): preserve state messages on subsequent get_state calls

landim:
## What does this PR do?

This PR fixes a critical bug in the Python SDK's `LangGraphAgent.get_state()` method where the messages key was being deleted from the cached state dictionary, causing data loss on subsequent calls.

### The Problem
In commit cf90f85, the `get_state()` method was modified to cache thread states in `self.thread_state[thread_id]`. However, at line 629, the code directly deletes the "messages" key from the cached state:

```python
state = self.thread_state[thread_id]
messages = langchain_messages_to_copilotkit(state.get("messages", []))
del state["messages"]  # This mutates the cached state\!
```

This causes:
- First call to `get_state()`: Works correctly, returns messages
- Second call to `get_state()`: Returns empty messages array because the "messages" key was deleted from the cached state

### The Solution
This PR introduces a deep copy of the state before removing the messages key:

```python
# Create a deep copy of the state to avoid modifying the original
state_copy = deepcopy(state)
messages = langchain_messages_to_copilotkit(state_copy.get("messages", []))
del state_copy["messages"]
```

This ensures the cached state remains intact for future calls while still providing the expected return format.

## Related PRs and Issues

- Related to commit: https://github.com/CopilotKit/CopilotKit/commit/cf90f85aad1d69f2257b70318f86ad052ad68a77#diff-a34f2fd44d5bc3c9e21e9ec55c4e7a3760def6a649d046e2eda6173c671d3580R616-R619

## Checklist

### DIFF
--- a/sdk-python/copilotkit/langgraph_agent.py
+++ b/sdk-python/copilotkit/langgraph_agent.py
@@ -2,6 +2,7 @@
 
 import uuid
 import json
+from copy import deepcopy
 from typing import Optional, List, Callable, Any, cast, Union, TypedDict, Literal
 
 from langgraph.graph.state import CompiledStateGraph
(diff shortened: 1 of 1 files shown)


ranst91:
Due to how our CI works (is being worked on) this is integrated in a separate PR mentioned here. Therefore I am closing this specific PR.

Thanks for the contribution!

PR #2273, opened 2025-08-04
docs(headless-ui): update messaging and guides

tylerslaton:


peterj35-dd:
Just one question.... why?

PR #2279, opened 2025-08-05
fix: limit ReplaySubject buffer to prevent OOM

suhasdeshpande:
### **ReplaySubject Constructor:**
```typescript
Rx.ReplaySubject([bufferSize], [windowSize], [scheduler])
```

**Arguments:**
1. `[bufferSize = Number.MAX_VALUE]` _(Number)_: Maximum element count of the replay buffer.

### **The Problem:**
- **Default buffer size = `Number.MAX_VALUE`** 
- **`Number.MAX_VALUE` = 1.7976931348623157e+308** (basically infinite)
- **Every single event is stored forever**

**Our code:**
```typescript
export class RuntimeEventSubject extends ReplaySubject<RuntimeEvent> {
  constructor() {
    super(); // ← DEFAULT = Number.MAX_VALUE buffer size!
  }
}
```

**Problem**: CopilotKit runtime experiencing OOM crashes during long-running LangGraph sessions (5+ minutes).

**Root Cause**: `RuntimeEventSubject` extends `ReplaySubject` without specifying a buffer limit. Per RxJS documentation, this defaults to `Number.MAX_VALUE`, causing **every single event to be retained in memory indefinitely**.

**Evidence**: Official RxJS docs state: "be cautious. ReplaySubject can potentially use more memory since it's storing multiple values. Ensure that you specify a limited buffer size if you're concerned about memory consumption."

**Solution**: Added `MAX_BUFFER_SIZE` limit to prevent unbounded memory growth while preserving recent event history for subscribers.

## Summary by CodeRabbit

* **Bug Fixes**
  * Limited the event buffer size to prevent potential out-of-memory errors during runtime.
  * Added warnings to notify when large numbers of events have been buffered.

### DIFF
--- a/CopilotKit/packages/runtime/src/service-adapters/events.ts
+++ b/CopilotKit/packages/runtime/src/service-adapters/events.ts
@@ -109,11 +109,19 @@ interface RuntimeEventWithState {
 
 type EventSourceCallback = (eventStream$: RuntimeEventSubject) => Promise<void>;
 
-const MAX_BUFFER_SIZE = 50;
+const BUFFER_SIZE = process.env.COPILOT_EVENT_BUFFER_SIZE
+  ? parseInt(process.env.COPILOT_EVENT_BUFFER_SIZE)
+  : undefined; // Keep unlimited by default
 
 export class RuntimeEventSubject extends ReplaySubject<RuntimeEvent> {
+  private static eventCount = 0;
+
   constructor() {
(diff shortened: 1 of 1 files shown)


tylerslaton (commented):
Can we get some documentation for this before merging? I want users to be aware they have this option and why they may want it. Perhaps in the trouble-shooting section?

PR #2491, opened 2025-09-24
fix out of sync pnpm lockfiles

maxkorp:
## What does this PR do?

Dunno how this happened, but a bunch of pnpm lock files in the examples folders are out of sync, and it's blowing up deploys. 

NOW THEY CAN BE DEDESYNCED (I believe some people call that synced?)


maxkorp:
Off it goes!

PR #2496, opened 2025-09-25
fix: support older browsers in markdown and code block components

ranst91:
This PR disabled "lookback expression" on older browsers, which is widely used by some of the dependencies we use for the code block and syntax highlighting

## Summary by CodeRabbit

- Bug Fixes
  - Fixed markdown and code block rendering issues on older browsers.

- Performance
  - Syntax highlighting is now lazy-loaded to reduce initial load.

- Reliability
  - Added automatic, runtime fallback for environments lacking newer regex support so code blocks and markdown remain readable.

- Chores
  - Added release metadata for a patch update.


ssnielsen:
@ranst91 @tylerslaton

👋 Hope it's ok that I tag you here. Tried to find an issue related to this, but found this PR instead.

We've had this exact issue with this on our own application after adding CopilotKit (version 1.10.6).

Looks like the issue should be resolved with this PR, but it seems like it's not actually part of any releases (nor on `main`). Was it reverted again later?

PR #2564, opened 2025-10-01
fix(render-message): return null for default case

### LINKED ISSUE #2542: 🐛 Bug: RenderMessage render returning nothing
### ♻️ Reproduction Steps

The render section of RenderMessage is returning nothing in case of default case matching. [Refer this](https://github.com/CopilotKit/CopilotKit/blob/main/CopilotKit/packages/react-ui/src/components/chat/messages/RenderMessage.tsx#L24). When any tool call message is sent, then it breaks/throws error.

This causes React to throw error at runtime. We are using React v17, which is unsupported. But render method in React should still return some default value.

### ✅ Expected Behavior

Ideally it should return null for default case.

### ❌ Actual Behavior

An error occurs in the console and breaks the page.

### 𝌚 CopilotKit Version

```shell
1.10.4
```

### 📄 Logs (Optional)

```shell
RenderMessage(...): Nothing was returned from render. This usually means a return statement is missing. Or, to render nothing, return null.
```

If you want then I can raise a PR handling the default case.

DexAsHisH:
## What does this PR do?

Added `return null` statement in the default case of the switch statement in `RenderMessage` component to prevent React runtime errors when encountering unexpected message roles.

**Changes:**

- Added `return null` in the default case of the switch statement in `RenderMessage.tsx`.
- This ensures the render function always returns a valid value, JSX or null.

## Related PRs and Issues

- Fixes #2542

## Checklist

### DIFF
--- a/CopilotKit/packages/react-ui/src/components/chat/messages/RenderMessage.tsx
+++ b/CopilotKit/packages/react-ui/src/components/chat/messages/RenderMessage.tsx
@@ -51,5 +51,7 @@ export function RenderMessage({
           ImageRenderer={ImageRenderer}
         />
       );
+    default:
+      return null;
   }
 }


brandonmcconnell:
@DexAsHisH Looks good! Thanks for opening this PR.

PR #2569, opened 2025-10-02
fix(react-ui): address hydration issue

tylerslaton:
Fixing an issue where we'd cause a hydration issue due to how we were reactive styling.

Screenshot 2025-10-02 at 3 08 37 PM

### DIFF
--- a/CopilotKit/packages/react-ui/src/components/chat/PoweredByTag.tsx
+++ b/CopilotKit/packages/react-ui/src/components/chat/PoweredByTag.tsx
@@ -1,8 +1,13 @@
-import React from "react";
+import React, { useState, useEffect } from "react";
 import { useDarkMode } from "../../hooks/use-dark-mode";
 
 export function PoweredByTag({ showPoweredBy = true }: { showPoweredBy?: boolean }) {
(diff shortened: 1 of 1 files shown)


CopilotKit/packages/react-ui/src/components/chat/PoweredByTag.tsx
```diff
@@ -1,9 +1,14 @@
-import React from "react";
+import React, { useState, useEffect } from "react";
 import { useDarkMode } from "../../hooks/use-dark-mode";
 
 export function PoweredByTag({ showPoweredBy = true }: { showPoweredBy?: boolean }) {
+  const [mounted, setMounted] = useState(false);
   const isDark = useDarkMode();
 
+  useEffect(() => {
```
maxkorp, line 16: We should probably add a comment as to why we're doing this. I know in context what the purpose for this is, but out of context why would I care what color my component is using if it hasn't been mounted yet? Perfect solution but not particularly intuitive.

tylerslaton: Good call out! Let me make a paper trial.

PR #2578, opened 2025-10-05
Setup React frontend and FastAPI backend with basic summarize endpoint

Meghana-2124:
Frontend + Backend Connected:
- React frontend with textarea and summarize button
- FastAPI backend `/summarize` endpoint working
- Currently uses placeholder logic (first 50 characters)
- Real summarization logic will be added in a follow-up PR


tylerslaton:
Hi @Meghana-2124!

Thanks for the work here. We won't be merging this as it is an example and being placed into the root. If you'd like to add an example, please checkout our contribution guidlines!

Thanks again.

Meghana-2124:
Okay  @tylerslaton. sure I'll check it out. Thanks.

PR #2580, opened 2025-10-06
fix: surface errors also in production

ranst91:
This PR enables surfacing errors even on production (so no dev console needed)


CopilotKit/packages/react-core/src/hooks/use-copilot-runtime-client.ts
```diff
@@ -115,20 +109,15 @@ export const useCopilotRuntimeClient = (options: CopilotRuntimeClientHookOptions
...
-            // TODO: if onError & renderError should work without key, insert here
-          }
+          // Route non-GraphQL errors to banner as well
+          const fallbackError = new CopilotKitError({
+            message: error?.message || String(error),
+            code: CopilotKitErrorCode.UNKNOWN,
+          });
+          setBannerError(fallbackError);
+          // Trace the non-GraphQL error
+          traceUIError(fallbackError, error);
```
tylerslaton, line 119: Does this mean that the ErrorBanner renders when `showDevConsole` is set to false as well? If so, I think what we really want is this:

```suggestion
          if(isDev) setBannerError(fallbackError);
          // Trace the non-GraphQL error
          traceUIError(fallbackError, error);
```

Do I understand this correctly?

ranst91: No, the banner is not rendered. It is set, and somewhere else, if this is dev mode, it will be rendered.

PR #2582, opened 2025-10-07
fix: allow disabling actions in the new action hooks

ranst91:
With the creation of the new convenience wrappers around useCopilotAction, the "available" got lost.
We should be able to set "disabled" on it


tylerslaton:
Would it make sense to just make this be a Boolean instead? Since this is a net new interface.

PR #2637, opened 2025-10-20
Feat/1.5 endpoint setup

ranst91:
In preparation of 1.5. This PR sets the new runtime

### DIFF
--- a/CopilotKit/packages/runtime/src/lib/runtime/copilot-runtime.ts
+++ b/CopilotKit/packages/runtime/src/lib/runtime/copilot-runtime.ts
@@ -377,8 +377,6 @@ export class CopilotRuntime {
     this.params = params;
     this.observability = params?.observability_c;
 
-    console.log(this.runtimeArgs);
-
     this.instance = new CopilotRuntimeVNext(this.runtimeArgs);
   }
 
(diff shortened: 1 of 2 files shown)


CopilotKit/packages/runtime-client-gql/src/client/types.ts
brandonmcconnell, line 71: Should there be a specified type in addition to just the default value for the `type` param?

Something like `type: TextMessageInput["type"] = "TextMessage" as const;`

CopilotKit/packages/runtime/src/lib/integrations/node-express/index.ts
brandonmcconnell, line 14: This looks fine – just making sure we're meaning to use `copilotRuntimeNodeHttpEndpoint` here and not `createCopilotEndpoint` like the others.

I assume this is probably right, since we are setting up the Express endpoint.

brandonmcconnell (approved):
Somehow, this comment didn't make it into my previous review

CopilotKit/packages/runtime-client-gql/src/client/types.ts
brandonmcconnell: Looks like this comment isn't relevant any anymore after your last commit/sync – marking resolved

CopilotKit/packages/runtime/src/lib/integrations/node-express/index.ts
ranst91: copilotRuntimeNodeHttpEndpoint is the one which would use the new endpoint under the hood.
So usually these files do: main func -> createEndpoint
This one does: main func -> some other file main func -> createEndpoint

brandonmcconnell: Makes sense! Good to merge then, I think. I'll resolve this comment.

Issue #2651, opened 2025-10-22
🚀 Feature Request: support `null`-unions in `convertJsonSchemaToParameter`


phryneas:
### 🎤 Tell us your idea

Currently, `convertJsonSchemaToParameter` doesn't support arrays in the `type` property. It should at least support `null`, since OpenAI actually doesn't support "optional" properties, but does allow something like `type: ["null", "string"]` to be passed in.

A schema like

```js
const schema = {
  type: "object",
  properties: {
    primaryTypeDisplayName: {
      type: ["object", "null"], // nullable union type
      properties: {
        text: { type: "string" },
        languageCode: { type: "string" },
      },
      required: ["text", "languageCode"],
    },
  },
};
```

would currently result in

```js
const parameters = [
  {
    name: "primaryTypeDisplayName",
    type: "string",
  },
];
```
which is 100% not the intended outcome.

PR #2652, opened 2025-10-22
add support for null-unions in `convertJsonSchemaToParameter`

phryneas:
## What does this PR do?

Currently, `convertJsonSchemaToParameter` doesn't support arrays in the `type` property. It should at least support `null`, since OpenAI actually doesn't support "optional" properties, but does allow something like `type: ["null", "string"]` to be passed in.

A schema like

```js
const schema = {
  type: "object",
  properties: {
    primaryTypeDisplayName: {
      type: ["object", "null"], // nullable union type
      properties: {
        text: { type: "string" },
        languageCode: { type: "string" },
      },
      required: ["text", "languageCode"],
    },
  },
};
```

would currently result in

```js
const parameters = [
  {
    name: "primaryTypeDisplayName",
    type: "string",
  },
];
```

which is 100% not the intended outcome.

With this PR, it results in

```js
const parameters = [
  {
    name: "primaryTypeDisplayName",
    description: undefined,
    type: "object", // Should be 'object', not 'string'
    required: false, // Should be optional since it can be null
    attributes: [
      { name: "text", description: undefined, type: "string" },
      { name: "languageCode", description: undefined, type: "string" },
    ],
  },
];
```

This doesn't support more complicated unions or nullable-elements in arrays (since there's no way in the current `parameters.type` to express that), but it should already improve a lot.

## Related PRs and Issues

#2651

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added support for nullable JSON schema types, enabling optional fields to be properly recognized and represented in parameter definitions.
  * Enhanced schema handling to support union types that include null values.

* **Improvements**
  * Improved error handling and messaging for unsupported schema scenarios involving complex nested types.

### DIFF
--- a/CopilotKit/packages/shared/src/utils/json-schema.test.ts
+++ b/CopilotKit/packages/shared/src/utils/json-schema.test.ts
@@ -4,6 +4,7 @@ import {
   actionParametersToJsonSchema,
   jsonSchemaToActionParameters,
   JSONSchema,
+  NullableJSONSchema,
 } from "../utils/json-schema";
 import { zodToJsonSchema } from "zod-to-json-schema";
 import { Parameter } from "../types";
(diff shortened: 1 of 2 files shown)


mme:
Thanks for the effort you put into this PR. We’re planning substantial changes to the JSON schema in the upcoming release, so this change will be superseded. Because of that, we’ll close the PR for now.

PR #2673, opened 2025-10-27
fix(react-ui): replace random key with stable hash in CodeBlock

### LINKED ISSUE #2669: 🐛 Bug: Random key in CodeBlock (@copilotkit/react-ui) causes flickering during streaming
### Description
While using `@copilotkit/react-ui@1.10.6` we noticed that code blocks in chat messages flicker whenever:
- the chat viewport is scrolled (React re-renders the list), or
- the assistant streams additional markdown into an existing message.

Root cause: the `CodeBlock` component rendered by `Markdown` sets `key={Math.random()}` on the underlying syntax highlighter. React therefore treats every update as a brand-new subtree, unmounting and remounting the code block on each render. The syntax highlighter starts over, producing a noticeable color flash.

### Expected behaviour
Code blocks should preserve their DOM/state across re-renders so that syntax highlighting remains stable during scrolling and streaming updates.

### Actual behaviour
Because the key changes on every render, React remounts the code block. Users see the highlighted code briefly disappear/reflash anytime the message list re-renders (e.g. when new tokens arrive).

### Suggested fix
Replace the random key with a stable value derived from the block’s content (e.g. a hash of the `language + value`). That keeps React from remounting the subtree while still allowing different snippets to coexist.

### Environment
- `@copilotkit/react-ui` v1.10.6
- React 18 / Next.js 14 (app router)

Thanks for all the work on CopilotKit! A stable key would immediately fix the UX regression without requiring downstream overrides.

ghost:
## Summary

Fixes the flickering issue in code blocks during streaming and scrolling by replacing `Math.random()` key with a stable hash based on language and code content.

## Problem

Previously, `CodeBlock` components rendered in Markdown used `key={Math.random()}` which caused React to unmount and remount the syntax highlighter on every render. This resulted in visible flickering whenever:
- The chat viewport was scrolled (React re-renders the list)
- The assistant streamed additional markdown into an existing message

## Solution

Replaced the random key with a deterministic hash function that generates stable keys from the combination of:
- The programming language
- The code content

This allows React to properly identify and maintain the same component instance across renders, eliminating the flickering.

## Changes

- Added `hashCode()` helper function to generate stable numeric hashes from strings
- Modified the code block rendering logic to use `${language}-${hashCode(codeValue)}` as the key
- Refactored to extract `codeValue` and `language` variables for better readability

## Testing

The JavaScript build completes successfully. The fix preserves React's reconciliation behavior while eliminating the visual flickering issue.

Fixes #2669

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced code block rendering in the chat interface to eliminate flickering and provide a more stable viewing experience.

### DIFF
--- a/CopilotKit/packages/react-ui/src/components/chat/Markdown.tsx
+++ b/CopilotKit/packages/react-ui/src/components/chat/Markdown.tsx
@@ -5,6 +5,17 @@ import remarkGfm from "remark-gfm";
 import remarkMath from "remark-math";
 import rehypeRaw from "rehype-raw";
 
+// Simple hash function to generate stable keys for code blocks
+function hashCode(str: string): number {
+  let hash = 0;
+  for (let i = 0; i < str.length; i++) {
+    const char = str.charCodeAt(i);
+    hash = (hash << 5) - hash + char;
+    hash = hash & hash; // Convert to 32bit integer
+  }
+  return Math.abs(hash);
+}
+
 const defaultComponents: Components = {
   a({ children, ...props }) {
     return (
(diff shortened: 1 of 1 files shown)


Dimitri-Accad:
Hi!
Is this still planned to be merged?

PR #2688, opened 2025-11-05
fix: address issue with turbopack on windows

tylerslaton:
Adding safe local storage access when it doesn't exists in certain environments

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved dev console stability in environments where browser storage is unavailable or restricted; prevents crashes and preserves graceful fallback for preference and version caching.
* **Refactor**
  * Centralized and hardened storage access and updated component prop handling to ensure safer state persistence and clearer defaults.


tylerslaton:
False alarm :(

PR #2978, opened 2026-01-08
Change req.on 'close' to req.socket.on 'close'

little-thing:
Node.js triggers the `req.close` event upon completing the request body parsing, which sets `request.signal.aborted` to true, preventing the result from being returned.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed request cancellation to trigger when the underlying socket closes, ensuring more reliable request cleanup on connection termination.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/src/v2.x/packages/runtime/src/endpoints/express-utils.ts
+++ b/src/v2.x/packages/runtime/src/endpoints/express-utils.ts
@@ -66,7 +66,7 @@ export function createFetchRequestFromExpress(req: ExpressRequest): Request {
   }
 
   const controller = new AbortController();
-  req.on("close", () => controller.abort());
+  req.socket.on("close", () => controller.abort());
   init.signal = controller.signal;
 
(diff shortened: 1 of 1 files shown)


jpr5:
Thanks for flagging this — the underlying issue (Node.js firing `req.close` on body-parse completion, not just on disconnect) was real.

However, this has already been addressed on `main` with a more robust approach — the `close` handler now guards with `if (req.aborted)` so it only fires on actual client disconnects:

```ts
req.on("close", () => {
    if (req.aborted) {
        abort();
    }
});
```

Additionally, the file has moved from `src/v2.x/packages/runtime/...` to `packages/v2/runtime/src/endpoints/express-utils.ts`, so this PR can't merge cleanly anymore.

Closing as already resolved. Thanks again for the contribution!

Issue #2946, opened 2026-01-02
🐛 Bug: 1.50 no longer renders AG-UI tool events correctly


ben-gineer:
### ♻️ Reproduction Steps

1. When running an AG-UI Agent using CopilotKit on 1.50, it no longer renders multiple tool call events correctly - instead only triggering the first tool call
2. An example page.tsx is attached

[page.tsx](https://github.com/user-attachments/files/24408156/page.tsx)

### ✅ Expected Behavior

I expect to see each successive tool call be rendered on the page. See example image using CopilotKit 1.10.6:

<img width="1591" height="1089" alt="Image" src="https://github.com/user-attachments/assets/160673df-ca8b-42d0-8ede-3b80c207d447" />

### ❌ Actual Behavior

Only the first tool call is shown. See example image using CopilotKit 1.50.1:

<img width="1591" height="1089" alt="Image" src="https://github.com/user-attachments/assets/db96ca52-ba63-4760-a8c3-af32131b2c9e" />

### 𝌚 CopilotKit Version

```shell
1.50.1
```

### 📄 Logs (Optional)

```shell

```

ben-gineer:
Why is this closed?

BenTaylorDev:
Triage note: the v2 tool-call render path has been substantially reworked since this was filed (e.g. `CopilotChatToolCallsView` now maps over *all* `message.toolCalls`, plus commits to dedupe rendered tool calls and harden the default renderer). We couldn't confirm the specific "only the first AG-UI tool call renders" repro from the current code. Could you retest on the latest release (1.63.x)? If it still reproduces, a minimal repro (or the AG-UI event sequence) would help a lot. Tracking #2927 alongside this as the same subsystem.

BenTaylorDev:
Correcting my earlier triage note on this one: it pointed at the wrong render path.

That note cited `CopilotChatToolCallsView` mapping over all `message.toolCalls`, which is the **v2** seam. You're on 1.50.1 with a v1 `page.tsx`, so your rendering goes through `useLazyToolRenderer`, and that seam is unchanged on `main`:

```tsx
// packages/react-core/src/hooks/use-lazy-tool-renderer.tsx
if (!message?.toolCalls?.length) return null;

const toolCall = message.toolCalls[0];
if (!toolCall) return null;
```

[...]

PR #3023, opened 2026-01-18
fix(react-core): return array of tool messages

ben-gineer:
## What does this PR do?

Attempt to apply recommended fix to enable all AG-UI tool call events to be intercepted in v1.50 (currently only first tool call message is shown). The fix was based upon the CopilotKit Support Bot's recommendation.

## Related PRs and Issues

- [Bug: 1.50 no longer renders AG-UI tool events correctly](https://github.com/CopilotKit/CopilotKit/issues/2946)

## Checklist

### DIFF
--- a/src/v1.x/packages/react-core/src/hooks/use-copilot-chat_internal.ts
+++ b/src/v1.x/packages/react-core/src/hooks/use-copilot-chat_internal.ts
@@ -509,9 +509,9 @@ export function useCopilotChatInternal({
 
       const lazyRendered = lazyToolRendered(message, allMessages);
       if (lazyRendered) {
-        const renderedGenUi = lazyRendered();
-        if (renderedGenUi) {
-          return { ...message, generativeUI: () => renderedGenUi };
+        const renderedGenUiArray = lazyRendered();
+        if (renderedGenUiArray && renderedGenUiArray.length > 0) {
+          return { ...message, generativeUI: () => renderedGenUiArray };
         }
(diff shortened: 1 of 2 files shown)


BenTaylorDev:
Thanks for this, and apologies it sat so long without a review.

The bug is real and still present on `main`: `use-lazy-tool-renderer.tsx` renders only `message.toolCalls[0]`, so an assistant message with parallel tool calls drops all but the first. Confirmed and captured in OSS-763.

I'm closing rather than asking you to rebase, for two reasons:

1. The tree moved under this branch — `src/v1.x/packages/react-core/` no longer exists. The file now lives at `packages/react-core/src/hooks/use-lazy-tool-renderer.tsx`, so the diff can't apply as authored.

2. There's an edge case to settle first. `renderToolCall` returns `null` when no renderer is registered for a tool, so the array can be `[null]`. That passes the `length > 0` guard and sets `generativeUI`, and because `[null]` is truthy, `AssistantMessage` then suppresses a caller's `props.subComponent` and renders an empty spacer div. Today's `if (renderedGenUi)` check happens to avoid this. Separately, the change alters the return type of `useLazyToolRenderer`, which is exported from `hooks/index.ts` — so someone needs to decide whether we take that as a breaking change.

Neither is a knock on the patch; mapping over all tool calls is the right shape. It just needs a couple of calls that are ours to make. The ticket carries your approach and credits this PR.

PR #3033, opened 2026-01-19
bump hono

maxkorp:
Theres a security hole with no known use in hono fixed in 4.11.4, so bumping to that everywhere in the name of snyk shhhhh


maxkorp:
Test failures are preexisting, gonna merge this and come back for those

PR #3049, opened 2026-01-21
feat(runtime): forward auth headers on agent.connect()

MikeRyanDev:
Add support for forwarding authorization and x-* headers to the AgentRunner.connect() method, matching the existing behavior in handleRunAgent().

- Add optional `headers` field to AgentRunnerConnectRequest interface
- Extract header forwarding logic into reusable header-utils module
- Update handleConnectAgent to extract and forward allowlisted headers
- Refactor handleRunAgent to use shared extractForwardableHeaders utility
- Add tests for header forwarding in connect and header utils

## What does this PR do?

## Related PRs and Issues

## Checklist

### DIFF
--- a/src/v2.x/packages/runtime/src/__tests__/handle-connect.test.ts
+++ b/src/v2.x/packages/runtime/src/__tests__/handle-connect.test.ts
@@ -1,0 +1,174 @@
+import { Observable } from "rxjs";
+import { describe, it, expect } from "vitest";
+import { BaseEvent } from "@ag-ui/client";
+import { handleConnectAgent } from "../handlers/handle-connect";
+import { CopilotRuntime } from "../runtime";
+import { AgentRunnerConnectRequest } from "../runner/agent-runner";
+
+describe("handleConnectAgent", () => {
+  const createMockRuntime = (
+    agents: Record<string, unknown> = {},
+    connectHandler?: (request: AgentRunnerConnectRequest) => Observable<BaseEvent>
+  ): CopilotRuntime => {
+    return {
+      agents: Promise.resolve(agents),
+      transcriptionService: undefined,
+      beforeRequestMiddleware: undefined,
+      afterRequestMiddleware: undefined,
+      runner: {
+        run: () =>
+          new Observable<BaseEvent>((subscriber) => {
+            subscriber.complete();
+          }),
+        connect: connectHandler ?? (() =>
+          new Observable<BaseEvent>((subscriber) => {
+            subscriber.complete();
+          })),
+        isRunning: async () => false,
+        stop: async () => false,
+      },
+    } as CopilotRuntime;
+  };
+
+  it("should return 404 when agent does not exist", async () => {
+    const runtime = createMockRuntime({});
+    const request = new Request("https://example.com/agent/nonexistent/connect", {
+      method: "POST",
+      headers: { "Content-Type": "application/json" },
+      body: JSON.stringify({
+        threadId: "thread-1",
+        runId: "run-1",
+        state: {},
+        messages: [],
+        tools: [],
+        context: [],
+        forwardedProps: {},
+      }),
+    });
+
+    const response = await handleConnectAgent({
+      runtime,
+      request,
+      agentId: "nonexistent-agent",
+    });
+
+    expect(response.status).toBe(404);
+    expect(response.headers.get("Content-Type")).toBe("application/json");
+
+    const body = await response.json();
+    expect(body).toEqual({
+      error: "Agent not found",
+      message: "Agent 'nonexistent-agent' does not exist",
+    });
+  });
+
+  it("forwards only authorization and custom x- headers to runner.connect()", async () => {
+    const recordedRequests: AgentRunnerConnectRequest[] = [];
+    let resolveConnect: (() => void) | undefined;
+    const connectInvoked = new Promise<void>((resolve) => {
+      resolveConnect = resolve;
+    });
+
+    const runtime = createMockRuntime(
+      { "test-agent": { clone: () => ({}) } },
+      (request: AgentRunnerConnectRequest) =>
+        new Observable<BaseEvent>((subscriber) => {
+          recordedRequests.push(request);
+          resolveConnect?.();
+          subscriber.complete();
+        })
+    );
+
+    const requestBody = {
+      threadId: "thread-1",
+      runId: "run-1",
+      state: {},
+      messages: [],
+      tools: [],
+      context: [],
+      forwardedProps: {},
+    };
+
+    const request = new Request("https://example.com/agent/test-agent/connect", {
+      method: "POST",
+      headers: {
+        "Content-Type": "application/json",
+        "X-Custom": "custom-value",
+        "X-Another-Header": "another-value",
+        Authorization: "Bearer forwarded-token",
+        Origin: "http://localhost:4200",
+        "User-Agent": "test-agent",
+      },
+      body: JSON.stringify(requestBody),
+    });
+
+    const response = await handleConnectAgent({
+      runtime,
+      request,
+      agentId: "test-agent",
+    });
+
+    expect(response.status).toBe(200);
+    await connectInvoked;
+
+    expect(recordedRequests).toHaveLength(1);
+    expect(recordedRequests[0].threadId).toBe("thread-1");
+    expect(recordedRequests[0].headers).toMatchObject({
+      authorization: "Bearer forwarded-token",
+      "x-custom": "custom-value",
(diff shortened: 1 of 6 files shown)


bobbyo:
manual audit: unable to post via automation now

bobbyo:
### RED-TEAM REVIEW (extended) for #3049

I re-reviewed this with a hostile/multi-lens pass and found several production risk points beyond the first-pass notes.

1) [P1] Over-broad header allowlist allows request-chain spoofing
-  forwards **all  headers**, including , , , and framework/proxy metadata. Those headers are commonly inserted by reverse proxies and are user-controllable in edge cases; forwarding them into  can let a client override trust/security assumptions in downstream adapters.
- Impact: request-origin identity confusion, SSRF / misrouting in custom adapters using forwarded headers for policy or routing.

2) [P1] Header overwrite can silently clobber server-trusted credentials
-  merges  and now shares the same utility.
- If a runner/agent preconfigures  for provider auth, any inbound request  header replaces it (first-class auth shadowing).
- Impact: downstream tool/provider calls can run under attacker-chosen token; hard-to-debug cross-tenant behavior changes under multi-tenant deployments.

3) [P1] Header normalization/canonicalization is inconsistent with existing runtime behavior
-  stores keys as iterated by .
- Fetch/undici/browser behavior is normalized by default today, but this is environment-sensitive and currently unprotected. Any runtime that emits non-lowercase  can result in duplicate keys ( + ) and nondeterministic overwrite order.
- Impact: request-level auth headers can be dropped or overwritten unexpectedly across environments.

4) [P1]  allowlist still includes proxy/infra/debug headers that can be high-noise and stateful
-  has no denylist and no value validation.
- Example , , ,  all forwarded verbatim.
- Impact: leakage of infra metadata and possible policy bypass in custom runners that treat all forwarded headers as trusted context.

5) [P2] Connect path now mutates request contract with only positive tests
-  now always adds  in  request object, but tests only validate normal SSE/vanilla happy path and do not validate connect streaming behavior under invalid/malformed headers or malformed input.
- Impact: regressions under back-pressured SSE/connect replay and malformed headers likely undetected until load/perf conditions.

6) [P2] API-compatibility regression risk for custom  implementations
-  changed to include  in interface only; custom runner implementations outside this repo that destructure request types or rely on strict runtime shape checks may break or start dropping fields.
- Impact: ecosystem/runtime breakage for adopters; this is exactly the kind of hidden compatibility risk that causes production surprises in enterprise setups.

7) [P2] No negative tests for parser/path interactions in shared utility
- New utility has only “happy-path” unit coverage. No cases for oversized headers, repeated header names, control chars, CR/LF injection, empty values, or invalid Unicode.
- Impact: could open parser instability or header injection surfaces in future adapters reading these values.

8) [P2]  and  now diverge by architecture visibility
- In built-in runners (, ),  is accepted but ignored, so behavior is inconsistent depending on runtime implementation (custom runner gets data, built-ins don’t).
- Impact: users may assume feature parity and silently no-op in prod while downstream logic expects headers to be present.

Given we had issues in this area before under concurrency, I’d like these fixed or at least documented as accepted behavior before merge.

Suggested mitigation baseline:
- Narrow allowlist to explicit header names (no bare ), and add explicit denylist for proxy/control headers.
- Validate/strip dangerous values and enforce header size/count caps.
- Add negative tests for malformed headers + concurrent connect calls.
- Add explicit contract note: built-in runners ignore forwarded headers in  currently.

bobbyo:
### Red-team review (extended) for #3049

I re-reviewed this with a hostile multi-lens pass and found additional blockers/risks.

1) P1 — Over-broad `x-*` allowlist can move proxy metadata into agent context
- `src/v2.x/packages/runtime/src/handlers/header-utils.ts:7-8` forwards all `x-*` headers.
- This includes request-chain headers such as `x-forwarded-*` and infra/debug headers that are often untrusted.
- Why it fails in prod: auth/routing logic in downstream runners can trust forwarded values and be spoofed by a client sending crafted headers.

2) P1 — Insecure header precedence allows request token override
- `src/v2.x/packages/runtime/src/handlers/handle-run.ts` merges `{...agent.headers, ...forwardableHeaders}`.
- Inbound `Authorization` now shadows runner-configured values.
- Why this can break: one malicious connect/run call can replace provider token used by downstream integrations.

3) P1 — No explicit allowlist/denylist policy and no normalization checks
- `shouldForwardHeader` allows any `x-*` with no denylist.
- Why breakage happens: high-volume deployments using gateway-generated headers (`x-trace-id`, `x-request-start`, etc.) get forwarded as mutable context and pollute agent state.

4) P2 — Connect path is not contract-consistent across runners
- `AgentRunnerConnectRequest.headers` is added in interface, but built-in runners ignore it.
  - `runner/in-memory.ts` connect payload is consumed without header use.
  - `runner/intelligence.ts` connect also ignores forwarded headers.
- Why this matters: custom runner implementers may expect behavior parity and silently get no-op forwarding in core environments.

5) P2 — Missing stress/abuse coverage in shared header extraction
- New utility tests (`header-utils.test.ts`) only cover happy paths.
- Missing tests: multi-value headers, empty/oversized header values, control characters, CR/LF, and malformed names.
- Why this can fail: payload size abuse or malformed headers can become runtime-level errors later in downstream adapters.

6) P2 — Connect-path validation coverage is too narrow
- `handle-connect.test.ts` only verifies successful request paths.
- Missing: invalid header combinations + malformed JSON + non-object request bodies + concurrent connect churn under SSE subscription lifecycle.
- Why this matters: intermittent production failures appear under load or adversarial traffic and slip past CI.

7) P2 — Unclear API compatibility contract
- Adding `headers` to connect request changes the public runtime contract of connect behavior, but no explicit doc/typing migration notes.
- Why this can break: enterprise integrators depending on exact request surface for runner implementations may behave unpredictably.

Recommended fixes:
- Replace bare `x-*` allowlist with explicit safe-header list + denylist for proxy/debug headers.
- Preserve deterministic header casing and validate/limit header size and count.
- Add explicit contract doc: built-in runners currently ignore connect headers; custom runners must opt in.
- Expand negative tests for malformed and abusive header inputs.

bobbyo:
### Round-2 red-team pass (hardened rubric)

Applied stricter format: severity + lens + concrete break scenario + prod impact.

- [P1][Security] Header trust boundary expansion via bare `x-*` allowlist
  - Evidence: `src/v2.x/packages/runtime/src/handlers/header-utils.ts:7-8`
  - Scenario: A client sends `X-Forwarded-For`, `X-User-Id`, or `X-Real-IP` along with request and a custom runner treats forwarded headers as truthy identity/routing inputs.
  - Why it breaks prod: Identity and routing can be poisoned before business logic, producing cross-tenant misrouting or audit corruption even when transport-level IP/CORS is correct.
  - Suggested fix: Explicit allowlist (`authorization`, a short list of app-specific headers) + explicit denylist for proxy/control headers.

- [P1][Security/Confidentiality] Request-provided `Authorization` overrides configured credentials
  - Evidence: `src/v2.x/packages/runtime/src/handlers/handle-run.ts` merge currently does `...agent.headers, ...forwardableHeaders`
  - Scenario: Internal default agent has provider token in `agent.headers`, but inbound request includes attacker-controlled `Authorization`.
  - Why it breaks prod: Downstream calls execute under caller-controlled token and can leak user tokens / exceed tenancy boundaries.
  - Suggested fix: Keep request headers immutable by default (e.g., never override existing non-empty `agent.headers.authorization`), or explicit precedence policy.

- [P2][Compatibility] `AgentRunnerConnectRequest.headers` added but no behavior guarantee in built-in runners
  - Evidence: `src/v2.x/packages/runtime/src/runner/agent-runner.ts`, `.../runner/in-memory.ts`, `.../runner/intelligence.ts`
  - Scenario: A custom runner depends on connect headers (for auth or correlation) and works, while in-memory/intelligence users see no effect in production deployments.
  - Why it breaks prod: Silent feature divergence leads to environment-dependent regressions that only appear after deployment.
  - Suggested fix: either implement/consume headers in built-ins or explicitly document unsupported behavior + gate with testable contract.

- [P2][Reliability] No validation caps on forwarded-header volume/shape
  - Evidence: `src/v2.x/packages/runtime/src/handlers/header-utils.ts:14-24`
  - Scenario: Malicious client sends many long headers (or malformed values), creating large objects during every connect/run.
  - Why it breaks prod: Under concurrency, per-request allocation and downstream string handling may increase tail latency or trigger parser exceptions in runners.
  - Suggested fix: enforce count/size caps + reject/trim invalid control chars before `agent.headers` injection.

### Net result of this round
I did not surface new logic-layer failures beyond the above; this is a stronger, narrower set with explicit failure semantics (less speculative than prior pass).

PR #3056, opened 2026-01-21
Fix: fix example e2e tests

maxkorp:
A few pieces to this
- Moves e2e folder to it's own pnpm workspace
- moves v1 examples INTO the root workspace
  - Most importantly, makes these use the local copilotkit deps (in TS, python to come)
- excludes those examples and the v2 storybooks from the default build command
- consolidates some pnpm overrides
- fixes some cors warnings
- fixes the e2e workflow given the above changes


MikeRyanDev (approved):
LGTM!

PR #3097, opened 2026-01-26
fix(python): support LangGraph 1.x interrupt() resume return values

### LINKED ISSUE #3096: copilotkit_interrupt assumes interrupt() returns message list; breaks on LangGraph 1.x resume returning string/dict
## Description
In `copilotkit==0.1.76` (and current main), `copilotkit.langgraph.copilotkit_interrupt` assumes that `langgraph.types.interrupt()` returns a list of messages after resuming, and does:

```python
answer = response[-1].content
```

However, with LangGraph 1.x HITL patterns, the resumed value is commonly the frontend's `resolve()` payload (often a string, e.g. `""` or a JSON string). In that case `interrupt()` returns a string/dict, and the indexing causes runtime errors (`TypeError` / `KeyError: -1` depending on wrappers).

This makes `copilotkit_interrupt` unusable for HITL in LangGraph 1.x unless users implement their own wrapper.

## Reproduction
1. Backend uses `copilotkit.langgraph.copilotkit_interrupt(message=...)`
2. Frontend uses `useLangGraphInterrupt` and resolves with `resolve("")` or `resolve(JSON.stringify({...}))`
3. After resuming, `copilotkit_interrupt` tries to index `response[-1]` and fails.

## Expected
`copilotkit_interrupt` should accept that `interrupt()` resume value may be a string/dict (not only a message list) and return it as answer without crashing.

## Actual
Crashes with `TypeError` / `KeyError: -1` on resume.

## Versions
- copilotkit: 0.1.76
- langgraph: 1.0.5 (also seen on 1.0.x)
- langchain: 1.2.0

## Proposed Fix
Make `copilotkit_interrupt` robust to `interrupt()` returning:
- `list[BaseMessage]` (existing behavior)
- `str` (treat as answer)
- `dict` (`json.dumps` to answer, or keep raw)
- `None`/other (stringify)

Also ensure GraphInterrupt is NOT swallowed (do not wrap `interrupt()` in try/except that catches GraphInterrupt).

weimingboya:
Fixes #3096

### What
Fix `copilotkit.langgraph.copilotkit_interrupt` to support LangGraph 1.x resume return values.

### Why
LangGraph's `interrupt()` may return the raw resume payload (string/dict) after resuming. The current implementation assumes a list of messages and accesses `response[-1].content`, which breaks HITL flows when frontend resolves with `resolve("")` / `resolve(JSON.stringify(...))`.

### How
- Keep interrupt payload format unchanged (`__copilotkit_interrupt_value__` / `__copilotkit_messages__`)
- Do NOT catch/swallow GraphInterrupt
- After resume, support response being list/str/dict/None/other and derive `answer` accordingly
- Add unit tests via monkeypatching `interrupt()`

### Backwards compatibility
If response is a message list, behavior is unchanged (still uses last message content).


weimingboya:
FYI: the failing checks here are Vercel deploy checks showing "Authorization required to deploy" (I don't have access to authorize deploys for the CopilotKit team). This PR only touches  and I verified the new unit tests pass locally ( in ). Happy to add a changeset if maintainers prefer a version bump for the Python SDK.

weimingboya:
Quick clarification: the failing checks are Vercel deploy checks showing "Authorization required to deploy" (I don't have access to authorize deploys for the CopilotKit team).

This PR only touches `sdk-python/` and I verified the new unit tests pass locally (`pytest` in `sdk-python`).

(Previous comment lost the backticked parts due to shell escaping.)

Happy to add a changeset if maintainers prefer a version bump for the Python SDK.

PR #3173, opened 2026-02-09
feat(runtime): add fetchRunHistory and agent to connect request

mme:
## Summary
- Add `agent` to `AgentRunnerConnectRequest` so runners can access agent configuration during connection
- Add `fetchRunHistory` support to in-memory and SQLite runners, enabling import of external conversation history (e.g. from LangGraph checkpointers)
- Add `convertMessagesToEvents` utility in the shared package with run/message-level deduplication


MikeRyanDev (approved):

packages/v2/shared/src/cloud-key.ts
```diff
@@ -0,0 +1,8 @@
+const COPILOT_CLOUD_PUBLIC_API_KEY_REGEX = /^ck_pub_[0-9a-f]{32}$/i;
+
+/**
+ * Returns true if the given string is a valid CopilotCloud public API key.
+ */
+export function isValidCopilotCloudKey(key: string): boolean {
```
MikeRyanDev, line 6: I think this is also implemented the Angular package. If you have a spare cycle, can you update there as well?

BenTaylorDev:
Closing as un-mergeable against the current tree — this targets the deleted `packages/v1/*` and `packages/v2/*` trees (the repo is now flat) plus version-bump churn; the changed symbols aren't present under the flat `packages/`. A fresh PR against the flat layout would be welcome if any part is still needed.

PR #3174, opened 2026-02-09
feat: use render tool on v2

ranst91:
This PR bring two new hooks to CPK v2:
useRenderTool - Registers a UI correspondant to a backend tool
useDefaultRenderTool - Just like the above, but is a wildecard, "render this as default for all tools if not specified..." kind of thing

### DIFF
--- a/.changeset/cyan-dots-promise.md
+++ b/.changeset/cyan-dots-promise.md
@@ -1,0 +1,5 @@
+---
+"@copilotkit/react-core": patch
+---
+
+fix: add dependencies to fix stale state of regenerate method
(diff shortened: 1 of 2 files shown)


src/v2.x/packages/react/src/hooks/use-render-tool.tsx
```diff
@@ -0,0 +1,116 @@
...
+          args: z.infer<S>;
+          status: "executing";
+          result: undefined;
+        }
+      | {
+          name: string;
+          args: z.infer<S>;
+          status: "complete";
+          result: string;
+        },
```
AlemTuzlak: these are repeated both here and below, maybe we could create interfaces for these 3 cases and also export them? Would allow users to create abstractions on top easier

PR #3186, opened 2026-02-11
fix: typing of useCopilotReadable

choqchoq-dkt:
## What does this PR do?

Fix typing issue in useCopilotReadable hook.

There is an issue in typing of the **convert** props. We pass in this method only value. 

It is is a type of 
```ts 
(value: any) => void
``` 
and not 
```ts 
convert?: (value: T) => string;
````

I also change typing:

- Remove useless dependencies props.
- Make hook generic.

That maybe imply a breaking change.

## Checklist


BenTaylorDev:
Closing as un-mergeable — targets the ancient deleted `src/v1.x/packages/react-core/` tree. If the `useCopilotReadable` convert-prop typing is still wanted, a fresh PR against the flat tree would be welcome.

PR #3188, opened 2026-02-11
fix: use the right properties for tool call renderers

ranst91:


jpr5:
Hey @ranst91 — thanks for the PR. I took a look at this against the current codebase and I think it can be closed. Here's the analysis:

**Property names**: The current code in `use-render-tool.tsx` uses `copilotkit.renderToolCalls` and `copilotkit.setRenderToolCalls`, which match the actual API defined in `react-core.ts` (getter at line 52, setter at line 66, private field `_renderToolCalls` at line 32). The proposed names `toolCallRenderers`/`setToolCallRenderers` don't exist anywhere in the codebase — renaming to them would cause a runtime error.

**Deps array change** (`config.name` → `config`): The current pattern `[config.name, copilotkit, extraDeps.length, ...extraDeps]` deliberately uses the stable primitive `config.name` rather than the `config` object. Since `config` is typically an object literal passed inline, it gets a new identity every render — using it as a dep would cause an infinite effect loop. This matches the pattern in `useFrontendTool` which similarly uses `[tool.name, tool.available, copilotkit, ...]`.

**File path**: The file has since moved from `src/v2.x/` to `packages/v2/`, so this PR also has merge conflicts.

Looks like this might have been from an earlier draft naming that didn't end up landing — the merged version in #3174 (from the day before) already had the correct names. Closing this out, but thanks for the attention to the code!

PR #3197, opened 2026-02-11
wip: add v2 interface docs

maxkorp:


tylerslaton:
Looking good but I notice that I can't see the old docs. Also there are references to copilotkitnext which should instead be copilotkit/react-core/v2 (as an example)

PR #3201, opened 2026-02-12
fix(runtime): add messages-tuple stream mode for LangGraph Platform streaming

tylerslaton:
## Summary
- Fixes streaming not working when connecting to LangSmith-hosted LangGraph deployments (e.g., graphs built with `create_agent` from `langchain`)
- The `events` stream mode produces no `on_chat_model_stream` data on LangGraph Platform for these graphs, causing all content to arrive as a single `MessagesSnapshot` at the end
- Adds `messages-tuple` as a fallback stream mode and converts `[AIMessageChunk, metadata]` tuples into AG-UI text/tool-call events
- Includes dedup logic so when `events` mode works (local dev), `messages-tuple` events are skipped — no regressions

## Test plan
- [ ] Build passes (`pnpm --filter @copilotkit/runtime build`)
- [ ] Local LangGraph dev server: streaming still works via `events` mode (no regressions)
- [ ] LangSmith-hosted deployment: token-by-token streaming now works via `messages-tuple` fallback
- [ ] Tool calls stream correctly through `messages-tuple` mode

### DIFF
--- a/src/v1.x/packages/runtime/src/lib/runtime/agent-integrations/langgraph/agent.ts
+++ b/src/v1.x/packages/runtime/src/lib/runtime/agent-integrations/langgraph/agent.ts
@@ -1,4 +1,4 @@
-import { map } from "rxjs";
+import { map, Subscriber } from "rxjs";
 import { LangGraphEventTypes } from "../../../../agents/langgraph/events";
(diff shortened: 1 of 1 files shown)


jpr5:
Superseded by #3394 — rebased onto current main and modernized for the current architecture.

PR #3207, opened 2026-02-12
refactor: migrate from Turborepo to Nx for task management

AlemTuzlak:
- Updated example.sh to use Nx commands instead of Turbo commands for linking and running tasks.
- Added new Nx task runner guidance in 40-nx.auto.mdc.
- Removed outdated Turborepo guidance in 40-turborepo.auto.mdc.
- Deleted turbo.json configuration as part of the migration to Nx.


tylerslaton (approved):

firxworx:
saw this in my github feed, I gotta say nx is the worst time sink when you actually use it, at least that's been my experience

Issue #1881, opened 2025-05-27
🚀 Feature Request: Mastra Message History Support


carlo-stanciu:
### 🎤 Tell us your idea

The main idea is to have CopilotChat components support message history for Mastra CoAgents similar to LangGraph.

At the moment, when specifying the thread in CopilotKit root component, the chat behaves as it should, meaning the Mastra server is updating the conversation history in the DB. However, when refreshing the app instance, we loose all previous messages in the UI. The CopilotKitChat component does not retrieve the conversation history based on thread ID nor can we specify an external list of messages to render. This makes the Mastra - CopilotKit integration not usable (for the moment).

Just wanted to say, you guys did an amazing job with CopilotKit!
Thank you!

orca-copilotkit:
Thank you for submitting this request. Our team will certainly review it and consider it for our roadmap.

bobbyo:
## This issue appears to be a 100% blocker to using Copilotkit frontend with a Mastra backend

### Request: If there is a workaround that allows reloading an app in a new browser and continuing chats using `CopilotKit` , please do share !

### TLDR details
For any app using a Mastra backend and Copilotkit frontend, when the app reloads users loose their old chat threads. They can only start new chat threads.

Technically, the data for the old chat threads exists--Mastra does save it to the db.
The user just can't see it in a `<CopilotChat> component.
Possibly worse, if the app does set the threadId in the `<CopilotKit threadId=old-threadId-here ..>`, the backend will keep updating the old thread, so the AI/Agent sees the old messages but the user cannot see them, a very confusing UX experience.

### Repro
Client-side code that does work as a new chat but not show historical chat content on app reload:
```
import { CopilotChat } from "@copilotkit/react-ui";
import { CopilotKit } from "@copilotkit/react-core";

export interface CopilotKitChatProps {
    threadId: string;
    agent: string;
}

export function CopilotKitChat({ threadId, agent }: CopilotKitChatProps) {
    return (
        <CopilotKit 
            runtimeUrl="https://localhost:3000/mastra/api/copilotkit"
            agent={agent}
            threadId={threadId}
        >
            <CopilotChat
                labels={{
                    title: "Your Assistant",
                    initial: "Hi! 👋 How can I assist you today?",
                }}
            />
        </CopilotKit>
    );
}
```



<img width="1055" height="1098" alt="Image" src="https://github.com/user-attachments/assets/9bb0659b-1be0-49de-b3f8-9317150cbae7" />

[...]

PR #3251, opened 2026-02-20
feat(runtime): add thread rehydration support for LangGraph agents

mxmzb:
## What does this PR do?

Enables loading previous conversation history when switching to an existing LangGraph thread. Adds two reusable utilities to @copilotkit/runtime/langgraph:

- convertLangGraphMessages — converts LangGraph messages to AG-UI format
- loadLangGraphPostgresState — reads thread state from LangGraph's PostgreSQL checkpoints

Also fixes a bug where the synthetic RUN_FINISHED event during state hydration was missing required threadId/runId fields, causing client-side validation failure.

## Demo app and implementation

- https://github.com/CopilotKit/with-langgraph-fastapi-persisted-threads
- https://with-langgraph-fastapi-persisted-th.vercel.app/

## Related PRs and Issues

- https://github.com/CopilotKit/CopilotKit/issues/1881 (this is for Mastra, but same underlying root cause)

### DIFF
--- a/.changeset/sour-kids-argue.md
+++ b/.changeset/sour-kids-argue.md
@@ -1,0 +1,6 @@
+---
+"@copilotkit/runtime": patch
+"@copilotkitnext/runtime": patch
+---
+
+feat(runtime): add thread rehydration support for LangGraph agents


MikeRyanDev:
We will implement this in a different way

PR #3260, opened 2026-02-23
feat(runtime): add framework-agnostic fetch-based runtime with adapte…

AlemTuzlak:
…rs and docs

Implement a framework-agnostic CopilotKit runtime built on the Fetch API. The core handler (createCopilotRuntimeHandler) takes a Request and returns a Response, enabling native support on Bun, Deno, Cloudflare Workers, Next.js App Router, React Router, and TanStack Start. Thin adapters are provided for Node.js HTTP (createCopilotNodeHandler/createCopilotNodeListener), Express (createCopilotExpressHandler), and Hono (createCopilotHonoHandler).

Key changes:
- Add core fetch handler, router, CORS, and lifecycle hooks
- Add Node.js, Express, and Hono adapters with multi-route/single-route modes
- Add @copilotkit/runtime/v2 subpath exports (v2, v2/express, v2/hono, v2/node)
- Deprecate old function names (createCopilotEndpoint, createCopilotEndpointExpress, createNodeFetchHandler, createCopilotEndpointSingleRoute)
- Add Runtime Server Adapter docs page with examples for 10 frameworks
- Add example projects for Express, Hono, Node.js, Bun/Elysia, Deno, CF Workers
- Fix docs dark mode dropdown styling and card layout issues
- Adds CI integration tests for every runtime

## What does this PR do?

## Related PRs and Issues

## Checklist


ranst91 (dismissed):

packages/v2/agent/src/uuid.ts
```diff
@@ -0,0 +1,35 @@
+let IDX = 256;
```
mme: There is a `generateRandomId()` function in the shared package, use that instead?

mme (changes requested):

packages/v2/agent/src/uuid.ts
AlemTuzlak: done @mme

tylerslaton (commented):
# Code Review: Framework-Agnostic Fetch-Based Runtime

Nice work on this, Alem — the architecture is solid. The `(Request) → Promise<Response>` core with thin adapters is exactly the right pattern. Hook system is clean, test coverage is extensive (~4,100 new test lines), and backward compat is well-handled with deprecated aliases.

Below are the issues I found, grouped by priority.

---

## High Priority

### 1. `__cpk_methodCall` mutation lost if hooks replace the request
**`fetch-handler.ts` — `resolveSingleRoute` / `dispatchSingleRoute`**

In single-route mode, `(request as any).__cpk_methodCall = methodCall` stashes parsed data on the request. If `onBeforeHandler` returns a new `Request`, the property is lost and `dispatchSingleRoute` reads `undefined`.

**Fix:** Return `methodCall` from `resolveSingleRoute` and pass it as an explicit parameter to `dispatchSingleRoute`.

### 2. `credentials: true` + wildcard origin silently produces invalid CORS
**`fetch-cors.ts` ~line 56-62**

Per the Fetch spec, `Access-Control-Allow-Origin: *` with `Access-Control-Allow-Credentials: true` causes browsers to reject the response. Should either throw a config error or auto-resolve `*` to the request origin when credentials are enabled.

### 3. `runOnError` can throw unhandled
**`fetch-handler.ts` — main catch block**

If the `onError` hook itself throws, it propagates unhandled. Wrap `runOnError` in a try/catch with a final fallback 500 response.

### 4. `decodeURIComponent` can throw `URIError` on malformed URLs
**`fetch-router.ts` ~lines 87, 97, 104**

Malformed percent-encoding (e.g., `%ZZ`) in path segments causes an uncaught `URIError`. Wrap in try/catch and return `null` (no match).

---

## Medium Priority

### 5. Silent error swallowing in Node/Express adapters
**`node-fetch-handler.ts` ~line 64, `express-fetch-bridge.ts` ~line 79**

Bare `catch {}` discards errors without logging. Should at minimum be `catch (err) { console.error(err) }`.

### 6. Missing `AbortSignal` forwarding in Express slow path
**`express-fetch-bridge.ts` — `buildPreParsedRequest`**

The old code had explicit abort signal wiring. The new pre-parsed-body path doesn't attach an `AbortSignal`, so client disconnects won't propagate. Also, the deleted `express-abort-signal.test.ts` was not replaced.

### 7. `synthesizeBody` null guard uses wrong comparison
**`express-fetch-bridge.ts` ~line 88**

`body !== undefined` should be `body !== null`. Since `typeof null === "object"`, a null body produces `JSON.stringify(null)` → the string `"null"`. Mitigated by `hasPreParsedBody` rejecting null earlier, but the function itself is logically incorrect.

### 8. Missing `Allow` header on 405 responses
**`fetch-handler.ts` — `validateHttpMethod`**

RFC 9110 §15.5.6 requires 405 responses to include an `Allow` header.

### 9. Error messages leaked to clients in 500 responses
**`fetch-handler.ts` ~line 215**

`error.message` is returned directly in the JSON body, which can expose internals (file paths, DB errors). Consider redacting in production or making configurable.

### 10. After-request middleware may double-consume response body
**`fetch-handler.ts` ~lines 170-176**

The after-middleware is fire-and-forget but reads `response.body` after the response may have been piped to the client. `Response` body can only be consumed once. Should `response.clone()` before reading.

---

## Low Priority / Nits

### 11. Suffix matching without `basePath` is overly permissive
**`fetch-router.ts`** — Without basePath, any URL ending in `/info` etc. matches regardless of prefix. Docs should recommend `basePath` in production.

### 12. Missing `Vary` headers on preflight
**`fetch-cors.ts`** — Should also set `Vary: Access-Control-Request-Headers, Access-Control-Request-Method` for CDN caching.

### 13. Forward-reference code ordering
**`node-fetch-handler.ts`** — `const createNodeFetchHandler = createCopilotNodeHandler` appears before the function declaration. Works due to hoisting but reads confusingly — swap the order.

### 14. Bun tests use vitest imports but run with `bun test`
**`test_runtime-servers.yml`** — Works today but is fragile. Consider `bunx vitest run` for consistency.

---

## Breaking Change Concern

### `CopilotKitRequestHandler` type removed without alias
The `CopilotKitRequestHandler` type was deleted from `handler.ts` with no re-export. This is an exported type — anyone importing it gets a build error. Since the changeset is `minor`, this should get a deprecated type alias to avoid a semver violation.

Similarly, `createFetchRequestFromExpress` and `sendFetchResponse` were removed. If these were considered public API, they should also get deprecated aliases.

---

## Summary

| Priority | Count |
|----------|-------|
| High | 4 |
| Medium | 6 |
| Low | 4 |
| Breaking change concern | 1 |

The high-priority items are all fixable without architectural changes. Once addressed, this is in great shape to merge. 🚀

tylerslaton (changes requested):
# Code Review: Framework-Agnostic Fetch-Based Runtime

Nice work on this, Alem — the architecture is solid. The `(Request) → Promise<Response>` core with thin adapters is exactly the right pattern. Hook system is clean, test coverage is extensive (~4,100 new test lines), and backward compat is well-handled with deprecated aliases.

Below are the issues I found, grouped by priority. Requesting changes for the high-priority items.

---

## High Priority

### 1. `__cpk_methodCall` mutation lost if hooks replace the request
**`fetch-handler.ts` — `resolveSingleRoute` / `dispatchSingleRoute`**

In single-route mode, `(request as any).__cpk_methodCall = methodCall` stashes parsed data on the request. If `onBeforeHandler` returns a new `Request`, the property is lost and `dispatchSingleRoute` reads `undefined`.

**Fix:** Return `methodCall` from `resolveSingleRoute` and pass it as an explicit parameter to `dispatchSingleRoute`.

### 2. `credentials: true` + wildcard origin silently produces invalid CORS
**`fetch-cors.ts` ~line 56-62**

Per the Fetch spec, `Access-Control-Allow-Origin: *` with `Access-Control-Allow-Credentials: true` causes browsers to reject the response. Should either throw a config error or auto-resolve `*` to the request origin when credentials are enabled.

### 3. `runOnError` can throw unhandled
**`fetch-handler.ts` — main catch block**

If the `onError` hook itself throws, it propagates unhandled. Wrap `runOnError` in a try/catch with a final fallback 500 response.

### 4. `decodeURIComponent` can throw `URIError` on malformed URLs
**`fetch-router.ts` ~lines 87, 97, 104**

Malformed percent-encoding (e.g., `%ZZ`) in path segments causes an uncaught `URIError`. Wrap in try/catch and return `null` (no match).

---

## Medium Priority

### 5. Silent error swallowing in Node/Express adapters
**`node-fetch-handler.ts` ~line 64, `express-fetch-bridge.ts` ~line 79**

Bare `catch {}` discards errors without logging. Should at minimum be `catch (err) { console.error(err) }`.

### 6. Missing `AbortSignal` forwarding in Express slow path
**`express-fetch-bridge.ts` — `buildPreParsedRequest`**

The old code had explicit abort signal wiring. The new pre-parsed-body path doesn't attach an `AbortSignal`, so client disconnects won't propagate. Also, the deleted `express-abort-signal.test.ts` was not replaced.

### 7. `synthesizeBody` null guard uses wrong comparison
**`express-fetch-bridge.ts` ~line 88**

`body !== undefined` should be `body !== null`. Since `typeof null === "object"`, a null body produces `JSON.stringify(null)` → the string `"null"`. Mitigated by `hasPreParsedBody` rejecting null earlier, but the function itself is logically incorrect.

### 8. Missing `Allow` header on 405 responses
**`fetch-handler.ts` — `validateHttpMethod`**

RFC 9110 §15.5.6 requires 405 responses to include an `Allow` header.

### 9. Error messages leaked to clients in 500 responses
**`fetch-handler.ts` ~line 215**

`error.message` is returned directly in the JSON body, which can expose internals (file paths, DB errors). Consider redacting in production or making configurable.

### 10. After-request middleware may double-consume response body
**`fetch-handler.ts` ~lines 170-176**

The after-middleware is fire-and-forget but reads `response.body` after the response may have been piped to the client. `Response` body can only be consumed once. Should `response.clone()` before reading.

---

## Low Priority / Nits

### 11. Suffix matching without `basePath` is overly permissive
**`fetch-router.ts`** — Without basePath, any URL ending in `/info` etc. matches regardless of prefix. Docs should recommend `basePath` in production.

### 12. Missing `Vary` headers on preflight
**`fetch-cors.ts`** — Should also set `Vary: Access-Control-Request-Headers, Access-Control-Request-Method` for CDN caching.

### 13. Forward-reference code ordering
**`node-fetch-handler.ts`** — `const createNodeFetchHandler = createCopilotNodeHandler` appears before the function declaration. Works due to hoisting but reads confusingly — swap the order.

### 14. Bun tests use vitest imports but run with `bun test`
**`test_runtime-servers.yml`** — Works today but is fragile. Consider `bunx vitest run` for consistency.

---

## Breaking Change Concern

### `CopilotKitRequestHandler` type removed without alias
The `CopilotKitRequestHandler` type was deleted from `handler.ts` with no re-export. This is an exported type — anyone importing it gets a build error. Since the changeset is `minor`, this should get a deprecated type alias to avoid a semver violation.

Similarly, `createFetchRequestFromExpress` and `sendFetchResponse` were removed. If these were considered public API, they should also get deprecated aliases.

---

## Summary

| Priority | Count |
|----------|-------|
| High | 4 |
| Medium | 6 |
| Low | 4 |
| Breaking change concern | 1 |

The high-priority items are all fixable without architectural changes. Once addressed, this is in great shape to merge.

maxkorp (dismissed):

jpr5 (changes requested):
## Prior Review Comments (Resolved)

- **Use** `generateRandomId()` **from shared package** instead of custom UUID implementation — resolved by author (now imports `randomUUID` from `@copilotkitnext/shared`)

---

## Critical Issues

### 1. Fire-and-forget observable subscription in intelligence run handler

**File**: `packages/v2/runtime/src/handlers/intelligence/run.ts`

The observable from `runtime.process()` is subscribed but the subscription is never stored or cleaned up. If the client disconnects or the request is aborted, the observable continues running in the background, leaking resources.

**Recommendation**: Store the subscription and tie its lifecycle to the request/response — unsubscribe on client disconnect or abort signal.

---

### 2. Type mismatch: `CopilotRuntime` vs `CopilotRuntimeLike` across handler signatures

**Files**: `packages/v2/runtime/src/core/runtime.ts`, handler files in `packages/v2/runtime/src/handlers/`

Some handlers accept `CopilotRuntime` (concrete class) while the fetch handler and adapters use `CopilotRuntimeLike` (structural interface). This creates a hidden coupling — any runtime passed through adapters must actually be a full `CopilotRuntime` instance, but the type system doesn't enforce this consistently.

**Recommendation**: Standardize on `CopilotRuntimeLike` for all handler signatures, or document explicitly which handlers require the concrete class and why.

---

### 3. `RouteInfo` type at hooks.ts:36 uses weak typing that defeats discriminated union benefits

**File**: `packages/v2/runtime/src/core/hooks.ts:36`

The `RouteInfo` type used in lifecycle hooks doesn't carry the discriminated union tag from the route matching, so hook consumers can't narrow on the route type. This means hooks that need route-specific behavior must use unsafe type assertions.

**Recommendation**: Thread the discriminated union type through to the hooks interface so consumers can pattern-match on route kind.

---

## Important Issues

### 4. Empty catch blocks silently swallow SSE errors

**File**: `packages/v2/runtime/src/handlers/shared/sse-response.ts`

Multiple `catch` blocks in SSE streaming code are empty or only log at debug level. SSE connection failures, encoding errors, or stream corruption will be invisible to operators.

**Recommendation**: Log at `warn` or `error` level. Consider emitting a final SSE error event before closing the stream so clients can distinguish "stream ended normally" from "stream died."

---

### 5. Multiple `as any` type casts bypass TypeScript safety

**Files**: `packages/v2/runtime/src/endpoints/express-fetch-bridge.ts`, `node-fetch-handler.ts`, and others

At least 6 instances of `as any` were found across the new code. These defeat the purpose of TypeScript and hide potential runtime type mismatches.

**Recommendation**: Replace with proper typing. Where bridge code genuinely cannot type external input, use `unknown` with explicit narrowing instead of `any`.

---

### 6. Inconsistent or missing structured logging

**Files**: Various handler and adapter files

Some files use `console.log`, others use a structured logger, and several critical paths have no logging at all (e.g., CORS rejection, route not found, hook errors).

**Recommendation**: Adopt the project's existing logger consistently. Add structured logging for: route matching decisions, CORS policy evaluation, hook lifecycle, and error paths.

---

### 7. Error message text doesn't match actual error condition

**File**: `packages/v2/runtime/src/endpoints/single-route-helpers.ts`

The method validation error message references incorrect HTTP methods for some routes, potentially confusing API consumers debugging integration issues.

**Recommendation**: Audit all error messages in single-route helpers to ensure they accurately describe the expected method and route.

---

### 8. Documentation contradictions between inline JSDoc and .mdx docs

**Files**: `docs/content/docs/(root)/runtime-server-adapter.mdx`, various source files

The .mdx documentation describes some adapter options that don't exist in the current implementation, and some implemented options are undocumented. Specific examples: lifecycle hook descriptions in docs don't match the actual hook signatures in code.

**Recommendation**: Reconcile documentation with implementation. Ensure every public API option is documented and every documented option exists.

---

### 9. Lifecycle hooks table in documentation is incomplete

**File**: `docs/content/docs/(root)/runtime-server-adapter.mdx`

The hooks documentation table doesn't cover all available hooks (`onError` is missing from the table but referenced in examples). The hook parameter types described in docs don't match the TypeScript interfaces.

**Recommendation**: Auto-generate or manually update the hooks reference table to match the actual TypeScript types.

---

### 10. CORS credentials + wildcard origin edge case

**File**: `packages/v2/runtime/src/core/fetch-cors.ts`

Per the Fetch spec, `Access-Control-Allow-Origin: *` cannot be combined with `Access-Control-Allow-Credentials: true`. The CORS implementation handles this correctly in most paths, but the `resolveOrigin` function has a code path where a wildcard `allowedOrigins: ["*"]` combined with `credentials: true` could produce an invalid CORS response.

**Recommendation**: Add an explicit guard that reflects the request's `Origin` header instead of `*` when credentials are enabled, or throw a configuration error at setup time.

---

## Suggestions

### 11. Extract shared request/response utilities

Adapter files (`express.ts`, `hono.ts`, `node.ts`) each contain similar request normalization and response writing logic. Extract common patterns into a shared utility to reduce duplication and ensure consistent behavior.

### 12. Unify single-route and multi-route adapter patterns

`express.ts` / `express-single.ts` and `hono.ts` / `hono-single.ts` have significant code duplication. The single-route variants could be thin wrappers that configure the multi-route variant with a single route.

### 13. Add request ID propagation

No request ID is generated or propagated through the handler chain. This makes correlating logs across the fetch handler → router → handler → runtime pipeline difficult in production.

### 14. Add AbortSignal threading

The `Request.signal` from the incoming fetch request is not threaded through to the runtime's agent execution. This means client disconnects don't cancel in-progress LLM calls.

### 15. Express fetch bridge body handling could be simpler

`express-fetch-bridge.ts` has complex logic to detect and re-serialize pre-parsed bodies. Consider using `express.raw()` middleware on CopilotKit routes instead, avoiding the need to re-serialize.

### 16. Add integration test for CORS preflight flow

The CORS unit tests cover individual functions but there's no integration test that sends an actual `OPTIONS` preflight followed by the real request, verifying the full browser CORS handshake.

### 17. Add test for hook error propagation

No tests verify what happens when a lifecycle hook throws. Does the error propagate to the client? Is it logged? Is the request still handled?

### 18. Test coverage gap: SSE stream interruption

No tests simulate a client disconnecting mid-SSE-stream to verify cleanup behavior.

### 19. Test coverage gap: concurrent requests

No tests verify behavior under concurrent requests to the same adapter instance (race conditions in shared state, if any).

### 20. Node adapter should document `createCopilotNodeListener` vs `createCopilotNodeHandler` distinction

Both are exported but the difference isn't documented. One returns a request listener, the other returns a handler — clarify when to use which.

### 21. Consider making `basePath` normalization consistent

`fetch-router.ts` handles basePath with both prefix and suffix strategies, but the normalization (trailing slash handling) differs between them.

### 22. Deno and Bun examples lack error handling

The example files for Deno and Bun servers don't demonstrate error handling or graceful shutdown, unlike the Express and Node examples.

### 23. Changeset should note breaking changes for V1 consumers

If any V1 re-exports changed signatures or removed options, the changeset should call this out explicitly.

### 24. CI workflow `test_runtime-servers.yml` should run on PR changes to adapter files

The new CI workflow triggers aren't scoped to the relevant file paths, meaning unrelated PRs will run these tests unnecessarily, or relevant changes might not trigger them.

jpr5 (changes requested):
## Prior Review Comments (Resolved)

- **Use** `generateRandomId()` **from shared package** instead of custom UUID implementation — resolved by author (now imports `randomUUID` from `@copilotkitnext/shared`)

---

## Critical Issues

### 1. Fire-and-forget observable subscription in intelligence run handler

**File**: `packages/v2/runtime/src/handlers/intelligence/run.ts`

The observable from `runtime.process()` is subscribed but the subscription is never stored or cleaned up. If the client disconnects or the request is aborted, the observable continues running in the background, leaking resources.

**Recommendation**: Store the subscription and tie its lifecycle to the request/response — unsubscribe on client disconnect or abort signal.

---

### 2. Type mismatch: `CopilotRuntime` vs `CopilotRuntimeLike` across handler signatures

**Files**: `packages/v2/runtime/src/core/runtime.ts`, handler files in `packages/v2/runtime/src/handlers/`

Some handlers accept `CopilotRuntime` (concrete class) while the fetch handler and adapters use `CopilotRuntimeLike` (structural interface). This creates a hidden coupling — any runtime passed through adapters must actually be a full `CopilotRuntime` instance, but the type system doesn't enforce this consistently.

**Recommendation**: Standardize on `CopilotRuntimeLike` for all handler signatures, or document explicitly which handlers require the concrete class and why.

---

### 3. `RouteInfo` type at hooks.ts:36 uses weak typing that defeats discriminated union benefits

**File**: `packages/v2/runtime/src/core/hooks.ts:36`

The `RouteInfo` type used in lifecycle hooks doesn't carry the discriminated union tag from the route matching, so hook consumers can't narrow on the route type. This means hooks that need route-specific behavior must use unsafe type assertions.

**Recommendation**: Thread the discriminated union type through to the hooks interface so consumers can pattern-match on route kind.

---

## Important Issues

### 4. Empty catch blocks silently swallow SSE errors

**File**: `packages/v2/runtime/src/handlers/shared/sse-response.ts`

Multiple `catch` blocks in SSE streaming code are empty or only log at debug level. SSE connection failures, encoding errors, or stream corruption will be invisible to operators.

**Recommendation**: Log at `warn` or `error` level. Consider emitting a final SSE error event before closing the stream so clients can distinguish "stream ended normally" from "stream died."

---

### 5. Multiple `as any` type casts bypass TypeScript safety

**Files**: `packages/v2/runtime/src/endpoints/express-fetch-bridge.ts`, `node-fetch-handler.ts`, and others

At least 6 instances of `as any` were found across the new code. These defeat the purpose of TypeScript and hide potential runtime type mismatches.

**Recommendation**: Replace with proper typing. Where bridge code genuinely cannot type external input, use `unknown` with explicit narrowing instead of `any`.

---

### 6. Inconsistent or missing structured logging

**Files**: Various handler and adapter files

Some files use `console.log`, others use a structured logger, and several critical paths have no logging at all (e.g., CORS rejection, route not found, hook errors).

**Recommendation**: Adopt the project's existing logger consistently. Add structured logging for: route matching decisions, CORS policy evaluation, hook lifecycle, and error paths.

---

### 7. Error message text doesn't match actual error condition

**File**: `packages/v2/runtime/src/endpoints/single-route-helpers.ts`

The method validation error message references incorrect HTTP methods for some routes, potentially confusing API consumers debugging integration issues.

**Recommendation**: Audit all error messages in single-route helpers to ensure they accurately describe the expected method and route.

---

### 8. Documentation contradictions between inline JSDoc and .mdx docs

**Files**: `docs/content/docs/(root)/runtime-server-adapter.mdx`, various source files

The .mdx documentation describes some adapter options that don't exist in the current implementation, and some implemented options are undocumented. Specific examples: lifecycle hook descriptions in docs don't match the actual hook signatures in code.

**Recommendation**: Reconcile documentation with implementation. Ensure every public API option is documented and every documented option exists.

---

### 9. Lifecycle hooks table in documentation is incomplete

**File**: `docs/content/docs/(root)/runtime-server-adapter.mdx`

The hooks documentation table doesn't cover all available hooks (`onError` is missing from the table but referenced in examples). The hook parameter types described in docs don't match the TypeScript interfaces.

**Recommendation**: Auto-generate or manually update the hooks reference table to match the actual TypeScript types.

---

### 10. CORS credentials + wildcard origin edge case

**File**: `packages/v2/runtime/src/core/fetch-cors.ts`

Per the Fetch spec, `Access-Control-Allow-Origin: *` cannot be combined with `Access-Control-Allow-Credentials: true`. The CORS implementation handles this correctly in most paths, but the `resolveOrigin` function has a code path where a wildcard `allowedOrigins: ["*"]` combined with `credentials: true` could produce an invalid CORS response.

**Recommendation**: Add an explicit guard that reflects the request's `Origin` header instead of `*` when credentials are enabled, or throw a configuration error at setup time.

---

## Suggestions

### 11. Extract shared request/response utilities

Adapter files (`express.ts`, `hono.ts`, `node.ts`) each contain similar request normalization and response writing logic. Extract common patterns into a shared utility to reduce duplication and ensure consistent behavior.

### 12. Unify single-route and multi-route adapter patterns

`express.ts` / `express-single.ts` and `hono.ts` / `hono-single.ts` have significant code duplication. The single-route variants could be thin wrappers that configure the multi-route variant with a single route.

### 13. Add request ID propagation

No request ID is generated or propagated through the handler chain. This makes correlating logs across the fetch handler → router → handler → runtime pipeline difficult in production.

### 14. Add AbortSignal threading

The `Request.signal` from the incoming fetch request is not threaded through to the runtime's agent execution. This means client disconnects don't cancel in-progress LLM calls.

### 15. Express fetch bridge body handling could be simpler

`express-fetch-bridge.ts` has complex logic to detect and re-serialize pre-parsed bodies. Consider using `express.raw()` middleware on CopilotKit routes instead, avoiding the need to re-serialize.

### 16. Add integration test for CORS preflight flow

The CORS unit tests cover individual functions but there's no integration test that sends an actual `OPTIONS` preflight followed by the real request, verifying the full browser CORS handshake.

### 17. Add test for hook error propagation

No tests verify what happens when a lifecycle hook throws. Does the error propagate to the client? Is it logged? Is the request still handled?

### 18. Test coverage gap: SSE stream interruption

No tests simulate a client disconnecting mid-SSE-stream to verify cleanup behavior.

### 19. Test coverage gap: concurrent requests

No tests verify behavior under concurrent requests to the same adapter instance (race conditions in shared state, if any).

### 20. Node adapter should document `createCopilotNodeListener` vs `createCopilotNodeHandler` distinction

Both are exported but the difference isn't documented. One returns a request listener, the other returns a handler — clarify when to use which.

### 21. Consider making `basePath` normalization consistent

`fetch-router.ts` handles basePath with both prefix and suffix strategies, but the normalization (trailing slash handling) differs between them.

### 22. Deno and Bun examples lack error handling

The example files for Deno and Bun servers don't demonstrate error handling or graceful shutdown, unlike the Express and Node examples.

### 23. Changeset should note breaking changes for V1 consumers

If any V1 re-exports changed signatures or removed options, the changeset should call this out explicitly.

### 24. CI workflow `test_runtime-servers.yml` should run on PR changes to adapter files

The new CI workflow triggers aren't scoped to the relevant file paths, meaning unrelated PRs will run these tests unnecessarily, or relevant changes might not trigger them.

PR #3264, opened 2026-02-24
feat: e2e selectors for components

AlemTuzlak:
## What does this PR do?

## Related PRs and Issues

## Checklist


packages/v1/runtime/package.json
```diff
@@ -87,9 +87,6 @@
   },
   "peerDependencies": {
     "@anthropic-ai/sdk": "^0.57.0",
-    "@copilotkit/shared": "workspace:*",
```
tylerslaton, line 90: This seems like a mistake, am I missing something?

AlemTuzlak: @tylerslaton it wasn't, these shouldn't be peer deps

AlemTuzlak:
@ranst91 let's create a linear ticket about this to see if we can improve the e2e test suites, they were using css selectors so that's even worse than test ids, but I agree it should use a11y testing but that requires more careful changes than just adding a prop

PR #3272, opened 2026-02-25
feat: accept agentId parameter in useLangGraphInterrupt

AlemTuzlak:
## Summary
- Adds an optional `agentId` parameter to `useLangGraphInterrupt` so the hook can be scoped to a specific agent
- The explicit `agentId` takes priority over the chat configuration context, then falls back to `"default"`
- Updates the docs reference page with the new parameter

## Test plan
- [ ] Verify `useLangGraphInterrupt({ agentId: "my-agent", render: ... })` targets the correct agent
- [ ] Verify omitting `agentId` preserves existing behavior (reads from chat config context)
- [ ] Build passes (`nx run @copilotkit/react-core:build`)

### DIFF
--- a/packages/v1/react-core/src/hooks/use-langgraph-interrupt.ts
+++ b/packages/v1/react-core/src/hooks/use-langgraph-interrupt.ts
@@ -40,7 +40,8 @@ export function useLangGraphInterrupt<TEventValue = any>(
   actionRef.current = action;
 
   const existingConfig = useCopilotChatConfiguration();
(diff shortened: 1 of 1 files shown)


AlemTuzlak:
@mxmzb I just want to bring it more in line with our other hooks until it's completely deprecated, we have the new v2 `useInterrupt` hook that should be used instead of this one, but this makes it easier to maintain the hook until migration as now it defaults to the "default" agent which is not the behavior you might want

PR #3280, opened 2026-02-26
feat(react-core): type-safe agent IDs via Register augmentation

AlemTuzlak:
Type-safe `agentId` props for `@copilotkit/react-core/v2`. Augment `Register` with your agent map, and every hook and chat component rejects unknown agent IDs at compile time. With no augmentation, `agentId` stays `string`.

This PR does **not** ship a CLI. The generator that writes `.copilotkit/types.ts` and `.copilotkit/register.d.ts` belongs in Intelligence.

## What does this PR do?

- Adds an empty `Register` interface as the typegen augmentation point.
- Adds `AgentId`, which is `keyof Register["agents"]` when that map exists, otherwise `string`.
- Wires `AgentId` through v2 hooks, renderers, `CopilotChat`, `CopilotThreadsDrawer`, and `CopilotChatConfigurationProvider`.
- Merges latest `main` and removes `@copilotkit/cli` from this repo.

## Related PRs and Issues

- CLI work: Intelligence repo (one level above this repo). Do not land a generator here.

## Checklist

## Testing

**Commands run.** Merge only. No `nx` test run in this worktree (no `node_modules` after the checkout). Lefthook was skipped on the merge commit for the same reason.

**Manual test.**

1. In an app that uses `@copilotkit/react-core/v2`, add this declaration file and include it in `tsconfig.json`:

```ts
// .copilotkit/register.d.ts
import type { Agents } from "./types.js";

declare module "@copilotkit/react-core/v2" {
  interface Register {
    agents: Agents;
  }
}
```

2. Add `.copilotkit/types.ts`:

```ts
export type Agents = {
  planner: { description: "Plans events" };
};
```

3. Write `useAgent({ agentId: "planner" })`. Type-check must pass.
4. Write `useAgent({ agentId: "typo" })`. Type-check must fail.
5. Remove the declaration file. `agentId` must accept any `string` again.

**How this PR makes testing easy.** This is a compile-time contract. The steps above are the proof. There is no new runtime test, because without a generated `Register` map the type is still `string`.

## Risk / rollback

Low. Without a `Register` augmentation, every `agentId` is still `string`, so existing apps keep compiling. Rollback is a revert of this PR.

## Public API change

**Before**

```ts
useAgent({ agentId: "anything" });
```

**After**

```ts
// Default (no typegen files): still any string
useAgent({ agentId: "anything" });

// With Register augmented from generated types:
useAgent({ agentId: "planner" }); // ok
useAgent({ agentId: "typo" });    // type error
```

## Summary by CodeRabbit

* **New Features**
  * Added type-safe agent ID support across chat components, hooks, tools, renderers, and configuration.
  * Agent IDs can be constrained to known application-defined values while remaining compatible with string IDs.
  * Added an extensibility point for registering application-specific agent definitions.

* **Improvements**
  * Standardized agent ID typing throughout the public API.
  * Added support for agent-specific generic types in frontend tools and human-in-the-loop integrations.


BenTaylorDev (commented):
Reviewed against current `main`. Verdict: **needs revision**. Up front: the mechanism itself works — I verified that rather than assuming it.

## Verified working

I applied the type mechanism on top of current `main`, built `@copilotkit/react-core`, and type-checked a consumer against the **published** `dist/v2/index.d.mts` rather than the source. That distinction matters, because tsdown bundles the v2 entry into a single name-mangled chunk and re-exports `Register` as an alias. Augmentation still merges through it:

```
WITH    declare module "@copilotkit/react-core/v2" { interface Register { agents: Agents } }
        AgentId === "planner"

WITHOUT the augmentation
        probe.ts(6,7): Type 'false' is not assignable to type 'true'
        probe.ts(7,7): Type 'true' is not assignable to type 'never'    -> AgentId === string
```

Both assertions flip when the augmentation is removed, so the narrowing and the `string` fallback are each real.

## Blocking

**1. `useAgent` narrows the wrong field.** `use-agent.tsx:70` applies `AgentId` to the thread-scoped `agentId`. The doc comment on that same property defines it as the name to register a *new* proxied agent under, one that "must not already be taken" — by definition a name absent from a generated agent map. Meanwhile `runtimeAgentId` at `use-agent.tsx:100`, the id that genuinely belongs in the map, stays `string`. The hook's own error strings at `:167` and `:197` suggest `agentId: "chat-1"`, which the narrowing then rejects.

**2. The library breaks its own round trip.** `Thread.agentId` is `string` (`use-threads.tsx:30`), so feeding a listed thread back into `useThreads` or `CopilotChat` stops compiling:

```
error TS2322: Type 'string' is not assignable to type '"planner"'.
```

`CopilotThreadsDrawer` lists threads and this PR narrows its `agentId` prop, so that is a supported flow today.

**3. No escape hatch for a runtime-derived id.** An agent chosen from a URL parameter, an API response, or a `<select>`:

```
error TS2322: Type 'string' is not assignable to type '"planner"'.
```

`AgentId | (string & {})` is the usual idiom here. Hard narrowing may well be the intent, but nothing documents it and there is no opt-out.

Those three are really one decision — closed union, or union plus `(string & {})`. Everything else follows from it.

**4. No test.** 22 files, zero tests. The package already has the harness: `src/v2/__tests__/headless-type-exports.test-d.ts` and `hooks/__tests__/use-default-render-tool-types.test-d.ts`. A `.test-d.ts` pinning both directions is cheap here. Without one, a later refactor can revert `AgentId` to `string` and CI stays green.

That risk is not theoretical. The CLI's `copilotkit typegen` already emits exactly this augmentation, and against a `react-core` that has no `Register` it type-checks completely clean and does nothing at all. Neither side notices today.

**5. No docs.** A new public augmentation contract ships with no shell-docs page.

## Rebase notes

The branch is stale against `main` and conflicts. Two things to watch while resolving:

- `use-render-tool.tsx` conflicts at three sites where `main` widened `render` to return `React.ReactElement | null` (`:41`, `:72`, `:109`). Taking the branch side reverts that on all three overloads.
- `use-human-in-the-loop.tsx` conflicts because `main` moved to `useLayoutEffect`. `use-suggestions.tsx` conflicts because `main` added `useDefaultAgentId`.

The `import {…}` to `import type {…}` hunks already landed on `main` and auto-merge away, so they can come out of the diff and leave just the feature.

## Nit

`register.ts` and `types/copilotkit-types.ts` carry `biome-ignore` comments. This repo lints with oxlint and has no other `biome-ignore` anywhere in the tree. I ran oxlint on both files — 0 warnings, 0 errors — so they can go.

BenTaylorDev (commented):
Re-checked at `42bbfa5` against current `main`. The new commit fixes a real inference bug and adds the type-test harness, but four of the five findings are still open, and three of them are one decision.

**Closed:** the agent generic no longer widens to `undefined`, and the two new `.test-d.ts` files pin that.

**1. `useAgent` still narrows the wrong field.** `use-agent.tsx:70` is `agentId: AgentId` on the thread-scoped variant, whose own doc comment says the name "must not already be taken" — by definition a name absent from the generated map. `runtimeAgentId` at `:100`, the id that does belong in the map, is still `string`. The type rejects the ids this overload is for and constrains nothing on the one it should.

**2. The round trip still does not compile.** `Thread.agentId` is `string` (`use-threads.tsx:31`), `UseThreadsInput.agentId` is `AgentId` (`:53`). Both lines are unchanged since my last pass, where I measured:

```
error TS2322: Type 'string' is not assignable to type '"planner"'.
```

`CopilotThreadsDrawer` lists threads and this PR narrows its `agentId` prop, so that is a supported flow today.

**3. No escape hatch.** `AgentId` (`types/copilotkit-types.ts:10`) is still a closed union, so an id from a URL parameter, an API response or a `<select>` needs a cast. `AgentId | (string & {})` is the usual idiom. A closed union may well be the intent — then it needs stating, and 1 and 2 resolved in its terms.

**4. Docs.** Still no shell-docs page for a new public augmentation contract.

On the tests: both new files assert that things resolve to `AgentId`, which holds whether `AgentId` is a union or `string`. What is not pinned is the part that can silently regress — that augmenting `Register` narrows, and that its absence falls back to `string`. That is the direction where the CLI already emits this augmentation against a `react-core` that has no `Register`, type-checks completely clean, and does nothing at all.

Nit, still present: the `biome-ignore` comments in `register.ts:1` and `types/copilotkit-types.ts`. oxlint is clean on both files.

1, 2 and 3 are one decision — closed union, or union plus `(string & {})`. Everything else follows from whichever you pick.

PR #3286, opened 2026-02-26
feat(core): add copilotkit.runTool() for programmatic tool execution

tylerslaton:
Adds a new `runTool()` method to `CopilotKitCore` that allows programmatic execution of registered frontend tools without requiring an LLM turn. This enables UI-driven tool invocations (e.g., button clicks triggering tool handlers) while maintaining proper message history and subscriber notifications.

Key changes:
- Extract `executeToolHandler` helper from duplicated logic in `executeSpecificTool` to share with `runTool`
- Add `runTool()` on `RunHandler` with tool/agent lookup, message creation, handler execution, and optional follow-up support
- Add `TOOL_NOT_FOUND` and `AGENT_NOT_FOUND` error codes
- Add `CopilotKitCoreRunToolParams` and `CopilotKitCoreRunToolResult` types
- Add delegation method on `CopilotKitCore`
- Add 16 comprehensive tests covering all runTool behaviors


packages/v2/core/src/core/run-handler.ts
```diff
@@ -284,121 +320,161 @@ export class RunHandler {
...
+        const result = await tool.handler!(parsedArgs as any, {
+          toolCall: toolCall as any,
+          agent,
+        });
+        if (result === undefined || result === null) {
+          toolCallResult = "";
+        } else if (typeof result === "string") {
+          toolCallResult = result;
+        } else {
+          toolCallResult = JSON.stringify(result);
```
AlemTuzlak: this code snippet has so many type casts, can we maybe remove this, llms pick this up constantly and think it's okay to do

tylerslaton: Yep, let me give that a try.

PR #3310, opened 2026-02-27
fix: remove server-side runtime dependency from runtime-client-gql

### LINKED ISSUE #3225: 🐛 Bug: Built Bundle Size Explosion with CopilotKit v1.5+
### ♻️ Reproduction Steps

1. Build an application using CopilotKit dependencies with v1.5.x
2. Check built js bundle size
...

### ✅ Expected Behavior

We used version 1.1.x to build a single JS IIFE bundle with Vite that includes a custom-built chat ontop of CopilotKit. After building we'd see ~600kb of code after brotli compression. We do not use Next.js and instead use Express as the backend for frontend adapter that communicates with our ADK backend.

### ❌ Actual Behavior

After updating to version 1.5.x, the new resulting bundle has exploded in size to ~4.3mb after build. Investigation reveals that there's new dependencies that are exported at the top-level via `*` barrel export, not allowing webpack/rolldown to treeshake dependecies out. This causes:
1. Backend JS to be bundled into our final build
2. Large 10mb+ packages to be bundled into builds that are not used
3. No way to remove server-side code from our frontend only build

### 𝌚 CopilotKit Version

```shell
├── @copilotkit/react-core@1.51.3
├── @copilotkit/react-ui@1.51.3
├── @copilotkit/runtime-client-gql@1.51.3
```

### 📄 Logs (Optional)

```shell
n/a
```

jpr5:
## Summary

- `@copilotkit/runtime-client-gql` declared `@copilotkit/runtime` as a dependency, but never imports from it at runtime
- This caused barrel re-exports to pull server-side code into client bundles (600KB → 4.3MB)
- Removed the package.json dependency and added an Nx `implicitDependencies` entry to preserve build ordering (graphql-codegen needs the schema file that runtime generates)

Fixes #3225

## Test plan

- [x] Verified no source imports from `@copilotkit/runtime` in runtime-client-gql
- [x] Nx build ordering preserved via `implicitDependencies`
- [x] Full build passes (including graphql-codegen)
- [x] All tests pass


justinnzhang:
🙏

PR #3311, opened 2026-02-27
fix: trigger agent invocation on MCP ui/message events

### LINKED ISSUE #3216: 🐛 Bug: `ui/message` event triggered from MCP apps updates the chat but doesn't invoke the agent
### ♻️ Reproduction Steps

While trying out the get-time app [example](https://github.com/modelcontextprotocol/ext-apps/blob/main/examples/basic-server-react/src/mcp-app.tsx), I notice that the `app.sendMessage` triggers adding the chat message on the UI but doesn't send it to the remote agent. 

Is there any way i can subscribe on my own and trigger `runAgent` on my own? 

### ✅ Expected Behavior

`app.sendMessage` should update UI and also trigger the chat. 

### ❌ Actual Behavior

`app.sendMessage` only updates UI.

### 𝌚 CopilotKit Version

```shell
"@copilotkitnext/react": "^1.51.3"
```

### 📄 Logs (Optional)

```shell

```

jpr5:
## Summary

- When an MCP app sends a `ui/message` event via `app.sendMessage()`, the message was added to the chat UI but the agent was never invoked to process it
- Now `copilotkit.runAgent()` is called through the proper RunHandler pipeline (not direct `agent.runAgent()`)
- Agent invocation is configurable via a `followUp` parameter in the `ui/message` params:
  - `followUp: true` → invoke agent after adding message
  - `followUp: false` → add message only (display-only)
  - Not specified → defaults to `true` for user-role messages, `false` for assistant-role

Fixes #3216

## Test plan

- [x] 5 new tests covering: user message invokes agent, assistant message doesn't, default role behavior, explicit followUp=false skips agent, explicit followUp=true on assistant invokes agent
- [x] All 20 tests in MCPAppsActivityRenderer pass
- [x] Build passes

### DIFF
--- a/packages/v2/react/src/components/MCPAppsActivityRenderer.tsx
+++ b/packages/v2/react/src/components/MCPAppsActivityRenderer.tsx
@@ -546,12 +546,27 @@ export const MCPAppsActivityRenderer: React.FC<MCPAppsActivityRendererProps> =
                         .map((c) => c.text)
                         .join("\n") || "";
 
+                    const role =
+                      (params.role as "user" | "assistant") || "user";
+
                     if (textContent) {
                       currentAgent.addMessage({
                         id: crypto.randomUUID(),
-                        role: (params.role as "user" | "assistant") || "user",
+                        role,
                         content: textContent,
                       });
+
+                      // Invoke the agent for user messages so the LLM
+                      // processes the message, matching CopilotChat behavior.
+                      // Assistant messages are display-only and don't need a run.
+                      if (role === "user") {
+                        currentAgent.runAgent().catch((err) => {
+                          console.error(
+                            "[MCPAppsRenderer] ui/message runAgent error:",
+                            err,
+                          );
+                        });
+                      }
                     }
                     sendResponse(msg.id, { isError: false });
                   } catch (err) {
--- a/packages/v2/react/src/components/MCPAppsActivityRenderer.tsx
+++ b/packages/v2/react/src/components/MCPAppsActivityRenderer.tsx
@@ -543,6 +543,7 @@ export const MCPAppsActivityRenderer: React.FC<MCPAppsActivityRendererProps> =
                     const params = msg.params as {
                       role?: string;
                       content?: Array<{ type: string; text?: string }>;
+                      followUp?: boolean;
                     };
 
                     // Extract text content from the message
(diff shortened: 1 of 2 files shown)


packages/v2/react/src/components/MCPAppsActivityRenderer.tsx
```diff
@@ -272,6 +273,11 @@ export const MCPAppsActivityRenderer: React.FC<MCPAppsActivityRendererProps> =
     const agentRef = useRef(agent);
     agentRef.current = agent;
 
+    // Get copilotkit core instance for proper RunHandler pipeline
+    const { copilotkit } = useCopilotKit();
+    const copilotkitRef = useRef(copilotkit);
+    copilotkitRef.current = copilotkit;
```
AlemTuzlak, line 279: the ref thing seems a bit redundant? you assign to it and then call it underneath, any reason why you don't call copilotkit on line 571?

packages/v2/react/src/components/MCPAppsActivityRenderer.tsx
```diff
@@ -537,6 +543,7 @@ export const MCPAppsActivityRenderer: React.FC<MCPAppsActivityRendererProps> =
                     const params = msg.params as {
                       role?: string;
                       content?: Array<{ type: string; text?: string }>;
+                      followUp?: boolean;
                     };
```
AlemTuzlak, line 547: I know this was here before but I'd argue this is the root cause of the bug, us telling typescript we know better than it does and forgetting to implement the functionalities that come with the params, this should be typesafe and not cast with as.

packages/v2/react/src/components/MCPAppsActivityRenderer.tsx
```diff
@@ -546,12 +553,31 @@ export const MCPAppsActivityRenderer: React.FC<MCPAppsActivityRendererProps> =
                         .map((c) => c.text)
                         .join("\n") || "";
 
+                    const role =
+                      (params.role as "user" | "assistant") || "user";
```
AlemTuzlak, line 557: let's see if there's a way to avoid casting here

AlemTuzlak (changes requested):

jpr5:
Closing — targets old `packages/v2/react/` path which no longer exists after repo restructure.

However, **the bug is still present** in the current code at `packages/react-core/src/v2/components/MCPAppsActivityRenderer.tsx:524-562`. The `ui/message` handler calls `addMessage()` but never invokes `runAgent()` — the agent doesn't process MCP-sent messages. Will open a new PR against the current file structure.

jpr5:
New PR with the fix against current codebase: #3705

PR #3315, opened 2026-02-28
fix: guard against non-object tool arguments from LLMs

giulio-leone:
## Problem

When an LLM returns non-object values (e.g. empty string `""`) as tool call arguments, `JSON.parse()` succeeds but produces a non-object value. This value is stored in CopilotKit's conversation history. On subsequent requests, providers like Anthropic reject the non-dictionary `tool_use.input` with a 400 error:

```
Anthropic API error: 400 {..."message":"messages.39.content.0.tool_use.input: Input should be a valid dictionary"}
```

The conversation becomes permanently broken because every subsequent LLM call includes the invalid message.

## Root Cause

`JSON.parse('""')` → `""` (a string, not an object). No validation ensures the parsed result is actually a `Record<string, unknown>` before it's stored as tool arguments.

## Fix

| Package | File | Change |
|---------|------|--------|
| `@copilotkit/runtime` (v1) | `service-adapters/conversion.ts` | `ensureObjectArgs()` wrapper around `JSON.parse()` |
| `@copilotkit/runtime-client-gql` (v1) | `client/conversion.ts` | Validate `getPartialArguments()` result |
| `@copilotkitnext/core` (v2) | `core/run-handler.ts` | Guard both specific and wildcard tool arg parsing |
| `@copilotkitnext/agent` (v2) | `index.ts` | `ensureObjectArgs()` for tool call input conversion |

## Tests

All 748 tests pass (2 consecutive clean runs).

Fixes #3300

PR #3320, opened 2026-03-01
fix(anthropic): validate tool_use input is object before sending to API

### LINKED ISSUE #3300: 🐛 Bug: Anthropic error 400 "Input should be a valid dictionary" when LLM returns non-object args
### ♻️ Reproduction Steps

We are using Anthropic adapter with local runtime

Occasionally the LLM returns tool use with `args` that are not an object, in our case it returns empty JSON string which gets saved in the CopilotKit conversation like this:

```
{
    "actionExecutionMessage": {
      "arguments": "\"\"",
      "name": "upsert-language-variant-mapi"
    },
    "createdAt": "2026-02-27T08:43:39.525Z",
    "id": "ck-ef174282-d8fb-4d11-998a-6074c1eb494d"
  },
```

The problem is that Anthropic API expects the arguments to be an object, not a simple value or array, and throws HTTP 400

```
"originalError": {
                    "message": "Anthropic API error: 400 {\"type\":\"error\",\"error\":{\"type\":\"invalid_request_error\",\"message\":\"messages.39.content.0.tool_use.input: Input should be a valid dictionary\"},\"request_id\":\"req_011CYWfkWgkbirSykhqMJWgC\"}",
                    "extensions": {
                        "name": "CopilotError",
                        "statusCode": 503,
                        "code": "CONFIGURATION_ERROR",
                        "visibility": "banner",
                        "severity": "critical",
                        "troubleshootingUrl": "https://docs.copilotkit.ai/troubleshooting/common-issues#i-am-getting-a-network-errors--api-not-found",
                        "originalError": {
                            "message": "Anthropic API error: 400 {\"type\":\"error\",\"error\":{\"type\":\"invalid_request_error\",\"message\":\"messages.39.content.0.tool_use.input: Input should be a valid dictionary\"},\"request_id\":\"req_011CYWfkWgkbirSykhqMJWgC\"}",
                            "stack": "Error\n    at new CopilotKitError (/app/node_modules/.pnpm/@copilotkit+shared@1.10.6/node_modules/@copilotkit/shared/dist/index.js:359:18)\n    at new CopilotKitLowLevelError (/app/node_modules/.pnpm/@copilotkit+shared@1.10.6/node_modules/@copilotkit/shared/dist/index.js:464:5)\n    at convertServiceAdapterError (/app/node_modules/.pnpm/                        }
...
                    }
                }
```

Probably the only stable way to reproduce it is to fake such response in the adapter

We tracked the problem down to here [CopilotKit/packages/runtime/src/service-adapters/conversion.ts](https://github.com/CopilotKit/CopilotKit/blob/c981974deeac2eb1aa6379d28213dec4e43b7e0f/packages/v1/runtime-client-gql/src/client/conversion.ts#L40)

Where the string version of the arguments retrieved from LLM is converted to input which is typed as `Record`, but it doesn't validate the type, so even other types of values that are properly serialized to JSON, e.g. empty string mentioned above passes the conversion

### ✅ Expected Behavior

CopilotKit's code, or the adapter handles non-compliant shapes of tool args in messages, e.g. by fallback to empty object, and let's the tool handle the invalid inputs

### ❌ Actual Behavior

When this happens, the conversation gets indefinitely errorneous as any subsequent LLM call keeps this invalid request data

### Screenshots

_No response_

### 𝌚 CopilotKit Version

```shell
1.10.6 but the same code seems to be in the latest version
```

### 📄 Logs (Optional)

```shell

```

sxu75374:
## Summary

When an LLM returns non-object tool call arguments (e.g., an empty string `""`), the Anthropic adapter forwards them as-is to the API, causing HTTP 400:

```
messages.X.content.0.tool_use.input: Input should be a valid dictionary
```

## Root Cause

In `convertMessageToAnthropicMessage()` (`utils.ts`), `message.arguments` is passed directly as `tool_use.input` without validating that it's a plain object. When the LLM hallucinates non-object arguments (empty string, array, null, etc.), Anthropic rejects the request.

## Fix

Added a type check before passing `input` to ensure it's always a plain object (`Record<string, any>`). Falls back to `{}` for any non-object value.

## Tests

Added 8 test cases covering:
- Valid object arguments (pass-through)
- Empty string, non-empty string → `{}`
- `null`, `undefined` → `{}`
- Array, number → `{}`
- Empty object `{}` (pass-through)

All existing tests pass.

Fixes #3300

Signed-off-by: sxu75374 <imshuaixu@gmail.com>

### DIFF
--- a/.changeset/fix-anthropic-tool-input.md
+++ b/.changeset/fix-anthropic-tool-input.md
@@ -1,0 +1,5 @@
+---
+"@copilotkit/runtime": patch
+---
+
+fix(anthropic): validate tool_use input is object before sending to API
--- a/packages/v1/runtime/src/service-adapters/anthropic/utils.ts
+++ b/packages/v1/runtime/src/service-adapters/anthropic/utils.ts
@@ -134,13 +134,22 @@ export function convertMessageToAnthropicMessage(
       ],
     };
   } else if (message.isActionExecutionMessage()) {
+    // Anthropic requires tool_use input to be a valid object (dict).
+    // If the LLM returned non-object arguments (e.g. empty string, array, null),
+    // fall back to an empty object to avoid 400 errors.
+    const input =
+      message.arguments &&
+      typeof message.arguments === "object" &&
+      !Array.isArray(message.arguments)
+        ? message.arguments
+        : {};
     return {
       role: "assistant",
       content: [
         {
           id: message.id,
           type: "tool_use",
-          input: message.arguments,
+          input,
           name: message.name,
         },
       ],
--- a/packages/v1/runtime/tests/service-adapters/anthropic/utils-tool-input-validation.test.ts
+++ b/packages/v1/runtime/tests/service-adapters/anthropic/utils-tool-input-validation.test.ts
@@ -1,0 +1,90 @@
+import { convertMessageToAnthropicMessage } from "../../../src/service-adapters/anthropic/utils";
+
+/**
+ * Tests for tool_use input validation in convertMessageToAnthropicMessage.
+ *
+ * Anthropic requires tool_use.input to be a valid dictionary (object).
+ * When an LLM returns non-object arguments (empty string, array, null, etc.),
+ * the converter must fall back to {} to avoid HTTP 400 errors.
+ *
+ * See: https://github.com/CopilotKit/CopilotKit/issues/3300
+ */
+
+function makeActionExecutionMessage(overrides: {
+  id?: string;
+  name?: string;
+  arguments?: any;
+}) {
+  return {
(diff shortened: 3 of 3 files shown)


BenTaylorDev:
Closing as superseded — tool-call arguments are now guarded before reaching the adapters on `main`: merged #3315 added `safeParseToolArgs` (`packages/shared/src/utils/index.ts`) applied upstream in `packages/runtime/src/service-adapters/conversion.ts`, so the value handed to the Anthropic adapter is already guaranteed to be an object. That covers this PR's adapter-level guard. Issue #3300 is closed as completed. Thanks @sxu75374!

PR #3327, opened 2026-03-02
fix(openai): force Chat Completions API in OpenAIAdapter.getLanguageModel()

### LINKED ISSUE #3317: 🐛 Bug: Second request incorrectly uses /responses endpoint instead of /chat/completions with OpenRouter
## Bug Description
  When using CopilotKit with OpenRouter, the first conversation request works correctly, but the second request fails because CopilotKit incorrectly sends the request to `/api/v1/responses` instead of `/api/v1/chat/completions`.

  ## Environment
  - CopilotKit version: 1.52.1
  - LLM Provider: OpenRouter
  - Model: google/gemini-2.5-flash-lite
  - Framework: Next.js 15.5.9 (App Router)

  ## Configuration
  ```typescript
  const openai = new OpenAI({
    apiKey: process.env.OPENROUTER_API_KEY,
    baseURL: "https://openrouter.ai/api/v1",
  });

  const serviceAdapter = new OpenAIAdapter({
    openai,
    model: "google/gemini-2.5-flash-lite",
  });

  Expected Behavior

  All requests should go to /api/v1/chat/completions

  Actual Behavior

  - First request: ✅ https://openrouter.ai/api/v1/chat/completions
  - Second request: ❌ https://openrouter.ai/api/v1/responses

  Error

  Invalid Responses API request
  url: 'https://openrouter.ai/api/v1/responses'

  Note

  OpenRouter has two separate APIs:
  - /api/v1/chat/completions - Standard Chat Completions API (OpenAI compatible)
  - /api/v1/responses - Responses API Beta (different request format)

  These endpoints don't auto-route and require different request formats.

### ✅ Expected Behavior

  All requests should go to /api/v1/chat/completions

### ❌ Actual Behavior

  - First request: ✅ https://openrouter.ai/api/v1/chat/completions
  - Second request: ❌ https://openrouter.ai/api/v1/responses

### Screenshots

_No response_

### 𝌚 CopilotKit Version

```shell
├── @copilotkit/react-core@1.52.1
├── @copilotkit/react-ui@1.52.1
├── @copilotkit/runtime@1.52.1
```

### 📄 Logs (Optional)

```shell

```

sxu75374:
## Summary

When using CopilotKit with OpenAI-compatible providers (e.g. OpenRouter), the second conversation request fails because `@ai-sdk/openai` auto-detects the Responses API and routes subsequent requests to `/responses` instead of `/chat/completions`.

## Root Cause

`OpenAIAdapter.getLanguageModel()` calls `provider(this.model)` which lets the AI SDK auto-select the API. For non-OpenAI providers like OpenRouter, `/responses` is a separate, incompatible endpoint.

## Fix

Use `provider.chat(this.model)` instead of `provider(this.model)` to explicitly force the Chat Completions API. This aligns `getLanguageModel()` with `process()` which already uses `openai.beta.chat.completions.stream()` (Chat Completions).

## Changes

- `openai-adapter.ts`: 1 line change + comment
- `openai-adapter-language-model.test.ts`: Updated mock to support `.chat()`, updated assertions, added OpenRouter-specific test case

## Testing

- Updated existing tests to verify `.chat()` is called
- Added dedicated test for OpenRouter base URL scenario

Fixes #3317

### DIFF
--- a/packages/v1/runtime/src/service-adapters/openai/openai-adapter.ts
+++ b/packages/v1/runtime/src/service-adapters/openai/openai-adapter.ts
@@ -136,7 +136,12 @@ export class OpenAIAdapter implements CopilotServiceAdapter {
       headers: options.defaultHeaders,
       fetch: options.fetch,
     });
-    return provider(this.model);
+    // Use .chat() explicitly to force the Chat Completions API.
+    // The default provider(model) auto-selects the Responses API for newer
+    // models, which breaks OpenAI-compatible providers (e.g. OpenRouter)
+    // that expose /responses as a separate, incompatible endpoint.
+    // See: https://github.com/CopilotKit/CopilotKit/issues/3317
+    return provider.chat(this.model);
   }
 
   private ensureOpenAI(): OpenAI {
--- a/packages/v1/runtime/tests/service-adapters/openai/openai-adapter-language-model.test.ts
+++ b/packages/v1/runtime/tests/service-adapters/openai/openai-adapter-language-model.test.ts
@@ -36,10 +36,14 @@ type _exhaustive =
       };
 const _check: _exhaustive = true;
 
-const { mockProviderFn, mockCreateOpenAI } = vi.hoisted(() => {
-  const mockProviderFn = vi.fn().mockReturnValue({ modelId: "test-model" });
+const { mockProviderFn, mockChatFn, mockCreateOpenAI } = vi.hoisted(() => {
+  const mockChatFn = vi.fn().mockReturnValue({ modelId: "test-model" });
+  const mockProviderFn = Object.assign(
+    vi.fn().mockReturnValue({ modelId: "test-model" }),
+    { chat: mockChatFn },
+  );
   const mockCreateOpenAI = vi.fn().mockReturnValue(mockProviderFn);
-  return { mockProviderFn, mockCreateOpenAI };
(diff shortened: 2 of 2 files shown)


BenTaylorDev:
Thanks for this, and sorry it went unreviewed for so long.

The diagnosis is correct and still holds on `main`. Verified against the installed `@ai-sdk/openai`:

```
provider(model)  -> OpenAIResponsesLanguageModel
provider.chat()  -> OpenAIChatLanguageModel
```

So `getLanguageModel()` does hand back a Responses-API model, and providers that only expose `/chat/completions` do break. Tracked in OSS-764.

Closing rather than asking for a rebase:

1. The tree moved — `packages/v1/runtime/` is gone; the adapter now lives at `packages/runtime/src/service-adapters/openai/openai-adapter.ts`.

2. `GroqAdapter.getLanguageModel()` has the identical `createOpenAI(...)` + `provider(this.model)`, and Groq is chat-completions-only — so this bug has two sites, and your PR found the first one.

3. Forcing `.chat()` unconditionally also removes Responses-API capability for users on real OpenAI, since `getLanguageModel()` feeds the default `BuiltInAgent`. We need to pick between switching on a non-OpenAI `baseURL` and exposing an explicit option, and that call is ours.

The ticket carries all three points and credits this PR.

Issue #3334, opened 2026-03-02
🐛 Bug: CopilotSidebar v2 — No external open/close control


deepakbatham572:
### ♻️ Reproduction Steps

CopilotSidebar v2 locks setModalOpen inside its own CopilotChatConfigurationProvider. There is no prop, ref, or callback to close it from outside. This is a regression from v1 which had open, onSetOpen, and clickOutsideToClose props.

Package: @copilotkitnext/react

Root cause: CopilotSidebarView always passes isModalDefaultOpen to its internal provider, so shouldCreateModalState is always true — the state is permanently sealed inside the component.

Fix requested: Add a controlled open / onOpenChange prop, or a forwarded ref with close() handle.

### ✅ Expected Behavior

in v2 the copilotsidebar should closed 

### ❌ Actual Behavior

in v2 the copilotsidebar is not closing at all 

### 𝌚 CopilotKit Version

```shell
│ ├── @copilotkit/runtime@1.52.0
│ ├── @copilotkit/react-core@1.52.0
│ ├── @copilotkit/react-ui@1.52.0
│ ├── @copilotkitnext/react@1.52.0
```

### 📄 Logs (Optional)

```shell

```

deepakbatham572:
Hi @mxmzb , @tylerslaton  could you please help on this , it's more than 3 week we have not received any resolution for this also one PR is raised but that is also closed.

could you please check on this ?

deepakbatham572:
Hi @mxmzb , @tylerslaton 
Do we have any update on that ?

PR #3340, opened 2026-03-02
Fix` isModalDefaultOpen` always `true` in `useCopilotChatConfiguration`

mxmzb:
## What does this PR do?

- Fix nested `CopilotChatConfigurationProvider` ignoring inner `isModalDefaultOpen` when parent already has modal state
- The `useState` initializer used `parentConfig?.isModalOpen ?? resolvedDefaultOpen`, so the parent's value always won via `??` — changed to use `resolvedDefaultOpen` directly since `shouldCreateModalState` already handles parent passthrough

## Related PRs and Issues

- https://copilotkit.slack.com/archives/C09C1BLEPC1/p1769021190286849
- https://github.com/CopilotKit/CopilotKit/issues/3334 (related but not resolving)


mxmzb:
Do not merge (for now).

deepakbatham572:
Hi @mxmzb , @jpr5 , @tylerslaton  could you please confirm the ETA for that we are facing issue because of this bug..

PR #3348, opened 2026-03-03
feat(runtime, react): Scaffold Intelligence runner and agent

BenTaylorDev:
## What does this PR do?

This PR adds an initial scaffolding of the CopilotKit Intelligence agent runner and React agent classes.

Both will be part of the new threads-enabled feature set, and will be iterated upon in subsequent PRs.

For this PR, the React agent is not tied into `useAgent`, so it cannot currently be interacted with.  

Also, in this PR, docs that describe the agent and runner architecture are not updated, as it is assumed that these classes may change substantially and rapidly in the very near term.

A big part of this PR is the integration of the Phoenix JS client SDK to enable the IntelligenceAgent and IntelligenceAgentRunner to connect to the Intelligence platform's Phoenix server.  That allows threaded AG-UI messages to be streamed from to and from the platform, instead of directly from agent runner to client agent hook.

The message structure for the Phoenix connection is still in flux.

Note: while the SQLite-backed agent runner is in a separate package, the IntelligenceAgentRunner was intentionally placed along side the `ProxiedCopilotRuntimeAgent` to make it that much easier to switch from the latter to the former if desired.

## Related PRs and Issues

- https://linear.app/copilotkit/issue/ENT-59
- https://linear.app/copilotkit/issue/ENT-54

## Checklist

### DIFF
--- a/packages/v2/core/src/intelligence-agent.ts
+++ b/packages/v2/core/src/intelligence-agent.ts
@@ -56,7 +56,7 @@ export class IntelligenceAgent extends AbstractAgent {
         : "http://localhost";
     const stopUrl = new URL(stopPath, new URL(runtimeUrl, origin));
 
-    void fetch(stopUrl.toString(), {
+    fetch(stopUrl.toString(), {
       method: "POST",
       headers: {
         "Content-Type": "application/json",


packages/v2/runtime/src/runner/intelligence.ts
```diff
@@ -0,0 +1,193 @@
...
+    channel
+      .join()
+      .receive("ok", () => {
+        // Ask the server to replay history via a CUSTOM event.
+        channel.push(EventType.CUSTOM, {
+          type: EventType.CUSTOM,
+          name: "connect",
+          value: { threadId },
+        });
+      })
```
MikeRyanDev: I think this could all be done in a more idomatic rxjs way. Avoid subscribe. Something like:

```ts
return new Observable(observer => {
  // join the channel and call the observer

  return () => { /* clean up channel join */ };
});
```

packages/v2/runtime/src/runner/intelligence.ts
```diff
@@ -0,0 +1,193 @@
...
+    }
+
+    return Promise.resolve(true);
+  }
+
+  private async executeAgentRun(
+    request: AgentRunnerRunRequest,
+    state: ThreadState,
+    threadId: string,
+  ): Promise<void> {
```
MikeRyanDev: This also strikes me as a method that could be more reactive and less async/await

packages/v2/runtime/src/runner/intelligence.ts
BenTaylorDev: Roger.  I'll make that change here and also in the `run` method above.

packages/v2/runtime/src/runner/intelligence.ts
BenTaylorDev: I'm not immediately picking up what you are putting down on this one.   There is an await-ed call to agent.runAgent - that could be wrapped into an observable, just for handling the async nature of it, which would mean we could remove Promise/async stuff from the the definition of `executeAgentRun`.  Is that what you mean?

MikeRyanDev: Yes, that's what I meant.

If changing the interface to be based on Observables is a no-go, then we could at least make the internals of this method be reactive and then use something like `lastValueFrom` to downlevel into promises.

BenTaylorDev: I've added a change for that.

packages/v2/core/src/intelligence-agent.ts
```diff
@@ -0,0 +1,237 @@
...
+
+    const { runtimeUrl, agentId, headers, credentials } = this.config;
+    const stopPath = `${runtimeUrl}/agent/${encodeURIComponent(agentId)}/stop/${encodeURIComponent(this.threadId)}`;
+    const origin =
+      typeof window !== "undefined" && window.location
+        ? window.location.origin
+        : "http://localhost";
+    const stopUrl = new URL(stopPath, new URL(runtimeUrl, origin));
+
+    void fetch(stopUrl.toString(), {
```
MikeRyanDev: `void` 👀

BenTaylorDev: Removed.

BenTaylorDev:
I got verbal approval from Mike that this is good to merge.

PR #3349, opened 2026-03-03
fix: replace @copilotkitnext with @copilotkit packages in big-hornets-shake changeset

mxmzb:
The `big-hornets-shake` changeset was using `@copilotkitnext/*` package names instead of `@copilotkit/*`, which caused the release workflow to fail due to mixing ignored and non-ignored packages.

Fixes: https://github.com/CopilotKit/CopilotKit/actions/runs/22644986711

### DIFF
--- a/.changeset/big-hornets-shake.md
+++ b/.changeset/big-hornets-shake.md
@@ -1,6 +1,5 @@
 ---
 "@copilotkit/runtime": patch
-"@copilotkit/docs": patch
 ---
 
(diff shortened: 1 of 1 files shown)


tylerslaton (changes requested):
We don't need the docs, it isn't a package

mxmzb:
> We don't need the docs, it isn't a package

True 😅

tylerslaton (approved):

tylerslaton:
Thanks for the fix!

PR #3358, opened 2026-03-04
Fix sidebar CSS override precedence in Chrome

### LINKED ISSUE #263: When overriding CSS properties, sometimes the overridden css property does not take precedence

mitre88:
## Summary
Fixes #263 by lowering selector specificity for sidebar-scoped UI styles so user overrides are more reliable across browsers (especially Chrome).

## What changed
Replaced high-specificity selectors like:
- `.copilotKitSidebar .copilotKitWindow`
- `.copilotKitSidebar .copilotKitHeader`
- `.copilotKitSidebar .copilotKitInputContainer`

with specificity-reduced equivalents using `:where(...)`, for example:
- `:where(.copilotKitSidebar) .copilotKitWindow`

This keeps behavior unchanged while making downstream CSS customizations easier to override.

## Validation
- `pnpm -C packages/v1/react-ui test`

### DIFF
--- a/packages/react-ui/src/css/header.css
+++ b/packages/react-ui/src/css/header.css
@@ -14,7 +14,7 @@
   z-index: 2;
 }
 
-.copilotKitSidebar .copilotKitHeader {
+:where(.copilotKitSidebar) .copilotKitHeader {
   border-radius: 0;
 }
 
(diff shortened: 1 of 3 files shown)


mitre88:
Closing this to keep my open PR queue focused on contributions that are currently reviewable and likely to proceed. This one has a blocking signal (failing-checks) and would need a deeper rework before it is worth maintainer time. Thanks for the review/context.

PR #3372, opened 2026-03-04
Fix/error-propagation

mxmzb:
## What does this PR do?

- Prevents error propagating out of CPK when an agent runtime is defined but not reachable
- Adds an `onError` callback to `CopilotChat` and `CopilotKitProvider`

## Related PRs and Issues

- https://copilotkit.slack.com/archives/C09C1BLEPC1/p1771375504525019

### DIFF
--- a/.changeset/wild-owls-mix.md
+++ b/.changeset/wild-owls-mix.md
@@ -1,0 +1,6 @@
+---
+"@copilotkit/react-core": patch
+---
+
+- Prevents error propagating out of CPK when an agent runtime is defined but not reachable
+- Adds an onError callback to CopilotChat and CopilotKitProvider


packages/v2/react/src/components/chat/CopilotChat.tsx
```diff
@@ -84,19 +125,8 @@ export function CopilotChat({
...
-        if (error instanceof AGUIConnectNotImplementedError) {
-          // connect not implemented, ignore
-        } else {
-          throw error;
-        }
-      }
-    };
     agent.threadId = resolvedThreadId;
-    connect(agent);
+    copilotkit.connectAgent({ agent });
```
tylerslaton: This looks like a change in behavior with the removal of await. Also, it may still be good to have a catch-all for thrown errors that trigger onError still. WDYT?

mxmzb: Nice catch!

PR #3377, opened 2026-03-05
Prevent binary artifacts in PRs

jpr5:
## Summary

- **`.gitignore`** — add patterns for common binary extensions (`.exe`, `.dll`, `.so`, `.dylib`, `.o`, `.obj`, `.a`, `.lib`, `.wasm`, `.dSYM/`)
- **CI workflow** (`static_check-binaries.yml`) — runs on all PRs to `main`, checks for binary file extensions, `build/` directories, `.dSYM/` directories, and individual files over 1 MB
- **Lefthook pre-commit** (`check-binaries`) — mirrors the CI check locally, catching binary artifacts, build outputs, and oversized files at commit time before they reach CI

Prompted by ag-ui PR #29 where ~15 MB of compiled macOS binaries were accidentally committed. This adds the same protection to CopilotKit.

## Test plan

- [ ] Verify `.gitignore` additions don't affect existing tracked files
- [ ] Confirm the `static / check binaries` workflow runs and passes on this clean PR
- [ ] Verify lefthook `check-binaries` fires on `git commit` and blocks binary/oversized files

### DIFF
--- a/lefthook.yml
+++ b/lefthook.yml
@@ -11,6 +11,43 @@ output:
 pre-commit:
   parallel: true
   commands:
+    check-binaries:
+      tags: binaries
+      run: |
+        VIOLATIONS=0
+        STAGED=$(git diff --cached --name-only --diff-filter=ACM)
+        [ -z "$STAGED" ] && exit 0
+
+        BINARIES=$(echo "$STAGED" | grep -iE '\.(exe|dll|so|dylib|o|obj|a|lib|wasm)$' || true)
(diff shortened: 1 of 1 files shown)


jpr5:
@AlemTuzlak Same gist as on AG-UI.

jpr5:
> can we also add the lefthook check, I think that one is really valuable as it stops you from committing it and having to fix it then

Back to you @AlemTuzlak

PR #3392, opened 2026-03-06
feat(runtime, react): Intelligence platform and integration implementation

BenTaylorDev:
## What does this PR do?
This PR is the result of a group integration and rapid iteration exercise to flesh out the features of the IntelligenceAgent, IntelligenceAgentRunner and CopilotKitIntelligence platform client.

Because the Intelligence* classes use websockets instead of SSE, there are new configuration options and branching paths for what Agent and AgentRunner is configured/used.

This PR also adds the useThreads hook, which encapsulates CRUD functionality for threads, including getting real time updates (also over websockets) and managing state using a dedicated RxJs-based store (with actions, selectors, etc.).

## Related PRs and Issues

- ENT-54
- ENT-57
- ENT-59

## Checklist


.github/workflows/static_check-binaries.yml
```diff
@@ -0,0 +1,87 @@
+name: static / check binaries
```
MikeRyanDev, line 1: What's this about?

dev-docs/architecture/setup-intelligence.md
```diff
@@ -0,0 +1,236 @@
+# Intelligence Setup Guide
```
MikeRyanDev, line 1: Let's get this updated with the latest names

examples/v2/react/demo/package.json
```diff
@@ -53,4 +53,4 @@
     "typescript": "^5",
     "vitest": "^3.2.3"
   }
-}
\ No newline at end of file
+}
```
MikeRyanDev, line 56: Revert changes to this file

packages/v1/runtime-client-gql/src/client/conversion.ts
```diff
@@ -243,8 +243,20 @@ function getPartialArguments(args: string[]) {
   try {
     if (!args.length) return {};
 
-    return JSON.parse(untruncateJson(args.join("")));
+    const parsed = JSON.parse(untruncateJson(args.join("")));
```
MikeRyanDev, line 246: Revert

packages/v1/runtime-client-gql/src/client/__tests__/conversion.test.ts
```diff
@@ -0,0 +1,64 @@
+import { describe, it, expect } from "vitest";
```
MikeRyanDev, line 1: Revert

packages/v1/runtime-client-gql/src/message-conversion/agui-to-gql.test.ts
```diff
@@ -135,6 +135,25 @@ describe("agui-to-gql", () => {
         aguiToolCallToGQLActionExecution(toolCall, "parent-id"),
       ).toThrow("Unsupported tool call type");
     });
+
+    test.each([
```
MikeRyanDev, line 139: Revert

packages/v1/runtime-client-gql/src/message-conversion/agui-to-gql.ts
```diff
@@ -198,8 +198,7 @@ export function aguiToolCallToGQLActionExecution(
       argumentsObj = JSON.parse(toolCall.function.arguments);
     } catch (error) {
       console.warn(
-        `Failed to parse tool call arguments for ${toolCall.function.name}:`,
-        error,
+        `[CopilotKit] Failed to parse tool arguments, falling back to empty object`,
```
MikeRyanDev, line 201: Revert all of these

packages/v1/runtime/src/graphql/message-conversion/agui-to-gql.test.ts
```diff
@@ -135,6 +135,25 @@ describe("agui-to-gql", () => {
         aguiToolCallToGQLActionExecution(toolCall, "parent-id"),
       ).toThrow("Unsupported tool call type");
     });
+
+    test.each([
```
MikeRyanDev, line 139: Revert all of this

packages/v1/runtime/src/graphql/message-conversion/agui-to-gql.ts
```diff
@@ -234,8 +234,7 @@ export function aguiToolCallToGQLActionExecution(
       argumentsObj = JSON.parse(toolCall.function.arguments);
     } catch (error) {
       console.warn(
-        `Failed to parse tool call arguments for ${toolCall.function.name}:`,
-        error,
+        `[CopilotKit] Failed to parse tool arguments, falling back to empty object`,
```
MikeRyanDev, line 237: Revert all of this

packages/v1/runtime/src/lib/runtime/copilot-runtime.ts
```diff
@@ -330,10 +332,7 @@ export class CopilotRuntime<const T extends Parameter[] | [] = []> {
   private runtimeArgs: CopilotRuntimeOptions;
   private _instance: CopilotRuntimeVNext;
 
-  constructor(
-    params?: CopilotRuntimeConstructorParams<T> &
-      PartialBy<CopilotRuntimeOptions, "agents">,
-  ) {
+  constructor(params?: CopilotRuntimeConstructorParams<T>) {
```
MikeRyanDev: Why did any of this change?

packages/v1/runtime/src/service-adapters/conversion.ts
```diff
@@ -67,3 +67,30 @@ export function convertGqlInputToMessages(
 
   return messages.filter((m) => m);
 }
+
+/**
+ * Safely parses a JSON string into a plain object for tool arguments.
+ * Mirrors the shared safeParseToolArgs in @copilotkitnext/shared (v2).
+ * Kept as a local copy because v1 does not import from v2 shared.
+ */
+function safeParseToolArgs(raw: string): Record<string, unknown> {
```
MikeRyanDev, line 76: Revert all of this?

packages/v1/runtime/src/service-adapters/conversion.test.ts
```diff
@@ -0,0 +1,56 @@
+import { describe, it, expect } from "vitest";
```
MikeRyanDev, line 1: Revert all of this

packages/v2/agent/src/index.ts
```diff
@@ -42,6 +42,7 @@ import { createOpenAI } from "@ai-sdk/openai";
 import { createAnthropic } from "@ai-sdk/anthropic";
 import { createGoogleGenerativeAI } from "@ai-sdk/google";
 import { randomUUID } from "crypto";
+import { safeParseToolArgs } from "@copilotkitnext/shared";
```
MikeRyanDev, line 45: Did we maybe grab unrelated commits or something...?

lefthook.yml
```diff
@@ -11,6 +11,43 @@ output:
 pre-commit:
   parallel: true
   commands:
+    check-binaries:
```
MikeRyanDev, line 14: Revert?

.gitignore
```diff
@@ -45,3 +45,15 @@ coverage
 .langgraph_api
 lefthook-local.yml
 .langgraph_api
+
+# Binary artifacts
```
MikeRyanDev, line 50: Revert?

pnpm-lock.yaml
```diff
@@ -2084,6 +2084,9 @@ importers:
       '@ai-sdk/openai':
```
MikeRyanDev, line 2250: If package.json didn't change, we should revert this back to the copy from `main`

.github/workflows/static_check-binaries.yml
maxkorp: This got added on main, and freaking github does a bad job showing diffs

https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-comparing-branches-in-pull-requests

(bad is subjective in this case)

packages/v1/runtime-client-gql/src/client/__tests__/conversion.test.ts
maxkorp: Nyet, ... diff's are just dumb

examples/v2/react/demo/package.json
maxkorp: Nyet, ... diff's are just dumb

packages/v1/runtime-client-gql/src/client/conversion.ts
maxkorp: Nyet, ... diff's are just dumb

packages/v1/runtime-client-gql/src/message-conversion/agui-to-gql.test.ts
maxkorp: Nyet, ... diff's are just dumb

packages/v1/runtime-client-gql/src/message-conversion/agui-to-gql.ts
maxkorp: Nyet, ... diff's are just dumb

lefthook.yml
maxkorp: Nyet, ... diff's are just dumb

.gitignore
maxkorp: Nyet, ... diff's are just dumb

pnpm-lock.yaml
BenTaylorDev: The package.json for the react package did change, to add phoenix.

packages/v1/runtime/src/lib/runtime/copilot-runtime.ts
BenTaylorDev: Gonna back this out.

packages/v2/core/src/core/agent-registry.ts
```diff
@@ -248,6 +268,11 @@ export class AgentRegistry {
             description: description,
             transport: this._runtimeTransport,
             credentials,
+            runtimeMode:
+              runtimeInfoResponse.mode === "intelligence"
+                ? "intelligence"
+                : "sse",
```
AlemTuzlak: shouldn't this be returned by the fetchRuntimeInfo correctly? I would expect this to just set the mode, otherwise if we in the future add a 3rd one this becomes a risk hazard

packages/v2/core/src/__tests__/phoenix-observable.test.ts
```diff
@@ -0,0 +1,174 @@
...
+const {
+  ɵobservePhoenixEvent$,
+  ɵobservePhoenixSocketHealth$,
+  ɵobservePhoenixSocketSignals$,
+  ɵphoenixChannel$,
+  ɵphoenixSocket$,
+} = await import("../utils/phoenix-observable");
+
+type PhoenixSocketSession =
+  import("../utils/phoenix-observable").ɵPhoenixSocketSession;
```
AlemTuzlak, line 27: why not just normal imports? I feel like our codebase is riddled with these and it encourages the LLM's to do these

packages/v2/core/src/agent.ts
```diff
@@ -146,59 +158,170 @@ export class ProxiedCopilotRuntimeAgent extends HttpAgent {
   }
 
   connect(input: RunAgentInput): Observable<BaseEvent> {
-    if (this.transport === "single") {
-      if (!this.singleEndpointUrl) {
-        throw new Error("Single endpoint transport requires a runtimeUrl");
+    if (this.runtimeMode !== "intelligence") {
```
AlemTuzlak: can we have smtn like this:
```ts
if(this.runtimeMode !== "intelligence"){
  return this.#handleSSEConnect();
}
return this.#handleIntelligenceConnect();
```

packages/v2/core/src/agent.ts
```diff
@@ -146,59 +158,170 @@ export class ProxiedCopilotRuntimeAgent extends HttpAgent {
...
+      ),
     );
-    return withAbortErrorHandling(transformHttpEventStream(httpEvents));
   }
 
   public run(input: RunAgentInput): Observable<BaseEvent> {
-    if (this.transport === "single") {
-      if (!this.singleEndpointUrl) {
-        throw new Error("Single endpoint transport requires a runtimeUrl");
+    if (this.runtimeMode !== "intelligence") {
```
AlemTuzlak: same question as above

packages/v2/core/src/agent.ts
```diff
@@ -146,59 +158,170 @@ export class ProxiedCopilotRuntimeAgent extends HttpAgent {
...
     }
 
-    const httpEvents = runHttpRequest(
-      `${this.runtimeUrl}/agent/${this.agentId}/connect`,
-      this.requestInit(input),
+    return defer(() => from(this.resolveDelegate())).pipe(
+      switchMap((delegate) =>
+        withAbortErrorHandling(
+          (
+            delegate as AbstractAgent & {
```
AlemTuzlak: is there a way to avoid the typecast here and  on 223?

packages/v2/core/src/agent.ts
```diff
@@ -146,59 +158,170 @@ export class ProxiedCopilotRuntimeAgent extends HttpAgent {
...
 
+  private async resolveDelegate(): Promise<AbstractAgent> {
+    await this.ensureRuntimeMode();
+
+    if (this.delegate) {
+      this.syncDelegate(this.delegate);
+      return this.delegate;
+    }
+
+    if (this.runtimeMode !== "intelligence") {
```
AlemTuzlak: I feel like htere should be a `const RUNTIME_MODE = { INTELLIGENCE: "intelligence", SSE: "sse" } as const` somewhere in the codebase that's used opposed to magical strings litered everywhere

packages/v2/core/src/agent.ts
```diff
@@ -146,59 +158,170 @@ export class ProxiedCopilotRuntimeAgent extends HttpAgent {
...
+    const response = await fetch(`${this.runtimeUrl}/info`, {
+      headers,
+      ...(this.credentials ? { credentials: this.credentials } : {}),
+    });
+    if (!response.ok) {
+      throw new Error(
+        `Runtime info request failed with status ${response.status}`,
+      );
+    }
+    return (await response.json()) as RuntimeInfo;
```
AlemTuzlak: small nit but we could extract these two branches into a private method:
```ts
const runtimeInfo = await this.fetchRuntime("url", body);
```
inside of there we check if response is okay and return the response.json

packages/v2/core/src/agent.ts
```diff
@@ -247,4 +367,44 @@ export class ProxiedCopilotRuntimeAgent extends HttpAgent {
...
+  private syncDelegate(delegate: AbstractAgent): void {
+    delegate.agentId = this.agentId;
+    delegate.description = this.description;
+    delegate.threadId = this.threadId;
+    delegate.setMessages(this.messages);
+    delegate.setState(this.state);
+
+    if ("headers" in delegate) {
+      (
+        delegate as AbstractAgent & { headers?: Record<string, string> }
```
AlemTuzlak: again, there should be no casts, especially for optional properties on the object

packages/v2/core/src/intelligence-agent.ts
```diff
@@ -30,19 +93,122 @@ export class IntelligenceAgent extends AbstractAgent {
...
+        subscriber ?? {},
+      ];
+
+      await this.onInitialize(input, subscribers);
+
+      self.activeRunDetach$ = new Subject<void>();
+      let resolveCompletion: (() => void) | undefined;
+      self.activeRunCompletionPromise = new Promise<void>((resolve) => {
+        resolveCompletion = resolve;
+      });
```
AlemTuzlak, line 171: correct me if I'm wrong but seems like this is going for:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/withResolvers

packages/v2/core/tsdown.config.ts
```diff
@@ -25,6 +26,7 @@ export default defineConfig([
       "rxjs",
       "zod",
     ],
+    inlineOnly: false,
```
AlemTuzlak: why are we using deprecated properties in bundling? what does this aim to do?

packages/v2/runtime/src/handlers/shared/json-response.ts
```diff
@@ -0,0 +1,12 @@
+const JSON_HEADERS = { "Content-Type": "application/json" } as const;
+
+export function jsonResponse(body: unknown, status = 200): Response {
+  return new Response(JSON.stringify(body), {
+    status,
+    headers: JSON_HEADERS,
+  });
+}
+
```
AlemTuzlak: why not just Response.json?

packages/v2/runtime/src/handlers/intelligence/run.ts
```diff
@@ -0,0 +1,147 @@
...
+        agentId,
+        sourceInput: input,
+        thread,
+        userId,
+      }).catch((nameError) => {
+        console.error("Failed to generate thread name:", nameError);
+      });
+    }
+  } catch (error) {
+    console.error("Failed to get or create thread:", error);
```
AlemTuzlak: why not use the logger from copilotkit/shared?

packages/v2/runtime/src/handlers/shared/agent-utils.ts
```diff
@@ -0,0 +1,136 @@
...
+        "string" ||
+        (requestBody as Record<string, unknown>).lastSeenEventId === null)
+    ) {
+      lastSeenEventId =
+        (requestBody as ConnectRequestBody).lastSeenEventId ?? null;
+    }
+
+    return { input, lastSeenEventId };
+  } catch (error) {
+    console.error("Invalid connect request body:", error);
```
AlemTuzlak: logger

packages/v2/runtime/src/index.ts
```diff
@@ -6,3 +6,6 @@ export * from "./runner";
 
 // Export transcription services
 export * from "./transcription-service/transcription-service";
+
+// Export intelligence platform client and types
+export * from "./intelligence-platform";
```
AlemTuzlak: can we not use export *, this can seriously impact bundler performance

BenTaylorDev: In our defense, that is convention for the rest of the file.   Also, there is a barrel in ./intelligence-platform that does narrow the exports, but its still "all the things other packages need".

It seems like the only way to usefully constrain in this file is to copy the contents of the barrel file into the level, but, it seems like that would just make maintenance harder for little gain in bundle size management.

Thoughts?

packages/v2/runtime/src/handlers/shared/agent-utils.ts
BenTaylorDev: Will fix.

packages/v2/runtime/src/handlers/intelligence/run.ts
BenTaylorDev: Will fix.

packages/v2/runtime/src/handlers/shared/json-response.ts
BenTaylorDev: Good catch.  I'll get those swapped out.

packages/v2/core/src/__tests__/phoenix-observable.test.ts
BenTaylorDev: This is some Jest stuff, where it waits to import the things its going to mock until after the lib is mocked below.  But you can't do that with types, so the type is imported the way it is to avoid the full import happening before it can be mocked.

packages/v2/core/tsdown.config.ts
BenTaylorDev: Ah, good catch.  It was added to fix a previous issue with Phoenix was being bundled, but that issue was resolved a different way, so this is vestigial.  I'll take it out.

packages/v2/core/src/intelligence-agent.ts
BenTaylorDev: It is a candidate for that, yes, but:
* we are targeting ES2022 and that requires ES2024
* this code was copied from AbstractAgent with the the only change from the base version being that events are not verified (this was a suggestion from Markus on (at least in the short term) handling some differences between SSE and WS observable behavior when connect()-ing).

So, I am inclined to leave it as is for now.

packages/v2/core/src/agent.ts
BenTaylorDev: That was because TS couldn't see the narrowing in the if statement.  Adding an explicit type predicate.

packages/v2/core/src/agent.ts
BenTaylorDev: I opted, instead, to factor out setting the URL and args, and just have one fetch + error check.

packages/v2/core/src/agent.ts
BenTaylorDev: Sure, I added an interface to get around them.

packages/v2/core/src/agent.ts
BenTaylorDev: There's an existing type union, but I can add named constants to avoid having the appearance of the magic strings.

packages/v2/core/src/agent.ts
BenTaylorDev: Done, and I changed the guard to be truthy.

packages/v2/core/src/agent.ts
BenTaylorDev: Done.

packages/v2/core/src/core/agent-registry.ts
BenTaylorDev: It's typed to be option in the RuntimeInfo response, but I think the intent was really to protect against undefined, so I changed it to nullish-coalesce on 'sse' (but as a constant).

packages/v2/core/src/agent.ts
maxkorp: I would agree with this, but gonna say we're not gonna block on it.

packages/v2/runtime/src/index.ts
maxkorp: I think underwhelming treeshaking performance has been an existing issue for us in the community so I'm thinking we follow Alem's thought here. It's a bit more maintenance, but in the age of claude go brrr I think the load picked up there is minimal.

BenTaylorDev: Yeah.  I can push that up real quick.

packages/v2/runtime/src/handlers/shared/agent-utils.ts
```diff
@@ -0,0 +1,136 @@
...
+    let lastSeenEventId: string | null = null;
+
+    if (
+      "lastSeenEventId" in (requestBody as Record<string, unknown>) &&
+      (typeof (requestBody as Record<string, unknown>).lastSeenEventId ===
+        "string" ||
+        (requestBody as Record<string, unknown>).lastSeenEventId === null)
+    ) {
+      lastSeenEventId =
+        (requestBody as ConnectRequestBody).lastSeenEventId ?? null;
```
AlemTuzlak, line 122: I think typeof requestBody === "object" could've avoided the crazy amount of casts here

packages/v2/runtime/src/handlers/handle-run.ts
```diff
@@ -1,182 +1,46 @@
...
-      status: 200,
-      headers: {
-        "Content-Type": "text/event-stream",
-        "Cache-Control": "no-cache",
-        Connection: "keep-alive",
-      },
-    });
+    return handleSseRun({ runtime, request, agent, input });
   } catch (error) {
     console.error("Error running agent:", error);
```
AlemTuzlak, line 45: we need a rule for this I think

packages/v2/runtime/src/runtime.ts
```diff
@@ -51,10 +54,48 @@ export interface CopilotRuntimeOptions extends CopilotRuntimeMiddlewares {
...
+  runner: AgentRunner;
+  a2ui: CopilotRuntimeOptions["a2ui"];
+  mcpApps: CopilotRuntimeOptions["mcpApps"];
+  intelligence?: CopilotKitIntelligence;
+  mode: RuntimeMode;
+}
+
+export interface CopilotSseRuntimeLike extends CopilotRuntimeLike {
+  intelligence?: undefined;
+  mode: "sse";
```
AlemTuzlak: I'd use the constant here as the type so if it drifts in the future or gets renamed this doesn't break

packages/v2/runtime/src/handlers/handle-run.ts
BenTaylorDev: For what aspect of it?

packages/v2/runtime/src/runtime.ts
BenTaylorDev: Ah, yeah, that was just a miss.

PR #3394, opened 2026-03-06
fix(runtime): add messages-tuple stream mode for LangGraph Platform streaming

jpr5:
## Summary
Supersedes #3201 — rebased onto current main and modernized for the current architecture.

- Fixes streaming not working when connecting to LangGraph Platform deployments (e.g., graphs built with `create_agent` from `langchain`)
- The `events` stream mode produces no `on_chat_model_stream` data on LangGraph Platform for these graphs, causing all content to arrive as a single `MessagesSnapshot` at the end
- Adds `messages-tuple` as a fallback stream mode and converts `[AIMessageChunk, metadata]` tuples into AG-UI text/tool-call events
- Includes dedup logic so when `events` mode works (local dev), `messages-tuple` events are skipped — no regressions

### How it works

Three method overrides in the CopilotKit `LangGraphAgent` wrapper:

1. **`runAgentStream`** — adds `messages-tuple` to the default stream modes (`["events", "values", "updates", "messages-tuple"]`)
2. **`handleStreamEvents`** — fixes a filter mismatch where `messages-tuple` mode produces SSE events with type `"messages"` (not `"messages-tuple"`), which the parent's `streamModes.includes()` check silently drops
3. **`handleSingleEvent`** — detects `[AIMessageChunk, metadata]` array tuples (vs regular event objects with `.event` property) and routes them to a new `handleMessagesTupleEvent` method
4. **`handleMessagesTupleEvent`** — converts chunk data into AG-UI `TEXT_MESSAGE_*` and `TOOL_CALL_*` events with proper start/content/end lifecycle

When `events` mode IS producing `on_chat_model_stream` data (normal local dev), `_eventsStreamActive` is set to `true` and messages-tuple events are automatically skipped to avoid duplicates.

### Concern: tight coupling to parent class internals

This fix works entirely through method overrides on `AGUILangGraphAgent` from `@ag-ui/langgraph`, which means it's tightly coupled to the parent class's internal method signatures and streaming pipeline. If the parent class refactors `runAgentStream`, `handleStreamEvents`, or `handleSingleEvent`, these overrides could silently break.

Ideally, the `messages-tuple` stream mode handling (including the `"messages"` filter fix) would live in `@ag-ui/langgraph` itself, since the mismatch between `messages-tuple` as a stream mode and `"messages"` as the SSE event type is a bug/gap in the parent class. This PR is a pragmatic workaround in the CopilotKit wrapper, but a proper upstream fix in ag-ui would be cleaner and less fragile.

## Test plan
- [ ] Build passes (`nx run @copilotkit/runtime:build`)
- [ ] Local LangGraph dev server: streaming still works via `events` mode (no regressions)
- [ ] LangSmith-hosted deployment: token-by-token streaming now works via `messages-tuple` fallback
- [ ] Tool calls stream correctly through `messages-tuple` mode

### DIFF
--- a/packages/v1/runtime/src/lib/runtime/agent-integrations/langgraph/agent.ts
+++ b/packages/v1/runtime/src/lib/runtime/agent-integrations/langgraph/agent.ts
@@ -1,4 +1,4 @@
-import { map } from "rxjs";
+import { map, Subscriber } from "rxjs";
 import { LangGraphEventTypes } from "../../../../agents/langgraph/events";
 import { RawEvent } from "@ag-ui/core";
 import {
@@ -11,7 +11,11 @@ import {
   StateEnrichment,
 } from "@ag-ui/langgraph";
 import { Message as LangGraphMessage } from "@langchain/langgraph-sdk/dist/types.messages";
-import { ThreadState } from "@langchain/langgraph-sdk";
+import { ThreadState, StreamMode } from "@langchain/langgraph-sdk";
+
+// Extract the RunAgentExtendedInput type from the parent class method signature
+// since it's not exported from @ag-ui/langgraph
+type RunAgentExtendedInput = Parameters<AGUILangGraphAgent["runAgentStream"]>[0];
 
 interface CopilotKitStateEnrichment {
   copilotkit: {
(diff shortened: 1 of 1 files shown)


jpr5:
Moving this fix upstream to @ag-ui/langgraph where it belongs — CopilotKit owns that repo too, so no need for a tightly-coupled workaround here.

jpr5:
Moved upstream to ag-ui where it belongs: https://github.com/ag-ui-protocol/ag-ui/pull/1257

PR #3399, opened 2026-03-07
fix(agent): auto-close reasoning lifecycle when SDK omits reasoning-end

### LINKED ISSUE #3323: 🐛 Bug: Reasoning / Thinking events in SSE stream cause agent to stall
## Summary

When Anthropic models produce reasoning/thinking tokens and CopilotKit emits `REASONING_*` events in the AG-UI SSE stream, the agent stalls permanently after the thinking phase completes. No further tool calls, text generation, or `RUN_END` events are emitted. The agent simply stops.

Stripping all `REASONING_*` events from the SSE response stream before they reach the frontend resolves the issue — the agent proceeds normally with tool calls and text output.

### Setup

We use `BuiltInAgent` from `@copilotkitnext/agent` with Anthropic models. Thinking/reasoning is configured in two complementary ways:

1. **Vercel AI SDK `providerOptions`** — passed to `createLanguageModelForAgent`:

   ```typescript
   providerOptions: {
     anthropic: {
       thinking: { type: 'adaptive' },
       effort: 'medium',
     }
   }
   ```

2. **HTTP-level injection** — a custom `fetch` wrapper that intercepts the outbound Anthropic API request and injects `thinking`, `max_tokens`, and model-level effort params into the JSON body (as a fallback / override for values the SDK may not pass through).

Both approaches work correctly at the Anthropic API level — thinking tokens are generated and streamed back.

### What we tried (chronologically)

1. **Verified `@ai-sdk/anthropic` version**: Initially had `2.0.56` which lacked `type: 'adaptive'` support in its Zod schema. Upgraded to `2.0.67` which natively supports adaptive thinking. **Did not fix the stall.**

2. **Verified CopilotKit version consistency**: Ensured all `@copilotkit/*` and `@copilotkitnext/*` packages are at `1.52.1` (the version that added reasoning support per [PR #3165](https://github.com/CopilotKit/CopilotKit/pull/3165)). **Did not fix the stall.**

3. **Increased `maxTokens`**: Set `maxTokens` to the model maximum (64000 for Sonnet 4.6) to rule out token budget exhaustion. **Did not fix the stall.**

4. **Disabled SDK-level thinking settings** (kept HTTP-level injection): Set `providerOptions.anthropic.thinking` to `undefined` so the SDK is unaware of thinking, but the HTTP-level fetch wrapper still injects `thinking: { type: 'adaptive' }` into the raw API request body. Anthropic still returns thinking tokens, and **the agent still stalled**. This proved the issue is not in how we configure the SDK, but in how CopilotKit handles the streamed reasoning events.

5. **Stripped all `REASONING_*` events from the SSE stream**: Added a `TransformStream` in our response pipeline that filters out any SSE `data:` line where the parsed JSON has `type` starting with `"REASONING"`. **This fixed the stall.** Agents proceed normally to tool calls and text generation.

### Workaround (current)

We strip `REASONING_*` events from the AG-UI SSE stream before the response reaches the frontend:

```typescript
// Inside our SSE TransformStream rewrite function
const parsed = JSON.parse(jsonPart)
if (parsed?.type?.startsWith?.('REASONING')) return '' // drop the line
```

This is behind a feature flag so we can re-enable reasoning events when a fix is available.

### Key observation

The Anthropic API call itself succeeds. Thinking tokens are generated and the model produces a complete response (tool calls and/or text). The issue is entirely in how CopilotKit's runtime translates the Vercel AI SDK's reasoning stream into AG-UI SSE events. It appears that `REASONING_END` is never emitted, which may leave the runtime's internal state machine stuck in a "reasoning" state, preventing it from emitting subsequent text or tool call events. When `REASONING_*` events are stripped from the stream, the agent works perfectly — even though Anthropic is still performing reasoning internally (the thinking just isn't surfaced to the frontend).

## Expected behavior

After reasoning content finishes streaming, a `REASONING_END` event should be emitted, followed by the agent continuing its run with `TEXT_MESSAGE_START`, `TOOL_CALL_START`, or `RUN_END` as appropriate. The presence of `REASONING_*` events in the stream should not cause the agent run to stall.

## Steps to reproduce

1. Configure a `BuiltInAgent` with an Anthropic model that supports thinking (e.g., Claude Sonnet 4.6).
2. Set `providerOptions.anthropic.thinking` to `{ type: 'adaptive' }` (or `{ type: 'enabled', budgetTokens: N }`).
3. Send a message that triggers agent tool use (not just a simple text reply).
4. Observe that `REASONING_START` and `REASONING_MESSAGE_CONTENT` events are emitted, but no `REASONING_END` ever arrives. The stream stalls — no further events of any kind.
5. Strip all `REASONING_*` events from the SSE stream and repeat — agent proceeds normally.

## Possibly related

- [CopilotKit PR #3165](https://github.com/CopilotKit/CopilotKit/pull/3165) — Added reasoning message support
- The issue may be in how the runtime converts Vercel AI SDK reasoning chunks into AG-UI events — specifically, `REASONING_END` never being emitted, which could block the state machine from transitioning to text/tool-call phases

### ✅ Expected Behavior

SSE returns REASONING_END which then resumes flow.

### What happens

1. Agent receives a user message and starts a run.
2. Anthropic returns thinking/reasoning content. CopilotKit's runtime translates these into AG-UI SSE events:
   - `REASONING_START`
   - `REASONING_MESSAGE_CONTENT` (multiple, streaming the thinking text)
3. After the reasoning content finishes streaming, **the stream stalls**. No `REASONING_END` is ever emitted. No `TEXT_MESSAGE_START`, no `TOOL_CALL_START`, no `RUN_END` — nothing. The stream hangs indefinitely.
4. The frontend shows the thinking content but the agent never proceeds to action.

### Screenshots

_No response_

### 𝌚 CopilotKit Version

```shell
## Environment

| Dependency | Version |
|---|---|
| `@copilotkit/runtime` | `1.52.1` |
| `@copilotkitnext/agent` | `1.52.1` |
| `@copilotkit/react-core` | `1.52.1` |
| `@ai-sdk/anthropic` | `2.0.67` |
| `ai` (Vercel AI SDK) | `5.0.116` |
| Node.js | v24 |

**LLM models used:** Anthropic Claude Sonnet 4.6, Claude Opus 4.6, Claude Opus 4.5 — all with `thinking` configured via the Vercel AI SDK's `providerOptions`.
```

### 📄 Logs (Optional)

```shell

```

giulio-leone:
## Summary

Fixes #3323 — Reasoning/Thinking events in SSE stream cause agent to stall when using Anthropic models.

## Problem

When Anthropic models produce reasoning/thinking tokens, the Vercel AI SDK (`@ai-sdk/anthropic`) streams `reasoning-start` and `reasoning-delta` events but **never emits `reasoning-end`**. CopilotKit's `BasicAgent` passively waits for this event in its `fullStream` loop, so the AG-UI state machine gets stuck in the "reasoning" phase:

1. `REASONING_START` is emitted ✅
2. `REASONING_MESSAGE_CONTENT` is emitted (streaming thinking text) ✅
3. `REASONING_END` is never emitted ❌ → stream stalls permanently

No `TEXT_MESSAGE_START`, `TOOL_CALL_START`, or `RUN_FINISHED` events follow, and the agent appears frozen.

## Root Cause

The event loop at `packages/v2/agent/src/index.ts` has separate `case` handlers for each stream event type, but there is no state tracking for whether reasoning is in progress, and no fallback to auto-close reasoning when the stream transitions to a different phase.

## Fix

- Added `isInReasoning` boolean flag to track the reasoning lifecycle
- Added `closeReasoningIfOpen()` helper that emits `REASONING_MESSAGE_END` + `REASONING_END` when reasoning is still open
- Called `closeReasoningIfOpen()` at every phase transition point:
  - `text-start` — reasoning → text output
  - `tool-input-start` — reasoning → tool call streaming
  - `tool-call` — reasoning → tool call (non-streaming)
  - `finish` — reasoning → run end
  - `error` — reasoning → error
  - `abort` — reasoning → abort

The fix is backward-compatible: when the SDK **does** emit `reasoning-end` normally (e.g., OpenAI models), the explicit handler sets `isInReasoning = false` first, making the guard a no-op at subsequent transitions.

## Tests Added

3 new test cases covering the exact scenarios from the bug report:

| Test | Scenario |
|------|----------|
| `should auto-close reasoning when SDK omits reasoning-end before tool call` | Anthropic thinking → tool call (no reasoning-end) |
| `should auto-close reasoning when SDK omits reasoning-end before text` | Anthropic thinking → text output (no reasoning-end) |
| `should auto-close reasoning when SDK omits reasoning-end before finish` | Reasoning-only stream → finish (no reasoning-end) |

All existing tests continue to pass (106/106).

## Validation

```
pnpm test        → 106 passed ✅
pnpm check-types → clean ✅
pnpm lint        → 0 errors (144 pre-existing warnings) ✅
```

### DIFF
--- a/.changeset/fix-reasoning-end-event.md
+++ b/.changeset/fix-reasoning-end-event.md
@@ -1,0 +1,7 @@
+---
+"@copilotkitnext/agent": patch
+---
+
+fix(agent): auto-close reasoning lifecycle when SDK omits reasoning-end
+
+Some AI SDK providers (notably @ai-sdk/anthropic) never emit the "reasoning-end" stream event. This fix adds defensive state tracking to auto-emit REASONING_MESSAGE_END + REASONING_END at phase transitions, preventing the agent from stalling.


giulio-leone:
Friendly ping — CI is green, tests pass, rebased on latest. Ready for review whenever convenient. Happy to address any feedback. 🙏

giulio-leone:
Friendly ping — rebased on latest and ready for review. Happy to address any feedback!

giulio-leone:
## ✅ Validation Evidence

**Branch:** `fix/reasoning-end-event` @ `4ba2e36`
**Status:** Rebased onto `origin/main` (2 commits applied cleanly)

**Double-pass test results** (strict identical runs, no code changes between passes):

```
Pass 1: 142 passed (9 test files)  ✅
Pass 2: 142 passed (9 test files)  ✅
```

Test runner: vitest · `packages/v2/agent/src/__tests__/`

Branch force-pushed to fork. CI is the authoritative gate.

PR #3407, opened 2026-03-09
chore: add skill for claude to describe our pre-commit hooks

ranst91:
This PR adds a simple, contextual, claude code skill that tells claude about this repo's pre commit hook and how to run them.
This is so you can tell it "I wish to commit a thing and the pre-commit hooks fail" and it'll have context quickly

### DIFF
--- a/.claude/skills/git-hooks/SKILL.md
+++ b/.claude/skills/git-hooks/SKILL.md
@@ -1,0 +1,70 @@
+---
+name: git-hooks
+description: CopilotKit pre-commit hook reference. Load automatically when the user mentions git hooks failing, pre-commit errors, lefthook issues, commit blocked, or "hooks don't work", or when user wants to commit/push anything Contains the full hook topology so debugging skips discovery and goes straight to diagnosis.
+---
+
+# CopilotKit Git Hooks Reference
+
+## Hook runner: Lefthook
+
+The repo uses **lefthook** (not husky). The git hook at `.git/hooks/pre-commit` calls lefthook, which reads `lefthook.yml` at the repo root.
+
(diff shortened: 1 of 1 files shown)


.claude/skills/git-hooks/SKILL.md
```diff
@@ -0,0 +1,78 @@
+---
+name: git-hooks
+description: CopilotKit pre-commit hook reference. Load automatically when the user mentions git hooks failing, pre-commit errors, lefthook issues, commit blocked, or "hooks don't work". Contains the full hook topology so debugging skips discovery and goes straight to diagnosis.
```
AlemTuzlak: I think this should also be loaded when user wants to commit/push something as well

.claude/skills/git-hooks/SKILL.md
```diff
@@ -0,0 +1,78 @@
...
+Only use this when the failure is a known flaky test unrelated to the change.
+
+## npm scripts involved
+
+| Script                    | Expands to                                           |
+| ------------------------- | ---------------------------------------------------- |
+| `pnpm run lint`           | `nx run-many -t lint --projects=packages/**`         |
+| `pnpm run format`         | `prettier --write "**/*.{ts,tsx,md}"`                |
+| `pnpm run test`           | `nx run-many -t test`                                |
+| `pnpm run check:packages` | `nx run-many -t publint,attw --projects=packages/**` |
```
AlemTuzlak: I wouldn't let it do this IMO

ranst91: True, missed that one

PR #3416, opened 2026-03-11
fix: render tool call registration lost under React Strict Mode

mxmzb:
## Summary

- **Root cause**: `CopilotKitProvider`'s `didMountRef` pattern breaks under React Strict Mode — refs persist across the simulated unmount/remount cycle, causing the provider's `setRenderToolCalls` effect to overwrite registrations from `useFrontendTool` and `useRenderTool` hooks
- **Fix**: Split render tool call storage in `CopilotKitCoreReact` into two independent layers (prop-based vs hook-based) so the provider and hooks never overwrite each other
- **Also includes**: Python SDK fix for `langchain_messages_to_copilotkit` always emitting assistant messages when AIMessage has tool_calls (independent correctness improvement)

## Details

The `didMountRef` guard in `CopilotKitProvider` skips setter effects on first mount to avoid overwriting child hook registrations. Under React Strict Mode (dev only), the mount→unmount→remount cycle leaves `didMountRef.current = true`, so on remount the provider's effect fires `setRenderToolCalls(allRenderToolCalls)` — which only contains prop-based entries — overwriting whatever `useFrontendTool`/`useRenderTool` just registered.

The fix introduces two new methods on `CopilotKitCoreReact`:
- `addHookRenderToolCall(entry)` — used by hooks
- `removeHookRenderToolCall(name, agentId)` — used by `useHumanInTheLoop` cleanup

The `renderToolCalls` getter merges both layers with a cached result for `useSyncExternalStore` referential stability. The `didMountRef` pattern continues to work as before for prop-based entries — this fix eliminates the race without changing the provider's effect ordering.

## Test plan

- [x] New test: `useFrontendTool` render registration survives StrictMode remount (was failing before fix)
- [x] New test: hook + prop render entries coexist through StrictMode remount
- [x] New test: hook render entries survive provider prop changes
- [x] All 799 existing tests pass (46 files)
- [x] Manual verification: "Get Help" button renders on initial load and after disconnect/reconnect

### DIFF
--- a/packages/v2/react/src/hooks/__tests__/standard-schema.test.tsx
+++ b/packages/v2/react/src/hooks/__tests__/standard-schema.test.tsx
@@ -22,6 +22,8 @@ vi.mock("../use-frontend-tool", () => ({
 type MockCore = {
   renderToolCalls: ReactToolCallRenderer[];
   setRenderToolCalls: ReturnType<typeof vi.fn>;
+  addHookRenderToolCall: ReturnType<typeof vi.fn>;
+  removeHookRenderToolCall: ReturnType<typeof vi.fn>;
 };
 
 const mockUseCopilotKit = useCopilotKit as ReturnType<typeof vi.fn>;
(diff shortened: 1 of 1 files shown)


packages/v2/react/src/hooks/use-frontend-tool.tsx
```diff
@@ -25,29 +25,12 @@ export function useFrontendTool<
...
         name,
         args: tool.parameters,
         agentId: tool.agentId,
         render: tool.render,
-      } as ReactToolCallRenderer;
-      mergedMap.set(keyOf(newEntry), newEntry);
-
-      // Set the merged list back
-      copilotkit.setRenderToolCalls(Array.from(mergedMap.values()));
+      } as ReactToolCallRenderer);
```
AlemTuzlak: can we try to remove this as

PR #3421, opened 2026-03-12
fix(react): yield to React scheduler before follow-up agent run to pr…

ranst91:


packages/v2/react/src/lib/react-core.ts
```diff
@@ -136,4 +136,25 @@ export class CopilotKitCoreReact extends CopilotKitCore {
...
+   * Awaiting a zero-delay timeout yields to the macrotask queue. In practice
+   * this gives React's scheduler tasks (re-render + useEffect) time to run
+   * before the follow-up runAgent call reads context. This is a best-effort
+   * heuristic: scheduler ordering between MessageChannel callbacks and
+   * setTimeout(0) is implementation-defined and not spec-guaranteed across all
+   * environments. It is reliable in the browser environments we target and in
+   * jsdom (Node.js), but edge cases (e.g. startTransition, concurrent mode
+   * suspense, React Native) may require additional handling in the future.
+   */
+  async waitForPendingFrameworkUpdates(): Promise<void> {
```
AlemTuzlak, line 153: I feel like this is what [flushSync](https://react.dev/reference/react-dom/flushSync) was built for

PR #3439, opened 2026-03-13
fix: handle empty tool arguments without crashing

### LINKED ISSUE #3263: 🐛 Bug: Facing "Unexpected end of JSON input" when human-in-the-loop action is called without parameters (Google ADK)
### ♻️ Reproduction Steps

Hi, first of all thanks for this great toolkit, really enjoy working with it. I'm trying to setup Google ADK with the CoPilot SDK following this documentation (https://docs.copilotkit.ai/adk/human-in-the-loop). 

While the HITL in this example here works with parameters, I had a use case which did not require input parameters. This however does not seem to work well; omitting the parameters sets the arguments to "" leaving to an "Unexpected en of JSON input" error here.  

### Steps to Reproduce

  **Backend — ADK agent (`main.py`)**

  Define a tool with an empty `properties` schema

  ```python
  ASK_FOR_COLOR_TOOL = {
      "type": "function",
      "function": {
          "name": "ask_for_circle_color",
          "description": "Asks the user to pick a color. Returns the color they
   picked.",
          "parameters": {
              "type": "object",
              "properties": {}   # ← no parameters
          }
      }
  }

  agent = Agent(
      model="gemini-2.5-flash",
      name="my_agent",
      instruction=f"""
          When the user asks to draw a circle, call ask_for_circle_color first,
          then call add_circle with the returned color.
          Tool reference: {ASK_FOR_COLOR_TOOL}
      """,
      tools=[]  
  )
  ```

  **Frontend — Next.js (`page.tsx`)**

  Register the same tool using `useHumanInTheLoop` with no `parameters` (which
  is documented as optional):

  ```tsx
  import { useHumanInTheLoop } from "@copilotkit/react-core";

  useHumanInTheLoop({
    name: "ask_for_circle_color",
    description: "Asks the user to pick a color. Returns the color they
  picked.",
    // parameters omitted — documented as optional when tool takes no args
    render: ({ respond, status }) => (
      <ColorPickerWidget respond={respond} status={status} />
    ),
  });

"use client";

import { useState } from "react";

const PRESET_COLORS = [
  "#ef4444", // red
  "#f97316", // orange
  "#eab308", // yellow
  "#22c55e", // green
  "#3b82f6", // blue
  "#8b5cf6", // violet
  "#ec4899", // pink
  "#14b8a6", // teal
];

interface ColorPickerWidgetProps {
  respond?: (color: string) => void;
  status: "inProgress" | "executing" | "complete";
}

export function ColorPickerWidget({ respond, status }: ColorPickerWidgetProps) {
  const [selected, setSelected] = useState("#3b82f6");
  const [custom, setCustom] = useState("#3b82f6");

  if (status === "inProgress") {
    return (
      <div className="rounded-xl border border-gray-200 bg-white p-4 shadow-sm w-72 animate-pulse">
        <div className="h-4 w-40 bg-gray-200 rounded mb-3" />
        <div className="flex gap-2 flex-wrap mb-3">
          {PRESET_COLORS.map((_, i) => (
            <div key={i} className="w-8 h-8 rounded-full bg-gray-200" />
          ))}
        </div>
        <div className="h-8 bg-gray-200 rounded" />
      </div>
    );
  }

  if (status === "complete") {
    return (
      <div className="flex items-center gap-2 text-sm text-gray-600 py-1">
        <span
          className="inline-block w-5 h-5 rounded-full border border-gray-300 flex-shrink-0"
          style={{ backgroundColor: selected }}
        />
        <span>Color chosen: <code className="font-mono text-xs">{selected}</code></span>
      </div>
    );
  }

  // status === "executing"
  return (
    <div className="rounded-xl border border-gray-200 bg-white p-4 shadow-sm w-72">
      <p className="text-sm font-medium text-gray-700 mb-3">Pick a color for the circle:</p>

      <div className="flex gap-2 flex-wrap mb-3">
        {PRESET_COLORS.map((color) => (
          <button
            key={color}
            onClick={() => setSelected(color)}
            className="w-8 h-8 rounded-full border-2 transition-transform hover:scale-110"
            style={{
              backgroundColor: color,
              borderColor: selected === color ? "#111" : "transparent",
            }}
            title={color}
          />
        ))}
      </div>

      <div className="flex items-center gap-2 mb-3">
        <label className="text-xs text-gray-500 shrink-0">Custom:</label>
        <input
          type="color"
          value={custom}
          onChange={(e) => {
            setCustom(e.target.value);
            setSelected(e.target.value);
          }}
          className="w-8 h-8 rounded cursor-pointer border border-gray-200"
        />
        <span className="text-xs font-mono text-gray-500">{selected}</span>
      </div>

      <button
        onClick={() => respond?.(selected)}
        className="w-full py-1.5 rounded-lg text-sm font-medium text-white transition-colors"
        style={{ backgroundColor: selected }}
      >
        Confirm
      </button>
    </div>
  );
}


  ```

In this case, the workaround I've made is adding a dummy argument to the call, but that imo should not be necessary, as this function call does not need parameters. 

### ✅ Expected Behavior

A tool with no parameters should be callable without error. An empty args object should convert to `{}` rather than '' and should parse cleanly and the `render` function should be invoked normally.

### ❌ Actual Behavior

  JSON parse error crashes the HITL render cycle. The color picker widget never
  appears and the agent gets stuck waiting for a response that never comes.

### Screenshots

Image

### 𝌚 CopilotKit Version

```shell
├── @copilotkit/react-core@1.51.4
├── @copilotkit/react-textarea@1.51.4
├── @copilotkit/react-ui@1.51.4
├── @copilotkit/runtime@1.51.4
```

### 📄 Logs (Optional)

```shell

```

giulio-leone:
## Problem

When a HITL (Human-in-the-Loop) tool has no parameters, some providers (e.g., Google ADK) send `arguments` as an empty string `""`. `JSON.parse("")` throws `"Unexpected end of JSON input"`, crashing the tool execution pipeline with an opaque error instead of executing the tool with empty args.

Fixes #3263

## Root Cause

In `run-handler.ts`, both the specific-tool and wildcard-tool execution paths call `JSON.parse(toolCall.function.arguments)` without checking for empty/falsy values:

```ts
// Line 309 — specific tool path
parsedArgs = JSON.parse(toolCall.function.arguments); // crashes on ""

// Line 448 — wildcard tool path  
parsedArgs = JSON.parse(toolCall.function.arguments); // crashes on ""
```

## Fix

Treat empty/falsy argument strings as `{}` (empty object) before parsing:

```ts
const rawArgs = toolCall.function.arguments;
parsedArgs = rawArgs ? JSON.parse(rawArgs) : {};
```

This matches the existing defensive patterns:
- **v1 `agui-to-gql.ts`**: already catches parse errors and falls back to `{}`
- **v2 `web-inspector`**: already checks `args === ""  ` before parsing (line 818)

## Changes

| File | Change |
|------|--------|
| `packages/v2/core/src/core/run-handler.ts` | Safe empty-args fallback at both parse sites |
| `packages/v2/core/src/__tests__/core-edge-cases.test.ts` | Updated test: empty args → handler called with `{}` |
| `packages/v2/core/src/__tests__/core-full.test.ts` | Updated test: empty args → handler called with `{}` |

## Testing

- All 230 passing tests still pass (61 pre-existing failures in unrelated packages unchanged)
- Updated 2 tests that previously expected a crash on empty args to verify the handler is called with `{}`

### DIFF
--- a/packages/v2/core/src/__tests__/core-edge-cases.test.ts
+++ b/packages/v2/core/src/__tests__/core-edge-cases.test.ts
@@ -187,7 +187,9 @@ describe("CopilotKitCore.runAgent - Edge Cases", () => {
   });
 
   it("should handle empty arguments via wildcard tool path", async () => {
-    const handler = vi.fn(async (args: any) => `Wildcard: ${JSON.stringify(args)}`);
+    const handler = vi.fn(
+      async (args: any) => `Wildcard: ${JSON.stringify(args)}`,
+    );
     const wildcardTool = createTool({
       name: "*",
       handler,


giulio-leone:
Thanks for the thorough review @jpr5! All feedback addressed:

**Issues fixed:**
1. ✅ Added test for wildcard tool path with empty args
2. ✅ Added tests for `null` and `undefined` arguments
3. ✅ Added inline comments explaining the fallback rationale
4. ✅ Replaced broad falsy check with explicit `=== '' || === null || === undefined`

**Suggestions applied:**
- Removed duplicate tests (TEST 7/8) from `core-full.test.ts`
- Removed all `console.log` debugging statements from tests
- Added inline comments in both code paths

Note: I kept the handler logic inline rather than extracting a helper — the two paths (`executeToolHandler` and `executeWildcardTool`) have different wrapping semantics so extraction would add indirection without reducing real duplication. Happy to revisit if you disagree.

Branch now starts clean from `main` with a single focused commit.

giulio-leone:
Thanks for the thorough review @jpr5! Pushed a follow-up commit addressing your feedback:

**Done:**
1. ✅ **Extracted `parseToolArguments()` helper** — both specific-tool and wildcard paths now share the same parsing logic (no duplication)
2. ✅ **Added debug-level logging** — `logger.debug()` fires when arguments fallback to `{}`, so it's no longer silent
3. ✅ **Typed `toolCall` parameter** — `executeSpecificTool` and `executeWildcardTool` now use `ToolCall` from `@ag-ui/client` instead of `any`
4. ✅ **Added JSDoc** on `parseToolArguments` explaining the rationale
5. ✅ **Removed console.log** from tests (done in previous commit)
6. ✅ **Removed duplicate TEST 7/8** from core-full.test.ts (done in previous commit)

**Already covered in edge-cases tests:**
- ✅ Wildcard tool path with empty args (`"should handle empty arguments via wildcard tool path"`)
- ✅ `null` arguments (`"should handle tool arguments as null"`)
- ✅ `undefined` arguments (`"should handle tool arguments as undefined"`)

**Re: falsy check breadth** — the check is intentionally `=== "" || === null || === undefined` (strict equality), not a truthiness check, so it won't catch `0`, `false`, or other falsy values. Only the three specific empty-argument cases that providers actually send are normalised.

Ready for re-review!

giulio-leone:
Hi @jpr5 — all requested changes have been addressed:
- ✅ Refactored to use shared `parseToolArguments()` utility
- ✅ Properly typed `toolCall` as `ToolCall`
- ✅ Both `executeToolHandler()` and `executeWildcardTool()` updated

Would you be able to re-review when you get a chance? Thanks!

PR #3443, opened 2026-03-13
fix(core): execute frontend tools when backend returns placeholder result

### LINKED ISSUE #3442: 🐛 Bug: useHumanIntheLoop not executing frontend tool after confirmation
### ♻️ Reproduction Steps

1. Create a Frontend tool using `useFrontendTool` 

```
useFrontendTool({
    name: "delete_orders",
    description: "Tool used to delete orders from the database",
    parameters: z.object({
      orderId: z.string().describe("The todo item text"),
    }),
    handler: async ({ orderId }) => {
      console.log("DELETING ORDER", orderId)
      return { status: "success", orderId }
    }
  })
```

2. Create a consent tool using `useHumanInTheLoop`

```
useHumanInTheLoop({
    name: "delete_approval",
    description: "Tool used to obtain approval from user by presenting a approval dialog",
    render: ({ args, status, respond, result }) => {
      console.log("DELETE APPROVAL TOOL CALLED", { args, status, result })

      if (status === ToolCallStatus.Executing && respond) {
        return (
          <>
            <div>Approve deletion?</div>
            <Button onPress={() => respond({ confirmed: false })}>Cancel</Button>
            <Button onPress={() => respond({ confirmed: true })}>Approve</Button>
          </>
        )
      }

      return null
    }
  })
```

3. Instruct Strands agent to always ask for consent before running any destructive tools like deletion using system prompt.

4. Prompt agent using `CopilotkitSidebar` - `Delete order 1`

### ✅ Expected Behavior

Agent should display approval dialog first because its a delete operation. One approved, the tool returns `{ confirmed: true }` to the agent and then the agent should proceed executing the `delete_orders` tool.

### ❌ Actual Behavior

Agent displays approval dialog. Upon approval agent says that it's executing the delete operation and has successfully deleted but it doesn't execute the delete tool. 

Model invocation logs shows that It tried to execute the tool and the execution was `forwarded to client` but seems like the `TOOL_CALL_START` ever was dropped by AG-UI? 

Model invocation logs - 
```
 [{
        "role": "assistant",
        "content": [{
                "text": "Now I'll proceed with deleting order 1."
            },
            {
                "toolUse": {
                    "toolUseId": "tooluse_Z12tEp5vpaRf1RZLa3xhsR",
                    "name": "delete_orders",
                    "input": {
                        "orderId": "1"
                    }
                }
            }
        ]
    },
    {
        "role": "user",
        "content": [{
            "toolResult": {
                "toolUseId": "tooluse_Z12tEp5vpaRf1RZLa3xhsR",
                "content": [{
                    "text": "Forwarded to client"
                }]
            }
        }]
    }
]
```

### Screenshots

https://github.com/user-attachments/assets/24e0fd16-48f5-43d5-9066-51db8ccd195c
Image
Image

### 𝌚 CopilotKit Version

```shell
`"@copilotkitnext/react": "^1.51.4"`
```

### 📄 Logs (Optional)

```shell

```

giulio-leone:
## Problem

When using `useHumanInTheLoop` with `useFrontendTool` and a remote backend agent (e.g., Strands), frontend tools are **never executed** after HITL approval.

**Flow:**
1. User asks to delete something
2. Agent calls `delete_approval` (HITL tool) → user sees approval dialog ✅
3. User clicks Approve → `respond({ confirmed: true })` is called ✅
4. Agent calls `delete_orders` (frontend tool) → tool handler **never fires** ❌

The agent's model invocation logs show the tool call was made and a result `"Forwarded to client"` was returned, but the frontend tool handler never executes.

## Root Cause

When a remote backend agent calls a frontend tool, the backend cannot execute it (frontend tools are registered without `execute` functions). The backend returns a placeholder result (e.g., `"Forwarded to client"`) via a `TOOL_CALL_RESULT` AG-UI event.

The `@ag-ui/client` library's `defaultApplyEvents` processes this event and creates a tool message in `messages`. Then in `processAgentResult` (run-handler.ts), the check at line 275-278:

```typescript
if (newMessages.findIndex(
  (m) => m.role === "tool" && m.toolCallId === toolCall.id,
) === -1) {
  // Execute frontend tool...
}
```

...sees the placeholder result already exists and **skips** the frontend tool execution entirely.

## Fix

Restructured `processAgentResult` to:

1. Look up the frontend tool **before** checking for existing results
2. If a registered frontend tool with a handler exists AND a result is already present, treat that result as a backend placeholder
3. Remove the placeholder from both `newMessages` and `agent.messages`
4. Execute the real frontend tool handler and insert the actual result

This is backward-compatible:
- For `BuiltInAgent` (local): frontend tools have no `execute` → Vercel AI SDK produces no `tool-result` part → no `TOOL_CALL_RESULT` event → no placeholder → original path works unchanged
- For remote agents: placeholder is now correctly replaced by the real frontend execution result

## Testing

The fix preserves all existing behavior:
- Tools without handlers: unchanged (wildcard fallback still works)
- Tools with `agentId` constraints: unchanged (check happens in `executeSpecificTool`)
- `followUp` logic: unchanged (triggers follow-up run with real results)
- Non-frontend tool results: unchanged (no handler → `shouldExecute` is false when result exists)

Closes #3442

### DIFF
--- a/.changeset/execute-frontend-tools-placeholder.md
+++ b/.changeset/execute-frontend-tools-placeholder.md
@@ -1,0 +1,6 @@
+---
+"@copilotkitnext/core": patch
+---
+
+fix(core): execute frontend tools when backend returns placeholder result — replace backend placeholder with real frontend execution
+


giulio-leone:
Closing to reduce PR volume. The fix is available in the branch if useful.

Vishal0203:
@giulio-leone  Why was this closed? 
Shouldn't this be the expected behavior? Please help me understand. Thanks!

giulio-leone:
@Vishal0203 good question — I closed it during a PR-volume cleanup, not because the behavior itself was considered wrong. The intended fix is still the same: when the backend returns a placeholder result for the HITL flow, the frontend tool should still execute instead of stopping at the placeholder.

The fix is still available in the branch. If this is still the expected behavior on your side, I am happy to rebase it onto current `main` and reopen/recreate the PR in the preferred form.

PR #3444, opened 2026-03-13
Fix Bedrock Converse API compatibility in CopilotKitMiddleware

mme:
## Summary

- Adds `_fix_messages_for_bedrock()` static method to `CopilotKitMiddleware` that sanitizes messages before model invocation
- Strips unanswered `tool_calls` (no matching `ToolMessage`) — Bedrock rejects `toolUse` blocks without a corresponding `toolResult`
- Syncs `msg.content` tool_use blocks with `msg.tool_calls` to remove stale content blocks left behind in checkpoints
- Fixes string-typed `args`/`input` in tool_calls and tool_use content blocks (Bedrock requires dict)

## Context

When using CopilotKit with Amazon Bedrock's Converse API, `after_agent` restores frontend tool_calls to the checkpoint. This creates three problems that cause Bedrock to reject the request:

1. AIMessages contain `toolUse` blocks without matching `toolResult` messages
2. Content blocks and `tool_calls` can get out of sync after `after_model` strips calls but leaves content blocks
3. Tool call arguments can be serialized as strings instead of dicts

This fix runs automatically in `awrap_model_call` before every model invocation.

## Test plan

- [ ] Verify existing CopilotKit tests still pass
- [ ] Test with Bedrock Converse API (Claude models) with frontend tool injection
- [ ] Test multi-turn conversations where frontend tool_calls are restored from checkpoints


ranst91:
# PR #3444 Review — Bedrock Converse API Compatibility

## Critical Issues

**1. Sync path (`wrap_model_call`) never gets the fix**
`_fix_messages_for_bedrock()` is only called in `awrap_model_call`. Synchronous Bedrock callers still get the same API rejections this PR is trying to fix.

**2. Return value discarded — relies on implicit mutation**
`copilotkit_lg_middleware.py:152`
```python
self._fix_messages_for_bedrock(request.messages)  # return value unused
```
If `request.messages` ever returns a copy, the fix silently becomes a no-op. Use the return value:
```python
fixed = self._fix_messages_for_bedrock(request.messages)
return await handler(request.override(messages=fixed))
```

**3. Silent data corruption with no logger**
`copilotkit_lg_middleware.py:127-130, 139-141` — Both `except` blocks replace invalid JSON with `{}` and swallow the error silently. The module doesn't import a logger (unlike `sdk.py`, `langgraph_agent.py`). Tools execute with empty args and the agent produces wrong results with zero log trail.

Fix: add `logger = get_logger(__name__)`, log a `WARNING` before each `{}` substitution including tool name, ID, and raw bad value.

**4. Unanswered tool_call stripping is silent + has a `KeyError` risk**
`copilotkit_lg_middleware.py:110-122` — Stripped tool_calls produce no log entry. Line 112 uses `tc['id']` (direct access) while the rest of the function uses `tc.get('id')` — a malformed checkpoint entry with no `'id'` key crashes `awrap_model_call` with a `KeyError`.

Fix: use `tc.get('id')` consistently; log a `WARNING` for every stripped tool_call.

**5. Zero test coverage for 86 lines of hot-path logic**
This method runs on **every** model invocation for every provider. A regression breaks OpenAI/Anthropic too. Minimum required tests:
- Mixed answered/unanswered tool_calls — only answered survives
- `ToolMessage(tool_call_id=None)` does not mark `None`-id calls as answered
- `tc['args']` as valid JSON string → dict; invalid → `{}`; already dict → unchanged
- `tc['args'] = None` → currently passes through to Bedrock unhandled (another rejection)
- Clean valid message list passes through unchanged (passthrough/regression guard)

---

## Important Issues

**6. `hasattr(m, 'tool_call_id')` always `True` — admits `None` into answered set**
`copilotkit_lg_middleware.py:78` — `tool_call_id` is a required Pydantic field, `hasattr` is always `True`. A `ToolMessage` deserialized via `model_construct()` can have `tool_call_id=None`, which then enters `answered_tc_ids` and incorrectly marks `None`-id tool_calls as answered.

Fix: `if isinstance(m, ToolMessage) and m.tool_call_id is not None`

**7. `tc['args'] = None` reaches Bedrock unhandled**
The `isinstance(tc.get('args'), str)` check is `False` for `None`. Bedrock rejects it.

PR #3445, opened 2026-03-13
fix: resolve stale agent closure in useCopilotChat reset/reload

### LINKED ISSUE #3430: useCopilotChat reloadMessages and reset not working
Migrated from archived demo repo during consolidation (PR #3418).

**Original issue:** https://github.com/CopilotKit/orca-CopilotKit-demo/issues/1

**Description:**
Core API bug in the `useCopilotChat` hook — `reloadMessages` and `reset` functions do not work as expected.

---
*This issue was automatically migrated from an archived demo repository.*

giulio-leone:
### Problem

`useCopilotChat` hook's `reset()` and `reloadMessages()` functions do not work as expected — messages are not cleared or reloaded.

### Root Cause Analysis

Two independent bugs in `use-copilot-chat_internal.ts`:

**1. Stale closure in `reload` callback**

The `reload` callback (created via `useAsyncCallback` / `useCallback`) captures `agent` in its closure, but the dependency array only lists derived properties:

```ts
[agent?.messages.length, agent?.isRunning, agent?.setMessages, copilotkit?.runAgent]
```

`agent?.setMessages` is a **prototype method** on `AbstractAgent`, so it has the **same function reference** across all instances. When the agent transitions from a provisional `ProxiedCopilotRuntimeAgent` to the real connected agent, these deps don't change (both have 0 messages, neither is running, same prototype method). The callback is **never recreated** and keeps operating on the stale provisional agent.

**2. Deferred ref update in `useUpdatedRef`**

`useUpdatedRef` updated `ref.current` inside a `useEffect` (fires post-render/paint). This created a window where `latestResetFunc` and `latestReloadFunc` would call stale closures that referenced the previous agent instance. The standard React pattern for "latest value refs" assigns synchronously during render.

### Fix

1. **Add `agent` to `reload`'s dependency array** — the callback is now recreated when the agent instance changes
2. **Update `useUpdatedRef` to assign synchronously** — eliminates the stale-ref window entirely

### Changes

- `packages/v1/react-core/src/hooks/use-copilot-chat_internal.ts`
  - Added `agent` to `reload` callback's dependency array (line 478)
  - Changed `useUpdatedRef` from `useEffect`-based to synchronous `ref.current = value` assignment

### Testing

- Verified `sendMessage` and `setMessages` already correctly include `agent` in their deps (consistent with this fix)
- The change is minimal and surgical — no behavioral change for correctly-functioning code paths

### DIFF
--- a/.changeset/stale-agent-reload.md
+++ b/.changeset/stale-agent-reload.md
@@ -1,0 +1,5 @@
+---
+"@copilotkit/react-core": patch
+---
+
+fix stale agent usage in `useCopilotChat` reset/reload flows
--- a/packages/v1/react-core/src/hooks/__tests__/use-copilot-chat-internal-connect.test.tsx
+++ b/packages/v1/react-core/src/hooks/__tests__/use-copilot-chat-internal-connect.test.tsx
@@ -17,6 +17,7 @@ import { CopilotKitCoreRuntimeConnectionStatus } from "@copilotkitnext/core";
 let mockRuntimeConnectionStatus: CopilotKitCoreRuntimeConnectionStatus =
   CopilotKitCoreRuntimeConnectionStatus.Disconnected;
 const mockConnectAgent = vi.fn().mockResolvedValue(undefined);
+const mockRunAgent = vi.fn().mockResolvedValue(undefined);
 
 const mockAgent: Record<string, unknown> = {
   messages: [],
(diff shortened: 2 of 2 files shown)


giulio-leone:
Added a follow-up push to reduce merge friction:

- added a patch changeset for `@copilotkit/react-core`
- added a regression test covering the agent-swap case behind the stale `reloadMessages()` closure

Local validation:

- `cd packages/v1/react-core && pnpm test -- src/hooks/__tests__/use-copilot-chat-internal-connect.test.tsx`
- `cd packages/v1/react-core && pnpm run publint && pnpm run attw`

The new regression test verifies that after the hook swaps from one agent instance to another with the same derived deps, `reloadMessages()` operates on the latest agent instance rather than the stale one.

giulio-leone:
Closing to reduce PR volume. The fix is available in the branch if useful.

PR #3458, opened 2026-03-17
fix(core): persist runId after HITL resolve

### LINKED ISSUE #3456: 🐛 Bug: runId changes after HITL resolve
### ♻️ Reproduction Steps

1. Define a Human-in-the-Loop frontend tool using useCopilotAction and renderAndWaitForResponse 
2. use the renderAndWaitForResponse parameter with any callback you prefer
3. Make the agent execute the Human-in-the-Loop frontend tool

### ✅ Expected Behavior

I would expect the runId to be the same in the backend since the fact that a frontend tool completed execution should differ from a runCompleted state.

### ❌ Actual Behavior

Instead of keeping the same runId when the human confirms the frontend tool execution, the runAgent function is recursively called inside RunHandler class (copilotkit_run_handler.ts) generating a different runId. This creates issues when tracking with external libraries like Langfuse and in general splits in half a run that is effectively not completed yet but simply halted due to user interaction.

### Screenshots

_No response_

### 𝌚 CopilotKit Version

```shell
1.52
```

### 📄 Logs (Optional)

```shell

```

BillionClaw:
## Summary

Fixes #3456

When a frontend tool using `renderAndWaitForResponse` (Human-in-the-Loop) completes and triggers a follow-up run, the `runId` was changing. This caused issues with external observability tools like Langfuse that track runs by their ID.

## Changes

- Add optional `runId` to `CopilotKitCoreRunAgentParams` interface
- Pass `runId` through `runAgent` to `agent.runAgent()` parameters
- Accept and forward `runId` in `processAgentResult` for follow-up runs

## Technical Details

The issue was in the `RunHandler.processAgentResult` method. When a HITL frontend tool completed execution and `needsFollowUp` was set to `true`, the code called `this.runAgent({ agent })` without passing the existing `runId`. This caused `AbstractAgent.runAgent` to generate a new `runId` via `prepareRunAgentInput`, breaking continuity for the same logical run.

The fix ensures that when `runAgent` is called for a follow-up after tool execution, the original `runId` is passed through to maintain continuity.

### DIFF
--- a/packages/v2/core/src/core/run-handler.ts
+++ b/packages/v2/core/src/core/run-handler.ts
@@ -15,6 +15,7 @@ import type { FrontendTool } from "../types";
 export interface CopilotKitCoreRunAgentParams {
   agent: AbstractAgent;
   forwardedProps?: Record<string, unknown>;
+  runId?: string;
 }
 
 export interface CopilotKitCoreConnectAgentParams {
@@ -212,6 +213,7 @@ export class RunHandler {
   async runAgent({
     agent,
     forwardedProps,
+    runId,
   }: CopilotKitCoreRunAgentParams): Promise<RunAgentResult> {
     // Agent ID is guaranteed to be set by validateAndAssignAgentId
     if (agent.agentId) {
@@ -244,10 +246,11 @@ export class RunHandler {
           },
           tools: this.buildFrontendTools(agent.agentId),
           context: Object.values(this._internal.context),
+          runId,
         },
         this.createAgentErrorSubscriber(agent),
       );
-      return this.processAgentResult({ runAgentResult, agent });
+      return this.processAgentResult({ runAgentResult, agent, runId });
     } catch (error) {
       const runError =
         error instanceof Error ? error : new Error(String(error));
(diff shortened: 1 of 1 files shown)


onthebed:
Friendly bump -- this PR has been open for a while. Happy to make any changes if needed, or close it if no longer relevant.

BenTaylorDev:
Thanks for this. Since it was opened, the repo migrated to a flat package layout and this area was substantially reworked, so the PR can't merge as authored and is largely superseded:

- **Path retired.** It edits `packages/v2/core/src/core/run-handler.ts`, but the `packages/v2/` tree no longer exists — the runtime core is now `packages/core/src/core/run-handler.ts` (and `git merge-tree` hard-conflicts against the rewritten file).
- **The plumbing already landed.** `runId?: string` is already on `CopilotKitCoreRunAgentParams` and forwarded into `agent.runAgent` on current `main` (`run-handler.ts:24`), so two of the three changes here are redundant.
- **The one residual behavior conflicts with the current design.** Reusing the prior `runId` on the post-HITL follow-up runs against how `main` now models this: follow-ups intentionally mint a new `runId` (grouped by the stable `threadId`), and HITL continuation is expressed through the newer `resume?: ResumeEntry[]` mechanism (`run-handler.ts:29`) rather than pinning one runId across two distinct executions.

Closing as superseded/retired-tree. If observability continuity across a HITL resolve is still a gap for you on latest `main`, it's worth a fresh issue framed around the current `runId`/`resume` semantics — happy to look. Thanks for surfacing #3456.

PR #3471, opened 2026-03-19
feat(runtime): Add a dedicated callback for identifying users in Intelligence (closes ENT-133)

MikeRyanDev:
## What does this PR do?

## Related PRs and Issues

## Checklist

### DIFF
--- a/packages/v2/runtime/src/handlers/intelligence/threads.ts
+++ b/packages/v2/runtime/src/handlers/intelligence/threads.ts
@@ -4,7 +4,7 @@ import {
   isIntelligenceRuntime,
 } from "../../runtime";
 import { logger } from "@copilotkitnext/shared";
-import { errorResponse } from "../shared/json-response";
+import { errorResponse, isHandlerResponse } from "../shared/json-response";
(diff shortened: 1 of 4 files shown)


packages/v2/runtime/src/handlers/intelligence/threads.ts
```diff
@@ -179,18 +207,16 @@ export async function handleDeleteThread({
...
-    const { userId, agentId } = body;
-
-    if (!isValidIdentifier(userId) || !isValidIdentifier(agentId)) {
-      return errorResponse("Valid userId and agentId are required", 400);
-    }
+    const mutation = await resolveThreadMutationContext(
+      intelligenceRuntime,
+      request,
+    );
+    if (mutation instanceof Response) return mutation;
```
BenTaylorDev: I can see why this is this way - it might be an ErrorResponse.  

But just to see this code in isolation, it _feels_ wrong.  Would it make sense to have this check for ErrorResponse on the instanceof?

MikeRyanDev: Clarified this guard

PR #3489, opened 2026-03-20
fix(runtime): isStreamConsumed false positive in async frameworks (Next.js pages router + bodyParser:false)

AlexNti:
## Problem

`isStreamConsumed()` in the node-http integration returns a false positive in async request pipelines — specifically **Next.js pages router with `bodyParser: false`**.

This caused the handler to silently skip the streaming path and forward an **empty body** to Hono, producing request failures for any user running CopilotKit behind the Next.js pages router without body parsing.

## Root Cause

`isStreamConsumed` previously checked four flags:

```ts
req.readableEnded ||
req.complete ||
readableState?.ended ||
readableState?.endEmitted
```

`req.complete` and `_readableState.ended` are set by the Node.js **HTTP parser** when all network bytes are received at the socket level — **not** when application code reads the stream. In frameworks that `await` before dispatching to the route handler (Next.js pages router performs async routing, component loading, etc.), the event loop ticks after the socket `data` callback completes. By that point, both flags are already `true` — even though the body is still sitting unread in `_readableState.buffer`.

The result: with `bodyParser: false`, `isStreamConsumed()` returned `true`, `req.body` was `undefined` (body parser never ran), and the code fell through to:

```
"Request stream consumed with no available body; sending empty payload."
```

...forwarding an empty body to Hono, finally the, `400 Invalid JSON payload` was sent to the client, making the agent completely  the agent was completely unreachable.

## Fix

Rely only on `req.readableEnded` (`=== _readableState.endEmitted`), which only becomes `true` after application code has fully drained the stream and the Node.js `end` event fires. This is the only reliable flag that distinguishes *"HTTP layer received bytes"* from *"application code consumed the stream"*.

## Verification

Reproduced and fixed with live Node.js HTTP experiments:

| Scenario | Before | After |
|---|---|---|
| `bodyParser:false` + async routing (Next.js) | `true` ❌ false positive → empty payload | `false` ✅ streaming path taken |
| `bodyParser:true` (middleware consumed stream) | `true` ✅ synthesis path | `true` ✅ synthesis path |
| `bodyParser:false` + sync handler (plain Node) | `false` ✅ | `false` ✅ |

## Files Changed

`packages/v1/runtime/src/lib/integrations/node-http/request-handler.ts`

### DIFF
--- a/.changeset/fix-stream-consumed-async-false-positive.md
+++ b/.changeset/fix-stream-consumed-async-false-positive.md
@@ -2,4 +2,4 @@
 "@copilotkit/runtime": patch
 ---
 
-fix(runtime): isStreamConsumed false positive in async frameworks — req.complete and _readableState.ended become true after the HTTP parser receives bytes, not after app code reads the stream. In async pipelines like Next.js pages router, this caused bodyParser:false requests to silently send an empty body to Hono. Now relies only on req.readableEnded, which correctly signals app-level stream consumption.
+fix(runtime): isStreamConsumed false positive in async frameworks — req.complete and \_readableState.ended become true after the HTTP parser receives bytes, not after app code reads the stream. In async pipelines like Next.js pages router, this caused bodyParser:false requests to silently send an empty body to Hono. Now relies only on req.readableEnded, which correctly signals app-level stream consumption.
--- a/.changeset/fix-stream-consumed-async-false-positive.md
+++ b/.changeset/fix-stream-consumed-async-false-positive.md
@@ -1,0 +1,5 @@
+---
+"@copilotkit/runtime": patch
+---
+
+fix(runtime): isStreamConsumed false positive in async frameworks — req.complete and _readableState.ended become true after the HTTP parser receives bytes, not after app code reads the stream. In async pipelines like Next.js pages router, this caused bodyParser:false requests to silently send an empty body to Hono. Now relies only on req.readableEnded, which correctly signals app-level stream consumption.
--- a/packages/v1/runtime/src/lib/integrations/node-http/__tests__/request-handler.test.ts
+++ b/packages/v1/runtime/src/lib/integrations/node-http/__tests__/request-handler.test.ts
@@ -20,29 +20,14 @@ describe("isStreamConsumed", () => {
   });
 
   it("returns false after an async tick with data in buffer (async framework scenario)", async () => {
-    // This is the regression test for Next.js pages router + bodyParser:false.
-    //
-    // The Node.js HTTP parser sets req.complete and _readableState.ended
-    // synchronously when all bytes arrive. In async frameworks (Next.js pages
-    // router, etc.) at least one event loop tick passes before the handler
-    // runs, so by then complete=true and ended=true — but the data is still
-    // unread in _readableState.buffer.
-    //
-    // The old implementation checked req.complete and _readableState.ended,
-    // which caused a false positive here, skipping the streaming path and
-    // sending an empty body to Hono.
     const stream = makeStream();
 
-    // Simulate HTTP parser pushing bytes and signaling end-of-message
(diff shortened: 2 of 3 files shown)


AlexNti:
Closing for now — will reopen once the fix is finalized.

AlexNti:
Hey team, could you please take a look at this one when you get a chance? It’s currently blocking our migration to the latest CopilotKit version. We have 6 agents running on the Pages Router, and all of them are failing because of this. For now, we’re bypassing the error with a custom patch that applies the exact same change as the one in this PR. Thanks!

@ranst91 
@ataibarkai 
@tylerslaton

BenTaylorDev:
Thanks for this — the diagnosis is correct and the bug is still live on `main`. `isStreamConsumed` really does over-report: in a Next.js pages-router setup with `bodyParser: false`, `req.complete` / `_readableState.endEmitted` can be true while the stream is still unread and `req.body` is `undefined`, so the request handler decides the body was already consumed and forwards an empty body upstream. Reducing the check to `req.readableEnded` (the accurate "the readable was actually drained to end" signal) is the right fix, and your test captures exactly the false-positive case.

The one problem is purely mechanical: this branch predates a large restructure of the repo, and the file you patched (`packages/v1/runtime/.../request-handler.ts`) no longer exists — the runtime now lives at `packages/runtime/src/lib/integrations/node-http/request-handler.ts` (with the same `isStreamConsumed` logic). So the PR can't be merged as-authored against the current tree.

Rather than ask you to rebase across the whole restructure, I'm going to close this and re-apply the same fix (plus your test) directly on `packages/runtime` so it lands promptly — full credit to you for finding and diagnosing it. Really appreciate the clear repro and the targeted test.

AlexNti:
> Thanks for this — the diagnosis is correct and the bug is still live on `main`. `isStreamConsumed` really does over-report: in a Next.js pages-router setup with `bodyParser: false`, `req.complete` / `_readableState.endEmitted` can be true while the stream is still unread and `req.body` is `undefined`, so the request handler decides the body was already consumed and forwards an empty body upstream. Reducing the check to `req.readableEnded` (the accurate "the readable was actually drained to end" signal) is the right fix, and your test captures exactly the false-positive case.
> 
> The one problem is purely mechanical: this branch predates a large restructure of the repo, and the file you patched (`packages/v1/runtime/.../request-handler.ts`) no longer exists — the runtime now lives at `packages/runtime/src/lib/integrations/node-http/request-handler.ts` (with the same `isStreamConsumed` logic). So the PR can't be merged as-authored against the current tree.
> 
> Rather than ask you to rebase across the whole restructure, I'm going to close this and re-apply the same fix (plus your test) directly on `packages/runtime` so it lands promptly — full credit to you for finding and diagnosing it. Really appreciate the clear repro and the targeted test.

@BenTaylorDev  Thanks for the update and for taking this forward. I’m glad the diagnosis, reproduction steps, and test were clear and useful. 
We are  looking forward to the fix landing so we can update our codebases and remove the custom patches we currently have in place.

BenTaylorDev:
Update on this: the fix is **merged to `main`** as **#6489** (`ba98b63`), applied to `packages/runtime/src/lib/integrations/node-http/request-handler.ts` with your test ported over, plus a live `http.IncomingMessage` case that stands up a real server and asserts the body is still readable after the parser has set `complete`. Your diagnosis is credited in the PR description. It isn't in a published release yet — I'll comment again once it ships so you know when to drop your patch.

Two notes for the interim. Your local patch is the same one-line change, so it will keep working until you pick up a release that includes the fix, and there's nothing to un-patch beyond deleting it. But if you carry it forward by path, note that the file moved in the restructure — `packages/runtime/...`, not `packages/v1/runtime/...` — so a patch pinned to the old location will silently stop applying rather than fail loudly.

BenTaylorDev:
v1.68.0 was released today.

PR #3498, opened 2026-03-22
fix(react): return null instead of throwing in useRenderCustomMessages when agent is absent

### LINKED ISSUE #3497: Bug: useRenderCustomMessages throws 'Agent not found' during runtime connection — no provisional agent fallback (v1.54.0)
## Description

After upgrading from `@copilotkit/react-core` and `@copilotkit/runtime` **1.53.0 to 1.54.0**, the application crashes with:

```
Error: Agent not found
    at use-render-custom-messages.tsx:43:13
```

The error is thrown from `useRenderCustomMessages` in `@copilotkitnext/react` when it calls `copilotkit.getAgent(agentId)` and the agent is not yet registered in the `AgentRegistry._agents` map.

## Root Cause

The `useAgent()` hook gracefully handles the connecting/error states by creating a **provisional** `ProxiedCopilotRuntimeAgent` (cached in a `useRef`). However, `useRenderCustomMessages()` has **no such fallback** — it directly calls `copilotkit.getAgent(agentId)` and hard-throws if it returns `undefined`:

```typescript
// use-render-custom-messages.tsx (line 42-43)
const agent = copilotkit.getAgent(agentId);
if (!agent) throw new Error("Agent not found");
```

This is triggered because:

1. The v1 `CopilotKit` wrapper from `@copilotkit/react-core` **always** injects `renderCustomMessages: [{ render: CoAgentStateRenderBridge }]` into the v2 `CopilotKitProvider`.
2. This causes `useRenderCustomMessages()` to return a non-null render function (the renderers list is non-empty since `CoAgentStateRenderBridge` has no `agentId` filter).
3. When any message is rendered during the connecting phase (or error state), the render function fires, calls `getAgent()`, gets `undefined`, and throws.

### Contributing Factor: `useSingleEndpoint` default changed to `true`

In 1.54.0, the v1 `CopilotKit` wrapper (`@copilotkit/react-core`) changed:

```javascript
// copilotkit-Yh_Ld_FX.mjs line 1554
useSingleEndpoint: props.useSingleEndpoint ?? true,  // was false in 1.53.0
```

This changes the runtime transport from `"rest"` (GET `/info`) to `"single"` (POST with `{ method: "info" }` envelope). If the single-endpoint `/info` POST fails for any reason (e.g., Express body-parsing consuming the stream before the Hono handler), the runtime enters `"error"` status, agents are never registered, and every subsequent message render throws.

## Steps to Reproduce

1. Use `@copilotkit/react-core` v1 API (`CopilotKit` component) with a self-hosted runtime
2. Upgrade both `@copilotkit/react-core` and `@copilotkit/runtime` from 1.53.0 to 1.54.0
3. Configure agents via `CopilotRuntime({ agents: { myAgent: new LangGraphHttpAgent({ url: ... }) } })`
4. Render `<CopilotChat agentId="myAgent" />` using the v2 API (`@copilotkit/react-core/v2`)
5. Load the page — the crash occurs immediately

## Expected Behavior

`useRenderCustomMessages` should handle the connecting/error state gracefully (return `null` instead of throwing), matching the pattern used by `useAgent()`.

## Actual Behavior

```
Error: Agent not found
    at use-render-custom-messages.tsx:43:13
```

The entire React tree crashes.

## Suggested Fix

In `use-render-custom-messages.tsx`, replace the hard throw with a graceful fallback:

```typescript
const agent = copilotkit.getAgent(agentId);
if (!agent) return null;  // Agent not yet registered — skip custom rendering
```

This matches the defensive pattern used elsewhere (e.g., `SuggestionEngine.reloadSuggestions` does `if (!agent) return;`).

## Workaround

Guard `CopilotChat` rendering until the agent is registered:

```tsx
import { useCopilotKit, CopilotKitCoreRuntimeConnectionStatus } from '@copilotkit/react-core/v2';

const { copilotkit } = useCopilotKit();
const agentReady = copilotkit.runtimeConnectionStatus === CopilotKitCoreRuntimeConnectionStatus.Connected
  && !!copilotkit.getAgent(myAgentId);

// Only render CopilotChat when agentReady is true
```

## Environment

- `@copilotkit/react-core`: 1.54.0
- `@copilotkit/runtime`: 1.54.0
- `@copilotkitnext/react`: 1.54.0 (transitive)
- `@copilotkitnext/core`: 1.54.0 (transitive)
- React: 18.x
- Node.js runtime: Express + `copilotRuntimeNodeExpressEndpoint`
- Backend: Python FastAPI + LangGraph + `copilotkit` SDK 0.1.78
- Previously working on: 1.53.0 (no code changes besides the version bump)

roman-rr:
## Summary

- Replace hard `throw new Error("Agent not found")` with `return null` in `useRenderCustomMessages` when `copilotkit.getAgent(agentId)` returns `undefined`
- Fixes #3497

## Problem

After upgrading from 1.53.0 to 1.54.0, the React tree crashes immediately with:

```
Error: Agent not found
    at use-render-custom-messages.tsx:43:13
```

The crash occurs because:

1. The v1 `CopilotKit` wrapper always injects `CoAgentStateRenderBridge` into `renderCustomMessages`, so the inner render function always executes for every message.
2. `useRenderCustomMessages` calls `copilotkit.getAgent(agentId)` and hard-throws if it returns `undefined`.
3. During runtime connecting/error states (or before the `/info` sync completes), agents are not yet in the registry, so `getAgent()` returns `undefined`.
4. The `useSingleEndpoint` default changed to `true` in 1.54.0, which can cause the `/info` fetch to fail, leaving the runtime in permanent error state with no agents registered.

## Fix

Replace the throw with `return null` — the same defensive pattern already used by:

- **`useRenderActivityMessage`** (line 54): passes `getAgent()` result through without throwing
- **`SuggestionEngine.reloadSuggestions`**: does `if (!agent) return;`
- **The hook's own early-return** (line 33-35): `if (!customMessageRenderers.length) return null;`

The component will re-render once the agent is available (the `CopilotKitProvider` forces re-renders on `onAgentsChanged` and `onRuntimeStatusChanged` events).

## Changed file

- `packages/v2/react/src/hooks/use-render-custom-messages.tsx` — 1 line changed (throw → return null) + comment explaining why

## Test plan

- [x] Verified the fix matches the existing defensive patterns in the codebase
- [ ] Upgrade from 1.53.0 to patched 1.54.x with a self-hosted Express runtime using `LangGraphHttpAgent` agents
- [ ] Confirm no crash on initial page load while the runtime is connecting
- [ ] Confirm custom message renderers (including `CoAgentStateRenderBridge`) work normally after the agent registers
- [ ] Confirm chat works end-to-end after the runtime connection establishes

### DIFF
--- a/packages/v2/react/src/hooks/use-render-custom-messages.tsx
+++ b/packages/v2/react/src/hooks/use-render-custom-messages.tsx
@@ -40,7 +40,13 @@ export function useRenderCustomMessages() {
     const runId = resolvedRunId ?? `missing-run-id:${message.id}`;
     const agent = copilotkit.getAgent(agentId);
     if (!agent) {
-      throw new Error("Agent not found");
+      // Agent may not be registered yet while the runtime is connecting,
+      // in an error state, or during React re-renders before the /info
+      // sync completes.  Return null to skip custom rendering for this
+      // cycle — the component will re-render once the agent is available.
+      // This matches the defensive pattern used by useRenderActivityMessage
+      // and SuggestionEngine.reloadSuggestions.
+      return null;
     }
 
     const messagesIdsInRun = resolvedRunId


BenTaylorDev:
Closing as superseded — merged #3797 already makes `useRenderCustomMessages` return `null` instead of throwing when the agent is absent (`packages/react-core/src/v2/hooks/use-render-custom-messages.tsx`). Issue #3497 is closed as completed. (Note the source moved from `packages/v2/react/` to `packages/react-core/src/v2/` in the flat-package refactor.) Thanks @roman-rr!

PR #3504, opened 2026-03-23
fix: expose containerRef in CopilotChatInput children render prop

### LINKED ISSUE #3500: CopilotChatInput containerRef not attached when using children render prop
## Description

When using the `children` render prop pattern on `CopilotChatInput`, the `containerRef` prop is never attached to a DOM element. This means `inputContainerHeight` in `CopilotChatView` stays at `0`, causing the dynamic `paddingBottom` on the scroll content to be too small — the absolutely-positioned input overlaps the last messages.

## Root Cause

In `CopilotChatInput.tsx`, the `containerRef` is only attached in the **default render path** (around line 497):

```tsx
// Default path — containerRef IS attached
return (
  <div ref={containerRef} className={cn(positioning === "absolute" && "cpk:absolute ...")} ...>
    ...
  </div>
);
```

But when `children` is provided, the component returns early (around line 278) **without attaching `containerRef`**:

```tsx
// Children render prop path — containerRef is NEVER attached
if (children) {
  const childProps = {
    textArea: BoundTextArea,
    sendButton: BoundSendButton,
    disclaimer: BoundDisclaimer,
    // ... no containerRef here
  };
  return (
    <div data-copilotkit style={{ display: "contents" }}>
      {children(childProps)}
    </div>
  );
}
```

Meanwhile, `CopilotChatView` measures the input container via `ResizeObserver`:

```tsx
const inputContainerRef = useRef(null);
// ...
resizeObserver.observe(element); // element is null → never observes
setInputContainerHeight(element.offsetHeight); // stays 0
```

And uses it for padding:

```tsx
<div style={{ paddingBottom: `${inputContainerHeight + FEATHER_HEIGHT + ...}px` }}>
```

With `inputContainerHeight = 0`, the padding is only `96 + 32 = 128px`, which is often not enough for a custom input layout.

## Reproduction

1. Use a custom input component with the children render prop:

```tsx
<CopilotChatInput {...props}>
  {({ textArea, sendButton }) => (
    <div className="cpk:absolute cpk:bottom-0 cpk:left-0 cpk:right-0">
      {textArea}
      {sendButton}
    </div>
  )}
</CopilotChatInput>
```

2. Send messages until the chat scrolls
3. The input overlaps the last messages because `paddingBottom` is too small

## Workaround

Manually extract `containerRef` from props and attach it to your custom wrapper:

```tsx
function MyInput({ containerRef: externalRef, ...props }) {
  const footerRef = useCallback((node) => {
    if (typeof externalRef === 'function') externalRef(node);
    else if (externalRef) externalRef.current = node;
  }, [externalRef]);

  return (
    <CopilotChatInput {...props}>
      {({ textArea, sendButton }) => (
        <div ref={footerRef} className="cpk:absolute cpk:bottom-0 ...">
          {textArea}
          {sendButton}
        </div>
      )}
    </CopilotChatInput>
  );
}
```

## Suggested Fix

Include `containerRef` in the `childProps` object so consumers can attach it:

```tsx
if (children) {
  const childProps = {
    textArea: BoundTextArea,
    sendButton: BoundSendButton,
    disclaimer: BoundDisclaimer,
    containerRef,  // ← add this
    // ...
  };
  return (
    <div data-copilotkit style={{ display: "contents" }}>
      {children(childProps)}
    </div>
  );
}
```

Or alternatively, attach `containerRef` to the `display: contents` wrapper div (since `display: contents` makes it layout-invisible, the ResizeObserver would need the actual positioned child — so exposing it in childProps is better).

## Environment

- `@copilotkit/react-core`: 1.54.0
- CopilotKit v2 API

marthakelly:
## Summary

- Adds `containerRef` to the `childProps` object in `CopilotChatInput`'s children render prop path
- Without this, `CopilotChatView`'s `ResizeObserver` never measures the input height, causing the input to overlap chat messages when using custom layouts

## Additional fix

- Fixes pre-existing prettier formatting issue in `docs/lib/integration-features.ts` (was already broken on main — the prettier CI check only runs on PRs, so it slipped through)

## Test plan

- [x] Added unit test verifying `containerRef` is passed to children render function
- [x] All 838 existing tests pass
- [x] Confirmed bug is not reproducible in default render path (demo app)

Closes #3500


tylerslaton (approved):
Looks great :shipit:

PR #3505, opened 2026-03-24
feat(useInterrupt): attach interrupt event metadata to resume command

mxmzb:
## Summary

- `useInterrupt`'s `resolve()` now includes `interruptEvent: pendingEvent.value` in `forwardedProps.command` alongside the user's `resume` response
- Uses a `useRef` for `pendingEvent` to avoid changing the `resolve` callback's dependency array (prevents handler effect double-fire)
- Updated test assertion to expect the new `interruptEvent` field

## Why

Agent adapters (like the Mastra AG-UI adapter in [ag-ui-protocol/ag-ui#feat/mastra-suspend-resume](https://github.com/ag-ui-protocol/ag-ui/compare/feat/mastra-suspend-resume)) need to round-trip interrupt context (`toolCallId`, `runId`, interrupt `type`) through the frontend without storing state between runs. This is required for Mastra's `tool-call-suspended` and `tool-call-approval` resume paths.

The LangGraph adapter is unaffected — it only reads `command.resume` and ignores the extra field.

## Test plan

- [x] All 15 existing `useInterrupt` tests pass (including updated resolve assertion)
- [x] Full CopilotKit test suite passes via pre-commit hook
- [x] Manual integration test with Mastra suspend/resume reproduction ticket (deep-agent `tkt-mastra-suspend`)

### DIFF
--- a/dev-docs/architecture/setup-intelligence.md
+++ b/dev-docs/architecture/setup-intelligence.md
@@ -70,7 +70,7 @@ import { CopilotKitIntelligence } from "@copilotkitnext/runtime";
 
 const intelligence = new CopilotKitIntelligence({
   apiKey: process.env.COPILOTKIT_INTELLIGENCE_API_KEY!,
-  tenantId: process.env.COPILOTKIT_INTELLIGENCE_TENANT_ID!,
+  organizationId: process.env.COPILOTKIT_INTELLIGENCE_ORGANIZATION_ID!,
   apiUrl: "https://your-intelligence-host/api",
   wsUrl: "wss://your-intelligence-host/socket",
 });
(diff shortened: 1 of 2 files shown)


packages/v2/react/src/hooks/use-interrupt.tsx
```diff
@@ -215,7 +223,12 @@ export function useInterrupt<
       setPendingEvent(null);
       copilotkit.runAgent({
         agent,
-        forwardedProps: { command: { resume: response } },
+        forwardedProps: {
+          command: {
+            resume: response,
+            interruptEvent: pendingEventRef.current?.value,
```
ranst91, line 229: Mastra doesn't keep a record of "there's a pending X right now"?
The reason we did not have an event being sent back so far is because with LG, there's only one interrupt ever happening at any given moment, which means a resume command that's being sent is acting on an already known "current interrupt", which is also fetchable through langgraph APIs at any given moment

mxmzb: Correct, that's why this is gonna provide the required data back to Mastra. Do you feel something should be done differently here @ranst91 ?

ranst91: Nope, if mastra does not have a record, let's provide it back

PR #3512, opened 2026-03-24
feat: rename tenantId to organizationId in intelligence 

maxkorp:
To match intelligence structure. THis should be removed entirely soon


maxkorp:
Goes with: https://github.com/CopilotKit/Intelligence/pull/52

Issue #3475, opened 2026-03-19
🐛 Bug: defaultOpen prop is ignored in UI components


alejandro-valle-uvi:
### ♻️ Reproduction Steps

1. Render CopilotSideBar with defaultOpen={false}
`import { CopilotSidebar } from "@copilotkit/react-core/v2";

export default function Page() {
  return <CopilotSidebar defaultOpen={false} />;
}`

2. Load the page
3. The sidebar is open
...

### ✅ Expected Behavior

The sidebar should start closed and only open when the user clicks the toggle button.

### ❌ Actual Behavior

Sidebar is open even when defaultOpen is "false".

Root cause: 

CopilotChat (line 194 of CopilotChat.tsx) wraps its children in CopilotChatConfigurationProvider without forwarding isModalDefaultOpen. The provider defaults to 

`useState(isModalDefaultOpen ?? true)` 
so isModalOpen starts as true. 

When CopilotSidebarView creates its own nested CopilotChatConfigurationProvider with isModalDefaultOpen: defaultOpen (correctly false), the inner provider resolves 

`resolvedIsModalOpen = parentConfig?.isModalOpen ?? internalModalOpen. `

The parent's true takes precedence, and the inner false is never used.

### 𝌚 CopilotKit Version

```shell
├── @copilotkit/react-core@1.54.0
├── @copilotkit/react-ui@1.54.0
├── @copilotkit/runtime@1.54.0
```

### 📄 Logs (Optional)

```shell

```

PR #3513, opened 2026-03-24
fix: respect defaultOpen={false} in CopilotSidebar and CopilotPopup

marthakelly:
## Summary

`<CopilotSidebar defaultOpen={false} />` was opening anyway because the prop was not being forwarded through the provider chain. This PR fixes that - and also addresses the architectural trade-off flagged in review.

### Who is affected

**User A** just uses `<CopilotSidebar />` with no outer provider
Before the fix, their sidebar opened by default. After the fix, it still opens by default. Nothing changed for them.                 
                                                            
**User B** uses `<CopilotSidebar defaultOpen={false} />`
Before: broken (opened anyway). After: works.
                                                                                                                                       
**User C** wraps their app in a bare `CopilotChatConfigurationProvider` and reads `isModalOpen` from it.
Before our fix `isModalOpen` was stuck at true forever. After: it reflects the sidebar's state.

### What changed

- The `defaultOpen` prop is forwarded correctly through the provider chain to both `CopilotSidebar` and `CopilotPopup`.
- Outer `useCopilotChatConfiguration()` hooks now reflect sidebar state. Fixed via bidirectional state sync between nested providers.

### Files changed

- `packages/v2/react/src/components/chat/CopilotChat.tsx` - accept and forward `isModalDefaultOpen`
- `packages/v2/react/src/components/chat/CopilotSidebar.tsx` - pass `isModalDefaultOpen={defaultOpen}`
- `packages/v2/react/src/components/chat/CopilotPopup.tsx` - same
- `packages/v2/react/src/providers/CopilotChatConfigurationProvider.tsx` - fix resolution logic + bidirectional sync
- `packages/v2/react/src/providers/__tests__/CopilotChatConfigurationProvider.test.tsx` - add tests

### Test plan

- [x] 843 tests pass (including 4 new bidirectional sync tests)
- [x] New tests cover all three reproduction scenarios from [tkt-modal-default-open](https://github.com/CopilotKit/deep-agent-cpk-experiments/tree/main/app/client/src/tickets/tkt-modal-default-open)

**Linear:** [CPK-7152](https://linear.app/copilotkit/issue/CPK-7152/fix-defaultopenfalse-prop-ignored-in-copilotsidebar)
**GitHub Issue:** #3475
**Related PRs:** #3477, #3496 (stalled community PRs addressing the same issue)

### DIFF
--- a/packages/v2/react/src/providers/CopilotChatConfigurationProvider.tsx
+++ b/packages/v2/react/src/providers/CopilotChatConfigurationProvider.tsx
@@ -1,8 +1,11 @@
 import React, {
   createContext,
+  useCallback,
   useContext,
   ReactNode,
+  useEffect,
   useMemo,
+  useRef,
   useState,
 } from "react";
 import { DEFAULT_AGENT_ID, randomUUID } from "@copilotkitnext/shared";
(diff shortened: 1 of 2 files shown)


packages/v2/react/src/providers/CopilotChatConfigurationProvider.tsx
```diff
@@ -88,9 +88,13 @@ export const CopilotChatConfigurationProvider: React.FC<
   const [internalModalOpen, setInternalModalOpen] =
     useState<boolean>(resolvedDefaultOpen);
 
-  const resolvedIsModalOpen = parentConfig?.isModalOpen ?? internalModalOpen;
-  const resolvedSetModalOpen =
-    parentConfig?.setModalOpen ?? setInternalModalOpen;
+  const hasExplicitDefault = isModalDefaultOpen !== undefined;
+  const resolvedIsModalOpen = hasExplicitDefault
```
mxmzb, line 126: I had worked on this before here https://github.com/CopilotKit/CopilotKit/pull/3340/changes and I don't remember right now what exactly was going wrong there and what the behaviors were, but basically:

- This is fixing behavior A
- It breaks behavior B

It's impossible to have A and B fixed at the same time, I think due to the two context providers or something (e.g. if we want to fix A and B we need to have a different solution).

marthakelly: Thanks for the flag. I looked into the history, #3313 reverted #3101 because it conditionally skipped creating modal state, making setModalOpen and isModalOpen undefined.

Is that the behavior B you encountered previously?

In this approach the modal state is always created, we just change which one gets read. Both values are always defined.

mxmzb: Yes, this PR 3513 breaks 3313. Also related is this: https://copilotkit.slack.com/archives/C09C1BLEPC1/p1769021190286849

I apologize for not being able to point out the issue exactly, but this PR will break again. I guess we have two groups of people composing components in two different ways or expecting two different things to happen, and we can only have one behavior working correctly (because there are two providers and one of them manages its state itself and then the child sidebar retrieves it from the closest provider, or something like that. Sorry again for being imprecise)

Maybe if you ask Claude to check my comment it may trace around and come up with more details. Also, you can take a look at https://github.com/CopilotKit/deep-agent-cpk-experiments/tree/main/app/client/src/tickets/tkt-modal-default-open and plug your fix into it and play around, I used that to reproduce issues and that was the reproduction example.

mxmzb: So, when I say it's "impossible to fix" of course I don't mean actually impossible, but the fix is more fundamental (probably cleaning up the two context providers).

marthakelly: Very helpful, thank you. I'll take a closer look in the morning. Appreciate the insight!

PR #3525, opened 2026-03-25
fix: multiple CopilotChat components with different threadIds share message state

### LINKED ISSUE #2957: 🐛 Bug: v2 - Multiple CopilotChat components with different threadId:s share same message history
### ♻️ Reproduction Steps

Issue: When rendering two CopilotChat components with the same agentId but different threadIds, sending a message to one chat updates both:

<CopilotChat threadId="123" />
<CopilotChat threadId="234" />

This happens when using the v2 components, and can reproduced in the [React demo app](https://github.com/CopilotKit/CopilotKit/tree/main/src/v2.x/apps/react/demo), for example by adding an additional CopilotChat component [here](https://github.com/CopilotKit/CopilotKit/blob/main/src/v2.x/apps/react/demo/src/app/single/page.tsx):

```javascript
 return <div><CopilotChat inputProps={{ toolsMenu }} threadId="xyz" />
         <CopilotChat inputProps={{ toolsMenu }} threadId="123" /></div> ;
```

The following workarounds have been suggested:

- **Use a unique agentId per chat instance (one per thread)** - My goal is to dynamically add CopilotChat components, and I need to be able to dynamically create new threads. Dynamic adding of agents is not possible, as I understand it (?).
- **Wrap each chat in its own CopilotKitProvider for isolated agent registries** - Same reason.

### ✅ Expected Behavior

 Each thread should maintain isolated message history.

### ❌ Actual Behavior

Image

### 𝌚 CopilotKit Version

```shell
Cloned the CopilotKit repo locally to repro the issue, so that is the code I am using. When running the command in my own project:

╬ô├╢┬ú╬ô├╢├ç╬ô├╢├ç @copilotkit/react-core@1.50.1 -> .\node_modules\.pnpm\@copilotkit+react-core@1.50_0d1db295f17c11d744b252d28582dfd1\node_modules\@copilotkit\react-core
╬ô├╢┬ú╬ô├╢├ç╬ô├╢├ç @copilotkit/react-ui@1.50.1 -> .\node_modules\.pnpm\@copilotkit+react-ui@1.50.1_90f70b2ba525fd17168ffe0e506d957c\node_modules\@copilotkit\react-ui
╬ô├╢┬ú╬ô├╢├ç╬ô├╢├ç @copilotkit/runtime-client-gql@1.50.1 extraneous -> .\node_modules\.pnpm\@copilotkit+runtime-client-_7c8018aaafcad5c230fb51e757b0ca5a\node_modules\@copilotkit\runtime-client-gql
╬ô├╢┬ú╬ô├╢├ç╬ô├╢├ç @copilotkit/runtime@1.50.1 -> .\node_modules\.pnpm\@copilotkit+runtime@1.50.1__6684207f0474cd65bd2ae85918beb446\node_modules\@copilotkit\runtime
```

### 📄 Logs (Optional)

```shell

```

marthakelly:
## Summary

Fixes #2957 (CPK-7155)

- **Root cause:** `useAgent` looked up agents by `agentId` only. Two `CopilotChat` instances with the same agent but different `threadId` values shared a single `AbstractAgent` instance — and thus shared `messages[]`, `state`, and `threadId`.
- **Fix:** `useAgent` now accepts an optional `threadId`. When provided, it clones the registry agent and caches per `agentId:threadId`, giving each chat its own isolated instance. Without `threadId`, behavior is unchanged (backward compatible).
- Updated V1 wrapper (`use-copilot-chat_internal`) to pass `threadId` through and removed the now-unnecessary threadId mutation guard.

## Test plan

- [x] New unit tests in `use-agent-thread-isolation.test.tsx` (6 tests):
  - Different threadIds → different agent instances
  - Same threadId → same cached instance (stable reference)
  - No threadId → shared registry agent (backward compat)
  - Message isolation between threads
  - ThreadId correctly set on clones
  - Composite key for provisional agents
- [x] Updated V1 connect guard tests to match new behavior
- [x] All existing tests pass
- [x] Manual: render two `<CopilotChat agentId="same" threadId="a" />` and `<CopilotChat agentId="same" threadId="b" />` side by side — messages should be isolated


ranst91:
## Critical Issues

### 1. Stale clone returned when registry agent is replaced

When `copilotkit.agents` changes (registry agent replaced — e.g. after reconnect, hot-reload, or config change), `useMemo` recomputes. But the cache-hit path ignores the new agent entirely:

```typescript
const cached = threadAgentCache.current.get(cacheKey);
if (cached) {
  if (cached instanceof HttpAgent) cached.headers = { ...copilotkit.headers };
  return cached;  // ← old clone from old agent, silently returned
}
```

If the new registry agent has a different URL, transport, credentials, or middleware chain, the clone will silently use the old configuration forever. Users see broken or stale responses with no error.

**Fix:** Store source reference alongside the clone and invalidate on mismatch:
```typescript
const threadAgentCache = useRef<Map<string, { source: AbstractAgent; clone: AbstractAgent }>>(new Map());

const entry = threadAgentCache.current.get(cacheKey);
if (entry && entry.source === existing) {
  if (entry.clone instanceof HttpAgent) entry.clone.headers = { ...copilotkit.headers };
  return entry.clone;
}
const clone = cloneForThread(existing, threadId, copilotkit.headers);
threadAgentCache.current.set(cacheKey, { source: existing, clone });
return clone;
```

---

### 2. V1 wrapper: removing threadId guard causes `connectAgent` on every Connected render

The old guard:
```typescript
if (agent && existingConfig?.threadId && agent.threadId !== existingConfig.threadId && Connected)
```
The new guard:
```typescript
if (agent && Connected)
```

The original guard served two purposes: preventing `agent.threadId` mutation (now handled by `useAgent`) AND as a dirty check to prevent redundant connections. Removing it means `connectAgent` fires on every render where status is `Connected` — including renders from unrelated state changes, StrictMode double-invocations, and `copilotkit` context re-renders.

If `connectAgent` sends a network request (handshake, session init), this is a behavioral regression. The updated tests explicitly confirm this fires in cases where it previously would not, including when `threadId` is missing entirely.

**Fix:** Track the last-connected agent instance instead:
```typescript
const lastConnectedAgentRef = useRef<AbstractAgent | null>(null);
if (agent && agent !== lastConnectedAgentRef.current && Connected) {
  lastConnectedAgentRef.current = agent;
  connect(agent);
}
```

---

### 3. No guard against `clone()` returning `this` — silent source state destruction

`cloneForThread` calls `clone.setMessages([])` and `clone.setState({})` without verifying `clone !== source`. If a custom `AbstractAgent` subclass implements `clone()` as `return this` (or returns a shallow copy sharing the internal `_messages` array reference), these calls silently destroy the source agent's state. Every other `CopilotChat` using the shared agent renders empty with no error or warning.

**Fix:**
```typescript
const clone = source.clone();
if (clone === source) {
  throw new Error(
    `useAgent: ${source.constructor.name}.clone() returned the same instance. ` +
    `clone() must return a new, independent object.`
  );
}
```

---

## Important Issues

### 4. Phantom provisional left in cache on threadId transition

1. Component mounts with no `threadId` → provisional created under key `"my-agent"`.
2. Component re-renders with `threadId="thread-a"` → `cacheKey` is now `"my-agent:thread-a"`.
3. Real agent appears → `provisionalAgentCache.delete("my-agent:thread-a")` runs, but the `"my-agent"` provisional is never removed.

If the component later remounts without a threadId, it receives the stale `"my-agent"` provisional (with `runtimeMode: "pending"`) instead of the now-available real agent. Chat silently non-functional.

**Fix:** Also delete the bare agentId key when clearing provisionals:
```typescript
provisionalAgentCache.current.delete(cacheKey);
provisionalAgentCache.current.delete(agentId);  // clear bare-key provisional too
```

---

### 5. `threadAgentCache` grows unboundedly — no eviction, no disposal

A `CopilotChat` that cycles through many threadIds (e.g. "new conversation" button without unmounting) accumulates one full `AbstractAgent` clone per unique threadId. If agents hold event listeners, WebSocket connections, or timers, none are disposed on threadId change or unmount.

**Fix:** At minimum, add a cleanup `useEffect` on unmount:
```typescript
useEffect(() => {
  return () => { threadAgentCache.current.clear(); };
}, []);
```

---

### 6. Mutation inside `useMemo` is unsafe in concurrent mode

```typescript
// Inside useMemo:
cached.headers = { ...copilotkit.headers };  // side effect in pure computation
```

React may invoke `useMemo` multiple times in concurrent mode and discard intermediate results. The mutation lands on the shared cache object regardless. Move the headers update to a `useEffect`, or apply it only inside `connect()` where it's already propagated.

---

## Test Coverage Gaps

**T-1 — Registry agent replacement: no test verifying stale clone is invalidated (Critical)**

A failing test: render with `registeredAgentV1`, capture clone; replace `mockCopilotkit.agents` with `registeredAgentV2` (different config); rerender; assert the returned agent reflects the new configuration. Currently this would fail.

**T-2 — `threadId` prop change on mounted component (Critical)**

No test verifies that switching `threadId` from `"a"` → `"b"` returns a fresh clone, and switching back to `"a"` returns the previously cached one.

**T-3 — Provisional → real agent transition for threadId path (Critical)**

The provisional → real agent lifecycle is only tested for the no-`threadId` path. No test verifies the threadId clone path gets a clean clone with correct `threadId` when the real agent appears.

**T-4 — V1 wrapper never asserts `useAgent` called with correct `threadId` (Critical)**

The V1 mocks `useAgent` at the module level and never asserts `expect(vi.mocked(useAgent)).toHaveBeenCalledWith(expect.objectContaining({ threadId: "config-thread-id" }))`. If the call site accidentally omits `threadId`, all V1 tests still pass.

**T-5 — Test 4 ("isolates messages") is vacuously true (Important)**

`CloneableAgent.clone()` creates a fresh empty agent, so agent-b having 0 messages is guaranteed by the constructor — not by `cloneForThread` clearing copied state. The isolation guarantee is never actually verified. `CloneableAgent.clone()` should copy `messages` and `state` from the source to make this test meaningful.

examples/v2/react/demo/src/app/api/copilotkit/[[...slug]]/route.ts
```diff
@@ -8,12 +8,17 @@ import { TranscriptionServiceOpenAI } from "@copilotkit/voice";
 import { handle } from "hono/vercel";
 import OpenAI from "openai";
 
+const isAnthropic =
+  !process.env.OPENAI_API_KEY?.trim() &&
+  !!process.env.ANTHROPIC_API_KEY?.trim();
+
```
Comment (AlemTuzlak) [inline examples/v2/react/demo/src/app/api/copilotkit/[[...slug]]/route.ts]: I'm pretty sure this would crash SSR in non-next envs because it's a module-scoped access to process, let's move it inside of determineModel

packages/core/src/core/state-manager.ts
```diff
@@ -36,45 +38,117 @@ export class StateManager {
...
+   * Subscribe to an agent's events to track state and messages.
+   *
+   * Registry agents (subscribed via `onAgentsChanged`) use the bare `agentId`
+   * key so that `unsubscribeFromAgent(agentId)` can remove them when they
+   * are replaced. Per-thread clones (subscribed via `subscribeAgentToStateManager`)
+   * pass `isClone = true` to use a composite `agentId:threadId` key, keeping
+   * their subscription independent of the registry agent's.
    */
-  subscribeToAgent(agent: AbstractAgent): void {
+  subscribeToAgent(agent: AbstractAgent, isClone = false): void {
```
AlemTuzlak: can we change this to be something like `SubscribeConfig = { isClone: boolean };`˛I hate magical booleans as params to functions, also JSDOC on top of the properties and what they do would be even better. Makes it easier to extend in the future and the code is self-documenting

packages/core/src/core/state-manager.ts
```diff
@@ -213,10 +297,21 @@ export class StateManager {
     message: Message,
     input?: RunAgentInput,
   ): void {
-    if (!agent.agentId || !input) return;
+    if (!agent.agentId) return;
 
-    const { threadId, runId } = input;
-    this.associateMessageWithRun(agent.agentId, threadId, message.id, runId);
+    if (input) {
```
AlemTuzlak: early return here if input not defined

packages/react-core/src/v2/__tests__/utils/test-helpers.tsx
```diff
@@ -51,9 +51,18 @@ export class MockStepwiseAgent extends AbstractAgent {
...
+    // Return a new instance that shares the same subject so tests can keep
+    // controlling events via the original reference while satisfying the
+    // clone() contract (must return a distinct object).
+    // Use the concrete constructor so subclasses (e.g. FailingConnectAgent)
+    // preserve their overridden methods.
+    const cloned = new (this
+      .constructor as new () => MockStepwiseAgent)() as this;
+    cloned.agentId = this.agentId;
+    (cloned as unknown as { subject: Subject<BaseEvent> }).subject =
+      this.subject;
```
AlemTuzlak, line 64: man this looks scary, we seriously need to fix our types, cc @tylerslaton

marthakelly: Agreed, tracked separately. This is a limitation of the current clone(): this polymorphic return type. Will follow up with Tyler on improving the upstream types https://linear.app/copilotkit/issue/CPK-7228/fix-abstractagentclone-return-type-to-eliminate-unsafe-casts-in

examples/v2/react/demo/src/app/api/copilotkit/[[...slug]]/route.ts
```diff
@@ -13,7 +13,8 @@ const determineModel = () => {
     return "openai/gpt-5.2";
   }
   if (process.env.ANTHROPIC_API_KEY?.trim()) {
-    return "anthropic/claude-sonnet-4.5";
+    // claude-3-7-sonnet supports extended thinking
+    return "anthropic/claude-3-7-sonnet-20250219";
```
Comment (tylerslaton) [inline examples/v2/react/demo/src/app/api/copilotkit/[[...slug]]/route.ts:17]: Any reason why this changed to a much earlier model?

PR #3526, opened 2026-03-25
fix(agent): skip empty reasoning deltas and auto-close reasoning lifecycle

### LINKED ISSUE #3323: 🐛 Bug: Reasoning / Thinking events in SSE stream cause agent to stall

marthakelly:
## Summary

Fixes two bugs in \`packages/v2/agent/src/index.ts\` that cause Anthropic model reasoning/thinking events to stall the agent pipeline.

Closes #3323

### Issue 1 — Empty delta kills the Observable (Critical)

The \`reasoning-delta\` case used \`?? \"\"\` which could emit \`delta: \"\"\`. \`@ag-ui/client\` calls \`EventSchemas.parse()\` on every SSE event and \`@ag-ui/core\` has \`delta: z.string().refine(s => s.length > 0)\` — an empty delta throws a ZodError that kills the RxJS Observable before any further events (tool calls, text, finish) can fire.

**Fix:** Skip emission when the delta is empty — \`if (!delta) break;\`

### Issue 2 — Missing \`reasoning-end\` stalls the state machine (Secondary)

\`@ai-sdk/anthropic\` never emits a \`reasoning-end\` stream part. Without it the downstream state machine stays stuck in the reasoning phase forever, preventing text output, tool calls, and run completion.

**Fix:** Track \`isInReasoning\` state and call \`closeReasoningIfOpen()\` at the start of every phase-transition case (\`abort\`, \`text-start\`, \`tool-input-start\`, \`tool-call\`, \`finish\`, \`error\`). This auto-emits the missing \`REASONING_MESSAGE_END\` + \`REASONING_END\` events so the stream can progress normally.

Additional hardening:
- \`closeReasoningIfOpen()\` is also called at the start of \`reasoning-start\` to correctly handle consecutive reasoning blocks (e.g. Anthropic extended thinking interleaved with tool calls) where no \`reasoning-end\` arrives between them
- \`closeReasoningIfOpen()\` is called in the \`!terminalEventEmitted\` fallback block so streams that end without any terminal event still close the reasoning lifecycle cleanly
- \`reasoning-start\` now generates a fresh \`randomUUID()\` when the provider supplies \`id="0"\` or no id, matching the \`text-start\` pattern and preventing consecutive blocks from sharing a \`messageId\`

## Changes

- \`packages/v2/agent/src/index.ts\` — all fixes
- \`packages/v2/agent/src/__tests__/basic-agent.test.ts\` — 6 new regression tests + updated existing "empty reasoning content" test + strengthened ordering assertions in auto-close tests
- \`packages/v2/agent/src/__tests__/test-helpers.ts\` — \`abort()\` stream event helper
- \`.changeset/fix-reasoning-stall.md\` — patch changeset for \`@copilotkitnext/agent\`

## Test plan

- [x] \`pnpm nx run @copilotkitnext/agent:test\` — all 153 tests pass (38 in \`basic-agent.test.ts\`, up from 32)
- [x] Full monorepo test suite passes locally (\`nx run-many -t test --projects="packages/**"\`)
- [x] Verify with Anthropic Claude 3.5/3.7 model that reasoning streams complete without stalling
- [x] Verify that streams with legitimate non-empty reasoning deltas are still emitted correctly


packages/runtime/src/agent/index.ts
```diff
@@ -991,6 +1010,7 @@ export class BuiltInAgent extends AbstractAgent {
           for await (const part of response.fullStream) {
             switch (part.type) {
               case "abort": {
+                closeReasoningIfOpen();
```
AlemTuzlak: from what I see we call this on every case? why not just execute it before the switch then?

packages/runtime/src/agent/index.ts
```diff
@@ -872,6 +872,28 @@ export class BuiltInAgent extends AbstractAgent {
...
+          isInReasoning = false;
+          subscriber.next({
+            type: EventType.REASONING_MESSAGE_END,
+            messageId: reasoningMessageId,
+          } as ReasoningMessageEndEvent);
+          subscriber.next({
+            type: EventType.REASONING_END,
+            messageId: reasoningMessageId,
+          } as ReasoningEndEvent);
+        };
```
AlemTuzlak: can we try to not have casts here?

packages/runtime/src/agent/__tests__/basic-agent.test.ts
```diff
@@ -1102,6 +1105,448 @@ describe("BasicAgent", () => {
...
+      vi.mocked(streamText).mockReturnValue(
+        mockStreamTextResponse([
+          reasoningStart(),
+          reasoningDelta("First thought"),
+          // NO reasoningEnd() — second block starts immediately
+          reasoningStart(),
+          reasoningDelta("Second thought"),
+          reasoningEnd(),
+          finish(),
+        ]) as any,
```
AlemTuzlak: I've been fighting a losing battle against llms and using as any, let's try to remove all of these casts in the test file, we have like 1k linting warnings atm about using casts and some other things. So whenever we see casts let's kill it with fire

packages/runtime/src/agent/index.ts
jpr5: @AlemTuzlak Fuel for `internal-skills`..

packages/runtime/src/agent/__tests__/basic-agent.test.ts
jpr5: @AlemTuzlak I have private CLAUDE.md logic to prevent these.  Incorporate into `internal-skills`...

packages/runtime/src/agent/index.ts
tylerslaton, line 1: Just want to take a step back - we're currently editing our `BuiltInAgent`, does the issue we're solving only surface here? Want to check if its a bug in `BuiltInAgent` or something more core to AG-UI.

AlemTuzlak:
From what I saw its anthropic related in ai sdk, so probably not on our side

Issue #3532, opened 2026-03-26
v2 runtime connect path bypasses agent.connect while run still preserves agent semantics


0xTomDaniel:
## Summary
The v2 runtime `connect` path bypasses agent-level connect semantics entirely.

`handleRunAgent` still preserves the agent abstraction: it clones/configures the registered agent and executes the run through the agent-backed path.

`handleConnectAgent`, by contrast, routes directly into runner-owned connect logic and never preserves agent-defined `connect` behavior. That creates an architectural asymmetry between `run` and `connect`, and it makes it impossible for agents with custom thread-follow / reconnect behavior to participate in the runtime contract in the same way they do for normal runs.

## Relevant commits
- First introduction of the bypass pattern: https://github.com/CopilotKit/CopilotKit/commit/7b838547a90663514c9d0c1628baded6c61be87a
- Later expansion/reference point for Intelligence support: https://github.com/CopilotKit/CopilotKit/commit/5b3037a2417b8b7edae1670bce874c9bf8384c15

## What looks wrong architecturally
Today, `run` and `connect` are not aligned:

- `run` preserves agent semantics.
- `connect` bypasses agent semantics and goes straight to runner-owned behavior.

That is surprising because AG-UI still models both sides of the contract:
- `AbstractAgent` has `runAgent(...)` / `run(...)`
- `AbstractAgent` also has `connectAgent(...)` / `connect(...)`

So from an extension/embedding perspective, `connect` still appears to be an agent capability, but the CopilotKit runtime no longer preserves that capability when servicing `/connect`.

## Why this matters
This is not just an implementation detail. It breaks a meaningful class of agents that need custom connect semantics at the agent layer.

A generic example is an agent that:
- owns its own notion of how to follow a thread over time
- can discover later runs on that thread
- can surface AG-UI state updates for runs not initiated by the current client request
- may need custom thread polling, run discovery, replay, or reconnection behavior that is specific to the agent/backend

When `connect` is forced through runner-owned logic only, that agent-level behavior is abandoned even though `run` still respects the agent abstraction.

## Expected behavior
The runtime should preserve agent-level `connect` semantics in the same way it preserves agent-level `run` semantics.

Concretely, the `connect` path should stay aligned with the `run` path as closely as possible:
- treat the registered agent as the primary semantic owner of connect behavior
- preserve agent configuration/state/thread initialization for connect just like run does
[...]

PR #3533, opened 2026-03-26
fix(runtime): preserve agent-owned connect semantics for sse

0xTomDaniel:
## Summary
- align the v2 SSE connect path with the run path by configuring and seeding the cloned agent before connect
- pass agent and parsed connect input through the SSE handler into the in-memory runner
- preserve agent-owned connect semantics in `InMemoryAgentRunner.connect` when runtime supplies an agent and connect input
- add regression coverage for handler configuration/seeding and agent-owned connect delegation

## Testing
- `pnpm nx run @copilotkitnext/runtime:test --output-style=stream`
- `pnpm nx run @copilotkit/runtime-client-gql:test --output-style=stream`

## Related
- Related to #3532
- Scope note: this PR fixes the SSE/in-memory runtime path. Intelligence-mode connect still follows its existing platform-managed path and likely needs a separate design decision if full symmetry is desired there.

### DIFF
--- a/.changeset/agent-connect-semantics.md
+++ b/.changeset/agent-connect-semantics.md
@@ -1,0 +1,5 @@
+---
+"@copilotkitnext/runtime": patch
+---
+
+fix: preserve agent-owned connect semantics for the sse runtime path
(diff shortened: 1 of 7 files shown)


BenTaylorDev:
Closing as un-mergeable against the current tree — targets the deleted `packages/v2/runtime/*` (flat is `packages/runtime/src/v2/runtime`), and the SSE connect path has since been reworked on main (per-request agent clone).

Issue #2731, opened 2025-11-18
🐛 Bug: Browser Build Fails Due to Accidental Inclusion of AWS SDK


zero8892084:
### ♻️ Reproduction Steps

Title: Browser Build Fails Due to Accidental Inclusion of AWS SDK via @langchain/core in @copilotkit/runtime@1.10.x

Description:

Hi team,

After upgrading to @copilotkit/runtime@1.10.6, our Vite-based React application fails at runtime in the browser with the following error:
Uncaught SyntaxError: Export 'import_getSSOTokenFromFile2' is not defined in module
This error originates from AWS SDK v3’s Node.js–specific module (@aws-sdk/shared-ini-file-loader), which is being bundled into the browser build.

Root Cause
We traced the dependency chain and found that:

@copilotkit/runtime@1.10.6
└─ @copilotkit/common@1.10.6
   └─ @langchain/core@^0.3.0
      └─ @aws-sdk/* (Node.js-only modules)
Although @langchain/core may be intended as an optional or backend-only dependency, Vite (and other bundlers) automatically include it during dependency pre-bundling, causing Node.js–specific code (e.g., file system operations for AWS SSO tokens) to be injected into the client-side bundle — which fails in the browser.

This breaks any frontend app using CopilotKit >=1.9.0, even when only using LangGraphAgent (which only needs HTTP/SSE and has no need for LangChain or AWS).

Impact
Frontend apps crash on load in the browser.
The issue persists even with a clean node_modules and cache because Vite re-prebundles the problematic dependencies.
Users cannot use the latest CopilotKit versions without workarounds.
Proposed Fixes
(Recommended) Mark @langchain/core as an optional peer dependency and exclude it from the main bundle.

If LangChain integration is needed, it should be opt-in via a separate package (e.g., @copilotkit/langchain-adapter).
Alternatively, ensure that any LangChain-related code is dynamically imported only in Node.js environments, or provide ESM builds with proper exports conditions that exclude Node.js modules from browser bundles.
At minimum, document this breaking change and provide guidance for frontend users to avoid the issue.
Temporary Workaround
Downgrading to @copilotkit/runtime@1.8.2 resolves the issue, as that version does not depend on @langchain/core.

npm install @copilotkit/runtime@1.8.2 @copilotkit/react-core@1.8.2 @copilotkit/react-ui@1.8.2

Thanks for your great work on CopilotKit! Let me know if you need more details or a minimal reproduction repo.

[...]

PR #3539, opened 2026-03-27
fix(runtime): keep browser bundles free of LangChain/AWS imports

raashish1601:
This adds a browser-safe runtime entry so Vite/browser builds no longer pull in LangChain-backed adapters and their AWS SDK transitive dependencies.

Validation:
- pnpm -C packages/v1/runtime build
- pnpm -C packages/v1/runtime exec vitest run src/browser.test.ts
- pnpm -C packages/v1/runtime publint .
- rg -n "@langchain/core|@langchain/aws|@langchain/community|langchain/runnables/remote|@aws-sdk" packages/v1/runtime/dist/browser.mjs packages/v1/runtime/dist/browser.cjs

### DIFF
--- a/.changeset/odd-bears-relax.md
+++ b/.changeset/odd-bears-relax.md
@@ -1,0 +1,5 @@
+---
+"@copilotkit/runtime": patch
+---
+
+Add a browser-safe runtime entry so browser bundles stop pulling LangChain-backed adapters and their AWS SDK transitive imports.
--- a/.changeset/odd-bears-relax.md
+++ b/.changeset/odd-bears-relax.md
@@ -1,5 +1,0 @@
----
-"@copilotkit/runtime": patch
----
-
-Add a browser-safe runtime entry so browser bundles stop pulling LangChain-backed adapters and their AWS SDK transitive imports.
--- a/.changeset/young-donkeys-bake.md
+++ b/.changeset/young-donkeys-bake.md
@@ -1,0 +1,5 @@
+---
+"@copilotkit/runtime": patch
+---
+
+fix runtime browser bundles to avoid pulling in LangChain and AWS browser-incompatible dependencies
--- a/packages/v1/runtime/package.json
+++ b/packages/v1/runtime/package.json
@@ -11,6 +11,8 @@
   },
   "version": "1.54.1-next.6",
   "sideEffects": [
+    "./dist/browser.mjs",
+    "./dist/browser.cjs",
     "./dist/index.mjs",
     "./dist/index.cjs",
     "./dist/v2/index.mjs",
@@ -20,8 +22,22 @@
   "module": "./dist/index.mjs",
   "exports": {
     ".": {
+      "browser": {
+        "import": {
+          "types": "./dist/browser.d.mts",
+          "default": "./dist/browser.mjs"
+        },
+        "require": {
+          "types": "./dist/browser.d.cts",
+          "default": "./dist/browser.cjs"
+        }
+      },
       "import": "./dist/index.mjs",
       "require": "./dist/index.cjs"
+    },
+    "./browser": {
+      "import": "./dist/browser.mjs",
+      "require": "./dist/browser.cjs"
     },
     "./langgraph": {
       "import": "./dist/langgraph.mjs",
--- a/packages/v1/runtime/src/browser-stubs.ts
+++ b/packages/v1/runtime/src/browser-stubs.ts
@@ -1,0 +1,26 @@
+const browserOnlyError = (name: string) =>
+  new Error(
+    `${name} is not available in browser bundles. Import it from a server-side runtime entry instead.`,
+  );
+
+export class LangChainAdapter {
+  constructor() {
+    throw browserOnlyError("LangChainAdapter");
+  }
+}
+
+export class GoogleGenerativeAIAdapter extends LangChainAdapter {}
+
+export class BedrockAdapter extends LangChainAdapter {}
+
+export class ExperimentalOllamaAdapter {
+  constructor() {
+    throw browserOnlyError("ExperimentalOllamaAdapter");
+  }
+}
+
+export class RemoteChain {
+  constructor() {
+    throw browserOnlyError("RemoteChain");
+  }
+}
--- a/packages/v1/runtime/src/browser.test.ts
+++ b/packages/v1/runtime/src/browser.test.ts
@@ -1,0 +1,26 @@
+import { describe, expect, it } from "vitest";
+import {
+  BedrockAdapter,
+  ExperimentalOllamaAdapter,
+  GoogleGenerativeAIAdapter,
+  LangChainAdapter,
+  RemoteChain,
+} from "./browser-stubs";
+
+describe("browser entry", () => {
+  it("keeps LangChain-backed adapters out of browser bundles", () => {
+    expect(() => new LangChainAdapter()).toThrow(
+      /not available in browser bundles/i,
+    );
+    expect(() => new GoogleGenerativeAIAdapter()).toThrow(
+      /not available in browser bundles/i,
+    );
+    expect(() => new BedrockAdapter()).toThrow(
+      /not available in browser bundles/i,
+    );
+    expect(() => new ExperimentalOllamaAdapter()).toThrow(
+      /not available in browser bundles/i,
+    );
+    expect(() => new RemoteChain()).toThrow(/not available in browser bundles/i);
+  });
+});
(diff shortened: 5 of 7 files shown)


marthakelly (commented):
# PR #3539 Review — CopilotKit/CopilotKit

**PR:** fix runtime browser bundles to avoid pulling in LangChain and AWS browser-incompatible dependencies **Branch:** `pr-3539` → `main`**Date:** 2026-04-01

---

## Critical Issues

**(1)** `browser.ts` **re-exports** `./langgraph`**, pulling in** `@langchain/langgraph-sdk`

`browser.ts:1` — `export * from "./langgraph"` re-exports from `./lib/runtime/agent-integrations/langgraph`, which imports `@langchain/langgraph-sdk`. This directly contradicts the PR's purpose. Furthermore, the main `index.ts` does NOT re-export `./langgraph` — it is a separate subpath export (`@copilotkit/runtime/langgraph`). The browser entry is exporting *more* than the main entry for this module.

**Fix:** Remove `export * from "./langgraph"` from `browser.ts`.

---

**(2)** `browser.ts` **re-exports** `./lib/runtime/copilot-runtime`**, which transitively imports** `node:crypto`

`browser.ts:2` — `copilot-runtime.ts` imports `TelemetryAgentRunner` from `./telemetry-agent-runner`, which imports `createHash` from `node:crypto` at the top level. It also imports `telemetry-client.ts` which uses `require("../../package.json")` and `node:crypto`. These are unconditional top-level imports that will crash in any browser environment.

**Fix:** Either lazy-import the telemetry dependencies in `copilot-runtime`, or provide a browser-safe wrapper that avoids the telemetry imports.

---

**(3)** `browser.ts` **re-exports** `./lib/integrations`**, which imports** `node:http`**,** `node:stream`**,** `graphql-yoga`**,** `type-graphql`

`browser.ts:4` — The integrations barrel exports `node-http`, `node-express`, `nest`, `nextjs/pages-router`, and `shared.ts`. `shared.ts` imports `graphql-yoga` and `type-graphql` at the top level. `node-http/index.ts` imports `node:http` and `node:stream`. All are server-only.

**Fix:** Exclude `./lib/integrations` from `browser.ts` entirely, or selectively re-export only browser-safe types.

---

**(4)** `browser.ts` **re-exports** `./lib/logger`**, which imports** `pino` **and** `pino-pretty`

`browser.ts:5` — `logger.ts` unconditionally imports `pino` and `pino-pretty`, both Node.js-specific libraries. These will fail to resolve in a browser bundler.

**Fix:** Remove `./lib/logger` from `browser.ts` or provide a browser-safe logger stub.

---

## Important Issues

**(5)** `exports["./browser"]` **is missing** `types` **conditions**

`package.json:38-41` — The explicit `./browser` subpath export lacks `types` fields:

```json
"./browser": {
  "import": "./dist/browser.mjs",
  "require": "./dist/browser.cjs"
}
```

TypeScript with `moduleResolution: "node16"` or `"bundler"` uses the `types` condition. Without it, `import {} from "@copilotkit/runtime/browser"` may fail to resolve types.

**Fix:** Add `types` sub-conditions matching the pattern used in the `browser` condition of `.`:

```json
"./browser": {
  "import": { "types": "./dist/browser.d.mts", "default": "./dist/browser.mjs" },
  "require": { "types": "./dist/browser.d.cts", "default": "./dist/browser.cjs" }
}
```

---

**(6) Missing deprecated exports from browser entry**

`browser.ts` — The main entry (via `lib/index.ts`) exports deprecated-but-browser-safe symbols that `browser.ts` omits:

- `LangGraphAgent`, `LangGraphHttpAgent` (deprecated classes that just throw)
- `TelemetryAgentRunner`, `TelemetryAgentRunnerConfig`
- Deprecated type aliases: `TextMessageEvents`, `ToolCallEvents`, `CustomEventNames`, `PredictStateTool`

A consumer switching from server to browser entry would get compile errors for these. The deprecated dummy classes are pure JS with no Node dependencies — they should be re-exported or stubbed.

---

**(7)** `GoogleGenerativeAIAdapter` **and** `BedrockAdapter` **stubs throw with the wrong class name**

`browser-stubs.ts:12-14` — These classes extend `LangChainAdapter` without their own constructor, so `new GoogleGenerativeAIAdapter()` throws *"LangChainAdapter is not available in browser bundles"* — naming the wrong class. This is a bug: the error message is misleading.

**Fix:** Give each subclass its own constructor:

```ts
export class GoogleGenerativeAIAdapter {
  constructor() { throw browserOnlyError("GoogleGenerativeAIAdapter"); }
}
export class BedrockAdapter {
  constructor() { throw browserOnlyError("BedrockAdapter"); }
}
```

Or use `new.target.name` in `LangChainAdapter`'s constructor to auto-detect.

---

**(8) Test imports from** `./browser-stubs`**, not** `./browser` **— the actual entry point is untested**

`browser.test.ts` — The test is titled "browser entry" but imports directly from `./browser-stubs`. The actual module that bundlers resolve (`browser.ts`) is never tested. If a future refactor changes `browser.ts` to stop re-exporting from `browser-stubs`, this test would still pass while the browser entry is broken.

**Fix:** Add at least one test that imports from `./browser` to verify stubs are reachable through the actual entry point.

---

**(9) No export parity test between** `browser.ts` **and** `index.ts`

The entire purpose of the browser entry is to be a drop-in replacement. If someone adds a new export to `index.ts` and forgets `browser.ts`, there is no safety net. A structural test comparing exported names between the two entries would prevent drift.

---

**(10) Constructor signature mismatches on stub classes**

The stub classes have no explicit constructors, so they advertise incompatible argument types vs. the real classes they replace. For example, the real `RemoteChain` requires a `RemoteChainParameters` argument, but the browser stub inherits a no-arg constructor from `LangChainAdapter`. Similarly, `ExperimentalOllamaAdapter` and others accept configuration objects the stubs silently ignore at the type level.

This is benign today because the `./browser` export is missing its `types` field (issue 5). Once that is fixed, any consumer passing constructor arguments — which is the normal usage pattern — will get compile errors against the browser entry's type declarations.

**Fix:** Give each stub class a constructor signature matching the real class, e.g.:

```ts
export class RemoteChain {
  constructor(_params?: any) { throw browserOnlyError("RemoteChain"); }
}
```

Or, for stricter type safety, import the parameter types (which are pure interfaces with no Node dependencies) and use them directly.

---

## Suggestions

**(11) Stub classes don't implement** `CopilotServiceAdapter` — The real adapter classes implement the `CopilotServiceAdapter` interface; the stubs don't. There is no runtime impact since constructors always throw before any method is called, but type-level assignments like `const adapter: CopilotServiceAdapter = new LangChainAdapter(...)` would fail against the browser entry's types. If the stubs are meant to be type-compatible drop-in replacements, consider adding `implements CopilotServiceAdapter` with method stubs that throw, or declaring the interface conformance via a type assertion in the export.

**(12) Error messages should be more actionable** — The current message *"Import it from a server-side runtime entry instead"* is vague. Consider specifying the import path: *"Use this adapter only in server-side code (e.g., API routes). If you are seeing this in a server context, ensure your bundler is not using the 'browser' export condition."*

**(13)** `browser.ts` **should have a module-level comment** explaining it is the browser-safe entry point, what it replaces, and why certain exports are stubbed. This is a non-obvious architectural boundary.

**(14)** `sideEffects` **marking on the browser entry may prevent tree-shaking** — The browser entry is marked as side-effectful (inherited pattern from the main entry which needs it for `reflect-metadata`). If the browser entry doesn't need `reflect-metadata`, removing it from `sideEffects` would enable better tree-shaking.

**(15) Changeset description is incomplete** — Mentions LangChain and AWS but the fix also stubs `ExperimentalOllamaAdapter` and `RemoteChain`. Consider: *"add browser-safe runtime entry that stubs out server-only adapters (LangChain, Bedrock, Ollama, RemoteChain)"*.

---

## Strengths

- The `browser` condition approach in `package.json` exports is the correct pattern for this problem — it lets bundlers transparently resolve the right entry.
- Stub classes that throw on construction is a clean fail-fast pattern with clear error messages.
- The 5 LangChain/AWS/Ollama stubs correctly cover the adapter classes that directly depend on those packages.
- The `tsdown.config.ts` change is minimal and correct — `dts: true` will auto-generate the `.d.mts`/`.d.cts` files referenced in the browser export condition.
- Co-located test file follows project conventions.

BenTaylorDev:
Closing as superseded — the "keep browser bundles free of LangChain" concern (#2731) was resolved on `main` via a different approach: `@langchain/*` are now optional peer dependencies (`packages/runtime/package.json` `peerDependenciesMeta`) and externalized in `packages/runtime/tsdown.config.ts` (with `unbundle: true`), so they no longer get pulled into browser bundles. Issue #2731 is closed as completed. This PR also targets the retired `packages/v1/runtime/` tree from before the flat-package refactor, so it no longer applies. Thanks @raashish1601!

PR #3550, opened 2026-03-29
feat(runtime): add framework-agnostic fetch-based runtime with adapters and docs

tylerslaton:
Implement a framework-agnostic CopilotKit runtime built on the Fetch API. The core handler (createCopilotRuntimeHandler) takes a Request and returns a Response, enabling native support on Bun, Deno, Cloudflare Workers, Next.js App Router, React Router, and TanStack Start. Thin adapters are provided for Node.js HTTP (createCopilotNodeHandler/createCopilotNodeListener), Express (createCopilotExpressHandler), and Hono (createCopilotHonoHandler).

Key changes:
- Add core fetch handler, router, CORS, and lifecycle hooks
- Add Node.js, Express, and Hono adapters with multi-route/single-route modes
- Add @copilotkit/runtime/v2 subpath exports (v2, v2/express, v2/hono, v2/node)
- Deprecate old function names (createCopilotEndpoint, createCopilotEndpointExpress, createNodeFetchHandler, createCopilotEndpointSingleRoute)
- Add Runtime Server Adapter docs page with examples for 10 frameworks
- Add example projects for Express, Hono, Node.js, Bun/Elysia, Deno, CF Workers
- Fix docs dark mode dropdown styling and card layout issues
- Adds CI integration tests for every runtime

> Original work by @AlemTuzlak — rebased onto current main.

## What does this PR do?

Rebased version of #3260 against current main, resolving conflicts from the `packages/v2/runtime/` → `packages/runtime/src/v2/runtime/` restructuring.

## Related PRs and Issues

- #3260 (original PR)

## Checklist

### DIFF
--- a/.oxlintrc.json
+++ b/.oxlintrc.json
@@ -10,9 +10,27 @@
     "react/react-in-jsx-scope": "off",
     "import/no-unassigned-import": "off",
     "unicorn/no-array-sort": "off",
-    "copilotkit/require-cpk-prefix": "off"
+    "copilotkit/require-cpk-prefix": "off",
+    "no-restricted-imports": "off"
   },
   "overrides": [
+    {
+      "files": ["packages/**/*.{ts,tsx}", "examples/**/*.{ts,tsx}"],
+      "rules": {
+        "no-restricted-imports": [
+          "error",
+          {
+            "paths": [
+              {
+                "name": "@a2ui/lit",
+                "message": "Do not use value imports from '@a2ui/lit' — it registers custom elements as a side effect, which breaks React Strict Mode. Use `import type` for types, or import from '@a2ui/lit/0.8' for value access (core only, no UI side effects).",
+                "allowTypeImports": true
+              }
+            ]
+          }
+        ]
+      }
+    },
     {
       "files": ["packages/react-ui/src/**/*.{ts,tsx}"],
       "rules": {
--- a/docs/content/docs/(root)/headless.mdx
+++ b/docs/content/docs/(root)/headless.mdx
@@ -36,6 +36,10 @@ Bring your own design system and render everything your way.
   </Tab>
 </Tabs>
 
+import PerformanceTip from "@/snippets/shared/performance-tip.mdx";
+
+<PerformanceTip components={props.components} />
+
 ## When should I use this?
 
 Use headless UI when you want to:
--- a/docs/content/docs/reference/v2/hooks/useAgent.mdx
+++ b/docs/content/docs/reference/v2/hooks/useAgent.mdx
@@ -263,6 +263,62 @@ function StreamingChat() {
 }
 ```
 
+## Performance
+
+### Two levels of throttle control
+
+CopilotKit provides throttle controls at two independent layers. They are **additive** — you can use one or both depending on your architecture:
+
+| Layer | Option | Where | What it throttles |
+|-------|--------|-------|-------------------|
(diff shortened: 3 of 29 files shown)


marthakelly:
## Highest-Priority Findings

1. **`RouteInfo` is not a discriminated union**: The eight non-null assertions in `fetch-handler.ts` (`route.agentId!`, `route.threadId!`) are a symptom of a type that permits illegal states. This is the single highest-leverage type fix in the PR and would clean up multiple files at once. Senior reviewers on a TypeScript codebase will notice this immediately.

2. **`addCorsHeaders` consumes `response.body`**: Reconstructing a `Response` from `response.body` after a hook may have read it is a latent crash. For SSE responses (the primary use case of this runtime), the body is a live `ReadableStream`. This deserves a look before merging.

3. **Hono adapter silently enables CORS when `cors` is omitted**: `toFetchCorsConfig(undefined)` returning `{}` instead of `undefined` causes CORS to be active-by-default for Hono users who didn't ask for it. The Express adapter handles this correctly; Hono should match.

4. **Hooks skipped on multi-route 404/405**: Users who register `onResponse` hooks for observability will silently miss 404 and 405 responses in multi-route mode. The inconsistency with single-route mode makes this hard to predict.

5. **Two parallel dispatch switch statements** : `dispatchSingleRoute` and `dispatchMultiRoute` have identical structure and will diverge as new routes are added. This is a maintenance bet against the future that a senior reviewer would flag for consolidation.

6. **`platform as any` in ~22 test locations**: Systematic casting in tests is casting debt that defeats the value of having TypeScript on the test suite. A shared `MockPlatform` type would make these tests worth something when the platform interface changes.

AlemTuzlak:
Addressed findings 1, 3, 4, 5, 6. Pushing back on finding 2.

**1. RouteInfo → discriminated union** — Done. Changed from interface with optional fields to:
```typescript
export type RouteInfo =
  | { method: "agent/run"; agentId: string }
  | { method: "agent/connect"; agentId: string }
  | { method: "agent/stop"; agentId: string; threadId: string }
  | { method: "info" }
  | { method: "transcribe" };
```
All 8 non-null assertions removed. TypeScript narrows correctly in switch cases now.

**2. `addCorsHeaders` consumes `response.body`** — Disagree. `new Response(response.body, { headers })` *transfers* the ReadableStream to the new Response per the Fetch spec — it does not read/consume it. The original response is discarded immediately (variable reassignment). For SSE streams specifically: the stream remains unread until the client pulls from it; wrapping it in a new Response with different headers is the standard pattern. If an `onResponse` hook reads the body and returns void, that's a hook authoring bug (documented: "Return a modified Response to replace the original"). No code change needed here.

**3. Hooks on multi-route 404/405** — Done. Multi-route 404 and 405 now `throw` the Response instead of returning early, so they flow through the same catch block as single-route (which already ran `onResponse` hooks for thrown Responses). This makes both modes consistent.

**4. Consolidated dispatch** — Done. Merged `dispatchMultiRoute` and `dispatchSingleRoute` into a single `dispatchRoute` function. Single-route mode now wraps the request body *before* calling dispatch, eliminating the parallel switch statements.

**5. Two parallel dispatch switch statements** — See finding 4 above.

**6. `platform as any` in tests** — Done. Defined a `MockIntelligencePlatform` interface in each test file. The `CopilotKitIntelligence` class uses private `#` fields which make `Partial<CopilotKitIntelligence>` unsatisfiable from object literals — that was the root cause of the `as any` casts. All 22 occurrences removed, unused imports cleaned up.

marthakelly (approved):
realllyyyy nice work

PR #3556, opened 2026-03-30
fix(threads): Improve useThreads DX with clearer names and direct typ…

BenTaylorDev:
…e hints

## What does this PR do?

This PR implements some feedback on the DX for useThreads, where type hints were indirect and property names (esp. for pagination) were likely to collide with other hooks.

## Related PRs and Issues

- ENT-183

## Checklist

### DIFF
--- a/packages/react-core/src/v2/hooks/use-threads.tsx
+++ b/packages/react-core/src/v2/hooks/use-threads.tsx
@@ -6,7 +6,6 @@ import {
   ɵselectThreadsIsLoading,
   ɵselectHasNextPage,
   ɵselectIsFetchingNextPage,
-  type ɵThread as CoreThread,
   type ɵThreadRuntimeContext,
   type ɵThreadStore,
 } from "@copilotkit/core";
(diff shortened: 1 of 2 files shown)


packages/react-core/src/v2/hooks/use-threads.tsx
```diff
@@ -68,18 +74,18 @@ export interface UseThreadsResult {
   error: Error | null;
   /**
    * `true` when there are more threads available to fetch via
-   * {@link fetchNextPage}. Only meaningful when `limit` is set.
+   * {@link fetchNextThreadsPage}. Only meaningful when `limit` is set.
    */
-  hasNextPage: boolean;
+  hasNextThreadsPage: boolean;
```
MikeRyanDev: Hmm, still not loving these. Sorry for the bikeshedding.

Thoughts on...
* `hasMoreThreads`
* `isFetchingMoreThreads`
* `fetchMoreThreads`

BenTaylorDev: I have no strong preference.  The current set is technically accurate - we use pages.

Your suggestion is more conversational.  I suppose the fact that it is pages is already known via configuration the hook.

BenTaylorDev: I can swap them out in a bit.

maxkorp:
Formatting error is from a different PR. I'll fix that once I get a few merges in

mxmzb (approved):

PR #3562, opened 2026-03-31
feat: add debug mode to runtime and client

AlemTuzlak:
## Summary
- Add `debug` option to `CopilotRuntime` constructor and `<CopilotKit>` provider for detailed event pipeline logging
- Server-side uses Pino structured logger; client-side uses `console.debug` at each AG-UI pipeline stage
- Accepts `true` for full output or a granular config: `{ events, lifecycle, verbose }`
- The two toggles are independent — enabling debug on the client does not affect the server

## Changes

**`@copilotkit/shared`**
- New `DebugConfig` type, `ResolvedDebugConfig` interface, `resolveDebugConfig()` normalizer

**`@copilotkit/runtime`**
- `debug` option on `CopilotRuntime({ debug: true })`
- Lifecycle logging: agent run started, SSE stream opened/completed/errored
- Event logging: every emitted event (verbose = full payload, otherwise key fields summary)

**`@copilotkit/react-core` + `@copilotkit/core`**
- `debug` prop on `<CopilotKit debug={true}>`
- Threads config through `CopilotKitCore` → `AgentRegistry` → `ProxiedCopilotRuntimeAgent`

**Companion PR:** ag-ui-protocol/ag-ui#1398 (AG-UI client-side debug logging)

## Test plan
- [x] All 647 runtime tests pass
- [x] All 352 AG-UI client tests pass
- [x] Lint clean (0 errors)
- [x] All affected packages build successfully
- [ ] Manual: set `debug: true` on runtime and provider, verify structured logs on server and console output on client


packages/runtime/src/v2/runtime/handlers/shared/sse-response.ts
```diff
@@ -50,10 +58,27 @@ export function createSseEventResponse({
...
+    }
+
+    let eventCount = 0;
+
     subscription = observable.subscribe({
       next: async (event) => {
         if (!request.signal.aborted && !streamClosed) {
           try {
+            if (debug?.events) {
+              eventCount++;
```
marthakelly: `eventCount` only increments when debug.events is true, so if someone has `lifecycle: true, events: false`, the completion log will always show `eventCount: 0` even if the agent emitted a bunch of events. Could be confusing if you're just trying to confirm your agent ran.

Two ideas: rename it to `loggedEventCount` so the scope is obvious, or count all emitted events unconditionally and just keep the per-event detail logging gated on `debug.events.` Either way works, just want to make sure folks don't misread it as "nothing happened."

packages/runtime/src/v2/runtime/handlers/handle-run.ts
```diff
@@ -65,7 +77,13 @@ export async function handleRunAgent({
       });
     }
 
-    return handleSseRun({ runtime, request, agent, input });
+    return handleSseRun({
+      runtime,
+      request,
+      agent,
+      input,
+      debug: runtime.debug ?? undefined,
```
marthakelly: `runtime.debug` is typed as `ResolvedDebugConfig` (non-optional on CopilotRuntimeLike), so `?? undefined` can never trigger. Looks like a leftover from when debug was optional, safe to drop.

packages/runtime/src/v2/runtime/handlers/shared/sse-response.ts
```diff
@@ -98,3 +132,26 @@ export function createSseEventResponse({
     },
   });
 }
+
+function summarizeEvent(event: BaseEvent): Record<string, unknown> {
+  const e = event as any;
```
marthakelly, line 149: Since `BaseEvent` is already a discriminated union on type, a type guard or switch on `event.type` would give you full type safety there for free. Not urgent, but given how careful the rest of the codebase is about casting, might be worth a quick pass.

packages/runtime/src/v2/runtime/__tests__/debug-sse-response.test.ts
```diff
@@ -0,0 +1,225 @@
...
+      observableFactory: () => createTestObservable([event]),
+      debug,
+    });
+
+    await drainResponse(response);
+
+    expect(mockDebug).toHaveBeenCalledWith("SSE stream opened");
+  });
+
+  it("logs lifecycle message on stream complete with event count", async () => {
```
marthakelly: This test sends 2 events but asserts `eventCount: 0`, which is correct since `events: false` means the counter never increments, but it means the test passes without verifying the counter works.

Might be worth either flipping to `events: true` so `eventCount` reflects the 2 sent events, or just dropping the count from the assertion and checking that "SSE stream completed" was logged. Either way you'd be testing something meaningful instead of a zero that can't fail.

marthakelly (dismissed):

packages/runtime/src/v2/runtime/handlers/shared/sse-response.ts
AlemTuzlak: Fixed. `eventCount` now increments unconditionally for all emitted events, and a separate `loggedEventCount` tracks only events logged when `debug.events` is true. The completion log reports both: `{ eventCount: 2, loggedEventCount: 0 }` so the total is always visible.

packages/runtime/src/v2/runtime/handlers/handle-run.ts
AlemTuzlak: Fixed. `runtime.debug` is non-optional `ResolvedDebugConfig`, so `?? undefined` was dead code. Removed.

packages/runtime/src/v2/runtime/handlers/shared/sse-response.ts
AlemTuzlak: Leaving as-is for now. `BaseEvent` is a discriminated union with 20+ event types — a switch on `event.type` would require exhaustively handling each variant just to pull optional summary fields. The `as any` here is scoped to a logging-only function that reads optional fields with truthiness guards, so the practical risk is low. A type guard approach would add significant boilerplate for the same result.

packages/runtime/src/v2/runtime/__tests__/debug-sse-response.test.ts
AlemTuzlak: Fixed. The existing test now verifies `eventCount: 2, loggedEventCount: 0` (events counted but not logged). Added a new test with `events: true` that verifies `eventCount: 2, loggedEventCount: 2` — confirming the counter works when event logging is enabled.

jpr5 (commented):
## Code Review: feat: add debug mode to runtime and client

Nice work on this, @AlemTuzlak — the type design (`boolean | granular object`) is clean and well-precedented, the `resolveDebugConfig()` normalizer is thorough, and the server-side logging (event summaries, lifecycle bookends, counters) is solid. marthakelly's feedback was properly addressed. Hot path is cheap when debug is disabled.

That said, the review found a few issues that should be addressed before merge.

---

## High

### 1. `debug` not stored on `ProxiedCopilotRuntimeAgent`

`ProxiedCopilotRuntimeAgentConfig` (`agent.ts:74-84`) has no `debug` field. `AgentRegistry.updateRuntimeConnection()` passes `debug: rawDebug` to the constructor (`agent-registry.ts:292`) — silently dropped. `clone()` reads `this.debug` (`agent.ts:374`) — always `undefined`.

The threading chain is broken at `AgentRegistry` → `ProxiedCopilotRuntimeAgent`. Debug config never reaches agents on the client.

### 2. Per-request Pino logger allocation

Both `createSseEventResponse` (`sse-response.ts:27`) and `handleRunAgent` (`handle-run.ts:61`) call `createLogger()` on every request when debug is enabled. `createLogger` creates a new pino instance with a `pino-pretty` stream each time. pino-pretty is synchronous. Under load with debug enabled, this is wasteful.

**Suggestion:** Cache the logger on the runtime instance or use a module-level singleton.

### 3. Client-side `debug` prop is plumbing-only — no actual logging

Grep for `console.debug` across the client packages returns zero hits. The `<CopilotKit debug={true}>` prop stores the config and passes it to ag-ui's `transformChunks`, but CopilotKit's own client has no logging infrastructure consuming it.

The JSDoc says "Enable debug logging for the client-side event pipeline" which is misleading if no client-side logging actually happens. Either add `console.debug` calls at each pipeline stage, or update the JSDoc to clarify this only controls ag-ui pipeline logging.

### 4. Missing changeset

New feature touching `@copilotkit/shared`, `@copilotkit/runtime`, `@copilotkit/react-core`, `@copilotkit/core`. All need minor version bumps. No `.changeset/` file exists. The changeset bot flagged this too.

---

## Medium

### 5. `verbose: true` by default when `debug: true` — PII risk

When `debug: true` (the boolean shorthand), `verbose` defaults to `true`, which logs entire `BaseEvent` objects via `debugLogger.debug({ event })` (`sse-response.ts:76`). Events carry user messages, tool call arguments, and state snapshots. Any log aggregator would ingest this.

Users who casually pass `debug: true` get full payload logging with no warning. Consider defaulting `verbose` to `false` even for the boolean case, or adding a `console.warn` noting that verbose mode logs full payloads.

### 6. `debug` not synced on prop changes

`CopilotKitProvider` passes `debug` only at `CopilotKitCore` construction (line 563). There's no `setDebug()` method, and the prop-sync effect (lines 662-683) doesn't include `debug`. Toggling debug at runtime is silently ignored. If this is intentional, a comment would help.

### 7. `verbose` naming is ambiguous

`verbose` reads like "more logging" but actually means "full payloads vs key-field summaries." A name like `fullPayloads` or `expandPayloads` would be more self-describing for users who expect `--verbose` CLI semantics.

### 8. ag-ui companion PR dependency

`transformChunks(this.debug)` passes debug config to ag-ui. If ag-ui#1398 isn't merged first, the current ag-ui release doesn't accept that argument. The merge order dependency should be documented in the PR description.

### 9. `debug-config-threading.test.ts` doesn't test threading

The test only verifies `CopilotKitCore` stores the `debug` property. It never instantiates `AgentRegistry` or `ProxiedCopilotRuntimeAgent` — which is where the broken chain lives (finding #1). Consider extending the test to cover the full chain, or renaming it to reflect what it actually tests.

---

## Low

- **No `NODE_ENV` production warning** — `debug: true` on a deployed runtime silently enables logging with no heads-up. A `console.warn` on `NODE_ENV === 'production'` would be cheap insurance.
- **`!debug` guard uses falsy coercion** (`debug.ts:42`) — collapses `false`, `undefined`, `null`, `0`, `""` into the same path. Strict `=== false || === undefined` would be more intentional.
- **No error-path test in SSE response** — no test for observable error mid-stream. The implementation logs "SSE stream error" but it's not verified.
- **Docs show `<CopilotKit>`** but actual v2 component is `CopilotKitProvider`. Minor — both names likely work via re-export.
- **`as any` on event summary** (`sse-response.ts:142`) — Alem's justification (debug-only, read-only, 20+ union types) is reasonable for v1. Non-blocking.

---

## What's Good

- **Type design**: `boolean | granular object` with `resolveDebugConfig()` normalizer — clean, extensible
- **Server-side logging**: Event summaries, lifecycle bookends, `eventCount`/`loggedEventCount` distinction
- **marthakelly's feedback**: All 3 actionable items properly addressed
- **Hot path**: Cheap when disabled — simple falsy checks, no string formatting
- **Pino is server-only**: No client bundle impact
- **Documentation**: Accurate troubleshooting page with correct defaults table
- **Backward compatible**: All additions are optional fields

## Recommended Actions

**Must-fix:**
1. Add `debug` field to `ProxiedCopilotRuntimeAgentConfig` and store it
2. Add changeset
3. Cache pino logger (singleton or on runtime instance)

**Should-fix:**
4. Add client-side logging or fix JSDoc to match reality
5. Default `verbose: false` for the boolean shorthand (PII)
6. Document ag-ui#1398 merge order

jpr5:
## CR Fixes Pushed

Addressed all findings from the code review. 6 commits on top of Alem's work:

### Fixes

| Commit | Finding | What Changed |
|--------|---------|--------------|
| `971a816` | **M5: PII risk** | `verbose` now defaults to `false` when `debug: true`. Users must explicitly opt in with `{ verbose: true }` to get full payload logging. |
| `6689de4` | **H1: Broken threading** | Added `debug` field to `ProxiedCopilotRuntimeAgentConfig`, stored in constructor. `AgentRegistry` now resolves the config before passing it. `clone()` preserves it. |
| `fde4e0b` | **H2: Per-request logger** | Pino logger is now created once in the runtime constructor and passed to handlers, instead of instantiating a new pino + pino-pretty per request. |
| `295a1e6` | **M6: Prop sync** | Added `setDebug()` to `CopilotKitCore` and wired it into the provider's prop-sync `useEffect`. |
| `39b4ea6` | **H3: Misleading docs** | Updated JSDoc and troubleshooting docs to accurately describe client-side behavior (config forwarded to AG-UI transport, not CopilotKit logging). Removed fabricated `console.debug` output examples. |
| `3a79f8a` | **H4: Missing changeset** | Added `.changeset/debug-mode.md` with minor bumps for shared, runtime, react-core, core. |

### New Tests

- **6 tests** in `debug-config-threading.test.ts` — verifies full chain: `CopilotKitCore` → `AgentRegistry` → `ProxiedCopilotRuntimeAgent`, including `clone()` preservation
- **3 tests** in `debug.test.ts` — verifies verbose defaults to false, explicit verbose opt-in works
- **1 test** in `debug-sse-response.test.ts` — verifies pre-created logger is reused instead of creating per-request

### CI

All green except `unit (20.x)` which fails on a pre-existing `clipboard.test.ts` bug (`navigator is not defined` — no jsdom environment). Unrelated to our changes. `unit (22.x)` and `unit (24.x)` both pass clean.

@AlemTuzlak ready for your re-review.

ranst91 (approved):

PR #3564, opened 2026-04-01
fix(sdk-python): align runtime info contract with JS runtime

kalyanaramansanthanam:
## Summary
This aligns `sdk-python` runtime discovery with the contract already used by the JavaScript runtime and consumed by the React/core agent registry.

## Problem
The JS runtime now returns runtime info in this shape:
- `version`
- `agents` as a record keyed by agent id/name
- `audioFileTranscriptionEnabled`
- `mode`
- optional flags like `a2uiEnabled`

But `sdk-python` still returns the older shape:
- `sdkVersion`
- `actions`
- `agents` as a list

That mismatch causes JS runtime consumers to treat array indexes as agent ids, which can surface as errors like:
- `Agent onboarding_assistant not found`
- `Known agents: [0]`

## What Changed
- change `sdk-python/copilotkit/sdk.py` so `info()` returns JS-style runtime metadata:
  - `version`
  - keyed `agents`
  - `audioFileTranscriptionEnabled: False`
  - `mode: "sse"`
  - `a2uiEnabled: False`
- add `Agent.runtime_info_repr()` in `sdk-python/copilotkit/agent.py` so the Python SDK can emit the same agent description shape as the JS runtime (`name`, `description`, `className`)
- keep legacy `actions` and `sdkVersion` keys temporarily for compatibility with older Python integrations
- update `sdk-python/copilotkit/html.py` so the browser info page can render both the new keyed-agent shape and the legacy list shape
- add a regression test for the runtime info contract
- update the troubleshooting snippet to show the current runtime discovery payload

## Why This Approach
The JS runtime and shared types have already migrated to a keyed runtime info contract. This PR makes `sdk-python` match that upstream contract instead of requiring application-specific response rewrites.

## Notes
I was not able to run the Python test suite in this environment because the local filesystem is effectively full, so this PR has not been executed end-to-end here.


sdk-python/copilotkit/html.py
```diff
@@ -147,17 +147,20 @@ def generate_info_html(info: InfoDict) -> str:
...
             arguments=json.dumps(action.get("parameters", []), indent=2),
         )
+
+    raw_agents = info["agents"]
+    agent_iterable = raw_agents.values() if isinstance(raw_agents, dict) else raw_agents
     agent_html = ""
-    for agent in info["agents"]:
-        agent_type = agent.get("type", "Unknown")
+    for agent in agent_iterable:
+        agent_type = agent.get("type") or agent.get("className", "Unknown")
```
marthakelly, line 163: New agents have `className: "LangGraphAGUIAgent"` but no `type` field, so the if `agent_type == "langgraph"` guard never fires and the badge renders the raw class name. Either update the guards to match the new `className` values, or have `runtime_info_repr()` emit a `type` field too.

sdk-python/copilotkit/sdk.py
```diff
@@ -247,21 +249,29 @@ def info(
         agents = self.agents(context) if callable(self.agents) else self.agents
 
         actions_list = [action.dict_repr() for action in actions]
-        agents_list = [agent.dict_repr() for agent in agents]
+        agents_dict = {
+            agent.name: agent.runtime_info_repr()
+            for agent in agents
```
marthakelly, line 254: nit: duplicate agent names silently overwrite each other. Worth a `logger.warning`

sdk-python/copilotkit/agent.py
```diff
@@ -59,8 +65,16 @@ async def get_state(
...
             'name': self.name,
             'description': self.description or ''
         }
+
+    def runtime_info_repr(self) -> RuntimeAgentDescriptionDict:
+        """Runtime info representation aligned with the JS runtime contract."""
+        return {
+            'name': self.name,
+            'description': self.description or '',
+            'className': self.__class__.__name__,
```
marthakelly, line 79: user subclasses will produce unexpected class names in the JS registry. A stable class attribute (`agent_type = "langgraph"`) would be safer than `__class__.__name__`.

sdk-python/copilotkit/html.py
```diff
@@ -147,17 +147,20 @@ def generate_info_html(info: InfoDict) -> str:
...
-    for action in info["actions"]:
+    for action in actions:
         action_html += ACTION_TEMPLATE.format(
             name=action["name"],
             description=action["description"],
             arguments=json.dumps(action.get("parameters", []), indent=2),
         )
+
+    raw_agents = info["agents"]
+    agent_iterable = raw_agents.values() if isinstance(raw_agents, dict) else raw_agents
```
marthakelly, line 160: `generate_info_html` shouldn't be deciding what shape agents is in, that's `info()`'s job. Normalize there once and every caller just gets a dict, and any future consumer of `InfoDict` won't need to copy this check.

There's also a type mismatch: `InfoDict` declares agents as always a dict, so the else branch is unreachable from mypy's perspective. The signature needs updating to reflect that this function accepts both shapes.

sdk-python/copilotkit/sdk.py
```diff
@@ -247,21 +249,29 @@ def info(
...
             ]
         )
 
         return {
+            "version": COPILOTKIT_SDK_VERSION,
+            "agents": agents_dict,
+            "audioFileTranscriptionEnabled": False,
+            "mode": "sse",
+            "a2uiEnabled": False,
+            # Legacy keys kept temporarily for older Python integrations.
```
marthakelly, line 272: nit: The codebase already uses `warnings.warn(..., DeprecationWarning)`, worth following the same pattern here.

sdk-python/copilotkit/sdk.py
```diff
@@ -30,12 +30,14 @@
...
-    actions: List[ActionDict]
-    agents: List[AgentDict]
+    """Runtime info dictionary for CopilotKit remote discovery."""
+    version: str
+    agents: Dict[str, RuntimeAgentDescriptionDict]
+    audioFileTranscriptionEnabled: bool
+    mode: Literal["sse"]
+    a2uiEnabled: bool
+    actions: NotRequired[List[ActionDict]]
+    sdkVersion: NotRequired[str]
```
marthakelly, line 40: nit: `actions` and `sdkVersion` are marked `NotRequired` but `info()` always returns them. That forces callers to use defensive `.get()` for keys that are guaranteed present. Either make them required in the `TypedDict`, or split into two types, one for the wire format and one for the Python return value.

sdk-python/tests/test_runtime_info.py
```diff
@@ -0,0 +1,81 @@
...
+
+    info = sdk.info(
+        context={
+            "properties": {},
+            "frontend_url": None,
+            "headers": {},
+        }
+    )
+
+    assert info["version"]
```
marthakelly, line 50: `assert info["version"]` checks truthiness. `COPILOTKIT_SDK_VERSION` falls back to "" when the package isn't pip-installed, which is typical in test environments.`assert info["version"] == COPILOTKIT_SDK_VERSION` is safer.

marthakelly (commented):
Looks good and it's close to landing! This aligns the Python SDK with the JS contract cleanly. The approach is solid, the test proves the new contract, and the legacy key preservation is thoughtful. My comments are mostly small quality of life things.

There are a few gaps in tests that would be good to address as well:

`sdk-python/tests/test_runtime_info.py`
No test for `description=None`. The or '' coercion in `runtime_info_repr` and the HTML rendering both touch this path and neither is exercised.

`sdk-python/tests/test_runtime_info.py`
No test for multiple agents, doesn't verify all agents are keyed correctly, and won't catch the silent-overwrite on duplicate names.

`sdk-python/tests/test_runtime_info.py`
Both HTML tests register zero actions, so only the "No actions found" fallback is exercised. The action-rendering path in `generate_info_html` is never hit.

BenTaylorDev:
Thanks for this, and apologies for the long silence.

I checked the change against the JS contract directly and the alignment is accurate — `RuntimeInfo` in `packages/shared/src/utils/types.ts` does require `version`, `agents: Record<string, AgentDescription>`, `audioFileTranscriptionEnabled` and `mode`, and `AgentDescription` is `{name, className, description}`. The Python shape here matches it.

I'm still closing it, because nothing reads that response anymore. `sdkVersion` has no consumers left in `packages/`, and the v1 remote-endpoint discovery that used to parse it is inert — `fetchRemoteInfo` is never called, and `assignEndpointsToAgents` returns an empty record. Meanwhile `copilotkit` is a published package, and changing `agents` from a list to a dict is a real break for anyone parsing `/info` today. That's a nonzero compatibility cost for no behavior change on our side, so the trade doesn't work out.

One piece is worth keeping regardless: the stray `print(info, flush=True)` you removed from `html.py` is a debug print that runs on every `/info` render in a published library. If you want to send that as a one-line PR I'll get it merged quickly — otherwise we'll pick it up.

The docs snippet doesn't need the accompanying change: it documents what the Python SDK emits today, which is unchanged if this doesn't land.

PR #3567, opened 2026-04-01
fix(sdk-python): align runtime info contract

deepshekhardas:
Aligns Python SDK runtime info with JS runtime keyed-agent contract.


BenTaylorDev:
Closing as a duplicate of #3564, which is the canonical version of this fix (identical diff). Let's keep the discussion there.

PR #3572, opened 2026-04-01
feat(runtime): add factory mode to BuiltInAgent for backend-agnostic LLM integration

AlemTuzlak:
## Summary

Adds **factory mode** to `BuiltInAgent`, enabling backend-agnostic LLM integration without introducing a new class. Users provide a factory function that creates the LLM stream — CopilotKit handles lifecycle events, stream-to-AG-UI conversion, error handling, and abort/cancellation.

### What's included

- **Three backend types**: `aisdk` (Vercel AI SDK), `tanstack` (TanStack AI), and `custom` (raw AG-UI events)
- **Stream converters**: `convertAISDKStream` and `convertTanStackStream` extract AG-UI events from backend-native streams
- **Input helpers**: `convertInputToTanStackAI` and `convertMessagesToVercelAISDKMessages` for message/prompt conversion
- **State management**: Automatic `STATE_SNAPSHOT`/`STATE_DELTA` event extraction from tool results
- **Reasoning support**: Auto-close logic for providers that never emit `reasoning-end` (e.g. `@ai-sdk/anthropic`)
- **273 tests** across 16 test files covering all backend types, converters, state tools, multimodal, and edge cases
- **Docs**: Comprehensive factory mode guide with 16 examples covering tools, reasoning, state, forwardedProps, structured output
- **Example app**: React Router example migrated to factory mode

### Architecture

Factory mode is a discriminated union on `BuiltInAgentConfiguration`:

```typescript
const agent = new BuiltInAgent({
  type: "aisdk",           // or "tanstack" | "custom"
  factory: ({ input, abortSignal }) =>
    streamText({ model: openai("gpt-4o"), messages: ..., abortSignal }),
});
```

No new public classes — `BuiltInAgent` handles both simple mode (model string) and factory mode (user-owned LLM call).

### Critical fixes
- **C1/M3**: Fixed TOCTOU window — `abortController` now set synchronously before Observable creation in classic `run()`, with concurrent run guard
- **C2/M2**: Classic `tool-result` now checks both `output` and `result` property names, with `try/catch` on `JSON.stringify`
- **C3**: Error case safely handles `undefined` `p.error` with proper fallback chain

### Important fixes
- **I2**: `RUN_ERROR` events now include `threadId` and `runId` in classic mode (matching factory mode)
- **I3**: State tool events guarded against `undefined` snapshot/delta before emitting
- **I4**: `JSON.stringify` wrapped in try/catch with descriptive fallback message
- **M1**: `assignToolsToAgents` now skips factory-mode agents (prevents config corruption)
- **M5**: TanStack converter's dropped event types documented with clear limitation note

### Suggestions addressed
- **S1**: `@ts-expect-error` on `clone()` middlewares documented with rationale
- **M4**: `AgentFactoryContext.abortController` kept (needed by TanStack AI) but JSDoc added discouraging direct `.abort()` calls

## Test plan

- [x] 273 unit tests pass across all 3 backend types
- [x] Build succeeds
- [x] React Router example compiles
- [x] Docs render correctly (Vercel preview)


jpr5 (changes requested):
## Code Review — Agent Class with Multi-Backend Support

Nice work on the architecture — the discriminated union pattern, clean separation of lifecycle management from stream conversion, and the reasoning auto-close logic for providers like `@ai-sdk/anthropic` are all well-designed. The 76 parameterized tests and the `eventField<T>` helper eliminating `as any` casts are solid.

Found 3 critical issues, 7 important issues, and several suggestions.

---

### 🔴 Critical — Must Fix

**C1. Concurrent `run()` overwrites `abortController` — first run becomes unabortable**

`agent.ts`: `this.abortController` is a single instance field. A second `run()` call overwrites it, and the `finally` block from whichever run finishes first nulls it for the other. Either guard at the top of `run()`:

```typescript
if (this.abortController) {
  throw new Error("Agent is already running. Call abortRun() first or create a new instance.");
}
```

…or scope the controller entirely within the closure and remove the instance field.

**C2. AI SDK `tool-result` may use wrong property name (`output` vs `result`)**

`aisdk.ts`:
```typescript
const toolResult = "output" in p ? p.output : null;
```

AI SDK historically uses `result`, not `output`. If the actual SDK property is `result`, this silently drops ALL tool results — state tools (`AGUISendStateSnapshot`/`AGUISendStateDelta`), tool call results, everything falls through to `null`. Every `ToolCallResultEvent.content` becomes `"null"`.

Please verify against the actual AI SDK types. The safe fix:
```typescript
const toolResult = "output" in p ? p.output : ("result" in p ? p.result : null);
```

**C3. `throw p.error` throws `undefined` when AI SDK error event lacks `.error` property**

`aisdk.ts`, `case "error"`: throws `p.error` which may be `undefined`. This produces `RUN_ERROR.message = "undefined"` — useless for debugging provider errors (rate limits, auth failures, etc.).

```typescript
case "error": {
  if (abortSignal.aborted) break;
  const err = p.error ?? p.message ?? p.cause;
  if (err instanceof Error) throw err;
  throw new Error(typeof err === "string" ? err : `AI SDK stream error: ${JSON.stringify(p)}`);
}
```

---

### 🟡 Important — Should Fix

**I1. TanStack converter silently drops `TOOL_CALL_RESULT` events**

`tanstack.ts` only handles 4 chunk types (`TEXT_MESSAGE_CONTENT`, `TOOL_CALL_START/ARGS/END`). `TOOL_CALL_RESULT` is silently ignored. Tool-using agents with TanStack backend will have results vanish from the event stream — the UI shows tool calls stuck pending, and multi-turn tool use breaks.

The AI SDK converter handles this; TanStack doesn't. At minimum add `TOOL_CALL_RESULT` handling. If other event types (state, reasoning) are deferred, add a TODO comment listing the gaps.

**I2. `RUN_ERROR` event missing `threadId` and `runId`**

`agent.ts`: `RUN_STARTED` and `RUN_FINISHED` both include `threadId` and `runId`. `RUN_ERROR` omits them, breaking run correlation for downstream consumers.

**I3. State tool events emitted with `undefined` data when tool result shape is unexpected**

`aisdk.ts`: `StateSnapshotEvent.snapshot` will be `undefined` if the tool returns `{ state: {...} }` instead of `{ snapshot: {...} }`. No validation before emit. Guard with `if (snapshot !== undefined)` before yielding.

**I4. `JSON.stringify(toolResult)` can throw on circular references**

`aisdk.ts`: bare `JSON.stringify` with no try/catch in the `tool-result` case. Circular tool results kill the entire run with a confusing TypeError. Wrap with try/catch and include tool name in the fallback message.

**I5. `convertInputToTanStackAI` has zero test coverage**

This is the primary user-facing API for TanStack integration — documented in the guide, used in the example app, exported publicly. It has 6+ distinct code paths (system message filtering, tool call mapping, context injection, state serialization) and 0 tests.

**I6. State management tool results (`AGUISendStateSnapshot`/`AGUISendStateDelta`) untested**

The special tool-name-to-state-event conversion in the AI SDK converter has no test coverage.

**I7. `collectEventsIncludingErrors` discards the error object**

`agent-test-helpers.ts`: the helper resolves identically whether the observable errors or completes normally (`error: () => resolve(events)`). Tests can't distinguish between the two, which could mask regressions where `RUN_ERROR` stops being emitted.

---

### 🔵 Suggestions

**S1.** `clone()` uses `@ts-expect-error` to access `this.middlewares` from parent — fragile. Document the coupling or expose `getMiddlewares()` on `AbstractAgent`.

**S2.** Missing exhaustiveness check in `switch(this.config.type)` — add a `default: never` guard for future-proofing.

**S3.** Docs comparison table says "Vercel AI SDK v6 only" — there is no v6. Fix version reference.

**S4.** Docs reference `useCopilotChatSuggestions` for shared state — wrong hook name.

**S5.** Async factory (Promise-returning) paths are documented and type-supported but untested.

**S6.** Abort test uses hardcoded 30ms `setTimeout` — potentially flaky on slow CI. Consider signaling-based synchronization.

---

### ✅ What's Working Well

- Discriminated union pattern with correct TypeScript narrowing
- Lifecycle management cleanly separated from stream conversion
- Reasoning auto-close handles `@ai-sdk/anthropic` never emitting `reasoning-end`
- 76 parameterized tests across all 3 backend types
- `eventField<T>` helper eliminates all `as any` casts in tests
- Concurrent tool call handling (toolCallStates map) is solid
- Example app migration demonstrates the simplification clearly
- Docs are comprehensive with 16 examples covering real use cases

---

**Summary:** 3 critical (concurrent run race, possibly wrong tool-result property, undefined error), 7 important. The most dangerous is C2 — please verify the AI SDK `tool-result` property name against the actual types. If it's `result` not `output`, all tool functionality is silently broken.

marthakelly (dismissed):
Approving so you can address whichever feedback you agree with and merge at-will

## Highest-Priority Findings

1. **`copilot-runtime.ts:483` — `assignToolsToAgents` corrupts factory-mode agent configs.** This is a silent data mutation bug that affects anyone using factory-mode agents in a `CopilotRuntime` with runtime-level actions. It'll be hard to debug because the effect only shows up at run time when the agent misbehaves. The fix is a two-line `isFactoryConfig` guard.

2. **`index.ts:1071` — Classic `run()` and `convertAISDKStream` have already diverged.** The tool-result path in classic mode is missing both the `"result"` fallback and the `JSON.stringify` try/catch. These are real behavioral differences for users on older AI SDK versions or with non-serializable tool results. The right fix is to have classic `run()` delegate to the converter — which also eliminates ~200 lines of duplication.

3. **`index.ts:951` — `this.abortController` set inside async IIFE in classic `run()`.** There's a microtask window where `abortRun()` silently does nothing. `runFactory` already has the correct pattern (`// Set synchronously before Observable creation to close TOCTOU window`). Classic mode should match.

4. **`index.ts:562` — `abortController` in `AgentFactoryContext` should be removed before this API ships.** Exposing write access to the abort controller to factory authors is a footgun. It's easy to remove now; it's a breaking change after publication.

5. **`tanstack.ts:165` — `STATE_SNAPSHOT`/`STATE_DELTA`/`REASONING` silently dropped by TanStack converter.** Users won't discover this until they try to use shared state with a TanStack backend and nothing happens. Surface it as a JSDoc limitation on the config type.

marthakelly (dismissed):

jpr5 (dismissed):
All 7 findings from Round 1 verified fixed:

- **C1** Concurrent run() guard — ✓
- **C2** tool-result checks both `output` and `result` — ✓
- **C3** Error fallback chain (`p.error ?? p.message ?? p.cause`) — ✓
- **I1** TanStack TOOL_CALL_RESULT handling added — ✓
- **I2** RUN_ERROR includes threadId/runId — ✓
- **I3** State events guarded against undefined — ✓
- **I4** JSON.stringify wrapped in try/catch — ✓

No new bugs introduced. LGTM.

PR #3576, opened 2026-04-01
feat: add AG-UI event emitter to web inspector via TanStack DevTools EventClient

AlemTuzlak:
## Summary

- **New package `@copilotkit/devtools-client`** — TanStack DevTools `EventClient` singleton for emitting AG-UI events, with localStorage-based snippet persistence for saving/replaying event configurations
- **Core `DevtoolsListener`** — subscribes to devtools events and expands high-level operations (tool call, text message, reasoning, state snapshot, custom event) into full AG-UI event sequences dispatched through the target agent's subscriber chain, with automatic run lifecycle wrapping
- **Web Inspector "Event Emitter" tab** — new menu section with agent picker, event type selector, structured forms per event type, emit/save-as-snippet buttons, and snippet management (replay/delete)

## How it works

The web inspector UI calls `devtoolsClient.emit("tool-call", { agentId, toolName, args, result })`. Core's `DevtoolsListener` receives this via the EventClient's local pub/sub (`listenToSelf` pattern) and expands it into the full AG-UI sequence: `RUN_STARTED → TOOL_CALL_START → TOOL_CALL_ARGS → TOOL_CALL_END → TOOL_CALL_RESULT → RUN_FINISHED`. These events flow through the real agent subscriber pipeline, triggering actual side effects (tool execution, state updates, chat UI rendering).

## Test plan

- [ ] `@copilotkit/devtools-client` — 11 tests: EventClient emit/on roundtrip for all 5 event types + snippet CRUD (load, save, delete, corrupt localStorage handling)
- [ ] `@copilotkit/core` — 7 new tests for DevtoolsListener event expansion (tool-call, text-message, reasoning, state-snapshot, custom-event, agent-not-found no-op, skip-run-lifecycle-when-mid-run) + 324 existing tests pass with zero regressions
- [ ] Manual: open web inspector, navigate to Event Emitter tab, select an agent, compose a tool call, emit it, verify it appears in AG-UI Events tab
- [ ] Manual: save a snippet, reload page, verify snippet persists, replay it


marthakelly (commented):
# PR #3576 Review — CopilotKit/CopilotKit

**Title:** feat: add AG-UI event emitter to web inspector via TanStack DevTools EventClient **Author:** AlemTuzlak | **Status:** Open (Draft) | **Base:** main

---

## Critical Issues

**(1)** `DevtoolsListener` **is never destroyed — resource/subscription leak**

`packages/core/src/core/core.ts:253` — `CopilotKitCore` creates and initializes a `DevtoolsListener` but never calls `destroy()`. The class has no teardown method. Since `devtoolsClient` is a module-level singleton, each `CopilotKitCore` construction (e.g., React re-mount) registers additional event handlers on the same global `devtoolsClient` without cleaning up old ones. The `active` guard mitigates crashes, but stale listeners accumulate as a memory and CPU leak.

**Fix:** Add a `destroy()` method to `CopilotKitCore` that calls `this.devtoolsListener.destroy()`, or make `DevtoolsListener.initialize()` idempotent by cleaning up previous handlers first.

---

**(2)** `notifySubscribers` **has no error handling — one failing subscriber aborts the entire event chain**

`packages/core/src/core/devtools-listener.ts:84-88` — If any subscriber callback throws, the iteration loop aborts. In `handleToolCall`, a throw in `onToolCallStartEvent` means `onToolCallArgsEvent`, `onToolCallEndEvent`, `onToolCallResultEvent`, and `onRunFinishedEvent` never fire. The agent receives `RUN_STARTED` but never `RUN_FINISHED`, leaving it in an inconsistent lifecycle state.

The existing `CopilotKitCore.notifySubscribers()` (`core.ts:282-295`) wraps each subscriber call in try-catch. `DevtoolsListener` bypasses this established pattern.

**Fix:** Wrap each subscriber callback in try-catch, consistent with `CopilotKitCore.notifySubscribers`. Consider having `withRunLifecycle` guarantee `RUN_FINISHED` fires via try-finally.

---

**(3) Missing UMD externals —** `@copilotkit/devtools-client` **will be inlined into UMD bundles**

`packages/core/tsdown.config.ts` and `packages/web-inspector/tsdown.config.ts` — Both packages now depend on `@copilotkit/devtools-client`, but neither adds it to the UMD build's `external` array or `globals` map. Every other `@copilotkit/*` workspace dependency is listed there. The UMD bundles will attempt to inline the entire devtools-client and its transitive dependency `@tanstack/devtools-event-client`, producing oversized or broken builds.

**Fix:** Add `"@copilotkit/devtools-client"` to `external` and a corresponding global entry in both UMD config blocks.

---

## Important Issues

**(4)** `window.prompt()` **blocks the main thread and fails silently in embedded contexts**

`packages/web-inspector/src/index.ts` — Snippet naming uses `window.prompt()`, which is a synchronous, thread-blocking call. Worse, `window.prompt()` silently returns `null` in many embedded webviews, mobile browsers, and sandboxed iframes — exactly the environments where a devtools inspector is likely to be rendered. In SSR, `window` is undefined entirely. The user clicks "Save Snippet," the browser either freezes momentarily or the call returns `null` with no feedback, and nothing is saved.

**Fix:** Replace `window.prompt()` with an inline text input or a small custom modal/popover form. This avoids thread blocking, works in all rendering contexts, and gives you a place to show validation feedback.

---

**(5)** `handleReplaySnippet` **bypasses validation — corrupt snippets go straight to emit**

`packages/web-inspector/src/index.ts` — `handleReplaySnippet()` reads a saved snippet's payload and calls `devtoolsClient.emit()` directly, without running `isEmitterFormValid()` or any equivalent guard. A snippet saved against an older schema version, or one with a manually edited/corrupt payload in `localStorage`, is emitted as-is. Compare with `handleEmit()`, which at least goes through the form state — replay has no such gate.

**Fix:** Run the same validation that `handleEmit()` uses (or a payload-level equivalent) before replaying. If the snippet is invalid, surface an error explaining that the snippet may be outdated.

---

**(6) Module-level singleton instantiated unconditionally at import time**

`packages/devtools-client/src/event-client.ts:25` — `export const devtoolsClient = new CopilotKitEventClient()` executes at module evaluation time. Since `@copilotkit/core` depends on `@copilotkit/devtools-client`, every application using `@copilotkit/core` — including production builds — will instantiate this `EventClient` and its underlying `EventTarget`, even when devtools are not being used.

**Fix:** Use lazy initialization or make `DevtoolsListener` opt-in via a config flag.

---

**(7)** `devtoolsListener.initialize()` **can crash** `CopilotKitCore` **constructor**

`packages/core/src/core/core.ts:262` — Called with no try-catch. If the `EventClient.on()` calls throw (e.g., `EventTarget` unavailable in SSR), the entire `CopilotKitCore` constructor fails. A developer convenience feature should never prevent the core SDK from initializing.

**Fix:** Wrap in try-catch, log a warning, continue without devtools.

---

**(8)** `getEmitterPayload()` **calls** `JSON.parse()` **without try-catch**

`packages/web-inspector/src/index.ts` (`getEmitterPayload()`) — Calls `JSON.parse` on `emitterArgs`, `emitterStateJson`, and `emitterCustomValue` with no error handling. While `validateJson()` runs on `@input`, the validation state could be stale. If `emitterJsonError` is null while JSON fields contain invalid text, an unhandled `SyntaxError` crashes the component.

**Fix:** Wrap `JSON.parse` calls in try-catch inside `getEmitterPayload()`, or re-validate at call time.

---

**(9)** `saveSnippet()` **and** `deleteSnippet()` **have no error handling**

`packages/devtools-client/src/snippets.ts:16-26` — `localStorage.setItem()` can throw `DOMException` (quota exceeded, private browsing, sandboxed iframes). These propagate unhandled through `handleSaveSnippet()` and `handleDeleteSnippet()` in the web inspector, potentially crashing the UI.

**Fix:** Wrap `localStorage.setItem` in try-catch and surface errors to the caller.

---

**(10)** `DevtoolsListener` **silently drops events for unknown agents with no logging**

`packages/core/src/core/devtools-listener.ts` (all `handleXxx` methods) — Every handler does `if (!agent) return;` with no warning. The user explicitly selected an agent and clicked "Emit Event" — they expect it to work. Silent drops make debugging nearly impossible.

**Fix:** Add `console.warn` when an agent is not found.

---

**(11)** `handleEmit()` **provides no success/failure feedback**

`packages/web-inspector/src/index.ts` (`handleEmit()`) — Calls `devtoolsClient.emit()` with no error handling and no visual confirmation. The user clicks "Emit Event" and has no idea if it worked. For a devtools feature, this defeats the purpose.

**Fix:** Add try-catch around emit. On success, show brief confirmation. On failure, display error.

---

## Suggestions

**(12) Identical** `text-message` **and** `reasoning` **cases can be merged**

`packages/web-inspector/src/index.ts` — In both `getEmitterPayload()` and `isEmitterFormValid()`, the `"text-message"` and `"reasoning"` cases return the exact same value. These can be combined into a single fall-through case. Beyond being shorter, this eliminates a divergence risk: if either event type later gains a field, having them in separate branches makes it easy to update one and forget the other.

---

**(13) Eight** `as any` **casts bypass the typed event system**

`packages/core/src/core/devtools-listener.ts:64` (3 casts), `packages/web-inspector/src/index.ts` (4 casts at emit sites), and `notifySubscribers` uses `sub: any`. The carefully constructed `CopilotKitDevtoolsEvents` → `StripPrefix` → `CopilotKitEventSuffixes` type pipeline is systematically bypassed at every usage site. Make `registerHandler` generic over the event key to eliminate most casts.

**(14)** `DevtoolsEventType` **can drift from** `CopilotKitDevtoolsEvents`

`packages/devtools-client/src/types.ts` — The union type and the interface keys are maintained independently. Derive the union: `type DevtoolsEventType = keyof CopilotKitDevtoolsEvents extends \`copilotkit:${infer S}\` ? S : never;\`

**(15) No double-initialize guard on** `DevtoolsListener`

`packages/core/src/core/devtools-listener.ts` — Calling `initialize()` twice registers duplicate handlers. Add `if (this.active) return;` at the top of `initialize()`.

**(16)** `DevtoolsSnippet.payload` **is** `Record<string, unknown>` **— loses type safety**

`packages/devtools-client/src/types.ts` — Use a discriminated union tying `eventType` to a specific payload shape, making `getEmitterPayload()` and `handleReplaySnippet()` type-safe without casts.

**(17) Common** `agentId` **field not factored out of event payloads**

`packages/devtools-client/src/types.ts` — Each event payload repeats `agentId: string`. Extract a `DevtoolsEventBase` and intersect it into each payload to prevent a new event from omitting it.

**(18) Event-client tests don't clean up listeners**

`packages/devtools-client/src/__tests__/event-client.test.ts` — Each test calls `devtoolsClient.on()` but never unsubscribes. Handlers accumulate on the module-level singleton across tests, risking false positives.

**(19) Missing test:** `destroy()` **prevents post-destroy event processing**

`packages/core/src/__tests__/devtools-listener.test.ts` — The `active` guard is explicitly coded but never tested. A test should emit after `destroy()` and assert no callbacks fire.

**(20) TOOL_CALL_RESULT payload assertions incomplete**

`packages/core/src/__tests__/devtools-listener.test.ts` — Commit 8dedcd1 fixed the TOOL_CALL_RESULT schema (`content`, `messageId`, `role`), but tests don't assert these fields. This is exactly the class of bug already fixed once in this PR.

**(21) New package missing** `engines` **field**

`packages/devtools-client/package.json` — Sibling packages declare `"engines": { "node": ">=18" }`. Add for consistency.

**(22) Unverifiable "library bug" claim in comment**

`packages/core/src/core/devtools-listener.ts:58-60` — References a bug in `@tanstack/devtools-event-client` with no issue link. Rewrite to describe the defensive pattern on its own merits, or add a link to a bug report.

**(23) No web-inspector logic tests**

`packages/web-inspector/src/index.ts` — `validateJson()`, `isEmitterFormValid()`, `getEmitterPayload()`, and `revalidateEmitterJson()` contain non-trivial branching logic with no test coverage. These are the gatekeepers for what gets emitted to agents.

---

## Strengths

- **Clean package separation**: `@copilotkit/devtools-client` is a well-scoped, independently testable package with proper barrel exports and dual ESM/CJS/UMD build targets.
- **Thoughtful** `withRunLifecycle` **abstraction**: Cleanly separates the run envelope from individual event handling, reducing duplication across five handlers.
- **Good** `DevtoolsListenerDeps` **interface**: Narrow, single-method dependency with `Readonly` modifier — easy to mock, minimal surface area.
- **Solid snippet persistence design**: `agentId` intentionally excluded from `DevtoolsSnippet` for cross-agent reusability is a non-obvious, well-documented design decision.
- **Good edge case test coverage**: "agent not found" and "already running" tests verify the two most important guard conditions with precise assertions.
- `StripPrefix` **type utility**: Well-crafted type-level machinery that correctly derives suffix-keyed event maps from prefixed ones.

jpr5 (commented):
## Code Review: feat: add AG-UI event emitter to web inspector

Good feature, @AlemTuzlak — the TanStack DevTools EventClient pattern is clean, the 5-event-type expansion in DevtoolsListener is well-structured, and the snippet persistence is a nice touch. The test suite (32 tests) covers happy paths thoroughly. Pushing fixes for all findings below.

---

## Critical — CI Blocker

### 1. Stale `pnpm-lock.yaml`

`packages/devtools-client/package.json` removed `@ag-ui/client` and `eslint` but the lockfile wasn't regenerated. `pnpm install --frozen-lockfile` fails, cascading into every CI job. Fix: regenerate lockfile.

---

## High

### 2. Subscriber/state dispatch mismatch in DevtoolsListener

`handleToolCall` (and all other handlers) uses `resolveAgent()` for subscriber notification but `resolveAllInstances()` for state mutation (addMessages/setState). These resolve differently: `resolveAgent` returns the registry agent, while `resolveAllInstances` returns clones when a resolver is set. This means:
- Subscriber events (TOOL_CALL_START, etc.) are dispatched on the **registry agent's** subscribers
- Messages are added to the **clones**

If clones have their own subscribers (which they will for per-thread agents), those subscribers never receive the AG-UI events. The registry agent's subscribers receive events about messages that were never added to the registry agent.

### 3. `deleteSnippet` returns `true` for nonexistent IDs

`deleteSnippet("nonexistent")` writes back the same array and returns `true`, implying success. Callers cannot distinguish "deleted" from "was never there."

### 4. No event ordering assertions in tests

Tests check "subscriber called once" but never verify events arrive in the correct AG-UI sequence. For tool-call, the sequence must be `RUN_STARTED → TOOL_CALL_START → TOOL_CALL_ARGS → TOOL_CALL_END → TOOL_CALL_RESULT → RUN_FINISHED`. Without ordering assertions, a reordering bug would pass all tests.

---

## Medium

### 5. DevtoolsListener is always-on with no opt-in gate

`DevtoolsListener` is created and initialized unconditionally in the `CopilotKitCore` constructor. Every CopilotKit user gets 5 event listeners registered on the EventClient's EventTarget regardless of whether devtools are in use. Production apps that never use devtools still pay the cost. Should be gated behind `showDevConsole` or similar.

### 6. No runtime payload validation on devtools events

Zero validation on event payloads in `devtools-listener.ts`. `handleToolCall` directly passes `payload.toolName` and `payload.args` to subscribers. If `args` is not a valid `Record<string, unknown>` at runtime, downstream consumers may behave unexpectedly. Since this is a devtools feature (not user-facing), the risk is lower, but defense-in-depth would help.

### 7. `listenToSelf` relies on upstream default

The entire devtools panel depends on `listenToSelf` being enabled so emitted events are received by the same client. This is not explicitly configured — it works only because `@tanstack/devtools-event-client` defaults `listenToSelf: true`. A single upstream change would silently break the package. Should be explicitly set.

### 8. `loadSnippets` silently drops malformed entries but never repairs storage

The `.filter()` strips invalid entries on every read, but the corrupt data remains in localStorage. Every subsequent call re-parses and re-filters the same junk. Should write back the cleaned array after filtering.

### 9. No snippet count/size limit

`saveSnippet` appends indefinitely. localStorage has a ~5MB limit per origin. No cap on count or total size — a user saving many large snippets will eventually hit `QuotaExceededError` with no warning.

### 10. Committed `excalidraw.log`

Empty file at repo root, tracked by git. Accidental — should be removed.

### 11. Missing changeset

New package `@copilotkit/devtools-client` plus changes to `core` and `web-inspector`. Changeset needed for the release pipeline.

### 12. License field mismatch

`devtools-client/package.json` says `"license": "MIT"` but CopilotKit packages use ELv2.

---

## Low / Nit

- `Record<string, any>` at two points in web-inspector `index.ts` — should be `Record<string, unknown>`
- UMD externals list doesn't match globals map in web-inspector tsdown config
- `parentMessageId` generated in tool-call handler but never appears in subscriber event payloads
- No tests for multiple listeners on same event, or unsubscribe behavior
- No integration test exercising full pipeline (emit → listener → subscriber in order)
- `emitDynamic` uses `as any` — justified with comment, but worth noting
- 5114-line monolithic `index.ts` in web-inspector (Event Emitter adds ~700 lines) — approaching split threshold

---

## What's Good

- **EventClient singleton + lazy init** — clean pattern, defers allocation until first use
- **Event expansion** — all 5 types correctly expanded into full AG-UI sequences
- **Snippet CRUD** — robust, handles QuotaExceededError, corrupt data
- **Type design** — `CopilotKitDevtoolsEvents` interface map with `StripPrefix` utility type is well-done
- **DevtoolsListener lifecycle** — properly destroyed, active guard, error resilience
- **Test coverage** — 32 tests with meaningful field-level assertions

Pushing fixes now.

jpr5:
## CR Fixes Pushed

Addressed all findings from the code review. Single commit on top of Alem's work:

### CI Blocker
- **Regenerated `pnpm-lock.yaml`** — stale after devtools-client deps changed, was failing `--frozen-lockfile`
- **Removed `excalidraw.log`** — accidentally committed empty file
- **Added changeset** — minor bumps for devtools-client, core, web-inspector

### Bug Fixes
- **`deleteSnippet` now returns `false` for nonexistent IDs** — previously returned `true` regardless
- **`loadSnippets` repairs corrupt localStorage** — writes cleaned array back after filtering invalid entries
- **Added `listenToSelf` documentation** — comment explaining dependency on upstream default

### Architecture
- **Gated `DevtoolsListener.initialize()` behind `showDevConsole`** — listener was always-on, registering 5 event listeners even in production. Now only initializes when devtools are explicitly enabled.

### New Tests (10 added)
- **Event ordering** (2): tool-call and text-message AG-UI sequence order verification
- **Integration** (2): full pipeline tests with toolCallId/messageId consistency across events
- **devtools-client** (4): deleteSnippet nonexistent ID, loadSnippets corrupt repair, multiple listeners, unsubscribe behavior
- Added `jsdom` devDep needed for snippet tests

### Known: snippet tests need env fix
The snippet tests require jsdom but the vitest `@vitest-environment jsdom` inline comment isn't being picked up by the current config. This is pre-existing on the branch — needs a vitest config file or environment setup. CI will validate once the main build passes.

### Not addressed (by design)
- **Subscriber/state dispatch mismatch** (CR finding #2) — `resolveAgent` vs `resolveAllInstances` targeting different instances. This is an architectural issue that needs Alem's input on the intended behavior for thread clones.

@AlemTuzlak ready for your review.

ranst91 (dismissed):

BenTaylorDev:
Closing this out — it's drifted well past the point where a rebase is realistic, not for any fault in the work.

Since April the inspector was restructured substantially: `packages/web-inspector/src/index.ts` has grown to over 20k lines and the navigation moved out to a typed structure in `src/lib/inspector-nav.ts` with grouped leaves. The Event Emitter tab here attaches to the old flat menu in a version of that file that no longer exists. A merge against current `main` also conflicts across the retired `docs/` tree and the `showcase/packages/` → `examples/` relocation, and CI is red.

For the record, nothing on `main` replaced this — the Playground leaf added later is a live chat surface, not an event injector, so composing and replaying arbitrary AG-UI sequences still isn't possible in the inspector. If it gets picked back up, `devtools-listener.ts` is the part worth carrying over; it's largely independent of the UI that moved.

One design note for whenever that happens: this put `@copilotkit/devtools-client` in `dependencies` of `@copilotkit/core`, so a devtools-only package became a runtime dependency for every consumer. Even with the `showDevConsole` gate, that wants a different shape — peer dep, separate entry point, or keeping the listener out of core entirely.

Thanks @AlemTuzlak — sorry it sat this long without anyone closing the loop.

PR #3587, opened 2026-04-02
feat: add throttleMs to useAgent for streaming render throttling

mxmzb:
## Summary

- Adds optional `throttleMs` parameter to `useAgent()` that coalesces rapid `onMessagesChanged` notifications into fewer React re-renders using a leading+trailing throttle
- Only `onMessagesChanged` is throttled (it fires per-token during streaming); `onStateChanged` and `onRunStatusChanged` always fire immediately
- `throttleMs=0` (default): exact current behavior, zero overhead
- Implemented as a **separate concern** — clean `useEffect` that overrides the `onMessagesChanged` subscription when `throttleMs > 0`

## Test plan

- [x] Default behavior unchanged (no throttleMs → every notification re-renders)
- [x] Leading edge fires immediately (first token shows instantly)
- [x] Coalescing within window (rapid tokens don't cause per-token re-renders)
- [x] Trailing edge fires (final state is never lost)
- [x] `onStateChanged` fires immediately regardless of `throttleMs`
- [x] Pending timer cleaned up on unmount (no setState-after-unmount)
- [x] All 930 existing tests pass


packages/react-core/src/v2/hooks/use-agent.tsx
```diff
@@ -94,7 +101,12 @@ function getOrCreateThreadClone(
   return clone;
 }
 
-export function useAgent({ agentId, threadId, updates }: UseAgentProps = {}) {
+export function useAgent({
+  agentId,
+  threadId,
+  updates,
+  throttleMs,
```
AlemTuzlak, line 124: instead of making this a react feature why not put this in core?

PR #3622, opened 2026-04-04
fix(react-core): preserve assistant text when multiple tool calls fire in one turn (CPK-7154)

### LINKED ISSUE #3470: Agent text response lost when multiple tool calls in single turn (dedup replaces text + earlier tools)
## Description

When an agent makes multiple tool calls in a single turn (e.g., `captureData` → `updateWorkingMemory`), the `CopilotChatMessageView` deduplicates all messages into a single view showing only the **last tool call indicator**. The initial text response and earlier tool call indicators are lost.

## Steps to Reproduce

1. Set up a CopilotKit sidebar with an agent that has multiple server-side tools
2. Send a message that triggers 2+ tool calls (e.g., "Record a headache symptom, severity 7/10")
3. Agent streams: text → tool call 1 (captureData) → tool call 2 (updateWorkingMemory) → final text
4. Observe the chat sidebar

## Expected Behavior

The chat should show:
- Brief acknowledgment text ("Let me record that...")
- Tool call 1 indicator (captureData — Done)
- Tool call 2 indicator (updateWorkingMemory — Done)  
- Final text response from the agent

## Actual Behavior

- Only the **last tool call indicator** is visible (updateWorkingMemory — Done)
- The initial text, first tool indicator, and final text response are all lost
- Console shows 100+ `CopilotChatMessageView: Deduplicated N messages` warnings
- The agent run completes successfully (`processAgentResult` is called)

## Context

- Using `@ag-ui/mastra` adapter with Mastra agents
- AG-UI multi-route protocol (`/agent/:id/connect` SSE streaming)
- `useDefaultRenderTool` registered for tool call indicators
- The AG-UI adapter emits proper `TEXT_MESSAGE_CHUNK`, `TOOL_CALL_START/END/RESULT` events with different `messageId`s per step (via `onFinishMessagePart`)

## Environment

- `@copilotkit/runtime`: 1.54.0
- `@copilotkitnext/react`: 1.54.0  
- `@ag-ui/mastra`: latest
- `@mastra/core`: 1.13.2
- Browser: Chrome (macOS)

## Related

Possibly related to #3041 (sequential tool rendering feature request). The core issue is that CopilotChatMessageView's deduplication logic collapses all messages within a single agent turn into one view, losing intermediate text and tool indicators.

## Workaround Attempted

- Instructing the agent to limit to one tool call per turn (agent ignores this when framework-injected tools like `updateWorkingMemory` are present)
- The text IS generated and streamed (visible momentarily before dedup replaces it)

marthakelly:
## Summary

- **Root cause**: `CopilotChatMessageView` deduplicated messages with `new Map(messages.map(m => [m.id, m]))`, which keeps only the *last* occurrence of each ID. During streaming, when an agent fires multiple tool calls in one turn, the same message ID appears multiple times: first with text content, then with empty content + tool calls appended. The last entry wins, wiping the text.
- **Fix**: Replace "keep last" with a **merge** strategy for assistant messages — recover non-empty `content` from any earlier occurrence while keeping the latest `toolCalls` (which accumulate). All other message roles retain "keep last" behavior.
- **Test**: Added a regression test covering the three-occurrence scenario (text → first tool call → second tool call) to confirm the merged message renders both the original text and the final tool call set.

## Test plan

- [x] New test `"preserves assistant text content when later duplicate has empty content (multi-tool-call scenario)"` passes
- [x] Existing dedup tests (single duplicate, order preservation, activity rendering) all pass unchanged
- [x] Full `@copilotkit/react-core` test suite: **950 passed, 0 failed** (baseline was 949)

Closes #3470


tylerslaton (dismissed):
Looks good to me! Posting a few comments from AI, would be good to resolve them but not required. Approving and you can decide how to proceed.

---

## Issues

1. **JSDoc inaccurately describes toolCalls behavior** — `CopilotChatMessageView.tsx:308-313`\
   \
   The JSDoc says "while keeping the latest toolCalls" but the implementation (lines 332-334) uses `??` which *recovers earlier toolCalls when the latest is undefined*. The inline comment on lines 330-331 is accurate and directly contradicts the JSDoc summary. Since the JSDoc is what consumers read first on this exported function, it should be corrected to say something like: "and similarly recovers toolCalls from earlier occurrences if the latest is undefined."


2) **Test "uses latest content" does not assert content value** — `CopilotChatMessageView.test.tsx:136-165`\
   \
   The test "uses latest content when all assistant duplicates have non-empty content" verifies message counts and absence of duplicate-key warnings, but never asserts *which* content value was kept. The test name claims latest content wins but nothing proves it. Should add:

```typescript
expect(assistantMessages[0].textContent).toContain("Full response from the assistant.");
```

3. **No test for empty-array** `[]` **vs** `undefined` **toolCalls** — `CopilotChatMessageView.test.tsx`\
   \
   The `??` operator for toolCalls is the critical design choice distinguishing "explicitly no tool calls" (`[]`) from "field not provided" (`undefined`). No test locks in this behavior. If someone later changes `??` to `||`, the regression would go undetected. Suggested test:

```typescript
it("keeps empty toolCalls array from later chunk (does not fall back)", () => {
  const messages: Message[] = [
    assistantMsg("a-1", "", [toolCall("tc-1", "fn")]),
    assistantMsg("a-1", "Done.", []),
  ];
  const result = deduplicateMessages(messages);
  expect(result).toHaveLength(1);
  expect((result[0] as AssistantMessage).toolCalls).toEqual([]);
});
```

---

## Suggestions

1. **Add test for** `content: undefined` **(distinct from** `""`**)** \
   \
   All tests use `""` to represent wiped content, but `AssistantMessage.content` is `string | undefined`. A test with explicitly `undefined` content would document that `||` handles both cases and guard against a future `||` → `??` refactor.
2. **Redundant** `as AssistantMessage` **type assertions** \
   \
   Lines 328-334. Since `Message` is a `ZodDiscriminatedUnion` on `role`, checking `message.role === "assistant"` already narrows the type. The casts on lines 328-329 and 333-334 are redundant (the one on line 340 for `acc.set()` is justified due to TypeScript spread-type limitations).
3. **Consider** `@internal` **annotation on export** \
   \
   `deduplicateMessages` is exported for testability but not re-exported through the barrel file. Adding `@internal` to the JSDoc would signal this isn't part of the public API.**(1) Missing changeset** — `CopilotChatMessageView.tsx`
4. No `.changeset/*.md` file is present. \
   \
   The changeset bot has flagged this. This is a patch-level bug fix for `@copilotkit/react-core` and must be added before merge.

ranst91 (approved):

PR #3629, opened 2026-04-06
fix: strip inline type modifiers from bundled .d.ts files

mxmzb:
## Summary

- Bundled `.d.ts` files use `export { type Foo }` syntax which is valid TypeScript but invalid JavaScript. Bundlers (webpack via tsconfig paths, Next.js) that accidentally resolve to these files crash with `JavaScript parse error: Expected ',', got 'Foo'`.
- The `copy-dts.mjs` post-build script now strips inline `type` modifiers from all declaration file exports: `export { type Foo, Bar }` → `export { Foo, Bar }`. The modifier is redundant in `.d.ts` files — TypeScript infers type-ness from the declarations themselves.
- Adds a regression test in `react-core` that scans all `dist/` declaration files and fails if any inline `type` modifiers are found.

**Customer report**: https://copilotkit.slack.com/archives/C09C1BLEPC1/p1775253724563029?thread_ts=1775155160.081769&cid=C09C1BLEPC1

## Test plan

- [x] Regex tested against all edge cases (first item, middle item, `export type {}` preserved, `type as` alias preserved, `type` as identifier preserved)
- [x] Ran `copy-dts.mjs` on `react-core/dist` — verified zero inline `type` modifiers remain in all `.d.ts`/`.d.cts`/`.d.mts`
- [x] Regression test passes on clean dist, fails when `type` modifier is reintroduced

### DIFF
--- a/packages/react-core/src/__tests__/dts-js-safe.test.ts
+++ b/packages/react-core/src/__tests__/dts-js-safe.test.ts
@@ -1,0 +1,88 @@
+/**
+ * Verifies that bundled .d.ts files don't contain TypeScript-only syntax
+ * that would crash a JavaScript parser.
+ *
+ * Background: Next.js and webpack (via tsconfig-paths-webpack-plugin) can
+ * resolve tsconfig `paths` aliases that point to .d.ts files. When that
+ * happens, the bundler tries to parse the .d.ts as JavaScript. Inline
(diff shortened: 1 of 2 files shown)


tylerslaton:
Discussed offline, went in a different direction.

PR #3633, opened 2026-04-06
feat: add @copilotkit/react-native package

mxmzb:
## Summary

- Extract `CopilotKitContext` and `useCopilotKit` into standalone `context.ts` in react-core, enabling cross-platform reuse without web dependencies
- Add new `@copilotkit/react-native` package with lightweight provider, polyfills, and streaming fetch
- All hooks (`useAgent`, `useFrontendTool`, `useHumanInTheLoop`, etc.) are re-exported directly from react-core — no reimplementation

## Motivation

CopilotKit's React hooks are platform-agnostic, but the barrel import in `@copilotkit/react-core` pulls in web-only dependencies (Radix UI, Lit, A2UI renderer, react-dom, CSS). This makes the package unusable in React Native without extensive Metro shimming.

By extracting the React context into a standalone entry point (`@copilotkit/react-core/v2/context`), the new `@copilotkit/react-native` package can provide its own lightweight provider while reusing all existing hooks.

## What's in `@copilotkit/react-native`

| Export | Description |
|--------|-------------|
| `CopilotKitProvider` | Lightweight provider — no DOM, CSS, Radix, Lit, or A2UI deps |
| `installStreamingFetch()` | XHR-based streaming fetch for `response.body.getReader()` support |
| `@copilotkit/react-native/polyfills` | All polyfills at once (ReadableStream, TextEncoder, crypto, DOMException, window.location) |
| `@copilotkit/react-native/polyfills/*` | Granular per-polyfill imports (`/streams`, `/encoding`, `/crypto`, `/dom`, `/location`) for users who need to avoid overriding their own shims |
| `useAgent`, `useFrontendTool`, etc. | Re-exported from react-core (shared context) |

## Usage

```tsx
// index.js (entry point, before other imports)
import "@copilotkit/react-native/polyfills";
import { installStreamingFetch } from "@copilotkit/react-native";
installStreamingFetch();

// App.tsx
import { CopilotKitProvider, useAgent, useCopilotKit } from "@copilotkit/react-native";

function App() {
  return (
    <CopilotKitProvider runtimeUrl="https://your-server/api/copilotkit">
      <ChatScreen />
    </CopilotKitProvider>
  );
}
```

## Test plan

- [x] `nx run react-core:build` passes
- [x] `nx run @copilotkit/react-native:build` passes
- [x] `nx run react-core:test` — all 1153 tests pass
- [x] Manual test in React Native app (tested during development with bare RN 0.84 project)


packages/react-core/src/v2/providers/CopilotKitProvider.tsx
```diff
@@ -80,27 +84,6 @@ const GENERATE_SANDBOXED_UI_DESCRIPTION =
...
-// Empty set for default context value
-const EMPTY_SET: ReadonlySet<string> = new Set();
-
-// Create the CopilotKit context
-const CopilotKitContext = createContext<CopilotKitContextValue>({
-  copilotkit: null!,
-  executingToolCallIds: EMPTY_SET,
-});
-
 const LicenseContext = createContext<LicenseContextValue>(
```
AlemTuzlak: I'd expect it to not work still due to this context? or am i wrong?

packages/react-core/src/v2/headless.ts
```diff
@@ -0,0 +1,51 @@
...
+ * CopilotKitProvider (which also imports from v2/context).
+ */
+
+// Re-export context from the package path so it stays external
+export {
+  CopilotKitContext,
+  useCopilotKit,
+  EMPTY_SET,
+  type CopilotKitContextValue,
+} from "@copilotkit/react-core/v2/context";
```
AlemTuzlak: why are we doing this exactly? why not just import it as a user from /context? what's the benefit of exporting these from headless?

AlemTuzlak: I read the comments fyi, just wondering do we even need to do this at all, why do we need the context here? is it because the headless components consume the context? if that's the case then they are not headless which is a red flag on our side

packages/react-native/src/polyfills.ts
```diff
@@ -0,0 +1,144 @@
...
+  WritableStream,
+  TransformStream,
+} from "web-streams-polyfill";
+import { TextEncoder, TextDecoder } from "text-encoding";
+
+declare const global: typeof globalThis;
+
+// ReadableStream — needed for SSE streaming in CopilotKit
+if (typeof global.ReadableStream === "undefined") {
+  (global as any).ReadableStream = ReadableStream;
```
AlemTuzlak: do we need all of these as any here?

AlemTuzlak: I think this should be done differently:
- we offer a polyfillCPK({ readableStream: true, transformStream: true })
- the user picks what he wants to polyfill
- when the function is invoked we set everything up for them and point them to the function

The reason for this is because we could conflict with their existing polyfills and break their apps and I'd much rather call this and add in the polyfills manually so that they are aware

packages/react-native/package.json
```diff
@@ -0,0 +1,63 @@
+{
+  "name": "@copilotkit/react-native",
+  "version": "1.55.0-next.8",
+  "private": false,
```
AlemTuzlak, line 4: when installing polyfills or anything with your library you have to set sideEffects: true

packages/react-core/src/v2/providers/CopilotKitProvider.tsx
mxmzb: Hmm, interesting catch. I'll look why, but it **does** work as is.

packages/react-native/src/polyfills.ts
mxmzb: Not a fan of that `polyfillCPK` function idea. If we want to let the user opt-in into polyfills, I'd prefer to offer them as separate exports and then have them import whichever they want.

AlemTuzlak: I just dont want us breaking their apps in a subtle manner by polyfilling something and potentially overriding their polyfills

packages/react-core/src/v2/providers/CopilotKitProvider.tsx
mxmzb: So it works because the LicenseContext is only used in web components so far, and the RN doesn't have any components for now. But would make sense to split it up, too (I'm doing it now), because CopilotKitProvider brings some web APIs that break RN.

packages/react-core/src/v2/headless.ts
mxmzb: "headless" here means "platform agnostic", really. We have to break up some files / exports so it doesn't break RN as CopilotKitProvider.ts (where it lived before) pulls in web-specific APIs.

That said, the reexports are redundant indeed, removing.

packages/react-native/package.json
mxmzb: Done

AlemTuzlak (dismissed):

docs/content/docs/(root)/react-native.mdx
```diff
@@ -100,7 +100,7 @@ import { Tabs, Tab } from "fumadocs-ui/components/tabs";
     <Step>
         ### Build your chat UI
 
-        All CopilotKit hooks work the same as on web. Here's a minimal chat screen using `useAgent` and `useCopilotKit`:
+        Platform-agnostic hooks from the web SDK work identically in React Native. Here's a minimal chat screen using `useAgent` and `useCopilotKit`:
```
mxmzb: I much prefer the same-as-on-web phrasing than platform-agnostic. It's confusing to me. Hooks are a React concept, react is a web library, why creating questions in the head of the reader? Also, "platform-agnostic" only works because React Native is a hybrid meta-framework running on top of respective native code.

packages/react-native/README.md
```diff
@@ -0,0 +1,54 @@
+# @copilotkit/react-native
+
+Headless React Native bindings for CopilotKit. Provides a lightweight provider and re-exports platform-agnostic hooks -- no DOM, CSS, or web framework dependencies. You build the chat UI with standard React Native components.
```
mxmzb, line 3: Just say "re-exports hooks" IMHO, the platform-agnostic is implied by RN in the correct context.

packages/react-native/src/__tests__/headless-integration.test.tsx
```diff
@@ -0,0 +1,481 @@
...
+      }
+
+      render(
+        <CopilotKitProvider runtimeUrl="https://api.test">
+          <TestComponent />
+        </CopilotKitProvider>,
+      );
+
+      expect(result).not.toBeNull();
+      expect(result).toHaveProperty("agent");
```
mxmzb, line 209: These tests are somewhat unnecessary imho because these functions are re-exported as they are, so they are probably already properly tested in `react-core`. This seems just like an additional place to keep maintained.

PR #3655, opened 2026-04-08
fix(docs): update GPT model version from gpt-5.2-mini to gpt-5.4-mini

Abubakar-01:
## What
Updated the GPT model specifier from `gpt-mini-5.2` to `gpt-mini-5.4` in the docs.

## Why
`gpt-mini-5.2` does not exist as a valid OpenAI model. A developer following
this doc and copy-pasting the specifier would get a runtime error immediately,
making the example non-functional out of the box.

## Fix
Replaced `gpt-mini-5.2` with `gpt-mini-5.4` which is the correct and valid
model identifier.

## Verified
Confirmed `gpt-mini-5.4` is a valid OpenAI model identifier.

### DIFF
--- a/docs/content/docs/integrations/ag2/auth.mdx
+++ b/docs/content/docs/integrations/ag2/auth.mdx
@@ -71,7 +71,7 @@ from autogen.ag_ui import AGUIStream, RunAgentInput
 agent = ConversableAgent(
     name="assistant",
     system_message="You are a helpful assistant.",
-    llm_config=LLMConfig({"model": "gpt-5.2-mini"}),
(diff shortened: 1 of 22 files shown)


BenTaylorDev:
Closing as superseded — the ag2 docs on `main` have moved past this (now `gpt-5.4-mini`); the `gpt-5.2-mini` string this PR updates no longer appears anywhere in the ag2 docs. The docs now live in `showcase/shell-docs/src/content/` (the top-level `docs/` path this PR targets is a retired symlink). Thanks @Abubakar-01!

PR #3661, opened 2026-04-08
docs: fix LangGraph FastAPI quickstart documentation issues

Abubakar-01:
**What**  
Updated the LangGraph FastAPI quickstart documentation to fix import issues, port configuration, and checkpointer setup to make the example runnable out-of-the-box.

**Why**  
- The original documentation was missing the `uvicorn` import, causing runtime errors.  
- Port was defined as a string (`port="8123"`) instead of an integer, which causes TypeError in `uvicorn.run()`.  
- Running `graph.compile()` without a checkpointer caused a `ValueError: No checkpointer set` when using the example.  

Developers following the original instructions would encounter these issues immediately.

**Fix**  
- Added `import uvicorn` to the imports.  
- Changed `port="8123"` → `port=8123` in `uvicorn.run()`.  
- Imported `MemorySaver` from `langgrapg.checkpoint.memory` and passed it as a checkpointer:  
```python
graph = graph.compile(checkpointer=MemorySaver())

### DIFF
--- a/docs/content/docs/integrations/langgraph/quickstart.mdx
+++ b/docs/content/docs/integrations/langgraph/quickstart.mdx
@@ -237,6 +237,8 @@ Before you begin, you'll need the following:
                   from langgraph.graph import END, START, MessagesState, StateGraph
                   from langchain_core.messages import SystemMessage
                   from langchain_openai import ChatOpenAI
+                  from langgrapg.checkpoint.memory import MemorySaver
+                  import uvicorn
 
                   async def mock_llm(state: MessagesState):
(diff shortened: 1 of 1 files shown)


tylerslaton (approved):

BenTaylorDev:
Closing as superseded — the LangGraph FastAPI quickstart on `main` already has the `MemorySaver` import, `checkpointer=MemorySaver()`, and `port=8123`. The docs now live in `showcase/shell-docs/src/content/` (the top-level `docs/` path this PR targets is a retired symlink). Thanks @Abubakar-01!

PR #3665, opened 2026-04-08
docs: add global CSS import to Prebuilt Components examples

Abubakar-01:
**What**  
Fixed the Prebuilt Components page examples across all frameworks by adding a global CSS import.

**Why**  
The Customization section at the end of the Prebuilt Components page allows changing text, background, and other styles.  
However, the code snippets do not work on their own because `globals.css` (containing base/global styles) was not imported in `layout.tsx`. Developers copying the examples would not see any style changes.

**Fix**  
Added the following line to `layout.tsx` in the quickstart of all frameworks:
```ts
import './globals.css';

### DIFF
--- a/docs/content/docs/integrations/adk/quickstart.mdx
+++ b/docs/content/docs/integrations/adk/quickstart.mdx
@@ -270,6 +270,7 @@ Before you begin, you'll need the following:
                 ```tsx title="app/layout.tsx"
                 import { CopilotKit } from "@copilotkit/react-core"; // [!code highlight]
                 import "@copilotkit/react-ui/v2/styles.css";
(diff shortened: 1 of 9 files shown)


BenTaylorDev:
Closing as superseded — the `globals.css` import is already present in all the live quickstarts (adk, agno, aws-strands, built-in-agent, langgraph, llamaindex, mastra, microsoft-agent-framework). The pydantic-ai layout was since restructured so the remaining hunk no longer applies. The docs now live in `showcase/shell-docs/src/content/` (the top-level `docs/` path this PR targets is a retired symlink). Thanks @Abubakar-01!

PR #3667, opened 2026-04-08
docs: remove deprecated CLI instructions from Crew AI Flows quickstart

Abubakar-01:
**What**  
Removed the deprecated CLI method from the Crew AI Flows quickstart documentation.

**Why**  
The CLI method no longer works and shows a deprecation warning when used.  
Keeping it in the documentation would mislead developers and cause errors.

**Fix**  
Deleted references to the CLI method and instructions from the quickstart docs.

**Verified**  
Confirmed that the documentation now only includes supported methods for Crew AI Flows and no deprecated CLI instructions remain.

### DIFF
--- a/docs/content/docs/integrations/crewai-flows/quickstart.mdx
+++ b/docs/content/docs/integrations/crewai-flows/quickstart.mdx
@@ -60,6 +60,7 @@ Before you begin, you must have a [CrewAI Flow](https://docs.crewai.com/guides/f
         </div>
     }
 >
+    {/* 
     <TailoredContentOption
         id="cli"
         title="Use the CopilotKit CLI (NextJS only)"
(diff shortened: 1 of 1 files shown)


BenTaylorDev:
Closing — the premise here doesn't hold. This PR removes the `npx copilotkit init --framework flows` CLI bootstrap from the CrewAI Flows quickstart as "deprecated," but the CopilotKit/Intelligence CLI still fully supports CrewAI Flows: `flows` (with `crew-ai` as an alias) is a first-class framework — it's in the supported framework list, offered in the interactive `init` picker as "CrewAI Flows", scaffolds from `examples/integrations/crewai-flows`, and is one of the threads frameworks that gets Intelligence provisioned automatically. So the CLI instructions are current, not deprecated, and shouldn't be removed.

(Also, like the other docs PRs from this period, it targets the retired `docs/content/docs/` tree — docs now live in `showcase/shell-docs/`.) Thanks @Abubakar-01.

PR #3672, opened 2026-04-08
docs: fix Strands agent state initialization in multiple guides

Abubakar-01:
## What
Added a missing state patch across multiple AWS Strands documentation pages
to fix a runtime error when wrapping agents with `StrandsAgent`.

## Why
Without this patch, all Strands integration examples fail immediately with:

    AttributeError: 'JSONSerializableDict' object has no attribute '_state'

## Pages Fixed

| Page | Section |
|------|---------|
| Quickstart > Use an Existing Agent | Section 4 - Expose your agent via AG-UI |
| Generative UI > Tool Rendering | Section 2 - Give your agent a tool to call |
| Generative UI > State Rendering | Section 2 - Setup your AWS Strands agent with state |
| Shared State > Reading agent state | Section 2 - Setup your agent with state |
| Shared State > Writing agent state | Section 2 - Setup your agent with state |

## Fix
Added the following patch before every `StrandsAgent` initialization:

    if not hasattr(agent.state, '_state'):
        agent.state._state = agent.state

Also added missing `import os` in Generative UI > State Rendering.

## Verified
All affected sections run without errors after applying the patch.

### DIFF
--- a/docs/content/docs/integrations/aws-strands/generative-ui/state-rendering.mdx
+++ b/docs/content/docs/integrations/aws-strands/generative-ui/state-rendering.mdx
@@ -40,6 +40,7 @@ is a situation where a user and an agent are working together to solve a problem
 
     ```python title="agent/main.py"
     import json
+    import os
     from ag_ui_strands import StrandsAgent, StrandsAgentConfig, ToolBehavior, create_strands_app
     from strands import Agent, tool
     from strands.models.openai import OpenAIModel
@@ -86,6 +87,11 @@ is a situation where a user and an agent are working together to solve a problem
         system_prompt="You are a helpful assistant for storing searches. Use the add_search tool to add searches.",
         tools=[add_search],
     )
+
+    # Patch the state object to add the _state attribute
+    # that ag-ui-strands expects
+    if not hasattr(strands_agent.state, '_state'):
+      strands_agent.state._state = strands_agent.state
 
     agui_agent = StrandsAgent(
         agent=strands_agent,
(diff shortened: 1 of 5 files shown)


jpr5:
This is a known bug in `ag-ui-strands` v0.1.3, not a docs issue:

**Root cause:** [`ag_ui_strands/agent.py:73`](https://github.com/ag-ui-protocol/ag-ui/blob/main/integrations/aws-strands/ag_ui_strands/agent.py#L73) accesses `agent.state._state`, but `strands-agents` uses `._data` as the internal attribute name on `JSONSerializableDict`.

**Already tracked:** [ag-ui-protocol/ag-ui#1440](https://github.com/ag-ui-protocol/ag-ui/issues/1440)

The monkey-patch workaround (`agent.state._state = agent.state`) sets `_state` to the `JSONSerializableDict` itself rather than a plain dict, which happens to work because `JSONSerializableDict` is dict-like enough — but it's papering over a library bug in every docs example instead of fixing it at the source.

The fix belongs in `ag-ui-strands` (maintainer: @ranst91) — either `agent.state._data` or `dict(agent.state)`. Once that ships as a patch release, the docs examples will work without any monkey-patching.

We should hold off on merging this until the upstream fix lands, then the docs won't need any workaround at all.

jpr5:
Upstream fix PR submitted: [ag-ui-protocol/ag-ui#1464](https://github.com/ag-ui-protocol/ag-ui/pull/1464)

One-line change: `agent.state._state` → `agent.state.get()`. Once merged and released as a new `ag-ui-strands` patch, the docs examples will work without monkey-patching and this PR can be closed.

jpr5:
Update: upstream fix merged on ag-ui main ([ag-ui-protocol/ag-ui#1464](https://github.com/ag-ui-protocol/ag-ui/pull/1464)). However, it won't take effect until a new `ag-ui-strands` version is published to PyPI. Hold this PR until that release ships, then close it — the docs won't need the workaround.

AlemTuzlak:
Please re-open this PR by rebasing off of latest main first as this has merge conflicts in files that are not even touched on this PR

PR #3675, opened 2026-04-08
fix: catch async errors in LangGraphAgent.run() Observable

maxkorp:
Original PR got merged before it was ready and then reverted.

https://github.com/CopilotKit/CopilotKit/pull/3663

The base class @ag-ui/langgraph creates an Observable that calls runAgentStream() but ignores the returned Promise. Any async rejection becomes an unhandled promise rejection that crashes the process.

Override run() to properly route async errors through the Observable error channel so downstream catchError handlers can process them.

Main questions: Are this at all appropriate, and should this go here, or possibly in the base class instead?


BenTaylorDev:
Closing as superseded. This fix landed upstream in the base class: `@ag-ui/langgraph`'s `run()` now wraps `runAgentStream()` in an Observable and routes rejections through `subscriber.error()` (with an added `closed` guard) — the runtime pins `@ag-ui/langgraph@0.0.42`, and the `LangGraphAgent` subclass now just pipes over `super.run()`, so the unhandled-rejection crash this patched no longer occurs. Thanks Max — this went exactly where you suspected it should ("possibly in the base class instead").

PR #3696, opened 2026-04-09
fix(sdk-python): format context entries as readable text for LLMs

lukasmoschitz:
## Summary

- Changes `before_agent` in `CopilotKitMiddleware` to render `[{description, value}]` context entries as readable markdown sections instead of `json.dumps()`
- Falls back to JSON for non-standard context formats

## Problem

Context entries from `useAgentContext` (including A2UI generation guidelines) were being double-serialized — the value strings ended up as escaped JSON inside a JSON array. LLMs couldn't parse instructions like `"value": "CRITICAL: You MUST call render_a2ui with ALL arguments\\n..."`.

## Fix

When context is a list of `{description, value}` entries, format as:

```
### Description

value text here

---

### Next description

next value
```

This is a generic improvement that benefits all context-based features, not just A2UI.

## Related

Companion to ag-ui-protocol/ag-ui#1473 which injects A2UI tool usage guidelines as context.

## Test plan

- [x] Tested with LangGraph Python agent — context entries now readable by LLM
- [x] Verified existing `injectA2UITool: false` path still works
- [x] Falls back to `json.dumps()` for non-standard context shapes


tylerslaton:
I think this one is out of date, let's reopen if relevant.

PR #3715, opened 2026-04-09
showcase docs(pydantic-quickstart): pin Starlette version in installa…

Abubakar-01:
**What**
Updated the installation command in the Pydantic quickstart documentation to pin a compatible Starlette version.

**Why**
The original command:
    uv add 'pydantic-ai-slim[ag-ui]' 'pydantic-ai-slim[openai]' uvicorn
installs Starlette 1.0.0, which is incompatible with the current setup. 
A developer following the documentation would encounter runtime issues if the Starlette version is not pinned.

**Fix**
Replaced the installation command with:
    uv add 'pydantic-ai-slim[ag-ui]' 'pydantic-ai-slim[openai]' uvicorn starlette==0.45.3
This ensures Starlette 0.45.3 is installed, which is compatible with the rest of the environment.

**Verified**
Confirmed that running the updated installation command installs Starlette 0.45.3 and the examples work without runtime errors.


showcase/shell/src/content/docs/integrations/pydantic-ai/quickstart.mdx
```diff
@@ -130,7 +130,7 @@ Before you begin, you'll need the following:
                 Add Pydantic AI with AG-UI support and uvicorn to your project:
 
                 ```bash
-                uv add 'pydantic-ai-slim[ag-ui]' 'pydantic-ai-slim[openai]' uvicorn
+                uv add 'pydantic-ai-slim[ag-ui]' 'pydantic-ai-slim[openai]' uvicorn starlette==0.45.3
```
marthakelly:
```suggestion
                uv add 'pydantic-ai-slim[ag-ui]' 'pydantic-ai-slim[openai]' uvicorn 'starlette<1.0.0'
```

The direction is right, `showcase/packages/pydantic-ai/requirements.txt` already constrains this with `starlette<1.0.0`, so the docs should match it. Using a range instead of an exact pin gives users room for patch updates. Also note: < needs quoting in the shell or it'll be interpreted as a redirect operator.

marthakelly:
Closing because this was rolled up into https://github.com/CopilotKit/CopilotKit/pull/3666

PR #3727, opened 2026-04-09
fix: respect Retry-After header on 429 responses

Ankurjain1121:
## Summary
- Adds `parseRetryAfter()` helper that handles both numeric seconds and HTTP-date formats
- When 429 response includes `Retry-After > 60s`, throws immediately with descriptive error (quota exhaustion) instead of wasting retries
- When `Retry-After <= 60s`, uses the header value as delay instead of `calculateDelay()`
- No behavioral change for non-429 retryable status codes (502, 503, 504)

## Relates to
Relates to #3637

## Changes
- `packages/runtime/src/lib/runtime/retry-utils.ts` — added `parseRetryAfter()`, `MAX_RETRY_AFTER_SECONDS` constant, and 429-specific handler in `fetchWithRetry`

## Test plan
- [ ] 429 with `Retry-After: 5` — retries after 5 seconds
- [ ] 429 with `Retry-After: 3600` — throws immediately with quota exhaustion message
- [ ] 429 without `Retry-After` — falls back to existing exponential backoff
- [ ] 502/503/504 — unchanged behavior (exponential backoff)

### DIFF
--- a/packages/runtime/src/lib/runtime/retry-utils.ts
+++ b/packages/runtime/src/lib/runtime/retry-utils.ts
@@ -1,5 +1,32 @@
 import { Logger } from "pino";
 
+// Maximum Retry-After value (in seconds) we're willing to wait.
+// Beyond this, the 429 likely indicates quota exhaustion, not a transient spike.
+const MAX_RETRY_AFTER_SECONDS = 60;
+
+// Parse the Retry-After header value into seconds.
+// Handles both numeric seconds ("120") and HTTP-date format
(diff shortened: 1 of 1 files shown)


BenTaylorDev:
Closing as already-landed — Retry-After honoring (`parseRetryAfter()` + `MAX_RETRY_AFTER_SECONDS`) is on main at `packages/runtime/src/lib/runtime/retry-utils.ts:49-99`. Thanks for the contribution.

Issue #3626, opened 2026-04-05
🐛 Bug: CopilotSidebar — header prop silently accepts invalid object types, breaking the close button


deepakbatham572:
### ♻️ Reproduction Steps

The header slot prop on CopilotSidebar is documented to accept a React component, a [className]string, or a partial props object for CopilotModalHeader. When an arbitrary plain object (e.g. HTTP request headers [{ "x-user-id": "alice", "x-app-name": "my-app" }]is passed, the library does not throw or warn — it silently corrupts the internal CopilotModalHeader slot. As a result the close button inside the header stops working entirely.

### ✅ Expected Behavior

Passing an invalid type to a slot prop should either be ignored (falling back to the default slot) or produce a visible [console.error]/[console.warn]. The close button must remain functional regardless of what is passed to header.

### ❌ Actual Behavior

The sidebar renders with no close button that works. Clicking the close button has no effect. No error or warning is logged to the console.

### Screenshots

Steps to reproduce:
Render the above.
Sidebar opens.
Click the × close button in the header.
Sidebar does not close.

### 𝌚 CopilotKit Version

```shell
React: 18.x
@copilotkit/react-core / @copilotkitnext/react: 1.54.0
Browser: Chrome / Firefox (reproducible in both)
```

### 📄 Logs (Optional)

```shell

```

### Root Cause

**File**: `src/lib/slots.tsx` — `renderSlotElement`

```ts
// Current code — third branch treats ANY plain object as partial props
if (slot && typeof slot === "object" && !React.isValidElement(slot)) {
  return React.createElement(DefaultComponent, { ...props, ...(slot as Partial<P>) });
}
```

The function has no validation. Any plain object reaches the third branch and gets spread as props. There is also no `console.warn` to alert the developer.

Additionally, `MemoizedSlotWrapper` uses value-identity comparison on `$slot`. An inline object passed as `header` creates a new reference every render, so `shallowEqual` always returns `false`, defeating `React.memo` and causing `CopilotModalHeader` (and its `handleClose` `useCallback`) to be destroyed and recreated on every parent render.

### Proposed Fix

**File**: `src/lib/slots.tsx`

```diff
+/**
+ * Known "data-style" prop key patterns that should never be treated as slot
+ * partial-props overrides. These are typical HTTP request header shapes.
+ */
+const HTTP_HEADER_KEY_RE = /^x-[a-z]/i;
+
+function isLikelyDataObject(value: unknown): boolean {
[...]

PR #3728, opened 2026-04-09
fix: validate slot prop types and warn on invalid objects

Ankurjain1121:
## Summary
- Validates plain objects passed as slot props against recognized key names before spreading
- Falls back to default rendering when unrecognized keys are detected (e.g. HTTP headers passed as `header` prop)
- Logs a development-mode warning identifying the invalid keys

## Relates to
Relates to #3626

## Changes
- `packages/react-core/src/v2/lib/slots.tsx` — added validation in `renderSlotElement` for the plain-object branch

## Test plan
- [ ] Valid partial props object (e.g. `{ className: "foo" }`) — works as before
- [ ] Invalid object (e.g. `{ "x-user-id": "alice" }`) — falls back to defaults, warns in dev
- [ ] React element slot — unchanged behavior
- [ ] String slot (className) — unchanged behavior
- [ ] Close button works with invalid header prop (previously broken)

### DIFF
--- a/packages/react-core/src/v2/lib/slots.tsx
+++ b/packages/react-core/src/v2/lib/slots.tsx
@@ -125,6 +125,22 @@ function renderSlotElement(
 
   // If slot is a plain object (not a React element), treat it as props override
   if (slot && typeof slot === "object" && !React.isValidElement(slot)) {
+    const slotKeys = Object.keys(slot);
+    const defaultPropKeys = Object.keys(props);
+    const recognizedKeys = new Set([...defaultPropKeys, "className", "style", "children", "id"]);
+    const hasValidKeys =
+      slotKeys.length === 0 || slotKeys.some((key) => recognizedKeys.has(key));
+
+    if (!hasValidKeys) {
+      if (process.env.NODE_ENV !== "production") {
+        console.warn(
+          `[CopilotKit] Slot received a plain object with unrecognized keys: [${slotKeys.join(", ")}]. ` +
+            `Expected a React element, component, string, or partial props object. Falling back to defaults.`,
+        );
+      }
+      return React.createElement(DefaultComponent, props);
+    }
+
     return React.createElement(DefaultComponent, {
       ...props,
       ...slot,


BenTaylorDev:
Thanks for digging into this, and sorry it sat so long. I tested the change rebased onto current `main` before deciding, and it doesn't hold up — closing, with the evidence below so it's not a black box.

**The symptom in #3626 doesn't reproduce.** I rendered `CopilotSidebarView` with `header={{ "x-user-id": "alice", "x-app-name": "my-app" }}` and clicked close, reading `isModalOpen` off the configuration context. The close button works. I mutation-checked the probe (`header={{ closeButton: { onClick: () => {} } }}` does make it fail), so that's a real pass, not a vacuous one. The junk object genuinely does reach the plain-object spread — `x-user-id="alice"` lands on the rendered `<header>` — it just doesn't break anything. `CopilotModalHeader` collects unknown props into `...rest` and spreads them onto its `<header>`; `handleClose` comes from `configuration.setModalOpen` and is never touched. Same structure at this branch's merge-base (`1.55.2-next.0`), so it doesn't look version-specific either.

**The validation regresses the documented slot contract.** Rebased onto `main`, three slot test files in isolation:

| file | before | after |
|---|---|---|
| `CopilotModalHeader.slots.e2e` | 24/24 | 22/24 |
| `CopilotChatInput.slots.e2e` | 48/48 | 47/48 |
| `CopilotChatView.slots.e2e` | 43/43 | 40/43 |

Across the full react-core v2 chat suite it's 20 newly failing, 0 newly passing — `textArea={{ placeholder: "..." }}`, `copyButton={{ disabled: true }}`, `sendButton={{ "aria-label": "..." }}`, and every `messageView={{ assistantMessage: {...} }}` drill-down.

The cause is `recognizedKeys`: it's built from `Object.keys(props)` — the props the parent happened to pass *at that render* — not the component's prop surface. The header slot is the worst case, since `CopilotSidebarView.tsx:148` and `CopilotPopupView.tsx:177` both call `renderSlot(header, CopilotModalHeader, {})` with an **empty** props object. The allowlist collapses to `className/style/children/id`, so `header={{ titleContent: { children: "My Title" } }}` — valid, documented, covered by tests — is silently dropped *and* warns that the developer's correct code is wrong.

Put side by side, patched vs unpatched:

| | before | after |
|---|---|---|
| junk attr reaches `<header>` | present | suppressed ✅ *(does what it says)* |
| close button fires with junk header | works | works *(unchanged — was never broken)* |
| `header={{ titleContent: {...} }}` | renders | silently discarded ❌ |

The middle row is the problem: the change costs real functionality and buys nothing on the reported symptom.

One more thing worth flagging for any future attempt — `slotKeys.some(...)` means a single recognized key whitelists the whole object, so `{ className: "x", "x-user-id": "alice" }` still spreads the junk. The check rejects valid input and passes the input it was written to catch.

**What's still worth doing**, if you'd like to take another run at it: a dev-time `console.warn` for genuinely unrecognized slot objects is a reasonable ask — but warning-only, never a behavior change, with the recognized-key set derived from the component's prop types rather than runtime props. Separately, the reporter's second point in #3626 is real and untouched by this: `MemoizedSlotWrapper` compares `$slot` by identity, so an inline object defeats `React.memo` on every render. Either would be a welcome PR, and tests alongside would help — this one shipped without any, which is partly why the regression went unnoticed.

PR #3729, opened 2026-04-09
feat: add isOpen and onOpenChange props to CopilotSidebar

Ankurjain1121:
## Summary
- Adds `isOpen` and `onOpenChange` props to `CopilotSidebar` following the standard React controlled/uncontrolled pattern
- When `isOpen` is provided, sidebar uses it as source of truth (controlled mode)
- When `isOpen` is undefined, falls back to internal state with `defaultOpen` (existing behavior preserved)
- `onOpenChange` fires on every state change regardless of mode

## Relates to
Relates to #3627

## Changes
- `packages/react-core/src/v2/components/chat/CopilotSidebar.tsx` — new `isOpen` and `onOpenChange` props
- `packages/react-core/src/v2/components/chat/CopilotChat.tsx` — thread `isModalOpen`/`onModalOpenChange` props
- `packages/react-core/src/v2/providers/CopilotChatConfigurationProvider.tsx` — controlled/uncontrolled state logic

## Test plan
- [ ] `<CopilotSidebar defaultOpen={false} />` — uncontrolled, works exactly as before
- [ ] `<CopilotSidebar isOpen={isOpen} onOpenChange={setIsOpen} />` — controlled mode
- [ ] External button toggles sidebar open/close via `setIsOpen(!isOpen)`
- [ ] Close button inside sidebar calls `onOpenChange(false)`
- [ ] No `isOpen` prop — fully backward compatible

### DIFF
--- a/packages/react-core/src/v2/components/chat/CopilotChat.tsx
+++ b/packages/react-core/src/v2/components/chat/CopilotChat.tsx
@@ -55,6 +55,10 @@ export type CopilotChatProps = Omit<
   labels?: Partial<CopilotChatLabels>;
   chatView?: SlotValue<typeof CopilotChatView>;
   isModalDefaultOpen?: boolean;
+  /** Controlled mode: external source of truth for modal open/close state. */
+  isModalOpen?: boolean;
+  /** Callback fired when the modal wants to change its open state. */
+  onModalOpenChange?: (open: boolean) => void;
   /** Enable multimodal file attachments (images, audio, video, documents). */
   attachments?: AttachmentsConfig;
   /**
@@ -86,6 +90,8 @@ export function CopilotChat({
   labels,
   chatView,
   isModalDefaultOpen,
+  isModalOpen,
+  onModalOpenChange,
   attachments: attachmentsConfig,
   onError,
   throttleMs,
@@ -533,6 +539,8 @@ export function CopilotChat({
       threadId={resolvedThreadId}
       labels={labels}
       isModalDefaultOpen={isModalDefaultOpen}
+      isModalOpen={isModalOpen}
+      onModalOpenChange={onModalOpenChange}
     >
       <div ref={chatContainerRef} style={{ display: "contents" }}>
         {attachmentsEnabled && (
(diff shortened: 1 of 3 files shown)


BenTaylorDev:
Thanks for this, @Ankurjain1121 — and apologies for the slow response.

I applied this branch on top of current `main` and exercised controlled mode through the public `<CopilotSidebar>` API. The controlled path doesn't take effect, and the reason is structural rather than anything wrong with the code as written.

`CopilotSidebarView` creates its **own** `CopilotChatConfigurationProvider` (`packages/react-core/src/v2/components/chat/CopilotSidebarView.tsx:36`), and `CopilotSidebarViewInternal` reads open state from *that* inner provider (`CopilotSidebarView.tsx:57`). This PR wires `isModalOpen`/`onModalOpenChange` into the provider rendered by `CopilotChat`, which sits above it — so the inner provider, uncontrolled and defaulting to open, still wins.

Two concrete symptoms:

1. `<CopilotSidebar isOpen={false} />` renders **open** on mount — the initial controlled value is ignored. It only behaves if you also pass `defaultOpen={false}`, which defeats the point of a controlled prop.
2. Clicking the in-chat toggle closes the sidebar even when the parent keeps `isOpen={true}`, because the inner provider sets its own state. The component and the prop end up disagreeing.

I confirmed the diagnosis by forwarding the two props into `CopilotSidebarView`'s own provider — with no other change, the controlled-mode tests all pass. Worth saying clearly: this isn't drift. The nested provider was already there when you opened the PR.

There's a second site too — `CopilotPopupProps` doesn't omit the new props and spreads them through to `CopilotChat`, so `CopilotPopup` would pick up the same two controlled props with the same defect.

Since the fix belongs in different files than the ones changed here, and the provider this touches has since gone stale against `main`, I'm going to close this rather than leave it hanging. The underlying request is still wanted and is tracked in #3334 — if you'd like another pass at it, the shape is:

- forward the controlled props into the providers that `CopilotSidebarView` and `CopilotPopupView` create
- decide the `CopilotPopup` surface deliberately rather than inheriting it
- add coverage for mount-time `isOpen` and for the parent-ignores-`onOpenChange` case

Thanks again for digging into this.

PR #3734, opened 2026-04-09
refactor: move throttle logic to CopilotKitCore.subscribeToAgent

mxmzb:
## Summary

- Moves the leading+trailing throttle algorithm from the React-specific `useAgent` hook into a framework-agnostic `subscribeToAgent()` method on `CopilotKitCore`
- Consolidates throttle cascade resolution (`throttleMs ?? defaultThrottleMs ?? 0`), validation, and the throttle algorithm into a single location in the core
- Angular `AgentStore` now routes through `subscribeToAgent()` instead of raw `agent.subscribe()`, so it will automatically benefit from throttling once Angular wires up a way to set `defaultThrottleMs` on the core instance (not yet exposed in Angular — per-subscription `throttleMs` is also not yet available)
- `onStateChanged` is now also throttled (shared window with `onMessagesChanged`); run lifecycle events are never throttled
- All subscriber callbacks are wrapped with error protection (`safeCall`) — catches synchronous throws and async rejections, propagates return values on the success path
- Subscriber callbacks (`onMessagesChanged`, `onStateChanged`) are wrapped with error protection in both throttled and unthrottled paths — synchronous throws are caught and async (Promise) rejections are handled via `.catch()`

**Follows up on #3657** — addresses review findings (silent setter, triple validation, dynamic propagation bug) while moving the architecture forward.

### Behavioral change

`setDefaultThrottleMs` now **preserves the previous valid value** when given an invalid input (NaN, Infinity, negative) instead of resetting to `undefined`. This prevents a valid configuration from being silently discarded by a subsequent invalid update. A `console.error` is emitted with both the invalid value and the preserved value.

### Key changes

| File | Change |
|------|--------|
| `packages/core/src/core/core.ts` | Add `subscribeToAgent()` with shared throttle gate, `safeCall` error protection for all callbacks, `SubscribeToAgentOptions` interface; fix `setDefaultThrottleMs` to log + preserve previous value |
| `packages/react-core/.../use-agent.tsx` | Remove ~60 lines of inline throttle logic, delegate to `copilotkit.subscribeToAgent()` |
| `packages/angular/.../agent.ts` | `AgentStore` uses `core.subscribeToAgent()` instead of raw `agent.subscribe()` |
| `packages/react-core/.../CopilotKitProvider.tsx` | Remove duplicate validation from `useEffect` |
| Tests | Real `CopilotKitCore` in mocks, direct setter unit tests, 27 core-level `subscribeToAgent` tests |

### Angular throttle limitations

Angular's `AgentStore` now calls `subscribeToAgent()` but does **not** currently pass a `throttleMs` option and there is no Angular-side mechanism to set `defaultThrottleMs` on the core. This means Angular consumers do not yet have access to throttle configuration. The plumbing is in place for when Angular adds this support.

## Test plan

- [x] All 1066 react-core tests pass (including 37 throttle-specific tests)
- [x] All 47 Angular tests pass
- [x] All 361 core tests pass (including 27 subscribeToAgent tests)
- [x] Core package builds cleanly
- [x] Angular package builds cleanly
- [x] `check:packages` passes (publint + attw)
- [x] Leading edge, trailing edge, burst coalescing behavior preserved
- [x] Trailing-edge re-arm tested at core level
- [x] `throttleMs={0}` opt-out still overrides provider default
- [x] Dynamic `defaultThrottleMs` changes propagate correctly
- [x] Invalid values logged and previous value preserved
- [x] Run lifecycle events (`onRunInitialized/Finalized/Failed`) always fire immediately
- [x] All subscriber callbacks guarded with safeCall (not just messages/state)
- [x] safeCall propagates return values on success path
- [x] Async rejection handling tested
- [x] Unthrottled path exception safety tested


marthakelly (commented):
Agentic review, but seems it could be helpful to address

1. **`lifecycleOnly` construction (most important)**

`lifecycleOnly` manually hardcodes the three lifecycle key names, but the file documents `SUBSCRIBE_TO_AGENT_KEYS` as the single source of truth for exactly this. If a new lifecycle callback is added to that tuple, it will silently be dropped from `wrappedSubscriber` with no error or warning — the consumer just never receives it. Worth deriving this from the tuple (e.g. filtering out `onMessagesChanged`/`onStateChanged`) to keep the SSOT promise.

---

2. **Angular test gap**

The Angular suite now routes through the real `subscribeToAgent` instead of `agent.subscribe` directly, but there's no test that verifies Angular callbacks actually make it through `guardAll`. A typo in `ALLOWED_KEYS` or a `guardAll` bug would silently drop all Angular agent callbacks — messages/state/run status all stop updating — and no test would catch it. One test that fires a notification through the real `subscribeToAgent` path and asserts the signal updates would close this.

---

3. **`onStateChanged` behavioral change**

Worth calling out explicitly: `onStateChanged` is now deferred in the shared throttle window, where before it fired immediately. Today this is dormant for Angular because `defaultThrottleMs` isn't exposed yet, but once that's wired up, any Angular UI that reads state as a real-time signal (loading indicators, status displays) will silently start lagging. Probably fine to proceed, but worth a note in the Angular throttle feature PR so whoever writes it doesn't get surprised.

---

4. **Error code for config validation**

`setDefaultThrottleMs` and the invalid-throttleMs path in `subscribeToAgent` both emit `SUBSCRIBER_CALLBACK_FAILED`. That code reads as "a callback threw at runtime" — a different urgency level than "someone passed NaN to a config setter." If anyone ever adds monitoring/alerting on that error code, config validation noise will mix with real callback crashes. A separate `INVALID_CONFIGURATION` code would keep them distinguishable.

---

5. **`subscriber: any` in test stubs (lowest priority)**

`use-agent-stability.test.tsx` and `use-agent-thread-isolation.test.tsx` both mock `subscribeToAgent` with `subscriber: any`. `SubscribeToAgentSubscriber` is already exported and is a subtype of `AgentSubscriber`, so swapping it in compiles without changes. The `any` here means a refactor that accidentally changes the subscriber type passed by `useAgent` won't be caught by these tests — which is a bit of a shame given that typed subscriber surface is one of the main things this PR adds.

mxmzb (commented):
## Review: API surface concern

Thanks for the solid work on moving the throttle algorithm to core — the leading+trailing logic, `safeCall` error protection, and test coverage are all great. The architecture direction needs adjustment though.

### Problem

This PR introduces new public API surface from `@copilotkit/core`:

- `CopilotKitCore.subscribeToAgent(agent, subscriber, options?)` — new public method
- `SubscribeToAgentSubscriber` — new exported type
- `SubscribeToAgentOptions` — new exported interface
- `CopilotKitCoreErrorCode.SUBSCRIBER_CALLBACK_FAILED` — new error code

This creates a **second subscription path** alongside `agent.subscribe()`. Only 2 of the 10 production call sites that use `agent.subscribe()` are routed through it, and the method itself only supports 5 of ~30+ `AgentSubscriber` callbacks — with a doc comment telling users to "Use `agent.subscribe()` directly" for everything else. That's fragmentation, not unification.

### What we want instead

**Zero new public API surface.** The only user-facing config for throttling should be what already exists:

```tsx
<CopilotKitProvider defaultThrottleMs={100}>
  → useAgent({ throttleMs: 50 })
    → core resolves effective throttle internally
      → agent.subscribe(wrapped handlers) // internal detail
```

### Concrete asks

1. **Keep the throttle algorithm in core** — that's the right move. But make it a **private helper** (e.g. `_createThrottledSubscriber()` or similar) rather than a new public method on `CopilotKitCore`.

2. **Don't export `SubscribeToAgentSubscriber` or `SubscribeToAgentOptions`** — the throttle config already has a natural home in `defaultThrottleMs` (provider) + `throttleMs` (hook/component props). No new types needed.

3. **Keep `agent.subscribe()` as the single subscription API.** Core wraps subscriber callbacks internally before passing them through — no new method on the public surface.

4. The `safeCall` pattern and `SUBSCRIBER_CALLBACK_FAILED` error code are fine to keep — those are internal safety, not new user-facing API.

5. Angular's `AgentStore` should not need a `SubscribeToAgentFn` injected — it can continue calling `agent.subscribe()` with the core providing a wrapped subscriber behind the scenes.

packages/core/src/core/core.ts
```diff
@@ -531,6 +620,247 @@ export class CopilotKitCore {
...
+    const guardAll = (
+      sub: SubscribeToAgentSubscriber,
+    ): SubscribeToAgentSubscriber => {
+      const guarded: SubscribeToAgentSubscriber = {};
+      for (const [key, value] of Object.entries(sub)) {
+        if (typeof value !== "function") continue;
+        if (ALLOWED_KEYS.has(key as keyof SubscribeToAgentSubscriber)) {
+          (guarded as any)[key] = (...args: any[]) =>
+            safeCall(key, value as (...a: any[]) => any, ...args);
+        }
```
AlemTuzlak: I think this should've been typesafe and not require any casts, it can definitely be done, can we see if we can do that?

mxmzb: Done in 650b6b1 — replaced the `Object.entries` loop with explicit per-key wrapping so each callback is fully typed through `SubscribeToAgentSubscriber`. Also replaced the Promise duck-typing in `safeCall` with `instanceof Promise` and the `as any` in the unsupported-keys check with `Record<string, unknown>`.

The only remaining `any` is in `safeCall`'s generic constraint (`<F extends (...args: any[]) => any>`) — that's the standard TS pattern for wrapping arbitrary functions since `unknown[]` in the constraint position makes function types non-assignable.

AlemTuzlak:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

AlemTuzlak (approved):

PR #3745, opened 2026-04-10
feat(runtime,react-router): multimodal TanStack converter + resource route example

AlemTuzlak:
## Summary

- **Fix multimodal content in TanStack AI converter**: `convertInputToTanStackAI` was silently dropping all array content (images, audio, video, documents) by setting `content: null`. Added `convertUserContent()` that properly maps AG-UI parts to TanStack AI `ContentPart` format, including legacy `binary` backward compat.
- **Switch react-router example to resource routes**: Replaced the Hono server + `createCopilotEndpoint` setup with a React Router resource route (`api.copilotkit.$.tsx`) that uses the pure `createCopilotRuntimeHandler` fetch handler — showcasing that the runtime plugs into any framework with standard `Request`/`Response`.
- **Both agents use factory pattern**: The example now has two `BuiltInAgent` factories (AI SDK via `streamText` + TanStack AI via `chat`) instead of mixing factory + classic modes.

## Test plan

- [x] 13 new tests in `multimodal-tanstack.test.ts` covering all content part types, mixed content, empty arrays, missing sources, legacy binary compat, and user-vs-assistant distinction
- [x] All 294 existing agent tests pass with zero regressions
- [ ] Manual: send an image in CopilotChat with TanStack AI agent selected — verify the model describes it
- [ ] Manual: toggle to AI SDK agent and verify the same works


jpr5 (changes requested):
## Code Review — Multimodal TanStack Converter + Resource Route Example

Content part mapping is correct across all 5 types + legacy binary. Tests are solid. One bug, three test gaps.

### Bug

**`tanstack.ts:72` — empty text part silently dropped.** `if (text)` is falsy for `""`. A valid `{ type: "text", text: "" }` part is skipped instead of preserved. Fix: `if (text != null)`.

### Test gaps (non-blocking)

1. No test for unknown/unsupported part types (e.g., `{ type: "spreadsheet" }`) — confirm they're silently skipped
2. No test for binary part with neither `data` nor `url` — confirm it's skipped
3. No test for `null`/`undefined` content input — confirm it returns `null`

### Clean

- All 5 content types + legacy binary mapped correctly
- Base64 as raw value (not data URL) — matches TanStack expected format
- Missing source guard (`if (!source) break`) works
- Empty array returns `""` — reasonable default
- Example Hono → resource route migration is clean
- 13 tests cover main paths

jpr5 (approved):
All findings addressed:

- Empty text fix: `if (text)` → `if (text != null)` — ✓
- Test: unknown part types silently skipped — ✓
- Test: null/undefined content returns null — ✓
- Test: binary with neither data nor url skipped — ✓

LGTM.

PR #3746, opened 2026-04-10
ci: optimize GitHub workflows — consolidate, add concurrency, fix bugs, matrix refactor

jpr5:
## Summary

Comprehensive optimization of all 25 GitHub Actions workflows. Net reduction of ~900 lines.

**Analysis & details:** [CopilotKit GitHub Workflows: Analysis & Optimization](https://www.notion.so/33e3aa38185281ffbe35e34ff2ac867f)

### Changes

**Path filters & waste reduction**
- `publish-commit.yml`: Added `packages/**` path filter — was firing on every push to every branch with no filters at all
- Upgraded `actions/checkout` from @v2 to @v4, added `--frozen-lockfile`, added pnpm cache

**Concurrency controls**
- Added `cancel-in-progress: true` concurrency groups to: `test_unit`, `e2e_dojo`, `e2e_examples`, `test_runtime-servers`, `static_quality`
- Prevents superseded CI runs from piling up on rapid pushes

**Consolidation**
- Merged standalone `static_commitlint.yml` into `static_quality.yml` as a 4th parallel job
- Same triggers, same setup — no reason for a separate workflow

**Bug fixes**
- `static_danger.yml`: Fixed phantom `matrix.node-version` reference (no matrix was defined), upgraded actions from @v2/@v3 to @v4, added timeout
- `test_unit-python-sdk.yml`: Removed `poetry lock` in CI (was regenerating lockfile instead of using committed one), added concurrency + timeout

**Auto-fix on PRs**
- Format job now auto-runs the formatter and commits fixes back to the PR branch
- Commitlint posts a helpful PR comment with valid prefixes and rebase instructions on failure

**showcase_deploy.yml matrix refactor**
- Replaced 23 structurally identical build jobs with a single dynamic matrix job
- **1,193 → ~230 lines** — service config encoded as JSON, matrix generated from path-filter results
- Special cases (shell LFS, build-args, custom Dockerfile) handled via matrix properties
- Railway deploy step conditionally runs only for services with a Railway service ID

**Action version standardization**
- Upgraded `actions/checkout` and `actions/setup-node` to @v4 across all modified workflows
- Added pnpm cache configuration where missing

### Files changed
- 10 workflow files modified, 1 deleted (`static_commitlint.yml`)
- Net: -900 lines

## Test plan

- [ ] Verify `publish-commit` no longer fires on docs/examples changes
- [ ] Verify concurrency cancels superseded runs on a PR with rapid pushes
- [ ] Verify `static_quality` runs all 4 jobs (format, oxlint, package-quality, commitlint)
- [ ] Verify format auto-fix commits back to PR on formatting violations
- [ ] Verify commitlint posts helpful comment on bad commit messages
- [ ] Verify `showcase_deploy` matrix correctly builds only changed services
- [ ] Verify `showcase_deploy` workflow_dispatch with specific service works
- [ ] Verify `showcase_deploy` Railway deploy triggers for services with IDs
- [ ] Verify `static_danger` no longer has empty node-version


marthakelly (dismissed):
Two questions on `showcase_deploy.yml`, want to make sure these are intentional before this merges.

---

### 1. Shell path filter (line 65)

Three paths from the original filter don't appear to be in this version:
- `showcase/shared/**`
- `showcase/scripts/**`
- `showcase/packages/*/manifest.yaml`

Was that intentional? The shell service is the only one with a repo-root Docker context, so changes to shared infrastructure wouldn't trigger a rebuild without those paths. If they were deliberately removed, would be good to understand the reasoning.

---

### 2. `curl` flag on Railway deploy (line 202)

Looks like `-f` was dropped, leaving just `-s`. Was that a deliberate change? Without `-f`, a non-2xx response from the Railway API (expired token, wrong service ID, 5xx) will still exit 0 — so a failed deploy would show green in CI. Happy to be wrong if there's something else catching that case.

---

Might be intentional, just wanted to flag before it goes in.

jpr5:
Both great catches — and both were caught and fixed during code review before the commit rewrite:

1. **Shell path filter**: `showcase/shared/**`, `showcase/scripts/**`, and `showcase/packages/*/manifest.yaml` are all present (lines 67-69 of the current `showcase_deploy.yml`).

2. **`curl -f` flag**: Restored — line 218 uses `curl -sf` with `&& echo "Deploy triggered"` confirmation. 

The 9 original commits were squashed into 5 clean `ci:`-prefixed commits, which is why the intermediate fix commits aren't visible in the history — the fixes are baked into the final state.

---

Separately, the `test / unit / python-sdk` failure is because `pyproject.toml` and `poetry.lock` are out of sync on main. The old workflow masked this with `poetry lock && poetry install` (regenerating the lockfile in CI). We removed `poetry lock` to enforce reproducible builds — fixing the lockfile in the next push.

marthakelly (dismissed):

.github/workflows/static_danger.yml
tylerslaton, line 1: What even is this file for? Seems unnecessary to begin with.

sdk-python/pyproject.toml
```diff
@@ -24,7 +24,7 @@ python = ">=3.10,<3.13"
 langgraph = { version = ">=0.3.25,<2" }
 langchain = { version = ">=0.3.0" }
 crewai = { version = "0.118.0", optional = true }
-ag-ui-langgraph = { version = ">=0.0.29", extras = ["fastapi"] }
+ag-ui-langgraph = { version = ">=0.0.29,<0.0.32", extras = ["fastapi"] }
```
tylerslaton, line 27: Seems smelly, why is this here? Was something breaking?

tylerslaton (approved):

jpr5:
Addressing @tylerslaton's review comments:

**`static_danger.yml`** — Pre-existing workflow, runs `danger ci` on LangGraph file changes. Only change was fixing a bug (referenced `matrix.node-version` but no matrix existed) and upgrading to v4. Happy to delete the file entirely in a follow-up if the team doesn't use Danger.

**`pyproject.toml` pin `<0.0.32`** — `ag-ui-langgraph 0.0.32` has a broken dep on `ImageInputContent` (not exported by current `ag-ui-protocol`). This breaks `test / unit / python-sdk` when `poetry lock` resolves to latest. Removing the pin — the upstream break isn't our problem to fix in this PR. The python test will fail until upstream publishes a fix, but it would also fail on main if re-run today.

PR #3747, opened 2026-04-10
fix(runtime): widen TanStack message content type for adapter compat

AlemTuzlak:
## Summary

- Widen `TanStackChatMessage.content` from `TanStackContentPart[]`
  to `any[]` so messages from `convertInputToTanStackAI` are directly
  passable to any TanStack AI adapter without `as any` casts
- Split `TanStackContentPart` into a proper discriminated union with
  separate variants per modality
- Add `env.d.ts` with `vite/client` reference for CSS import types
- Fix `onError` callback shape in the example

## Test plan

- [x] All 17 multimodal TanStack tests pass
- [x] All 294 agent tests pass
- [ ] Verify no TS errors in react-router example IDE

### DIFF
--- a/packages/runtime/src/agent/converters/tanstack.ts
+++ b/packages/runtime/src/agent/converters/tanstack.ts
@@ -11,24 +11,32 @@ import {
 } from "@ag-ui/client";
 import { randomUUID } from "@copilotkit/shared";
 
+type ContentPartSource =
+  | { type: "data"; value: string; mimeType: string }
+  | { type: "url"; value: string; mimeType?: string };
+
 /**
  * A TanStack AI content part (text, image, audio, video, or document).
  */
 export type TanStackContentPart =
   | { type: "text"; content: string }
-  | {
-      type: "image" | "audio" | "video" | "document";
-      source:
-        | { type: "data"; value: string; mimeType: string }
-        | { type: "url"; value: string; mimeType?: string };
-    };
+  | { type: "image"; source: ContentPartSource }
+  | { type: "audio"; source: ContentPartSource }
+  | { type: "video"; source: ContentPartSource }
+  | { type: "document"; source: ContentPartSource };
 
 /**
  * Message format expected by TanStack AI's `chat()`.
+ *
+ * Content is typed as `any[]` for the multimodal case so messages are directly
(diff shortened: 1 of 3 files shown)


packages/runtime/src/agent/converters/tanstack.ts
```diff
@@ -11,24 +11,32 @@ import {
...
+ * Content is typed as `any[]` for the multimodal case so messages are directly
+ * passable to any adapter without casts — different adapters constrain which
+ * modalities they accept (e.g. OpenAI only allows text + image).
+ * Use `TanStackContentPart` to inspect individual parts if needed.
  */
 export interface TanStackChatMessage {
   role: "user" | "assistant" | "tool";
-  content: string | null | TanStackContentPart[];
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  content: string | null | any[];
```
marthakelly, line 39: Two options to consider here depending on how far you want to take it:

**Quick fix:** swap `any[]` for `unknown[]`. Callers will need to narrow before accessing elements, but it signals "intentionally opaque" rather than "type checking off," and it's a zero-API-break change.

**Preferred:** make it generic with a default. `TanStackChatMessage<T extends readonly unknown[] = TanStackContentPart[]>` with `content: string | null | T`. Existing callers are untouched, adapter authors can parameterize with their own part type, and we can drop the `eslint-disable` comment entirely.

marthakelly (approved):

packages/runtime/src/agent/converters/tanstack.ts
AlemTuzlak: the reason this is typed as any is because tanstack ai is very specific about its typesafety, eg you can't pass a document message into a model that doesn't support it, which means it's impossible to implement a function that is a convenience method on top that can return the proper types without taking in the model the user provided and make it super unergonomic. And I'd rather have the users NOT use this function and opt into just validating input.messages themselves if they want ultimate typesafety on their side. so i'm not really sure how I feel about doing either of the 2, the 2nd one is just lying to typescript with a generic 😓

marthakelly: that makes sense! thanks for the follow up.

marthakelly (approved):

PR #3759, opened 2026-04-10
fix(react-ui): restore chat window layout broken in 1.55

marthakelly:
## Summary

Two UI regressions introduced in 1.55, both in `react-ui`:

**1. Input not pinned to bottom of chat window**
The attachments drag-and-drop feature (`fc1cfa09d`) wrapped messages+input in a `<div className={dragOver ? "copilotKitDragOver" : ""}>`. When not dragging that's `class=""` — no CSS rule targets it, so it collapses to content height inside the flex-column window. Fix: always apply `copilotKitChatBody` class and give it `flex: 1 1 0%; display: flex; flex-direction: column; min-height: 0`.

**2. Empty space below app content when sidebar is mounted**
`818e4e767` added `min-height: 100vh; height: 100%` to `.copilotKitSidebarContentWrapper`, causing a full-viewport gap below page content even when collapsed. Fix: remove those height rules — the wrapper only needs `margin-right` to push content left.

## Test plan

- [ ] Open `CopilotSidebar` — input pinned to bottom, messages fill remaining space
- [ ] Open `CopilotPopup` — same
- [ ] Mount `CopilotSidebar` collapsed — no empty space below app content
- [ ] Open/close sidebar — margin-right transitions correctly, no blank space
- [ ] Drag file over chat — `copilotKitDragOver` class still applies
- [ ] CI passes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3760, opened 2026-04-10
fix(react-ui): remove min-height from CopilotSidebarContentWrapper

marthakelly:
## Summary

- `818e4e767` added `min-height: 100vh`, `height: 100%` to `.copilotKitSidebarContentWrapper` and its children to support full-height content inside the sidebar
- This caused empty space equal to the viewport height to appear **below the app content** whenever `CopilotSidebar` was mounted, even when collapsed
- The wrapper only needs `margin-right` to push content left when the sidebar opens — it doesn't need to control its own height
- Remove the three height rules; child layout is the app's responsibility

## Test plan

- [ ] Mount `CopilotSidebar` — no empty space below app content when sidebar is collapsed
- [ ] Open sidebar — page content shifts left correctly via `margin-right`
- [ ] Close sidebar — margin-right transitions back to 0, no blank space
- [ ] CI passes

### DIFF
--- a/packages/react-ui/src/css/sidebar.css
+++ b/packages/react-ui/src/css/sidebar.css
@@ -37,19 +37,6 @@
   overflow: visible;
   margin-right: 0px;
   transition: margin-right 0.3s ease;
-  min-height: 100vh;
-  min-height: 100dvh;
-  height: 100%;
-}
-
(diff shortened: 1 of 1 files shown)


marthakelly:
Combining into #3759.

PR #3764, opened 2026-04-11
fix: pin @ag-ui/client to 0.0.52 in CLI templates

marthakelly:
## Summary

- All CLI-scaffolded templates imported `HttpAgent` from `@ag-ui/client` but pinned it at stale versions (`0.0.40`–`0.0.46`)
- `@copilotkit/runtime@1.55.2` ships with `@ag-ui/client@0.0.52`, which added `private _debug`, `private _debugLogger`, and a `debugLogger` getter to `AbstractAgent`
- npm installs two copies; TypeScript's nominal typing for private class members rejects `HttpAgent` (old copy) as assignable to `AbstractAgent` (new copy), breaking `next build` in every CLI-scaffolded project
- Fix: pin all templates to `@ag-ui/client@0.0.52` to match runtime

## Root cause (precise)

`AbstractAgent` in `0.0.52` gained private fields. TypeScript requires that private members in a class type come from the **same class declaration** — so `HttpAgent` from `0.0.42` (which extends `AbstractAgent@0.0.42`) is not assignable to `AbstractAgent@0.0.52`, even though the shapes are otherwise identical. The error surfaces as `HttpAgent is missing: _debug, _debugLogger, debugLogger`.

## Test plan

- [ ] Scaffold a new project from any of the affected templates (adk, agno, crewai-flows, pydantic-ai, strands, ms-agent-*)
- [ ] Run `npm install && npm run build` — should complete without type errors
- [ ] Verify `HttpAgent` resolves from the same `@ag-ui/client` copy as `@copilotkit/runtime`

### DIFF
--- a/examples/integrations/a2a-middleware/package.json
+++ b/examples/integrations/a2a-middleware/package.json
@@ -16,7 +16,7 @@
   "dependencies": {
     "@a2a-js/sdk": "latest",
     "@ag-ui/a2a-middleware": "0.0.2",
-    "@ag-ui/client": "0.0.40",
+    "@ag-ui/client": "0.0.52",
     "@copilotkit/react-core": "latest",
     "@copilotkit/react-ui": "latest",
     "@copilotkit/runtime": "latest",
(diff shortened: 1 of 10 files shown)


marthakelly:
Folded into #3761 — needed @copilotkit/* and @ag-ui/client bumped together to avoid peer dep conflict.

PR #3777, opened 2026-04-12
fix: set human_input_mode=NEVER on ag2 showcase agent

jpr5:
## Summary

Two fixes:

### 1. AG2 showcase agent blocks on human feedback
The `ConversableAgent` defaults to `human_input_mode="TERMINATE"`, which prompts for user feedback when the conversation terminates. In the AG-UI streaming context via `AGUIStream`, this blocks with:
> "Please give feedback to the sender. Press enter to skip and use auto-reply, or type 'exit' to stop the conversation:"

This was broken from day one (showcase deployed April 8, first L4 test ran April 9 and failed). The ag2 image was never rebuilt since — the unpinned `ag2>=0.9.0` dep resolved to 0.11.5 at build time.

Fix: set `human_input_mode="NEVER"`.

### 2. Showcase E2E Slack alerts lack detail
The Slack notification only said "Showcase E2E suite failed" with a link. Now captures playwright output and includes failed test names + error messages:

```
❌ Showcase E2E suite failed
View run

  1) [L4: tools] ag2 renders tool results @tools
    Error: ag2 response doesn't contain weather info: "Please give feedback..."
```

Test names now carry their level prefix (`[L1: health]`, `[L2: agent]`, `[L3: chat]`, `[L4: tools]`) so you can tell what broke from the Slack message alone.

### DIFF
--- a/.github/workflows/showcase_drift-detection.yml
+++ b/.github/workflows/showcase_drift-detection.yml
@@ -40,17 +40,37 @@ jobs:
         run: npx playwright install --with-deps chromium
 
       - name: Run L1-L3 tests
+        id: l1_l3
         working-directory: showcase/tests
         env:
           CI: true
-        run: npx playwright test integration-smoke --grep "@health|@agent|@chat" --reporter=github
+        run: |
+          npx playwright test integration-smoke --grep "@health|@agent|@chat" --reporter=github 2>&1 | tee /tmp/e2e-output.log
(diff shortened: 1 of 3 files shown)


marthakelly (approved):
The fix is right. The line, `human_input_mode="NEVER"`, cuts out the `check_termination_and_human_reply` path. I saw it reproduce before, and pass after. LGTM.

Issue #1691, opened 2025-04-18
🐛 Bug: TypeError: Cannot read properties of undefined (reading 'forEach') in filterAdjacentAgentStateMessages even when defined on window


yshimamoto:
### ♻️ Reproduction Steps

Hi CopilotKit team,

I am experiencing a persistent error when using CopilotKit in my React project:

```
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'forEach')
    at filterAdjacentAgentStateMessages (chunk-XXXX.js:XXXX)
```

## What I did

- I followed the official sample and defined a global dummy function in my code:
  ```js
  window.filterAdjacentAgentStateMessages = function (messages) {
    return Array.isArray(messages) ? messages : [];
  };
  ```
- I confirmed via `console.log(window.filterAdjacentAgentStateMessages)` that the function is correctly defined and available on the window object.
- I import this initialization at the very top of my entry point (`index.tsx`), before any CopilotKit components are used.
- All CopilotKit packages are version `1.8.5`.

## What happens

Despite the above, I still get the error in the browser console, and the stack trace points to a CopilotKit bundle file (e.g., `chunk-XXXX.js`).  
It seems that CopilotKit's internal code is not referencing the global `window.filterAdjacentAgentStateMessages`, or is calling `forEach` on an undefined value.

## Additional info

- This does not happen in the official sample, but happens in my project even with the same versions and initialization order.
- The error occurs when using CopilotKit chat features.
- I have tried clearing cache, reinstalling node_modules, and updating all packages.

## Request

- Please check if CopilotKit is correctly referencing the global `window.filterAdjacentAgentStateMessages` in all cases.
- Please add a null/undefined check before calling `forEach` on messages in the relevant code.
- If you need more details (logs, reproduction repo, etc.), I am happy to provide them.

Thank you for your support!

### ❌ Actual Behavior

I still get the error in the browser console, and the stack trace points to a CopilotKit bundle file (e.g., `chunk-XXXX.js`).  
It seems that CopilotKit's internal code is not referencing the global `window.filterAdjacentAgentStateMessages`, or is calling `forEach` on an undefined value.

### 𝌚 CopilotKit Version

```shell
- All CopilotKit packages are version `1.8.5`.
```

### 📄 Logs (Optional)

```shell
copilotkit.ts:60 window.filterAdjacentAgentStateMessages ƒ (messages) {
  return Array.isArray(messages) ? messages : [];
}
[...]

PR #3792, opened 2026-04-12
fix: guard filterAdjacentAgentStateMessages against undefined messages (#1691)

jpr5:
## Summary
- `filterAdjacentAgentStateMessages` called `.forEach` on the `messages` parameter without null-checking, causing a `TypeError` when messages was `undefined` or `null`
- Added `(messages || [])` guard so the function returns an empty array instead of crashing
- Added tests for both `undefined` and `null` inputs

## Test plan
- [x] New test: call with `undefined` -- returns `[]`
- [x] New test: call with `null` -- returns `[]`
- [x] All existing conversion tests still pass (131 tests)


jpr5:
Closing — this touches v1 code which is not worth the effort per Tyler.

PR #3802, opened 2026-04-12
fix: prevent TypeError when abort reason is string, not Error (#2596)

### LINKED ISSUE #2596: 🐛 Bug: Cannot read properties of undefined (reading 'includes') when stopping agent early
### ♻️ Reproduction Steps

1. Use local runtime (not sure if this happens also with cloud runtime or not)
2. Initiate the conversation with agent by sending user message (we call `onSend` from custom `ChatInput`)
3. ASAP stop the conversation with the stop button (we call `onStop` from custom `ChatInput`)
4. When you stop the conversation before the stream from CopilotKit runtime is established, the following error occurs:

`Error: [Network] Cannot read properties of undefined (reading 'includes')`

I tracked this problem down to this code which assumes that `error` is an `Error` object and casts it that way:

https://github.com/CopilotKit/CopilotKit/blob/7e83dfdefb0bd36094be5d06a1176ac2a0e918d0/CopilotKit/packages/runtime-client-gql/src/client/CopilotRuntimeClient.ts#L53

But the aborting code uses `cause` with string value `"Stop was called"` that the underlying call to `fetch` passes directly to the above `catch` block

https://github.com/CopilotKit/CopilotKit/blob/5ea96ab80ce52dc74a1861ce250eadb0c31f4813/CopilotKit/packages/react-core/src/hooks/use-chat.ts#L1015

Violating this explicit cast, ending up with `error.message` being `undefined`

Image

The error handling code should be more robust, not assuming specific type and/or the abort reason should be an error object, e.g. `new Error("Stop was called")`

This code seems the same in the latest codebase, so assuming the problem persists in the newest version

### ✅ Expected Behavior

The stop conversation action is handled properly, just as in the case the stream is aborted

### ❌ Actual Behavior

Error handling fails on null reference trying to access non-existing property of the error, and the error incorrectly changes to a more general error

`Error: [Network] Cannot read properties of undefined (reading 'includes')`

### 𝌚 CopilotKit Version

```shell
1.10.4
```

### 📄 Logs (Optional)

```shell

```

jpr5:
## Summary

- `CopilotRuntimeClient.ts` fetch wrapper crash when `AbortController.abort()` is called with a string reason instead of an Error object
- `(error as Error).message.includes(...)` throws TypeError when `.message` is undefined
- Added optional chaining: `(error as Error)?.message?.includes(...)`

## Test plan

- [x] Added `abort-error-handling.test.ts` demonstrating the TypeError with the old pattern and safe behavior with the fix
- [x] All 131 runtime-client-gql tests pass

Fixes #2596

### DIFF
--- a/packages/runtime-client-gql/src/client/CopilotRuntimeClient.ts
+++ b/packages/runtime-client-gql/src/client/CopilotRuntimeClient.ts
@@ -53,8 +53,8 @@ const createFetchFn =
     } catch (error) {
       // Let abort error pass through. It will be suppressed later
       if (
-        (error as Error).message.includes("BodyStreamBuffer was aborted") ||
-        (error as Error).message.includes("signal is aborted without reason")
+        (error as Error)?.message?.includes("BodyStreamBuffer was aborted") ||
+        (error as Error)?.message?.includes("signal is aborted without reason")
       ) {
         throw error;
       }
--- a/packages/runtime-client-gql/src/client/__tests__/abort-error-handling.test.ts
+++ b/packages/runtime-client-gql/src/client/__tests__/abort-error-handling.test.ts
@@ -1,0 +1,70 @@
+import { describe, it, expect, vi, afterEach } from "vitest";
+
+/**
+ * Test for #2596: Abort before stream crashes on undefined .message
+ *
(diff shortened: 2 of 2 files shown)


jpr5:
Closing — this touches v1 code which is not worth the effort per Tyler.

PR #3808, opened 2026-04-12
fix: prevent Enter key from sending during IME composition (#1862)

### LINKED ISSUE #1862: 🐛 Bug: Press enter for sending message isn't compatible with some input keyboard
### ♻️ Reproduction Steps

1.  Most of the asian characters input methods rely onthe  Enter key for confirming the character input.  
2.  Copilotkit doesn't handle this well when user press enter the message is sent, but user just want to confirm the character selection. 
3.
...

### ✅ Expected Behavior

 https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/isComposing       detect `isComposing` before sending can prevent this bug 

### ❌ Actual Behavior

message is sent with wrong words

### 𝌚 CopilotKit Version

```shell
1.8.13
```

### 📄 Logs (Optional)

```shell

```

jpr5:
Fixes #1862

Red-green tested locally.

### DIFF
--- a/packages/react-core/src/v2/components/chat/CopilotChatInput.tsx
+++ b/packages/react-core/src/v2/components/chat/CopilotChatInput.tsx
@@ -434,7 +434,7 @@ export function CopilotChatInput({
       }
     }
 
-    if (e.key === "Enter" && !e.shiftKey) {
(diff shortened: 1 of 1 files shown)


jpr5:
Closing — this is a subset of #3835 which covers the same fix more completely.

PR #3809, opened 2026-04-12
fix: allow maxTokens override for unrecognized OpenAI-compatible models (#1950)

### LINKED ISSUE #1950: 🐛 Bug: OpenRouter models default to 128k token limit instead of actual model limits
# Issue: OpenRouter models default to 128k token limit instead of actual model limits

## Problem Description

When using OpenRouter with models like `google/gemini-2.5-pro-preview`, CopilotKit's token validation throws "Not enough tokens for system message" error because it defaults to the conservative `DEFAULT_MAX_TOKENS = 128000` limit, even though Gemini 2.5 Pro Preview actually supports up to 1M tokens.

## Current Behavior

1. Using OpenRouter with `google/gemini-2.5-pro-preview` model
2. Setting OpenRouter base URL as the OpenAI endpoint
3. CopilotKit uses OpenAI adapter but doesn't recognize the model name
4. Falls back to `DEFAULT_MAX_TOKENS = 128000`
5. Throws error when system message + files exceed 128k tokens

## Expected Behavior

CopilotKit should recognize OpenRouter model names and apply the correct token limits for the underlying models.

## Environment Details

- **Model**: `google/gemini-2.5-pro-preview` via OpenRouter
- **Base URL**: OpenRouter's OpenAI-compatible endpoint
- **Actual Model Limit**: 2M tokens
- **CopilotKit Applied Limit**: 128k tokens
- **Error**: "Not enough tokens for system message"

## Workaround

Currently, users must reduce input size instead of utilizing the model's full capacity.

Thank you for considering this enhancement!

jpr5:
Fixes #1950

Red-green tested locally.

### DIFF
--- a/packages/runtime/src/service-adapters/openai/utils.ts
+++ b/packages/runtime/src/service-adapters/openai/utils.ts
@@ -60,8 +60,11 @@ export function limitMessagesToTokenCount(
   return result;
 }
 
-export function maxTokensForOpenAIModel(model: string): number {
-  return maxTokensByModel[model] || DEFAULT_MAX_TOKENS;
+export function maxTokensForOpenAIModel(
+  model: string,
+  overrideMaxTokens?: number,
+): number {
+  return maxTokensByModel[model] || overrideMaxTokens || DEFAULT_MAX_TOKENS;
 }
 
 const DEFAULT_MAX_TOKENS = 128000;


jpr5:
Closing — dead code removal with no functional impact. Not worth the review overhead.

PR #3813, opened 2026-04-12
fix: add null check for navigator.clipboard across all copy-to-clipboard calls (#2114)

### LINKED ISSUE #2114: Bug: navigator.clipboard undefined in non-localhost environments causes copy functionality to fail
### **Description**
CopilotKit's copy functionality fails in non-localhost environments because it directly calls `navigator.clipboard.writeText()` without checking if `navigator.clipboard` exists.

### **Environment**
- **CopilotKit version**: 1.8.13
- **Browser**: Chrome/Firefox/Safari (all affected)
- **Environment**: Non-localhost (IP address, custom domain, production)
- **Framework**: Next.js/React

### **Steps to Reproduce**
1. Deploy CopilotKit app to non-localhost environment (e.g., `192.168.1.100:3000` or custom domain)
2. Generate AI response with code blocks
3. Click the copy button in code block toolbar
4. See error in console: `Cannot read properties of undefined (reading 'writeText')`

### **Expected Behavior**
Copy functionality should work in all environments, including non-localhost deployments.

### **Actual Behavior**
Copy button throws JavaScript error and fails to copy content.

### **Error Details**
Uncaught TypeError: Cannot read properties of undefined (reading 'writeText')
at handleCopy (chunk-XMNTLIK5.mjs:31:27)

### **Root Cause**
- `navigator.clipboard` is `undefined` in non-localhost environments due to browser security policies
- CopilotKit directly calls `navigator.clipboard.writeText()` without checking availability
- Browsers apply stricter security policies for non-localhost domains

### **Proposed Solution**
Add clipboard availability check and fallback mechanism:

```javascript
// Before
navigator.clipboard.writeText(text)

// After
if (navigator.clipboard?.writeText) {
  navigator.clipboard.writeText(text)
} else {
  // Fallback using document.execCommand
  fallbackCopyToClipboard(text)
}
Workaround
Currently using a polyfill in application code:
if (typeof window !== 'undefined' && !navigator.clipboard) {
  navigator.clipboard = {
    writeText: (text) => {
      // Fallback implementation using document.execCommand
    }
  };
}
Impact
Affects all CopilotKit users deploying to production environments
Breaks core copy functionality in real-world deployments
Forces users to implement workarounds
Additional Context
This is a common issue with Clipboard API that requires proper feature detection and fallback mechanisms for production environments.

jpr5:
Fixes #2114

Red-green tested locally.

### DIFF
--- a/packages/angular/src/lib/components/chat/copilot-chat-assistant-message-buttons.ts
+++ b/packages/angular/src/lib/components/chat/copilot-chat-assistant-message-buttons.ts
@@ -20,6 +20,7 @@ import {
 import { CopilotTooltip } from "../../directives/tooltip";
 import { cn } from "../../utils";
 import { injectChatLabels } from "../../chat-config";
+import { copyToClipboard } from "@copilotkit/shared";
 
 // Base toolbar button component
 @Component({
@@ -121,21 +122,13 @@ export class CopilotChatAssistantMessageCopyButton {
     event?.stopPropagation();
     if (!this.content()) return;
 
-    if (!navigator.clipboard?.writeText) {
-      console.error("Clipboard API is not available");
-      return;
-    }
-
-    navigator.clipboard.writeText(this.content()!).then(
-      () => {
+    copyToClipboard(this.content()!).then((success) => {
+      if (success) {
         this.copied.set(true);
         this.clicked.emit();
         setTimeout(() => this.copied.set(false), 2000);
-      },
-      (err) => {
-        console.error("Failed to copy message:", err);
-      },
-    );
+      }
+    });
   }
 }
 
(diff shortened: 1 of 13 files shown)


jpr5:
Addressed the feedback about duplicated clipboard code blocks.

**What changed:** Extracted a shared `copyToClipboard(text: string): Promise<boolean>` utility into `@copilotkit/shared` (`packages/shared/src/utils/clipboard.ts`). It handles:
- Clipboard API availability check
- `writeText` call with error handling
- Returns `true`/`false` for success/failure

All 9 call sites across `@copilotkitnext/angular`, `@copilotkit/react-core`, and `@copilotkit/react-ui` now import and use this shared utility instead of duplicating the same check+write+catch pattern.

Added unit tests for the utility itself in `packages/shared/src/utils/__tests__/clipboard.test.ts`. All existing tests continue to pass (1076 react-core tests, 136 shared tests, 2 react-ui tests).

AlemTuzlak (approved):

PR #3814, opened 2026-04-12
fix: ensure TextMessageOutput.id is never null by falling back to randomId() (#2118)

### LINKED ISSUE #2118: 🐛 Bug:  runtime return error:   Cannot return null for non-nullable field TextMessageOutput.id
### ♻️ Reproduction Steps

use latest version （1.9.x） and implement it according to the node-http demo, the server returns the following error：

`
---
Content-Type: application/json; charset=utf-8
Content-Length: 164

{"data":{"generateCopilotResponse":{"threadId":"ck-014359a9-6787-40e3-bbae-eaad840e821f","runId":null,"__typename":"CopilotResponse","messages":[]}},"hasNext":true}
---
Content-Type: application/json; charset=utf-8
Content-Length: 274

{"incremental":[{"items":null,"path":["generateCopilotResponse","messages",0],"errors":[{"message":"Cannot return null for non-nullable field TextMessageOutput.id.","locations":[{"line":23,"column":9}],"path":["generateCopilotResponse","messages",0,"id"]}]}],"hasNext":true}
---
Content-Type: application/json; charset=utf-8
Content-Length: 175

{"incremental":[{"data":{"__typename":"CopilotResponse","status":{"code":"Success","__typename":"SuccessResponseStatus"}},"path":["generateCopilotResponse"]}],"hasNext":false}
-----
`
The runtime uses the openai adapter and uses the internal chat interface. I have tested the llm interface and there is no problem。Not sure if it is an internal format verification problem

Using a lower version, such as 1.5 or 1.6, also has the same error
There is no error only when using version 1.3.x，but the 1.3.x version lacks some APIs

### ✅ Expected Behavior

Return the correct llm response

### ❌ Actual Behavior

Always returns the above error

### 𝌚 CopilotKit Version

```shell
1.9.x
```

### 📄 Logs (Optional)

```shell

```

jpr5:
Fixes #2118

Red-green tested locally.


AlemTuzlak (changes requested):
don't red-green test it locally, add tests

ranst91 (approved):

AlemTuzlak (approved):

PR #3818, opened 2026-04-12
fix: add isActive toggle parameter to onThumbsUp/onThumbsDown callbacks (#2615)

### LINKED ISSUE #2615: 🚀 Feature Request: expand `onThumbsUp`/`onThumbsDown` type to pass the button state
## Problem

In custom implementations of `AssistantMessage` developer most likely will be adding state for the feedback buttons (aka thumbs up/down):

```tsx
<Button
    label={translations.actions.thumbsUp}
    icon={reactionValue === "like" ? ThumbsUpFilled : ThumbsUp}
    disabled={isGenerating}
    onClick={(e) => {
      setReactionValue((val) => (val === "like" ? null : "like"))
      onThumbsUp?.(message) // 👈  no way to provide info about the value
      e.currentTarget.blur()
    }}
/>
```

However there is no way to pass the state to the callback because the type is strictly set to:

```ts
onThumbsUp?: (message: Message) => void;
```

As a result, it is not possible to derive whether the button was clicked or un-clicked. The only way to understand this is to count those clicks for each message in memory, which is cumbersome

## Proposed solution

Add an optional parameter to the callback: `onThumbsUp?: (message: Message, value?: string | boolean) => void;`
Another alternative is to allow generic payload `onThumbsUp?: (message: Message, extra: unknown) => void;` which developer can use after refining the type

jpr5:
## Summary

Fixes #2615

Adds an optional `isActive` boolean parameter to `onThumbsUp` and `onThumbsDown` callback signatures across `MessagesProps`, `AssistantMessageProps`, and `RenderMessageProps`. This allows consumers to know whether the feedback button is being toggled on or off.

## Test plan

- [ ] Verify thumbs up/down callbacks receive `isActive` parameter
- [ ] Verify existing consumers without the parameter still work

### DIFF
--- a/packages/react-ui/src/components/chat/props.ts
+++ b/packages/react-ui/src/components/chat/props.ts
@@ -121,12 +121,12 @@ export interface MessagesProps {
   /**
    * Callback function for thumbs up feedback
    */
-  onThumbsUp?: (message: Message) => void;
+  onThumbsUp?: (message: Message, isActive?: boolean) => void;
 
   /**
    * Callback function for thumbs down feedback
    */
-  onThumbsDown?: (message: Message) => void;
+  onThumbsDown?: (message: Message, isActive?: boolean) => void;
 
   /**
    * Map of message IDs to their feedback state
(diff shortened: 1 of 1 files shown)


jpr5:
Closing — targets a dead/unused API surface. No user impact.

PR #3819, opened 2026-04-12
fix: accept function returning headers in addition to static Record (#2159)

### LINKED ISSUE #2159: 🚀 Feature Request: Add a way to provide headers in async
### 🎤 Tell us your idea

When using the self-hosted method, it is possible to do the following:

```
    <CopilotKit
      runtimeUrl={`${env.NEXT_PUBLIC_API_URI}/copilotkit`}
      headers={{
        Authorization: `Bearer ${token}`,
      }}
    >
```

yet, if you use Clerk, getting the token itself is an async function. This is also true for localStorage operations

in component
```
const { getToken } = useAuth()
```

then to get the token
```
await getToken()
```

which is not possible to do where `<CopilotKit />` is available.

### Example solution

instead of `headers: Record<string, string>` if we have `getHeaders: () => Promise<Record<string, string>>`
this can be achieved.

jpr5:
## Summary

Fixes #2159

The `headers` prop on `CopilotKit` only accepted a static `Record<string, string>`. This adds support for passing a function `(() => Record<string, string>)` that is evaluated at request time, enabling dynamic headers (e.g., for rotating auth tokens).

## Test plan

- [ ] Verify static `headers` Record still works
- [ ] Verify function-returning headers are called and values used

### DIFF
--- a/packages/react-core/src/components/copilot-provider/copilotkit-props.tsx
+++ b/packages/react-core/src/components/copilot-provider/copilotkit-props.tsx
@@ -76,7 +76,7 @@ export interface CopilotKitProps extends Omit<
    * }
    * ```
    */
-  headers?: Record<string, string>;
+  headers?: Record<string, string> | (() => Record<string, string>);
 
   /**
    * The children to be rendered within the CopilotKit.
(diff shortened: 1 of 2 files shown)


jpr5:
Superseded by #3853 which covers both v1 and v2 providers. Closing to avoid merge conflict.

PR #3821, opened 2026-04-12
fix: use div instead of p tag in Markdown to prevent hydration errors (#2234)

### LINKED ISSUE #2234: 🐛 Bug: In HTML, <div> cannot be a descendant of <p>. This will cause a hydration error.
### ♻️ Reproduction Steps

1. I am using ollama to run a dashboard chart function. The CopilotSidebar components throws a exception, **In HTML, <div> cannot be a descendant of <p>.This will cause a hydration error.**

Image

2.
3.
...

### ✅ Expected Behavior

When the chat return the answer, it should call the function and update the data on the chart function.

### ❌ Actual Behavior

The error happened on the browser console

### 𝌚 CopilotKit Version

```shell
├── @copilotkit/react-core@1.9.3
├── @copilotkit/react-ui@1.9.3
├── @copilotkit/runtime@1.9.3
```

### 📄 Logs (Optional)

```shell

```

jpr5:
## Summary

Fixes #2234

React's hydration fails when block-level elements (like `<div>`) are nested inside `<p>` tags in the Markdown component. Replaces the `<p>` wrapper with `<div>` (adding `copilotKitParagraph` class for styling) to prevent SSR hydration mismatches.

## Test plan

- [ ] Verify Markdown rendering in SSR context has no hydration errors
- [ ] Verify paragraph styling is preserved via `copilotKitParagraph` class

### DIFF
--- a/packages/react-ui/src/components/chat/Markdown.test.ts
+++ b/packages/react-ui/src/components/chat/Markdown.test.ts
@@ -1,6 +1,5 @@
 import { readFileSync } from "fs";
-import { resolve, dirname } from "path";
-import { fileURLToPath } from "url";
+import { resolve } from "path";
 
 /**
  * These tests verify that CSS selectors in markdown.css correctly target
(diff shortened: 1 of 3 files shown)


jpr5:
@AlemTuzlak Good catch on the semantic difference between `<p>` and `<div>`. I audited the CSS and confirmed this change is safe:

The existing `p.copilotKitMarkdownElement` rule (now `.copilotKitParagraph`) already had `margin: 0px; padding: 0px` explicitly set, so `<p>`'s default browser margins (`margin-top: 1em; margin-bottom: 1em`) were never in play. The inter-paragraph spacing was always handled by the `.copilotKitParagraph:not(:last-child) { margin-bottom: 1.25em }` rule, which still works identically with `<div>`.

Both `<p>` and `<div>` default to `display: block`, and `line-height: 1.75` and `font-size: 1rem` were already explicit, so there's no visual regression — the CSS was fully self-contained and never relied on `<p>` defaults.

AlemTuzlak (approved):

PR #3822, opened 2026-04-12
fix: add remarkPlugins/rehypePlugins passthrough support (#2296)

### LINKED ISSUE #2296: 🚀 Feature Request:
### 🎤 Tell us your idea

Instead of pass only "markdownTagRenderers" prop to Markdown component, can you implement the feature to pass all props of Markdown component. It's helpful!
{content && <Markdown content={content} components={markdownTagRenderers} />}

jpr5:
## Summary

Fixes #2296

The Markdown component had hardcoded remark/rehype plugins with no way for consumers to add their own. This adds optional `remarkPlugins` and `rehypePlugins` props that are spread into the existing plugin arrays, enabling custom markdown processing.

## Test plan

- [ ] Verify custom remarkPlugins are applied alongside built-in plugins
- [ ] Verify custom rehypePlugins are applied alongside built-in plugins
- [ ] Verify Markdown renders correctly without custom plugins (no regression)


AlemTuzlak (changes requested):
I've seen 2 PR's up to this point where we're adding props and passing them through, why not just use ...props and then provide them to the underlying component instead of adding props 1 by 1 through prs

AlemTuzlak (approved):

ranst91 (approved):

PR #3826, opened 2026-04-12
fix: remove unstable key on CodeBlock to prevent flickering during streaming

### LINKED ISSUE #2669: 🐛 Bug: Random key in CodeBlock (@copilotkit/react-ui) causes flickering during streaming

jpr5:
## Summary

- Removes the `key` prop from `CodeBlock` in `Markdown.tsx` entirely
- The original `Math.random()` key caused React to remount the component on every render, producing visible flickering
- A content-based key (language + content prefix) still changes every streaming token, causing the same problem
- Without an explicit key, React uses positional identity — stable across re-renders while content streams in

Closes #2669

## Test plan

- [x] `@copilotkit/react-ui` tests pass
- [x] `@copilotkit/react-ui` build succeeds
- [ ] Manual: verify code blocks no longer flicker during streaming in chat UI

### DIFF
--- a/.changeset/stable-codeblock-keys.md
+++ b/.changeset/stable-codeblock-keys.md
@@ -1,5 +1,0 @@
----
-"@copilotkit/react-ui": patch
----
-
-fix: remove unstable key on CodeBlock to prevent flickering during streaming
--- a/.changeset/stable-codeblock-keys.md
+++ b/.changeset/stable-codeblock-keys.md
@@ -1,0 +1,5 @@
+---
+"@copilotkit/react-ui": patch
+---
+
+fix: remove unstable key on CodeBlock to prevent flickering during streaming
--- a/packages/react-ui/src/components/chat/Markdown.tsx
+++ b/packages/react-ui/src/components/chat/Markdown.tsx
@@ -59,7 +59,6 @@ const defaultComponents: Components = {
 
     return (
       <CodeBlock
-        key={Math.random()}
         language={(match && match[1]) || ""}
         value={String(children).replace(/\n$/, "")}
         {...props}


AlemTuzlak (approved):

jpr5:
No — this won't cause duplicate key warnings.

CodeBlock is rendered as a react-markdown component override (`components={{ code: CodeBlock }}`), not in a `.map()` or array. react-markdown handles keying internally based on AST node position — each code block is a separate AST node at a different position and gets its own key from react-markdown.

The `Math.random()` key was actually harmful — it forced React to unmount/remount CodeBlock on every re-render, destroying internal state (copy button state, syntax highlighting). Removing it lets React use stable positional identity, which is correct for component overrides.

AlemTuzlak (approved):

PR #3837, opened 2026-04-12
fix: add Zod v4 compatibility for schema-to-JSON conversion

jpr5:
## Summary
- Detects Zod v4 schemas by checking for `toJSONSchema()` method on the schema object
- Calls `schema.toJSONSchema()` directly for Zod v4 (avoids `zod-to-json-schema` v3, which can't handle Zod v4 internals)
- Preserves existing Zod v3 path via injected `zodToJsonSchema` fallback
- Removes dead `vendor === "zod4"` check (Zod v4 reports vendor as `"zod"`, not `"zod4"`)
- Removes `_def` duck-typing fallback (unnecessary with `toJSONSchema` detection)

### Priority order:
1. Standard JSON Schema V1 (`~standard.jsonSchema.input`)
2. Zod v4 native (`schema.toJSONSchema()`)
3. Zod v3 fallback (injected `zodToJsonSchema`)

### Tests added:
- Verifies `toJSONSchema()` is called when present
- Verifies it works without `zodToJsonSchema` option
- Verifies it takes priority over `zodToJsonSchema` fallback
- Verifies `~standard.jsonSchema` still takes top priority over `toJSONSchema()`

Closes #3636

### DIFF
--- a/packages/shared/src/standard-schema.ts
+++ b/packages/shared/src/standard-schema.ts
@@ -48,9 +48,10 @@ function hasStandardJsonSchema(
  * Strategy:
  * 1. If the schema implements Standard JSON Schema V1 (`~standard.jsonSchema`),
  *    call `schema['~standard'].jsonSchema.input({ target: 'draft-07' })`.
- * 2. If the schema is a Zod v3 schema (`~standard.vendor === 'zod'`), use the
+ * 2. If the schema exposes a `toJSONSchema()` method (Zod v4), call it directly.
+ * 3. If the schema is a Zod v3 schema (`~standard.vendor === 'zod'`), use the
  *    injected `zodToJsonSchema()` function.
- * 3. Otherwise throw a descriptive error.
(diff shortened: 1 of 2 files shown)


AlemTuzlak (approved):

packages/shared/src/standard-schema.ts
```diff
@@ -61,7 +62,12 @@ export function schemaToJsonSchema(
     return schema["~standard"].jsonSchema.input({ target: "draft-07" });
   }
 
-  // 2. Zod v3 fallback
+  // 2. Zod v4 native — exposes toJSONSchema() on the schema itself
+  if (typeof (schema as any).toJSONSchema === "function") {
+    return (schema as any).toJSONSchema() as Record<string, unknown>;
```
stnguyen90, line 67: Are you sure the schema has the method? I thought the method was a static method on zod

https://zod.dev/json-schema?id=ztojsonschema

PR #3838, opened 2026-04-12
fix: batch fixes for issues #3141, #3217, #3631, #3637, #3741, #1748, #1921, #1936

jpr5:
## Summary

Batch of 8 bug fixes (5 HIGH, 3 MEDIUM priority):

- **#3141** — Forward credentials on all agent registration paths (HIGH)
  - `applyCredentialsToAgents`/`applyCredentialsToAgent` were defined but never called from `initialize()`, `setAgents__unsafe_dev_only()`, or `addAgent__unsafe_dev_only()`

- **#3217** — Prevent "Unknown provider" regression for LangChainAdapter (HIGH)
  - When `getLanguageModel()` returns null and `provider`/`model` are undefined, code constructed `"undefined/undefined"` as model string. Now checks each source explicitly and throws a clear error

- **#3631** — Patch defu CVE prototype pollution (HIGH)
  - Added pnpm override for `defu@<=6.1.4` to `>=6.1.5`

- **#3637** — Honor Retry-After header on 429 responses (HIGH)
  - Parse Retry-After as seconds or HTTP-date, use as retry delay. Throw if server requests wait > 60s

- **#3741** — Pass toolCallId to useRenderTool render components (HIGH)
  - Added `toolCallId` prop to all three status branches and updated `ReactToolCallRenderer` type

- **#1748** — Extract all text parts when AIMessage content is a list (MEDIUM)
  - Python SDK now iterates all content parts instead of only taking `content[0]`

- **#1921** — Pass urlTransform prop through to ReactMarkdown (MEDIUM)
  - Added `urlTransform` as optional prop with passthrough and memo comparison

- **#1936** — Preserve jsonSchema structure in MCP tool parameter extraction (MEDIUM)
  - `extractParametersFromSchema` now preserves enum values, recursively extracts nested object attributes, and handles object arrays

**Skipped:** #1937 (async headers) — overlaps with #2159, deferred to X2

## Test plan

- [x] MCP tools unit tests pass (17/17)
- [x] All affected packages build successfully
- [x] Formatted with oxfmt

PR #3842, opened 2026-04-12
fix: honor Retry-After header on 429 responses

jpr5:
Closes #3637

Parse Retry-After as seconds or HTTP-date and use it as the retry delay instead of exponential backoff. Throw if the server requests a wait longer than `maxRetryAfterSeconds` (60s default) to avoid hanging.

Split from #3838.


AlemTuzlak (dismissed):

AlemTuzlak:
@jpr5 let's add tests for this, fix looks good

ranst91 (approved):

PR #3847, opened 2026-04-12
fix: batch of community issue fixes (blitz X5)

jpr5:
## Summary

Batch of fixes for 7 community-reported issues:

- **#2137**: Replace `rm -rf` with `shx rm -rf` in runtime package.json for Windows compatibility
- **#2220**: Add `anyOf`/`oneOf`/`$ref` resolution and `integer`/`null` type handling in JSON schema to Zod conversion
- **#2920**: Scope `.dark` CSS selectors to CopilotKit elements (`.poweredBy`, `.copilotKitDevConsole`) instead of global selectors
- **#1955**: Sanitize `NaN`/`Infinity` float values in LangGraph state before JSON serialization
- **#2893**: Preserve graph-owned state keys during frontend state merge in `LangGraphAGUIAgent`
- **#2779**: Support dynamic headers function prop (`() => Record<string, string>`) for auth token refresh
- **#2137** (lockfile): Update pnpm-lock.yaml with shx dependency

## Test plan

- [x] All 132 `@copilotkit/shared` tests pass
- [x] All 1221 `@copilotkit/runtime` tests pass
- [x] `@copilotkit/react-core` builds successfully
- [x] `@copilotkit/react-ui` builds successfully
- [ ] Verify dark mode CSS scoping doesn't break `.poweredBy` styling
- [ ] Verify dynamic headers function receives fresh token on each request
- [ ] Verify LangGraph state merge preserves graph-owned keys

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3851, opened 2026-04-12
fix: sanitize NaN/Infinity values in LangGraph state before JSON serialization

### LINKED ISSUE #1955: NaN Breaks Proxy
Hi - im trying to surface an issue where we have a dictonary being returned in the state / messages stack mid graph, that contains a value of NaN.  Long story, but it breaks the way the JSON is handled in the proxy, looking for some ideas on how / why NaN seems to make its way via LangGraph stream event, into a message successfully, but then on the node side gets hung up.  I suspect its something deep inside this:

https://github.com/CopilotKit/CopilotKit/blob/77dfb43c55176d4b49966472e2e1ebfece3f42c2/sdk-python/copilotkit/langgraph_agent.py#L337

This comes up when trying to wrap the create_pandas_dataframe_agent from langchain into a basic graph.

jpr5:
## Summary
- Add sanitization pass over LangGraph agent state to replace NaN and Infinity with null before JSON serialization
- Prevents `ValueError: Out of range float values are not JSON compliant` in Python SDK

Closes #1955

---
*Split from #3847*

### DIFF
--- a/sdk-python/copilotkit/langgraph_agent.py
+++ b/sdk-python/copilotkit/langgraph_agent.py
@@ -2,6 +2,7 @@
 
 import uuid
 import json
+import math
 from typing import Optional, List, Callable, Any, cast, Union, TypedDict, Literal
 
 from langgraph.graph.state import CompiledStateGraph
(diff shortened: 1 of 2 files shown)


ranst91 (dismissed):
Approved, this but this implementation is kept for backwards compatibility and no longer maintained, the one that's maintained is available in AG-UI

AlemTuzlak (approved):

PR #3866, opened 2026-04-13
docs(mastra): align Agent snippet with requestContext, id, and contex…

Abubakar-01:
## Summary

Updates the Mastra integration **Agent app context** (“Readables”) docs so the
`Agent` example matches current Mastra behavior and CopilotKit’s injected
AG-UI context shape.

## Why

1. **API rename** — Mastra’s Agent `instructions` callback now documents
   **`requestContext`** (replacing the older `runtimeContext` usage shown in the
   previous snippet).
2. **Strict `Agent` config** — Mastra’s `Agent` type requires a unique **`id`**
   field; the example now includes it so copy-paste builds succeed.
3. **Correct context lookup** — The `find()` comparison string is aligned with
   the `useAgentContext` `description` value (`"The current user's colleagues"`)
   so the example actually resolves the injected item.

## Changes

- `docs/content/docs/integrations/mastra/agent-app-context.mdx` — Agent
  constructor snippet under the Mastra agent / readables flow.

## Notes

- If this page is mirrored under `showcase/shell/src/content/docs/...`, keep
  the showcase copy in sync with the same commit or a follow-up.

## Test plan

- [ ] Skim rendered MDX; confirm code block matches Mastra docs for
  `requestContext` / `Agent` options.


marthakelly:
Thanks for the contribution! Closed — mirrored to https://github.com/CopilotKit/CopilotKit/pull/4040 (#4040) where CI can run and the team will pick up review. Please follow up there.

PR #3870, opened 2026-04-13
fix: inspector not receiving AG-UI events from per-thread agent clones

NathanTarbert:
## Summary

Fixes the inspector showing no AG-UI events.

`useAgent()` returns per-thread clones (via `getOrCreateThreadClone`) that are not in the agent registry. The inspector subscribed to registry agents via `onAgentsChanged`, so it never received events from clones — resulting in an empty event panel.

Adds an `onAgentRunStarted` subscriber event to `CopilotKitCoreSubscriber` that fires before `connectAgent()` and `runAgent()`, passing the actual agent instance (which may be a per-thread clone). The inspector subscribes to that agent's events, matching the existing `StateManager` pattern (`subscribeAgentToStateManager`).

Per-thread cloning remains enabled — this does **not** touch `useAgent` or revert any cloning logic.

## Changes

| File | What |
|------|------|
| `packages/core/src/core/core.ts` | Add `onAgentRunStarted` to `CopilotKitCoreSubscriber` interface |
| `packages/core/src/core/run-handler.ts` | Fire `onAgentRunStarted` in both `connectAgent()` and `runAgent()` |
| `packages/web-inspector/src/index.ts` | Handle `onAgentRunStarted` by subscribing to the clone agent |

## Test plan

- [x] All 337 `@copilotkit/core` tests pass
- [x] All 2 `@copilotkit/web-inspector` tests pass
- [x] Both packages build cleanly
- [x] `oxfmt --check` passes
- [ ] Manual: open an app with `showDevConsole={true}`, send a message, verify AG-UI events appear in the inspector

### DIFF
--- a/packages/core/src/core/core.ts
+++ b/packages/core/src/core/core.ts
@@ -153,6 +153,15 @@ export interface CopilotKitCoreSubscriber {
     code: CopilotKitCoreErrorCode;
     context: Record<string, any>;
   }) => void | Promise<void>;
+  /**
+   * Fired when an agent run or connect begins. The `agent` may be a per-thread
+   * clone that is not present in `core.agents`. Subscribers (e.g. the inspector)
+   * can use this to subscribe to the clone's AG-UI events.
+   */
+  onAgentRunStarted?: (event: {
+    copilotkit: CopilotKitCore;
+    agent: AbstractAgent;
+  }) => void | Promise<void>;
 }
 
 // Subscription object returned by subscribe()
--- a/packages/core/src/core/run-handler.ts
+++ b/packages/core/src/core/run-handler.ts
@@ -206,6 +206,18 @@ export class RunHandler {
         };
       }
 
+      // Notify subscribers (e.g. the inspector) about the agent that is about
+      // to run. This is critical for per-thread clones that are not present in
+      // the agent registry and would otherwise be invisible to subscribers.
+      await this._internal.notifySubscribers(
+        (subscriber) =>
+          subscriber.onAgentRunStarted?.({
+            copilotkit: this.core,
+            agent,
+          }),
+        "Subscriber onAgentRunStarted error:",
+      );
+
       const runAgentResult = await agent.connectAgent(
         {
           forwardedProps: this._internal.properties,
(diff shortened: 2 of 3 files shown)


jpr5 (approved):
Clean regression fix. Per-thread cloning was intentionally disabled in ae48d476 (race condition with activity message renderers), accidentally re-enabled by the throttleMs merge, broke the inspector (subscribes to registry agents, not clones).

Single file, net -4 lines, returns to the original `return existing` behavior. Comment explains the history well.

Vercel "failures" are just authorization required from external fork — not code issues.

jpr5 (changes requested):
## Withdrawing approval — this would re-introduce bug #2957

Traced the full history of per-thread cloning:

1. **PR #3525** (marthakelly) added cloning to fix #2957 — multiple CopilotChat components with same agentId but different threadIds share message state. Real user-reported bug.

2. **ae48d476** was a temporary workaround that disabled cloning because `useRenderActivityMessage` was reading the registry agent instead of the clone (A2UI button responses vanished).

3. **PR #3630** (marthakelly, Apr 6) was the **proper fix** — updated `useRenderActivityMessage` to resolve per-thread clones via `getThreadClone()`. Both fixes are on main today.

4. The throttleMs commits (`af98091b` et al.) **do not touch cloning logic** — they only modify `effectiveThrottleMs` calculation and provider integration. There was no accidental re-enable.

**Cloning is enabled on main intentionally** because PR #3630 fixed the root cause that ae48d476 was working around.

### The inspector issue is real, but the fix is wrong

If the inspector subscribes to registry agents via `onAgentsChanged` and doesn't see events from clones, the fix should be:
- Make the inspector resolve per-thread clones via `getThreadClone(registryAgent, threadId)`, same pattern used in `useRenderActivityMessage` and `useRenderCustomMessages`
- OR have the inspector subscribe to clone events in addition to registry agents

Disabling cloning would re-introduce cross-thread message bleeding (#2957).

NathanTarbert:
Closing in favor of a new PR from an upstream branch so CI can check out the code.

PR #3872, opened 2026-04-13
fix: inspector not receiving AG-UI events from per-thread agent clones

NathanTarbert:
## Summary

Fixes the inspector showing no AG-UI events.

`useAgent()` returns per-thread clones (via `getOrCreateThreadClone`) that are not in the agent registry. The inspector subscribed to registry agents via `onAgentsChanged`, so it never received events from clones — resulting in an empty event panel.

Adds an `onAgentRunStarted` subscriber event to `CopilotKitCoreSubscriber` that fires before `connectAgent()` and `runAgent()`, passing the actual agent instance (which may be a per-thread clone). The inspector subscribes to that agent's events, matching the existing `StateManager` pattern (`subscribeAgentToStateManager`).

Per-thread cloning remains enabled — this does **not** touch `useAgent` or revert any cloning logic.

## Changes

| File | What |
|------|------|
| `packages/core/src/core/core.ts` | Add `onAgentRunStarted` to `CopilotKitCoreSubscriber` interface |
| `packages/core/src/core/run-handler.ts` | Fire `onAgentRunStarted` in both `connectAgent()` and `runAgent()` |
| `packages/web-inspector/src/index.ts` | Handle `onAgentRunStarted` by subscribing to the clone agent |

## Test plan

- [x] All 337 `@copilotkit/core` tests pass
- [x] All 2 `@copilotkit/web-inspector` tests pass
- [x] Both packages build cleanly
- [x] `oxfmt --check` passes
- [ ] Manual: open an app with `showDevConsole={true}`, send a message, verify AG-UI events appear in the inspector

### DIFF
--- a/packages/core/src/core/core.ts
+++ b/packages/core/src/core/core.ts
@@ -153,6 +153,15 @@ export interface CopilotKitCoreSubscriber {
     code: CopilotKitCoreErrorCode;
     context: Record<string, any>;
   }) => void | Promise<void>;
+  /**
+   * Fired when an agent run or connect begins. The `agent` may be a per-thread
+   * clone that is not present in `core.agents`. Subscribers (e.g. the inspector)
+   * can use this to subscribe to the clone's AG-UI events.
+   */
+  onAgentRunStarted?: (event: {
+    copilotkit: CopilotKitCore;
+    agent: AbstractAgent;
+  }) => void | Promise<void>;
 }
 
 // Subscription object returned by subscribe()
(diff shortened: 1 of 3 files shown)


jpr5 (approved):
This is the correct fix for the inspector issue — keeps per-thread cloning intact and makes the inspector subscribe to clone events.

- New `onAgentRunStarted` subscriber event fires before `connectAgent()` and `runAgent()`, passing the actual agent instance (may be a clone)
- Inspector subscribes to the clone's AG-UI events via existing `subscribeToAgent()` pattern
- Follows the same approach as `subscribeAgentToStateManager` — consistent with existing architecture
- 3 files, 41 lines, focused and clean

This should supersede #3870 (which proposed disabling cloning entirely, re-introducing bug #2957).

PR #3879, opened 2026-04-14
fix: smoke monitor rebuilds include shared tools copy

jpr5:
## Summary

The smoke monitor's drift-triggered rebuilds fail because the deploy workflow doesn't copy shared modules (shared_python/, shared_frontend/src/, shared_typescript/tools/) into each package's Docker build context before building.

- Adds a "Copy shared modules into build context" step to `showcase_deploy.yml`, placed between the build-args prep and the Docker build step
- The step conditionally copies each shared directory only if it exists, so it's safe for packages that don't use shared modules
- Fixes rebuilds triggered via `workflow_dispatch` (from smoke monitor drift detection) and manual dispatches alike

### DIFF
--- a/.github/workflows/showcase_deploy.yml
+++ b/.github/workflows/showcase_deploy.yml
@@ -199,6 +199,21 @@ jobs:
           echo "$ARGS" >> $GITHUB_OUTPUT
           echo "BUILDARGS_EOF" >> $GITHUB_OUTPUT
 
+      - name: Copy shared modules into build context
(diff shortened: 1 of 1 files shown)


jpr5:
Closing — this is superseded by main. The same functionality landed via 810f8a4f4 and was refined by 5 subsequent commits that fixed a trailing-slash cp bug, removed dead shared_frontend copy, and added starter guards.

PR #3885, opened 2026-04-14
fix(core): subscribe to ToolCallResultEvent in StateManager

### LINKED ISSUE #3884: AG-UI: tool result messages missing from conversation history on subsequent turns (causes LLM 400)
### ♻️ Reproduction Steps

1. Set up a CopilotKit v2 `<CopilotChat>` with an AG-UI `HttpAgent` backend that has server-side tools (e.g., MCP tools)
2. Send a message that triggers a tool call (e.g., "What's the weather in Seattle?")
3. Agent streams: `TOOL_CALL_START` → `TOOL_CALL_ARGS` → `TOOL_CALL_END` → `TOOL_CALL_RESULT` → `TEXT_MESSAGE_*` → `RUN_FINISHED`
4. First turn completes successfully — user sees the tool result and text response
5. Send a **second message** (e.g., "What is 5 * 33?")
6. Backend receives the conversation history and returns **400 Bad Request** from the LLM provider

### Root Cause

When CopilotKit replays the conversation history on the second turn, it sends the messages array to the AG-UI backend **without the tool result message**. The history looks like:

```
user: "What's the weather in Seattle?"
assistant: [tool_calls: [{id: "call_abc", function: {name: "get_weather", ...}}]]
assistant: "The weather in Seattle is 15°C and partly cloudy."
user: "What is 5 * 33?"
```

The correct history should be:

```
user: "What's the weather in Seattle?"
assistant: [tool_calls: [{id: "call_abc", function: {name: "get_weather", ...}}]]
tool: {tool_call_id: "call_abc", content: "Seattle: 15°C, partly cloudy"}
assistant: "The weather in Seattle is 15°C and partly cloudy."
user: "What is 5 * 33?"
```

The `tool` result message for `call_abc` is completely absent. This causes both OpenAI and Azure OpenAI to reject the request with a 400 error because `tool_calls` in the assistant message have no corresponding `tool` result.

### ✅ Expected Behavior

CopilotKit should persist `TOOL_CALL_RESULT` events in its internal conversation state and include the corresponding `tool` role messages when replaying history to the backend on subsequent turns.

### ❌ Actual Behavior

Tool result messages are dropped from the conversation state. On the second turn, the backend receives `assistant(tool_calls)` followed directly by `assistant(text)` with no `tool(result)` in between. The LLM provider (OpenAI/Azure OpenAI) rejects this with a 400 error.

Error from Azure OpenAI Responses API:
```
Code: agent_run_error_event
Message: An internal error has occurred while streaming events.
```

The underlying cause is a 400 from the LLM because the message history violates the constraint that every `tool_calls` entry must have a matching `tool` result message.

### Workaround

We subclass `AgentFrameworkAgent` on the backend to inject synthetic tool result dicts into the raw messages before they enter the library pipeline:

```python
class ToolGapFixAgent(AgentFrameworkAgent):
    async def run(self, input_data):
        messages = input_data.get("messages")
        if messages:
            input_data = {**input_data, "messages": fix_tool_gaps(messages)}
        async for event in super().run(input_data):
            yield event
```

Where `fix_tool_gaps` scans for `assistant` messages with `tool_calls` not followed by a `tool` result, and injects synthetic ones.

### Environment

```
@copilotkit/react-core: 1.55.3
@ag-ui/client: 0.0.52
agent-framework-ag-ui: 1.0.0b260311 (Python backend)
LLM provider: Azure OpenAI (Responses API)
Transport: AG-UI HttpAgent via agents__unsafe_dev_only prop
```

### Frontend Setup

```tsx
const agent = new HttpAgent({ url: "/ag-ui" });

<CopilotKit
  runtimeUrl="/ag-ui"
  agent="my-agent"
  agents__unsafe_dev_only={{ "my-agent": agent }}
  useSingleEndpoint={false}
>
  <CopilotChat />
</CopilotKit>
```

### Related Issues

- #2504 — `tool_use` ids without `tool_result` blocks (different root cause but same symptom)
- #3470 — tool calls lost in single turn dedup (UI-side, different bug)
- #3462 — sub-agent tool calls disappear from UI after completion

nebrass:
## Summary

The `StateManager` subscribes to several AG-UI agent events (`onNewMessage`, `onStateSnapshotEvent`, `onStateDeltaEvent`, etc.) but was **missing a handler for `onToolCallResultEvent`**.

When an AG-UI server emits `TOOL_CALL_RESULT` events for tool outputs, the tool-result message was never associated with the active run. On the next conversation turn, CopilotKit replays the assistant `tool_calls` message without the matching tool-result message, causing the LLM API to reject the request with a missing-tool-result error.

This makes **multi-turn conversations with tool calls fail** on the second turn.

## Changes

- **Import** `ToolCallResultEvent` from `@ag-ui/client`
- **Add** `onToolCallResultEvent` handler in the `subscribe()` block, following the same revocation-guard and `effectiveInput` pattern used by the other handlers
- **Add** `handleToolCallResult()` private method that reads `event.messageId` and calls `associateMessageWithRun()` — mirroring `handleNewMessage()`
- **Add 3 tests**: basic association, cross-run tracking, and mixed messages + tool results in a single run

## Test plan

- [x] All 340 existing tests pass (`pnpm --filter @copilotkit/core exec vitest run` — 34 files, 0 failures)
- [x] 3 new tests covering `ToolCallResultEvent` handling pass
- [x] Verified the `web-inspector` package already subscribes to `onToolCallResultEvent` (confirming the event is actively emitted by the AG-UI pipeline)

Closes #3884


BenTaylorDev:
Thanks for digging into #3884. I don't think this lands the fix, though, for a mechanism reason: the new `onToolCallResultEvent` handler calls `associateMessageWithRun`, which only writes to the `messageToRun` map — that's run-isolation bookkeeping (consumed by `getRunIdForMessage` and per-run cleanup), **not** `agent.messages`, which is the conversation history sent to the LLM that the issue is about. So associating the tool-result id with a run can't restore a missing tool message in history.

The history assembly is owned by `@ag-ui/client`, which already handles `TOOL_CALL_RESULT` — it splices a `role:"tool"` message into `messages` right after the matching assistant `toolCalls` message and then fires `onNewMessage` — so the result should already be in history via the existing path. The added tests fire only `onToolCallResultEvent` on a mock agent and never run the client's real apply pipeline, so they assert the run-association but not the LLM history the issue describes. It also currently hard-conflicts with `main`.

Before investing further, could you confirm #3884 still reproduces on current `main` — inspect `agent.messages` after a `TOOL_CALL_RESULT` on the second turn? If it does, the fix belongs in the client's message assembly (or how the run input is built from `agent.messages`), not StateManager's run map. Closing this PR for now (the issue can stay open) — happy to reopen with a current-`main` repro. Appreciate the investigation.

nebrass:
Thanks @BenTaylorDev. I tested the reproduction steps from #3884 again.

I used `@copilotkit/core@1.55.3` with `@ag-ui/client@0.0.55`. I tested the same two-turn flow over HTTP/SSE.

The second request contained:

`user → assistant(tool_calls) → tool(result) → assistant(text) → user`

The tool result was present and correctly ordered. The second turn completed without a 400 error. I also tested when `TOOL_CALL_RESULT` arrives after the assistant text. It worked too.

So I cannot reproduce #3884 with `@ag-ui/client@0.0.55`. The upstream fix seems to solve the issue.

You were right about this PR. The extra StateManager handler is not needed. Thanks for pointing me to the real message pipeline.

PR #2890, opened 2025-12-19
docs: remove broken Copilot Runtime concept link

omprakash0702:
## What does this PR do?

Removes a broken documentation link to the Copilot Runtime concept page, which no longer exists.

## Related PRs and Issues

- Fixes #2847

## Checklist

- [x] I have read the [Contribution Guide](https://github.com/copilotkit/copilotkit/blob/master/CONTRIBUTING.md)
- [ ] If the PR changes or adds functionality, I have updated the relevant documentation

## Documentation

* Updated CopilotRuntime reference documentation with clarified wording for improved readability.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #3889, opened 2026-04-14
chore: use latest langgraph and fix deployment

ranst91:


AlemTuzlak (approved):

botillar:
## Incident update: AKS MCP model-mismatch crash loop (prod/canary)

Same root cause as #2890 thread: this is a rollout skew between prompt/model config and runtime model registry.

- MCP selector/prompt path is using `gpt-5-nano-bulk`.
- Runtime in current canary/prod image asserts model support and throws `Unsupported model name` before serving, then pods crash-loop.

### Recovery sequence
1. Replace canary/prod image with a SHA where model-family registry and prompty config are in sync.
2. Validate canary pod readiness + startup logs and confirm no `Unsupported model name` assertion.
3. Run synthetic checks (`summary_missing`, `word_shared_chat_e2e`) to green.
4. Only then promote canary SHA to prod.

### Hardening
- Ensure prompty defaults and model-family `LLL_MODEL_DATA` changes are always shipped together in the same image.
- Add a simple smoke assertion in startup to catch unsupported model names before service registration and before alerting traffic.

I can split this into an explicit rollback + forward-hotfix checklist in a follow-up patch if you want to run it as a runbook.

PR #2105, opened 2025-07-04
feat (react-ui) : add audio format configuration options in push to talk

gtalha07:
## What does this PR do?

This PR addresses issue #2049 by implementing the ability to configure media type settings for the `usePushToTalk` hook. This enhancement provides developers with greater flexibility when implementing voice input functionality in their CopilotKit applications.

The implementation allows developers to specify media type configurations such as `audio/wav`.

## Related PRs and Issues

- #2049 

## Checklist

- [x] I have read the [Contribution Guide](https://github.com/copilotkit/copilotkit/blob/master/CONTRIBUTING.md)
- [x] If the PR changes or adds functionality, I have updated the relevant documentation

Issue #1852, opened 2025-05-22
📚 Documentation: typo on the intro blog: "canneries" -> "canaries" on https://www.copilotkit.ai/blog/introducing-ag-ui-the-protocol-where-agents-meet-users/


amikofalvy:
### 💬 Let us know how we can improve our documentation.

On this page https://www.copilotkit.ai/blog/introducing-ag-ui-the-protocol-where-agents-meet-users/

In section "Coding tools (Devin vs. Cursor)"

> Throughout the adoption of generative AI, coding tools have been **canneries** in the coal mine, and Cursor is the best example of a user-interactive agent. An AI agent that works alongside users, in a shared workspace.

should be 

> Throughout the adoption of generative AI, coding tools have been **canaries** in the coal mine, and Cursor is the best example of a user-interactive agent. An AI agent that works alongside users, in a shared workspace.

Definitions:
- canneries: factories where food is canned.
- canaries: a type of small bird (finch)

orca-copilotkit:
Hey @amikofalvy , Thanks for pointing it out, we will get that corrected!

AlemTuzlak:
fixed! thank you

PR #3895, opened 2026-04-14
fix(runtime): fall back to upstream agent on connect when in-memory store is empty

### LINKED ISSUE #3553: 🐛 Bug: InMemoryAgentRunner.connect() fails to restore sessions on serverless platforms (Vercel, Cloud Run)
## Summary

When using CopilotKit with a proxied AG-UI agent on serverless platforms (Vercel, Google Cloud Run), session restoration via `threadId` fails intermittently because `InMemoryAgentRunner` stores thread events in an in-process `GLOBAL_STORE` (`Map` on `globalThis`) that is lost on cold starts or when requests hit different instances.

**Bugfix / Feature request:** Allow `InMemoryAgentRunner.connect()` to fall back to the upstream agent's `/connect` endpoint (or a configurable middleware API) when a `threadId` is not found in the in-memory store, so the backend agent can provide the conversation history.

## Environment

- **CopilotKit**: `@copilotkit/runtime` ^1.54.0 (resolved 1.54.1), `@copilotkit/react-core` ^1.54.0
- **AG-UI Client**: `@ag-ui/client` ^0.0.48
- **Framework**: Next.js 16 (App Router)
- **Deployment**: Vercel (serverless) / Google Cloud Run
- **Backend agent**: Python (Google ADK), connected via `HttpAgent` / `AGENT_URL`

## Architecture

```
Browser (CopilotKit React)
  → POST /api/copilotkit (Next.js API route)
    → CopilotRuntime (with HttpAgent pointing to Python backend)
      → Python ADK agent at AGENT_URL
```
### More info about backend environment (Python Google ADK part) for this issue described in https://github.com/ag-ui-protocol/ag-ui/issues/1243 

API route setup:

```ts
  agents: {
    default: new HttpAgent({
      url: process.env.AGENT_URL || 'http://localhost:8000/',
    }),
  },
})
```

Session restoration is initiated client-side by passing `threadId` as a URL param:

```tsx
const threadId = new URLSearchParams(globalThis.location?.search)
  .get('sessionId') || undefined

<CopilotKit key={threadId} runtimeUrl="/api/copilotkit" threadId={threadId}>
```

## Impact

This affects **any CopilotKit deployment** using the proxied agent pattern on serverless platforms — which is likely the most common production deployment model. The "Share Conversation URL" feature (a key UX feature in our client's project deployed to production) becomes unreliable, and users cannot depend on session continuity across page reloads or shared links.

## Related

- `InMemoryAgentRunner` and `GLOBAL_STORE` in `@copilotkit/runtime`
- `ProxiedCopilotRuntimeAgent.connect()` flow
- AG-UI protocol's `/connect` endpoint specification

### ✅ Expected Behavior

When `InMemoryAgentRunner.connect()` does not find a `threadId` in `GLOBAL_STORE`, it should **fall back** to the upstream agent (or a configurable middleware endpoint) to attempt session restoration from the backend's persistent storage.

## Proposed Solution: Fallback to upstream agent's connect endpoint

When the thread is not found in `GLOBAL_STORE`, `InMemoryAgentRunner` (or `ProxiedCopilotRuntimeAgent`) should forward the `connect` request to the configured `HttpAgent`'s `/connect` endpoint, similar to how `run()` already proxies to the agent's `/run` endpoint.

Pseudocode:

```ts
class InMemoryAgentRunner {
  async connect({ threadId, agent }) {
    // 1. Try in-memory first (current behavior)
    const stored = GLOBAL_STORE.get(threadId)
    if (stored) {
      return replayEvents(stored)
    }

    // 2. NEW: Fall back to upstream agent's connect endpoint
    if (agent?.connect) {
      return agent.connect({ threadId })
    }

    // 3. Return empty if no fallback available
    return emptyStream()
  }
}
```

## Current Behavior

1. User shares a conversation URL containing `?sessionId=<threadId>`
2. Recipient opens the URL; CopilotKit sends an `agent/connect` request with the `threadId`
3. `CopilotRuntime` handles the connect request via `runtime.runner.connect({ threadId })`
4. `InMemoryAgentRunner.connect()` looks up `threadId` in `GLOBAL_STORE` (an in-memory `Map` stored on `globalThis`)
5. **On cold start or different instance:** `GLOBAL_STORE` is empty → no events returned → session appears blank
6. **On warm instance (same process):** `GLOBAL_STORE` has the data → session restores correctly

This makes session restoration **intermittent and unreliable** on any serverless/multi-instance deployment.

## Root Cause

The `connect()` flow in `InMemoryAgentRunner` is entirely self-contained — it only checks the in-process memory store. It **never reaches the upstream agent** (the Python backend at `AGENT_URL`), even though:

- The Python ADK agent has its own thread/session persistence
- The `run()` flow correctly proxies to the backend agent
- The backend agent's AG-UI `/connect` endpoint could serve the conversation history

The asymmetry is:
| Operation | Reaches Python backend? | Serverless-safe? |
|-----------|------------------------|-----------------|
| `run()` (send message) | ✅ Yes, via `HttpAgent` | ✅ Yes |
| `connect()` (restore session) | ❌ No, in-memory only | ❌ No |

### Screenshots

_No response_

### 𝌚 CopilotKit Version

```shell
├── @copilotkit/react-core@1.54.1 -> ./node_modules/.pnpm/@copilotkit+react-core@1.54.1_@ag-ui+core@0.0.48_@types+mdast@4.0.4_@types+react-dom@19_c70e6720046be18078d01f4dc322c726/node_modules/@copilotkit/react-core
├── @copilotkit/react-ui@1.54.1 -> ./node_modules/.pnpm/@copilotkit+react-ui@1.54.1_@ag-ui+core@0.0.48_@types+mdast@4.0.4_@types+react-dom@19.2_87473fd4311520c6bffa467f72ff0ad2/node_modules/@copilotkit/react-ui
├── @copilotkit/runtime@1.54.1 -> ./node_modules/.pnpm/@copilotkit+runtime@1.54.1_@ag-ui+encoder@0.0.48_@cfworker+json-schema@4.1.1_@copilotki_111661355162fbdc80297afa697d54d2/node_modules/@copilotkit/runtime
```

### 📄 Logs (Optional)

```shell

```

AlemTuzlak:
## Summary

Closes #3553

- On serverless platforms (Vercel, Cloud Run), `InMemoryAgentRunner.connect()` fails to restore sessions because `GLOBAL_STORE` is lost on cold starts or when requests hit different instances
- The `connect()` flow never reached the upstream agent, unlike `run()` which correctly proxies through `HttpAgent`
- Added an optional `agent` field to `AgentRunnerConnectRequest` and delegate to the agent's `connect()` method when the in-memory store has no entry for the requested `threadId`
- When the in-memory store exists for the thread, it is still preferred over the upstream fallback

## Test plan

- [x] New test: falls back to agent `connect()` when `threadId` is not in `GLOBAL_STORE`
- [x] New test: prefers in-memory store over agent fallback when store exists
- [x] New test: returns empty when no store and no agent provided
- [x] All 1232 existing runtime tests pass (0 regressions)
- [x] SQLite runner tests pass (15/15)
- [x] Runtime build succeeds


MarkoIteratec:
Hi there! 
we are currently facing the same issue and would appreciate this fix! Any option so speed up the review process to get this fix soon?
best regards
Marko

BenTaylorDev (changes requested):
Thanks for this — the bug is real and still live on `main`: `InMemoryAgentRunner.connect()` returns empty when the store is cold (the serverless cold-start case in #3553). But the fallback mechanism as written would crash that exact path, so it needs rework:

- The fallback calls the agent's **raw protected `connect()`** (`request.agent as unknown as { connect(...) }).connect(input)`). Base `AbstractAgent.connect()` throws `AGUIConnectNotImplementedError`, and `HttpAgent` — the agent used for HTTP/Python/serverless backends — **doesn't override `connect()`**. So for the real upstream agent on an empty store, this throws instead of restoring the session — turning today's benign empty-replay into a cold-start crash. The graceful entry point is `connectAgent()` (or explicitly catching `AGUIConnectNotImplementedError` → `EMPTY`), which the base runtime already uses.
- The tests pass only because they introduce a bespoke `ConnectableAgent` that overrides the protected `connect()` to emit pre-baked events — no shipped agent does that, so the suite validates a path production never takes.
- It also currently hard-conflicts across several files; the `agent`→`handleSseConnect`→`runner.connect` plumbing has since landed separately (the #5712 header-forwarding rework), so a re-target would only need the `AgentRunnerConnectRequest` field + a corrected `in-memory.ts` fallback.

Please re-author against current `main` so the fallback goes through `connectAgent()` (or catches `AGUIConnectNotImplementedError`), and test it against a real `HttpAgent`, not a `connect()`-implementing subclass.

BenTaylorDev (changes requested):
Sorry for the slow turnaround on the rework — you pushed this on Aug 12 and it sat.

The crash risk from my last review is genuinely gone. Routing through `connectAgent()` is the right entry point: it catches `AGUIConnectNotImplementedError` and falls through to `EMPTY`, so an agent without `connect()` no longer throws. Using the per-request clone for `agent.threadId` is correct too.

Two things still blocking:

**1. `check-types` is failing on a real error in the diff.** `packages/runtime/src/v2/runtime/handlers/sse/connect.ts` has `agent,` twice in the same object literal (lines 50 and 72) → `TS1117`. Looks like a leftover from the Aug 12 merge resolution.

**2. The fallback is a no-op for the configuration in #3553.** `HttpAgent` doesn't override `connect()`, so `connectAgent()` hits the base implementation, throws `AGUIConnectNotImplementedError`, and resolves with zero events — no request ever reaches the upstream. Confirmed against the pinned `@ag-ui/client@0.0.57`: `connectAgent({}, …)` on an `HttpAgent` resolves in ~1ms with `events: []` and no network call. Same for every agent that can be registered on the SSE connect path — `BuiltInAgent`, `SanitizingHttpAgent`, `@ag-ui/langgraph`, `@ag-ui/a2a`. (The one class in the repo that does implement `connect()` is `ProxiedCopilotRuntimeAgent` in `@copilotkit/core`, but that's the client-side agent talking *to* the runtime — it's not what sits in `CopilotRuntime({ agents })`.)

The new test says as much: `gracefully falls back to an HttpAgent when no store exists` asserts `expect(events).toEqual([])`. Good safety test, but it documents that nothing is restored. The reporter's setup is exactly `new HttpAgent({ url: AGENT_URL })`, and the ask is to forward the connect request to the upstream `/connect` endpoint the way `run()` proxies to `/run`.

Closing #3553 needs an `HttpAgent`-level `connect()` that POSTs to that endpoint (the ag-ui-side work in ag-ui-protocol/ag-ui#1243) — `ProxiedCopilotRuntimeAgent.connect()` is a decent reference for the shape. If that's out of scope here, the alternative is a CopilotKit-side connecting agent the runtime can use. Happy to help scope either direction.

Smaller notes:
- `from` added to the rxjs import in `in-memory-runner.test.ts:26` is unused (lint doesn't catch it, just tidiness).
- Fallback events aren't written back into `sharedStore`, so a later `connect()` re-fetches and a later `run()` still starts cold — worth a comment if intentional.
- Upstream errors now error the SSE stream rather than completing empty. Probably what you want, but it is a change.

NathanTarbert:
Hey @AlemTuzlak,

Quick update: we're working through this now. Looking at the layer below turned it into an architectural decision rather than a patch, so here's what we found.

Your rework is right where it counts — routing through `connectAgent()` instead of the protected `connect()` is the correct entry point, and using the per-request clone for `agent.threadId` is right too.

What's blocking it isn't in this PR. As @BenTaylorDev found, no agent in `@ag-ui/client` implements `connect()`: `HttpAgent` overrides only `runAgent`, so `connectAgent()` reaches the base implementation, throws `AGUIConnectNotImplementedError`, and resolves empty. Your own test records that honestly with `expect(events).toEqual([])`. The fallback is correct code that currently has nothing to fall back to.

Adding that missing piece is where it gets architectural:

- An AG-UI server is documented as a single `POST /` taking `RunAgentInput`, and `HttpAgent` posts to its configured URL with no path suffix, so there's no route namespace a connect endpoint could hang off without inventing a convention.
- Thread-history replay and mid-run stream resumption are two different features sharing one `connect()` seam today.
- Replay streams also have to get past `verifyEvents`, which enforces single-run lifecycle rules. CopilotKit already carries a fork for exactly this — `packages/core/src/utils/connect-replay.ts` is `connectAgent` minus the verifier, with a TODO asking AG-UI for a supported opt-out.

The design thread is ag-ui-protocol/ag-ui#2105 and that's where it gets settled.

One correction worth having on the record: the review points at ag-ui-protocol/ag-ui#1243 as the upstream work, but that issue is "[Feature]: Allow passing threadId as the ADK session_id" and closed in March via #1383 and #1271 — same reporter, adjacent symptom, different layer. The related-but-closed one is #1852, and its live successor is #2105.

Two small things from the review that are independent of all this, whenever you're next in here: `connect.ts` has `agent,` twice in the same object literal at lines 50 and 72, which is what's failing `check-types` (TS1117, looks like Aug 12 merge fallout), and the `from` added to the rxjs import in `in-memory-runner.test.ts:26` is unused.

Happy to keep this as the CopilotKit half and land it once the client side can answer. Your call — it's your PR.

PR #3909, opened 2026-04-14
fix: align starter Dockerfiles and entrypoints with working demo packages

jpr5:
## Summary
- Python starters (non-langgraph) now include `agent_server.py` copied from demo packages, with imports rewritten from `agents.*` to `agent.*` for the starter directory layout
- All Python uvicorn starters use `agent_server:app` as the uvicorn module instead of directly referencing framework agent modules that don't export a FastAPI `app` object
- langgraph-fastapi starter now uses `langgraph_cli dev` (matching its demo) instead of uvicorn, with `langgraph.json` properly included
- Dockerfile.python template conditionally includes `COPY agent_server.py` for non-langgraph frameworks

## Test plan
- [x] All 537 showcase tests pass (248 generate-starters + 270 starter-consistency + others)
- [x] Verified `agent_server.py` exists in all 10 non-langgraph Python starters
- [x] Verified langgraph starters correctly omit `agent_server.py`
- [x] Verified import rewrites (`from agents.` -> `from agent.`) in all copied `agent_server.py` files
- [x] Verified langgraph-fastapi now has `agent/langgraph.json` and uses `langgraph_cli dev`

### DIFF
--- a/.github/workflows/showcase_deploy.yml
+++ b/.github/workflows/showcase_deploy.yml
@@ -53,8 +53,8 @@ on:
           - shell-dojolike
 
 concurrency:
-  group: showcase-deploy-${{ github.ref }}
-  cancel-in-progress: true
+  group: showcase-deploy-${{ github.ref }}-${{ github.event.inputs.service || 'auto' }}
+  cancel-in-progress: ${{ github.event_name == 'push' }}
 
 env:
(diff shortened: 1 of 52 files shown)


jpr5:
Superseded — will reapply on current main

PR #3912, opened 2026-04-14
fix: clean up Slack alert formatting in showcase workflows

jpr5:
## Summary

- **drift-detection**: Multiline failure details were interpolated directly into JSON payload strings, causing `%0A` literals and broken JSON when playwright output contained newlines. Now uses `jq` to build properly escaped JSON and `payload-file-path` to deliver it.
- **starter-smoke**: `toJSON(format(...))` double-encoded multiline content. Replaced with `jq --arg` payload builder. Also truncated error capture to 3 lines / 300 chars and added ANSI code + path stripping at the extraction step.
- **deploy**: Long inline GitHub Actions ternary expression was fragile and unreadable. Replaced with shell if/elif logic and `jq` payload builder.

All three workflows now follow the same safe pattern: extract summary (truncated, cleaned) -> build JSON via `jq --arg` (handles all escaping) -> write to file -> use `payload-file-path`.

The remaining showcase workflows (docs-sync, qa-sync, validate, smoke-monitor) were already using safe patterns (simple one-line strings or `jq --rawfile`).

## Test plan

- [ ] Trigger `showcase_drift-detection.yml` manually and verify Slack message renders cleanly
- [ ] Trigger `starter-smoke.yml` manually and verify failure alerts are readable
- [ ] Trigger `showcase_deploy.yml` manually and verify deploy notification is correct

### DIFF
--- a/.github/workflows/showcase_deploy.yml
+++ b/.github/workflows/showcase_deploy.yml
@@ -280,12 +280,29 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 2
     steps:
-      - name: Collect results and notify Slack
+      - name: Build Slack payload
+        run: |
+          DETECT_RESULT="${{ needs.detect-changes.result }}"
+          LOCKFILE_RESULT="${{ needs.check-lockfile.result }}"
+          BUILD_RESULT="${{ needs.build.result }}"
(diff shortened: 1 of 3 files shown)


jpr5:
Superseded — reapplying on current main

PR #3926, opened 2026-04-15
docs: fix useCopilotContext import in thread persistence guides

Protocol-zero-0:
## Summary
This fixes the `useCopilotContext` import path in the LangGraph and CrewAI Flows thread persistence guides. The docs currently reference `@copilotkit/react-core/v2`, but `useCopilotContext` is exported from `@copilotkit/react-core`.

I also synced the showcase doc copies so the examples stay aligned with the main docs source.

## Validation
I verified that the stale `@copilotkit/react-core/v2` import no longer appears in these guides.


BenTaylorDev:
Closing as moot — the buggy `@copilotkit/react-core/v2` import this PR targets no longer exists in the live `loading-message-history` pages (that code block was restructured and the import removed), so the specific fix no longer applies. The docs now live in `showcase/shell-docs/src/content/` (the top-level `docs/` path this PR targets is a retired symlink). Thanks @Protocol-zero-0!

PR #3947, opened 2026-04-16
fix(adk): add AGUIToolset to agent tools so client tools are available

marthakelly:
## Summary

- `ag_ui_adk` v0.5+ requires `AGUIToolset()` to be explicitly in an agent's tools list — it's replaced at runtime with a `ClientProxyToolset` that proxies frontend tools (`useFrontendTool`, `useCopilotAction`) from the browser
- Both ADK examples were added to the repo after this became required and were never updated, so client tools silently did nothing when triggered from the UI
- Adds `AGUIToolset()` to `examples/integrations/adk` (fixes `setThemeColor`) and `examples/showcases/adk-dashboard` (fixes chart/search tools)

## Why no regression test

AIMock is not properly wired to the Gemini API in the ADK `docker-compose.test.yml` — the `@chat` smoke test currently passes as a false positive (agent throws `RUN_ERROR`, test reads the catch-all error text). A proper regression test requires fixing the AIMock routing first, tracked in [CPK-7399](https://linear.app/copilotkit/issue/CPK-7399/adk-smoke-tests-pass-as-false-positive-aimock-not-wired-to-gemini-api).

## Test plan

- [ ] Run `examples/integrations/adk` locally, ask agent to "Set the theme to green" — background color should change
- [ ] Run `examples/showcases/adk-dashboard` locally, ask agent to show a chart — chart component should render

### DIFF
--- a/examples/integrations/adk/agent/main.py
+++ b/examples/integrations/adk/agent/main.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 import json
 from typing import Dict, Optional
 
-from ag_ui_adk import ADKAgent, add_adk_fastapi_endpoint
+from ag_ui_adk import ADKAgent, AGUIToolset, add_adk_fastapi_endpoint
 from dotenv import load_dotenv
 from fastapi import FastAPI
 from google.adk.agents import LlmAgent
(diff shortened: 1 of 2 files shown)


marthakelly:
Closing — this encodes the workaround as intended usage. The correct fix is auto-injection in ag_ui_adk upstream so frontend tools work without any manual AGUIToolset() step. Will file a PR to ag-ui-protocol/ag-ui instead.

marthakelly:
closing, agent miscommunication. will follow up with proper fix upstream

PR #3992, opened 2026-04-16
chore(ci): rename starter_deployed_smoke workflow to test_smoke-starter-deployed

jpr5:
Part of workflow naming standardization Bundle 6. Internal name goes from 'Starter Deployed Smoke Tests' to 'test / smoke / starter-deployed'. Also fixes the dash/underscore inconsistency with its sibling starter-smoke.yml (renamed separately by N9). Branch-protection audit confirmed drop-in safe. Consumer grep found zero hits for the old name.


jpr5:

PR #3998, opened 2026-04-16
chore(ci): rename test_runtime-servers → test_integration-runtime

jpr5:
Bundle 6 workflow rename. BOTH filename AND internal name change: 'test / integration' → 'test / integration / runtime'. Consumer grep verified no remaining references to either old name. Branch-protection audit confirmed drop-in safe ([Notion](https://www.notion.so/3443aa38185281a3ad07e5ec63debd0f)).


jpr5:

PR #4005, opened 2026-04-16
chore(ci): rename e2e_dojo.yml → test_e2e-dojo.yml (filename only)

jpr5:
Bundle 6 workflow rename. Filename-only change — internal name 'test / e2e / dojo' unchanged. Lowest-risk rename since check-name is stable. Brings filename in line with test_<layer>-<target> convention.


jpr5:

PR #4010, opened 2026-04-16
docs(showcase): multi-frontend consolidation strategy

jpr5:
## Summary

- Adds `showcase/FRONTEND-STRATEGY.md` documenting why multiple shells exist, what each one replaces, the rollout order, and how per-package shell selection works today (spoiler: there is no explicit mechanism — every shell embeds every package via iframe).
- Calls out `shell-dojolike/` as the **intended ag-ui Dojo replacement**, not cleanup waste. This came from user correction during the QA/E2E audit (J5 in the action inventory), where the alternate shell was nearly mislabeled as abandoned.
- Enumerates open questions (DNS cutover, shared vs. per-shell content, visibility filtering) that need decisions before the next consolidation target.

## Context

During the QA/E2E audit we flagged `showcase/shell-dojolike/` as unclear — possibly abandoned, possibly a spike. The user corrected this: it is the active replacement target for the external ag-ui Dojo site, and the broader strategy is to consolidate multiple external frontends into this repo one at a time. This doc captures that strategy so the next person (or agent) doesn't make the same mistake.

## Test plan

- [ ] Verify claims in the doc match reality:
  - [ ] Both shells consume the same `registry.json`
  - [ ] Both shells embed via iframe pointing at `integration.backend_url + demo.route`
  - [ ] `.github/workflows/showcase_deploy.yml` builds both shells independently
  - [ ] Per-package E2E runs against the package dev server, not a shell
- [ ] No source-code changes — docs-only PR, CI should pass unchanged


jpr5:
Consolidated into https://github.com/CopilotKit/CopilotKit/pull/4021 (Bundle 5).

PR #4015, opened 2026-04-16
docs(showcase/aimock): document fixture sync policy

jpr5:
## Summary

Adds `showcase/aimock/README.md` — the missing operator doc for the showcase aimock fixture server. Action inventory reference: **J6**.

Covers:
- **What aimock is** — deterministic LLM fixture server, deployed as a Railway sidecar in `--proxy-only` mode so unmatched requests fall through to the real provider.
- **Fixture layout** — `feature-parity.json` (per-package `test_e2e-showcase-on-demand` fixtures for the 9 demos across 17 packages) vs `smoke.json` (minimal `/api/smoke` round-trip fixture).
- **Sync policy** — **fixtures are hand-maintained.** No automated capture, no scheduled re-recording, no behavioral drift detection. Two load-time safety nets (Dockerfile `--validate-on-load` from #3973, CI `aimock-fixtures.test.ts` schema check) catch schema breakage but not semantic drift.
- **Add/update workflow** — manual; includes local validation command and the per-package E2E runner script.
- **Drift risk** — drift surfaces as flaky E2E failures, not as a dedicated signal. Doc enumerates symptoms + fix paths.
- **Related workflows** — `test_e2e-showcase-on-demand.yml` (formerly `showcase_aimock-e2e.yml`), `showcase_validate.yml`, `showcase_deploy.yml`, `showcase_smoke-monitor.yml`.

## Key finding

The sync policy is **not automated**. When a package's agent changes what it prompts the LLM with (new tool name, new prompt prefix), nothing in CI notices that the existing `feature-parity.json` fixture now matches the wrong request or returns a now-inappropriate response. The only signal is a downstream Playwright assertion failure, and a human has to trace the failure back to the fixture. Surfacing this explicitly is the main value of the doc; if we want real drift detection, the path is to wire aimock's `--record` mode into a periodic workflow that re-captures against live providers and diffs — but that is not built today.

## Test plan

- [x] Local: `pnpm --filter @copilotkit/showcase-scripts test aimock-fixtures` passes (no fixture changes)
- [x] Doc renders cleanly on GitHub
- [ ] Reviewer spot-checks that the described policy matches reality (no hidden automation missed)

### DIFF
--- a/showcase/aimock/README.md
+++ b/showcase/aimock/README.md
@@ -1,0 +1,66 @@
+# Showcase aimock
+
+Deterministic LLM fixture server for showcase E2E testing. Replaces real LLM API calls (OpenAI, Anthropic, Gemini) with pre-recorded responses so Playwright tests can run PR-gated in CI without API keys and without rate limits or non-determinism.
+
+The image built from this directory is [`ghcr.io/copilotkit/showcase-aimock`](https://github.com/orgs/CopilotKit/packages/container/package/showcase-aimock) and is deployed to Railway as a sidecar alongside the showcase packages.
+
+## What aimock is
+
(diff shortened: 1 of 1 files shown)


jpr5:
Consolidated into https://github.com/CopilotKit/CopilotKit/pull/4021 (Bundle 5).

PR #4034, opened 2026-04-17
fix(react-core): stabilize auto-generated threadId across remounts

restareaByWeezy:
Fixes #3628

When `threadId` isn't passed to `CopilotSidebar` / `CopilotChat`, the provider generates a fallback ID with `randomUUID()` inside `useMemo`. Problem is `useMemo` doesn't guarantee stability across remounts — so every page navigation or parent key change creates a brand new thread, silently dropping conversation history.

The fix moves the fallback into a `useRef` so the auto-generated ID survives for the full lifetime of the provider instance. When `threadId` is passed explicitly, nothing changes — it still wins.

```diff
+  const autoThreadId = useRef<string | null>(null);
+  if (!autoThreadId.current) {
+    autoThreadId.current = randomUUID();
+  }
+
   const resolvedThreadId = useMemo(() => {
     if (threadId) return threadId;
     if (parentConfig?.threadId) return parentConfig.threadId;
-    return randomUUID();
+    return autoThreadId.current!;
   }, [threadId, parentConfig?.threadId]);
```

One file, 6 lines added, 1 removed. All 1095 tests in `@copilotkit/react-core` pass.

### DIFF
--- a/packages/react-core/src/v2/providers/CopilotChatConfigurationProvider.tsx
+++ b/packages/react-core/src/v2/providers/CopilotChatConfigurationProvider.tsx
@@ -82,6 +82,11 @@ export const CopilotChatConfigurationProvider: React.FC<
 
   const resolvedAgentId = agentId ?? parentConfig?.agentId ?? DEFAULT_AGENT_ID;
 
+  const autoThreadId = useRef<string | null>(null);
+  if (!autoThreadId.current) {
+    autoThreadId.current = randomUUID();
+  }
+
   const resolvedThreadId = useMemo(() => {
     if (threadId) {
       return threadId;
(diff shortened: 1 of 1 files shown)


BenTaylorDev:
Thanks for this, and for the clear diagnosis of #3628. Since it was opened, `CopilotChatConfigurationProvider` has been substantially restructured and the auto-generated threadId is now stabilized through the current resolution order (`threadIdPropIsAuthoritative` → parent/override → last-resort mint) plus explicit-threadId propagation onto the agent (#5041), with `threadid-propagation.contract.test.tsx` and `CopilotChat.absentThreadConnect.test.tsx` covering the absent-thread case.

As a result this patch no longer applies — it hard-conflicts on exactly the `resolvedThreadId` block it rewrites, which has since been replaced. One note on the mechanism too: swapping `useMemo` for `useRef` guards against re-render, but neither survives a true unmount/remount (a remount is a fresh instance), so the current `useState`-seed-in-a-persistent-provider design is what actually addresses the remount scenario.

Closing as superseded — please reopen if you can still reproduce a lost thread on latest `main` (mount `CopilotSidebar` without a `threadId` under a persistent `<CopilotKit>`, then remount via navigation). Thanks again for surfacing it.

PR #4036, opened 2026-04-17
ci: add workflow to update PR branches via qa:update-branch label

marthakelly:
## Summary
- Adds `.github/workflows/update-branch.yml` — a labeled-event workflow that updates a PR's branch with its base when the `qa:update-branch` label is applied.
- Removes the label after running so the bot can re-apply it on the next cycle.
- Infrastructure-only change; enables a QA bot to keep PRs current without human intervention.

## Behavior
1. Bot adds `qa:update-branch` label to any PR.
2. Workflow calls the GitHub `updateBranch` API (same as the "Update branch" button).
3. Workflow removes the label (in a `finally` block, so it clears even if the update fails).

Uses the namespaced `qa:` prefix to avoid collisions and leave room for future bot actions (`qa:triaged`, etc.).

## Test plan
- [ ] Merge this PR.
- [ ] On a separate out-of-date PR, apply the `qa:update-branch` label and confirm the workflow runs, updates the branch, and removes the label.
- [ ] Confirm the workflow does nothing when unrelated labels are added.

### DIFF
--- a/.github/workflows/update-branch.yml
+++ b/.github/workflows/update-branch.yml
@@ -1,0 +1,40 @@
+name: Update PR branch
+
+on:
+  pull_request:
+    types: [labeled]
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  update-branch:
+    if: github.event.label.name == 'qa:update-branch'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Update PR branch with base
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const { owner, repo } = context.repo;
+            const pull_number = context.payload.pull_request.number;
+
+            try {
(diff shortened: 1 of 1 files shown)


MikeRyanDev:
Important: this is wired to `pull_request`, which means PRs opened from forks will run with a read-only `GITHUB_TOKEN` even if the workflow requests write permissions. In that case both `pulls.updateBranch` and the label removal can fail, so the QA bot won't actually be able to keep external-contributor PRs current.

Because this job doesn't check out or execute PR code, `pull_request_target` is the safer fit here if the intent is to mutate PR state from labels across all PRs, including forks.

MikeRyanDev (approved):
Reviewed the workflow update. The `pull_request_target` switch addresses the fork-permission issue, and the 422 handling looks appropriate for no-op update cases.

PR #4037, opened 2026-04-17
ci: add workflows: write permission to update-branch workflow

marthakelly:
## Summary
- Adds `workflows: write` to the `.github/workflows/update-branch.yml` permissions block.
- Follow-up to #4036.

## Why
Without `workflows: write`, the labeled update-branch action fails on any PR whose base-merge would touch files under `.github/workflows/**` — GitHub rejects the merge with a "refusing to allow a GitHub App to create or update workflow" error. Adding the permission is harmless for PRs that don't touch workflow files and unblocks the QA bot for PRs that do.

## Test plan
- [ ] After merge: apply the \`qa:update-branch\` label to a PR that has workflow-file changes relative to base, confirm the workflow runs, updates the branch, and removes the label.
- [ ] Confirm unrelated PRs still get updated as before.

### DIFF
--- a/.github/workflows/update-branch.yml
+++ b/.github/workflows/update-branch.yml
@@ -7,6 +7,7 @@ on:
 permissions:
   contents: write
   pull-requests: write
+  workflows: write
 
 jobs:
   update-branch:


MikeRyanDev:
Important: `workflows: write` does not appear to be a valid `permissions` key for `GITHUB_TOKEN` in GitHub Actions workflow syntax. The current docs list keys like `actions`, `contents`, `pull-requests`, etc., but not `workflows`.

If that's right, this won't grant the intended capability and may make the workflow invalid instead of fixing the merge path for PRs that touch `.github/workflows/**`. I'd verify against the Actions permission docs and, if this capability really isn't available to `GITHUB_TOKEN`, switch to a GitHub App/PAT-based token for the `updateBranch` call rather than extending the workflow-level `permissions` block.

Issue #3644, opened 2026-04-07
🐛 Bug: Duplicate assistant message IDs when multiple tool calls are interleaved with tool results


caibirdme:
### ♻️ Reproduction Steps

  When using CopilotKit with an AG-UI event stream that contains multiple `TOOL_CALL_START` events for the same assistant message, interleaved with `TOOL_CALL_RESULT` events, the
  frontend can end up with duplicate assistant message IDs in `messages[]`.

  In development mode, this triggers:

  ```txt
  CopilotChatMessageView: Deduplicated 1 message(s) with duplicate IDs.
```

This seems to happen with a sequence like this:

  1. An assistant text message is created with messageId = msg_x
  2. TOOL_CALL_START(tool_a, parentMessageId = msg_x)
  3. TOOL_CALL_START(tool_b, parentMessageId = msg_x)
  4. TOOL_CALL_RESULT(messageId = tool-result:msg_x:tool_a, toolCallId = tool_a)
  5. TOOL_CALL_START(tool_c, parentMessageId = msg_x)

### ✅ Expected Behavior

tool_c should be appended to the existing assistant message msg_x

### ❌ Actual Behavior

  - a new assistant message is created with the same id msg_x
  - now there are two assistant messages with the same id
 
  CopilotChatMessageView deduplicates messages by message.id:
  ```const deduplicatedMessages = [...new Map(messages.map((m) => [m.id, m])).values()];```

 So once duplicate IDs appear, one of the messages is dropped from rendering. In my case, the later duplicate assistant message had empty content and only the last tool call, so deduplication effectively hid the earlier assistant content/tool calls.

### 𝌚 CopilotKit Version

```shell
- CopilotKit React: 1.54.1
  - @ag-ui/client: 0.0.48
```

### 📄 Logs (Optional)

```shell
The issue appears to come from the event-application logic for TOOL_CALL_START.
The current logic seems to only attach the tool call to the parent assistant message if the last message in the array has `id === parentMessageId`.
That works until a TOOL_CALL_RESULT is appended as a separate role: "tool" message. After that, the last message is no longer the parent assistant message, so a later TOOL_CALL_START for the same parent creates a new assistant message with the same ID instead of merging into the existing one.

  In other words, the logic is too dependent on "last message" position and does not search for the existing assistant message by parentMessageId.
```

BenTaylorDev:
[...]

PR #4058, opened 2026-04-18
fix(react-core): deduplicate tool calls in CopilotChatToolCallsView to prevent React key warning

### LINKED ISSUE #3928: 🐛 Bug: `CopilotChatToolCallsView` produces React duplicate-key warning when `connect()` replays historic tool call events over pre-populated message state
### ♻️ Reproduction Steps

1. Use CopilotKit v2 with `InMemoryAgentRunner` and register at least one frontend tool via `useFrontendTool`
2. Send a message that triggers a tool call and let the run complete (the tool call should appear in the chat UI)
3. Reload the page — or otherwise cause `connectAgent()` to be called on an agent whose `messages` state already contains the completed assistant message with the tool call
4. Open the browser DevTools console

### ✅ Expected Behavior

No console warnings. Each tool call should appear exactly once in the chat.

### ❌ Actual Behavior

The browser console shows:

```
Warning: Encountered two children with the same key, `call_uUF5OZtugHE6tyoC9PPtpAs6`.
Keys should be unique so that components maintain their identity across updates.
Non-unique keys may cause children to be duplicated and/or omitted —
the behavior is unsupported and could change in a future version.
    at CopilotChatToolCallsView
    at MemoizedSlotWrapper
    at CopilotChatAssistantMessage
```

**Root cause:** When `connectAgent()` is called, `InMemoryAgentRunner.connect()` replays compacted historic events (including `TOOL_CALL_START` for completed tool calls) via `@ag-ui/client`'s `defaultApplyEvents`. However, at that point the agent's `messages` state already contains those tool calls from the prior run. Because `defaultApplyEvents` pushes to `message.toolCalls` unconditionally — with no duplicate guard — the same `toolCallId` ends up in the array twice:

```ts
message.toolCalls = [
  { id: "call_abc123", ... },  // ← already in initial state
  { id: "call_abc123", ... },  // ← pushed again by replayed TOOL_CALL_START
]
```

`CopilotChatToolCallsView` then maps this array using `toolCall.id` as a React `key`, producing the duplicate-key warning.

The existing `deduplicateMessages()` in `CopilotChatMessageView` only deduplicates at the message level; it does not dedup entries within each message's `toolCalls` array.

**Fix suggestion:** In `CopilotChatToolCallsView`, deduplicate `message.toolCalls` by `id` before rendering:

```ts
const uniqueToolCalls = [
  ...new Map(message.toolCalls.map((tc) => [tc.id, tc])).values(),
];
```

Or equivalently, extend `deduplicateMessages()` to also dedup within `toolCalls` arrays.

Note: the underlying idempotency issue in `@ag-ui/client`'s `defaultApplyEvents` has been separately reported to [ag-ui-protocol/ag-ui](https://github.com/ag-ui-protocol/ag-ui).

### Screenshots

_No response_

### 𝌚 CopilotKit Version

```shell
❯ pnpm ls | grep "@copilotkit"
├── @copilotkit/a2ui-renderer@1.55.3
├── @copilotkit/react-core@1.55.3
```

### 📄 Logs (Optional)

```shell

```

octo-patch:
Fixes #3928

## Problem

When `connectAgent()` is called, `InMemoryAgentRunner.connect()` replays compacted historic events (including `TOOL_CALL_START` for completed tool calls) via `@ag-ui/client`'s `defaultApplyEvents`. However, at that point the agent's `messages` state already contains those tool calls from the prior run. Because `defaultApplyEvents` pushes to `message.toolCalls` unconditionally — with no duplicate guard — the same `toolCallId` ends up in the array twice.

`CopilotChatToolCallsView` then maps this array using `toolCall.id` as a React `key`, producing the duplicate-key warning:

```
Warning: Encountered two children with the same key, `call_uUF5OZtugHE6tyoC9PPtpAs6`.
```

## Solution

Deduplicate `message.toolCalls` by `id` before rendering in `CopilotChatToolCallsView`, keeping the last occurrence so the most up-to-date arguments are used:

```ts
const uniqueToolCalls = [
  ...new Map(message.toolCalls.map((tc) => [tc.id, tc])).values(),
];
```

This is a defensive fix at the view layer. The underlying idempotency issue in `@ag-ui/client`'s `defaultApplyEvents` has been separately reported upstream.

## Testing

- The fix eliminates the React duplicate-key console warning when `connectAgent()` is called on an agent whose `messages` state already contains completed tool calls.
- Existing tool call rendering behavior is preserved — only the deduplication step is added.

### DIFF
--- a/packages/react-core/src/v2/components/chat/CopilotChatToolCallsView.tsx
+++ b/packages/react-core/src/v2/components/chat/CopilotChatToolCallsView.tsx
@@ -17,9 +17,18 @@ export function CopilotChatToolCallsView({
     return null;
   }
 
+  // Deduplicate tool calls by id. When connectAgent() replays historic
+  // TOOL_CALL_START events over a pre-populated messages state, the same
+  // toolCallId can appear more than once in the array, which causes React to
+  // emit a duplicate-key warning. Keep the last occurrence so that the most
+  // up-to-date arguments are used.
+  const uniqueToolCalls = [
+    ...new Map(message.toolCalls.map((tc) => [tc.id, tc])).values(),
+  ];
+
   return (
     <>
-      {message.toolCalls.map((toolCall) => {
+      {uniqueToolCalls.map((toolCall) => {
         const toolMessage = messages.find(
           (m) => m.role === "tool" && m.toolCallId === toolCall.id,
         ) as ToolMessage | undefined;


BenTaylorDev:
Thanks for the detailed writeup here — the root-cause analysis in the description is accurate, and it's the analysis that shaped how this got fixed.

Closing as superseded. Tool-call dedup now lives one layer up, in `deduplicateMessages` in `packages/react-core/src/v2/components/chat/CopilotChatMessageView.tsx`: a `dedupeToolCalls` helper is applied both when two occurrences of an assistant message merge and, separately, to any assistant message that carries `toolCalls` at all — that second branch covers exactly the `connect()`-replay case described here, where the message itself was never duplicated. `CopilotChatMessageView` renders from the deduplicated array, so `CopilotChatToolCallsView` receives already-deduped `toolCalls` on both the plain and virtualized paths.

One behavioral difference worth recording, since it's the reason the fix landed there rather than here: this PR keeps the *last* occurrence of a duplicated id, while the merged version prefers whichever occurrence actually carries `function.arguments`. In the replay scenario the blank copy tends to arrive second, so "keep last" would render an empty tool-call card — the same visible symptom in a different form.

Tracking for the wider tool-call dedup family is consolidated in #3644 (this branch's linked issue, #3928, was folded into it). If you still see a duplicate-key warning for tool calls on a current release, a reproduction on #3644 would be genuinely useful — the message-level and within-array facets were separate bugs and it's worth knowing if a third one exists.

PR #4059, opened 2026-04-18
fix(react-core): merge toolCalls by id when deduplicating duplicate assistant messages

### LINKED ISSUE #3644: 🐛 Bug: Duplicate assistant message IDs when multiple tool calls are interleaved with tool results

octo-patch:
Fixes #3644

## Problem

When a `TOOL_CALL_RESULT` is interleaved between `TOOL_CALL_START` events for the same `parentMessageId`, `@ag-ui/client` creates a second assistant message with the same id instead of appending to the existing one (because the parent is no longer the last message in the array).

This produces a sequence like:
1. `msg_x` created with `toolCalls: [tool_a, tool_b]`
2. A `TOOL_CALL_RESULT` for `tool_a` is inserted as a separate `"tool"` message
3. `TOOL_CALL_START(tool_c)` — parent `msg_x` is no longer last, so a **new** `msg_x` entry is created with `toolCalls: [tool_c]`

The previous `deduplicateMessages` logic:
```ts
const toolCalls = message.toolCalls ?? existing.toolCalls;
```
picks the later occurrence's `toolCalls` when it's defined — silently dropping `tool_a` and `tool_b`. The dev-mode warning fires but the data is already wrong.

## Solution

When both occurrences have non-empty `toolCalls` arrays, merge them by id (later occurrence wins for any given id):

```ts
const merged = new Map(existing.toolCalls.map((tc) => [tc.id, tc]));
for (const tc of message.toolCalls) {
  merged.set(tc.id, tc);
}
toolCalls = [...merged.values()];
```

Semantics preserved:
- `undefined` → fall back to existing (no tool call activity in this chunk)
- `[]` → intentional reset ("all tool calls finished"), resets the set
- Two non-empty arrays → union by id, later wins

## Testing

Two new unit tests added to the existing `deduplicateMessages` suite:
1. Verifies that toolCalls from both occurrences are merged correctly (reproduces the #3644 scenario)
2. Verifies that when the same tool call id appears in both, the later entry's arguments win

### DIFF
--- a/packages/react-core/src/v2/components/chat/CopilotChatMessageView.tsx
+++ b/packages/react-core/src/v2/components/chat/CopilotChatMessageView.tsx
@@ -309,8 +309,8 @@ const MemoizedCustomMessage = React.memo(
  * Deduplicates messages by ID. For assistant messages, merges occurrences:
  * recovers non-empty content from any earlier occurrence if the latest wiped it
  * (empty string means the streaming update cleared the field, not blank text),
- * and similarly recovers toolCalls from earlier occurrences if the latest is
- * undefined (an empty array [] is treated as intentional and kept as-is).
+ * and merges toolCalls from all occurrences by id so that no tool call is lost
+ * (an explicit [] is treated as intentional and resets the set).
  * For all other roles, keeps the last entry.
  *
  * @internal Exported for unit testing only — not part of the public API.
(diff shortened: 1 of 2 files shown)


BenTaylorDev:
Closing as already-landed — the `deduplicateMessages()` merge-by-id fix (with `toolCalls ?? existing.toolCalls` recovery) is on main at `packages/react-core/src/v2/components/chat/CopilotChatMessageView.tsx:324-339`. Note: #3644 stays open as its own tracker, since the interleaved-tool-calls case may not be fully covered by the dedup.

PR #4062, opened 2026-04-18
feat(showcase): bring 15 packages to 9 QA markdown files (Bundle 0)

jpr5:
Bundle 0 from the [Full Action Inventory](https://www.notion.so/3443aa38185281b5a1dfc6e0890264e1). 75 files generated mechanically from langgraph-python/qa/ templates with only the first-line framework label customized per package. Matches the ag2 precedent already in main.


jpr5:
Consolidated into #4064 (Bundle 0 + Bundle 1 + hitl rename in one PR).

PR #4082, opened 2026-04-19
ci: close showcase deploy automation gap for starters

jpr5:
## Summary

Two related defects in the showcase deploy pipeline let stale images sit live on Railway while Slack stayed green. This PR fixes both.

### Defect 1 — Drift detector skipped all starter services

`.github/workflows/showcase_smoke-monitor.yml` listed only 19 **package** slugs in its `SERVICES=(...)` array (ag2, mastra, llamaindex, ...). Zero **starter** slugs. As a result:

- GHCR `showcase-starter-<svc>` tags were never checked for drift.
- `gh workflow run showcase_deploy.yml -f service=starter-*` was never auto-dispatched.
- Starter services could run with weeks-old images and no alert would fire.

`showcase_deploy.yml` already supports `starter-*` dispatch names and already calls `serviceInstanceRedeploy` for any service with a `railway_id`, so no change is required there. The fix is extending `SERVICES=(...)` to include all 17 starter slugs via a sparse-checkout-driven filesystem enumeration (no more literal duplication between workflow and `showcase/starters/`).

### Defect 2 — Silent deploy failures reported green

`.github/workflows/showcase_deploy.yml` emitted `::warning::` and exited 0 when a service never returned 200 on its health path within 360s. The legacy justification (`# Don't fail — sleep-on-idle services take time to wake`) no longer applies: Railway is on the Pro tier with no sleep-on-idle, so a 6-minute failure to become healthy is a real failure. Changed to `::error::` + `exit 1`.

## Round 2 fixes

Round 2 CR raised six findings against the original smoke-monitor + validator changes. All fixed in this PR:

- **BLOCKING 1/2 — smoke-monitor guard.** Replaced the magic `-eq 19` sentinel with `grep -c '^starter-'` so adds/removes to the literal non-starter list can't silently disable the guard. Added `shopt -s nullglob` around the `showcase/starters/*/` loop so an empty starters tree no longer corrupts `SERVICES` with a `starter-*` literal.
- **BLOCKING 3 — GHCR stderr isolation.** Dropped `2>&1` on the `gh api -i` call; captured stderr to a temp file and surfaced it only when `gh` returns a non-zero RC with no HTTP status. Auth / rate-limit / network noise can no longer splice into the HTTP header block and poison `HTTP_STATUS` / `API_BODY` parsing.
- **BLOCKING 4 — validator tests.** Added `showcase/scripts/__tests__/validate-workflow-starters.test.ts` (12 specs): happy path, missing-from-options-only, missing-from-matrix-only, missing-from-both, empty starters dir (exit 3), template/ excluded, substring-spoof (starter-ag2 vs starter-ag2-extended), missing workflow file (exit 3). Also extended `VALIDATE_WORKFLOW_STARTERS_REPO_ROOT` to re-home the starters dir for testability.
- **MEDIUM 1 — YAML parsing.** Replaced the fragile regex-over-YAML options scanner with a real `yaml.parse()` + typed navigation down `on.workflow_dispatch.inputs.service.options`. ALL_SERVICES stays regex-scanned (embedded JSON in a bash heredoc, with `${{ ... }}` interpolations that aren't valid JSON pre-execution), but the surrounding step is now located via YAML.
- **MEDIUM 2 — Slack list truncation.** Replaced `cut -c1-200` with a `truncate_csv` helper that drops whole comma-separated entries until under budget and appends `…` when truncated. No more `starter-claude-sdk-pyth` mid-slug corruption.
- **MEDIUM 3 — template/ exclusion cross-references.** Both the TS validator's `EXCLUDED_DIRS` and `showcase_smoke-monitor.yml`'s `[ "$slug" = "template" ] && continue` now carry `# keep in sync with ...` comments pointing at each other.
- **NIT 2 — entry-check simplification.** Dropped the belt-and-suspenders `import.meta.url === \`file://${argv[1]}\`` branch; kept only the canonical `fileURLToPath(import.meta.url)` form.
- **NIT 3 — jq pipeline collapse.** Single-pass `.jobs[]? | select | "\(...)"` replaces the three-pass `map | map | .[]` chain in the notify step.

## Files changed

- `.github/workflows/showcase_deploy.yml` — warning → error + exit 1 on unhealthy deploy; `truncate_csv` replaces `cut -c1-200` (3 sites); single-pass jq pipeline in notify step.
- `.github/workflows/showcase_smoke-monitor.yml` — filesystem-driven `SERVICES=(...)`, starter-count guard, `nullglob` loop, stderr-isolated `gh api` call, cross-reference comment.
- `.github/workflows/showcase_validate.yml` — wires `validate-workflow-starters` into CI.
- `showcase/scripts/validate-workflow-starters.ts` — YAML-aware presence checks; env-var override homes both starters dir and workflow path.
- `showcase/scripts/tsconfig.json` — scripts-local tsconfig for LSP type resolution.
- `showcase/scripts/__tests__/validate-workflow-starters.test.ts` — 12 specs covering the full matrix of drift scenarios.

## Test plan

- [ ] Next scheduled `showcase_smoke-monitor` run includes starter services in its drift scan.
- [ ] A deliberately-unhealthy deploy (simulate by pointing health_path at a 404) fails the job and fires the Slack alert.
- [ ] `showcase_validate` CI job runs `validate-workflow-starters` and `npx vitest run scripts/__tests__/validate-workflow-starters.test.ts` green.

### DIFF
--- a/.github/workflows/showcase_deploy.yml
+++ b/.github/workflows/showcase_deploy.yml
@@ -402,7 +402,7 @@ jobs:
                 # Reset streak if we had a partial streak and then a non-200
                 HEALTHY_STREAK=0
               fi
-              # SUCCESS but not (yet) responding — may be sleep-on-idle waking up
+              # SUCCESS but not yet responding — app process still booting
             else
               HEALTHY_STREAK=0
             fi
@@ -418,6 +418,9 @@ jobs:
   notify:
     needs: [detect-changes, check-lockfile, build]
     if: always()
+    permissions:
+      contents: read
+      actions: read
     runs-on: ubuntu-latest
     timeout-minutes: 2
     steps:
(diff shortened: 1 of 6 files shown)


jpr5:
### Round 1 CR fixes applied

All findings addressed on top of the existing history (no rewrites, no force-push).

| # | Severity | Area | Commit |
|---|---|---|---|
| 1 | Blocking | `notify` job conflated partial & full matrix failure | `c6ceb27b2` |
| 2 | Blocking | Shell `/` health probe regression risk (no `/api/health` exists for `shell`/`shell-dojolike`) | `583db441a` |
| 3 | Medium  | Drift detector swallowed GHCR 401/403/5xx as "no versions yet" | `820e3493a` |
| 4 | Medium  | Unanchored SHA grep in drift detector | `820e3493a` (same diff) |
| 5 | High    | No parity enforcement between `showcase/starters/*` and workflow lists | `ecce840e2` |
| 6 | Nit     | Stale "sleep-on-idle waking up" comment (contradicts the Pro-tier rationale a few lines below) | `c6ceb27b2` (folded) |
| 7 | Nit     | Notify cancel-handling comment cited the old `exit 0` warn-not-fail path | `c6ceb27b2` (folded) |

**Notes on implementation choices**
- **Blocking 1** — per-leg detail comes from `gh api /actions/runs/$RUN_ID/jobs`, bucketing each `build (<leg>)` job by conclusion. If the API lookup fails, the fallback message is a softer "1+ of N service(s) failed (<list> targeted)" so the alert never lies about how many legs were affected.
- **Blocking 2** — neither `showcase/shell` nor `showcase/shell-dojolike` ships `/api/health` (verified locally), so option 2 (retry loop) was the only viable fix. The probe now retries up to 3× within each polling iteration before the streak logic observes a non-200 — transient 5xx/301 chains no longer trash accumulated progress but genuinely dead services still fail.
- **High 1** — the new `validate-workflow-starters.ts` uses regex presence checks (not a full YAML parse) with word-boundary anchoring, and was tested locally against the current tree ("all 17 starter(s) registered") and against a synthetic drift ("missing from: <source>" for each gap).

PR #4154, opened 2026-04-22
fix(threads): Improve flickering when switching threads (and connecting)

BenTaylorDev:
This PR includes several fixes that improve the UX of the Langraph + Threads example.

Some are part of the example itself, while others are in the core framework (most of which deal with avoiding flickering in the chat window when switching (so, connecting) to threads.

Example changes:
- package-lock.json included (hence, line count)
- restore button + tooltips on archive/restore/delete
- segmented Active/All filter
- graceful error state if threads don't load
- skeleton rows on initial load
- stable scrollbar gutter
- pre-paint dark-mode class
- logo position stable across app/chat modes
- drop dynamic-import drawer wrapper that caused null first paint
- archived-row dimming via child colors instead of opacity.

Framework changes:
- Skip copilotkit.connectAgent when CopilotChat lacks a caller-supplied threadId — a locally-minted UUID has no backend record, so /connect would always 404 on the intelligence platform.
- Suppress the welcome screen while a connect is in flight and unconditionally when the caller has supplied a threadId (hasExplicitThreadId). Prevents the "How can I help you today?" flash on thread switch.
- Gate suggestions on !isConnecting && !isRunning to avoid painting them against a mid-replay message tree.
- Defer the isConnecting release by one animation frame so trailing bootstrap renders commit before the flag flips.
- Reserve room for the "Powered by CopilotKit" license badge via a new --copilotkit-license-banner-offset CSS var published by the banner on mount; chat input consumes it only when bottom-anchored.
- Sort and display threads by lastRunAt (fallback to updatedAt → createdAt) so metadata-only actions like archive/rename don't reshuffle the list.
- useThreads waits for runtimeConnectionStatus === Connected before dispatching the store context, eliminating the speculative /threads fetch that fired before /info returned wsUrl.

Tests:
- CopilotChat.absentThreadConnect: connect is skipped without a threadId, fires when supplied via prop or config.
- CopilotChatView.connectingGate: isConnecting suppresses welcome; hasExplicitThreadId suppresses welcome on empty chat.
- threads (core): lastRunAt sort fallback ordering.
- use-threads: Connecting-state gate defers /threads until Connected.

## What does this PR do?

## Related PRs and Issues

- ENT-313
- ENT-317
- ENT-314

## Checklist

### DIFF
--- a/.changeset/ent-314-thread-connect-ux.md
+++ b/.changeset/ent-314-thread-connect-ux.md
@@ -1,0 +1,14 @@
+---
+"@copilotkit/react-core": minor
+"@copilotkit/core": patch
+---
+
+fix(threads): stabilize thread-switch UX, skip /connect for absent threads
+
+- Skip `connectAgent` when `CopilotChat` is rendered without a caller-supplied `threadId`. A locally-minted UUID has no backend record, so `/connect` would always 404 on the intelligence platform.
+- Expose `lastRunAt` on `Thread` and use it (with fallback to `updatedAt`, then `createdAt`) as the `useThreads` sort key so metadata-only actions like archive/rename no longer reshuffle the list.
+- `useThreads` waits for `runtimeConnectionStatus === Connected` before dispatching the store context, eliminating a speculative `/threads` fetch that fired before `/info` returned `wsUrl`.
+- Reserve room for the fixed "Powered by CopilotKit" license badge via a new `--copilotkit-license-banner-offset` CSS var published by the banner; chat input consumes it when bottom-anchored so the two no longer overlap.
+- Add a new `bottomAnchored` prop on `CopilotChatInput` for callers rendering it as a flex-last-child.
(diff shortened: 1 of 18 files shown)


MikeRyanDev (changes requested):
Blocking issue found:

- `packages/react-core/src/v2/hooks/use-threads.tsx` now returns `isLoading = false` while `runtimeConnectionStatus !== Connected` and no context has ever been dispatched yet. Because the store starts with `isLoading: false`, the hook sits in an empty/non-loading state until `/info` resolves, then fetches threads afterward. In the threads example this causes a visible empty-state flash on first mount instead of the intended skeleton/loading state.

The fetch deferral itself makes sense, but the hook still needs a pre-connected loading state (or equivalent) until the first context is dispatched / first list fetch settles.

MikeRyanDev (approved):

PR #4162, opened 2026-04-22
fix(shell-dashboard): resync scripts lockfile + unsilence npm ci

jpr5:
## Summary

- `showcase/scripts/package.json` bumped `vitest` to `^4.1.5` and added `@vitest/coverage-v8 ^4.1.5`, but `showcase/scripts/package-lock.json` still carried `vitest@4.1.4` and was missing the coverage package entirely. The Dockerfile's builder stage runs `npm ci` (strict), which refuses to reconcile drift and fails the `build (shell-dashboard)` job on main.
- Regenerated `showcase/scripts/package-lock.json` with `npm install --package-lock-only`.
- Also removes `--silent` from the builder-stage `npm ci` calls so a future same-class failure surfaces the error directly in CI (the original failure logged only `exit code: 1`).

## Why

Job `build (shell-dashboard)` on main commit 5ed233f01 failed with:

```
npm error `npm ci` can only install packages when your package.json and package-lock.json or npm-shrinkwrap.json are in sync.
npm error Missing: @vitest/coverage-v8@4.1.5 from lock file
npm error Invalid: lock file's vitest@4.1.4 does not satisfy vitest@4.1.5
```

The `--silent` flag hid this — stripping it on the builder stage makes the next drift self-diagnosing.

## Test plan

- [x] `docker build -f showcase/shell-dashboard/Dockerfile .` completes end-to-end locally (all stages, through `next build`).
- [ ] CI `build (shell-dashboard)` goes green on this PR.

### DIFF
--- a/showcase/scripts/package-lock.json
+++ b/showcase/scripts/package-lock.json
@@ -20,7 +20,68 @@
       "devDependencies": {
         "@types/node": "^22.0.0",
         "@types/prompts": "^2.4.9",
-        "vitest": "^4.1.3"
+        "@vitest/coverage-v8": "^4.1.5",
+        "vitest": "^4.1.5"
+      }
+    },
+    "node_modules/@babel/helper-string-parser": {
+      "version": "7.27.1",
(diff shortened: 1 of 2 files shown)


jpr5:

PR #4164, opened 2026-04-22
fix(shell-docs): resolve findFrameworksWithCell + SidebarLink scope type errors

jpr5:
## Summary

Two pre-existing TypeScript errors from commit 9b05ed41 (\"feat(showcase/shell-docs): error boundary consolidation + docs components\") were surfacing on main's Docker build — fixed here in parallel with #4160 (which resolved a separate missing-import bug).

## Why

1. `unscoped-docs-page.tsx` imports `findFrameworksWithCell` from `@/lib/docs-render`, but the helper was only declared locally in the two page.tsx routes — with a 1-arg signature (and dead code referencing an undeclared `demos` in one case). The component called it with 3 args.
2. The Step-2 overview cards in `[[...slug]]/page.tsx` render `<SidebarLink>` without the required `scope` prop; the component already ignored `scope` internally (destructured as `_scope`).

## Changes

- Export a shared 3-arg `findFrameworksWithCell(cell, integrationSlugs, demos)` from `src/lib/docs-render.tsx` so the helper stays framework-agnostic (no registry imports).
- Drop the dead local in `src/app/[[...slug]]/page.tsx` (referenced undeclared `demos` — would have been a runtime error if ever called).
- Rewire the live caller in `src/app/[framework]/[[...slug]]/page.tsx` to use the shared export, passing `getIntegrations()` slugs + local `demos` map.
- Relax `SidebarLinkProps.scope` to optional — it was already effectively unused internally.

## Test plan

- [x] `cd showcase/shell-docs && npm run build` — clean compile, all 26 static pages generate.
- [x] `pnpm -w format` — no formatter churn.
- [ ] CI green on the PR.

### DIFF
--- a/showcase/shell-docs/src/app/[[...slug]]/page.tsx
+++ b/showcase/shell-docs/src/app/[[...slug]]/page.tsx
@@ -25,14 +25,6 @@ import {
   type Integration,
 } from "@/lib/registry";
 
-function findFrameworksWithCell(cell: string): string[] {
-  const matches: string[] = [];
-  for (const integration of getIntegrations()) {
-    if (demos[`${integration.slug}::${cell}`]) matches.push(integration.slug);
-  }
-  return matches;
-}
-
 // Category ordering for the framework picker grid is imported from
 // @/lib/docs-render so the landing grid, sidebar dropdown, and this
 // overview share a single source of truth.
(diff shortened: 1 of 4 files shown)


jpr5:

PR #4171, opened 2026-04-22
debug(showcase/langgraph-typescript): instrument probe + event-loop to diagnose recurring Railway kill-loop

jpr5:
## Summary
- Watchdog probe now logs curl's actual stdout/stderr + exit code + HTTP status (not silenced).
- Watchdog emits `ss`/`netstat` snapshot for :8123/:8124 on probe failure.
- `liveness.mjs` emits 5-second event-loop heartbeats.
- Dockerfile installs `iproute2` so `ss` is available in the runtime image.

## Why
Temporary instrumentation. Local smoke works (`/ok` in ~14ms, liveness ticks every 5s, ss reports both listeners) but Railway's 36 grace-window probes all silently fail — we can't see why because the existing curl suppresses all output with `> /dev/null 2>&1`. Ship this, let one kill-cycle run on Railway, read the logs, then revert.

## Test plan
- [ ] Docker build green locally (verified: amd64 build succeeds with depot)
- [ ] Local smoke: liveness binds, heartbeat ticks, /ok returns 200 (verified)
- [ ] After merge + Railway redeploy, next kill-cycle logs show either (a) curl's actual error OR (b) event-loop heartbeat gaps revealing a blocking import

### DIFF
--- a/examples/integrations/langgraph-python-threads/README.md
+++ b/examples/integrations/langgraph-python-threads/README.md
@@ -14,11 +14,11 @@ This project is a monorepo with three services:
 
 When threads are enabled, additional infrastructure runs via Docker Compose:
 
-| Service          | Port       | Description                                                                                                                                        |
-| ---------------- | ---------- | -------------------------------------------------------------------------------------------------------------------------------------------------- |
-| **PostgreSQL**   | 5432       | Thread and event storage                                                                                                                           |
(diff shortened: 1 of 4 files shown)


jpr5:
Closing — instrumentation served its purpose. Root cause found: curl missing from TS runtime stages (node:22-slim). Fixed in #4174.

PR #4172, opened 2026-04-22
fix(react-core): Restore welcome screen when a generated thread UUID is pass from wrappers

BenTaylorDev:
## What does this PR do?

Fresh empty chats were hitting `/connect` against thread IDs the backend had never seen (404s) and the
  welcome screen stayed hidden. Root cause: the v1 `<CopilotKit>` wrapper pipes a
  `ThreadsProvider`-minted UUID down as `threadId`, and downstream consumers were treating `!!threadId`
  as "caller-chosen," so an auto-minted placeholder looked identical to a real thread.

  Fix: introduce an explicit `hasExplicitThreadId` signal and plumb it end-to-end, so consumers that only
   make sense against a real backend thread (`/connect`, switch-flash suppression, welcome-screen gating)
   can distinguish a caller-chosen thread from an internal placeholder.

 ## Changes

  - **`ThreadsProvider`** — exposes `isThreadIdExplicit`. True when the `threadId` prop is supplied or
  `setThreadId()` has been called; false for the first-mount auto-minted UUID. Explicitness via
  `setThreadId` is sticky across prop churn.
  - **`CopilotChatConfigurationProvider`** — accepts an optional `hasExplicitThreadId` prop (overrides
  inference), otherwise infers from `!!threadId`. OR-inherits from parent config so explicitness
  propagates through nested/bare providers.
  - **v1 `CopilotKit` bridge** (`copilotkit.tsx`) — reads `isThreadIdExplicit` from `useThreads()` and
  forwards it to `CopilotChatConfigurationProvider`, so the auto-minted UUID no longer masquerades as
  caller-chosen.
  - **`CopilotChat`** — `isConnecting` and the `/connect` effect now gate on `hasExplicitThreadId`
  instead of `!!providedThreadId`, so a caller-chosen thread from a wrapping provider is honored while a
  leaked auto-UUID is not.

  ## Test plan

  - [x] `threads-context.test.tsx` — `isThreadIdExplicit` transitions: auto→explicit via `setThreadId`,
  prop-driven explicitness, stickiness after prop removal
  - [x] `CopilotChatConfigurationProvider.test.tsx` — inference, explicit `false` override (v1 bridge
  case), OR-inheritance through nested chains
  - [x] `v1-explicit-threadid-bridge.test.tsx` (new) — regression coverage for the v1 wrapper
  specifically
  - [x] `CopilotChat.welcomeGate.test.tsx` (new) — welcome screen stays visible on a bare mount;
  `/connect` is not called without an explicit thread
  - [x] `packages/react-core` full test suite passes (1149 tests)
  - [ ] Manual smoke in `examples/integrations/langgraph-python-threads`: fresh app → welcome screen
  shows → first message starts a real thread

## Related PRs and Issues

## Checklist

### DIFF
--- a/examples/integrations/langgraph-python-threads/apps/app/src/lib/a2ui-theme.css
+++ b/examples/integrations/langgraph-python-threads/apps/app/src/lib/a2ui-theme.css
@@ -153,7 +153,6 @@ body {
   --font-family-mono:
     "Google Sans Code", "Helvetica Neue", Helvetica, Arial, sans-serif;
 
-  background: var(--background-light);
   font-family: var(--font-family);
   margin: 0;
   padding: 0;
--- a/packages/cli/src/utils/version.ts
+++ b/packages/cli/src/utils/version.ts
@@ -1,2 +1,2 @@
 // This is auto generated!
-export const LIB_VERSION = "1.0.0";
+export const LIB_VERSION = "1.0.2";
--- a/packages/react-core/src/components/copilot-provider/__tests__/v1-explicit-threadid-bridge.test.tsx
+++ b/packages/react-core/src/components/copilot-provider/__tests__/v1-explicit-threadid-bridge.test.tsx
@@ -1,0 +1,107 @@
+import React from "react";
+import { render, screen, act } from "@testing-library/react";
+import { describe, it, expect, beforeEach, afterEach, vi } from "vitest";
+import { CopilotKit } from "../copilotkit";
+import { useCopilotContext } from "../../../context/copilot-context";
+import { useCopilotChatConfiguration } from "../../../v2";
+import type { CopilotKitProps } from "../copilotkit-props";
+
+/**
+ * Probe that reads hasExplicitThreadId from the CopilotChatConfigurationProvider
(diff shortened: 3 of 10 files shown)


examples/integrations/langgraph-python-threads/apps/app/src/lib/a2ui-theme.css
```diff
@@ -153,7 +153,6 @@ body {
   --font-family-mono:
     "Google Sans Code", "Helvetica Neue", Helvetica, Arial, sans-serif;
 
-  background: var(--background-light);
```
BenTaylorDev, line 156: This fixes a visual issue specific to the threads example found while testing the broader change.

packages/cli/src/utils/version.ts
```diff
@@ -1,2 +1,2 @@
 // This is auto generated!
-export const LIB_VERSION = "1.0.0";
+export const LIB_VERSION = "1.0.2";
```
BenTaylorDev, line 2: This bumps the version to match version.d.ts - no actual version change occurs.

tylerslaton (approved):

PR #4173, opened 2026-04-22
feat(showcase/ops): L1-L4 alert rules for agent/chat/tools depth signals

jpr5:
## Summary
- New alert rule YAMLs for the L2/L3/L4 side-channel signals emitted by the smoke + e2e-smoke drivers (`agent:<slug>`, `chat:<slug>`, `tools:<slug>`).
- Each rule mirrors `smoke-red-tick.yml` exactly: `green_to_red` / `sustained_red` / `red_to_green` triggers, `!channel` escalation at failCount=4, 20-minute minDeployAgeMin guard, 15-minute per-key rate limit, and suppression of chatty `sustained_red` ticks within a 15-minute window.
- Extends the closed `DIMENSIONS` enum in `src/types/index.ts` with `agent`, `chat`, `tools` so the Zod schema derived from it accepts the new rules at load; typo'd dimensions continue to fail with a listed-valid-members error.

## Why
Slots 1 + 2 of the L1-L4 buildout extend the smoke + e2e-smoke drivers to emit per-starter depth signals. Without dedicated rules those ticks land in the status store but never page anyone — L2-L4 regressions would be silent even though the probes went red. These rules close that gap.

## Scope guardrails
- Pure YAML work plus the single-line enum extension the schema requires.
- No driver, discovery, orchestrator, or template-filter changes.
- Templates use only double-brace for `signal.*` fields (no new `slackSafeFields` wiring needed); triple-brace is restricted to loader-known-safe `env.dashboardUrl` / `event.runId`.

## Test plan
- [x] `pnpm exec vitest run src/rules/rule-loader.test.ts` — 64 passed (new rules load cleanly alongside the existing 8).
- [x] `pnpm exec vitest run` — 692 passed across 37 files.
- [x] `pnpm exec tsc --noEmit` — clean.
- [x] `pnpm -w format` — no drift.
- [ ] Smoke that a synthesised `agent:<slug>` red tick renders the expected template in staging before flipping discovery to emit these dimensions in production.

### DIFF
--- a/showcase/ops/config/alerts/agent-red-tick.yml
+++ b/showcase/ops/config/alerts/agent-red-tick.yml
@@ -1,0 +1,40 @@
+id: agent-red-tick
+name: "Starter agent reachability failing"
+owner: "@oss"
+
+signal:
+  dimension: agent
+
+triggers:
+  - green_to_red
+  - sustained_red
+  - red_to_green
+
+conditions:
+  escalations:
+    - whenFailCount: 4
+      mention: "!channel"
+      severity: critical
+
+  guards:
+    - minDeployAgeMin: 20
+
+  rate_limit:
+    perKey: "{{signal.slug}}:{{triggerName}}"
+    window: 15m
+
+  suppress:
+    when: 'trigger == "sustained_red" && lastAlertAgeMin < 15'
+
+targets:
+  - kind: slack_webhook
+    webhook: oss_alerts
+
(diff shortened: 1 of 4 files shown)


jpr5:
Superseded by #4178 — consolidated L1-L4 buildout with 7 grouped commits, overlap on railway-services.ts nameExcludes resolved.

PR #4182, opened 2026-04-22
chore(showcase): untrack generated registry + demo-content bundles

samjulien:
## Summary

- `registry.json`, `constraints.json`, `demo-content.json` under each shell's `src/data/` are rewritten by `generate-registry` + `bundle-demo-content` on every dev cycle and build — they were churning constantly in `git status` and leaking into unrelated PRs. Added them to `.gitignore` and removed from the index (7 files total).
- Each shell now runs `generate-registry` in a `predev` hook. `bundle-demo-content --watch` already does an initial emit, so fresh clones can run `npm run dev` without manual bundling.
- `shell-dojo` previously had no bundler step in its `build` — it relied on committed files to deploy. Now runs `generate-registry` + `bundle-demo-content` before `next build`, matching `shell` and `shell-docs`.
- `shell-dojo` and `shell-docs` `dev` scripts gain the bundle watcher so local dev regenerates on demo-source edits, matching the pattern `shell` already uses.

## Scope notes

Left tracked for now: `search-index.json` (shell + shell-docs), `starter-content.json` (shell), `docs-status.json` (shell). Their generators only run at build time or from `shell-dashboard` (`probe-docs`), so they're much quieter in day-to-day dev. Worth a follow-up if we want them untracked too.

## Test plan

- [x] Deleted all 7 regenerated files locally → ran `generate-registry` + `bundle-demo-content` → all 7 regenerated cleanly
- [x] Confirmed `git status` is clean of the generated files after regeneration (`.gitignore` rules active)
- [x] Verify Vercel preview deploy succeeds for each shell (shell, shell-docs, shell-dojo) — **please force rebuild without cache on at least one preview to rule out cached-JSON false positives**
- [x] Verify `npm run dev` works from a fresh clone / no pre-existing `src/data/` for each shell


samjulien:
Fixed by #4236

PR #4187, opened 2026-04-23
fix(docs): update copilot cloud link and fix lefthook when running on docs

tylerslaton:
Updating a link on the docs site and also fixing an issue with doc changes not going through lefthook properly.


mxmzb (approved):

MikeRyanDev (approved):
Reviewed locally. The navbar label update is consistent with the existing responsive behavior, and the lefthook change correctly skips docs-only staged files instead of letting oxfmt fail on an empty target set.

PR #4192, opened 2026-04-23
fix(sdk-python): relax partialjson version pin to allow 1.x

### LINKED ISSUE #4131: [python-sdk] Relax partialjson<0.0.9 pin to allow 1.x
## Summary

The `copilotkit` Python SDK pins `partialjson<0.0.9` in its dependencies. This blocks downstream projects from upgrading to `partialjson` 1.x, which has been published for months with backward-compatible API additions.

## Impact

Downstream projects cannot accept Dependabot bumps for `partialjson>=1.0` as long as this constraint holds. We are currently blocked on two Dependabot PRs in our repo pending this upstream relaxation.

## Proposed fix

Widen the constraint in the python-sdk `pyproject.toml` to `partialjson>=0.0.8,<2.0` (or `<3.0` given the project's semver cadence). `partialjson` 0.0.9 → 1.x is a non-breaking expansion per its changelog; existing `copilotkit` usage of `JSONParser.parse` remains compatible.

## Evidence

- partialjson 1.0.0 release: https://pypi.org/project/partialjson/1.0.0/
- partialjson changelog shows additive changes only between 0.0.8 and 1.x
- No API surface used by copilotkit appears broken by the bump

## Context

Filed from the LexAggregator project. Happy to submit a PR widening the pin if maintainers confirm the direction.

octo-patch:
Fixes #4131

## Problem

The `partialjson` dependency in `sdk-python/pyproject.toml` was pinned as `^0.0.8`, which in Poetry's semver notation translates to `>=0.0.8, <0.0.9`. This effectively blocks downstream projects from using `partialjson` 1.x, and prevents Dependabot from bumping the package.

## Solution

Widen the constraint to `>=0.0.8,<2.0.0`. The only usage in the SDK is `JSONParser().parse()` (in `copilotkit/runloop.py`), which is fully backward-compatible across the 0.0.8 → 1.x range per the partialjson changelog.

## Testing

- Verified `JSONParser` is the only partialjson API consumed by the SDK
- The additive changes between 0.0.8 and 1.x do not affect the existing API surface
- The upper bound `<2.0.0` guards against potential future breaking changes

### DIFF
--- a/sdk-python/pyproject.toml
+++ b/sdk-python/pyproject.toml
@@ -16,18 +16,24 @@ keywords = [
   "langsmith",
   "langserve",
 ]
+classifiers = [
+  "Programming Language :: Python :: 3.10",
+  "Programming Language :: Python :: 3.11",
+  "Programming Language :: Python :: 3.12",
+  "Programming Language :: Python :: 3.13",
+]
 packages = [{ include = "copilotkit" }]
-include = ["copilotkit/*.json"]
+include = ["copilotkit/*.json", "copilotkit/py.typed"]
 
 [tool.poetry.dependencies]
-python = ">=3.10,<3.13"
+python = ">=3.10,<3.14"
 langgraph = { version = ">=0.3.25,<2" }
 langchain = { version = ">=0.3.0" }
-crewai = { version = "0.118.0", optional = true }
+crewai = { version = ">=0.118.0", optional = true, python = ">=3.10,<3.13" }
 ag-ui-langgraph = { version = ">=0.0.29", extras = ["fastapi"] }
 ag-ui-protocol = ">=0.1.15"
 fastapi = ">=0.115.0,<1.0.0"
-partialjson = "^0.0.8"
+partialjson = ">=0.0.8,<2.0.0"
 toml = "^0.10.2"
 
 [tool.poetry.extras]


BenTaylorDev (changes requested):
Thanks — the underlying ask is valid: `main` still has `partialjson = "^0.0.8"` (i.e. `<0.0.9`), which does block 1.x and Dependabot. But this can't merge as-is:

1. **Bundled changes would regress `main`.** Beyond the pin, the diff also lowers `python` to `>=3.10,<3.14` (main is `<3.15`), lowers the `crewai` marker to `<3.13` (main is `<3.14`), and drops the `Python :: 3.14` classifier. Those all move backward from where `main` already is, and they're what causes the merge conflict on `pyproject.toml`. Please drop them and keep the change to just the `partialjson` constraint.
2. **The lockfile needs regenerating.** `sdk-python/poetry.lock` (and `uv.lock`) still pin `partialjson 0.0.8`, and CI installs from the lock — so relaxing `pyproject` alone is inert; it'll still resolve `0.0.8`. Regenerate the locks so `>=0.0.8,<2.0.0` can actually pick up 1.x.
3. **Confirm 1.x compatibility.** The only consumer is `runloop.py` (`JSONParser().parse(...)`). Worth a quick check that `partialjson` 1.x preserves that API/behavior before widening the range.

Simplest path is a fresh, minimal PR against current `main`: the one-line `partialjson` bump plus a lock regen. Thanks for flagging it.

BenTaylorDev:
Following up on my review from a few weeks back.

Closing this branch, but not the underlying ask — #4131 stays open and the fix is still wanted.

What changed in the meantime is that `main` has independently picked up most of what this diff carried, and gone further: the Python classifiers now run through 3.14, `copilotkit/py.typed` is in `include`, `python` is `>=3.10,<3.15`, and `crewai` is `{ version = ">=0.118.0", python = ">=3.10,<3.14" }`. Merging as-authored would now walk those backward — to `<3.14`, `<3.13`, and a dropped 3.14 classifier — and that's exactly what the `pyproject.toml` conflict is.

The one piece that has *not* landed is the part this PR is named for: `main` still has `partialjson = "^0.0.8"`. I've confirmed `partialjson` 1.0.0 and 1.1.0 are on PyPI, so the constraint really does block them, and your reading of the API surface matches — `JSONParser().parse()` in `runloop.py` is the only consumer.

We'll take that one line directly on `main`, along with the lockfile regen it needs (the lock still resolves `partialjson` independently of the `pyproject` constraint, so relaxing the constraint alone would have been inert). Thanks for flagging it and for the clear writeup on #4131 — that's what made it easy to act on.

PR #4193, opened 2026-04-23
fix(react-core): prevent cursor jump when typing in mobile viewport

### LINKED ISSUE #4150: 🐛 Bug: [Critial] Cannot insert text in the chatbox, it always jump to the End when  shrink the width of the browser windows to the smallest
### ♻️ Reproduction Steps

1. open https://dojo.ag-ui.com/langgraph/feature/agentic_chat
2. shrink the width of the windows to the smallest 
3. input "ABCD", then put cursor to the front of "A" and input "E",  the Cursor will jump to the End of "D"
...

### ✅ Expected Behavior

can insert text in any position

### ❌ Actual Behavior

always jump to the End

### Screenshots

Image

### 𝌚 CopilotKit Version

```shell
1.54.0
```

### 📄 Logs (Optional)

```shell

```

octo-patch:
Fixes #4150

## Problem

When the browser window is narrowed below 768px (mobile viewport), typing in the CopilotChat input box causes the cursor to jump to the end of the text after every keystroke. This makes it impossible to edit text in the middle of a string on mobile-sized viewports.

**Root cause:** The `evaluateLayout()` function is triggered on every `ResizeObserver` tick (which fires as the textarea grows while the user types). On the mobile viewport path, it unconditionally called `ensureMeasurements()`. Inside that function, `textarea.value` is temporarily set to `""` to measure single-line height, then restored — but this programmatic assignment resets the browser's cursor position to the end of the string.

## Solution

Guard the `ensureMeasurements()` call in the mobile viewport path so it only runs when measurements haven't been taken yet (`singleLineHeight === 0`), the same guard already used in `adjustTextareaHeight()` and in the non-mobile code path. After the first evaluation, subsequent calls skip the destructive measurement and leave the cursor untouched.

## Testing

- Reproduce by resizing the browser to <768px and typing text, then trying to insert characters in the middle — the cursor no longer jumps to the end
- Textarea height still adjusts correctly on mobile viewports (first-time measurement still runs via the `singleLineHeight === 0` guard)
- Desktop behavior is unchanged

### DIFF
--- a/packages/react-core/src/v2/components/chat/CopilotChatInput.tsx
+++ b/packages/react-core/src/v2/components/chat/CopilotChatInput.tsx
@@ -87,6 +87,19 @@ type CopilotChatInputRestProps = {
   containerRef?: React.Ref<HTMLDivElement>;
   /** Whether to show the disclaimer. Default: true for absolute positioning, false for static */
   showDisclaimer?: boolean;
+  /**
+   * Set to `true` when the input sits at the bottom of its container as a
+   * flex-last-child (visible position is driven by layout, not CSS
+   * positioning). Triggers reservation of bottom space for the fixed
+   * CopilotKit license banner via the
+   * `--copilotkit-license-banner-offset` CSS var so the two don't overlap.
+   *
+   * Not needed when `positioning === "absolute"`; that mode already pins the
+   * input to the bottom and picks up the same reservation automatically.
+   * Leave unset (default `false`) for inputs rendered mid-layout such as the
+   * welcome screen, where the banner offset would push the input off-center.
+   */
+  bottomAnchored?: boolean;
 } & Omit<React.HTMLAttributes<HTMLDivElement>, "onChange">;
 
 type CopilotChatInputBaseProps = WithSlots<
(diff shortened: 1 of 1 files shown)


kenizhou:
Thanks for your efforts, may I the know the plan to release?

octo-patch:
Hi @kenizhou — thanks for the interest! As an external contributor I can't speak to the release timeline; that's up to the CopilotKit maintainers, who would also need to review and merge this PR first.

kenizhou:
> Hi @kenizhou — thanks for the interest! As an external contributor I can't speak to the release timeline; that's up to the CopilotKit maintainers, who would also need to review and merge this PR first.

Noted and thanks, thanks for you great works!

BenTaylorDev:
Thanks for this — you correctly identified the cause: on a narrow/mobile viewport the input's `ensureMeasurements()` runs on every ResizeObserver tick and does `textarea.value = ""` then restores it, resetting the caret to the end mid-typing. It still reproduces on current `main`.

I'm consolidating on the competing #5432, which fixes the same bug on the same v2 `CopilotChatInput` surface but takes the more minimal route — it *removes* the redundant `ensureMeasurements()` call from the mobile branch (the following `adjustTextareaHeight()` already handles cold-cache measurement via its own guard) and adds a regression test. This PR's equivalent guard duplicates logic that already exists, and its bundled banner-offset change has since landed separately on `main`.

Since both fix the same issue and #5432 is the leaner, tested version, I'll close this as a duplicate and advance #5432 — but the diagnosis credit is yours. Thanks again.

PR #4196, opened 2026-04-23
fix(showcase-shell): separate CLI init command from live demos

AlemTuzlak:
## Summary

On integration profile pages (e.g. `/integrations/langgraph-python`), the `cli-start` demo entry — which has a `command` field instead of a `route` — was being rendered inside the Live Demos grid. Clicking it opened `DemoDrawer` with an undefined `demoRoute`, so the iframe tried to load `${backend_url}undefined`.

This PR:
- Splits `integration.demos` into `liveDemos` (runnable, have `route`) and `commandDemos` (CLI-only, have `command`).
- Filters command-only entries out of the Live Demos grid.
- Adds a new **Get Started** section above Live Demos that renders each `commandDemos` entry as a description + copy-able `<code>` block (mirrors how `shell-dashboard` already handles these via `CommandCell`).
- Makes `Demo.route` optional and adds `Demo.command?: string` on the `Demo` type.

Only `langgraph-python` currently has a command-only demo, so this is the only profile page that changes visibly.

## Test plan

- [x] Ran `showcase/shell` dev server, visited `/integrations/langgraph-python`, confirmed:
  - New "Get Started" card shows `npx copilotkit@latest init --framework langgraph-python` with a working Copy button.
  - "Live Demos" grid no longer contains the broken "CLI Start Command" tile.
  - Other integrations (no `command` demos) render unchanged — Live Demos still present, no empty "Get Started" section.
- [x] Pre-commit hooks (lint-fix, check-binaries, test-and-check-packages — 1149 tests, commitlint) all pass.


marthakelly (commented):
## PR #4196 Review — fix(showcase-shell): separate CLI init command from live demos

**Author:** AlemTuzlak | **Branch:** `fix/showcase-shell-cli-start-not-live-demo` → `main` | **Files:** 3 changed | **State:** OPEN

---

### Critical Issues

**(1)** `demo.route` **used without null check in** `[slug]/[demo]/page.tsx:62` **— broken iframe URL for command-only demos**

```typescript
const iframeSrc = `${integration.backend_url}${demo.route}`;
```

This PR makes `Demo.route` optional in `registry.ts`, but this unmodified consumer constructs an iframe URL by concatenating `demo.route` with no fallback. Navigating to `/integrations/langgraph-python/cli-start` (via URL bar, bookmark, or link from the matrix page) finds the `cli-start` demo which has no `route`, producing `"https://backend.example.com/undefined"`. The `demo` state is typed as `any` (line 29), so TypeScript provides zero protection.

**Fix:** Add a guard after the existing `!demo` check — if `!demo.route`, show a "not a live preview" message instead of rendering the iframe.

---

**(2) Same broken URL in** `[slug]/[demo]/preview/page.tsx:37`

```typescript
const src = `${base}${demo.route}`;
```

This file properly imports the typed `Demo` from `@/lib/registry`, so `demo.route` is now `string | undefined`. Same `"undefined"` URL concatenation for command-only demos accessed via `/integrations/langgraph-python/cli-start/preview`. TypeScript doesn't flag it because template literals accept `undefined` — they just stringify it.

**Fix:** Same pattern — guard for `!demo.route` before constructing the URL.

---

### Important Issues

**(3)** `Demo` **interface duplicated between** `profile-client.tsx` **(lines 12-20) and** `registry.ts` **(lines 15-23)**

The same `Demo` interface is defined independently in both files. `preview/page.tsx` already imports from `@/lib/registry` (line 5), proving shared imports work. This duplication means the types can drift silently — if one copy is updated without the other, consumers see inconsistent shapes.

**Recommendation:** Delete the local interfaces in `profile-client.tsx` and import from `@/lib/registry`.

---

**(4)** `Demo` **type permits invalid states — no invariant enforcement**

Both `route` and `command` are independently optional, allowing four combinations: route-only, command-only, both, and neither. The runtime filter logic in `profile-client.tsx` (lines 90-91) reveals the intent is a two-state model (live vs. command), but this invariant is invisible in the type definition. A demo with neither field silently vanishes from both sections; a demo with both fields appears only as a command demo.

**Recommendation (future improvement):** Consider a discriminated union (`{ type: "live"; route: string } | { type: "command"; command: string }`) to make invalid states unrepresentable. This isn't a blocking concern for this PR but would prevent the class of bugs in issues (1) and (2).

---

**(5)** `page.tsx` **uses** `any` **type for demo state (line 29), bypassing all type safety**

```typescript
const [demo, setDemo] = useState<any>(null);
```

Even with a perfectly designed `Demo` type, this consumer gets zero compiler protection. The `any` predates this PR but becomes more dangerous now that `route` is optional.

**Recommendation:** Change to `useState<Demo | null>(null)` and import the type.

---

### Suggestions

**(6)** `copyDemoCommand` **catch handler swallows errors silently (**`profile-client.tsx:128`**)**

The `.catch(() => { window.prompt(...) })` block discards the error without logging. If `navigator.clipboard` is unavailable (non-HTTPS) and `window.prompt` is also blocked (cross-origin iframe, popup blocker), the user clicks "Copy" and nothing happens — no feedback, no log. The existing `copyCloneCommand` has the same pattern, but at minimum the new function should accept the error parameter and log it.

---

**(7) Demos with neither** `route` **nor** `command` **silently disappear**

Both filters on lines 90-91 would exclude such a demo. Currently no such entry exists in the data, but a misconfigured registry entry would be invisible with no console warning to aid debugging.

---

**(8) JSX comment** `{/* Demos */}` **(line 493) is now ambiguous**

Post-split, this comment guards the `liveDemos` section but still says "Demos." The rendered heading says "Live Demos" — updating the comment to match would reduce ambiguity.

---

### Strengths

- The core fix is correct and well-structured — command-only demos are cleanly separated from live demos with proper filter logic.
- Conditional rendering (`commandDemos.length > 0`, `liveDemos.length > 0`) prevents empty sections from appearing.
- The `activeDemo.route ?? ""` fallback on line 551 is technically unreachable (since `openDemo` is only called from `liveDemos.map` where route is guaranteed truthy), making it a safe TypeScript-satisfying pattern rather than a real fallback path.
- The new "Get Started" section with copy-to-clipboard mirrors the existing `copyCloneCommand` pattern, maintaining consistency.
- JSON formatting changes are purely cosmetic (array compaction) with no data modifications.

---

### Recommended Action

**Request changes.** The two critical issues — unguarded `demo.route` usage in `page.tsx:62` and `preview/page.tsx:37` — mean that direct navigation to a command-only demo's detail page will render a broken iframe with a URL containing the literal string "undefined." These are the same class of bug the PR set out to fix in `profile-client.tsx`, just in two other downstream consumers that were missed. Adding a `!demo.route` guard with appropriate fallback UI in both files would complete the fix.

AlemTuzlak:
reopening to nudge CI

ranst91 (approved):

Issue #3141, opened 2026-02-02
🐛 Bug: HTTP Only Cookies - `credentials` not forwarded to all requests


luke-pk:
### ♻️ Reproduction Steps

1. Set up CopilotKit with React front-end, Node.js Express back-end
2. Enable HTTP-only cookies for application
3. Set `credentials: "include"` for `<CopilotKit ...>`
4. Make requests & observe that some requests do not send the cookies as expected

Note: 

Up until the latest release 1.51.3, CopilotKit did not support HTTP-only cookies. In 1.51.3 a PR was merged in to resolve this issue: https://github.com/CopilotKit/CopilotKit/pull/2970

However as described above, some requests still do not send the cookies.

I suspect there are one or more locations in the code that are still missing the required argument. For example the provisional agent created in useAgent definition only sets the headers, not the credentials property (see screenshot).

<img width="798" height="777" alt="Image" src="https://github.com/user-attachments/assets/227163a3-20dc-409b-be77-7f929ec877e7" />

### ✅ Expected Behavior

All requests originating from copilotkit to the runtime URL should forward the HTTP-only cookies.

### ❌ Actual Behavior

Initial requests with `method: "agent/connect"` do not send the cookies. Further requests to send messages do send the cookies.

### 𝌚 CopilotKit Version

```shell
├── @copilotkit/react-core@1.51.3
├── @copilotkit/react-ui@1.51.3
├── @copilotkitnext/react@1.51.3
```

### 📄 Logs (Optional)

```shell

```

luke-pk:
I agree with the bot, I tried manually patching `useAgent` to set `credentials` for the provisional agent and the cookies are properly forwarded.

PR #4199, opened 2026-04-23
fix(core): forward credentials on REST transport in ProxiedCopilotRuntimeAgent

### LINKED ISSUE #4198: 🐛 Bug: ProxiedCopilotRuntimeAgent.requestInit() drops credentials on default (rest/auto) transport
### ♻️ Reproduction Steps

1. Install `@copilotkit/react-core@1.56.3` and point it at a cross-origin runtime URL.
2. Pass `credentials="include"` to `<CopilotKitProvider>` for HTTP-only cookie auth.
3. Leave `useSingleEndpoint` unset (default `"auto"` → REST routing).
4. Trigger any chat interaction that hits `POST /agent/:id/connect` or `/agent/:id/run`.
5. DevTools Network tab: no `Cookie` header on the outgoing request.

### ✅ Expected Behavior

With `credentials="include"` on the provider, every runtime HTTP request from `ProxiedCopilotRuntimeAgent` should include `credentials: "include"` in its `RequestInit` — just like on the single-route transport.

### ❌ Actual Behavior

Only `createSingleRouteRequestInit` spreads `this.credentials` into the returned `RequestInit`. The default `/agent/:id/connect` and `/agent/:id/run` paths in `#connectViaHttp`/`#runViaHttp` call `this.requestInit(input)` inherited from `@ag-ui/client`'s `HttpAgent`, which builds `{ method, headers, body, signal }` and never reads `this.credentials`. The browser falls back to `credentials: "same-origin"`, so cookies are dropped on cross-origin calls.

The symptom is the same class as #3141 (closed by #3839). #3839 fixed the *assignment* side (`applyCredentialsToAgent` on every registration path), but the *consumption* side for REST transport still ignores the property. Custom `headers` work fine on the same path because they're read from `this.headers` inside `HttpAgent.requestInit`.

A minimal fix — adding a `requestInit` override on `ProxiedCopilotRuntimeAgent` alongside the existing `createSingleRouteRequestInit` — restores cookies on all transports:

```ts
requestInit(input: RunAgentInput): RequestInit {
  const baseInit = super.requestInit(input);
  return {
    ...baseInit,
    ...(this.credentials ? { credentials: this.credentials } : {}),
  };
}
```

Verified against 1.56.3 locally as a yarn patch — cookies flow on `/connect` after applying.

### 𝌚 CopilotKit Version

```shell
@copilotkit/core@1.56.3
@copilotkit/react-core@1.56.3
```

### 📄 Logs

```shell
POST https://<runtime>/agent/<id>/connect 401 (Unauthorized)
```

Request headers on the failing request include custom `X-*` headers configured on the provider but no `Cookie` header, confirming `headers` propagate while `credentials` are dropped.

nekitk:
## Summary

Fixes #4198.

`HttpAgent.requestInit()` from `@ag-ui/client` returns `{ method, headers, body, signal }` and does **not** include `credentials`. `ProxiedCopilotRuntimeAgent` declares `credentials?: RequestCredentials` and correctly spreads it inside `createSingleRouteRequestInit` (single-route transport), but the default REST transport — `#connectViaHttp` (`/agent/:id/connect`) and `#runViaHttp` (`/agent/:id/run` via `super.run`) — calls the inherited `this.requestInit(input)` and drops `credentials`. As a result, cookies are lost on cross-origin requests even when `<CopilotKitProvider credentials="include">` is set.

This PR adds a `requestInit` override on `ProxiedCopilotRuntimeAgent` that mirrors the existing pattern in `createSingleRouteRequestInit`: take the base init from `super.requestInit(input)` and spread `this.credentials` on top when configured. Behavior is unchanged when `credentials` is not set.

```ts
override requestInit(input: RunAgentInput): RequestInit {
  const baseInit = super.requestInit(input);
  return {
    ...baseInit,
    ...(this.credentials ? { credentials: this.credentials } : {}),
  };
}
```

### Related prior work

- #3141 — original bug report (assignment side)
- #3839 — fixed `applyCredentialsToAgent` on every registration path but didn't address the consumption side on the REST transport, which is what this PR does.

## Test plan

Added three tests in `packages/core/src/__tests__/core-credentials.test.ts` (mirroring the existing `core-credentials.test.ts` / `proxied-runtime-transport.test.ts` mocking conventions):

- [x] REST transport `runAgent` forwards `credentials: "include"` to `fetch`
- [x] REST transport `connectAgent` forwards `credentials: "same-origin"` to `fetch`
- [x] REST transport `runAgent` omits `credentials` when not configured

All 399 tests in `@copilotkit/core` pass locally. `pnpm -F @copilotkit/core build` succeeds.

### DIFF
--- a/.changeset/forward-credentials-rest-transport.md
+++ b/.changeset/forward-credentials-rest-transport.md
@@ -1,0 +1,14 @@
+---
+"@copilotkit/core": patch
+---
+
+fix: forward credentials on REST transport in ProxiedCopilotRuntimeAgent
+
+The REST transport (default `/agent/:id/connect` and `/agent/:id/run` paths)
+inherited `requestInit` from `HttpAgent`, which does not include the agent's
+`credentials` field. As a result, cookies were dropped on cross-origin
+requests even when `credentials` was configured on the provider. Override
+`requestInit` in `ProxiedCopilotRuntimeAgent` so the REST paths honor the
+configured credentials, matching the single-route transport behavior.
+
+Fixes #4198.
--- a/packages/core/src/__tests__/core-credentials.test.ts
+++ b/packages/core/src/__tests__/core-credentials.test.ts
@@ -1,6 +1,39 @@
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import { CopilotKitCore } from "../core";
+import { ProxiedCopilotRuntimeAgent } from "../agent";
 import { waitForCondition } from "./test-utils";
+
+const encoder = new TextEncoder();
+
+function createSseResponse(): Response {
+  const stream = new ReadableStream({
+    start(controller) {
+      const events = [
+        {
+          type: "RUN_STARTED",
+          threadId: "test-thread",
+          runId: "test-run",
+        },
+        {
+          type: "RUN_FINISHED",
+          threadId: "test-thread",
+          runId: "test-run",
+          result: { newMessages: [] },
+        },
+      ];
+      const payload = events
+        .map((event) => `data: ${JSON.stringify(event)}\n\n`)
+        .join("");
+      controller.enqueue(encoder.encode(payload));
+      controller.close();
+    },
+  });
+
+  return new Response(stream, {
+    status: 200,
+    headers: { "content-type": "text/event-stream" },
+  });
+}
 
 describe("CopilotKitCore credentials", () => {
   const originalFetch = global.fetch;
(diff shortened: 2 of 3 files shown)


BenTaylorDev:
Thanks for this, and for diagnosing the root cause correctly — the REST/`auto` transport really does drop `credentials` because `ProxiedCopilotRuntimeAgent` inherits `HttpAgent.requestInit` (which omits it), and your `override requestInit()` is the right fix. The bug still reproduces on current `main`, so this is a real issue.

There's a competing PR, #5433, that implements the same `override requestInit()` approach and currently applies cleanly to `main` with tests on both the run and connect paths. Rather than keep two identical fixes open, I'm consolidating on that one and closing this as a duplicate.

Credit for the diagnosis is yours — thank you for catching it. Closing in favor of #5433.

PR #4206, opened 2026-04-23
chore(claude): add SessionStart hook to install external skills in cloud sessions

marthakelly:
## Summary

Adds a hook that auto-installs skills from an external git repo at the start of every **Claude Code on the web** session (the cloud version, not the desktop app). Keeps skills in sync with the source repo's default branch automatically.

This public repo contains no reference to any specific private source.

## Where this runs

- ✅ **Claude Code on the web** (cloud sandboxes) — hook runs, skills install into `~/.claude/skills/` at session start.
- ❌ **Claude Code desktop app / CLI on your laptop** — hook is gated behind `CLAUDE_CODE_REMOTE=true` and exits immediately. Your local `~/.claude/skills/` is untouched. Manage local skills however you already do.

## Files added in this PR

**`.claude/settings.json`** — Claude Code's standard config file. Registers the script below to run on the `SessionStart` event. This is a supported Claude Code extension point.

**`.claude/hooks/session-start.sh`** — the script that runs at session start. On each cloud session it:
- Exits early (silent no-op) if not in a cloud session, or if the required env vars aren't set — so non-configured sessions, including external contributors cloning this OSS repo, just start normally.
- Clones the configured skills repo (or fetch + reset if already cached) using `SKILLS_REPO` + `SKILLS_TOKEN`, over HTTPS with the token.
- Runs the skills repo's own `scripts/install.sh` to install skills into their expected locations (reuses the canonical install path instead of reinventing it).
- Scrubs the token from the on-disk git config after each run.

## Admin config required to activate

In Github:

1. Generate a read-only GitHub token for the `` repo

In Anthropic / Claude Code on the web admin console:

1. Add env vars `SKILLS_REPO` (e.g. ``) and `SKILLS_TOKEN`
2. Add the `` repo to the list of repos sessions are allowed to clone.

Without these the hook no-ops silently.

## Test plan

- [x] `bash -n` syntax check
- [x] Dry-run with `CLAUDE_CODE_REMOTE` unset → silent no-op (desktop/CLI path)
- [x] Dry-run in remote mode without env vars → silent no-op
- [x] Dry-run in remote mode with partial vars → silent no-op
- [ ] Verify end-to-end in a real cloud session once admin has set the token + added the repo to the allowed list

https://claude.ai/code/session_01Y8KwSRqQNZVCEe1tAjCoe1


BenTaylorDev:
Closing this as superseded — thanks for building it, and sorry it sat unreviewed this long.

Skill distribution moved off the SessionStart-clone approach shortly after this was opened:

- `c2b01c7182` — "feat(plugin): add Claude Code plugin + marketplace manifests" added `.claude-plugin/marketplace.json` and `plugin.json`, so skills ship through Claude Code's plugin mechanism.
- `179242cd82` — "fix(skills): remove internal-only skills from public repo" removed the staff-only skills from `.claude/skills/`, noting they "now live in the internal-skills plugin."

That covers the same need this PR was solving — getting internal skills into sessions without publishing them here — through a supported extension point rather than a repo-level hook that clones a private repo at session start.

The branch has also drifted well behind `main` and no longer merges cleanly; a re-land would be a fresh branch rather than a rebase.

If the plugin path turns out not to reach cloud sessions and we do want a hook, two things from the original script are worth carrying forward into a new PR:

1. The token is written into `$CACHE_DIR/.git/config` before the scrub, and with `set -euo pipefail` any failure in the intervening `fetch`/`reset` exits before that scrub runs — leaving it on disk. A `trap … EXIT`, or avoiding persistence via `-c http.extraHeader=`, closes that.
2. A committed `.claude/settings.json` is shared with every contributor and fork (`.gitignore` only covers `.claude/*.local.json`), so a repo-level SessionStart hook wants an explicit owner sign-off before landing.

PR #4220, opened 2026-04-23
chore(showcase): make registry + demo-content bundles content-addressable

samjulien:
## Summary

Alternative to #4182 — simpler approach, per @atai-s suggestion to "remove generated_at."

The committed `registry.json` / `constraints.json` / `demo-content.json` under each shell were churning in `git status` every dev/build cycle. Two root causes:

1. **Timestamp.** Both generators stamped `generated_at: new Date().toISOString()` on every run, so every regeneration produced a diff even with zero content change.
2. **Formatter drift.** CI's `static_quality` workflow runs `oxfmt --write .` and commits any diff as "style: auto-fix formatting". Oxfmt collapses short arrays (`"tags": ["chat-ui"]`) onto one line; the generator emits vanilla `JSON.stringify` output (always multi-line arrays). Every fresh generator run disagreed with the committed, oxfmt-processed version.

This PR makes the generator output content-addressable — same inputs, byte-identical output, every time:

- Drop `generated_at` from `generate-registry.ts` and `bundle-demo-content.ts`
- Sort `fs.readdirSync(PACKAGES_DIR)` in `bundle-demo-content.ts` (platform-dependent otherwise: alphabetical on macOS/APFS, hash-order on Linux/ext4 — would permute top-level `demos` keys between dev machines and CI)
- Add the 7 generated data JSONs to `.oxfmtrc.json` `ignorePatterns` so CI stops reformatting them; the generator's output is now canonical
- Regenerate the committed baseline
- Drop `generated_at: string` from 5 consumer TS interfaces (nothing actually reads the field; verified via grep for `.generated_at` property access)
- Drop 2 test assertions on `generated_at`

## Why this instead of #4182

#4182 untracks the JSONs (`.gitignore` + `git rm --cached`) and wires `predev` hooks to keep fresh clones working. It's correct but heavier: it changed package.json scripts, added `.gitignore` entries, modified test hygiene plumbing, and leaves readers asking "where do these files come from?" This PR removes the actual source of churn and keeps the committed JSONs as a readable, testable baseline.

## Scope limits

`probe-docs.ts` (→ `docs-status.json`) and `bundle-starter-content.ts` (→ `starter-content.json`) still stamp timestamps. Same pattern, trivial follow-up — excluded from this PR to keep the diff focused.

## Test plan

- [x] Local: regenerated twice in a row, `diff -q` confirms byte-identical output
- [x] Local: full scripts vitest suite (24 files / 1141 tests) passes
- [ ] CI: `Showcase: Validate` + `static / quality` pass
- [ ] CI: `static_quality` does NOT produce a "style: auto-fix formatting" commit on top of this PR (confirms the oxfmt ignorePatterns addition is working)


samjulien:
Fixed by #4236

PR #4133, opened 2026-04-21
fix(showcase): prod-mode migration for mastra + claude-sdk-typescript starters

jpr5:
## Summary

- starter-claude-sdk-typescript: precompile TS at image build, run `node /app/dist/agent/index.js` at boot (was: `npx tsx agent/index.ts`).
- starter-mastra: `mastra build` at image build, `node /app/.mastra/output/index.mjs` at boot (was: `npx mastra dev`).
- package-claude-sdk-typescript: cold-start instrumentation to disambiguate mystery health-probe failure.
- starter-langgraph-typescript: pure regen from generator to pick up PR #4132 package changes.

## Why

Mirrors #4132 (langgraph-typescript). Dev-mode cold start exceeds 180s watchdog grace on Railway cold containers, producing 502 smoke test failures. Grace bumps do not fix it — the dev-time work (tsx-based compile / mastra rollup) needs to move to image build time so boot is a straight `node` invocation.

Per-slug opt-in lives in `getAgentBuildSteps()` / `getAgentBuildCopy()` in `showcase/scripts/generate-starters.ts`. Other TS starters (langgraph-typescript, etc.) get empty substitutions and their Dockerfile cache surface is unchanged.

## Test plan

- [x] `docker build --platform linux/amd64` green locally for starter-claude-sdk-typescript (tsc ~1.4s), starter-mastra (mastra build ~12.9s), and package-claude-sdk-typescript.
- [x] `pnpm -C showcase/scripts test` — 1079/1079 pass (2 tests updated to assert new prod-mode behavior).
- [x] `pnpm -C showcase/scripts validate-pins` — FAIL count held at 110, hash matches `fail-baseline.json` (no drift, no baseline update needed).
- [ ] Railway deploy + smoke test green for `showcase-starter-claude-sdk-typescript` and `showcase-starter-mastra`.
- [ ] Next deploy logs of `showcase-package-claude-sdk-typescript` reveal which phase consumes the 180s+ budget (pre-node / module-load / pre-Anthropic / listening timestamps).

PR #4235, opened 2026-04-24
fix(showcase/starters): reduce mastra watchdog grace 180s to 30s

jpr5:
## Summary

- Reduces mastra starter watchdog grace from 180s to 30s in
  `getWatchdogGraceSeconds()` and regenerates the starter
- The mastra starter already builds at image time via `mastra build`
  (landed in #4133) and boots via `node .mastra/output/index.mjs`,
  so startup is <10s — 180s grace was ~6x longer than needed

## Test plan

- [x] `generate-starters.test.ts` passes (266 tests)
- [x] `--check --slug mastra` confirms no drift
- [ ] Deploy to Railway and observe agent healthy within 30s grace

### DIFF
--- a/showcase/scripts/generate-starters.ts
+++ b/showcase/scripts/generate-starters.ts
@@ -663,7 +663,9 @@ function getAgentHealthPath(fw: FrameworkDef): string {
  */
 function getWatchdogGraceSeconds(fw: FrameworkDef): number {
   if (fw.slug.startsWith("langgraph-")) return 180;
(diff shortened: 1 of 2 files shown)


jpr5:
Closing — the pre-built index.mjs still takes ~450s to init on Railway (V8 JIT + Mastra framework startup on resource-constrained container). Grace=30 would re-enter the kill loop. PR #4234 (grace=600) is the correct band-aid. The real fix is profiling the Mastra framework's startup path.

PR #4248, opened 2026-04-24
docs: improve README documentation

bakrsabeeh:
Added contributing guidelines and support section to README.

---
*This contribution was made to improve the project's documentation and help new contributors.*

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,3 +196,8 @@ Here are a few useful resources to help you get started:
 ## 📄 License
 
 This repository's source code is available under the [MIT License](https://github.com/CopilotKit/CopilotKit/blob/main/LICENSE).
+
+
+## Getting Started
+
+Clone the repo and follow the installation instructions in the docs.


BenTaylorDev:
Thanks for taking a look at the README.

Closing this one. The substantive change is the five lines appended at the end:

```md
## Getting Started

Clone the repo and follow the installation instructions in the docs.
```

That lands after the License section rather than in the document flow, and it restates `## Quick Start`, which already sits near the top of the README and links out to the actual install steps — so it reads as a second, less specific entry point rather than an improvement on the existing one.

Separately, the branch has accumulated a long run of upstream merges, so the diff GitHub shows here spans several hundred files well outside the README and no longer resolves against `main`. That's not something worth untangling for this change.

If you'd like to improve the README's on-ramp, a small PR from a fresh branch that edits the existing `## Quick Start` section directly would be very reviewable — that's genuinely the part most worth sharpening.

PR #4258, opened 2026-04-24
refactor(runtime): Simplify connect and run handlers in the Intelligence runtime

MikeRyanDev:
## What does this PR do?

## Related PRs and Issues

## Checklist

### DIFF
--- a/packages/core/src/__tests__/intelligence-agent.test.ts
+++ b/packages/core/src/__tests__/intelligence-agent.test.ts
@@ -118,6 +118,7 @@ interface IntelligenceAgentTestAccess {
   canonicalRunId: string | null;
   config: unknown;
   connect(input: RunAgentInput): Observable<BaseEvent>;
+  messages: RunAgentInput["messages"];
   socket: MockSocket | null;
   threadId: string | undefined;
 }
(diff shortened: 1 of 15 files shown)


BenTaylorDev:
Actually, these are from claude:

handlers/intelligence/connect.ts previously returned 204 No Content when the platform raised a 404
  (isPlatformNotFoundError). The new implementation returns 404 with body { error: "Connect request
  rejected", message } for platform 400/401/403/404/409. Only ɵconnectThread resolving to null produces a
   204.

  On the client, IntelligenceAgent.requestJoinCredentials$ treats 204 as null → EMPTY (silent skip) but
  throws on any non-OK status. So a client calling /connect for a thread that doesn't exist on the
  platform will now error instead of silently no-op. Consumers that relied on the silent 404→204 behavior
   (e.g., speculative connects during first-load) will now surface an error.

  If this is intentional (i.e., the platform 404 now only happens in misconfiguration, not "thread not
  yet replicated"), 👍. If it isn't, keep mapping platform 404 → 204 so the client still sees null.

MikeRyanDev:
This is intentional under the new Intelligence connect contract.

In the current app-api implementation, `POST /api/threads/:threadId/connect` is credentials-only and will get-or-create the thread before minting the one-time join token. So a missing thread should no longer be represented as a normal platform `404` for speculative connect. The normal silent/no-stream path is still preserved separately: `ɵconnectThread` returning `null` maps to `204`, and `IntelligenceAgent.requestJoinCredentials$` treats `204` connect as `EMPTY`.

Given that, preserving platform `400/401/403/404/409` as non-OK responses is deliberate: those now indicate validation/auth/ownership/platform contract failures rather than “nothing to connect to.” Mapping platform `404` back to `204` would hide real app-api/runtime misconfiguration or ownership failures.

BenTaylorDev (approved):

PR #4259, opened 2026-04-24
fix: frontend tools missing on cross-page navigation

mxmzb:
## Summary

- **`useFrontendTool` and `useCopilotReadable` now use `useLayoutEffect`** instead of `useEffect` for registration. This ensures tools and context are in the store before any sibling `useEffect` (e.g. CopilotChat's `connectAgent`) fires in the same render cycle. `useAgentContext` already used `useLayoutEffect` — this aligns the others. The work is just array push/remove, so the synchronous cost is negligible.

- **`useAgent()` now returns `{ agent, runAgent }`**. Calling `agent.runAgent()` directly bypasses CopilotKit core and sends requests without frontend tools or context. The new `runAgent()` delegates to `copilotkit.runAgent({ agent })`, which collects all registered tools and context. This is additive — `agent.runAgent()` still works as before.

## Context

Reported by Deutsche Telekom: their HR onboarding app has a start page that navigates to a chat page. The chat page registers frontend tools via `useHumanInTheLoop` / `useFrontendTool` and context via `useAgentContext`. The first CopilotKit request fired without the tools because the registration effects hadn't synced by the time the connect/run fired.

Reproduction: https://github.com/CopilotKit/deep-agent-cpk-experiments/tree/main/app/client/src/tickets/tkt-page-nav-tool-loss

## Test plan

- [x] `use-frontend-tool-timing.test.tsx` — verifies tools are visible in `copilotkit.tools` during a sibling component's `useEffect`
- [x] `use-agent-run-agent.e2e.test.tsx` — verifies `runAgent()` from the hook includes frontend tools in the request payload
- [x] Existing e2e tests pass (`use-agent.e2e.test.tsx`, `use-agent-context-timing.e2e.test.tsx`)

### DIFF
--- a/packages/angular/src/lib/agent.ts
+++ b/packages/angular/src/lib/agent.ts
@@ -123,8 +123,8 @@ export class CopilotkitAgentFactory {
             transport: runtimeTransport,
           });
           // Apply current headers so runs/connects inherit them
-
           (provisional as any).headers = { ...headers };
+          this.#copilotkit.core.ensureToolMiddleware(provisional);
           lastAgentStore = new AgentStore(
             provisional,
             destroyRef,
@@ -146,6 +146,7 @@ export class CopilotkitAgentFactory {
         );
       }
 
+      this.#copilotkit.core.ensureToolMiddleware(abstractAgent);
       lastAgentStore = new AgentStore(
         abstractAgent,
         destroyRef,
--- a/packages/core/src/core/agent-registry.ts
+++ b/packages/core/src/core/agent-registry.ts
@@ -295,6 +295,7 @@ export class AgentRegistry {
               debug: rawDebug ? resolveDebugConfig(rawDebug) : undefined,
             });
             this.applyHeadersToAgent(agent);
+            this.core.ensureToolMiddleware(agent);
             return [id, agent];
           },
         ),
--- a/packages/core/src/core/core.ts
+++ b/packages/core/src/core/core.ts
@@ -943,6 +943,15 @@ export class CopilotKitCore {
   }
 
   /**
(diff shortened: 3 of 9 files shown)


BenTaylorDev:
**Closing as superseded — thank you, the diagnosis here was correct and it landed.**

**The layout-effect half shipped.** `f9b306aa4e` ("fix(react-core): register frontend tools before sibling effects run") re-derived this against current main and credits this PR. `packages/react-core/src/v2/hooks/use-frontend-tool.tsx` now registers in `useLayoutEffect`, with a regression test that mounts the consumer first. That closes the Deutsche Telekom cross-page-navigation report.

**The `useCopilotReadable` half was still open.** That hook moved to `packages/react-core/src/v1-deprecated/hooks/` and was still on `useEffect`. I confirmed the race reproduces there, and it is now fixed the same way in a follow-up.

**The `ensureToolMiddleware` half we are not taking.** Since this PR opened, `docs/programmatic-control.mdx` documented `agent.runAgent()` as the low-level path that deliberately does **not** execute frontend tools or chain follow-ups, with `copilotkit.runAgent({ agent })` as the recommended default. The middleware would push tool *definitions* into a raw `agent.runAgent()` request, but frontend tool handlers run in `RunHandler.processAgentResult`, which a raw run never reaches. The model could then emit a tool call with nothing on the client to execute it. That is a worse failure than the documented behavior.

The branch is also well behind main and conflicting, and it edits `packages/react-core/src/hooks/use-copilot-readable.ts`, a path that no longer exists.

If you still want tool injection on the direct path, that is worth its own issue — the design question is where the handler loop lives, not where the definitions get attached.

PR #4278, opened 2026-04-24
feat(shell-docs): Built-in Agent as first-class integration + variant-slug resolution

samjulien:
## Summary

- Registers Built-in Agent (BIA) as a first-class integration in the showcase registry (slug `built-in-agent`, `popular` category, sort=0, logo asset). `/built-in-agent/*` URLs now resolve through the framework-scoped router with BIA-specific content merged into the sidebar under a dedicated section positioned after \"App Control\".
- Resolves registry slug → docs folder mismatches end-to-end. The three LangGraph variants (`langgraph-python`, `langgraph-typescript`, `langgraph-fastapi`), the two Microsoft Agent Framework variants (`ms-agent-dotnet`, `ms-agent-python`), plus the legacy `google-adk` / `strands` renames all now reach their framework-unique content (e.g. `/langgraph-python/auth`, `/ms-agent-dotnet/auth`), and in-page `<Tabs groupId=\"...\">` pre-select the tab that matches the URL variant.
- Replaces bare 404s with a framework-scoped \"not available for this framework\" fallback page when the requested slug exists under a *different* integration (e.g. visiting `/mastra/advanced-configuration` when the topic only lives under BIA). Sidebar stays intact; the body lists the integrations where the page does exist.
- Drops the `deployed`-gated \"coming soon\" styling from the docs framework picker and sidebar selector. That flag tracks whether a live showcase demo exists — a showcase concern, not a docs concern.
- Wires the real `TailoredContent` / `TailoredContentOption` components into the MDX registry. They were stubbed as passthrough `<div>{children}</div>`, which made pages that use the switcher (Readables, display-components, etc.) render every option stacked — the source of the visible \"duplicate content\" on those pages.
- Deletes 8 per-framework `integrations/<fw>/(other)/` subtrees whose content was byte-identical (or trivially diverged) to root's canonical `(other)/contributing/` + `(other)/telemetry/` copy (~2.5k lines removed). Adds a path-exclusion filter to `sync-docs-from-main.ts` so future upstream edits don't resurrect them.
- Strips dangling MDX import tails from 18 synced pages — `Symbol, Symbol, } from \"...\"` orphans left at the top of files by an older sync-stripper version. Current stripper handles multi-line imports correctly.
- Drops per-framework section headers from the override nav to avoid React key collisions with root-nav sections of the same name (e.g. every framework's `---Getting Started---` used to collide with root's).

## Test plan

[...]

PR #4279, opened 2026-04-24
ci(showcase): exempt stub packages from e2e baseline check

samjulien:
## Summary

- The `Showcase: Validate` workflow's \"Enforce e2e spec count\" step iterates every `showcase/packages/*/` directory and requires each to have a `tests/e2e/` directory with N specs. Built-in Agent was registered as a manifest-only stub in #4278 so the docs layer could reference it; the showcase implementation is owned by a different workstream and hasn't landed. The baseline check therefore fails every run on `main`.
- Add a skip in the bash loop for packages marked `deployed: false` in `manifest.yaml`. That flag is the existing semantic marker for \"registered but no showcase implementation\" — validated as a boolean by `showcase/scripts/lib/manifest.ts`, already consumed by shell and shell-docs to gate pickability and filtering. Stub packages legitimately have no demos / specs / qa to enforce a baseline on.
- `validate-parity.ts` still emits a SHOULD warning (\"demo count 0 deviates from baseline 9\") for stub packages so the gap stays visible to the team without gating CI.

## Test plan

- [ ] CI passes on this branch (Showcase: Validate is the only step that was failing on main).
- [ ] Run locally against every manifest:
  ```bash
  for m in showcase/packages/*/manifest.yaml; do
    pkg=$(basename $(dirname \"$m\"))
    if grep -qiE '^deployed:[[:space:]]*false[[:space:]]*$' \"$m\"; then echo \"SKIP  $pkg\"
    elif grep -qiE '^deployed:[[:space:]]*true[[:space:]]*$' \"$m\"; then echo \"CHECK $pkg\"
    else echo \"???   $pkg\"; fi
  done | sort
  ```
  Result: 17 CHECK, `built-in-agent` alone SKIP.

### DIFF
--- a/.github/workflows/showcase_validate.yml
+++ b/.github/workflows/showcase_validate.yml
@@ -121,6 +121,19 @@ jobs:
             [ -d "$pkg_dir" ] || continue
             found=$((found + 1))
             pkg=$(basename "$pkg_dir")
+            # Skip stub packages — those registered in the integration
+            # registry for docs-side references but without a showcase
+            # implementation yet. `deployed: false` in manifest.yaml is the
+            # semantic marker for that state (validated as a boolean by
+            # showcase/scripts/lib/manifest.ts). These packages legitimately
+            # have no demos / specs / e2e tests to enforce a baseline on;
+            # validate-parity still emits a SHOULD warning for them so the
+            # gap stays visible without gating CI.
+            manifest="${pkg_dir}manifest.yaml"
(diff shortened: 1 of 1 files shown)


samjulien:
Closing as obsolete. The same problem was solved on `main` in commit b9d56c2fe (Jordan, 2026-04-24) with a slightly different approach — checking for absence of `src/` instead of `deployed: false` in manifest.yaml. That approach is structurally cleaner for our goal: once we add code under `showcase/packages/built-in-agent/src/`, enforcement auto-enables, no manual flag-flip required. Superseded.

PR #4314, opened 2026-04-27
Remove obsolete demo asset

VZBYang:
Remove the obsolete demo.gif asset from the main branch.


BenTaylorDev:
Thanks for spotting this — you were right, and it's been dead weight for a long time. I dated it: `assets/demo.gif` was referenced by the README until `446ad06afc` (2023-11-29) removed the `` tag, so it's been unreferenced for about two years and eight months. Nothing else in the tree points at it, nothing builds or copies the root `assets/` dir, and no CI check depends on it.

I'm closing this PR but taking the deletion — it's in #6448, along with three more unreferenced files in the same directory, and your name is on it. The reason for moving it rather than merging here is purely mechanical: this branch carries 746 commits that aren't on `main` (234 merges, plus 512 commits whose subjects don't appear in `main`'s history at all — most likely a history rewrite upstream of your fork point). Squash and rebase are both disabled on this repo, so the only available merge method would graft all of those onto `main` in exchange for a three-line deletion.

Nothing you did wrong — that's fork-sync drift, not a mistake. If you'd rather the commit land under your own authorship, a fresh branch off current `main` with just that one deletion would merge immediately and I'll drop `demo.gif` from #6448. Either way, thanks for the cleanup.

PR #4354, opened 2026-04-28
feat(langgraph): add state injection to copilotkit middleware

ranst91:


packages/sdk-js/src/langgraph/middleware.ts
```diff
@@ -234,9 +333,33 @@ const middlewareInput = {
...
+ * const middleware = createCopilotkitMiddleware({
+ *   exposeState: ["liked", "todos"],
+ * });
+ * ```
+ */
+export const createCopilotkitMiddleware = (
+  options: { exposeState?: ExposeStateOption } = {},
+) => {
+  const exposeState = options.exposeState ?? false;
+  return createMiddleware(buildMiddlewareInput(exposeState) as any);
```
AlemTuzlak, line 357: why the as any here?

AlemTuzlak (approved):

PR #4420, opened 2026-04-29
feat(runtime): configurable MCP server on BuiltInAgent (Bearer + per-call user header)

lukasmoschitz:
## Summary

Closes [CPK-7526](https://linear.app/copilotkit/issue/CPK-7526).

`@copilotkit/runtime/v2`'s `BuiltInAgent` already supported MCP servers, but only with **static** headers set at client construction. CopilotKit Intelligence's `/mcp` endpoint requires a two-axis auth contract that breaks under static headers:

1. **Static project Bearer** — `Authorization: Bearer cpk-{projectId}_{shortToken}_{longToken}`. Same on every call.
2. **Per-call end-user header** — `X-Cpki-User-Id`. Fresh per outbound MCP HTTP request — never cached. Determines which user's persistent bash sandbox + `/threads` view a call sees.

Without per-call user-id, every demo session for a given project would share one bash sandbox, breaking the user-isolation contract that `cpki.shell_contexts` and `ThreadsFs` enforce on the platform side.

This PR adds the missing per-call resolver hook and is the precondition for [CPK-7527](https://linear.app/copilotkit/issue/CPK-7527) (Phase 4b SL integration), which wires it into `Intelligence/demos/simple-agent`.

### Public API surface

```ts
import {
  BuiltInAgent,
  INTELLIGENCE_USER_ID_HEADER,
  // re-exported from @ai-sdk/mcp:
  type MCPClient,
  type MCPTransport,
  type OAuthClientProvider,
} from "@copilotkit/runtime/v2";

new BuiltInAgent({
  model: "openai/gpt-4o",
  mcpServers: [{
    type: "http",
    url: `${process.env.INTELLIGENCE_API_URL}/mcp`,
    authToken: process.env.INTELLIGENCE_API_KEY!,
    getHeaders: ({ requestHeaders }) => {
      const userId = requestHeaders[INTELLIGENCE_USER_ID_HEADER]?.trim();
      if (!userId) throw new Error("missing user-id");
      return { [INTELLIGENCE_USER_ID_HEADER]: userId };
    },
  }],
});
```

- Unified `MCPClientConfig` shape mirrors `@ai-sdk/mcp`'s `MCPTransportConfig` (single discriminated `type: "http" | "sse"`, `url`, `headers`, `authProvider`) plus two CopilotKit extensions: **`authToken`** (static Bearer shorthand) and **`getHeaders`** (per-call resolver invoked on every outbound HTTP request — initialize, tools/list, tools/call, reconnects).
- **`MCPRequestContext`** carries `requestHeaders` (snapshot of the agent's per-run forwarded headers), `input`, and `mcpServerUrl`.
- **`MCPHeaderResolverError`** wraps resolver throws so `RUN_ERROR` attributes the failure to the resolver via ES2022 `Error.cause`.
- **`BuiltInAgent.headers: Record<string, string> = {}`** is now declared on the class. The runtime's existing `extractForwardableHeaders` feature-detect (`if (agent.headers)` in `configureAgentForRequest`) only forwards headers to agents that already declare a truthy field — initializing to `{}` activates that path so the BFF's per-request headers reach the resolver via context.
- **`INTELLIGENCE_USER_ID_HEADER`** constant exported as the canonical header name (`"x-cpki-user-id"`) — re-exported through both `intelligence-platform` and `v2/runtime` barrels.

### Architecture

The agent's MCP layer now sits directly on `@ai-sdk/mcp`'s stable surface (`createMCPClient`, `MCPClient`, `MCPTransport`) instead of bypassing it to `@modelcontextprotocol/sdk`. The agent module no longer imports `@modelcontextprotocol/sdk` at all — that's confined to a small `CopilotKitMCPTransport` class that implements Vercel's `MCPTransport` interface and is only instantiated when a `getHeaders` resolver is present.

For static-only configs (no per-call hooks), the config is handed straight to `createMCPClient` and Vercel's built-in `HttpMCPTransport` / `SseMCPTransport` handle the wire.

`MCPClient`, `MCPTransport`, `OAuthClientProvider`, `OAuthTokens`, and `UnauthorizedError` are re-exported from `@copilotkit/runtime/v2` so consumers don't need a direct `@ai-sdk/mcp` dependency.

### Side-effect bug fix: SSE static headers

The pre-existing direct-SDK construction passed `serverConfig.headers` as the wrong-shape options arg to `SSEClientTransport`, and the SDK silently ignored it — auth on SSE simply didn't work. Vercel's `SseMCPTransport` correctly applies `headers` via its `commonHeaders()` pipeline. The new architecture inherits the fix; a regression test in `mcp-servers-integration.test.ts` proves static `headers` now reach the wire on the SSE path.

### What's intentionally NOT in here

- **Demo wiring** in `Intelligence/demos/simple-agent` (CPK-7527).
- **MCP support for LangGraph / Mastra** runtimes — different integration points; not this ticket.
- **Server-side `/mcp` route changes** in Intelligence (already shipped on `mme/integrate-sl`).

## Notes for reviewers

- **Public-API addition:** `BuiltInAgent.headers` is a real public field, not just an ad-hoc property. It's load-bearing for the runtime's existing header-forwarding feature-detect.
- **Test header inspection:** `aimock` redacts `Authorization` to `[REDACTED]` in its journal, so tests that need to verify the actual outgoing Bearer use a `vi.spyOn(globalThis, "fetch")` recorder instead of the journal. The `x-cpki-user-id` header isn't redacted, so per-call user-id assertions read from the journal directly.
- **No `toMCPServer()` helper.** Earlier iterations of this PR shipped a `CopilotKitIntelligence.toMCPServer()` shortcut. It conflated thread management (the class's actual job) with MCP transport configuration and baked in opinionated behavior that not every deployment will want; dropped in favor of the inline pattern shown above using the exported `INTELLIGENCE_USER_ID_HEADER` constant. Same line count as the helper-call form, more flexible.
- **Pre-existing `tsc --noEmit` OOM:** `pnpm nx run @copilotkit/runtime:check-types` runs out of heap on `main` at 8GB. Reproduces on a clean checkout without any changes from this PR. Worth tracking as a separate issue. Build (`tsdown`) is unaffected; the full vitest suite (1419 tests, 101 files) passes.

## Test plan

- [x] `pnpm nx run @copilotkit/runtime:test -- mcp-servers-integration` — 15 cases pass (8 existing + 6 per-call/static-header + 1 new SSE regression test)
- [x] `pnpm nx run @copilotkit/runtime:test -- mcp-clients` — 8 cases pass (existing user-managed-clients suite, including `MCPClient` ↔ `MCPClientProvider` type-compat check)
- [x] `pnpm nx run @copilotkit/runtime:test --skip-nx-cache` — full suite 1419/1419
- [x] `pnpm nx run @copilotkit/runtime:build` — `tsdown` clean
- [x] `oxlint` on changed files — 0 errors
- [ ] Manual end-to-end against a local Intel server + simple-agent BFF — deferred to CPK-7527 since the demo wiring lives there


docs/content/docs/integrations/built-in-agent/mcp-servers.mdx
```diff
@@ -69,6 +69,100 @@ const builtInAgent = new BuiltInAgent({
...
+Some MCP servers need a header that varies per call — typically an **end-user identifier** so each user gets their own isolated server-side state. Use `getHeaders` for this. The resolver runs on **every** outbound MCP HTTP request (initialize, tools/list, tools/call, reconnects), so the value is never cached across calls:
+
+```typescript
+const builtInAgent = new BuiltInAgent({
+  model: "openai:gpt-5.4-mini",
+  mcpServers: [
+    {
+      type: "http",
+      url: "https://intel-platform.example.com/mcp",
+      authToken: process.env.INTELLIGENCE_API_KEY,
```
mme: Give a more generic example, i.e. my-mcp server like the prev one, also MCP_API_KEY etc

docs/content/docs/integrations/built-in-agent/mcp-servers.mdx
```diff
@@ -69,6 +69,100 @@ const builtInAgent = new BuiltInAgent({
...
+import {
+  BuiltInAgent,
+  INTELLIGENCE_USER_ID_HEADER,
+} from "@copilotkit/runtime/v2";
+
+const builtInAgent = new BuiltInAgent({
+  model: "openai:gpt-5.4-mini",
+  mcpServers: [
+    {
+      type: "http",
```
mme: Put this into a helper function

docs/content/docs/integrations/built-in-agent/mcp-servers.mdx
```diff
@@ -69,6 +69,100 @@ const builtInAgent = new BuiltInAgent({
...
+        }
+        return { [INTELLIGENCE_USER_ID_HEADER]: userId };
+      },
+    },
+  ],
+});
+```
+
+The throw on missing user-id is deliberate: silent fallthrough to an empty user-id would collapse every browser session into one shared server-side state, defeating the very isolation guarantee the per-call header exists to enforce. Make sure your BFF sets `X-Cpki-User-Id` on its request to the runtime.
+
```
mme: Don't document `X-Cpki-User-Id` here - CPK Intelligence has a callback that passes returns the user id. Get it from there and pass it in the MCP config.

docs/content/docs/integrations/built-in-agent/mcp-servers.mdx
```diff
@@ -214,5 +316,11 @@ const builtInAgent = new BuiltInAgent({
 |----------|-----|------|-------------|
 | `type` | `"sse"` | `"http"` | Transport type |
 | `url` | string | string | MCP server URL |
-| `headers` | `Record<string, string>` | — | Custom HTTP headers (SSE only) |
-| `options` | — | `StreamableHTTPClientTransportOptions` | Transport options (HTTP only) |
+| `headers` | `Record<string, string>` | `Record<string, string>` | Static HTTP headers, sent on every outbound request to the server. |
+| `authProvider` | `OAuthClientProvider` | `OAuthClientProvider` | OAuth client provider for credential refresh flows. The provider's `tokens()` is called per outbound request. Re-exported from `@ai-sdk/mcp`. |
+| `authToken` | — | string | Static Bearer-token shorthand. Expands to `Authorization: Bearer <authToken>` on the outgoing headers. |
```
mme: Remove `authToken` shorthand

docs/content/docs/integrations/built-in-agent/mcp-servers.mdx
```diff
@@ -214,5 +316,11 @@ const builtInAgent = new BuiltInAgent({
 |----------|-----|------|-------------|
 | `type` | `"sse"` | `"http"` | Transport type |
 | `url` | string | string | MCP server URL |
-| `headers` | `Record<string, string>` | — | Custom HTTP headers (SSE only) |
-| `options` | — | `StreamableHTTPClientTransportOptions` | Transport options (HTTP only) |
```
mme, line 218: Don't remove options (no breaking changes)

packages/runtime/src/agent/index.ts
```diff
@@ -113,58 +111,119 @@ export type BuiltInAgentModel =
...
+ * transport config does not cover natively.
+ *
+ * - When `getHeaders` is set, the agent uses {@link CopilotKitMCPTransport}
+ *   so the resolver runs on every outbound HTTP request (initialize,
+ *   tools/list, tools/call, reconnects). HTTP only.
+ * - Otherwise, the config is passed straight to `@ai-sdk/mcp`'s
+ *   `createMCPClient`, which selects its built-in HTTP or SSE transport.
+ *   Static `headers` and the OAuth-style `authProvider` are honored on both.
  */
-export interface MCPClientConfigHTTP {
```
mme, line 119: Ensure no breaking changes- double check no type or config option gets removed (in the whole file)

packages/runtime/src/agent/index.ts
```diff
@@ -113,58 +111,119 @@ export type BuiltInAgentModel =
...
+  authProvider?: OAuthClientProvider;
   /**
-   * URL of the MCP server
+   * Static auth token. Shorthand that expands to
+   * `Authorization: Bearer <authToken>` before the config reaches the
+   * transport. Use this for project-level credentials that don't vary per
+   * call (e.g. a CopilotKit Intelligence project key).
    */
-  url: string;
+  authToken?: string;
```
mme: No `authToken` shorthand if this is also possible via plain `headers`

packages/runtime/src/v2/runtime/intelligence-platform/client.ts
```diff
@@ -1,5 +1,12 @@
 import { logger } from "@copilotkit/shared";
 
+/**
+ * Header name carrying the per-call end-user identity that the
+ * CopilotKit Intelligence `/mcp` endpoint requires. Lower-case to match the
+ * runtime's forwarded-header normalization.
+ */
+export const INTELLIGENCE_USER_ID_HEADER = "x-cpki-user-id";
+
```
mme, line 14: encapsulate this in the intelligence helper

BenTaylorDev:
Things Claude wanted me to tell you:
 - The cast in handlers/intelligence/run.ts:87 — (agent as unknown as { user?: { id, name } }).user = ....
  MiddlewareCapableAgent already extends AbstractAgent with user? in agent-utils.ts:16. The handler should
  accept that type (or a narrow shared type) instead of double-casting. Loses type safety for no real reason.
  - INTELLIGENCE_USER_ID_HEADER is hard-coded inside client.ts:9 — fine for encapsulation, but the helper is
  closed over a single header name. If/when a self-hosted Intelligence variant wants a different name, this
  becomes a fork point. Probably worth a comment that it's deliberately not configurable.
  - Concurrency assumption is implicit: setting agent.user before agent.run() only works because
  cloneAgentForRequest returns a fresh instance per request. Worth a one-liner on BuiltInAgent.user saying
  "set per request on a cloned agent; do not set on a shared singleton."
  - Doc-table footnote: the SSE row says "if you need it, use HTTP" but the SSE bug-fix commit means SSE now
  applies static headers correctly. A tiny note that SSE supports static headers and only getHeaders is
  HTTP-only would prevent confusion.

AlemTuzlak (approved):

PR #4445, opened 2026-04-29
WIP spike: OSS-69 host-portal pattern for shared gen-ui in inspector

marthakelly:
## What this is

Recon prototype for [OSS-69](https://linear.app/copilotkit/issue/OSS-69/recon-options-for-sharing-gen-ui-components-in-intelligence-inspector). Not feature-ready. Posted to make the pattern concrete for the team discussion. The full options analysis (with pros/cons of iframe, schema-driven DSL, snapshot capture, etc.) lives in the [Notion recon doc](https://app.notion.com/p/3513aa381852810890daecbf2ed4addf).

Recommendation in the recon: Option A, the host-portal pattern. This PR proves it works.

## How it works

The inspector renders the customer's own gen-ui without taking a React or Angular dependency.

1. **`@copilotkit/core`**: `ToolRenderBridge` interface + `ToolRenderRegistry` class, with `addToolRenderBridge` / `canRenderTool` / `attachToolRender` / `detachToolRender` on `CopilotKitCore`. Pure DOM contract: `(req, hostEl: HTMLElement) => mount/unmount`. No framework references anywhere in core.

2. **`@copilotkit/react-core`**: `HostPortalBridge` component mounted from `CopilotKitProvider`. On mount, registers a bridge with core. When asked to fill a host element, it `createPortal`s the matching `renderToolCalls` renderer into that element. Portal preserves React context, so the customer's hooks, providers, and theming all work. Subscribes to `onRenderToolCallsChanged` so live status flows through.

3. **`@copilotkit/web-inspector`**: `cpk-tool-render-slot` custom element. The inspector probes `core.canRenderTool` and emits the slot when a host renderer exists. Falls back to today's JSON view when none is registered. Wired into both rendering paths: the legacy "Current Messages" view and the threads-tab Conversation sub-tab from the threads PR.

4. **Demo**: `examples/v2/react/demo/src/app/sidebar` registers a `showChart` frontend tool plus a `useRenderTool` returning an inline-SVG bar chart, so the chart shows up in the chat AND in the inspector's tool-call view.

## Why this option

Option A uniquely satisfies all three constraints from the recon:
- Zero new render code in the inspector bundle (just a slot custom element + a `canRenderTool` probe).
- Zero drift with the live customer app, because it IS the customer's renderer.
- Works for in-memory runner (no Intelligence dependency).
- Graceful JSON fallback when the host has no renderer for a given tool name.

## Future framework support

The bridge contract is pure DOM, so Vue / Svelte / Solid / vanilla all map onto it without changes to core or inspector. Adding a new SDK is one bridge file in that SDK's package. See the [Notion doc's "Future framework support" section](https://app.notion.com/p/3513aa381852810890daecbf2ed4addf) for the per-framework mount-API mapping.

## What's intentionally out of scope

- Angular bridge. Same shape as the React bridge (`ViewContainerRef.createComponent`), not built here.
- Vue / Svelte / Solid bridges.
- Intelligence-runner demo. The pattern is runtime-agnostic by design (the bridge is client-side; tool-call data shape is the same), but this demo currently uses `InMemoryAgentRunner`. Proving the Intelligence path end-to-end depends on CPK-7453 anyway.
- Tests.
- Tool result threading. The slot currently passes args; tool result is plumbed through but conversation builder doesn't always populate it for frontend tools (see "Known issues" below).

## Demo-only changes that aren't part of the pattern

- `openGenerativeUI: false` in `examples/v2/react/demo/src/app/api/copilotkit/[[...slug]]/route.ts`. The built-in `generateSandboxedUI` tool out-ranks `showChart` for any chart request. Flipping it off lets the demo's tool fire so we can see the slot light up. Flip back to `true` to restore default runtime behavior.
- `BarChart` in the sidebar demo uses inline styles, not Tailwind classes. Tailwind class names from the host page do not penetrate into the inspector's shadow DOM (Lit). Inline styles render correctly across the boundary. This is a per-renderer concern customers will hit when their gen-ui shows up in the inspector; worth a docs note in the eventual feature PR.

## Known issues that reproduce without this spike

- Tool block shows "PENDING" badge even after a frontend tool's handler returns. Threads PR conversation-builder logic (`item.result` truthiness check). Not touched by the spike.
- Chat conversation appears to clear on run completion. Reproduces on the threads PR branch without the spike applied (`HostPortalBridge` is a sibling of `children`, not a wrapper, so its state changes do not cascade into the chat subtree).

## How to run the demo

```bash
cd examples/v2/react/demo
pnpm dev
```

Then open `http://localhost:3000/sidebar` and prompt:

> Use the showChart tool with title "Fruit Sales" and bars apples 12, bananas 8, cherries 15

You should see the chart in the sidebar chat AND in the inspector's tool block (auto-expanded since a host renderer is wired). Comment out the `useRenderTool({ name: "showChart", ... })` block in the sidebar page and the inspector falls back to today's JSON view.

## Diff stat

8 files, ~535 lines. Three new files (bridge contract, React portal, inspector slot), three small edits (`CopilotKitCore`, `CopilotKitProvider`, threads-tab `renderToolBlock`), two demo edits.

## Base branch

This PR targets `feat/CPK-7193-inspector-threads-clean` because the spike wires the slot into the threads-tab conversation view. Mergeable in sequence after the threads PR.

### DIFF
--- a/examples/v2/react/demo/src/app/sidebar/page.tsx
+++ b/examples/v2/react/demo/src/app/sidebar/page.tsx
@@ -6,6 +6,7 @@ import {
   defineToolCallRenderer,
   useConfigureSuggestions,
   useFrontendTool,
+  useRenderTool,
 } from "@copilotkit/react-core/v2";
 import { z } from "zod";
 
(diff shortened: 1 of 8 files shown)


BenTaylorDev:
Closing as abandoned/superseded — this WIP spike (OSS-69 host-portal prototype) was overtaken by the inspector direction that shipped as `packages/web-inspector`; the prototype path is dead. Reopen if this analysis missed something.

PR #4462, opened 2026-04-30
fix(suggestions): show available:"always" pills on welcome screen with runtimeUrl

tylerslaton:
## Summary

`available: "always"` suggestion configs (static and dynamic) didn't render on the welcome screen when the chat used `runtimeUrl` to fetch agents instead of registering them locally with `agents__unsafe_dev_only`.

The bug has been latent since v2 first landed (Dec 2025); a per-thread cloning change masked it for some flows from Mar 31 → Apr 23, and the Apr 23 revert (#3525 backout) re-exposed it.

This PR is welcome-screen only — the `!isConnecting && !isRunning` UI gate is unchanged, so suggestions still hide during connect/replay and during runs as before.

## What was broken

With `runtimeUrl`, the agents registry is empty during the initial `/info` fetch. Two compounding issues meant `available: "always"` configs never got off the ground:

1. **`SuggestionEngine.reloadSuggestions`** bailed early when the consumer agent wasn't in the registry yet. The first reload fires from `useConfigureSuggestions` on mount — at that moment the registry is empty, so every config got skipped. Static pills never appeared, and dynamic pills never even started generating.
2. **`useConfigureSuggestions`'s global-config path** (no `consumerAgentId` or `"*"`) only iterated the current agents map. Empty map → zero reload calls → suggestions stuck empty until something else triggered a reload.

## Fix

Three small changes, scoped to the welcome screen path:

1. `SuggestionEngine.reloadSuggestions` no longer bails when the agent's missing — defaults `messageCount` to 0 and processes static configs anyway. Dynamic configs still skip until a real agent arrives.
2. `useConfigureSuggestions`'s global path also calls `reloadSuggestions(targetAgentId)` directly (covers the empty-map case where the agent the chat is bound to isn't yet in the registry).
3. `useConfigureSuggestions` subscribes to `onAgentsChanged` *only* for dynamic configs, *only* when the target agent isn't yet present, and *unsubscribes after firing once*. Dynamic pills catch up after the runtime fetch completes, without piling up overlapping generations as multiple hooks mount.

## What's preserved

- `hasSuggestions` keeps `!isConnecting && !isRunning` — bootstrap replay and run-in-flight both still hide pills (no mid-replay layout jump, no stale-context flash mid-run).
- `available: "always"` is the *eligibility window* (welcome screen vs after first message), not a "render through transitions" override.
- Threading behavior (thread switch, explicit `threadId`, multi-chat) is unchanged. None of the new code paths fire on thread connects.

## Test plan

- [x] `packages/core` — engine unit tests for `reloadSuggestions` when no agent is present + when only static "always" config exists. All 59 core-suggestions tests pass.
- [x] `packages/react-core` — 4 integration tests in `CopilotChat.suggestionsAlways.test.tsx`:
  - shows on welcome screen with explicit `consumerAgentId`
  - shows on welcome screen with global config
  - hides during a run, reappears after (default scroll)
  - hides during a run, reappears after (pin-to-send)
- [x] All 1157 react-core tests pass.
- [ ] Manual smoke in `examples/v2/react/demo` with both static and dynamic `available: "always"` configs — welcome screen pills, hide during run, regenerate after.

### DIFF
--- a/packages/core/src/core/suggestion-engine.ts
+++ b/packages/core/src/core/suggestion-engine.ts
@@ -67,14 +67,15 @@ export class SuggestionEngine {
 
     // Use the provided agent instance when available; fall back to the registry agent.
     // Per-thread clones hold the actual conversation messages; the registry agent does not.
+    // The agent may legitimately be missing here (e.g. runtime info still loading,
+    // or the consumer agent is configured but not yet registered) — static suggestions
+    // don't need it, only dynamic generation does. Treat that case as "no messages yet"
+    // and process static configs anyway.
     const agent =
       consumerAgent ??
       (this.core as unknown as CopilotKitCoreFriendsAccess).getAgent(agentId);
-    if (!agent) {
-      return;
-    }
 
-    const messageCount = agent.messages?.length ?? 0;
+    const messageCount = agent?.messages?.length ?? 0;
     let hasAnySuggestions = false;
 
     for (const config of Object.values(this._suggestionsConfig)) {
@@ -95,6 +96,12 @@ export class SuggestionEngine {
       const suggestionId = randomUUID();
 
       if (isDynamicSuggestionsConfig(config)) {
+        // Dynamic suggestions need a real agent (provider + consumer messages).
+        // Skip when the agent isn't ready yet — `reloadSuggestions` will be
+        // called again once it is.
+        if (!agent) {
+          continue;
+        }
         if (!hasAnySuggestions) {
           hasAnySuggestions = true;
           void this.notifySuggestionsStartedLoading(agentId);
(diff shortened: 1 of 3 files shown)


packages/react-core/src/v2/components/chat/CopilotChatView.tsx
```diff
@@ -268,6 +268,11 @@ export function CopilotChatView({
   // Otherwise, mid-replay (bootstrap stream from /connect) or mid-run, the
   // suggestions would render against a still-assembling message tree and
   // visibly jump as each final text chunk reflows the layout.
+  //
+  // `available: "always"` controls *eligibility windows* (welcome screen vs
+  // after first message), not whether to render through these transitions —
+  // we still wait for the connect/run to settle and the end-of-run reload
+  // to repopulate against the new context.
```
BenTaylorDev: Is this an orphaned comment?

tylerslaton: Good catch — yes, orphaned. I added that paragraph when the PR was scoped to also override the gate for `available:"always"`, but after we pulled that back the comment was justifying behavior that doesn't exist. Removed it; the file now has zero net diff.

BenTaylorDev (approved):

PR #4490, opened 2026-04-30
fix(showcase): stop infinite tool-call loop in beautiful-chat + restore brand styling

AlemTuzlak:
## Summary

Two related Beautiful Chat bugs in the integration showcases:

### Bug 1 — Infinite tool-call loop on every suggestion click

In `showcase/aimock/feature-parity.json`, 10 tool-calling fixtures (`pieChart`, `barChart`, `render_pie_chart`×3, `render_bar_chart`×2, `scheduleTime`, `search_flights`, `toggleTheme`) lacked an explicit `id` on the `toolCall` and lacked a paired `toolCallId` followup fixture.

aimock's matcher (`router.js:36-46`) matches `userMessage` by **substring** against the **last** user message. After the agent ran the returned toolCall and re-prompted the LLM with the tool result attached, the last user message was unchanged → the same fixture matched → the same toolCall was returned forever. Repro: click any of the Pie Chart / Bar Chart / Schedule Meeting / Search Flights / Toggle Theme suggestions on `langgraph-python`'s `beautiful-chat` demo and watch the chart re-render until manually stopped.

Fix follows the existing convention in the same file (compare the `Ada Lovelace` / `show_card` pair):

- Added explicit `id` to each broken `toolCall` (`call_fp_pie_chart_001`, `call_fp_bar_chart_001`, …)
- Added 11 new `toolCallId` followup fixtures at the top of the file that match each id and respond with a brief `content` summary instead of another toolCall

This shared file is loaded by every integration's `beautiful-chat` demo, so the fix benefits all of them.

### Bug 2 — Black/white background split + broken CopilotKit logo

8 integrations use the full `ExampleLayout` pattern with `ThemeProvider` and an `` (crewai-crews, langgraph-fastapi, langgraph-python, langgraph-typescript, mastra, ms-agent-dotnet, ms-agent-python, pydantic-ai). Two issues:

1. `globals.css` hardcoded `body { background: #fafaf9 }` and never defined the brand tokens (`--background`, `--foreground`, `--card`, `--primary`, `--secondary`, `--muted`, `--accent`, `--destructive`, `--border`, `--input`, `--ring`, `--radius`) that the layout, mode-toggle, todo card/column, and chart components reference via Tailwind 4 arbitrary values. `ThemeProvider` was also adding `dark` to `<html>` from system preference, so CopilotKit's chat went dark via `[data-copilotkit] { background: var(--background) }` while body stayed cream — that's the visible split the user reported.
2. `example-layout/index.tsx` renders `` but the file did not exist in any integration's `public/`.

Fix per integration:

- Added the full token set (light + dark) under `:root` and `:root.dark, .dark`, registered the Tailwind 4 dark variant with `@custom-variant dark (&:where(.dark, .dark *));`, switched `body` to `var(--background)`/`var(--foreground)` so it tracks the same `dark` class CopilotKit's chat is already responding to.
- Copied `copilotkit-logo.svg` and `copilotkit-logo-mark.svg` into each integration's `public/` from `examples/integrations/langgraph-python/public/`.

The 10 simplified-layout integrations (ag2, agno, built-in-agent, claude-sdk-*, google-adk, langroid, llamaindex, spring-ai, strands) don't render the logo or canvas split, so they don't need either fix.

## Verification

- `feature-parity.json`, `d5-all.json`, `smoke.json` parse cleanly
- `pnpm tsx showcase/scripts/validate-fixture-tool-surface.ts` → `104 fixtures × 603 demos — no drift`
- Programmatic check: every `toolCalls[].id` in `feature-parity.json` now has a paired `toolCallId` followup fixture (no orphans)
- All 8 updated `globals.css` files have balanced braces and the `:root.dark` block

## Test plan

- [ ] Click each of the 9 suggestion pills in `langgraph-python` `beautiful-chat`; confirm tool-using suggestions no longer loop and end with a text summary
- [ ] Toggle system theme → confirm chat side and canvas/body side stay in sync (both light or both dark) on `langgraph-python` `beautiful-chat`
- [ ] CopilotKit logo renders in the chat header (no broken-image placeholder)
- [ ] Spot-check one Pattern A integration (`langgraph-fastapi` or `pydantic-ai`) and one Pattern B (`mastra` or `crewai-crews`) for the same behavior


AlemTuzlak:
CI was failing due to two pre-existing issues on `main` (unrelated to this PR's changes):

1. `Python unit tests (3.12)` — three langroid `test_agui_adapter.py` tests failing
2. `Validate Showcase` — validate-pins drift hash mismatch (count held at 134, hash shifted)

Both were already fixed on the open #4489 chore branch. Cherry-picked the two CI-fix commits (`d34bb1ac` langroid debug-log fix, `bea43fb9` ratchet baseline hash) onto this branch to unblock the run. They'll dedupe cleanly when #4489 merges first or this rebases onto main.

ranst91 (approved):

PR #4503, opened 2026-04-30
Add Perplexity service adapter (Agent API)

jliounis:
## Summary

Adds a `PerplexityAdapter` for the Copilot Runtime, mirroring the existing `GroqAdapter` line-for-line. Both providers expose an OpenAI-compatible chat completions endpoint, so the adapter wraps the official `openai` SDK pointed at `https://api.perplexity.ai`.

- New adapter: `packages/runtime/src/service-adapters/perplexity/perplexity-adapter.ts`
- Default model: `sonar-pro`
- Sets the `X-Pplx-Integration: copilotkit/<version>` attribution header on the lazy-initialized client
- Exported from `packages/runtime/src/service-adapters/index.ts`
- Vitest unit tests under `packages/runtime/tests/service-adapters/perplexity/` covering forwarded options, default model, and presence of the attribution header

## Usage

```ts
import { CopilotRuntime, PerplexityAdapter } from "@copilotkit/runtime";
import OpenAI from "openai";

const perplexity = new OpenAI({
  apiKey: process.env.PERPLEXITY_API_KEY,
  baseURL: "https://api.perplexity.ai",
});

const copilotKit = new CopilotRuntime();
return new PerplexityAdapter({ perplexity, model: "sonar-pro" });
```

If no client is supplied, the adapter lazy-instantiates one against the Perplexity base URL with the attribution header pre-attached, picking up `PERPLEXITY_API_KEY` from the environment.

## Testing

- [x] `pnpm --filter=@copilotkit/runtime build` — TypeScript compiles cleanly; emitted `dist/service-adapters/perplexity/*` artifacts match other adapters.
- [x] Unit tests added under `tests/service-adapters/perplexity/` mirror the existing groq tests.
- [x] No new dependencies — reuses the `openai` SDK already declared in `@copilotkit/runtime`.

### DIFF
--- a/packages/runtime/src/service-adapters/index.ts
+++ b/packages/runtime/src/service-adapters/index.ts
@@ -12,6 +12,7 @@ export * from "./google/google-genai-adapter";
 export * from "./openai/openai-assistant-adapter";
 export * from "./unify/unify-adapter";
 export * from "./groq/groq-adapter";
+export * from "./perplexity/perplexity-adapter";
 export * from "./anthropic/anthropic-adapter";
 export * from "./experimental/ollama/ollama-adapter";
 export * from "./bedrock/bedrock-adapter";
--- a/packages/runtime/src/service-adapters/perplexity/perplexity-adapter.ts
+++ b/packages/runtime/src/service-adapters/perplexity/perplexity-adapter.ts
@@ -1,0 +1,247 @@
+/**
+ * Copilot Runtime adapter for Perplexity (Agent API).
+ *
+ * Wraps the official `openai` SDK pointed at `https://api.perplexity.ai`,
+ * which exposes an OpenAI-compatible chat completions endpoint.
+ *
+ * ## Example
+ *
+ * ```ts
+ * import { CopilotRuntime, PerplexityAdapter } from "@copilotkit/runtime";
+ * import OpenAI from "openai";
+ *
+ * const perplexity = new OpenAI({
+ *   apiKey: process.env["PERPLEXITY_API_KEY"],
+ *   baseURL: "https://api.perplexity.ai",
+ * });
+ *
+ * const copilotKit = new CopilotRuntime();
+ *
+ * return new PerplexityAdapter({ perplexity, model: "sonar-pro" });
+ * ```
+ */
+import type { LanguageModel } from "ai";
+import { createOpenAI } from "@ai-sdk/openai";
+import type OpenAI from "openai";
+import Openai from "openai";
+import {
+  CopilotServiceAdapter,
+  CopilotRuntimeChatCompletionRequest,
+  CopilotRuntimeChatCompletionResponse,
+} from "../service-adapter";
+import {
+  convertActionInputToOpenAITool,
+  convertMessageToOpenAIMessage,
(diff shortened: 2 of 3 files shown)


jliounis:
Friendly ping — the Vercel preview is blocked because the build runs from a fork (it needs a team member to click the authorize link in the vercel[bot] comment). The PR itself is purely code: a thin `PerplexityAdapter` mirroring `GroqAdapter`, no new deps. Claude bot review found no issues. Could a maintainer authorize the Vercel preview and take a look?

BenTaylorDev:
Hi @jliounis — first off, an apology: you pinged in May and nobody got back to you. That's on us, not you, and I'm sorry you were left waiting on a fork-preview authorization that was never the real blocker.

I've now gone through this properly. The adapter itself is a clean, faithful mirror of `GroqAdapter` — no scope creep, no new dependencies, tests included. Thank you for that. Unfortunately I'm going to close it, and I want to explain why rather than just letting it lapse.

**The v1 service-adapter surface has since been deprecated.** In August we formally deprecated the v1 SDK (#6582) and moved its source under `packages/runtime/src/v1-deprecated/`. The path this PR targets — `packages/runtime/src/service-adapters/` — no longer exists on `main`, and every file in that tree now carries a "use v2 instead" notice. Your PR predates all of that by several months, so this is purely a timing problem, not a quality one. But it does mean merging it would add brand-new public API to a surface we've just told people to migrate off, which isn't something I can do without a broader product decision.

**Two things worth knowing regardless of where this lands**, since they'd bite whoever picks this up next:

1. The lazy-init path doesn't read `PERPLEXITY_API_KEY`. `new OpenAI({ baseURL, defaultHeaders })` with no `apiKey` falls back to `process.env.OPENAI_API_KEY` — that's the OpenAI SDK's own default. So `new PerplexityAdapter()` either throws `Missing credentials … set the OPENAI_API_KEY environment variable` (confusing, wrong provider named), or — if the user happens to have `OPENAI_API_KEY` set, which is common in a CopilotKit runtime — it silently attaches their **OpenAI** key to a client pointed at `api.perplexity.ai`. I verified both against the real SDK. It's a one-line fix (`apiKey: process.env.PERPLEXITY_API_KEY`), but the test covering that path `vi.mock`s the `openai` module and stubs the constructor's `apiKey` default, so it can't catch it.

2. The `X-Pplx-Integration` attribution header only attaches on the lazy path. The usage shown in the PR description — passing your own pre-configured `OpenAI` client — bypasses it, so in practice attribution wouldn't fire for most users.

**If Perplexity support is something you'd still like to see**, the place for it now is the v2 runtime rather than a v1 service adapter, and I'd genuinely welcome that conversation — please open an issue and tag me. Before writing any code, though, it'd be worth confirming that the `sonar` models handle OpenAI-style function calling: this adapter forwards `tools`/`tool_choice` and has a full tool-call streaming branch, and CopilotKit's core feature is frontend actions. If `sonar-pro` ignores `tools`, that branch never fires and actions silently no-op. Nothing in the current tests exercises a live call, so that's unverified either way — you're much better placed than I am to answer it.

Thanks again for the contribution, and sorry again for the silence.

Issue #4215, opened 2026-04-23
CopilotChat textarea can't be driven by Playwright in headless Chromium (React onChange not firing)


mmercuri:
## Summary

`<CopilotChat>` from `@copilotkit/react-ui@1.56.3` cannot be reliably driven by Playwright in headless Chromium. Four standard input-driving patterns all fail to enable the Send button, blocking automated end-to-end tests of custom LangGraph / AG-UI integrations that use the default chat surface.

This is a testability request, not a runtime bug — the chat works fine for real users. But every CopilotKit customer who wants CI regression coverage for their custom agents hits this wall.

## What I was doing

End-to-end test for a LangGraph agent fix ([stratix-python#92](https://github.com/LayerLens/stratix-python/pull/92)): backend is FastAPI + `ag-ui-langgraph`, frontend is Next.js 16.2.4 + `@copilotkit/react-core@1.56.3` + `@copilotkit/react-ui@1.56.3` + `@copilotkit/runtime@1.56.3` wired via `LangGraphHttpAgent`. Playwright 1.59.1, Chromium 147 headless.

The page renders `<CopilotChat>` inline (not a popup). Test drives three user turns through an interrupt/resume cycle.

## Repro

```tsx
// page.tsx
"use client";
import { CopilotChat } from "@copilotkit/react-ui";

export default function Page() {
  return (
    <main style={{ height: "100vh", display: "flex", flexDirection: "column" }}>
      <div data-testid="harness-chat" style={{ flex: 1, minHeight: 0 }}>
        <CopilotChat
          labels={{ title: "Evaluator", initial: "Ask me to evaluate your traces." }}
        />
      </div>
    </main>
  );
}
```

```ts
// interrupt-resume.spec.ts
const input = page.locator('textarea[placeholder*="message" i]').first();

// ATTEMPT 1: standard Playwright fill
await input.fill("evaluate my traces");
// FAILS: "element is not visible" (36 retries over ~10s)

// ATTEMPT 2: scrollIntoView + focus + keyboard.type + Enter
await input.scrollIntoViewIfNeeded();
await input.focus();
await page.keyboard.type("evaluate my traces");
await page.keyboard.press("Enter");
// FAILS: no network request is made. Send button stays disabled.
// (Verified via `page.on("response", ...)` — zero /api/copilotkit calls.)

// ATTEMPT 3: pressSequentially (delay: 10ms per char)
[...]

PR #4519, opened 2026-04-30
fix(react-ui): add stable test ids to chat input

717986230:
## What does this PR do?

Adds stable test IDs to the CopilotChat input area so automated tests can reliably target the default chat UI.

Changes included:

- Adds `data-testid="copilot-chat-input"` to the chat textarea.
- Adds standard `data-testid` values to the send and stop button.
- Keeps the existing `data-test-id` values for backwards compatibility.

This is a small non-breaking testability improvement for the React UI package.

## Related PRs and Issues

Related to #4215

## Checklist


contextablemark (changes requested):
Hi @717986230, 

Thanks for this — the direction is right and keeping the legacy `data-test-id` for back-compat is exactly the conservative approach we wanted here. The send/stop button change lands correctly on the DOM (one naming note below).

One blocking issue on the input, though: **the `data-testid="copilot-chat-input"` never reaches the DOM.**

The textarea here isn't a plain element — it's the `AutoResizingTextarea` wrapper in `packages/react-ui/src/components/chat/Textarea.tsx`. That component declares a fixed props interface, destructures only those named props, and renders its inner `<textarea>` without a `{...rest}` spread. So the `data-testid` passed in `Input.tsx` is silently dropped and won't appear in the rendered markup. (It doesn't trip a type error because TypeScript skips checking hyphenated JSX attributes, so it compiles clean and looks correct.)

Net effect: the main goal of this PR (i.e., letting tests reliably locate the chat input by a stable test id) doesn't appear to be actually achieved yet.

Suggested fix: add an explicit `"data-testid"?: string` prop to `AutoResizingTextarea` and apply it to the inner `<textarea>`, so the value passed from `Input.tsx` flows through. (Prefer the explicit prop over a blanket `...rest` spread — it keeps the internal component's contract narrow and intentional rather than implicitly forwarding arbitrary attributes.)

Second thing : please align the selector **names** with the convention the V2 components already established (`packages/react-core/src/v2/components/chat/CopilotChatInput.tsx`):

- V2 uses `copilot-chat-textarea` for the textarea itself and reserves `copilot-chat-input` for the input container. This PR puts `copilot-chat-input` on the textarea, which collides with the container's name in V2. Suggest renaming the textarea selector to `copilot-chat-textarea` for cross-package consistency.
- V2 uses `copilot-send-button` for the send control, whereas the new `data-testid` here reuses the state-based legacy names (`copilot-chat-ready` / `copilot-chat-request-in-progress`). Keep the legacy `data-test-id` values as-is for back-compat, but consider giving the new `data-testid` the conventional `copilot-send-button` name.

That way a test recipe written against V2 carries over to the default `<CopilotChat>` surface unchanged.

Two smaller asks:

- Could you add a quick test asserting both the textarea and send-button test ids render? That would have caught the dropped prop and guards against regressions.
- Heads-up for expectation-setting: stable selectors let tests *locate* the controls, but they don't address the underlying report in #4215 — that programmatic input in headless Chromium doesn't update React state, so the Send button stays disabled. That's fine for this PR's scope (selectors first, no imperative API), but, as a result, #4215 should be kept open with a note that driving the input remains a separate follow-up.

Once the input id actually renders in the DOM and there's a test covering it, this is should be good to merge. Thanks again!

contextablemark:
Thanks again @717986230 — the direction here was right, so rather than leave it stalled I pushed the remaining review items directly onto this branch (`450d47f9`, since edits-by-maintainers is enabled). Credit for the original change stays yours.

What I added on top:

- **Fixed the dropped-prop bug.** `data-testid` now threads through `AutoResizingTextarea` (`Textarea.tsx`) and lands on the inner `<textarea>`. The component destructures a fixed prop set with no `{...rest}` spread, so the id passed from `Input.tsx` was previously silently dropped and never reached the DOM.
- **Aligned selector names with the V2 components** in `@copilotkit/react-core` so a test recipe carries over unchanged: `copilot-chat-textarea` on the textarea and `copilot-send-button` on the send control. The legacy `data-test-id` values (`copilot-chat-ready` / `copilot-chat-request-in-progress`) are preserved as-is for back-compat.
- **Added a test** (`input.testids.test.ts`) asserting the `Input.tsx` wiring *and* that `Textarea.tsx` forwards the prop — the dropped-prop guard. It mirrors the existing `testids.test.ts` source-match convention since the package vitest runs in a node env with no DOM harness. Verified it fails if the forwarding line is removed.

`nx run react-ui:test` is green (37 pass, 4 new) and a fresh build emits both new testids into `dist`.

One expectation-setting note for anyone following from #4215: stable selectors let tests *locate* the controls, but they do not address the underlying report — programmatic input in headless Chromium not updating React state, so Send stays disabled. That remains a separate follow-up, so #4215 should stay open with a note even after this merges.

Should be ready for another look — re-requesting review.

contextablemark (approved):
Approve after making changes.

tylerslaton (approved):

tylerslaton (changes requested):
For our project, it should be `data-test-id` as per our other components.

tylerslaton (approved):

PR #4526, opened 2026-04-30
feat(shell-docs): telemetry parity port for docs cutover

samjulien:
## Summary
Ports the client-side telemetry stack from `docs/` into `showcase/shell-docs/` and tracks every subsequent telemetry-related upstream merge so the cutover from `docs.copilotkit.ai` to shell-docs can ship without going dark on funnel attribution.

### Telemetry stack port
- **Client-side PostHog** — `posthog-js` provider mounted in `app/layout.tsx`, complements server-side middleware capture
- **GA4** — `react-ga4` pageview hook
- **HubSpot, Reo.dev** — Script tags in `<head>`
- **Scarf pixel** — page-load OSS attribution
- **RB2B** — canonical `<Script id="reb2b-script">` in `<head>` (matches upstream after `#4646` LCP-trim refactor; see env var note below)
- **Framework selector instrumentation** — fires `framework_selected` (net-new event, no upstream equivalent)
- **POSTHOG env rename** — `POSTHOG_PROJECT_KEY` → `POSTHOG_KEY` across both shell and shell-docs middlewares, env-drives `POSTHOG_HOST` with `eu.i.posthog.com` fallback

### Open Graph + routing
- **OG image route** — `app/og/[...slug]/route.tsx` for per-page social-share images
- **vercel.json** — `app/og/**` `maxDuration: 60`
- **Internal-link sweep** — 30 `https://docs.copilotkit.ai/...` self-references across 15 mdx files rewritten to relative paths so they don't redirect-loop post-cutover

### Upstream parity tracking (each commit mirrors a merged docs/ PR)
- **Reverse-proxy PostHog through `/ingest`** (mirrors `#4554`) — `next.config.ts` rewrites + provider hardcodes `/ingest` + middleware excludes `/ingest` from matcher
- **Talk to Our Engineers nav button** (mirrors `#4590`) — fires `talk_to_us_clicked` with `{ location: "docs_nav" }`
- **Cloud CTA repointed to dashboard.operations.copilotkit.ai** (mirrors `#4592`) — UTM-tagged + fires `try_for_free_clicked` with `{ location }`. Distinguishes `docs_navbar` (desktop) from `docs_navbar_mobile`. `LinkToCopilotCloud` is preserved for MDX content references that still point at `cloud.copilotkit.ai`
- **CLI command copy tracking** (mirrors `#4643` + follow-ups) — `track-command-copy.ts` + `CopyTracker` provider monkey-patches `navigator.clipboard.writeText`. Final shape: `cli_command_copied` with `{ install_type, location? }`. `install_type` covers `npx`/`pnpm`/`pip`/`uv`/`docker`/`curl`/`brew`/`helm`/`kubectl`/`make`/`bash`/`sh` and friends
- **`capture_dead_clicks: false`** (mirrors `#4646` LCP trim) — drops dead-clicks-autocapture.js
- **Drop `useRB2B` hook + use canonical `<Script>`** (mirrors `#4646`) — single source of truth for RB2B injection

## Env vars required on the deployment target

Same project IDs as the current docs deployment for data continuity:

- `NEXT_PUBLIC_POSTHOG_KEY`, `POSTHOG_KEY`, `NEXT_PUBLIC_POSTHOG_HOST`
- **`NEXT_PUBLIC_REB2B_KEY`** — renamed from `NEXT_PUBLIC_RB2B_ID` to match upstream's canonical Script. Same key value, different env var name. **Deployment ops needs to set this.**
- `NEXT_PUBLIC_SCARF_PIXEL_ID`
- `NEXT_PUBLIC_REO_KEY`
- `NEXT_PUBLIC_GOOGLE_ANALYTICS_TRACKING_ID`

## Architecture notes

- Hook-based telemetry (GA4) and `CopyTracker` are consolidated into `<AnalyticsClient />`, a single client boundary mounted from the server-component layout. Mirrors upstream's `ProvidersWrapper` pattern.
- `<head>` carries the Reo.dev, HubSpot, and (gated on `REB2B_KEY`) RB2B Script tags. PostHog provider wraps the body inside Suspense; Scarf pixel sits at end of body.

## Test plan
- [ ] `nx run shell-docs:typecheck`, `lint`, `build` clean
- [ ] Cloud CTA click → confirm dashboard URL with UTMs lands and `try_for_free_clicked` fires with the right `location`
- [ ] Talk to Our Engineers click → confirm `talk_to_us_clicked` fires with `location: "docs_nav"` and routes to copilotkit.ai/contact-us
- [ ] PostHog dashboard: docs session continues to attribute via `/ingest` reverse proxy
- [ ] Copy any code block → confirm `cli_command_copied` fires once with the correct `install_type`
- [ ] RB2B, Reo.dev, HubSpot, GA4 scripts load (when their env vars are set) and report events
- [ ] OG image route renders for 5 random pages
- [ ] Framework selector emits `framework_selected`


showcase/shell-docs/src/lib/providers/posthog-provider.tsx
```diff
@@ -0,0 +1,156 @@
...
+          : null;
+      const initSessionId = params?.get("session_id") ?? sessionId;
+
+      // Suppress all PostHog-related console errors and warnings
+      const originalError = console.error;
+      const originalWarn = console.warn;
+      const originalLog = console.log;
+
+      const suppressPostHogErrors = () => {
+        console.error = (...args: unknown[]) => {
```
BenTaylorDev: Claude says (and this is true of other repos from which this came):

Global console patching in PostHog provider — src/lib/providers/posthog-provider.tsx:48-100

  suppressPostHogErrors() permanently monkey-patches console.error/console.warn with no teardown. Concrete
  failure modes:
  - Substring match on "posthog" swallows any unrelated message that happens to contain it.
  - On HMR or StrictMode double-mount, isInitializedRef.current only guards the same component instance — a
  fresh mount calls suppressPostHogErrors() again, treating the previously-patched console.error as
  originalError. Each remount layers another wrapper.
  - originalLog is captured but console.log is never patched — dead capture.

  If this is verbatim from upstream docs/, fine to track separately. If it's new in the port, worth
  simplifying — most ad-blocker noise can be silenced with posthog.debug(false) and loaded: () => {} callbacks
   instead of patching globals.

showcase/shell-docs/src/components/framework-selector.tsx
```diff
@@ -9,6 +9,7 @@
 
 import React, { useEffect, useRef, useState } from "react";
 import { usePathname, useRouter } from "next/navigation";
+import posthog from "posthog-js";
```
BenTaylorDev: > uses the posthog-js singleton import while the rest of the codebase
  uses usePostHog() from posthog-js/react. Both work after posthog.init() runs, but the inconsistency is a
  future-you trap. Switch to the hook.

showcase/shell-docs/src/components/brand-nav.tsx
```diff
@@ -347,17 +366,31 @@ export function BrandNav(_props: BrandNavProps = {}) {
                   {label}
                 </Link>
               ))}
-              <Link
-                href={CLOUD_CTA.href}
+              <a
+                href="https://dashboard.operations.copilotkit.ai/?utm_source=docs&utm_medium=cta&utm_campaign=intelligence&utm_content=navbar"
```
BenTaylorDev: > CLOUD_CTA keeps only label after href was removed; the dashboard URL
  with full UTM string is duplicated inline for desktop and mobile. Pull the URL back into the constant so UTM
   tweaks are single-touch.

showcase/shell-docs/src/app/layout.tsx
```diff
@@ -77,16 +82,62 @@ export default function RootLayout({
...
   return (
     <html
       lang="en"
       className={`${plusJakartaSans.variable} ${splineSansMono.variable}`}
     >
+      <head>
+        <Script
+          id="reo-init-script"
+          strategy="afterInteractive"
+          dangerouslySetInnerHTML={{
```
BenTaylorDev: dangerouslySetInnerHTML with unescaped env interpolation + missing gate — src/app/layout.tsx:97-115

  The Reo init script interpolates REO_KEY directly into a <script> body. Two issues:
  - If REO_KEY is undefined, the script still runs with e = "undefined" and fetches
  https://static.reo.dev/undefined/reo.js. The sibling REB2B_KEY is correctly gated ({REB2B_KEY ? <Script ...
  /> : null}); Reo should match.
  - Bare interpolation into dangerouslySetInnerHTML — any future env value containing " or </script> breaks
  out. Use JSON.stringify(REO_KEY) for the inline literal.

showcase/shell-docs/src/components/framework-selector.tsx
```diff
@@ -150,6 +151,21 @@ export function FrameworkSelector({
 
   function selectFramework(slug: string) {
     setStoredFramework(slug);
+    // Fire a PostHog event so analytics dashboards can see which
+    // backend readers pick. Wrapped in try/catch — PostHog can be
+    // blocked by ad blockers or fail to initialize, and a broken
+    // analytics call must never break navigation.
+    try {
+      const opt = options.find((o) => o.slug === slug);
+      posthog.capture("framework_selected", {
```
BenTaylorDev: >  Recommendation on framework_selected

  By the website's convention, surface-/feature-specific interactions get a <surface>.<feature>.<action> prefix —
  pricing.cta_clicked is the closest precedent (a top-level page-specific action that still gets prefixed). The
  framework selector is a docs-only widget that doesn't exist anywhere else, so a flat name will sit oddly next to
  the existing homepage.*/ag-ui.*/pricing.* clusters in PostHog Insights.

  Suggested rename: docs.framework_selector.selected (or docs.framework_selected if you don't want the widget-name
  level). Cheap to do now since it's a brand-new event with zero historical data — much harder to migrate after
  it's been firing for a quarter.

showcase/shell-docs/src/lib/providers/posthog-provider.tsx
samjulien: Fixed in 601b72663. Removed the suppressor block entirely — the /ingest reverse proxy in this PR eliminates the ERR_BLOCKED_BY_CONTENT_BLOCKER noise it was masking, so it's dead code now. All three concrete bugs (substring overreach, HMR wrapper layering, dead originalLog capture) go with it. PDX-112 closed as resolved here.

showcase/shell-docs/src/components/framework-selector.tsx
samjulien: Fixed in 945a69799. Switched framework-selector to usePostHog() to match the rest of the codebase.

showcase/shell-docs/src/components/brand-nav.tsx
samjulien: Fixed in f64b957ee. Pulled the dashboard URL + UTM string back into CLOUD_CTA.href so desktop and mobile share one source of truth.

showcase/shell-docs/src/app/layout.tsx
samjulien: Fixed in 72794959f. Gated the Reo Script on REO_KEY and switched to JSON.stringify for the inline interpolation. The escaped value is bound to a local var inside the IIFE so the key only appears once.

showcase/shell-docs/src/components/framework-selector.tsx
samjulien: Fixed in 945a69799. Renamed framework_selected → docs.framework_selected to match the dotted-prefix convention in PostHog Insights. Brand-new event with no historical data, so this is a clean rename.

BenTaylorDev (approved):

PR #4592, opened 2026-05-01
docs: add sign-up CTAs and refresh framework CLI walkthrough

samjulien:
## Summary

- Adds a reusable `OpsPlatformCTA` component with four visual variants (card, info, inline, tile), PostHog `cta_clicked` instrumentation, and UTM-tagged links to the dashboard
- Places sign-up CTAs across topical docs pages, integration overview pages, manual-setup quickstarts, premium pages, and the shared prebuilt-components snippet (covers all per-integration prebuilt pages)
- Repoints the navbar "Free Developer Access" link to the dashboard with matching click tracking
- Adds an `afterFeatures` slot to `FrameworkOverview` so MDX can inject content below the supported-features section
- Rewrites the CLI step on the eight framework quickstarts that have a scaffold (ADK, Agno, AWS Strands, LangGraph, LlamaIndex, Mastra, Microsoft Agent Framework, Pydantic AI) to use the interactive flow and describe the prompts (project name, platform Yes/No with sign-up link, framework picker)

Coverage: ~13% of docs pages, focused on the highest-traffic surfaces and topical landing pages.

## Test plan

- [ ] `/`, `/quickstart`, `/prebuilt-components`, `/shared-state`, `/inspector` show their CTAs in the expected position
- [ ] Integration overview pages (`/built-in-agent`, `/langgraph`, `/adk`, `/aws-strands`, `/microsoft-agent-framework`) render the CTA above the fold
- [ ] Each of the 13 per-integration `/{framework}/prebuilt-components` pages renders the inline CTA via the shared snippet
- [ ] Each of the 8 framework quickstarts shows the new CLI walkthrough (Yes-first prompt, dashboard sign-up link, correct framework label)
- [ ] Navbar still shows "Talk to Our Engineers" alongside the repointed "Free Developer Access" link
- [ ] DevTools Network: clicking any CTA sends one `cta_clicked` PostHog event with `{surface, variant, target, page_path}`
- [ ] Destination URL on every CTA ends with `?utm_source=docs&utm_medium=cta&utm_campaign=intelligence&utm_content=<surface>`

### DIFF
--- a/docs/app/(home)/[[...slug]]/page.tsx
+++ b/docs/app/(home)/[[...slug]]/page.tsx
@@ -22,6 +22,7 @@ import { Cards, Card } from "fumadocs-ui/components/card";
 import { PropertyReference } from "@/components/react/property-reference";
 import { InsecurePasswordProtected } from "@/components/react/insecure-password-protected";
 import { LinkToCopilotCloud } from "@/components/react/link-to-copilot-cloud";
+import { OpsPlatformCTA } from "@/components/react/ops-platform-cta";
 import { Accordions, Accordion } from "fumadocs-ui/components/accordion";
 import { NavigationLink } from "@/components/react/subdocs-menu";
 import { getSnippetTOCForPage } from "@/lib/snippet-toc";
@@ -40,6 +41,7 @@ const mdxComponents = {
   ...defaultMdxComponents,
   InsecurePasswordProtected: InsecurePasswordProtected,
   LinkToCopilotCloud: LinkToCopilotCloud,
+  OpsPlatformCTA: OpsPlatformCTA,
   Accordions: Accordions,
   Accordion: Accordion,
   Tabs: Tabs,
--- a/docs/app/integrations/[[...slug]]/page.tsx
+++ b/docs/app/integrations/[[...slug]]/page.tsx
@@ -21,6 +21,7 @@ import { Cards, Card } from "fumadocs-ui/components/card";
 import { PropertyReference } from "@/components/react/property-reference";
 import { InsecurePasswordProtected } from "@/components/react/insecure-password-protected";
 import { LinkToCopilotCloud } from "@/components/react/link-to-copilot-cloud";
+import { OpsPlatformCTA } from "@/components/react/ops-platform-cta";
 import { Accordions, Accordion } from "fumadocs-ui/components/accordion";
 import { NavigationLink } from "@/components/react/subdocs-menu";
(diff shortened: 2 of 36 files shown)


BenTaylorDev (commented):
## Inconsistencies worth resolving before merging

### Imports — two PostHog access patterns in one file

`docs/components/layout/navbar.tsx` now mixes both:

- `const posthog = usePostHog()` (the React hook — was already there for `handleTalkToEngineersClick`)
- `import posthog from "posthog-js"` (newly added module-level import)

The hook-bound `const` shadows the module import inside the component body, so the new top-level `import posthog` is effectively unused inside `Navbar` — `handleIntelligenceCtaClick` ends up calling through the same hook value as the existing handler. Either drop the new import and reuse the hook, or commit to the module pattern and drop `usePostHog()`. Whichever you pick, `ops-platform-cta.tsx` should follow suit (it currently uses the module import only).

Related nit in the same file: `handleTalkToEngineersClick` calls `posthog?.capture(...)` with the optional chain (the hook can return `undefined`), but `handleIntelligenceCtaClick` calls `posthog.capture(...)` without it. Both call sites should agree.

### Event names — two click taxonomies in the same file

Within `navbar.tsx`, navbar clicks now fire under two different conventions:

- `talk_to_us_clicked` — specific event name, flat properties
- `cta_clicked` — generic event with `{ surface, variant, target }` taxonomy

Same surface (the docs navbar), two different event shapes. Pick one — either migrate `talk_to_us_clicked` into `cta_clicked` (e.g. `surface: \"docs:navbar:talk-to-us\"`), or give the new sign-up CTAs their own specific name. As-is, anyone building a \"navbar CTA\" funnel has to union two events.

### `surface` strings — implicit grammar, mixed segment types

The `surface` taxonomy is a good idea, but the values across the MDX files would benefit from a written convention:

- Segment separator is a colon, route paths inside a segment use slashes (e.g. `docs:reference/v2/components/CopilotChat:inline`). Sensible, but undocumented — a one-line comment in `OpsPlatformCTA` describing `docs:<route-without-leading-slash>:<placement>` would prevent drift.
- The trailing segment mixes positional and semantic names:
  - positional: `top`, `inline`, `left`, `right`, `hero-tile`
  - semantic slot: `after-cli`, `signup-step`, `whats-next`

  Both styles are reasonable individually, but encoding both in the same field makes \"all top-of-page CTAs\" or \"all post-quickstart CTAs\" queries fragile. Consider splitting into two props (e.g. `placement` + `slot`) or picking one style.

BenTaylorDev (approved):

PR #4620, opened 2026-05-03
fix: trigger re-render when agent node changes

aartisonigra:
I fixed an issue in useAgentNodeName where the hook was only updating a useRef, which did not trigger React re-renders when the agent node changed. This could result in stale or undefined values being accessed during runtime.

To resolve this, I introduced useState alongside useRef so that updates to the node name properly trigger re-renders while still maintaining a reference for internal consistency.

The nodeName is now updated through a shared updateNodeName function used across all event handlers (onStepStartedEvent, onRunStartedEvent, onRunFinishedEvent, and onRunErrorEvent), ensuring consistent state updates across the agent lifecycle.

This change improves reliability and ensures UI always reflects the latest agent node state.


aartisonigra:
Hi maintainers, I’ve synced the branch with the latest upstream changes.

The fix ensures proper re-rendering when the agent node changes by using useState along with useRef.

Could you please review this PR? Also, some Vercel checks are awaiting authorization.

Thank you!

BenTaylorDev:
Thanks for this, and sorry it sat. I've verified the fix and it's correct: `useAgentNodeName` was returning `nodeNameRef.current`, and a ref write triggers no re-render, so both consumers — `useCoAgent().nodeName` (`use-coagent.ts:215`) and `use-langgraph-interrupt.ts:43` — kept showing a stale node. Adding state is the right fix. I also checked the risk I was most worried about, that node-name churn would re-register interrupt handlers, and it doesn't: both consumers assign it into a ref during render with `useCallback([], …)`, so the extra renders make `agentMetadata.nodeName` more accurate rather than looping.

Two small things and I think this is ready:

**1. Missing trailing newline** — this fails the `format` CI job. Confirmed locally with the repo's formatter:

```
$ oxfmt --check packages/react-core/src/hooks/use-agent-nodename.ts
Format issues found in above 1 files.
```

The whole fix:

```diff
   return nodeName;
-}
\ No newline at end of file
+}
```

**2. `nodeNameRef` is now unused** — it's declared and written in `updateNodeName`, but nothing reads it any more since the hook returns the state value. Worth deleting the ref and having `updateNodeName` just call `setNodeName`, so there's no second source of truth to drift.

A small test would be welcome if you're up for it — there's no test file for this hook yet, and the behavior is observable through `useCoAgent().nodeName`. Not a blocker though; if you'd rather just do the two items above, say so and we'll add coverage separately.

One thing on our side, not yours: this means every `useCoAgent` consumer now re-renders on each `STEP_STARTED`. That's per graph node rather than per token, and it's the behavior the hook always advertised, so we're happy to take it — just noting it's a deliberate call, not an oversight.

BenTaylorDev:
Closing this as superseded — but the diagnosis here was right, it was first, and I handled the follow-through badly.

You found it correctly: `useAgentNodeName` returned `nodeNameRef.current`, ref writes schedule no render, so `useCoAgent().nodeName` (and `use-langgraph-interrupt`) kept reporting whichever node was current at the last render. That was a real bug with two live consumers, sitting in the codebase since the hook was written, and you spotted it from a five-line read. Backing it with state was exactly the right fix.

What happened next is on me. I confirmed your fix in August, asked for two cosmetic changes, and then six days later landed the same change myself in #6520 (`ffd15801d6`) without referencing this PR or adding you as a co-author. I've now corrected #6520's description to record the credit, since the merged commit can't be rewritten to carry a `Co-Authored-By` trailer. What shipped is your change plus the two nits I'd asked for and the tests the hook never had — no third difference.

For completeness on why it can't be merged as-is now, independent of the above: `e499c65dce` moved `packages/react-core/src/hooks/` under `v1-deprecated/hooks/`, so this branch is 3,887 commits behind and `git merge-tree` reports a modify/delete conflict — merging would resurrect the file at a path that no longer exists.

Sorry for the eight-month gap and for not crediting you when it landed. The instinct you showed here — reading a hook and noticing the ref/render mismatch rather than working around the symptom — is the useful kind, and I'd be glad to see more from you. If you want something current with a clearer runway, say so and I'll point you at one.

PR #4622, opened 2026-05-03
fix(react-ui): allow sidebar children to use height 100%%

### LINKED ISSUE #261: When a sidebar is the parent, the child can not be made full screen with height 100%
**Problem**:
When putting child nodes under `CopilotSidebar`, these nodes can't be made `height: 100%` easily because of the way the DOM node are generated.

**Proposed Solution**:
In `CopilotSidebar`, change the parent DOM nodes so that the child elements can be set to `height: 100%`.

ashish4143:
## Summary
Fixes #261

When child nodes are placed under <CopilotSidebar>, they can't use height: 100%% because the wrapper DOM elements don't propagate height properly. This doesn't happen on Safari but occurs on Chrome.

## Problem
The .copilotKitSidebarContentWrapper and .copilotKitModalChildrenWrapper divs don't have height/flex properties, so children can't inherit height: 100%%.

## Changes
- Added height: 100%%, display: flex, and lex-direction: column to .copilotKitSidebarContentWrapper
- Added lex: 1 and height: 100%% to .copilotKitModalChildrenWrapper (child selector scoped)

## How to Test
1. Wrap content inside <CopilotSidebar>
2. Set a child element to height: 100%%
3. Verify the child fills full height in Chrome (previously failed)

### DIFF
--- a/packages/react-ui/src/css/sidebar.css
+++ b/packages/react-ui/src/css/sidebar.css
@@ -37,6 +37,14 @@
   overflow: visible;
   margin-right: 0px;
   transition: margin-right 0.3s ease;
+  height: 100%;
+  display: flex;
+  flex-direction: column;
+}
+
+.copilotKitSidebarContentWrapper > .copilotKitModalChildrenWrapper {
+  flex: 1;
+  height: 100%;
 }
 
 @media (min-width: 640px) {


BenTaylorDev (changes requested):
Thanks, and apologies this old one sat. The bug is real and still present, and you've correctly targeted the v1 `@copilotkit/react-ui` sidebar CSS (the file is unchanged on `main` and this merges cleanly). My concern is that the fix is applied too broadly and its effectiveness isn't verified:

- **It changes layout for every react-ui sidebar consumer, opt-in or not.** `.copilotKitSidebarContentWrapper` wraps the *entire* consumer app when `CopilotSidebar` is used. Flipping it to `display:flex; flex-direction:column; height:100%` and forcing the app-content child to `flex:1; height:100%` imposes a flex-column layout on everyone — apps whose content previously flowed naturally can get compressed/reflowed, and there's no way to opt out.
- **`height:100%` may not even resolve.** It only takes effect if the whole ancestor chain (`html`/`body`/`#root`) also has an explicit height, which react-ui doesn't set and can't guarantee — so in a stock app (e.g. Next.js with `body` height auto) the child still collapses to content height, and the reporter's full-height goal isn't met while everyone still inherits the flex change. (The issue's "works in Safari, not Chrome" symptom points at a flexbox `min-height:auto` difference this doesn't clearly address.)

Could you re-scope this as an **opt-in** (e.g. a modifier class a consumer adds to enable full-height children) rather than a default on all consumers, and verify it in a real Chrome app where `html/body/#root` don't declare `height:100%`? Happy to re-review once it's scoped and confirmed live.

BenTaylorDev:
Closing in favor of the superseding: https://github.com/CopilotKit/CopilotKit/pull/6410

lukasmoschitz:
Thanks for this, @ashish4143 — and sorry it stayed open so long after it was already resolved.

Your diagnosis was correct and it's what unblocked #261 after it had been open since March 2024: the two auto-height wrappers (`.copilotKitSidebarContentWrapper` and `.copilotKitModalChildrenWrapper`) leave a child's percentage height with no definite containing block to resolve against.

That landed in #6410 (merged 2026-08-07), which credits you as co-author on the commit. It took the same diagnosis in a slightly different direction:

- **Opt-in via a new `fullHeightChildren` prop** rather than a default, since the content wrapper wraps the entire consumer app and the change would otherwise affect every `CopilotSidebar` user.
- **`100dvh`/`100vh` instead of `height: 100%`** — a percentage here only resolves if `html`/`body`/`#root` also declare a height, which `react-ui` can't guarantee.
- **`min-height: 0`** on the children wrapper so tall content scrolls inside the child instead of stretching past the viewport.

Usage:

```tsx
<CopilotSidebar fullHeightChildren>
  <div style={{ height: "100%" }}>...</div>
</CopilotSidebar>
```

Closing as superseded. Thanks again for the report and the fix.

PR #4629, opened 2026-05-04
feat(core): registerProxiedAgent — explicit multi-agent mounting against one runtime agent

mme:
## Summary

Replaces the implicit per-thread agent cloning from #3525 / #3630 with an explicit registration API. Three commits:

1. **revert** — strips the cloning machinery (`useAgent({ threadId })` per-thread clones, `getThreadClone`, `globalThreadCloneMap`, `cloneForThread`), the inspector hooks added only to handle clones (`onAgentRunStarted` from #3869, the connect-time emission from #3872, the `agentRunThreadId` map), the state-manager `isClone` composite-key path, and the `consumerAgent` parameter on `SuggestionEngine`. Restores `agent.threadId = resolvedThreadId` in `CopilotChat` (pre-#3525 behavior). Re-opens issue #2957 (CPK-7155): two `<CopilotChat>` instances sharing an `agentId` will share state again.

2. **feat** — adds `CopilotKitCore.registerProxiedAgent({ agentId, remoteAgentId })` which mints a `ProxiedCopilotRuntimeAgent` under a local registry id and routes its outbound HTTP requests to the named runtime agent. Returns `{ agent, unregister }` for React `useEffect` cleanup. Throws on duplicate `agentId` (collisions with `agents__unsafe_dev_only` or another `registerProxiedAgent` are loud, not silent).

   `ProxiedCopilotRuntimeAgent` gains a `remoteAgentId` field used only for outbound routing — URL paths (`/agent/<id>/run`, `/connect`, `/stop`), single-route envelopes, and the `IntelligenceAgent` delegate's `agentId`. The local `agentId` remains the registry key and source of truth for state-manager subscriptions, `useAgent` lookups, and `onAgentsChanged`. So multiple proxies (`chat-1`, `chat-2`) targeting the same runtime agent (`default`) don't cross-talk in any subscriber bookkeeping.

3. **test** — re-adds the isolation coverage that the revert deleted, rewritten against the explicit-registration model (10 tests total).

## Usage

```tsx
const { copilotkit } = useCopilotKit();
useEffect(() => {
  const { agent, unregister } = copilotkit.registerProxiedAgent({
    agentId: "chat-1",          // local registry id (subscriber bookkeeping)
    remoteAgentId: "default",   // runtime id (URL routing only)
  });
  return unregister;
}, [copilotkit]);
// then <CopilotChat agentId="chat-1" />
```

## Test plan

- [x] `nx run @copilotkit/core:test` — **424 passed** (was 409 + 15 new across feat & test commits)
- [x] `nx run @copilotkit/react-core:test` — **1151 passed** (was 1150 + 1 new)
- [x] `nx run @copilotkit/web-inspector:test` — **7 passed**
- [x] Build: `core`, `react-core`, `web-inspector` all clean

### New tests cover

- routing — proxy at `agentId` routes outbound to `remoteAgentId`; URL path encodes the remote id, body envelopes use the remote id
- duplicate-throw — register twice with same `agentId` throws (whether against another registered proxy or an `agents__unsafe_dev_only` entry)
- idempotent `unregister` — calling twice doesn't throw or re-emit; stale handles don't strip replacements
- `onAgentsChanged` notification on register and unregister
- header inheritance from core
- two proxies → same `remoteAgentId` are distinct instances with isolated messages, isolated state, independent threadIds, but shared outbound URL
- `getAgent` returns the same proxy instance across calls (no per-call clone)
- registering before runtime connects yields a proxy that's still usable for in-memory ops
- registering with a `remoteAgentId` the runtime doesn't yet expose still works for local bookkeeping
- re-register after unregister yields a fresh proxy (no carry-over)
- regression: activity renderers receive the agent under the local `agentId`, never another agent in the registry (replaces the deleted clone-vs-registry trap test)

## Caveats

- Issue #2957 (CPK-7155) is re-opened by the revert. The new API is the supported path forward — callers that previously relied on `useAgent({ threadId })` for isolation should move to `registerProxiedAgent`.
- Runtime `/info` sync still uses last-write-wins, not throw, on collisions with manually-registered ids. Easy follow-up if we want strict throw there too.


mxmzb:
# PR #4629 Review Summary

## Overall Assessment

The PR is **well-designed and well-executed**. The shift from implicit WeakMap-based per-thread cloning to explicit `registerProxiedAgent` is a sound architectural improvement. However, five specialized review agents found several issues worth addressing before merge.

---

## Critical Issues (2 found)

### C1. Cleanup `try/catch` blocks swallow errors with no subscriber visibility
`packages/core/src/core/core.ts:415-446`

The new try-catch blocks around `threadStoreRegistry.unregister()` and `stateManager.unsubscribeFromAgent()` log to `console.error` but skip the subscriber error bus (`logAndEmitError` / `onError`). Production users with structured logging will never see these. If cleanup throws once, the subscription leak becomes **permanent and silent** — exactly the bug the new cleanup is supposed to prevent.

**Fix:** Route both catches through `logAndEmitError` so subscribers (including the web inspector) are notified.

### C2. `void this.notifyAgentsChanged()` — unhandled rejection risk
`packages/core/src/core/agent-registry.ts:248, 259`

Both the register and unregister paths fire-and-forget the async `notifyAgentsChanged()`. If it rejects (e.g. corrupted subscriber set, partially-initialized core via friends cast), the rejection is silently dropped. In Node 15+ this terminates the process.

**Fix:** Add `.catch()` handlers to both new `void this.notifyAgentsChanged()` calls.

---

## Important Issues (5 found)

### I1. `routedAgentId()` is never asserted on the wire for connect/stop paths
Rating 9/10 — `packages/core/src/agent.ts:157-165, 201, 233, 355, 373`

The new test only asserts the constructor-baked `url` field (which covers REST `run`). It does NOT exercise that `#connectViaHttp`, `abortRun` (stop), or single-route envelopes use `remoteAgentId` instead of `agentId`. The existing `proxied-runtime-transport.test.ts` matrix has zero occurrences of `remoteAgentId`. A regression where any of these paths fall back to `agentId` would silently route traffic to the wrong runtime agent and pass every test.

**Fix:** Add `remoteAgentId` rows to `proxied-runtime-transport.test.ts` matrix across run/connect/stop × REST + single.

### I2. Inspector reads live `agent.threadId` — race condition
`packages/web-inspector/src/index.ts:3004`

The old code captured `threadId` from the `onAgentRunStarted` event payload (frozen at run-start). The new code reads it *live* off the agent instance when a message event fires. If an agent switches threads between a RUN_STARTED and a late MESSAGES_SNAPSHOT event, the inspector ticks `liveMessageVersion` for the **wrong thread**.

**Fix:** Capture `threadId` from the RUN_STARTED event in the per-agent subscription, not from the mutable agent instance.

### I3. Optional `remoteAgentId` allows `(undefined, undefined)` state
`packages/core/src/agent.ts:75-93, 95-165`

`ProxiedCopilotRuntimeAgentConfig` inherits optional `agentId` from `HttpAgentConfig` and adds optional `remoteAgentId`. Constructing with neither set produces `runUrl = ".../agent//run"` — malformed but doesn't throw until first runtime call.

**Fix:** Make `agentId` required on `ProxiedCopilotRuntimeAgentConfig` (override the parent's optional). Move `routedAgentId()` throw into the constructor for fail-fast.

### I4. `stateManager.unsubscribeFromAgent` cleanup is untested
Rating 8/10 — `packages/core/src/core/core.ts:430-447`

This is brand-new logic. No test registers a proxy, subscribes it, unregisters, then asserts the StateManager subscription was torn down. A future refactor could silently reintroduce the subscription leak.

**Fix:** Add a test that spies on `unsubscribeFromAgent` after `unregister()`.

### I5. No empty-string validation on `agentId` / `remoteAgentId`
`packages/core/src/core/agent-registry.ts:197-262`

`registerProxiedAgent({ agentId: "", remoteAgentId: "default" })` happily installs an agent under key `""`. `remoteAgentId: ""` is worse — produces `/agent//run`.

**Fix:** Throw on empty/whitespace ids in `registerProxiedAgent`.

---

## Suggestions (8 found)

| # | Location | Suggestion |
|---|----------|------------|
| S1 | `core-register-proxied-agent.test.ts` | Test assertions peek at private fields via `unknown` cast (`url`, `runtimeMode`). Mock `fetch` and assert outbound URLs instead — also closes I1. |
| S2 | `agent.ts:157` | Enrich `routedAgentId()` error message with `agentId`, `remoteAgentId`, `threadId` values for debuggability. |
| S3 | `agent-registry.ts` vs `core.ts` | Naming inconsistency: `CopilotKitCoreAddAgentParams` uses `id`, while new types use `agentId`. Align to one convention. |
| S4 | `agent-registry.ts:226-228` | `runtimeMode` resolution comment describes 2 of 3 ternary branches. The "no runtimeUrl → SSE" fallback is the most surprising and undocumented. |
| S5 | docs + `core.ts` JSDoc | Note that the `useEffect` + `unregister` pattern is React StrictMode-safe (register/unregister/register is supported). |
| S6 | `agent-registry.ts` | `unregister: () => void` discards whether removal actually happened. Consider `() => boolean` for test/debug visibility. |
| S7 | `use-agent.tsx` | JS callers passing removed `threadId` prop get no runtime feedback. Add a dev-mode warning for unknown props. |
| S8 | Breaking changes | `onAgentRunStarted` removal from `CopilotKitCoreSubscriber` and `threadId` removal from `UseAgentProps` are breaking. Deserves CHANGELOG entry. |

---

## Strengths

- **Explicit > implicit** — the `registerProxiedAgent` API eliminates an entire class of clone-vs-registry identity bugs
- **Prototype-safe key check** — `hasOwnProperty.call` instead of `in`, with inline comment explaining why
- **Idempotent unregister with replacement guard** — identity check prevents stale closures from killing fresh registrations, tested thoroughly
- **Clean revert** — zero stale references to `getThreadClone`, `globalThreadCloneMap`, `cloneForThread`, `onAgentRunStarted`, or `isClone` remain in source
- **Comprehensive isolation tests** — 10+ tests covering routing, duplicates, idempotency, isolation, pending mode, offline core, and re-registration

---

## Recommended Action

```
1. Fix C1 + C2 (error handling visibility)     — before merge
2. Fix I2 (inspector threadId race)             — before merge
3. Add I1 tests (wire-level routing assertions) — before merge
4. Fix I3 + I5 (type validation)                — before merge or fast-follow
5. Add I4 test (cleanup assertion)              — fast-follow OK
6. Address S1-S8                                — at your discretion
```

---
🤖 Generated with [Claude Code](https://claude.com/claude-code) — 5-agent parallel review (code quality, test coverage, error handling, type design, comment accuracy)

AlemTuzlak (dismissed):

PR #4635, opened 2026-05-04
feat(vscode-extension): chat playground for the CopilotKit dev experience

AlemTuzlak:
## Summary

- Migrates the chat tab from the OpenAI/Anthropic + API-key + Node-subprocess flow to VS Code's built-in Language Model API (`vscode.lm`), with optional `vscode.lm.tools` forwarding so the model can use Copilot's web-fetch tool out of the box.
- Replaces the stock `CopilotChat` with a custom `PlaygroundChat` that drives the runtime directly over SSE, executes frontend tool handlers locally, supports `followUp: false`, renders markdown, and animates fixture replay messages.
- Adds a Tailwind v4 compile pass to the playground bundler so user renders pick up their own utility CSS, plus a workspace file watcher for hot-reload on `.ts/.tsx/.css` edits.
- Fixes a `toLmMessages` bug that was dropping assistant tool calls and tool results, causing the model to loop on the same tool with slightly tweaked args.
- Restructures the right "Components" panel into expandable component → hook rows with clickable line-jumps, and rebuilds the left sidebar around three clearly-labeled sections (Model / Recording / Saved replays). Both panels are now resizable + collapsible (state persisted to `localStorage`).
- Rebuilds the test-workspace as a v2-only weather demo: every hook becomes `useFrontendTool` with a deterministic mock-data handler so the chat actually renders populated cards. Adds `displayCurrentWeather` and the verb-prefix naming convention so the LLM treats them as UI tools rather than data fetchers.

## Test plan

- [ ] `nx test copilotkit-vscode-extension` passes locally (300 tests, 1 pre-existing flaky `runtime-host` test).
- [ ] `nx build copilotkit-vscode-extension` produces a clean dist with no rolldown warnings beyond the existing `define` deprecation notice.
- [ ] Open the dev host on `packages/vscode-extension/test-workspace`, open the Chat tab, ask "what's the weather in Sarajevo?" — the `displayCurrentWeather` card renders with populated values, conversation ends with a brief text reply (no 5-step loop).
- [ ] Save the chat as a fixture, click ▶ on the saved row — recorded conversation animates back into the chat surface.
- [ ] Edit a hook file (e.g. add a className to `WeatherTool.tsx`) — chat tab rebundles within ~300ms without a manual refresh.
- [ ] Resize the left/right side panels via the boundary handles; collapse each via the floating chevrons; verify state survives a reload.
- [ ] Click a tool-name pill in a chat message → opens the registering hook at its exact line in the editor.
- [ ] Diagnostics panel shows `Tailwind: styles/globals.css` and `VS Code tools: N available` for the test-workspace.


AlemTuzlak:
already added and mvoed out of this repo into copilotkit/vscode-extension

PR #4648, opened 2026-05-05
feat(angular): port CopilotChatReasoningMessage

marthakelly:
## Summary

Ports `CopilotChatReasoningMessage` from `@copilotkitnext/react-core` to `@copilotkitnext/angular`. Net new component (plus three slot-overridable sub-components), behavioral spec, and public API surface — no other packages touched. The Angular implementation expresses the same behavioral contract as the React source while idiomatic-converting hooks → signals/effects.

## Behavioral contract (preserved from React)

- `isStreaming` is true iff `isRunning && message.id === messages[messages.length - 1].id`.
- While streaming: header label is `"Thinking…"`, with a small pulsing dot when there is no content yet.
- After streaming ends: header label is `"Thought for X seconds"` using a ported `formatDuration` helper (`< 1s` → `"a few seconds"`, `< 60s` → `"N seconds"`, whole minutes → `"N minute(s)"`, otherwise `"Nm Ms"`). Elapsed time is captured at the moment streaming flips off.
- Auto-collapse: opens automatically when streaming starts, auto-collapses when streaming ends — UNLESS the user has manually clicked the header during this streaming session. Manual flag resets each time streaming restarts.
- Header click toggles open/closed only when `hasContent`. When no content, click is a no-op (and the rendered `<button>` is `disabled`).
- When `hasContent` is false AND not streaming, the Content area renders nothing.
- Header has `aria-expanded` only when expandable (i.e. content exists).
- Three slot-overridable sub-components: Header, Content, Toggle.
- Content rendering uses the existing Angular markdown renderer (`CopilotChatAssistantMessageRenderer`), mirroring how React uses Streamdown.
- While streaming AND content exists, a small pulsing cursor renders at the end of the content.

## What differs from React (and why)

- **Hooks → signals/effects.** React's `useState` + `useRef` + `useEffect` map to Angular `signal()` + private fields + a single `effect()` that calls `onStreamingChange()`. The state-transition logic was extracted into a method (`onStreamingChange`) so it can be exercised deterministically from tests without relying on the effect/zone tick.
- **`renderSlot` + namespace → standalone components + `<copilot-slot>`.** Mirrors the existing `copilot-chat-assistant-message` pattern. Sub-components are exported as `CopilotChatReasoningMessageHeader`, `CopilotChatReasoningMessageContent`, `CopilotChatReasoningMessageToggle`.
- **`onClick` slot prop → `clickHandler`.** Angular treats `[onClick]` as a DOM event property (NG0306), so the input is renamed `clickHandler`. The slot context object still uses `onClick` to match React's React-friendly shape.
- **`tw-merge` (`twMerge`) → existing `cn` util** (clsx + tailwind-merge wrapper already used across this package). Tailwind classes drop the React project's `cpk:` variant prefix because the Angular package's stylesheet pipeline does not use that prefix (matches sibling components like `copilot-chat-assistant-message-renderer`).
- **No top-level `children` render-prop.** Angular's content-projection API differs from React render-props; consumers override pieces via slot inputs (`headerComponent` / `contentComponent` / `toggleComponent`) or `<ng-template #header>` / `#contentView` / `#toggle`.

## Test coverage → contract bullet mapping

`packages/angular/src/lib/components/chat/__tests__/copilot-chat-reasoning-message.component.spec.ts`

- `formatReasoningDuration` (4 tests) → duration helper exact branches (sub-second / sub-minute / whole minutes / mixed).
- `isStreaming is true only when running and latest` → contract bullet 2.
- `shows 'Thinking…' label while streaming` → contract bullet 3.
- `shows 'Thought for X seconds' label after streaming ends` → contract bullet 4.
- `auto-opens when streaming starts` → contract bullet 5 (open-on-start).
- `auto-collapses when streaming ends without manual interaction` → contract bullet 5 (auto-collapse).
- `manual toggle overrides auto-collapse when streaming ends` → contract bullet 5 (manual override).
- `manual-toggle flag resets when streaming restarts` → contract bullet 5 (flag reset).
- `ignores toggle calls when there is no content` → contract bullet 6.
- `provides slot context computed values while streaming` → contract bullet 9 (slot wiring).
- Header sub-component tests (`label`, `isExpandable`, `handleClick`, `chevronClass`) → contract bullets 3/6/8/9.
- Content sub-component tests (`shouldRender` matrix) → contract bullet 7.

## Test plan

- [x] `nx run @copilotkitnext/angular:test` — 14 files / 68 tests passing (21 new).
- [x] `nx run @copilotkitnext/angular:build` — green.
- [ ] `nx run @copilotkitnext/angular:check-types` — Angular task itself green; the nx target run is blocked by a **pre-existing** failure in `@copilotkit/core:check-types` (`packages/core/src/threads.ts` + `phoenix-observable.ts` errors reproduce on a clean `main` checkout, no Angular sources involved). Reported but not fixed here (out of scope for an Angular-only capability).

## Notes / observations for the worker template

- `ComponentFixture#componentRef.setInput(...)` does **not** propagate to signal-based `input()` declarations in this Angular 19.2 + AnalogJS + vitest jsdom setup. Tests therefore follow the existing assistant-message idiom: `runInInjectionContext` + overwrite the input function with a plain `() => value` closure for unit assertions, with sub-component-level structural assertions split out where DOM coverage was important.
- `[onClick]` cannot be used as a custom-component input name — Angular blocks it as a DOM event property. Worker template should call out: avoid input names matching DOM event handlers.
- Angular's pre-existing repo-wide `nx check-types` is blocked by `@copilotkit/core` failures unrelated to Angular. Workers should treat "Angular task itself green" as the relevant signal until that upstream failure is fixed.

https://claude.ai/code/session_01B1zJ4kMRgLuN1r5T3iiGbZ

---
_Generated by [Claude Code](https://claude.ai/code/session_01B1zJ4kMRgLuN1r5T3iiGbZ)_

### DIFF
--- a/packages/angular/src/lib/components/chat/__tests__/copilot-chat-reasoning-message.component.spec.ts
+++ b/packages/angular/src/lib/components/chat/__tests__/copilot-chat-reasoning-message.component.spec.ts
@@ -87,7 +87,10 @@ describe("CopilotChatReasoningMessage", () => {
 
   it("isStreaming is true only when running and latest", () => {
     const message = makeReasoning("hi");
-    const { component, bindings } = buildComponent({ message, isRunning: true });
+    const { component, bindings } = buildComponent({
+      message,
+      isRunning: true,
+    });
 
     expect(component.isStreaming()).toBe(true);
 
--- a/packages/angular/src/lib/components/chat/__tests__/copilot-chat-reasoning-message.component.spec.ts
+++ b/packages/angular/src/lib/components/chat/__tests__/copilot-chat-reasoning-message.component.spec.ts
@@ -1,0 +1,341 @@
+import {
+  EnvironmentInjector,
+  runInInjectionContext,
+  signal,
+} from "@angular/core";
+import { TestBed } from "@angular/core/testing";
+import { beforeEach, describe, expect, it } from "vitest";
+import type { Message } from "@ag-ui/core";
+import {
+  CopilotChatReasoningMessage,
+  CopilotChatReasoningMessageHeader,
+  CopilotChatReasoningMessageContent,
+  formatReasoningDuration,
+} from "../copilot-chat-reasoning-message";
+import type { ReasoningMessage } from "../copilot-chat-reasoning-message.types";
+
+const reasoningId = "reasoning-1";
+
+function makeReasoning(content = ""): ReasoningMessage {
+  return {
+    id: reasoningId,
+    role: "reasoning",
+    content,
+  } as ReasoningMessage;
+}
+
+interface Bindings {
+  message: ReturnType<typeof signal<ReasoningMessage>>;
+  messages: ReturnType<typeof signal<Message[]>>;
+  isRunning: ReturnType<typeof signal<boolean>>;
+}
+
+function buildComponent(initial: {
+  message: ReasoningMessage;
+  messages?: Message[];
+  isRunning?: boolean;
+}): { component: CopilotChatReasoningMessage; bindings: Bindings } {
+  const injector = TestBed.inject(EnvironmentInjector);
+  const component = runInInjectionContext(
+    injector,
+    () => new CopilotChatReasoningMessage(),
+  );
+
+  const bindings: Bindings = {
+    message: signal(initial.message),
+    messages: signal(initial.messages ?? [initial.message]),
+    isRunning: signal(initial.isRunning ?? false),
+  };
+
+  (component as unknown as { message: () => ReasoningMessage }).message = () =>
+    bindings.message();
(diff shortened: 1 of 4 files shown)


kitos13:
@marthakelly wondering if you are planning to move this from draft? would like to have this feature available in the angular component.

BenTaylorDev:
Closing — Angular parity for this shipped on `main` via the later `inject*`-idiom effort (+ the ENT-1051 drawer work), which supersedes this May draft. Reasoning messages are on main at `packages/angular/src/lib/components/chat/copilot-chat-reasoning-message.ts` (+ utils). Reopen if this analysis missed something.

PR #4656, opened 2026-05-05
feat(angular): port Interrupt API (useInterrupt parity)

marthakelly:
## Summary

Ports React's `useInterrupt` (`packages/react-core/src/v2/hooks/use-interrupt.tsx`) to Angular as `injectInterrupt()`.

## Contract

Listens for `on_interrupt` custom events on the active agent, buffers the latest payload until the run finalizes, optionally filters via `enabled`, optionally pre-processes via `handler`, and surfaces a signal-based store. Calling `resolve(response)` resumes the agent run via `core.runAgent({ agent, forwardedProps: { command: { resume, interruptEvent } } })`.

Lifecycle mirrors React 1:1:
- Custom `on_interrupt` events are buffered locally and only exposed on `onRunFinalized`.
- `onRunStartedEvent` clears state.
- `onRunFailed` discards the buffer (no exposure).
- Sync / async / thenable handlers all accepted; rejection or supersession by a newer interrupt falls back to `result=null` via a per-run token.
- `enabled()` predicate filters interrupts before exposure or handler invocation.
- The active agent is resolved through `injectAgentStore`, so the same runtime sync / `agentId` signal semantics apply.

## Signal + service shape

```ts
injectInterrupt<TValue, TResult>({
  agentId?: string | Signal<string | undefined>,
  enabled?: (event: InterruptEvent<TValue>) => boolean,
  handler?: (props) => TResult | PromiseLike<TResult>,
}): InterruptStoreSignal<TValue, TResult>

interface InterruptStoreSignal<TValue, TResult> {
  event: Signal<InterruptEvent<TValue> | null>;
  result: Signal<TResult | null>;
  isPending: Signal<boolean>;
  resolve: (response: unknown) => void;
}
```

Why no `render` / `setInterruptElement`?
- React's `renderInChat: true` mode publishes a `React.ReactElement` into core for `<CopilotChat>` to slot in. That depends on a React-only extension (`CopilotKitCoreReact.setInterruptElement`) that the Angular `CopilotKit` service does not have, and forcing it into Angular would require a template-form directive — explicitly forbidden by the dispatch.
- The idiomatic Angular alternative is to expose pure signals; consumers render via `@if (interrupt.event(); as event) { ... }` in their own templates and call `interrupt.resolve(...)` from `(click)` handlers. Auto-rendering in `<CopilotChat>` (if desired later) is a separate component-level concern.

## Test → contract mapping

`packages/angular/src/lib/interrupt.spec.ts` (16 behavioral tests, all pass):
- subscribe/teardown lifecycle (subscribers added on construction, released on destroy)
- ignores non-`on_interrupt` custom events
- gates exposure on `onRunFinalized`
- clears on `onRunStartedEvent`
- `resolve()` clears state and dispatches `core.runAgent` with the correct `forwardedProps.command.{resume, interruptEvent}` payload
- `resolve()` is a no-op when no interrupt is pending
- `enabled() === false` blocks exposure AND handler invocation
- `null` result when no handler provided
- sync handler result reflected in `result()`
- async handler resolved value reflected in `result()` after microtask flush; null while pending
- async rejection → `result()=null`, `event()` still pending
- thenable (non-Promise) handler return values supported
- run-failed before finalize discards buffered interrupt
- multiple interrupts in one run keep the latest
- stale handler resolution after a newer interrupt is ignored (per-run token)
- `interruptEvent` in resume payload is the original `event.value` (not the user's response)

## Test plan

- [x] `nx run @copilotkitnext/angular:test` — 63/63 passing (16 new)
- [x] `nx run @copilotkitnext/angular:check-types` — zero errors in the new files; only pre-existing `@copilotkit/core` failures remain (noted in dispatch as pre-existing on main)
- [x] `nx run @copilotkitnext/angular:build` — green

## Out of scope

- A UI component that auto-renders interrupts inside `<CopilotChat>` (separate work — would need a non-directive Angular slot mechanism).
- Refactoring existing services or any non-Angular packages.

https://claude.ai/code/session_01B1zJ4kMRgLuN1r5T3iiGbZ

---
_Generated by [Claude Code](https://claude.ai/code/session_01B1zJ4kMRgLuN1r5T3iiGbZ)_

### DIFF
--- a/packages/angular/src/lib/interrupt.spec.ts
+++ b/packages/angular/src/lib/interrupt.spec.ts
@@ -385,6 +385,33 @@ describe("injectInterrupt", () => {
     expect(interrupt.result()).toBe("handled:second");
   });
 
+  it("discards async handler result when resolve was called before it settled", async () => {
+    let settleHandler: ((value: string) => void) | null = null;
+    const { interrupt } = setUp<string, string>({
+      handler: () =>
+        new Promise<string>((r) => {
+          settleHandler = r;
+        }),
+    });
+
+    agent.emitCustomEvent("on_interrupt", "pending");
+    agent.emitRunFinalized();
+    expect(interrupt.event()?.value).toBe("pending");
+    expect(interrupt.result()).toBeNull();
+
+    // User resolves before the handler promise settles.
+    interrupt.resolve("user-response");
+    expect(interrupt.event()).toBeNull();
+    expect(interrupt.result()).toBeNull();
+
+    // Now the handler's promise settles — its result should be discarded.
+    settleHandler?.("late-handler-result");
+    await flushMicrotasks();
+
+    expect(interrupt.result()).toBeNull();
+    expect(interrupt.event()).toBeNull();
+  });
+
   it("resolve passes the interrupt's value (not the response) as interruptEvent", () => {
     const payload = { question: "approve?", id: 42 };
     const { interrupt } = setUp<typeof payload>();
--- a/packages/angular/src/lib/interrupt.spec.ts
+++ b/packages/angular/src/lib/interrupt.spec.ts
@@ -1,0 +1,407 @@
+import { Component, signal } from "@angular/core";
+import { TestBed } from "@angular/core/testing";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import {
+  AbstractAgent,
+  type AgentSubscriber,
+  type BaseEvent,
+  type RunAgentInput,
+} from "@ag-ui/client";
+import { Observable } from "rxjs";
+import {
+  CopilotKitCore,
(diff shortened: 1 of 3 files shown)


BenTaylorDev:
Closing — this landed independently while the draft sat.

`injectInterrupt()` is on `main` now via `ba41a31d7c` ("feat: implement interrupt handling in AgentStore and add injectInterrupt function"), split across `packages/angular/src/lib/interrupt.ts` and `inject-interrupt.ts`, both re-exported from `public-api.ts`. The shipped version covers the same contract with a bit more: an `InterruptController` class, an explicit `InterruptExpiredError`, and overloads on `injectInterrupt` where this draft had a single signature.

No action needed — just tidying the draft queue.

PR #4684, opened 2026-05-06
fix(angular): add type narrowing for message properties

jpr5:
## Summary
- Fix 7 TS2339 errors in the Angular package caused by `Message` from `@ag-ui/core` resolving to `unknown` (zod v4 lacks `ZodDiscriminatedUnion`, breaking `z.infer`)
- Add structural `BaseMessage` interface in `copilot-chat-message-view.ts` with the minimal shape needed (`id`, `role`, `content`)
- Cast `messagesValue` computed signal through `BaseMessage[]` to fix template and method type errors
- Use `Record<string, unknown>` intermediate cast in `render-tool-calls.ts` type guard for `#getToolMessage`

## Test plan
- [x] `nx run @copilotkitnext/angular:build` passes locally

### DIFF
--- a/packages/angular/src/lib/components/chat/copilot-chat-message-view.ts
+++ b/packages/angular/src/lib/components/chat/copilot-chat-message-view.ts
@@ -18,6 +18,18 @@ import { CopilotChatMessageViewCursor } from "./copilot-chat-message-view-cursor
 import { cn } from "../../utils";
 
 /**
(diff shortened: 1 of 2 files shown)


jpr5:
Closing — the root cause was our branch's lockfile resolving zod@4.4.1 instead of zod@3.25.76. With the corrected lockfile, main's angular code builds without type workarounds.

Issue #84, opened 2023-11-28
useMakeCopilotDocumentReadable - external categories not respected


ataibarkai:
**Describe the bug**
`useMakeCopilotDocumentReadable` supports an `externalContextCategories` variable which lets one bind specific documents to specific textareas (Rather than having a single global state).

The default categories set is `["global"]` and it is respected.
However manually overriding this set is not respected.

**To Reproduce**
```
  useMakeCopilotDocumentReadable(
    clientTranscriptSummaryDocument,
    [salesReplyCategoryId],
    []
  );

// ...

<CopilotTextarea
        ...
        autosuggestionsConfig={{
          textareaPurpose: textareaPurpose,
          contextCategories: [salesReplyCategoryId],
          ...
/>
          ```
**Expected behavior**
When brining up the external context (e.g. using command-k), and typing "@", the document referenced above should appear under 'available sources'. it does not.

PR #83, opened 2023-11-28
bugfix: useMakeCopilotDocumentReadable

ataibarkai:

PR #4719, opened 2026-05-08
feat(inspector): add anonymous interaction telemetry (OSS-96)

marthakelly:
Refs https://linear.app/copilotkit/issue/OSS-96

## What this does

Adds three anonymous interaction events to the inspector: banner viewed, banner clicked, and threads tab clicked. These events are sent to `https://telemetry.copilotkit.ai/ingest` via a direct fetch POST from the browser (no shared dependencies).

Also adds a privacy panel in the inspector settings showing opt-out status, plus a one-time console message after the runtime connects.

Inspector telemetry documentation is folded into the existing `/telemetry` page on `main` — no new docs surface.

**Opt-out:** Set `COPILOTKIT_TELEMETRY_DISABLED=true` or `DO_NOT_TRACK=1` as an environment variable on the runtime process. This disables all telemetry events, URL parameters, and console disclosures in the inspector.

## Acceptance criteria walkthrough

| AC | Code | Status |
|---|---|---|
| Inspector emits `banner_viewed` + `banner_clicked` to `telemetry.copilotkit.ai/ingest` with banner variant identifier | `lib/telemetry.ts` (`TELEMETRY_INGEST_URL`), `index.ts` `fetchAnnouncement` (banner_viewed with `banner_id:timestamp`) + `handleDismissAnnouncement` / `handleAnnouncementContentClick` (banner_clicked with `cta`+`banner_id`+optional `cta_label`) | ✅ |
| PostHog anon distinct ID set on Inspector load | `getOrCreateTelemetryDistinctId()` (UUID v4 in localStorage), sent as `X-CopilotKit-Telemetry-Id` header on every ingest POST (Ben confirmed: lambda parses this as `telemetry_id`). | ✅ |
| Console log on Inspector load + runtime startup | `onRuntimeConnectionStatusChanged` (status=connected) → `maybeShowDisclosure()` (once per browser, skipped when `core.telemetryDisabled`); `BaseCopilotRuntime` + v1 + v2 ctors → `logRuntimeTelemetryDisclosure()` (once per process) | ✅ |
| Inspector UI opt-out (location documented) | Privacy panel (cog settings button) shows ✅ telemetry active or ❌ telemetry disabled based on `core.telemetryDisabled` — display-only. Opt-out is via `COPILOTKIT_TELEMETRY_DISABLED=true` env var on the runtime. | ✅ |
| Opt-out blocks all sends | All three `track*()` call sites in `index.ts` check `core.telemetryDisabled` and short-circuit; `appendRefParam` skips the URL param; `maybeShowDisclosure()` skips the console log. | ✅ |
| Doc page collects/doesn't-collect/opt-out, linked from log + disclosure | Updated `docs/content/docs/(root)/(other)/telemetry/index.mdx`. `TELEMETRY_DOCS_URL` referenced from console log + privacy panel. | ✅ |
| No content in payload | Properties limited to event metadata; **negative test** (`telemetry.test.ts`) greps the wire body for forbidden patterns (`messages`/`completion`/`prompt`/`state_snapshot`/`agent_state`/`content`/`user_id`). Reviewers: please grep call sites in `index.ts` for any unintended payload. | ✅ |

## Verification

- `nx run @copilotkit/web-inspector:test` — 29/29 pass
- `nx run @copilotkit/web-inspector:build` — clean
- `nx run @copilotkit/runtime:test` — 13/13 pass (includes `telemetryDisabled` flag propagation for `COPILOTKIT_TELEMETRY_DISABLED` and `DO_NOT_TRACK`)
- End-to-end playwright: with `COPILOTKIT_TELEMETRY_DISABLED=true`, no telemetry POST in network, privacy panel shows ❌, console disclosure suppressed

## Test plan

- [x] **CORS fix (merged):** PR #83 (lambda CORS headers) + PR #84 (OPTIONS route on API Gateway) merged
- [x] **Distinct ID header (confirmed):** sending `X-CopilotKit-Telemetry-Id` header per Ben Taylor
- [x] verify privacy disclosures (in-product + console + docs page) meet the non-negotiable bar from the ticket
- [ ] confirm consent-banner / CMP alignment and P2 pixel-review status before merge
- [x] allowlist the three `oss.inspector.*` events on the lambda (oss-path-to-production)

---
_Generated by [Claude Code](https://claude.ai/code/session_01Dx4KgZn51RWG8JvysBdpXK)_


BenTaylorDev:
Claude's notes:

1. Race: banner_viewed can fire before core.telemetryDisabled is known

  ensureAnnouncementLoading() is called from connectedCallback (index.ts:4153), and fetchAnnouncement awaits
  the CDN, then fires trackBannerViewed if !this.core?.telemetryDisabled. The runtime-info handshake that
  sets _telemetryDisabled (default false in agent-registry.ts:72) runs concurrently — whichever response wins
   decides. If the announcement CDN beats the runtime-info endpoint, COPILOTKIT_TELEMETRY_DISABLED=true is
  bypassed for the very first banner_viewed event.

  Fix: gate trackBannerViewed (and the disclosure) on core.runtimeConnectionStatus === "connected", or hold
  the call until the first onRuntimeConnectionStatusChanged notification arrives.

  2. No opt-out path when there is no runtime

> how is the telemetry in the inspector related to their being a runtime?  I would think them separate,
  since the inspector runs in the local client

⏺ You're right that they're architecturally separate — the inspector POSTs directly from the browser to
  telemetry.copilotkit.ai/ingest. The runtime is not in the data path at all.

  What the runtime actually does in this PR

  Only one thing: serves GET /info with a new telemetryDisabled: boolean field, set from
  isTelemetryDisabled() (reads COPILOTKIT_TELEMETRY_DISABLED / DO_NOT_TRACK env vars on the runtime process).
   The inspector reads that during its connection handshake and stores it on core.telemetryDisabled, then
  gates each track*() call site on that flag.

  So the runtime is a configuration relay, not a transport. The browser still talks to the lambda on its own.

  Why couple them at all?

  The choice is UX, not necessity:

  - Single knob: one env var on the runtime turns off telemetry everywhere (runtime + inspector). The
  operator doesn't have to set a separate flag for the browser.
  - Reuses existing convention: the runtime already had COPILOTKIT_TELEMETRY_DISABLED for its own
  (pre-existing) server-side telemetry. Threading it into the inspector avoids inventing a second opt-out
  mechanism.
  - Operator-controlled, not end-user-controlled: framing the opt-out as a server-side env keeps the decision
   with the developer running the runtime, not whichever browser happens to load the inspector.

  Why the coupling is awkward (and what I flagged earlier)

  The two issues in my review both fall out of treating a browser-side privacy gate as a server-side config:

  1. Race: the inspector mounts, fetches the announcement CDN, and can fire banner_viewed before the /info
  handshake returns the telemetryDisabled value. Until then, core.telemetryDisabled defaults to false.
  2. No-runtime case: if the inspector is attached with agents__unsafe_dev_only (no runtimeUrl), no handshake
   happens, and there is no opt-out path at all — because the only opt-out lives on a runtime that isn't
  there.

  A cleaner separation would be: inspector owns its own opt-out (localStorage + maybe an in-product toggle +
  a URL/query param the developer can set), and the runtime's telemetryDisabled is just a hint the inspector
  OR-s in. The PR already has the localStorage scaffolding (isTelemetryOptedOut, setTelemetryOptOut) sitting
  unused — making the privacy panel a real toggle instead of display-only would close both gaps without
  giving up the "single knob" convenience.

  3. track()'s own isTelemetryOptedOut() check is effectively dead

  telemetry.ts:62 short-circuits on isTelemetryOptedOut() reading cpk:inspector:telemetry:opt_out. Nothing in
   the source ever calls setTelemetryOptOut(true) — the privacy panel is "display-only" per the PR
  description. So the wrapper-level check fires only if a developer manually sets the key in devtools. All
  real opt-out enforcement lives at the five call sites in index.ts that read core.telemetryDisabled. A new
  call site that forgets the check leaks. Either remove the dead setTelemetryOptOut and the dead localStorage
   check (and rename isTelemetryOptedOut to be explicit), or move the gate inside track() so callers can't
  forget.

BenTaylorDev (approved):

PR #4720, opened 2026-05-08
fix(core): dedup cpki_event_id deliveries inside IntelligenceAgent

dustywusty:
## Summary

When the realtime gateway answers a thread-channel join, it replays the topic's full event history to the joining client. Under React effect-dep churn or transient reconnects, the chat opens 3-5 sequential WebSockets to the same `thread:<id>` topic in a few seconds, so every persisted event hits the page 2-3 times — same `metadata.cpki_event_id` each time. Without a gate, every duplicate re-fires every `agent.subscribe(...)` callback, which surfaces as duplicate rows in the web inspector's AG-UI Events tab and as redundant `notify*Event` mutations downstream (eventually trips a "Message not found" loop).

This adds a per-thread `Set<cpki_event_id>` on `IntelligenceAgentSharedState` and short-circuits inside `createThreadNotifications` when a verbatim replay is detected. Bounded at 2000 entries per thread with FIFO eviction; lives on `sharedState` so clones and recreated delegates share dedup memory.

The dedup gate is **the smallest possible diff** that prevents the user-visible bug. Behavior change is purely additive — no callsite is removed or rerouted, no existing test is altered.

## What's covered

- **6 new unit tests** in `intelligence-agent.test.ts`:
  - duplicate `cpki_event_id` is delivered once, not twice/three times
  - distinct event ids on the same payload shape are both delivered
  - payloads without `cpki_event_id` are passed through unchanged
  - dedup is scoped per-thread (same id on a different thread is delivered)
  - duplicate `RUN_FINISHED` in run mode still completes the observable cleanly (terminator preserved)
  - duplicate `RUN_ERROR` in run mode still errors the observable cleanly

- All 425 existing core tests pass unchanged.

- Pre-commit hook (full nx test suite across packages) passed.

## Reproduction + A/B verification

Reproduced **13/13 deterministically** with a Playwright recipe that opens two threads, sends a long-running prompt on thread A, switches to thread B mid-flight, then switches back. Each run produces the same fixed-shape set of duplicates the dispatcher would otherwise propagate:

```
1 RUN_STARTED       (matches Tyler's f11 — two RUN_STARTED rows w/ same id)
1 RUN_FINISHED      (matches Tyler's f02 — three RUN_FINISHED rows w/ same id)
1 MESSAGES_SNAPSHOT
22 STATE_SNAPSHOT   (matches Tyler's f11 — three STATE_SNAPSHOT rows w/ same id)
4 STEP_STARTED + 4 STEP_FINISHED
1 TEXT_MESSAGE_START + 1 TEXT_MESSAGE_END
~70-100 TEXT_MESSAGE_CONTENT (stream-length dependent)
~150-220 RAW
```

A/B-tested by toggling the early-return inside the dedup branch:

| State | WS dups (wire) | Dispatched dups (subscribers) |
|---|---|---|
| Fix on  | 568-955 / run | **0** (~290 caught) |
| Fix off | ~500 / run | **~290** propagate |

## Memory / safety

- `Set<eventId>` is bounded at `DEDUP_MAX_PER_THREAD = 2000` with FIFO eviction (`Set` insertion order).
- `cpki_event_id` is a UUID minted per persisted event by the runner, so collisions are vanishingly improbable; no false-positive dedup risk.
- Dedup is keyed by `(threadId, eventId)` — events on different threads never collide.
- Terminator semantics for RUN_FINISHED / RUN_ERROR preserved on duplicates, so run-mode observables still complete cleanly when a duplicate happens to be the terminal event.

## Out of scope (separate tickets to file)

1. The underlying React effect-dep churn that opens 3-5 sockets per thread switch. The dispatcher dedup is defense in depth; tracing why `useAgent`'s WeakMap key is being invalidated is a separate excavation.
2. Dropping `clearReconnectCursor(input.threadId)` inside `connect()` so reconnects resume from `lastSeenEventId` instead of forcing a full historical replay. Cheap and meaningful efficiency win, but changes existing test coverage and is independent of the user-visible fix here.

## Test plan

- [ ] Pull the canary (`pkg.pr.new` build for this PR) into `gen-ui-studio` (a fresh `npx copilotkit create`) and re-run Tyler's repro: open the inspector, two threads, send + switch + switch-back. AG-UI Events tab should show one row per `cpki_event_id`, not 2-3.
- [ ] Validate against existing Tyler-affected workflow on staging.


MikeRyanDev (changes requested):
Blocking issue found in the fresh thread-restore path.

`connect()` deliberately clears the replay cursor (`packages/core/src/intelligence-agent.ts:314` -> `clearReconnectCursor`, with `last_seen_event_id: null`) so switching back to a thread can request a full historical replay. The new `deliveredEventIds` table is also stored on shared per-thread state, but it is never cleared on that fresh restore path. Once an event has been seen, `createThreadNotifications` drops the same `cpki_event_id` for that thread (`packages/core/src/intelligence-agent.ts:658-659`).

That means the A -> B -> A restore case can now fail to rehydrate: the gateway's full replay for thread A will contain the same persisted `cpki_event_id`s that were delivered the first time A was opened, and this client will suppress them before `connectAgent` can rebuild messages/state. The existing restore test does not catch it because the second thread-A replay uses `event-a-2`; a persisted replay should reuse `event-a-1`.

Please either clear the delivered-event table when intentionally clearing the reconnect cursor for a fresh restore, or scope dedup to a connection/replay window instead of the whole thread lifetime. Add a regression test that opens A, switches away, clears local messages, then replays the same A event id and still rehydrates.

Validation I ran locally:
- `pnpm nx run @copilotkit/core:test -- src/__tests__/intelligence-agent.test.ts` passed: 51 tests
- `pnpm nx run @copilotkit/core:build` passed

PR #4740, opened 2026-05-08
ENT-510 - fix(core): only reset agent state on actual thread switch (supersedes #4720)

dustywusty:
## Summary

Supersedes #4720, which Mike correctly identified as introducing a blocking regression in the fresh-thread-restore path. This PR takes Mike's framing as the authoritative restatement of the bug and lands the proxy/orchestrator-level fix that addresses both halves of Tyler's symptom without that regression.

## The actual bug, restated

When the chat's connect effect re-fires due to React effect-dep churn, it calls `copilotkit.connectAgent({ agent })` on the same thread again. `RunHandler.connectAgent` previously did `agent.setMessages([])` + `agent.setState({})` and (transitively) called `clearReconnectCursor` on every invocation, **including the churn re-connects**. That forced the realtime gateway to replay the topic's full event history every time the chat re-opened a socket — sending the same persisted `cpki_event_id`s 2-3 times per thread switch — and produced both halves of Tyler's video:

- **Cosmetic surface**: duplicate `cpki_event_id` rows in the inspector AG-UI Events tab.
- **Functional surface**: `agent.messages` got cleared, the replay started rebuilding it, then another churn re-connect came along, called `setMessages([])` again, and the next `runAgent` fired with `messages.length = 1` while the server had N. `prepareRegenerateStream` triggered, the new user-message id wasn't in any checkpoint, and the server threw "Message not found".

PR #4720 only addressed the cosmetic surface and made the functional surface deterministic. Mike's review caught it.

## What this PR changes

- **`RunHandler.connectAgent`** tracks `_lastConnectedThreadId` across calls. The `setMessages([])` + `setState({})` + replay-cursor clear are now gated on the threadId actually changing. Same-thread churn re-connects preserve local state and let the gateway resume from `lastSeenEventId` instead of replaying. Thread switches still wipe state and ask for a full replay.
- **`IntelligenceAgent.connect()`** no longer auto-clears the replay cursor. Cursor management is the caller's decision now. `clearReconnectCursor` is made `public` so the orchestrator (and tests) can call it explicitly.
- **`ProxiedCopilotRuntimeAgent`** exposes a public `clearReplayCursor(threadId)` method that delegates to the IntelligenceAgent's `clearReconnectCursor`. Non-Intelligence runtime modes are a safe no-op.

The dispatcher-level dedup from #4720 is **removed entirely**. The orchestrator-level gate is sufficient and a much smaller, more obviously correct surface.

## Why this approach

Mike's two suggestions in the #4720 review were "clear the dedup table when the cursor is cleared" or "scope dedup to a connection/replay window." Both alone left a regression in the other direction:
- Coupling dedup to `clearReconnectCursor` (which was unconditional) reset dedup on every churn re-connect, so the original symptom returned.
- Per-connection-window dedup couldn't catch dups across the multiple sockets that fire during churn.

The right move is one level up: stop the underlying state-reset cascade that was forcing the replay storm in the first place. Once same-thread churn no longer wipes local state, the gateway can resume cleanly, the wire-level replays don't happen, and both Tyler's symptoms disappear naturally — no dispatcher dedup needed.

## Tests

- **`intelligence-agent.test.ts`**: 4 existing tests updated to reflect the new contract. `connect()` preserves the cursor by default; `clearReconnectCursor()` is the explicit primitive. Includes the regression test Mike asked for: replay the same `cpki_event_id` after `A → B → A` and verify the rehydrate still produces the user message.
- **`core-connect-thread-switch.test.ts`** (new): 4 focused tests for the orchestrator gate — same-thread churn does not reset, cross-thread switch does reset, `A → B → A` resets each transition, and a non-Intelligence runtime path is handled cleanly.

All 430 core tests pass:
```
Test Files  40 passed (40)
     Tests  430 passed (430)
```

## Test plan

- [ ] Pull the canary build into `gen-ui-studio` (a fresh `npx copilotkit create`).
- [ ] Manually walk Tyler's recipe with the inspector open: open thread A, switch to B, switch back to A. AG-UI Events tab should show one row per `cpki_event_id`. No "Message not found" toast.
- [ ] Send a message after the switch-back. Run completes cleanly.
- [ ] (Repro harness) `recipe-rehydrate.spec.ts` from kingston's transport-dup-repro asserts on user-visible state (DOM messages + absence of "Message not found"). Should pass against this canary; fails against `origin/main` and against #4720's commit.

## Why "supersedes #4720" not "build on top of"

The dispatcher dedup in #4720 isn't load-bearing under this fix — once the orchestrator stops cascading state-resets, there's no replay storm for the dispatcher to filter. Keeping the dedup would just be unused machinery with non-trivial memory accounting. Cleaner to drop it entirely and revisit only if a real overlap scenario surfaces later (network blips, retries, etc.).

### DIFF
--- a/.changeset/fix-thread-switch-state-reset.md
+++ b/.changeset/fix-thread-switch-state-reset.md
@@ -1,0 +1,13 @@
+---
+"@copilotkit/core": patch
+---
+
+fix(core): only reset agent state and replay cursor on actual thread switch
+
+`RunHandler.connectAgent` previously called `agent.setMessages([])` + `agent.setState({})` and the IntelligenceAgent delegate's `clearReconnectCursor` on every connect, including effect-dep churn re-connects on the same thread. That forced the realtime gateway to replay the topic's full event history every time the chat re-opened a socket — which under React effect churn happens 3-5 times per thread switch — and produced both halves of Tyler's bug: duplicate `cpki_event_id` rows in the inspector and intermittent "Message not found" toasts.
+
+This change tracks the most recent connected threadId on `RunHandler` and gates the state-reset + cursor-clear on that threadId actually changing. Same-thread churn re-connects now preserve local messages/state and the gateway can resume from `lastSeenEventId`. Actual thread switches still wipe local state and ask for a full historical replay.
+
+Supersedes #4720, which attempted to solve only the inspector duplicate-row symptom by adding a dispatcher dedup but introduced a regression in the A → B → A restore path (the dedup blocked the gateway's restore replay for thread A on second visit). The dispatcher dedup is removed entirely; the orchestrator-level gate is sufficient.
+
+Adds focused unit tests covering: same-thread churn does not reset, cross-thread switch does reset, A → B → A resets each transition, and a regression test that replays the same `cpki_event_id` after a thread switch and verifies rehydrate still works.
--- a/packages/core/src/__tests__/core-connect-thread-switch.test.ts
+++ b/packages/core/src/__tests__/core-connect-thread-switch.test.ts
@@ -1,0 +1,136 @@
+import type { AbstractAgent } from "@ag-ui/client";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { CopilotKitCore } from "../core";
+
+/**
+ * Tests for RunHandler.connectAgent's thread-switch detection. The
+ * orchestrator gates `setMessages([])` + `setState({})` + replay-cursor
+ * clear on the agent's threadId actually changing between successive
+ * connectAgent calls. Same-thread churn re-connects must preserve
+ * local state so the gateway can resume from `lastSeenEventId` instead
+ * of forcing a full historical replay.
+ */
+
+class StubAgent {
+  public agentId: string;
+  public threadId: string | undefined;
+  public messages: unknown[] = [{ id: "preexisting" }];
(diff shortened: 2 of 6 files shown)


MikeRyanDev (approved):
Reviewed the state-reset/cursor ownership change in detail. This looks merge-ready to me.

What I checked:
- `RunHandler.connectAgent` now gates `setMessages([])`, `setState({})`, and replay-cursor clearing on an actual `agent.threadId` change, while still detaching the active run on every reconnect.
- `IntelligenceAgent.connect()` now preserves the reconnect cursor by default, and `clearReconnectCursor(threadId)` gives the caller explicit control over full replay vs resume.
- `ProxiedCopilotRuntimeAgent.clearReplayCursor(threadId)` delegates cursor clearing for Intelligence mode and no-ops for non-Intelligence transports.
- The #4720 regression is covered: A -> B -> A now clears the cursor intentionally and rehydrates from a replay that reuses the same persisted `cpki_event_id`.

Validation:
- `pnpm nx run @copilotkit/core:test -- src/__tests__/core-connect-thread-switch.test.ts src/__tests__/intelligence-agent.test.ts` passed: 50 tests
- `pnpm nx run @copilotkit/core:build` passed
- GitHub checks are green

Only note: `git diff --check origin/main...HEAD` still reports the existing CRLF/trailing-whitespace noise in `showcase/harness/fixtures/d5/agent-config.json`, unrelated to this PR's touched files.

PR #4748, opened 2026-05-09
fix(showcase/langgraph-python): collapse gen-ui-agent to single set_steps call

tylerslaton:
## Summary

`gen-ui-agent` (Agent State) D5 cell has been intermittently flapping at **~22% red rate** with two distinct error modes:
- `timeout: assistant did not respond within 60000ms (baseline=0, current=0)`
- `expected [data-testid="agent-state-card"] to mount within 60000ms`

## Root cause

The system prompt instructed the agent to call `set_steps` **seven times**:
1. Plan all 3 with `status="pending"`
2. Step 1: in_progress → completed (2 calls)
3. Step 2: in_progress → completed (2 calls)
4. Step 3: in_progress → completed (2 calls)
5. Final assistant message

But the **aimock fixture returns all 3 steps already at `status="completed"`** in a single response. The LLM (gpt-4o-mini @ temp=0) sees both the iterate-7-times prompt AND the all-completed return — sometimes it follows the prompt and tries to iterate (burning supersteps + emitting tool calls aimock can't disambiguate), sometimes it reads the all-completed state and terminates early. Non-deterministic.

## Fix

Simplify the system prompt to a **single `set_steps` call** with all three steps emitted at once, matching the fixture shape. Trade-off: lose progressive visual transitions in the demo (which the fixture wasn't producing anyway), gain deterministic agent-loop behavior across both fixture-driven probe runs and real-LLM testing. The agent-state-card renders the same either way.

## Test plan

- [x] System prompt now describes a single set_steps call with all 3 steps at appropriate final status
- [ ] Post-merge: `gen-ui-agent` should hold green across consecutive e2e-deep ticks (was 7/9 green before; should be 9/9 with the deterministic flow)

### DIFF
--- a/showcase/integrations/langgraph-python/src/agents/gen_ui_agent.py
+++ b/showcase/integrations/langgraph-python/src/agents/gen_ui_agent.py
@@ -69,28 +69,33 @@ def set_steps(
 
 
 SYSTEM_PROMPT = (
-    "You are an agentic planner. For each user request, follow this exact "
-    "sequence:\n"
-    "1. Plan exactly 3 concrete steps and call `set_steps` ONCE with all "
-    'three steps at status="pending".\n'
-    '2. Step 1: call `set_steps` with step 1 at status="in_progress", '
-    'then call `set_steps` again with step 1 at status="completed".\n'
-    '3. Step 2: call `set_steps` with step 2 at status="in_progress", '
-    'then call `set_steps` again with step 2 at status="completed".\n'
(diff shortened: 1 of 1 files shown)


tylerslaton:
Closing — diagnosis was wrong. The system prompt doesn't reach aimock's fixture matcher, so changing it has no effect on the probe-driven flow. Aimock just matches by userMessage substring + optional toolName/toolCallId/hasToolResult, and the agent loop is fully determined by fixture responses (regardless of prompt). Need to dig into a different root cause for the gen-ui-agent flake.

PR #4749, opened 2026-05-09
fix(docs): generate proper markdown llms.txt

tamish-max:
## Summary
- generate `docs.copilotkit.ai/llms.txt` from the Fumadocs source instead of redirecting to the MCP index
- add `/llms-full.txt` and per-page `.md` markdown routes for AI-friendly docs access
- update docs copy to point to the new docs-hosted llms endpoints

## Validation
- `npx next build`

### DIFF
--- a/docs/app/llms-full.txt/route.ts
+++ b/docs/app/llms-full.txt/route.ts
@@ -1,0 +1,20 @@
+import { source } from "@/app/source";
+import { getLLMText } from "@/lib/get-llm-text";
+
+export const revalidate = false;
+
+export async function GET() {
+  const scanned = await Promise.all(
+    source
+      .getPages()
+      .slice()
+      .sort((a, b) => a.url.localeCompare(b.url))
+      .map(getLLMText),
+  );
+
+  return new Response(scanned.join("\n\n"), {
+    headers: {
+      "Content-Type": "text/markdown; charset=utf-8",
+    },
+  });
+}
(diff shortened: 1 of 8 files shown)


lukasmoschitz:
Thanks for this, @tamish-max — and apologies for the long silence on what is your first PR here.

The idea was right, and it shipped. `/llms.txt` no longer redirects to `mcp.copilotkit.ai`; the docs site generates all three surfaces you proposed:

- **`/llms.txt`** — markdown index, built from the same page enumerator the sitemap uses so the index can't drift from the actual route table
- **`/llms-full.txt`** — full corpus in one file, with `<Snippet />` tags resolved into fenced code blocks so a model sees real code rather than JSX it can't interpret
- **per-page markdown** — `<path>.md` (and `.mdx`) rewrite onto a route handler that reuses the same content resolution as the live page

Those live under `showcase/shell-docs/src/app/` and landed in `5728611dfd` on 2026-05-20, eleven days after you opened this. Independent work rather than a port of your branch, so I can't credit you on the commit in good conscience — but you identified the gap correctly and ahead of us, and the copy changes you made (pointing at `docs.copilotkit.ai` instead of `mcp.copilotkit.ai`) are in place too.

Separately, this branch can no longer be merged in any form: it targets `docs/app/`, `docs/lib/` and `docs/next.config.mjs`, and the standalone docs app was retired in `33a79485b6` on 2026-06-18. `docs/` is now just a symlink to `showcase/shell-docs`.

Closing as already-shipped. Sorry again for the wait, and thanks for the contribution — if you want to pick something else up, the `shell-docs` tree is where that work lives now.

PR #4754, opened 2026-05-10
docs: fix useCopilotContext import path in persistence guides

### LINKED ISSUE #3860: 📚 Documentation: Inconsistence on LangChain persistence hook documentation vs Typescript library on useCopilotContext

In https://docs.copilotkit.ai/integrations/langgraph/advanced/persistence/loading-message-history#using-setthreadid the docs mentioned 'import { useCopilotContext } from "@copilotkit/react-core/v2";'

However, Module '"@copilotkit/react-core/v2"' has no exported member 'useCopilotContext'.

So useCopilotContext is only available as v1 but not as v2?

serhiizghama:
## Problem

In the LangGraph and CrewAI Flows persistence guides, the "Using setThreadId" code samples import `useCopilotContext` from `@copilotkit/react-core/v2`:

```tsx
import { useCopilotContext } from "@copilotkit/react-core/v2";
```

This subpath does not export `useCopilotContext` — only the V1 entry (`@copilotkit/react-core`) does, via [`packages/react-core/src/context/index.ts`](https://github.com/CopilotKit/CopilotKit/blob/main/packages/react-core/src/context/index.ts). Copy-pasting the snippet fails to compile with:

```
Module '"@copilotkit/react-core/v2"' has no exported member 'useCopilotContext'.
```

Reported in #3860.

## Solution

Update the import path in both guides to `@copilotkit/react-core` so the snippet matches the actual public API.

Files changed:
- `docs/content/docs/integrations/langgraph/advanced/persistence/loading-message-history.mdx`
- `docs/content/docs/integrations/crewai-flows/persistence/loading-message-history.mdx`

## Testing

Doc-only change; no runtime behavior. Verified that `useCopilotContext` is exported from `@copilotkit/react-core` and not from the `/v2` subpath.

Closes #3860

### DIFF
--- a/docs/content/docs/integrations/crewai-flows/persistence/loading-message-history.mdx
+++ b/docs/content/docs/integrations/crewai-flows/persistence/loading-message-history.mdx
@@ -69,7 +69,7 @@ const YourApp = ({ setThreadId }) => {
 CopilotKit provides the current `threadId` and a `setThreadId` function from the `useCopilotContext` hook:
 
 ```tsx
-import { useCopilotContext } from "@copilotkit/react-core/v2";
+import { useCopilotContext } from "@copilotkit/react-core";
 
 const ChangeThreadButton = () => {
   const { threadId, setThreadId } = useCopilotContext();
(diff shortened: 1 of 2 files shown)


serhiizghama:
Closing — the two files this PR modifies were moved during a docs migration and no longer exist at the original paths. The new docs location doesn't contain the buggy import path, so the issue is effectively resolved by the restructure. Thanks for the patience!

PR #4768, opened 2026-05-11
Docs(DeepAgents): Interrupt Flow Python Code Fixes

MalaikaAbb:
## Summary
In python variant of DeepAgents, fixed the middleware to work correctly by accepting agent_name, and using expose_state on CopilotKitMiddleware to inject dynamic agent_name in system prompt. Also fixed missing agentId in useInterrupt Hook

## Changes

1. used `CopilotKitMiddleware(expose_state=["agent_name"])` in agent configuration so agent_name can be injected in agent's system prompt and can be recognized. Otherwise, it was only visible in AG-UI events inspector, and agent would not respond to the given agent_name that the user would set

2. Added `agentId: "sample_agent"` in `useInterrupt()` without it, the following error is thrown 

```
useAgent: Agent 'default' not found after runtime sync (runtimeUrl=/api/copilotkit).
Known agents: [sample_agent]
```
This is because `useInterrupt` calls `useAgent({ agentId: config.agentId })` and `useAgent` falls back to the literal string `"default"` when no `agentId` is given.

### DIFF
--- a/docs/content/docs/integrations/deepagents/generative-ui/your-components/interrupt-based.mdx
+++ b/docs/content/docs/integrations/deepagents/generative-ui/your-components/interrupt-based.mdx
@@ -104,7 +104,7 @@ We're going to have the agent ask us to name it, so we'll need a state property 
         <Tab value="Python">
             ```python title="agent.py"
             from typing import Any
-            from copilotkit import CopilotKitState
+            from copilotkit import CopilotKitState, CopilotKitMiddleware
             from langchain.agents.middleware import AgentMiddleware # [!code highlight]
             from langgraph.runtime import Runtime
             from langgraph.types import interrupt # [!code highlight]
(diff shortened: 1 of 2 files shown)


BenTaylorDev:
Thanks for improving the DeepAgents interrupt-flow example — turning the illustrative fragment into a more complete, runnable snippet is exactly the right direction.

Closing this for now, though: These DeepAgents pages were substantially rewritten after this PR was opened (the examples moved to the v2 `@copilotkit/react-core/v2` imports), and docs now live in `showcase/shell-docs/src/content/` rather than the `docs/content/docs/` tree this PR targets — so it can no longer be applied as-is. The underlying gap is still worth filling, so if you're up for it, a fresh PR against the current `showcase/shell-docs/src/content/docs/integrations/deepagents/human-in-the-loop/interrupt-flow.mdx` (and the `generative-ui/your-components/interrupt-based.mdx` page) would be very welcome.

PR #4773, opened 2026-05-11
feat: add X-AIMock-Strict header propagation across SDKs and integrations

jpr5:
## Summary

Per-request `X-AIMock-Strict` header propagation through the full CopilotKit stack — from Playwright browser contexts through AG-UI to outgoing LLM API calls. Enables the same aimock instance to serve both strict fixture-only probes and live demo traffic simultaneously.

- **Harness**: All probe browser contexts send `X-AIMock-Strict: true` via `extraHTTPHeaders` (9 sites across 5 driver files)
- **Python SDK**: `ContextVar` + `httpx` event hook — incoming `x-aimock-*` headers forwarded to outgoing LLM calls automatically via FastAPI integration (covers 12/17 showcase integrations)
- **TypeScript SDK**: `AsyncLocalStorage` + per-call headers merged into `wrapModelCall` middleware (covers 3 more, 15/17 total)
- **.NET integration**: `AsyncLocal` + ASP.NET middleware + `PipelinePolicy` stopgap — all 12 agent factories wired through `CreateOpenAIClientOptions()`. TODO markers for migration to copilotkit-sdk-dotnet when Microsoft delivers it.
- **Java/Spring integration**: `InheritableThreadLocal` + `HandlerInterceptor` + `ClientHttpRequestInterceptor` + WebClient `ExchangeFilterFunction` — covers both RestClient and WebClient paths

## Test plan

- [x] Python SDK: 14 new tests (ContextVar isolation, httpx hook injection, key lowercasing, warning on bad client), 116 total pass
- [x] TypeScript SDK: 8 new tests (AsyncLocalStorage isolation, header filtering, concurrent requests), 92 total pass
- [x] Harness: 1590 tests pass (1 pre-existing failure unrelated)
- [x] CR converged: 2 rounds × 14 agents, 6 bucket(a) findings fixed, 0 remaining

### DIFF
--- a/sdk-python/copilotkit/header_propagation.py
+++ b/sdk-python/copilotkit/header_propagation.py
@@ -1,7 +1,8 @@
-"""Header propagation for forwarding x-aimock-* headers to outgoing LLM calls.
+"""Header propagation for forwarding x-* prefixed headers to outgoing LLM calls.
 
 Uses Python contextvars for per-request ambient state in async FastAPI handlers.
 An httpx event hook reads the ContextVar and injects headers on outgoing requests.
+Matches the CopilotKit runtime's extractForwardableHeaders() behavior.
 """
 
 import contextvars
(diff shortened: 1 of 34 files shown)


sdk-python/copilotkit/header_propagation.py
```diff
@@ -0,0 +1,57 @@
+"""Header propagation for forwarding x-aimock-* headers to outgoing LLM calls.
+
+Uses Python contextvars for per-request ambient state in async FastAPI handlers.
+An httpx event hook reads the ContextVar and injects headers on outgoing requests.
+"""
```
tylerslaton: I think we should make correctly forwarding the headers generally work - we don't want this to be specific to only AI Mock. Forwarding headers is something we already want to work. 

1. I'm surprised this doesn't already work - are we sure it doesn't right now?
2. If it does not already work, we want all headers to go through. Then it is just a side-effect that we can pass AI Mock headers.
3. If it does already work, none of this code is necessary and we should just send the header like normal.

tylerslaton (changes requested):
Seeing that we're making this logic specific to AI Mock headers. We want all headers to forward correctly, not just AI Mock.

PR #4802, opened 2026-05-13
Docs(DeepAgents): Improved State Rendering Python Code Example

MalaikaAbb:
## Summary
The page's Python snippet defines `emit_research_progress` as a free-floating function there's ambiguity on how to register it as a tool or in middleware, at least one example should be elaborated, we used a tool approach. A reader following the page exactly ends up with a function nobody calls. When they prompt the chat with *"add the search for cats vs dogs"*, the agent falls back to the built-in `grep` tool, searches an empty virtual filesystem, and replies *"No matches found in the files."*

## Changes
- Added `AgentSearchMiddleware` because without it, 'searches' are not received on the frontend and cannot be seen in AG-UI inspector either during tool invocation. We see the state section go empty before tool call finishes, without any viable emitted state. 

- Changed `emit_research_progress` to `research `because  `emit_research_progress` reads like internal logging code — the LLM avoids it. `research` matches what the user actually says ("research X", "research for X"). Without the rename, even after registration the agent may still skip this tool in favor of the built-in `task` or `grep`.

- Added docstring because when a plain Python function is passed to `tools=[...]`, the framework uses the docstring as the tool's description for the LLM

- Defined Agent with relevant middlewares

## Results
- Following these changes, each emitted state change event is recognized by the frontend and relevant symbols with queries are rendered on the UI while the state is visible.

### DIFF
--- a/docs/content/docs/integrations/deepagents/generative-ui/state-rendering.mdx
+++ b/docs/content/docs/integrations/deepagents/generative-ui/state-rendering.mdx
@@ -44,9 +44,14 @@ Use state rendering when you want to:
       <Tab value="Python">
         ```python title="agent.py"
         from copilotkit import CopilotKitState
+        from langchain.agents.middleware import AgentMiddleware
+        from typing import Any
 
         class AgentState(CopilotKitState):
             searches: list[dict]
+
+        class AgentSearchMiddleware(AgentMiddleware[AgentState, Any]):
+            state_schema = AgentState
         ```
       </Tab>
       <Tab value="TypeScript">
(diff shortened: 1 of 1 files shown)


BenTaylorDev:
Thanks for this — fleshing out the state-rendering example with a real tool + middleware + `create_deep_agent(...)` wiring addresses a genuine gap (the live page still shows just a bare `emit_research_progress` fragment).

Closing for now, with a couple of notes for a refreshed version: These DeepAgents pages were substantially rewritten after this PR was opened (the examples moved to the v2 `@copilotkit/react-core/v2` imports), and docs now live in `showcase/shell-docs/src/content/` rather than the `docs/content/docs/` tree this PR targets — so it can no longer be applied as-is. Also, in the reworked `research` example the state update reassigns a local `state = { ... }` rather than mutating/emitting the agent state (the original used `state["searches"] = ...`), and the function isn't registered as a tool — worth tightening so the example actually streams state. A new PR against `showcase/shell-docs/src/content/docs/integrations/deepagents/generative-ui/state-rendering.mdx` would be welcome.

PR #4818, opened 2026-05-14
Docs(DeepAgents): State Rendering - Typescript Improvements

MalaikaAbb:
## Summary
The page's typescript snippet defines `emitResearchProgress` as a free-floating function there's ambiguity on how to register it as a tool or in middleware, at least one example should be elaborated, we used a tool approach. A reader following the page exactly ends up with a function nobody calls. When they prompt the chat with *"add the search for cats vs dogs"*, the agent falls back to the built-in `grep` tool, searches an empty virtual filesystem, and replies *"No matches found in the files."*

## Changes
- Added `searchesStateMiddleware` because without it, 'searches' are not received on the frontend and cannot be seen in AG-UI inspector either during tool invocation. We see the state section go empty before tool call finishes, without any viable emitted state. used zodState() so 'searches' can be seen by AG-UI and STATE_SNAPSHOT

- Changed `emitResearchProgress` to `research` for better readability and tool call invocation  

- Defined Agent with relevant middlewares

## Results
- Following these changes, each emitted state change event is recognized by the frontend and relevant symbols with queries are rendered on the UI while the state is visible.

### DIFF
--- a/docs/content/docs/integrations/deepagents/generative-ui/state-rendering.mdx
+++ b/docs/content/docs/integrations/deepagents/generative-ui/state-rendering.mdx
@@ -54,15 +54,22 @@ Use state rendering when you want to:
         import { createMiddleware } from "langchain";
         import { copilotkitMiddleware } from "@copilotkit/sdk-js/langgraph"; // [!code highlight]
         import { z } from "zod";
+        import { zodState } from "@copilotkit/sdk-js/langgraph";
+        import { StateSchema } from "@langchain/langgraph";
+
+        // Define the agent's custom state and include schema in middleware.
+        const AgentStateSchema = new StateSchema({
+            searches: zodState(
+                            z.array(z.object({
+                                query: z.string(),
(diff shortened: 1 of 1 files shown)


BenTaylorDev:
Thanks for the TypeScript state-rendering improvements. Note that part of this already landed — the current page adopted `zodState` independently during a rewrite. The remaining pieces (the runnable `research` tool + `createDeepAgent` wiring) are still absent and worth adding.

Closing this PR since These DeepAgents pages were substantially rewritten after this PR was opened (the examples moved to the v2 `@copilotkit/react-core/v2` imports), and docs now live in `showcase/shell-docs/src/content/` rather than the `docs/content/docs/` tree this PR targets — so it can no longer be applied as-is. A fresh PR against the current `showcase/shell-docs/src/content/docs/integrations/deepagents/generative-ui/state-rendering.mdx` covering just the still-missing example wiring would be a clean follow-up.

PR #4820, opened 2026-05-14
chore(ci): tighten CI/CD security — SHA pins, zizmor, pnpm v11, audit fixes

AlemTuzlak:
## Summary

Comprehensive CI/CD security hardening pass over the 33 workflows in `.github/workflows/`. Inspired by the TanStack security posture: pin every action to a SHA, bump to pnpm v11 for stronger supply-chain protections, add zizmor static analysis, and run an agentic security audit on top.

This PR is **CI-only** — no package code, no docs. The only non-`.github/` changes are `package.json` (`packageManager` field) and `pnpm-lock.yaml` (regenerated for the pnpm 10→11 bump).

### What changed

1. **All `uses:` SHA-pinned** (167 occurrences across 33 workflows). Each pin carries an inline `# vX.Y.Z` comment so reviewers can still tell what version is wired up. Tag-style refs (`@v4`) are mutable and have been weaponized in past supply-chain attacks; SHAs are immutable.

2. **`packageManager: pnpm@11.1.2`** + regenerated lockfile. pnpm v11 ships hardened script-execution defaults (the existing `onlyBuiltDependencies: [better-sqlite3]` in `pnpm-workspace.yaml` continues to be honored). Workflow-level `version: "10.13.1"` hardcodes were removed so `pnpm/action-setup` inherits from `packageManager` — single source of truth, no more silent drift.

3. **Dependabot for `github-actions`** (`.github/dependabot.yml`). Without this, SHA pins rot fast and new upstream advisories never reach us. Minor/patch bumps are grouped; majors stay separate for real review.

4. **Zizmor static analysis** (`.github/zizmor.yml` + `.github/workflows/security_zizmor.yml`). Blocking on PR, runs on push to main, weekly schedule for advisory drift. Catches the well-known classes of Actions footguns: template injection, dangerous triggers, unpinned uses, excessive token scopes, secret exfil patterns.

5. **Workflow hardening from the zizmor baseline + audit-agent findings**:
   - `persist-credentials: false` on every `actions/checkout` except 7 workflows that legitimately push back via the workflow token (each carries a justifying comment).
   - 17 attacker-controllable expansions routed through `env:` with quoted shell vars instead of direct `${{ }}` interpolation.
   - Per-job `permissions:` blocks on 14 workflows; `id-token: write` demoted from workflow-level to the specific Depot-runner jobs; `pull-requests: write` / `actions: write` narrowed to the jobs that actually call those APIs.

### Audit-driven fixes

Two parallel security agents reviewed the post-zizmor state. Real findings remediated in this PR:

| Finding | Fix |
| --- | --- |
| `publish-release.yml` build job uploaded its workspace as a 1-day artifact with `.git/config` (containing a persisted GITHUB_TOKEN) — recoverable by anyone with `actions:read`. | Dropped `token:` and added `persist-credentials: false`; publish job sets up its own `insteadOf` auth. |
| `auto_merge_showcases.yml` checked team membership of `context.actor`, not the PR author. A team member synchronizing or reopening an outsider's PR would auto-merge unreviewed code. | Now authorizes on `pull_request.user.login` (immutable for the PR's lifetime). |
| `static_quality.yml` format job ran `pipx install ruff` (unpinned) with a persisted write-scoped PR token in the workspace. | Pinned `ruff==0.15.13`. |
| `showcase_capture-previews.yml` concatenated workflow_dispatch inputs into a space-joined arg string. Whitespace or shell metacharacters in a slug would re-tokenize. | Switched to a bash array. |

Final zizmor result: **0 high, 0 medium** (down from 28 high + 54 medium baseline). All suppressions carry per-finding justifications in `zizmor.yml`.

## Test plan

- [ ] Zizmor workflow runs and passes on this PR
- [ ] All existing CI workflows still complete (static_quality, test_unit, integrations_parity, etc.) — confirms `permissions:` tightening hasn't broken any job's actual API needs
- [ ] `pnpm install --frozen-lockfile` succeeds in CI with pnpm 11.1.2
- [ ] Dependabot picks up the config (visible in repo Insights → Dependency graph → Dependabot)
- [ ] Spot-check that `actions/checkout` jobs that previously needed credentials (release tagging, auto-format push-back) still function

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -12,11 +12,10 @@ updates:
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
-      interval: "weekly"
-      day: "monday"
-      time: "09:00"
-      timezone: "Etc/UTC"
-    open-pull-requests-limit: 5
+      interval: "daily"
+    open-pull-requests-limit: 10
+    cooldown:
+      default-days: 1
     commit-message:
       prefix: "chore(ci)"
       include: "scope"
@@ -24,10 +23,19 @@ updates:
       - "dependencies"
       - "github-actions"
       - "security"
-    # Group low-risk version bumps so reviewers don't drown in tiny PRs.
-    # Major bumps stay separate so they get a real review.
     groups:
-      actions-minor-and-patch:
+      minor-and-patch:
+        patterns:
+          - "*"
         update-types:
           - "minor"
           - "patch"
+      # Workaround for dependabot/dependabot-core#14202: without an explicit
+      # major group, major updates matching the minor-and-patch pattern are
+      # silently suppressed. Remove this group when #14202 is fixed to get
+      # individual (ungrouped) PRs per major bump instead.
+      major:
+        patterns:
+          - "*"
+        update-types:
+          - "major"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,33 @@
+# Dependabot config — keeps SHA-pinned GitHub Actions current.
+#
+# Without this, every action pinned to a commit SHA in .github/workflows
+# silently rots: vulnerabilities discovered upstream don't reach us, and
+# our pinned SHAs drift further from the action's release tags every week.
+#
+# Scope: GitHub Actions only. Node/Python/pnpm dependency updates are
+# already handled by Renovate / manual upgrades and don't belong here.
+
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+      day: "monday"
+      time: "09:00"
+      timezone: "Etc/UTC"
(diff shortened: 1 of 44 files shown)


MikeRyanDev (changes requested):
Blocking publish-hardening issue: the release publish jobs still execute repository build code with npm publishing credentials present. In .github/workflows/publish-release.yml, the Publish to npm step passes NPM_TOKEN and NODE_AUTH_TOKEN into pnpm tsx scripts/release/publish-release.ts; that script then runs pnpm run build before pnpm publish. The prerelease workflow has the same pattern through scripts/release/prerelease.ts. This breaks the required build/publish separation: publish jobs should consume build artifacts without rerunning build or package lifecycle code in the same process tree as publish tokens. Please move the build/test work fully into the no-secret build job, or add a publish-only script path for the publish jobs.

PR #4835, opened 2026-05-15
fix(runtime): agent-defined headers take precedence over forwarded request headers

viktor-matic:
## Problem

`configureAgentForRequest()` merges headers from the incoming browser request **after** the agent's own headers:

```typescript
agent.headers = {
  ...agent.headers,
  ...extractForwardableHeaders(request), // ← overwrites agent headers
};
```

This silently overwrites any `Authorization` header set on the agent. A common use case is service-to-service auth where the agent is configured with an OIDC token (e.g. for Google Cloud Run), but the browser's own `Authorization` header clobbers it, causing downstream 401/403 errors.

## Fix

Reverse the spread order so agent-defined headers always win:

```typescript
agent.headers = {
  ...extractForwardableHeaders(request), // browser headers as defaults
  ...agent.headers,                       // agent headers take precedence
};
```

## Tests

Added `agent-utils.test.ts` covering:
- Agent-defined `Authorization` is preserved when the browser request also has an `Authorization` header
- Non-conflicting `x-*` headers from the request are still merged in

## Notes

The caller (application code) is still responsible for stripping infrastructure-specific headers (e.g. `x-serverless-authorization` from Cloud Run) from the request before it reaches CopilotKit — that is infrastructure knowledge CopilotKit should not need to have. This fix only addresses the merge order so that explicitly-set agent headers cannot be overwritten.

### DIFF
--- a/packages/runtime/src/v2/runtime/__tests__/agent-utils.test.ts
+++ b/packages/runtime/src/v2/runtime/__tests__/agent-utils.test.ts
@@ -1,0 +1,66 @@
+import { describe, it, expect } from "vitest";
+import { configureAgentForRequest } from "../handlers/shared/agent-utils";
+import type { AbstractAgent } from "@ag-ui/client";
+import type { CopilotRuntimeLike } from "../core/runtime";
+
+function makeRuntime(overrides: Partial<CopilotRuntimeLike> = {}): CopilotRuntimeLike {
+  return {
+    agents: {},
+    ...overrides,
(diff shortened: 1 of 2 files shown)


BenTaylorDev:
Closing as superseded — the same fix landed via #5782: `agent-utils.ts` now uses `mergeForwardableHeaders(agent.headers, request, …)` so server-set agent headers win over forwarded inbound headers, which is this PR's goal.

PR #4842, opened 2026-05-15
fix(runtime): throw typed RetryAfterExceededError for oversized Retry-After

### LINKED ISSUE #4772: retry-utils.ts — Retry-After > 60s throws instead of surfacing STOP condition
## Finding

PR #3842 added `Retry-After` parsing to `packages/runtime/src/util/retry-utils.ts` — correct for delay timing. One gap remains:

When `Retry-After` exceeds 60 seconds, the implementation throws instead of waiting. This is STOP behavior on a WAIT signal.

A quota exhaustion 429 with `Retry-After: 10800` should hold and resume. Instead it terminates. Under concurrent agent load this compounds — every worker hits the wall and throws.

## Expected behavior

- `Retry-After` values above threshold should surface the quota exhaustion condition to the caller
- Classify as STOP and surface with reset timing
- Let the caller decide whether to wait or abort

## Related pattern

`retry-after-overridden`

Corpus reference: https://github.com/SirBrenton/pitstop-truth

shiminshen:
## What does this PR do?

Closes #4772.

PR #3842 taught `fetchWithRetry` (in `packages/runtime/src/lib/runtime/retry-utils.ts`) to honor `Retry-After` on 429 responses, but capped the wait at `RETRY_CONFIG.maxRetryAfterSeconds` (60s by default). When the server requested a longer delay, the code threw a generic `Error`:

```ts
if (retryAfterMs > maxMs) {
  throw new Error(
    `Server requested Retry-After of …s which exceeds the maximum of 60s`,
  );
}
```

Callers had no good way to distinguish quota exhaustion from any other fetch failure — concurrent agent workers all hit the wall and threw indistinguishable generic errors with no programmatic access to the requested reset window.

This PR replaces the generic throw with a typed `RetryAfterExceededError` (modeled on the existing `PlatformRequestError`) that exposes the parsed `retryAfterMs` and the original 429 `Response` as `readonly` fields. The message is unchanged, so existing call sites that only catch the throw continue to work; callers that want to handle rate-limit quota exhaustion specifically can now discriminate on `instanceof RetryAfterExceededError` and decide whether to wait the full reset window or abort.

```ts
try {
  await fetchWithRetry(url, options);
} catch (err) {
  if (err instanceof RetryAfterExceededError) {
    // err.retryAfterMs, err.response are typed
    if (err.retryAfterMs < acceptableWait) {
      await sleep(err.retryAfterMs);
      // retry caller-side
    } else {
      // abort and surface to the user
    }
  }
}
```

## Related PRs and Issues

- Closes #4772
- Builds on #3842 (initial Retry-After parsing)

## Checklist

### DIFF
--- a/.changeset/retry-after-exceeded-error.md
+++ b/.changeset/retry-after-exceeded-error.md
@@ -1,0 +1,16 @@
+---
+"@copilotkit/runtime": patch
+---
+
+Surface 429 quota exhaustion as a typed `RetryAfterExceededError`
+
+When a 429 response carries a `Retry-After` value that exceeds
+`RETRY_CONFIG.maxRetryAfterSeconds` (60s by default), `fetchWithRetry` now
+throws a `RetryAfterExceededError` instead of a generic `Error`. The new
+class carries the parsed `retryAfterMs` and the original `response`, so
+callers can discriminate the rate-limit-quota-exhausted condition from
+ordinary fetch failures and decide whether to wait the full reset window
+or abort.
+
+Existing call sites that only catch the throw continue to work — the
+new class still extends `Error` and the message is unchanged.
--- a/packages/runtime/src/lib/runtime/__tests__/retry-utils.test.ts
+++ b/packages/runtime/src/lib/runtime/__tests__/retry-utils.test.ts
@@ -1,5 +1,10 @@
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
-import { fetchWithRetry, parseRetryAfter, RETRY_CONFIG } from "../retry-utils";
+import {
+  fetchWithRetry,
+  parseRetryAfter,
+  RETRY_CONFIG,
+  RetryAfterExceededError,
+} from "../retry-utils";
 
 function responseWithRetryAfter(
   headerValue: string | null,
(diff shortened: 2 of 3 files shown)


shiminshen:
Bumping this as well — happy to rebase or make any changes if there's feedback. The failing check is the Vercel fork-deploy false-red, same as my other open PR.

BenTaylorDev (changes requested):
Thanks for this — the typed error itself is clean and well-tested. Two problems, though, and the first is fundamental:

- **The code path has no consumers.** `fetchWithRetry` (and the whole `retry-utils` module) is orphaned — `git grep` across the monorepo finds it referenced *only* by its own test file; it's never wired into the runtime request path. So the PR's stated value ("callers can discriminate on `instanceof RetryAfterExceededError`") has no caller — the typed error is thrown on a branch production never executes. Until the retry utility is actually integrated, this is polishing dead code.
- **Retired-tree debris.** It adds `.changeset/retry-after-exceeded-error.md`, but `.changeset/` no longer exists on `main` (we've moved to conventional-commit-driven releases) — that file can't merge; please drop it.
- On direction: #4772 asked to "surface a STOP condition," and the issue's own suggestion was richer (a `RATE_LIMIT_EXCEEDED` `FailedResponseStatusReason` returning a structured status) — worth aligning with, though it's moot while the function is unreachable.

**Before revising, please confirm whether `fetchWithRetry` is meant to be wired into the runtime fetch path** (and show the real call site that would catch this error). If it's genuinely orphaned, this is probably better closed until the retry utility is integrated. Happy to reconsider if there's a pending integration.

shiminshen:
You're right, it's orphaned. I checked and `fetchWithRetry` only turns up in its own test file. No integration pending on my side — I picked this up from #4772 and took the file at face value without checking whether anything actually called it. My mistake.

One thing worth flagging on the issue itself: it points at `packages/runtime/src/util/retry-utils.ts`, which doesn't exist. The real file is under `src/lib/runtime/`. Whoever picks it up next will run into the same thing.

Closing this one. Happy to revisit if the retry util gets wired into the request path later.

PR #4845, opened 2026-05-15
fix(showcase-docs/prebuilt-components): Added imports and missing components in code examples 

MalaikaAbb:
## Summary
In Prebuilt Components section for all frameworks, CopilotChat, CopilotSidebar, CopilotPopup, imports are missing and custom component 'Suggestions' is not defined either. Therefore, added the sample definition for component and added the missing imports

## Changes
**CopilotChat:**
-  `chat-component.snippet.tsx` exists for all frameworks besides pydantic AI, Google ADK, and langgraph python. This component holds the code snippet used in documentation. For existing components, I simply moved the defined imports after the region `@region[chat-component]` to include it in code snippet for docs
- Added `chat-component.snippet.tsx` for all missing frameworks (pydantic AI, Google ADK, and langgraph python) in the same format for cohesion

**CopilotSidebar:**
- Added region `@region[sidebar-imports]` in all `prebuilt-sidebar/page.tsx` files and used the snippet in `sidebar.mdx`
- Added region  @region[suggestions-component-example] to show Suggestions component on docs and used it in `sidebar.mdx`

**CopilotPopup:**
- Added region `@region[popup-imports] `in all prebuilt-popup/page.tsx to use codfe snippet in  `sidebar.mdx`


samjulien:
Hi @MalaikaAbb — thanks for the careful QA pass on the prebuilt-components docs.

A few things to flag before closing this one:

1. **The "missing imports" issue you found has been fixed at the region-marker layer.** #4829 moved every demo cell's `@region[name]` marker so the region wraps imports + body as **one contiguous block**. The shell-docs pages (`/prebuilt-components/{chat,sidebar,popup}`) render those regions via `<Snippet>`, so after #4829 merged, the snippet bodies already include the imports — verify on `/{framework}/prebuilt-components/chat`, the `CopilotChat` import is visible in the snippet body.

2. **The pattern in this PR splits markers** — e.g. `@region[popup-imports]` wraps only the imports, then a separate marker is expected to wrap the body. We deliberately moved *away* from split markers in #4829: one MDX `<Snippet>` call should resolve to one contiguous, runnable code block. The right fix when imports are missing is to expand the existing `@region[<name>]` upward to cover imports, not to introduce a sibling import-only region.

3. **The `Suggestions` custom component definition** — this should be solved at the demo level rather than the snippet level. If a demo references a component without a visible definition, the demo file itself should expose it (with a region marker if it should appear in docs).

Closing this one — thanks for the QA work.

PR #4884, opened 2026-05-17
fix(web-inspector): unblock tests on Node 22.4+

ataibarkai:
## Summary

\`packages/web-inspector\` tests fail on Node 22.4+ with:
\`\`\`
TypeError: window.localStorage.clear is not a function
\`\`\`

Node 22.4 shipped an experimental \`Storage\` global that gets installed BEFORE vitest's jsdom env runs, leaving \`window.localStorage\` as a stub without \`.clear()\`. On Node ≤22.3, jsdom owns localStorage and tests pass — which is why this looked like a phantom issue when previously fixed by a polyfill (commit \`63d1d9e0\`) and then reverted in #4878.

This PR avoids the polyfill entirely by passing \`--no-experimental-webstorage\` to Node in the test script. That disables Node's experimental Storage API and lets jsdom own localStorage cleanly. Long comment in \`vitest.config.ts\` records the why so it doesn't get reverted again.

## Verification

- \`pnpm run test\` (full monorepo): **18/18 packages, 1170 + 29 tests pass** on Node 25.9.0.
- Without this fix: \`@copilotkit/web-inspector\` has 22/29 tests failing.

### DIFF
--- a/packages/web-inspector/vitest.config.ts
+++ b/packages/web-inspector/vitest.config.ts
@@ -1,5 +1,55 @@
 import { defineConfig } from "vitest/config";
 
+// Node 25.0 unflagged the experimental Web Storage API (nodejs/node#57658),
+// so on Node 25+ a stub `window.localStorage` is installed BEFORE vitest's
+// jsdom environment runs. vitest's jsdom env does NOT replace it, so the
+// stub stays in place — and it has no `.clear()` / `.getItem()` /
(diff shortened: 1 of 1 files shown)


BenTaylorDev:
Closing as already-landed — `packages/web-inspector/vitest.config.ts` on main (commit e8104fa2d2) already carries the version-gated `--no-experimental-webstorage` block this proposes.

PR #4886, opened 2026-05-17
fix(showcase/a2ui-dynamic): drop hardcoded catalog names from SYSTEM_PROMPTs

ataibarkai:
## Summary

Every dynamic-schema A2UI showcase agent had a chat `SYSTEM_PROMPT` that hardcoded `Card`, `StatusBadge`, `Metric`, `InfoRow`, `PrimaryButton`, `PieChart`, `BarChart` and steered toward specific chart shapes ("pie for part-of-whole, bar for comparisons"). That biased the primary LLM toward those exact names regardless of what the frontend's A2UI schema context actually advertises — renaming a catalog component on the frontend would silently degrade the agent.

Each chat `SYSTEM_PROMPT` now:
- tells the model to call `generate_a2ui` when a visual would help,
- names no specific components or chart shapes,
- points at the A2UI schema context attached to the conversation as the source of truth for component names + shapes.

Fixed-schema agents are untouched (the whole point of the fixed flow is that the catalog is known up front).

## Integrations covered

- `langgraph-python`
- `langgraph-typescript`
- `langgraph-fastapi`
- `ag2`
- `agno`
- `claude-sdk-python`
- `llamaindex`
- `ms-agent-python`
- `pydantic-ai`

## Test plan

- [ ] Run each dynamic-A2UI showcase and confirm dashboards still render with the existing React catalog components
- [ ] Rename a component in the React catalog (e.g. `PieChart` → `Donut`) and confirm the agent produces a surface using the new name without any agent-side change


BenTaylorDev:
Closing this one — the design decision went the other way on `main` since May.

**What changed.** `4de75ff900` (OSS-136) reworked the declarative-gen-ui demo into the Vantage Threads sales analyst. The header comment on `showcase/integrations/langgraph-python/src/agents/a2ui_dynamic.py:26-31` now makes the component-name set an explicit parity contract with the ADK sibling: "the structural rules and the component name set must match the catalog above." ADK's `_INSTRUCTION` names the components directly (`showcase/integrations/google-adk/src/agents/declarative_gen_ui_agent.py:42-44`). The `strands` prompt went further and pinned them hard: "Use ONLY these exact component names ... named exactly `Metric` (NOT `MetricTile` or `MetricCard`)" (`showcase/integrations/strands/src/agents/a2ui_dynamic.py:61`). That line exists because the model invents names the catalog does not register, which is the failure this PR's change would bring back.

**One factual issue with the replacement text.** The new prompt tells the primary model to rely on "the A2UI schema context attached to this conversation." That context does not reach the primary model on this path. The catalog and `compositionGuide` are merged into the secondary A2UI subagent's params (`packages/sdk-js/src/langgraph/middleware.ts:497-511`), and the frontend forwards only a boolean `a2uiCatalogAvailable` (`packages/react-core/src/v2/providers/CopilotKitProvider.tsx:809-813`). The primary model sees a zero-argument `generate_a2ui` tool and nothing more. The original comment in the diff had this right — the secondary LLM is the only place catalog names matter — but the prompt itself points the primary model at something it never receives.

Also worth noting: the D5 declarative probe drives both LLM calls through aimock fixtures (`showcase/harness/fixtures/d5/gen-ui-declarative.json`), so prompt quality is invisible to CI either way. Green here would not have told us anything.

**The real finding underneath.** The eight integrations listed in this PR still run the pre-OSS-136 prompt verbatim, while `langgraph-python` and `google-adk` run the sales-analyst prompt. That is a genuine showcase parity gap, and I will file it separately. The fix is to raise those eight to the sales-analyst prompt, not to remove catalog names from them.

Branch is 13,282 commits behind with 6 of 9 files conflicting, so there is nothing to rebase toward.

PR #4899, opened 2026-05-18
fix(showcase/built-in-agent): repair reasoning demos — Responses API param shape + reasoning-streaming model

samjulien:
## Summary

All three built-in-agent reasoning demos (`agentic-chat-reasoning`, `reasoning-default-render`, `tool-rendering-reasoning-chain`) currently error 400 on every prompt and surface no reasoning chain. Two independent misconfigurations in `reasoning-factory.ts` stack to produce the failure:

- OpenAI's Responses API rejects the flat `reasoning_effort` parameter with: *"Unsupported parameter: 'reasoning_effort'. In the Responses API, this parameter has moved to 'reasoning.effort'."* Every run aborts before producing any output.
- Even after fixing the parameter shape, `gpt-5.2` does not stream `response.reasoning_summary_text.delta` events with `reasoning.summary: "auto"` or `"detailed"`, so reasoning happens internally but never reaches the frontend. `o4-mini` (and other o-series models) do stream summary deltas, which the tanstack converter translates to AG-UI `REASONING_MESSAGE_CONTENT` events.

This PR:

- Switches the default `REASONING_MODEL` from `gpt-5.2` to `o4-mini`.
- Passes the parameter as `reasoning: { effort: "medium", summary: "auto" }`.
- Updates the JSDoc to describe why o-series models are required for a "visible reasoning" demo.

## Reproduction (before the fix)

Direct SSE probe against the deployed showcase:

```
curl -sN -XPOST "https://showcase-built-in-agent-production.up.railway.app/api/copilotkit-reasoning" \
  -H "Content-Type: application/json" -H "Accept: text/event-stream" \
  -d '{"method":"agent/run","params":{"agentId":"agentic-chat-reasoning"},"body":{"threadId":"t","runId":"r","messages":[{"id":"m1","role":"user","content":"hi"}],"state":{},"tools":[],"context":[],"forwardedProps":{}}}'
```

Returns `RUN_STARTED` → `RUN_ERROR` ("Unsupported parameter: 'reasoning_effort'"). Zero text or reasoning content emitted.

## Verification (after the fix)

Same probe against the patched factory emits the full envelope:

`RUN_STARTED` → `REASONING_START` → `REASONING_MESSAGE_START` → ~237× `REASONING_MESSAGE_CONTENT` → `REASONING_MESSAGE_END` → `REASONING_END` → `TEXT_MESSAGE_START` → ~119× `TEXT_MESSAGE_CONTENT` → `TEXT_MESSAGE_END` → `RUN_FINISHED`.

## Scope note

This PR fixes the factory-level misconfiguration that is the most visible cause of the broken demos. A separate runtime-level schema-conversion bug (AG-UI state-tool injection produces an OpenAI-strict-mode schema with `type: "None"`) also surfaces on these routes once the factory is unblocked; that is being tracked as a separate Linear ticket for OSS triage.

## Test plan

- [ ] Pull the branch, set `OPENAI_API_KEY`, run the built-in-agent integration, open `/demos/agentic-chat-reasoning`.
- [ ] Send a multi-step prompt (e.g. "What is 17 * 23? Walk through it.").
- [ ] Confirm the "Thinking…" card appears, streams content, and collapses to "Thought for X seconds" once the answer completes.
- [ ] Repeat for `/demos/reasoning-default-render` and `/demos/tool-rendering-reasoning-chain`.

### DIFF
--- a/showcase/integrations/built-in-agent/src/app/demos/agentic-chat-reasoning/page.tsx
+++ b/showcase/integrations/built-in-agent/src/app/demos/agentic-chat-reasoning/page.tsx
@@ -5,7 +5,7 @@
 // The built-in tanstack/openai factory normally uses a non-reasoning
 // model (gpt-4o) so REASONING_* events never flow. This demo points at
 // a dedicated route (`/api/copilotkit-reasoning`) whose factory uses a
-// reasoning-capable model (`gpt-5.2`) with `reasoning_effort: "low"`.
+// reasoning-capable model (`gpt-5.2`) with `reasoning.effort: "xhigh"`.
 // The runtime's tanstack converter translates the upstream reasoning
 // events into AG-UI REASONING_START / REASONING_MESSAGE_CONTENT /
 // REASONING_END, and CopilotKit renders them via the
--- a/showcase/integrations/built-in-agent/src/app/demos/reasoning-default-render/page.tsx
+++ b/showcase/integrations/built-in-agent/src/app/demos/reasoning-default-render/page.tsx
@@ -3,7 +3,7 @@
 // Reasoning (Default Render) — built-in-agent variant.
 //
 // Same backend (`/api/copilotkit-reasoning`, agent
-// `reasoning-default-render`, gpt-5.2 with reasoning_effort=low) as the
+// `reasoning-default-render`, gpt-5.2 with reasoning.effort=xhigh) as the
 // agentic-chat-reasoning demo, but this page passes NO custom
 // `reasoningMessage` slot. CopilotKit's built-in
 // `CopilotChatReasoningMessage` renders the chain as a collapsible card
--- a/showcase/integrations/built-in-agent/src/app/demos/tool-rendering-reasoning-chain/page.tsx
+++ b/showcase/integrations/built-in-agent/src/app/demos/tool-rendering-reasoning-chain/page.tsx
@@ -12,7 +12,7 @@
 //
 // Backed by the shared reasoning runtime
 // (`/api/copilotkit-reasoning`, agent `tool-rendering-reasoning-chain`)
-// using gpt-5.2 with `reasoning_effort: "low"`.
+// using gpt-5.2 with `reasoning.effort: "xhigh"`.
 
 import React from "react";
 import {
--- a/showcase/integrations/built-in-agent/src/lib/factory/reasoning-factory.ts
+++ b/showcase/integrations/built-in-agent/src/lib/factory/reasoning-factory.ts
@@ -6,14 +6,27 @@ import { baseServerTools } from "./server-tools";
 /**
  * Reasoning model used by all three reasoning demos.
  *
- * GPT-5.2 is a reasoning-capable variant; OpenAI's chat completions API
- * accepts the `reasoning_effort` parameter for it. If GPT-5.2 isn't
- * accessible in your environment swap to another reasoning-capable model
- * such as `o3` or `gpt-5-thinking`. The runtime's tanstack converter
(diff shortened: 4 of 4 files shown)


samjulien:
Direct Responses-API probe today (2026-05-19) to verify whether the model swap was necessary.

**Probe matrix:** same prompt ("What is 17 times 23? Walk through the calculation step by step, then state the final answer."), `stream: true`, against `api.openai.com/v1/responses`.

| Model | reasoning config | reasoning_summary_text.delta | reasoning_text.delta | reasoning_tokens (final) |
| --- | --- | --- | --- | --- |
| gpt-5.2 | effort=low, summary=auto | 0 | 0 | 0 |
| gpt-5.2 | effort=medium, summary=detailed | 0 | 0 | 0 |
| gpt-5.2 | effort=high, summary=auto | 0 | 0 | 0 |
| o4-mini | effort=medium, summary=auto | 166 | 0 | 320 |

gpt-5.2 emitted zero reasoning events in every configuration tested and reported `reasoning_tokens: 0` in the final response, so the model swap is necessary for a "visible reasoning chain" showcase demo.

Pushed a follow-up commit (08dbef98c) addressing the two concedable points:

1. Reverted `reasoning.effort` from `medium` back to `low` (matches the original factory; the model swap alone restores visibility; the silent bump added cost without benefit).
2. Updated the three demo `page.tsx` comments that still referenced `gpt-5.2` and `reasoning_effort: "low"` so they reflect the actual `o4-mini` + nested `reasoning.effort` wiring.
3. Tightened the factory JSDoc to cite the empirical probe results.

AlemTuzlak (dismissed):
I'm 99% sure this is a hallucination on the model part, the fix should be the config in modelOptions that you passed in

samjulien:
Re-probe after Alem's review. Empirical re-investigation today (2026-05-20) covering the configurations the first pass missed:

| Probe | reasoning_summary_text.delta | reasoning_tokens |
| --- | --- | --- |
| gpt-5.2 + effort=medium + summary=auto + include=reasoning.encrypted_content | 0 | 0 |
| gpt-5.2 + effort=high + summary omitted | 0 | 0 |
| gpt-5.2 + effort=medium + summary=auto + store=true | 0 | 0 |
| gpt-5.2 + effort=high + summary=auto on a Bayes-theorem prompt | 0 | 0 |
| **gpt-5.2 + effort=xhigh + summary=detailed** | **79** | **57** |
| gpt-5-pro + effort=high + summary=auto (control) | 167 | 384 |

So Alem was right that the model swap is unnecessary. The minimal fix is to keep gpt-5.2 and bump effort to `xhigh`.

Pushed `f08d5f94a` to do exactly that:

- `REASONING_MODEL` reverted to `gpt-5.2` (was `o4-mini`).
- `modelOptions.reasoning.effort` bumped from `low` to `xhigh`.
- JSDoc rewritten to reflect the new empirical picture and note the `xhigh` documentation caveat.
- Demo `page.tsx` comments updated to match.

Trade-off acknowledged in the JSDoc: gpt-5.2 + xhigh produces a shorter visible "Thinking..." card than o4-mini + medium would (57 vs 320 reasoning tokens on the same prompt), but staying on a current-gen model wins on brittleness grounds. The `REASONING_MODEL` env var override remains the escape hatch if `xhigh` support on gpt-5.2 ever regresses.

Worth flagging: `xhigh` is documented as "supported for all models after gpt-5.1-codex-max" so its gpt-5.2 support is empirically verified but not formally guaranteed by OpenAI docs. Worth keeping an eye on.

samjulien:
Closing this, doesn’t seem to be the right fix

PR #4947, opened 2026-05-21

### LINKED ISSUE #4935: 🐛 Bug: useFrontendTool render `args` uses LLM-generated tool call args instead of AG-UI TOOL_CALL_ARGS event data
### ♻️ Reproduction Steps

1. Set up an AG-UI backend agent (using `ag-ui-strands`) with an `args_streamer` configured for a frontend tool. The `args_streamer` enriches the tool call args with server-side data (e.g., address lookup results).

2. Define a `useFrontendTool` on the frontend with a parameter (e.g., `addresses`) that the `args_streamer` populates server-side.

3. When the LLM calls the tool, it generates its own (hallucinated) values for the `addresses` parameter since it's required in the schema.

4. The `args_streamer` correctly overrides these with real data — confirmed by:
   - The `TOOL_CALL_ARGS` event in the CopilotKit dev console showing correct data
   - The `TOOL_CALL_END` event's `toolCallArgs` containing the correct server-enriched values

5. However, the `render` function's `args` parameter contains the **LLM-generated (hallucinated) values**, not the streamed values from `TOOL_CALL_ARGS`.

**Backend (ag-ui-strands with args_streamer):**

```python
async def address_args_streamer(ctx: ToolCallContext):
    args = json.loads(ctx.args_str)
    postcode = args.get("postcode", "")
    addresses = get_addresses_for_postcode(postcode)  # real server-side lookup
    yield json.dumps({"postcode": postcode, "addresses": addresses})
```

**Frontend (useFrontendTool):**

```tsx
useFrontendTool({
  name: 'select_address_for_postcode',
  parameters: z.object({
    postcode: z.string(),
    addresses: z.array(z.object({
      id: z.string(),
      line1: z.string(),
      line2: z.string(),
      city: z.string(),
      postcode: z.string(),
    })),
  }),
  handler: async () => { /* ... */ },
  render: ({ args, status }) => {
    // args.addresses contains LLM hallucinated data, NOT the args_streamer data
    console.log(args)
  },
})
```

### ✅ Expected Behavior

The `render` function's `args` should reflect the data from the `TOOL_CALL_ARGS` and `TOOL_CALL_END` AG-UI protocol events — i.e., the server-enriched args from the `args_streamer`.

The AG-UI protocol's `TOOL_CALL_ARGS` event is the authoritative source for tool call arguments. When an `args_streamer` overrides the LLM-generated args, the render function should use the streamed values.

### ❌ Actual Behavior

The `render` function's `args` contains the LLM-generated tool call arguments from the `MESSAGES_SNAPSHOT`, ignoring the `TOOL_CALL_ARGS` event data entirely.

Evidence from the AG-UI event stream:

**TOOL_CALL_ARGS event (correct, from args_streamer):**

```json
{
  "type": "TOOL_CALL_ARGS",
  "toolCallId": "e2404443-...",
  "delta": "{\"postcode\": \"M1 1AA\", \"addresses\": [{\"id\": \"6\", \"line1\": \"1 Piccadilly\", \"line2\": \"City Centre\", \"city\": \"Manchester\", \"postcode\": \"M1 1AA\"}, {\"id\": \"7\", \"line1\": \"15 Portland Street\", \"line2\": \"City Centre\", \"city\": \"Manchester\", \"postcode\": \"M1 1AA\"}, {\"id\": \"8\", \"line1\": \"42 Deansgate\", \"line2\": \"City Centre\", \"city\": \"Manchester\", \"postcode\": \"M1 1AA\"}, {\"id\": \"9\", \"line1\": \"8 Albert Square\", \"line2\": \"City Centre\", \"city\": \"Manchester\", \"postcode\": \"M1 1AA\"}]}"
}
```

**TOOL_CALL_END (correct):**

```json
{
  "toolCallArgs": {
    "postcode": "M1 1AA",
    "addresses": [
      {"id": "6", "line1": "1 Piccadilly", "line2": "City Centre", "city": "Manchester", "postcode": "M1 1AA"},
      {"id": "7", "line1": "15 Portland Street", "line2": "City Centre", "city": "Manchester", "postcode": "M1 1AA"},
      {"id": "8", "line1": "42 Deansgate", "line2": "City Centre", "city": "Manchester", "postcode": "M1 1AA"},
      {"id": "9", "line1": "8 Albert Square", "line2": "City Centre", "city": "Manchester", "postcode": "M1 1AA"}
    ]
  }
}
```

**But render receives (LLM hallucinated):**

```json
{"postcode": "M1 1AA", "addresses": [{"id": "1", "line1": "1 Piccadilly Gardens", "line2": "", "city": "Manchester", "postcode": "M1 1AA"}, {"id": "2", "line1": "2 Piccadilly Gardens", "line2": "", "city": "Manchester", "postcode": "M1 1AA"}, {"id": "3", "line1": "3 Piccadilly Gardens", "line2": "Apartment 5", "city": "Manchester", "postcode": "M1 1AA"}]}
```

The render `args` is sourced from the tool call within the `MESSAGES_SNAPSHOT` (raw LLM output before args_streamer override), rather than from the `TOOL_CALL_ARGS`/`TOOL_CALL_END` events.

Console log confirms args never includes streamed addresses:

```
RENDER args: {"postcode":"SW1A 1AA"} status: inProgress
RENDER args: {"postcode":"SW1A 1AA"} status: executing

### Screenshots

<img width="1480" height="706" alt="Image" src="https://github.com/user-attachments/assets/aa7053dc-c0fa-4f88-9d17-affa44c21203" />

### 𝌚 CopilotKit Version

```shell
@copilotkit/react-core@1.57.1
@copilotkit/react-ui@1.57.1
@copilotkit/runtime@1.57.1

Using: useFrontendTool from '@copilotkit/react-core/v2'
AG-UI backend: ag-ui-strands 0.1.9, ag-ui-protocol 0.1.18
```

### 📄 Logs (Optional)

```shell
RENDER args: {"postcode":"M1 1AA","addresses":[{"id":"1","line1":"1 Piccadilly Gardens","line2":"","city":"Manchester","postcode":"M1 1AA"},{"id":"2","line1":"2 Piccadilly Gardens","line2":"","city":"Manchester","postcode":"M1 1AA"},{"id":"3","line1":"3 Piccadilly Gardens","line2":"Apartment 5","city":"Manchester","postcode":"M1 1AA"}]} status: inProgress
```

Tushar-Khandelwal-2004:
## What does this PR do?

Fixes #4935.

Updates CopilotKit core message state so streamed `TOOL_CALL_ARGS` are treated as the authoritative tool-call arguments for frontend rendering. This prevents `useFrontendTool` from rendering with hallucinated args from `MESSAGES_SNAPSHOT` when the backend later streams enriched tool-call args.

### Changes

-> Track streamed tool-call args in `StateManager` by agent/thread/toolCallId.
-> Return updated AG-UI message mutations from `onToolCallArgsEvent` and `onToolCallEndEvent`.
-> Preserve accumulated streamed args independently from existing snapshot args.
-> Add regression tests covering:
  -> replacement of hallucinated snapshot args
  -> multi-delta streamed args accumulation

## Related PRs and Issues

-> Fixes #4935

## Testing

->  `pnpm nx run @copilotkit/core:test -- --run src/__tests__/state-manager.test.ts`
->  `pnpm nx run @copilotkit/core:build`

Note: local monorepo-wide tests currently fail due to unrelated Vitest timeout issues in other packages.

### DIFF
--- a/packages/core/src/__tests__/state-manager.test.ts
+++ b/packages/core/src/__tests__/state-manager.test.ts
@@ -1,12 +1,7 @@
-import { describe, it, expect, beforeEach, vi } from "vitest";
+import { describe, it, expect, beforeEach } from "vitest";
 import { CopilotKitCore } from "../core";
-import {
-  AbstractAgent,
-  Message,
-  State,
-  RunAgentInput,
-  EventType,
-} from "@ag-ui/client";
+import type { Message, State, RunAgentInput } from "@ag-ui/client";
+import { AbstractAgent, EventType } from "@ag-ui/client";
 import { randomUUID } from "@copilotkit/shared";
 
 /**
@@ -23,7 +18,7 @@ class EventEmittingMockAgent extends AbstractAgent {
     });
   }
 
-  protected run(input: RunAgentInput): any {
+  protected run(_input: RunAgentInput): any {
     // Not used in these tests
     throw new Error("run() should not be called in these tests");
   }
@@ -175,6 +170,56 @@ class EventEmittingMockAgent extends AbstractAgent {
     }
   }
 
+  // Helper to emit tool call args event
+  public async emitToolCallArgs(
+    runId: string,
+    toolCallId: string,
+    delta: string,
+  ) {
+    const toolCall = this.findToolCall(toolCallId);
+    for (const sub of this.subscribers) {
+      if (sub.onToolCallArgsEvent) {
+        const mutation = await sub.onToolCallArgsEvent({
+          event: {
+            type: EventType.TOOL_CALL_ARGS,
+            toolCallId,
+            delta,
+          },
+          toolCallBuffer: toolCall?.function.arguments ?? "",
+          toolCallName: toolCall?.function.name ?? "",
+          partialToolCallArgs: {},
+          messages: this.messages,
+          state: this.state,
+          agent: this,
+          input: this.createRunInput(runId),
+        });
+        this.applyMutation(mutation);
+      }
+    }
+  }
+
+  // Helper to emit tool call end event
+  public async emitToolCallEnd(runId: string, toolCallId: string) {
+    const toolCall = this.findToolCall(toolCallId);
+    for (const sub of this.subscribers) {
+      if (sub.onToolCallEndEvent) {
+        const mutation = await sub.onToolCallEndEvent({
+          event: {
+            type: EventType.TOOL_CALL_END,
+            toolCallId,
+          },
+          toolCallName: toolCall?.function.name ?? "",
+          toolCallArgs: this.parseToolCallArgs(toolCall?.function.arguments),
+          messages: this.messages,
+          state: this.state,
+          agent: this,
+          input: this.createRunInput(runId),
+        });
+        this.applyMutation(mutation);
+      }
+    }
+  }
+
   // Override subscribe to track subscribers
   public override subscribe(subscriber: any) {
     this.subscribers.push(subscriber);
@@ -195,6 +240,38 @@ class EventEmittingMockAgent extends AbstractAgent {
       state: this.state,
       messages: this.messages,
     };
+  }
+
+  private findToolCall(toolCallId: string) {
+    for (const message of this.messages) {
+      if (message.role !== "assistant") continue;
+      const toolCall = message.toolCalls?.find((candidateToolCall) => {
+        return candidateToolCall.id === toolCallId;
+      });
+      if (toolCall) return toolCall;
+    }
+  }
+
+  public getToolCallArguments(toolCallId: string) {
+    return this.findToolCall(toolCallId)?.function.arguments;
+  }
+
+  private parseToolCallArgs(args?: string) {
+    if (!args) return {};
+    try {
+      return JSON.parse(args);
+    } catch {
+      return {};
+    }
+  }
+
+  private applyMutation(mutation: any) {
+    if (mutation?.messages) {
+      this.messages = mutation.messages;
+    }
+    if (mutation?.state) {
+      this.state = mutation.state;
+    }
   }
 }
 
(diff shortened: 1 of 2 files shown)


BenTaylorDev:
Thanks @Tushar-Khandelwal-2004 — the gap is real (state-manager doesn't sync tool-call args on main) and the mechanism (returning a mutation from `onToolCallArgsEvent` reaches renderers) is sound. One correctness concern before this can land:

`handleToolCallEnd` deletes the cached streamed args (`toolCallArgsById.delete(...)`), and this PR removes the `messages` param from `handleMessagesSnapshot` — so a `MESSAGES_SNAPSHOT` arriving **after** `TOOL_CALL_END` would re-stamp the hallucinated args with nothing left to re-apply. Issue #4935 specifically says the bad args come from the tool call *within the snapshot*, so event ordering is load-bearing and that path isn't covered. The tests also drive a hand-rolled `EventEmittingMockAgent` that simulates AG-UI's mutation-apply and only assert `agent.messages` — they don't exercise the real `useFrontendTool` render path the bug is about.

Could you confirm the actual ag-ui-strands event order (whether a snapshot lands after `TOOL_CALL_END`) and add a test that hits a real render, not the simulated mutation?

BenTaylorDev (changes requested):
Thanks for this. The bug it targets is a real class of problem, but this can't merge as authored and the mechanism needs rethinking:

- **Unmergeable / targets rewritten code.** This branch is stale and `git merge-tree` reports a hard conflict in both `state-manager.ts` and its test. Main rewrote the exact region you edit — `subscribe()` gained run-isolation machinery (`subRunId`/revocation/`runFinished`), so the callback edits and the `handleMessagesSnapshot` signature change collide head-on. It needs re-authoring against current `main`.
- **Mechanism doesn't match the stated bug.** The PR frames the issue as a `MESSAGES_SNAPSHOT` overwriting enriched streamed args — but it drops the `messages` param from `handleMessagesSnapshot` and does nothing there; it only re-applies stored args on `TOOL_CALL_ARGS`/`TOOL_CALL_END`. A snapshot arriving *after* `TOOL_CALL_END` would still win, so the fix doesn't demonstrably protect the ordering it claims to. Also worth confirming first: `@ag-ui/client` already accumulates `TOOL_CALL_ARGS` deltas into `message.toolCalls[].function.arguments`, which is exactly what the render site reads (`use-render-tool-call.tsx:38-40`) — so on a normal stream this re-application is redundant.
- Minor: `toolCallArgsById` entries are only deleted on `TOOL_CALL_END`, so a run that errors between ARGS and END leaks entries; and there's an `import` → `import type` reflow bundled in.

**Before re-authoring:** please reproduce #4935 against current `main` and confirm the render site actually shows snapshot/hallucinated args despite the client's default delta accumulation. If it reproduces, the real defect is a snapshot-overwrite ordering issue that must be handled in `onMessagesSnapshotEvent` (which this PR stopped touching), not at `TOOL_CALL_END`.

BenTaylorDev:
Closing this branch — the underlying gap is real and #4935 stays open, so nothing here is being dismissed.

Where it stands: the two questions from my review are still unanswered, and both are load-bearing rather than cosmetic. `handleToolCallEnd` deletes the cached streamed args, and the PR stops passing `messages` into `handleMessagesSnapshot`, so a `MESSAGES_SNAPSHOT` arriving *after* `TOOL_CALL_END` would re-stamp the hallucinated args with nothing left to re-apply — and #4935 describes the bad args as coming from the tool call *inside* the snapshot, which makes event ordering decisive. The tests also drive a hand-rolled `EventEmittingMockAgent` that simulates AG-UI's mutation-apply and assert only `agent.messages`, so they don't exercise the `useFrontendTool` render path the bug is actually about.

Meanwhile `main` has moved: `state-manager.ts` gained run-isolation machinery in the exact region this edits, and the branch now conflicts in both `state-manager.ts` and its test.

To be clear about what I'm *not* saying: the approach of returning a mutation from `onToolCallArgsEvent` is sound, and `main` still has no tool-call-args sync in `state-manager` — so this is worth doing. If you'd like to pick it up again, a fresh branch off current `main` with the snapshot-ordering path covered and one test through a real render would be very welcome, and I'll review it promptly. Thanks for the work so far.

PR #4948, opened 2026-05-21

deepshekhardas:
Applying changes from upstream PR #4947 by @Tushar-Khandelwal-2004


BenTaylorDev:
Closing as a duplicate of #4947, which is the canonical version of this fix (identical diff, and it links the issue it resolves). Let's keep the discussion there.

Issue #4953, opened 2026-05-21
🐛 Bug: `useHumanInTheLoop` silently fails under the default multi-route transport (CopilotKitProvider)


MalaikaAbb:
### ♻️ Reproduction Steps
issue is that A `useHumanInTheLoop` tool call is abandoned when the app uses the default multi-route REST/SSE transport (i.e. `<CopilotKitProvider>` without `useSingleEndpoint`, runtime built with `createCopilotEndpoint`). The tool call streams in, `RUN_FINISHED` fires almost immediately, the HITL UI appears for a moment and then disappears, and the run never waits for `respond()`. No error is shown.

### Environment

- `<CopilotKitProvider>` **without** `useSingleEndpoint`
- Runtime endpoint built with the multi-route handler (`createCopilotEndpoint`)
- A `useHumanInTheLoop` tool that the agent invokes

### Steps to reproduce

1. `<CopilotKitProvider runtimeUrl="…">` with no `useSingleEndpoint`.
2. Register a `useHumanInTheLoop` tool with a render component exposing a `respond` button.
3. Ask the agent to do something that triggers the tool.
4. Observe the AG-UI events and the chat.

### ✅ Expected Behavior

**Expected:** the run pauses at the HITL tool, the render UI stays on screen, and the run resumes only after `respond()` is called.

### ❌ Actual Behavior

**Actual:** events show `TOOL_CALL_START → TOOL_CALL_ARGS → TOOL_CALL_END → RUN_FINISHED` within ~70 ms. The HITL UI flashes briefly and vanishes; `respond()` is never awaited; the chat clears.

<img width="2444" height="562" alt="Image" src="https://github.com/user-attachments/assets/329ca501-e197-4825-b501-be2140b0ac16" />

The tool I used is below

```
useHumanInTheLoop({
    agentId: "my_agent",
    name: "book_call",
    description:
      "Ask the user to pick a time slot for a call. The picker UI presents fixed candidate slots; the user's choice is returned to the agent.",
    parameters: z.object({
      topic: z
        .string()
        .describe("What the call is about (e.g. 'Intro with sales')"),
      attendee: z
        .string()
        .describe("Who the call is with (e.g. 'Alice from Sales')"),
    }),
    render: ({ args, status, respond }: any) => (
      <TimePickerCard
        topic={args?.topic ?? "a call"}
        attendee={args?.attendee}
        slots={DEFAULT_SLOTS}
        status={status}
[...]

Issue #4952, opened 2026-05-21
🐛 Bug:  useFrontendTool tools silently dropped when the runtime has `openGenerativeUI` enabled


MalaikaAbb:
### ♻️ Reproduction Steps

When a user registers frontend tools using `useFrontendTool` / `useHumanInTheLoop` tools silently dropped when the runtime has `openGenerativeUI` enabled but the provider is not given the `openGenerativeUI` prop. Here are the steps to reproduce it. 

1. Configure the runtime with `openGenerativeUI: true`.
2. Render `<CopilotKitProvider runtimeUrl="…">` **without** an `openGenerativeUI` prop. (Note: The issue persists with `<CopilotKit runtimeUrl="...">` as well, so you can use either)
3. In a child component, register a tool with `useFrontendTool({ name: "myTool", … })`
4. Send a chat message.
5. Inspect the `RUN_STARTED` event (or the dev console "Frontend Tools" tab).

### ✅ Expected Behavior

In RUN_STARTED, the frontend tools registered using `useFrontendTool` / `useHumanInTheLoop` should be visible,  and propogated to agent side through AG-UI. 

<img width="1971" height="582" alt="Image" src="https://github.com/user-attachments/assets/ebdb91e4-484e-45a4-81aa-88f9ed928100" />

### ❌ Actual Behavior

In RUN_STARTED, the frontend tools registered using `useFrontendTool` / `useHumanInTheLoop` are **NOT** visible,  and cannot be seen. Hence, they are not registered on agent side and we cannot use them 
<img width="2239" height="995" alt="Image" src="https://github.com/user-attachments/assets/46efd7c1-2401-40da-8081-fb05e017c7f5" />

### Screenshots

When we use <CopilotKit runtimeUrl='...'>

where, my runtime is configured like this
```
const runtime = new CopilotRuntime({
  agents: {
    my_agent: new HttpAgent({ url: "http://localhost:8000/" }),
  },
  openGenerativeUI: true
});
```
and my CopilotKit Wrapper is like this
```
<CopilotKit runtimeUrl="/api/copilotkit" agent="my_agent">
          {children}
</CopilotKit>
```
I have the following frontend/HITL tools in page.tsx

```
useFrontendTool({
    name: "sayHello",
    description: "Say hello to the user",
    parameters: z.object({
      name: z.string().describe("The name of the user to say hello to"),
    }),
    handler: async ({ name }) => {
      alert(`Hello, ${name}!`);
[...]

Issue #3154, opened 2026-02-05
🚀 Feature Request: Allow users to chat with the agent while a `useFrontendTool` UI is pending


bajayo:
### Summary

Currently, when `useFrontendTool` renders an interactive UI component (e.g., a form, questionnaire, or confirmation dialog), the chat is effectively "locked" - users cannot send text messages to the agent until they complete the tool interaction.

**We'd like to request support for a more flexible interaction model** where users can have a conversation with the agent while a tool UI is displayed, and still be able to complete the tool interaction afterward.

### Use Case

Consider this scenario:

1. Agent calls a tool that renders a question: _"What is your pet's name?"_
2. User sees the form but wants to ask: _"Can I add multiple pets later?"_
3. User types this question in the chat
4. Agent responds: _"Yes, you can add more pets after completing the initial setup."_
5. User, now satisfied, clicks the Submit button on the original form with their pet's name

**Current behavior**: Step 3-5 results in `CopilotKitError: Run ended without emitting a terminal event`

**Desired behavior**: The conversation flows naturally, and the tool submission still works.

### Why This Matters

The current "locked" behavior forces a rigid interaction pattern that doesn't match how users naturally interact with AI assistants:

- **Users ask clarifying questions** - Before filling out a form, users often want to understand what they're being asked
- **Users change their mind** - They might want to ask about alternatives before committing to a choice
- **Users multitask** - In longer flows, users might have unrelated questions mid-way
- **Users make mistakes** - If they accidentally send a message, the entire tool state is lost

This is especially problematic for:

- Multi-step form wizards
- Insurance/financial quote flows
- Human-in-the-loop approval processes
- Any interactive UI that requires user decisions

### Current Workaround

We implemented a pattern that bypasses the default `useFrontendTool` behavior:

1. Set `followUp: false` to prevent CopilotKit from managing the tool response lifecycle
2. Manually send a `ToolMessage` via `sendMessage` when the user submits

```tsx
import {
  useFrontendTool,
  useCopilotChatInternal,
} from "@copilotkit/react-core";
import { useAgent, useCopilotChatConfiguration } from "@copilotkitnext/react";
import { v4 as uuidv4 } from "uuid";

function MyInteractiveTool() {
  const { sendMessage } = useCopilotChatInternal();
  const chatConfig = useCopilotChatConfiguration();
  const { agent } = useAgent({ agentId: chatConfig?.agentId });

  useFrontendTool({
    name: "ask_question",
[...]

PR #4954, opened 2026-05-21
feat(react-core): allow respond() to override followUp per call

### LINKED ISSUE #3206: 🚀 Feature Request: Respond to tool call without followUp
### Idea

I'd like to use `AppendMessageOptions` or only the `followUp` option when using the `respond` function from `useHumanInTheLoop` to have better control over the messages before an agent response is generated.

### Use Case

I'm currently using `useHumanInTheLoop` to enable the agent to propose a list of elements to be created to the user that the user can edit and apply.
However the user should also be able to send a message with additional information and instructions if the proposition didn't meet their expectations.

If there is an unapproved tool call and the user sends their message, I want to first set a "rejected" tool result so that the tool call is completed gracefully and then send the message. Both - the user message and the tool result - should be send together to give the agent the full picture before a reply is generated.

In the `sendMessage` method I can use the optional `AppendMessageOptions` which offer the `followUp` option to tell CopilotKit to not immediately send the message.
In the `respond` method offered by `useHumanInTheLoop` these options are not exposed - instead `followUp` can currently only be configured in the arguments of `useHumanInTheLoop` beforehand.

If I understand it correctly, if the `followUp` option could be set when calling `respond`, I should be able to register the tool result "silently" and have it automatically be sent as soon as `sendMessage` is called with `followUp: true`.

To summarize my ideal scenario:
1. Agent executes frontend tool
2. User sends a message
2.1. Frontend tool is gracefully completed calling `respond` with "rejection" result and `followUp: false`
2.2. User message is sent calling `sendMessage` with `followUp: true`
2.3. CopilotKit sends both messages and triggers a new generation
3. Agent can reference the user decision to reject the proposition from the first tool call and their new instructions to continue the dialogue or execute frontend tool again with improved proposition

In a nutshell:
```typescript
// at design time (currently supported)
useHumanInTheLoop({
  followUp: true, // sets the default value
});

// at runtime (currently not supported)
respond(result, {
  followUp: false, // overwrites the default value if supplied
})

// at runtime (currently supported)
sendMessage(userMessage, {
  followUp: true, // sends two messages (the tool result above and the user message created here)
})
```

### Related Issues
#3154 Might be related, because they also needed to set `followUp: false` when calling `useHumanInTheLoop` and then manually call `sendMessage` with `followUp: true` inside the render function

sikandar100:
## What does this PR do?

Closes #3206.

`useHumanInTheLoop`'s `followUp` setting can currently only be configured once, at design time, on the hook. This PR exposes an optional `RespondOptions` argument on the `respond` callback so a handler can override `followUp` for a **single** response.

Motivating use case from #3206: a rejection path that should halt the agent loop (`respond(result, { followUp: false })`) while a normal approval still lets it continue — without registering two separate tools or falling back to the `sendMessage`-based workaround described in #3206 and #3154.

### Approach

The `copilotkit-support-bot` analysis on #3206 outlined two possible paths: **(A)** update the `respond` signature in `useHumanInTheLoop`, or **(B)** change the tool `handler` contract to carry the override. This PR takes **Path A** — the smaller, non-breaking change.

- Adds a `RespondOptions` interface (`{ followUp?: boolean }`), exported from `@copilotkit/react-core`.
- `respond` now accepts an optional second argument: `respond(result, options?)`.
- On each call, `respond` stamps the registered tool's `followUp` to either the runtime override (when provided) or the tool's design-time default (when not). Core reads it through the existing `followUp` gate in the run handler — **no core changes required**.

The "stamp design-time-or-override on every call" approach was chosen over snapshot/restore because core reads `followUp` after an async chain (handler resolution → subscriber notifications → run-handler continuation). Restoring synchronously would race that read; stamping on every call means a later `respond()` without options correctly falls back to the design-time value with no leak.

### Backward compatibility

Fully backward-compatible:

- The new argument is optional; `respond(result)` with no options behaves exactly as before.
- `undefined` is an explicit no-op — the design-time `followUp` is preserved.
- The v1 hook and `ActionRenderPropsWait` types are widened the same way, so the v1 shim gains the API with no behavior change.
- Verified: every existing `respond(...)` call site across the repo is single-argument.

### Changes

Six files, organized into three commits for review:

1. **`feat`** — `RespondOptions` type + `respond` implementation (v2 hook + types, v1 hook + `frontend-action` types)
2. **`test`** — 4 new e2e cases in `use-human-in-the-loop.e2e.test.tsx`
3. **`docs`** — `useHumanInTheLoop.mdx` signature update + new "Overriding followUp at response time" section

### Tests

Four new e2e cases:

- `respond({ followUp: false })` stops the agent loop when `followUp` defaults to `true`
- `respond({ followUp: true })` re-runs the agent on a tool whose design-time `followUp` is `false`
- `respond()` without options preserves the design-time `followUp`
- `respond({ followUp: undefined })` is a no-op

Re-run behavior is asserted by spying on `agent.run()`. All 4 pass; the 11 existing `useHumanInTheLoop` tests and the full `@copilotkit/react-core` suite (1174 tests) remain green.

Out of scope: Vue parity (`packages/vue/src/v2/hooks/use-human-in-the-loop.ts`) — happy to follow up separately.

## Related PRs and Issues

- Closes #3206 — the feature request this PR implements
- Related: #3154 — describes the same `followUp: false` + manual `sendMessage` workaround this PR removes the need for
- #4952, #4953 — two pre-existing, unrelated bugs found while developing this PR (filed separately). The automated tests here are unaffected by both, since the mock-agent harness bypasses the runtime/transport layers involved. For manual end-to-end HITL verification, HITL currently requires the single-route transport (`useSingleEndpoint`), per #4953.

## Checklist


BenTaylorDev (changes requested):
Thanks — this is a genuinely useful capability and the design is sound: `main` still forces the design-time `followUp` with no per-call override (`run-handler.ts:723` reads `tool?.followUp !== false` after the handler), `respond` is still single-arg (no `RespondOptions` anywhere), so it's not superseded, and the mechanism is right — mutating the registered tool via `getTool(...)` before resolving means core reads the overridden value. Two things before it can land:

1. **The docs edit targets a retired tree.** `docs/content/docs/reference/v2/hooks/useHumanInTheLoop.mdx` no longer exists on `main` (the old docs app was removed; docs now live under `showcase/shell-docs/src/content/`). That edit drops out on merge, so the new option would ship undocumented — please move it there.
2. **The `use-human-in-the-loop.tsx` change needs re-applying onto main's current `respond`, carefully.** Since this branch was opened, `respond` was restructured to tear down the abort listener before resolving (`use-human-in-the-loop.tsx:22-25`, `cleanupAbortRef.current?.()` + clearing `resolvePromiseRef`). Your `respond` rewrite predates that, so a straight merge/"take theirs" would **drop the abort cleanup → listener leak / double-reject**. Please graft the per-call `followUp` override onto the current `respond` without removing that cleanup, keeping the `followUp` stamp before `resolve()` so `run-handler.ts:723` reads it.

With the docs re-targeted and the override folded into the current hook, the rest looks good.

BenTaylorDev:
Closing this branch, but the gap is real: #3206 stays open, and I've learned something from your patch that will shape how it gets fixed.

First, credit where it's due. Reaching for `copilotkit.getTool()` and stamping `followUp` before resolving is the only channel the current API gives you — you found the one place a per-call decision can currently be expressed. The gate lives at `run-handler.ts:933`:

```ts
if (!handlerResult.error && tool?.followUp !== false) {
  return true; // Needs follow-up
}
```

It reads the **registration** at the moment your handler's promise settles, so a per-call decision genuinely has nowhere else to live today. Your code comment ("Restoring later is unsafe due to async chains in core") is an accurate diagnosis of a missing seam. The reason I don't want to merge it is that the seam is what needs adding: writing shared registration state to express a per-invocation choice breaks down with two responses in flight for the same tool, or an unmount between the write and core's read.

Second, something useful I found while reviewing this. Per-call follow-up control **already exists** in core, on the programmatic path — `copilotkit.runTool()` accepts it:

```ts
export interface CopilotKitCoreRunToolParams {
  followUp?: string | false;  // false | "generate" | any other string → inject a user message, then re-run
}
```

consumed at `run-handler.ts:1182`. That means two things. The feature is clearly right — the LLM-driven HITL path is simply missing what the programmatic path already has. And the shape should match: `RespondOptions.followUp?: boolean` would leave the codebase with two different per-call follow-up contracts, a boolean here and `string | false` there, where a string carries real meaning.

So the design I'd want is for the decision to travel **with the response** into the `:933` gate — the resolved handler result carrying an optional per-invocation follow-up, typed to match `runTool` — which drops the shared-state mutation and handles concurrency for free, since each invocation owns its own value.

Practically, this branch also can't be rebased into shape: beyond content conflicts in `use-human-in-the-loop.tsx` and its e2e test, the docs change targets `docs/content/docs/reference/v2/hooks/useHumanInTheLoop.mdx`, and `docs` on `main` is now a symlink to `showcase/shell-docs` — git reports `directory in the way of docs`, which a rebase doesn't cleanly resolve. Docs now live under `showcase/shell-docs/src/content/`.

I'll write the design up on #3206 so it's not lost. If you want to take another run at it with that shape, I'd be glad to review — and if you'd rather not, no problem at all; the analysis in this PR is what pinned down where the seam belongs. Thanks for it.

PR #4955, opened 2026-05-21
fix(react-core): wait for provider HITL responses

### LINKED ISSUE #4953: 🐛 Bug: `useHumanInTheLoop` silently fails under the default multi-route transport (CopilotKitProvider)

Tushar-Khandelwal-2004:
## What does this PR do?

Fixes provider-level `humanInTheLoop` tools so they properly wait for user interaction before resolving under the default multi-route transport.

Previously, provider-registered HITL tools used a placeholder handler that resolved immediately with `undefined`. In REST/multi-route flows, this caused the HITL UI to briefly render and disappear before `respond()` was invoked.

This PR introduces proper pending resolver handling for provider-level HITL flows and ensures tool execution only resolves after the user responds.

## Related PRs and Issues

-> Fixes #4953

## Changes

-> Store pending provider-level HITL resolvers by `toolCall.id`
-> Pass a functional `respond` callback to HITL renderers while the tool call is executing
-> Resolve the matching pending HITL handler only after the user responds
-> Handle aborted HITL tool calls by resolving and cleaning up pending resolvers
-> Update provider HITL tests to cover wait-until-respond behavior
-> Adjust wildcard/provider test assertions for wrapped HITL renderers

## Testing

```bash
node ..\..\node_modules\vitest\vitest.mjs run src/v2/providers/__tests__/CopilotKitProvider.test.tsx src/v2/providers/__tests__/CopilotKitProvider.wildcard.test.tsx
```

## Notes

-> Full pre-commit lint passed for the changed files
-> Repo-wide pre-commit tests were not fully clean locally due to unrelated/flaky failures in:

  -> `A2UIMessageRenderer.test.tsx`
  -> `@copilotkit/sqlite-runner:test`

### DIFF
--- a/packages/react-core/src/v2/providers/CopilotKitProvider.tsx
+++ b/packages/react-core/src/v2/providers/CopilotKitProvider.tsx
@@ -1,6 +1,7 @@
 "use client";
 
 import type { AbstractAgent } from "@ag-ui/client";
+import { ToolCallStatus } from "@copilotkit/core";
 import type { FrontendTool } from "@copilotkit/core";
 import type React from "react";
 import {
@@ -464,6 +465,9 @@ export const CopilotKitProvider: React.FC<CopilotKitProviderProps> = ({
     openGenerativeUI?.sandboxFunctions,
     "openGenerativeUI.sandboxFunctions must be a stable array.",
   );
+  const humanInTheLoopResolversRef = useRef<
+    Map<string, (result: unknown) => void>
+  >(new Map());
 
   // Note: warnings for array identity changes are handled by useStableArrayProp
 
@@ -480,16 +484,32 @@ export const CopilotKitProvider: React.FC<CopilotKitProviderProps> = ({
         parameters: tool.parameters,
         followUp: tool.followUp,
         ...(tool.agentId && { agentId: tool.agentId }),
-        handler: async () => {
-          // This handler will be replaced by the hook when it runs
-          // For provider-level tools, we create a basic handler that waits for user interaction
+        handler: async (_args, context) => {
           return new Promise((resolve) => {
-            // The actual implementation will be handled by the render component
-            // This is a placeholder that the hook will override
-            console.warn(
-              `Human-in-the-loop tool '${tool.name}' called but no interactive handler is set up.`,
-            );
-            resolve(undefined);
+            const toolCallId = context?.toolCall?.id;
+            if (!toolCallId) {
+              console.warn(
+                `Human-in-the-loop tool '${tool.name}' called without a tool call id.`,
+              );
+              resolve(undefined);
+              return;
+            }
+
+            const finish = (result: unknown) => {
+              humanInTheLoopResolversRef.current.delete(toolCallId);
+              resolve(result);
+            };
+
+            humanInTheLoopResolversRef.current.set(toolCallId, finish);
+
+            if (context.signal?.aborted) {
+              finish(undefined);
+              return;
+            }
+
+            context.signal?.addEventListener("abort", () => finish(undefined), {
+              once: true,
+            });
           });
         },
       };
@@ -497,10 +517,34 @@ export const CopilotKitProvider: React.FC<CopilotKitProviderProps> = ({
 
       // Add the render component to renderToolCalls
       if (tool.render) {
+        const RenderComponent: ReactToolCallRenderer<unknown>["render"] = (
+          props,
+        ) => {
+          const ToolComponent = tool.render;
+          const respond =
+            props.status === ToolCallStatus.Executing
+              ? async (result: unknown) => {
+                  humanInTheLoopResolversRef.current.get(props.toolCallId)?.(
+                    result,
+                  );
+                }
+              : undefined;
+
+          const enhancedProps = {
+            ...props,
+            name: tool.name,
+            description: tool.description || "",
+            agentId: tool.agentId,
+            respond,
+          } as React.ComponentProps<typeof ToolComponent>;
+
+          return <ToolComponent {...enhancedProps} />;
+        };
+
         processedRenderToolCalls.push({
           name: tool.name,
           args: tool.parameters,
-          render: tool.render as React.ComponentType<any>,
+          render: RenderComponent,
           ...(tool.agentId && { agentId: tool.agentId }),
         });
       }
--- a/packages/react-core/src/v2/providers/CopilotKitProvider.tsx
+++ b/packages/react-core/src/v2/providers/CopilotKitProvider.tsx
@@ -485,7 +485,7 @@ export const CopilotKitProvider: React.FC<CopilotKitProviderProps> = ({
         followUp: tool.followUp,
         ...(tool.agentId && { agentId: tool.agentId }),
         handler: async (_args, context) => {
-          return new Promise((resolve) => {
+          return new Promise((resolve, reject) => {
             const toolCallId = context?.toolCall?.id;
             if (!toolCallId) {
               console.warn(
@@ -495,21 +495,28 @@ export const CopilotKitProvider: React.FC<CopilotKitProviderProps> = ({
               return;
             }
 
+            const signal = context.signal;
+            const abort = () => {
+              humanInTheLoopResolversRef.current.delete(toolCallId);
+              reject(new Error("Human-in-the-loop interaction aborted"));
+            };
+            const cleanup = () => {
+              signal?.removeEventListener("abort", abort);
+            };
             const finish = (result: unknown) => {
+              cleanup();
               humanInTheLoopResolversRef.current.delete(toolCallId);
               resolve(result);
             };
 
             humanInTheLoopResolversRef.current.set(toolCallId, finish);
 
-            if (context.signal?.aborted) {
-              finish(undefined);
+            if (signal?.aborted) {
+              abort();
               return;
             }
 
-            context.signal?.addEventListener("abort", () => finish(undefined), {
-              once: true,
-            });
+            signal?.addEventListener("abort", abort, { once: true });
           });
         },
       };
--- a/packages/react-core/src/v2/providers/__tests__/CopilotKitProvider.test.tsx
+++ b/packages/react-core/src/v2/providers/__tests__/CopilotKitProvider.test.tsx
@@ -1,7 +1,14 @@
-import { render, renderHook, waitFor } from "@testing-library/react";
+import {
+  fireEvent,
+  render,
+  renderHook,
+  screen,
+  waitFor,
+} from "@testing-library/react";
 import type React from "react";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import { z } from "zod";
+import { ToolCallStatus } from "@copilotkit/core";
 import type { ReactFrontendTool } from "../../types/frontend-tool";
 import type { ReactHumanInTheLoop } from "../../types/human-in-the-loop";
 import { HttpAgent } from "@ag-ui/client";
@@ -225,11 +232,18 @@ describe("CopilotKitProvider", () => {
         (rc) => rc.name === "approvalTool",
       );
       expect(approvalTool).toBeDefined();
-      expect(approvalTool?.render).toBe(TestComponent);
+      expect(approvalTool?.render).toBeDefined();
     });
 
-    it("creates placeholder handlers for humanInTheLoop tools", async () => {
-      const TestComponent: React.FC<any> = () => <div>Test</div>;
+    it("waits for humanInTheLoop renderers to respond", async () => {
+      const TestComponent: React.FC<any> = ({ respond }) => (
+        <button
+          data-testid="hitl-respond"
+          onClick={() => respond?.({ approved: true })}
+        >
+          Respond
+        </button>
+      );
       const humanInTheLoopTools: ReactHumanInTheLoop[] = [
         {
           name: "interactiveTool",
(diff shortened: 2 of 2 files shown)


BenTaylorDev:
Thanks @Tushar-Khandelwal-2004 — the bug is still live on main (the provider-level placeholder handler resolves `undefined` immediately and provider renders aren't passed `respond`), and the approach matches the current render-props contract. The concern is drift: this branch is a long way behind main, and the HITL rendering area has changed a lot since — the `respond`/`toolCallId` render-props contract now exists centrally.

Before merging, could you rebase onto current main and confirm the current tool-call rendering pipeline doesn't **already** inject `respond` for provider-level renderers? If it does, this manual wrapper would double-wire it. A quick repro: register a HITL tool via the `humanInTheLoop` prop on CopilotKitProvider (not the `useHumanInTheLoop` hook) and verify the run still hangs without the patch and isn't double-injected with it.

Tushar-Khandelwal-2004:
Thanks for flagging the drift. I’ve rebased the PR onto the latest main and updated it.
I verified that the central tool-call rendering pipeline does not inject respond for humanInTheLoop tools registered through CopilotKitProvider; that behavior is currently only handled by the useHumanInTheLoop hook wrapper. The provider-level path was therefore still resolving immediately with undefined.
The updated patch adds the same pending-resolver behavior for provider-registered HITL tools, keyed by toolCallId, and only exposes respond while the tool is executing. I also added a regression test that confirms the handler remains pending until respond() is called, so it should not be double-wired

BenTaylorDev (changes requested):
Thanks for this. It does address a real, still-present gap — the provider-*prop* HITL path (`<CopilotKitProvider humanInTheLoop={[...]}>`) still has the immediate-`resolve(undefined)` placeholder on main (`CopilotKitProvider.tsx:483-493`). Two things to reconcile before merge, both stemming from the ~199 commits of HITL evolution since this was authored:

1. **Abort semantics invert the reference.** This resolves `undefined` on abort, but #5554 (`7d644f22f`) deliberately changed the canonical hook to *reject* (`new Error("Human-in-the-loop interaction aborted")`) so core records an explicit error tool result instead of silently resolving to empty. Please match that — reject on abort — rather than reintroducing the empty-result-on-abort behavior #5554 removed.
2. **Confirm the target issue.** #4953 reproduces with the `useHumanInTheLoop` *hook*, which on current v2 main already has the wait+abort mechanism and e2e coverage (`use-human-in-the-loop.tsx:30-58`). This PR fixes the separate provider-prop path — a real bug, but possibly not the one #4953 describes. Worth checking whether #4953 still reproduces on main, and re-attaching to the right issue.

Also: the test wires the resolver plumbing directly but doesn't drive a real agent/multi-route transport or cover the abort branch — please add coverage that reproduces the actual scenario (agent emits TOOL_CALL → RUN_FINISHED → UI must persist) and exercises abort.

Tushar-Khandelwal-2004:
Thanks @BenTaylorDev I’ve addressed both points and pushed the update.

The provider-level handler now matches the canonical hook behavior on abort: it rejects with `Human-in-the-loop interaction aborted` rather than resolving `undefined`, so core records an explicit error result.

I also added coverage for the provider-prop flow itself:

- an abort test for a pending provider HITL handler; and
- an end-to-end event-sequence test that simulates `TOOL_CALL → RUN_FINISHED`, verifies the HITL UI remains interactive, and confirms it completes only after `respond()` is called.

I rechecked #4953 as well: the current `useHumanInTheLoop` hook path already has the wait/abort handling and e2e coverage, so this PR is more accurately fixing the separate `<CopilotKitProvider humanInTheLoop={[...]}>` path rather than the hook-specific reproduction.

BenTaylorDev (changes requested):
Really close — the abort handling is exactly right now (rejects with the same error as the canonical hook, matching #5554), and the new transport-driven + abort tests are great. One stale assertion is failing, though: `CopilotKitProvider.test.tsx:597` ("combines render components from all sources") still asserts `humanRenderTool?.render).toBe(TestComponent2)`, but since the render is now wrapped (`RenderComponent`), that identity no longer holds — you updated the sibling assertion in the "waits for…" test but missed this one. Change it to `.toBeDefined()` (or assert the wrapped render's behavior) and it's green. Verified the rest locally: 33/34 pass, lint clean, no type errors in the changed file.

Tushar-Khandelwal-2004:
Thanks @BenTaylorDev - fixed the stale assertion and pushed b7853fa. It now verifies that the provider HITL renderer is present, rather than comparing it to the original component by identity, since the provider path intentionally wraps that renderer.

BenTaylorDev (changes requested):
Thanks — the abort handling and the new transport-driven/abort tests are right, and `CopilotKitProvider.test.tsx` is 34/34 green now. One thing left, and it's more than a stale assertion.

`CopilotKitProvider.wildcard.test.tsx:150` fails on this branch (main is 9/9, branch is 8/9). The wildcard assertion adjustment your PR description mentions isn't in the current diff — looks like it was lost in the rebase.

Please don't fix it by relaxing to `.toBeDefined()` — it's catching a real regression. The wrapper's `name: tool.name` overwrites the actual invoked tool name with `"*"` for wildcard registrations. The render pipeline always passes the real tool name (`use-render-tool-call.tsx:49/59/69`), and there's a contract test for it (`defineToolCallRenderer.test.tsx:204-236`, "wildcard receives actual tool name, not `*`"). On main the provider path passed it through unwrapped; with this patch a wildcard HITL renderer can't tell which tool it's approving. Verified: main → `"deleteFile"`, this branch → `"*"`.

Also needs a rebase — the branch no longer merges cleanly (small import conflict in the test file from the React 18 + 19 matrix commit `4049434b3b`).

FYI, the same wildcard name-clobber already exists in the `useHumanInTheLoop` hook on main, and the Vue provider still has the immediate-`resolve(undefined)` placeholder — both out of scope here, I'll file them separately.

BenTaylorDev:
Thanks for this, and sorry it sat so long. Closing it in favor of #6851, which lands the same fix.

To be clear about what happened: your diagnosis was right and your implementation was right. The provider-prop path really did never wait — `resolve(undefined)` fired the moment the agent invoked the tool — and the parking handler keyed by tool call id plus the abort-rejects-with-an-error semantics in this PR are exactly what the fix needs. #6851 carries both of those over unchanged in substance.

Two things kept this branch from being the one that merges. The wildcard blocker from my last review is still open: the wrapper sets `name: tool.name`, which for a `"*"` registration hands the renderer `"*"` instead of the tool actually invoked, so a catch-all HITL render cannot tell what it is approving. And the branch has fallen behind `main` far enough that it no longer merges cleanly. #6851 adds the wildcard branch and pipeline-level tests that mount the real chat → tool-call → renderer path, on top of current `main`.

The credit for this one stays with you — #6851 says so in its description, and it links back here as the earlier implementation. I would genuinely welcome your next PR.

PR #4960, opened 2026-05-21
fix(react-core): wrap CopilotChatInput disclaimer slot in pointer-events-auto

jerelvelarde:
## Why this matters

Every CopilotKit consumer that puts interactive content in the disclaimer slot below the input pill hits the same trap: buttons and links render but don't respond to clicks or hover. The cause isn't obvious from outside the framework — the outer input container is intentionally `pointer-events: none` so the absolutely-positioned input doesn't block clicks on chat messages behind it, the pill itself is re-enabled, but the disclaimer slot is a sibling that never gets re-enabled.

For a canonical starter, this is friction every customizer hits the first time they add an agent selector, model picker, or any interactive disclaimer content. They debug for 10–30 minutes, find the workaround by reading the framework source, and ship a `pointerEvents: "auto"` override they didn't need to know about.

image

image

## How it's implemented

Single-file change to `packages/react-core/src/v2/components/chat/CopilotChatInput.tsx`: the disclaimer render now wraps in `<div className="cpk:pointer-events-auto">`, mirroring the existing wrap around the input pill in the same file a few lines above.

The outer container's `pointer-events: none` is unchanged — pass-through over the chat column above the input continues to work as before.

## Migration

Consumers who currently work around this by manually setting `pointerEvents: "auto"` on their disclaimer slot will see their override become a redundant no-op. No functional change for them; they can drop the override when convenient.

Vue counterpart `packages/vue/src/v2/components/chat/CopilotChatInput.vue`
has the same disclaimer pointer-events trap. To be addressed in a
follow-up PR — left out of this scope to keep the React fix isolated.

### DIFF
--- a/.changeset/fix-disclaimer-pointer-events.md
+++ b/.changeset/fix-disclaimer-pointer-events.md
@@ -1,0 +1,5 @@
+---
+"@copilotkit/react-core": patch
+---
+
+fix(react-core): wrap CopilotChatInput disclaimer slot in pointer-events-auto so interactive content (links, buttons, agent selectors) inside a custom disclaimer stays clickable. The outer container's pointer-events:none is preserved.
--- a/packages/react-core/src/v2/components/chat/CopilotChatInput.tsx
+++ b/packages/react-core/src/v2/components/chat/CopilotChatInput.tsx
@@ -1125,7 +1125,9 @@ export function CopilotChatInput({
       <div className="cpk:max-w-3xl cpk:mx-auto cpk:py-0 cpk:px-4 cpk:sm:px-0 cpk:[div[data-sidebar-chat]_&]:px-8 cpk:[div[data-popup-chat]_&]:px-4 cpk:pointer-events-auto">
         {inputPill}
       </div>
-      {shouldShowDisclaimer && BoundDisclaimer}
+      {shouldShowDisclaimer && (
+        <div className="cpk:pointer-events-auto">{BoundDisclaimer}</div>
+      )}
     </div>
   );
 }
--- a/packages/react-core/src/v2/components/chat/__tests__/CopilotChatInput.slots.e2e.test.tsx
+++ b/packages/react-core/src/v2/components/chat/__tests__/CopilotChatInput.slots.e2e.test.tsx
@@ -925,5 +925,34 @@ describe("CopilotChatInput Slot System E2E Tests", () => {
         expect(disclaimer.textContent).toContain("Custom Disclaimer");
       }
     });
+
+    it("should wrap disclaimer in pointer-events-auto so interactive content stays clickable", () => {
+      // Regression guard: the outer CopilotChatInput container is
+      // cpk:pointer-events-none so it doesn't block clicks on chat messages
+      // behind the absolutely-positioned input. The disclaimer slot is a
+      // sibling of the input pill wrapper and must explicitly re-enable
+      // pointer events; otherwise buttons/links rendered into the slot
+      // silently fail to receive clicks. Lock the wrapper class in so a
+      // future JSX refactor can't quietly regress it.
+      const CustomDisclaimer: React.FC<any> = () => (
+        <button data-testid="disclaimer-action">Click me</button>
+      );
+
+      const { container } = render(
+        <TestWrapper>
+          <CopilotChatInput
+            showDisclaimer={true}
+            disclaimer={CustomDisclaimer}
+          />
+        </TestWrapper>,
+      );
+
+      const action = container.querySelector(
+        '[data-testid="disclaimer-action"]',
+      );
+      expect(action).not.toBeNull();
+      const wrapper = action!.closest(".cpk\\:pointer-events-auto");
+      expect(wrapper).not.toBeNull();
+    });
   });
 });


jerelvelarde:
## Automated review

**Verdict: LGTM** — minor optional nit on test coverage.

### Overview
Single-file fix in `packages/react-core/src/v2/components/chat/CopilotChatInput.tsx` (+3 / -1). The outer input container (line 1101) has `cpk:pointer-events-none` so it doesn't block clicks on chat messages behind the absolutely-positioned input. The input pill wrapper at line 1125 already re-enables pointer events with `cpk:pointer-events-auto`, but the sibling disclaimer slot at line 1128 never was — so interactive content rendered inside the disclaimer slot (agent selectors, model pickers, links) couldn't receive clicks/hover. This PR wraps `{BoundDisclaimer}` in a `<div className="cpk:pointer-events-auto">` mirroring the existing input-pill wrapper pattern.

### Correctness
- Right shape: `pointer-events-auto` only re-enables inside the wrapper, preserving the outer container's pass-through over the chat column above the input.
- Wrapper exists only on the default render path (no `children` prop). The render-props branch at lines 597–628 passes `BoundDisclaimer` through to consumers as a slot, so headless consumers retain full control — consistent with how outer pointer events are handled.
- `shouldShowDisclaimer` (line 595) gates rendering, so no empty wrapper div is added when nothing is shown.
- Consumers currently overriding `pointerEvents: "auto"` inline (per PR description) will see their override become a redundant no-op — no regression.

### Style / Conventions
- Uses the `cpk:` Tailwind prefix consistently.
- Pattern mirrors the input-pill wrapper directly above (line 1125), which is the correct reference. Visual symmetry makes the relationship discoverable.
- No new exports, no API surface changes, no breaking changes.

### Test coverage
- Existing disclaimer tests in `packages/react-core/src/v2/components/chat/__tests__/CopilotChatInput.slots.e2e.test.tsx` lines 836–928 cover render presence/absence and class application but do **not** assert pointer-events on the wrapper. Optional regression assertion (cheap insurance against future JSX refactors):
  ```ts
  const wrapper = container.querySelector('.cpk\\:pointer-events-auto');
  expect(wrapper).not.toBeNull();
  ```
  Scoped to the disclaimer container. Not blocking — the visual symmetry with the pill wrapper makes accidental removal unlikely, and asserting compiled Tailwind `pointer-events: auto` through jsdom is awkward.

### Risks / edge cases
- None on the default render path. The change is additive and scoped to one conditional branch.
- Headless `children`-prop consumers are unaffected (they manage their own pointer events around the disclaimer slot — consistent with the existing slot contract).
- A custom disclaimer component that internally relies on `pointer-events: none` (very unlikely) could see its inner pointer behavior shadowed; child elements can still opt out with their own `pointer-events: none`, so the escape hatch remains.

### Performance / Security
No impact on either. One extra div in the tree only when the disclaimer is shown.

BenTaylorDev:
Verified locally: on current `main` the outer `[data-copilotkit]` container is `pointer-events-none` and the disclaimer renders raw, so interactive content in a custom disclaimer is genuinely unclickable (a hit-test at a disclaimer button's center resolves to a `div`, not the button). The wrapper fixes it.

Two changes before merge:
1. Prefer putting `cpk:pointer-events-auto` on the `Disclaimer` slot's own className (`CopilotChatInput.tsx:1401`, already `max-w-3xl mx-auto`) rather than a new full-width wrapper. That covers both the default layout **and** the `children` render-prop path, and constrains the clickable band to the disclaimer width instead of the full row.
2. Drop `.changeset/fix-disclaimer-pointer-events.md` — the repo removed `.changeset/` in `5afa55f06`.

BenTaylorDev (commented):
Thanks for this — the diagnosis is right and the fix is in the right place. I verified it end to end rather than taking CI's word for it:

- The bug is **still live on current `main`**: `CopilotChatInput.tsx:1116` sets `cpk:pointer-events-none` on the container and `:1139` still renders `{shouldShowDisclaimer && BoundDisclaimer}` unwrapped. Nothing has superseded it.
- The branch is stale but **merges clean** against `main` — the only nearby drift is a context line (`cpk:sm:px-0` → `cpk:@3xl:px-0`), which doesn't collide.
- **Mechanism confirmed, not just asserted**: `pointer-events` is inherited, `.cpk\:pointer-events-auto{pointer-events:auto}` really is in the shipped `dist/v2/index.css`, and the same pattern already works for the input pill at `:1136` under the *same two* `pointer-events:none` ancestors (`CopilotChatView.tsx:429` + `CopilotChatInput.tsx:1116`). I also checked the fix is on the real render path — `CopilotChatView` passes the disclaimer *through* `CopilotChatInput` (`CopilotChatView.tsx:268,274` and `:345,346`), so there's no bypass.
- **Tests**: 49/49 pass on the merged tree, and reverting just the source hunk makes the new test fail — it's a real regression guard.
- **Scope check**: Angular isn't affected (`copilot-chat-view-input-container.ts` never sets `pointer-events-none`), so React + the Vue follow-up you flagged are the complete set. No layout regression from the extra div — the only child-combinator CSS in `globals.css` is scoped to `[data-streamdown="table-wrapper"]`.

The `format` and Vercel reds are not on you: `format` failed with `A branch or tag with the name 'jerel/fix-disclaimer-pointer-events' could not be found` (fork checkout, not a formatting violation), and the Vercel jobs are stale example previews this diff doesn't touch. I ran `oxfmt --check` on both changed files locally — clean.

Two things before this can go in:

**1. Please drop `.changeset/fix-disclaimer-pointer-events.md`.** Changesets are retired in this repo — there's no `.changeset/config.json`, no `@changesets/cli` dependency, and no release workflow that consumes them (see `1e5ba689e0`, "fix: remove all changesets"). Releases go through the release workflow's package scopes instead, so the file is inert. Just the fix plus its test, please.

**2. Consider constraining the wrapper's width.** The wrapper is full container width, whereas the pill wrapper at `:1136` is `max-w-3xl mx-auto`. That turns the ~40px disclaimer strip into a full-width click-blocker on the absolute input overlay. It's invisible at rest — `CopilotChatView.tsx:303` pads the scroll content by the measured container height — but while the message list is scrolled up, content passing under that strip (a link, a copy button, a generative-UI card) was clickable before and won't be now, and wheel-scroll over the strip no longer reaches the scroller since none of the band's ancestors are scrollable.

```diff
-        <div className="cpk:pointer-events-auto">{BoundDisclaimer}</div>
+        <div className="cpk:pointer-events-auto cpk:mx-auto cpk:max-w-3xl">
+          {BoundDisclaimer}
+        </div>
```

That keeps the new blocking region to the same footprint the pill already occupies. If you'd rather keep it full-width, that's defensible — but then please amend "pass-through over the rest of the chat column is preserved" in the description, since the disclaimer band itself is exactly where pass-through is now lost.

PR #4961, opened 2026-05-21
fix(react-core): reserve safe-area bottom padding for in-flow CopilotChatInput

jerelvelarde:
## Why this matters

Any consumer embedding `CopilotChat` in a custom column layout — anything other than the full-screen absolute-positioned default — sees the input pill butt directly against the bottom edge of its container. The framework reserves `paddingBottom` only when `positioning === "absolute"` or `bottomAnchored` is set; the in-flow static mode (welcome-screen branch, most column embeds) gets nothing.

The visual cost is small but cumulative: every embed-in-a-card pattern, every glass-card chat panel, every two-pane app layout independently re-invents the `pb-6` workaround. For a canonical starter, the framework should default to a sensible inset that survives every positioning mode.

Issue:
image

image

## How it's implemented

Single-file change to `packages/react-core/src/v2/components/chat/CopilotChatInput.tsx`: the existing `paddingBottom` conditional now has a static-branch arm that reserves `var(--copilotkit-input-safe-area, 12px)`. Absolute and bottom-anchored modes are unchanged — they already use `var(--copilotkit-license-banner-offset, 0px)` and lay against the viewport edge.

The new `--copilotkit-input-safe-area` CSS variable defaults to 12px (chosen to match common manual workarounds minus existing internal pill padding). Consumers in tight popovers can set it to `0px`; consumers in roomy glass cards can raise it.

## Migration

- Apps that today work around this with `className="pb-6"` on a static-positioned CopilotChatInput will see `12px + 24px = 36px` stacking. Recommend dropping the workaround.
- Apps that depended on zero bottom inset in static mode can restore by setting `--copilotkit-input-safe-area: 0px` on their app root.
- Absolute / bottom-anchored modes are untouched — no regression possible there.

Vue counterpart `packages/vue/src/v2/components/chat/CopilotChatInput.vue` carries the same `paddingBottom`-only-when-absolute gate, and `packages/vue/src/v2/components/chat/__tests__/CopilotChatInput.bottomAnchored.test.ts:47-53` actively asserts the broken contract (`paddingBottom === ""` for welcome static input). `packages/vue/PARITY.md:379` documents that contract.

To be addressed in a follow-up PR — left out of this scope to keep the React fix isolated. The Vue parity test will start to fail once that follow-up lands; this PR is React-only by design.

### DIFF
--- a/.changeset/fix-input-safe-area-padding.md
+++ b/.changeset/fix-input-safe-area-padding.md
@@ -1,0 +1,5 @@
+---
+"@copilotkit/react-core": patch
+---
+
+fix(react-core): reserve safe-area bottom padding for in-flow CopilotChatInput via new `--copilotkit-input-safe-area` CSS variable (default `12px`). Absolute-positioned and bottom-anchored modes are unchanged.
--- a/packages/react-core/src/v2/components/chat/CopilotChatInput.tsx
+++ b/packages/react-core/src/v2/components/chat/CopilotChatInput.tsx
@@ -1118,7 +1118,7 @@ export function CopilotChatInput({
         // must stay still when the banner is present.
         ...(positioning === "absolute" || bottomAnchored
           ? { paddingBottom: "var(--copilotkit-license-banner-offset, 0px)" }
-          : {}),
+          : { paddingBottom: "var(--copilotkit-input-safe-area, 12px)" }),
       }}
       {...props}
     >
--- a/packages/react-core/src/v2/components/chat/__tests__/CopilotChatInput.bottomAnchored.test.tsx
+++ b/packages/react-core/src/v2/components/chat/__tests__/CopilotChatInput.bottomAnchored.test.tsx
@@ -1,0 +1,50 @@
+import React from "react";
+import { render } from "@testing-library/react";
+import { describe, expect, it } from "vitest";
+import { CopilotChatInput } from "../CopilotChatInput";
+import { CopilotChatConfigurationProvider } from "../../../providers/CopilotChatConfigurationProvider";
+
+const TEST_THREAD_ID = "test-thread";
+
+const BANNER_OFFSET_VAR = "var(--copilotkit-license-banner-offset, 0px)";
+const SAFE_AREA_VAR = "var(--copilotkit-input-safe-area, 12px)";
+
+function renderInput(props: React.ComponentProps<typeof CopilotChatInput>) {
+  return render(
+    <CopilotChatConfigurationProvider threadId={TEST_THREAD_ID}>
+      <CopilotChatInput {...props} />
+    </CopilotChatConfigurationProvider>,
+  );
+}
+
+describe("CopilotChatInput bottom padding", () => {
+  it('reserves license-banner offset when positioning="absolute" and bottomAnchored=false', () => {
+    const { container } = renderInput({ positioning: "absolute" });
+    const outer = container.firstChild as HTMLElement;
+    expect(outer.style.paddingBottom).toBe(BANNER_OFFSET_VAR);
+  });
+
+  it('reserves license-banner offset when positioning="absolute" and bottomAnchored=true', () => {
+    const { container } = renderInput({
+      positioning: "absolute",
+      bottomAnchored: true,
+    });
+    const outer = container.firstChild as HTMLElement;
+    expect(outer.style.paddingBottom).toBe(BANNER_OFFSET_VAR);
+  });
+
+  it('reserves license-banner offset when positioning="static" and bottomAnchored=true', () => {
+    const { container } = renderInput({
+      positioning: "static",
+      bottomAnchored: true,
+    });
+    const outer = container.firstChild as HTMLElement;
+    expect(outer.style.paddingBottom).toBe(BANNER_OFFSET_VAR);
+  });
+
+  it('reserves safe-area padding when positioning="static" and bottomAnchored=false (welcome / in-flow)', () => {
+    const { container } = renderInput({ positioning: "static" });
+    const outer = container.firstChild as HTMLElement;
+    expect(outer.style.paddingBottom).toBe(SAFE_AREA_VAR);
+  });
+});


jerelvelarde:
## Automated review

**Verdict:** Minor nits — the core change is correct and well-scoped, but there are a few cross-cutting concerns worth addressing before merge.

### Overview

Single-line change in `packages/react-core/src/v2/components/chat/CopilotChatInput.tsx` (line 1121) that fills in the previously-empty fall-through arm of the `paddingBottom` conditional:

```tsx
...(positioning === "absolute" || bottomAnchored
  ? { paddingBottom: "var(--copilotkit-license-banner-offset, 0px)" }
  : { paddingBottom: "var(--copilotkit-input-safe-area, 12px)" }),
```

Branch coverage across the four `positioning` x `bottomAnchored` combinations:

| `positioning` | `bottomAnchored` | Before | After |
| --- | --- | --- | --- |
| `absolute` | `false` | `license-banner-offset` | unchanged |
| `absolute` | `true` | `license-banner-offset` | unchanged |
| `static` | `true` | `license-banner-offset` | unchanged (CopilotChatView main run state) |
| `static` | `false` | none | `safe-area, 12px` (welcome screen + custom embeds) |

Only the last row changes — which matches the PR description.

### Correctness

- Logic is sound: `||` correctly short-circuits the three bottom-anchored branches to the banner-offset arm; only the in-flow static-non-anchored case picks up the new variable.
- `12px` default is conservative and matches the "common manual workaround minus existing internal pill padding" rationale.
- `inputContainerHeight` is measured via `ResizeObserver` in `CopilotChatView.tsx` (line 200) and feeds `paddingBottom` on the scroll content. Since this PR's new padding only applies when `bottomAnchored` is `false`, and the main run-state input in `CopilotChatView` sets `bottomAnchored: true` (CopilotChatView.tsx:263), the new padding cannot compound into the scroll-content offset on the main run-state path. The welcome-screen input is rendered separately and is not the one being measured by the overlay observer, so no cascade there either. Good.

### Nits / concerns

1. **Vue/React parity drift (minor blocker for the contract).** The Vue counterpart in `packages/vue/src/v2/components/chat/CopilotChatInput.vue` (lines 959-961) is unchanged — it still has an empty `else` branch. More importantly, `packages/vue/src/v2/components/chat/__tests__/CopilotChatInput.bottomAnchored.test.ts:47-53` explicitly asserts:

   ```ts
   it("does NOT reserve padding-bottom for the welcome (static, non-bottomAnchored) input", () => {
     // ...
     expect(container.style.paddingBottom).toBe("");
   });
   ```

   This means the documented cross-framework contract (per `packages/vue/PARITY.md:379`) says the welcome static input gets no bottom inset. This PR breaks that contract on the React side without updating the Vue side or the parity doc. Either (a) port the change to Vue and update the parity test, or (b) document the intentional React-only behavior.

2. **No React-side test for the new behavior.** `packages/react-core/src/v2/components/chat/__tests__/CopilotChatInput.test.tsx` and the `.slots.e2e.test.tsx` file exercise the `positioning` prop but not the `paddingBottom` assertion for any of the four cases. Adding a small unit test that mirrors the Vue parity suite — checking each of the four combinations — would lock the contract in and prevent regression. The patch is one line, but the contract surface is four cases.

3. **No changeset.** Sibling single-line fixes in `.changeset/` (e.g. `little-pears-tell.md`, `five-avocados-visit.md`) ship with a patch-level changeset. Since this is a user-visible visual change to `@copilotkit/react-core`, a changeset would let downstream consumers see this in release notes — especially given the migration note about stacking with `pb-6` workarounds.

4. **`12px` default — reasonable, but consider documenting.** The CSS variable is undocumented anywhere outside this code site. A one-line comment near the conditional, or a `--copilotkit-input-safe-area` mention in whichever doc lists CopilotKit CSS variables, would help consumers discover the knob (especially the "set to `0px` to opt out" pattern from the PR description). The `--copilotkit-license-banner-offset` variable has supporting comments at `license-warning-banner.tsx:6`; the new variable could use the same treatment.

5. **Stacking edge cases.** Worth a brief look in a showcase:
   - Welcome-screen embeds that already have a `pb-*` Tailwind utility on a parent — the new 12px stacks with parent padding, which is a small visual shift but not a bug.
   - Tight popovers / sidebars where the existing pill already kisses the bottom border by design — those consumers will see a 12px gap appear unless they set `--copilotkit-input-safe-area: 0px`. This is called out in the PR migration note; just confirming it'll likely surface a couple of cosmetic adjustments in existing showcases.

### Risks

- Low. Absolute / bottom-anchored modes are untouched. The only behavioral change is in the welcome-screen static-non-anchored path. Worst-case downstream impact is a 12px visual shift that's easily overridden via the CSS variable.

### Suggestions

- Mirror the change in `packages/vue/src/v2/components/chat/CopilotChatInput.vue` and update `packages/vue/src/v2/components/chat/__tests__/CopilotChatInput.bottomAnchored.test.ts:47-53` + `packages/vue/PARITY.md:379` to keep the contract aligned.
- Add a React-side test in `CopilotChatInput.test.tsx` (or a new `.paddingBottom.test.tsx`) asserting all four `positioning` x `bottomAnchored` combinations resolve to the expected CSS variable.
- Add a patch-level changeset for `@copilotkit/react-core` (and `@copilotkit/vue` if you port it).
- Consider a brief code comment near line 1121 explaining the new variable's intent (analogous to the existing comment at 1114-1118).

LGTM on the core fix once parity and tests are squared away.

BenTaylorDev:
Mechanism + escape hatch verified locally (static input `paddingBottom` 0→12px; overriding `--copilotkit-input-safe-area` takes effect).

One design change: the var is named `--copilotkit-input-safe-area` but defaults to a flat `12px`, while the codebase uses real `env(safe-area-inset-bottom)` for popup/sidebar (`CopilotPopupView.tsx:196`, `CopilotSidebarView.tsx:187`). Default it to `env(safe-area-inset-bottom, 12px)` so it honors notched devices and matches the existing pattern. Also drop the `.changeset/` file (`.changeset/` was removed from the repo in `5afa55f06`).

PR #4962, opened 2026-05-21
fix(react-core): scope [data-copilotkit] background to chat root only

jerelvelarde:
## Why this matters

The framework applies `background-color: var(--background)` to every element carrying the `data-copilotkit` attribute — which today is dozens of chat-internal nodes (assistant messages, suggestion containers, input wrappers, attachment chips). The attribute exists on those descendants for CSS-variable scoping, not because each one should paint its own opaque background.

The consequence: any consumer wrapping the chat in a translucent surface — glass cards, gradient backdrops, dark overlays — sees every chat-internal element paint solid `--background` (default white) directly over their backdrop. Custom theming is silently defeated. The workaround is a per-element `background-color: transparent !important` override applied across the whole chat tree.

For a canonical starter, the framework's CSS-variable token surface should be ergonomic to override. Today it isn't — every translucent theme starts with a one-page CSS prelude before it can render correctly.

image

## How it's implemented

Two changes:

1. **Split the CSS rule** in `packages/react-core/src/v2/styles/globals.css`. The `color` declaration stays on `[data-copilotkit]` (so descendant elements inherit foreground correctly). The `background-color` declaration moves to a new `[data-copilotkit-root]` selector that only chat root containers carry.

2. **Add `data-copilotkit-root` attribute** to chat root containers — `CopilotChatView`'s outer container, `CopilotSidebar`, `CopilotPopup`, and any other top-level chat surface. Each root keeps its existing `data-copilotkit` and additionally gets the new attribute.

Default-theme appearance is unchanged (the root still paints `--background`). The cascade is also unchanged for `color`. Only the per-descendant background paint is removed.

## Migration

Apps relying on `CopilotSidebar` / `CopilotPopup` defaults: no change. Apps with custom translucent embeds: gain transparent descendants by default — exactly the theming-friendly behavior they were working around.

The rare consumer that depended on every internal node painting `--background` can restore the old behavior with a one-line override in their own CSS:
```css
[data-copilotkit] { background-color: var(--background); }
```

Vue and Angular packages carry the same combined CSS rule and same `data-copilotkit`-on-internal-nodes pattern:

- `packages/vue/src/styles/globals.css:142-145`
- `packages/angular/src/styles/globals.css:142-145`

To be addressed in a follow-up PR — left out of this scope to keep the React fix isolated and reviewable. Custom translucent themes will continue to need per-element overrides in Vue/Angular consumers until that lands.

### DIFF
--- a/.changeset/fix-data-copilotkit-bg-scope.md
+++ b/.changeset/fix-data-copilotkit-bg-scope.md
@@ -1,0 +1,9 @@
+---
+"@copilotkit/react-core": patch
+---
+
+fix(react-core): scope [data-copilotkit] background-color to chat root only
+
+The `data-copilotkit` attribute is applied to dozens of chat-internal nodes (assistant messages, suggestion containers, input wrappers, attachment chips) for CSS-variable scoping. The base rule painted `background-color` on all of them, which defeated any consumer wrapping the chat in a translucent surface — every internal element painted opaque `--background` over their backdrop.
+
+Split the rule: `color` stays on `[data-copilotkit]` for descendant inheritance; `background-color` moves to a new `[data-copilotkit-root]` selector that only chat root containers carry. `CopilotChatView`, `CopilotSidebarView`, and `CopilotPopupView` are tagged with the new attribute, so default-theme appearance is unchanged. Custom translucent themes no longer need per-element overrides.
--- a/packages/react-core/src/v2/components/chat/CopilotChatView.tsx
+++ b/packages/react-core/src/v2/components/chat/CopilotChatView.tsx
@@ -366,6 +366,7 @@ export function CopilotChatView({
     return (
       <div
         data-copilotkit
+        data-copilotkit-root
         data-testid="copilot-chat"
         data-copilot-running={isRunning ? "true" : "false"}
         onDragOver={onDragOver}
@@ -399,6 +400,7 @@ export function CopilotChatView({
   return (
     <div
       data-copilotkit
+      data-copilotkit-root
       data-testid="copilot-chat"
       data-copilot-running={isRunning ? "true" : "false"}
       onDragOver={onDragOver}
--- a/packages/react-core/src/v2/components/chat/CopilotPopupView.tsx
+++ b/packages/react-core/src/v2/components/chat/CopilotPopupView.tsx
@@ -1,12 +1,14 @@
 import React, { useEffect, useMemo, useRef, useState } from "react";
-import CopilotChatView, {
+import type {
   CopilotChatViewProps,
   WelcomeScreenProps,
 } from "./CopilotChatView";
+import CopilotChatView from "./CopilotChatView";
 import CopilotChatToggleButton from "./CopilotChatToggleButton";
 import { CopilotModalHeader } from "./CopilotModalHeader";
 import { cn } from "../../lib/utils";
-import { renderSlot, SlotValue } from "../../lib/slots";
+import type { SlotValue } from "../../lib/slots";
+import { renderSlot } from "../../lib/slots";
 import {
   CopilotChatConfigurationProvider,
   CopilotChatDefaultLabels,
@@ -208,6 +210,7 @@ function CopilotPopupViewInternal({
   const popupContent = isRendered ? (
     <div
       data-copilotkit
+      data-copilotkit-root
       className={cn(
         "cpk:fixed cpk:inset-0 cpk:z-[1200] cpk:flex cpk:max-w-full cpk:flex-col cpk:items-stretch",
         "cpk:md:inset-auto cpk:md:bottom-24 cpk:md:right-6 cpk:md:items-end cpk:md:gap-4",
--- a/packages/react-core/src/v2/components/chat/CopilotSidebarView.tsx
+++ b/packages/react-core/src/v2/components/chat/CopilotSidebarView.tsx
@@ -1,9 +1,10 @@
 import React, { useEffect, useLayoutEffect, useRef, useState } from "react";
 
-import CopilotChatView, {
+import type {
   CopilotChatViewProps,
   WelcomeScreenProps,
 } from "./CopilotChatView";
+import CopilotChatView from "./CopilotChatView";
 import {
   CopilotChatConfigurationProvider,
   useCopilotChatConfiguration,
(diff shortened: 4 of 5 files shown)


jerelvelarde:
## Automated review

**Verdict:** Low-risk, well-scoped CSS surface fix. Default rendering is preserved, translucent consumer themes regain a clean override surface, and the implementation matches the stated migration story.

### Overview
Splits the single `[data-copilotkit] { color; background-color; }` rule in `packages/react-core/src/v2/styles/globals.css` so `color` continues to apply to every CopilotKit-scoped element (descendants inherit foreground), while `background-color: var(--background)` now applies only to a new `[data-copilotkit-root]` selector. The new attribute is added to the outermost containers of `CopilotChatView` (both branches), `CopilotSidebarView`, and `CopilotPopupView`.

### Code correctness
- **Default-theme parity holds.** The three components that previously painted the chat background — `CopilotChatView`, `CopilotSidebarView` aside, `CopilotPopupView` outer wrapper — all carry both `data-copilotkit` and `data-copilotkit-root`. The first paint of `--background` against the consumer's surface still happens.
- **Descendant background loss is benign.** Components that previously inherited `var(--background)` purely from the `[data-copilotkit]` rule and that *need* a background already declare it explicitly via Tailwind: `CopilotSidebarView` aside has `cpk:bg-background`, `CopilotPopupView` content card has `cpk:bg-background`, `CopilotModalHeader` has `cpk:bg-background/95`, `CopilotChatSuggestionPill` has `cpk:bg-background`, ui `Button` has `cpk:bg-background`. Portaled UI (Radix tooltip, dropdown) uses `cpk:bg-primary` / `cpk:bg-popover`, so portals rendered outside the chat root are unaffected.
- **CSS-variable scoping is preserved.** The token block (`[data-copilotkit] { --background: ...; --foreground: ...; ... }`) and the dark-mode override (`.dark [data-copilotkit], [data-copilotkit].dark { ... }`) still target `[data-copilotkit]`, so descendants resolving `var(--background)` via their own `bg-background` classes get the correct value in both light and dark modes.
- **Cascade is safe.** The only `:hover`/`:focus` rule in the same file that touches `background` is `[data-copilotkit] ::-webkit-scrollbar-thumb:hover`, which uses a literal `oklch(...)` color, not the token. No interaction with the new rule.

### Completeness of root coverage
Verified `grep -rn 'data-copilotkit' packages/react-core/src/v2/components`. Three categories exist:
1. **Chat root surfaces** — `CopilotChatView` (2 returns), `CopilotSidebarView`, `CopilotPopupView`. All now carry `data-copilotkit-root`. Complete.
2. **`display: contents` wrappers** (e.g. `CopilotChatView:389`, `CopilotChatInput:624`, `CopilotChatMessageView:640`, etc.). These intentionally do not paint anything; correct to leave unmarked.
3. **Visible internal nodes** (`CopilotChatInput:1102`, `CopilotChatMessageView:656`, `CopilotChatUserMessage:211`, `CopilotChatAssistantMessage:191`, `CopilotChatSuggestionView:16`, `CopilotChatSuggestionPill:29`, `CopilotChatReasoningMessage`, `CopilotChatToggleButton:147`, ui `tooltip` and `dropdown-menu` content). Toggle button has `cpk:bg-primary`; tooltip has `cpk:bg-primary`; dropdown has `cpk:bg-popover`; everything else does not need a background. Correct to leave unmarked.

No missing root surface within `react-core/v2`. Suggestion pills, modal header, and the popup's inner content card already declare `cpk:bg-background` themselves, so they continue to paint in default theming without `data-copilotkit-root`.

### Migration impact
- Apps using the default `Sidebar` / `Popup` / `Chat` views: no visible change (the root paints `--background` either way; descendants going transparent stack invisibly against the opaque root).
- Apps with translucent or gradient backdrops: now work without per-element overrides — the documented goal.
- Apps that *depended* on every descendant painting `--background` opaquely on top of a custom backdrop: visual change. The PR body documents the one-line restoration (`[data-copilotkit] { background-color: var(--background); }`). This is reasonable for a fix labeled "scope," but worth calling out in the changeset / release notes so consumers see it before upgrading.

### Lefthook autofix changes
Three benign cleanups appeared in the chat-view files:
- **`import type { ... } from "./CopilotChatView"` / `import type { SlotValue }`** (in `CopilotPopupView.tsx` and `CopilotSidebarView.tsx`): pure type-import separation, identical runtime emit. Standard `verbatimModuleSyntax` / `isolatedModules` fix.
- **`["--sidebar-width" as string]: widthToCss(...)` → `"--sidebar-width": widthToCss(...)`** in `CopilotSidebarView.tsx`: the surrounding object is already cast to `React.CSSProperties`, which accepts CSS custom property keys as plain string literals. The dynamic `[expr as string]` form was redundant. No runtime change, identical emitted CSS, identical type-checking outcome.

All autofixes are safe.

### Suggestions
- **Cross-framework parity.** `packages/vue/src/styles/globals.css:142-145` and `packages/angular/src/styles/globals.css:142-145` still ship the combined rule that paints `var(--background)` on every `[data-copilotkit]` descendant. Vue components similarly nest the attribute (e.g. `CopilotChatMessageView.vue:347`, `CopilotChatInput.vue:952`, `CopilotChatAssistantMessage.vue:744`, plus chat-root, sidebar, and popup view internals). The same regression and same fix apply there. Out of scope for this PR's title, but a follow-up `fix(vue): ...` and `fix(angular): ...` would close the gap.
- **Tests.** Optional but cheap: a CSS-class assertion test that the `CopilotChatView` root carries both `data-copilotkit` and `data-copilotkit-root`, and that `CopilotChatAssistantMessage` carries only `data-copilotkit`, would lock the contract for future refactors. Files in `packages/react-core/src/v2/components/chat/__tests__/` already use that style (`CopilotChatCssClasses.test.tsx`, etc.).
- **Doc / migration note.** Worth a one-line entry in the changeset or `CHANGELOG.md` so consumers who customized translucent themes see the new attribute name as the override anchor. Also a candidate for a brief callout in any "theming" doc page if one exists.

### Risks
- Visible change is limited to consumers who explicitly relied on the descendant-painted background — documented and reversible with one CSS line.
- No bundle, runtime, security, or performance implication; CSS-only and one HTML attribute per root.

BenTaylorDev:
Verified locally: default appearance is unchanged (roots keep `--background`), and internal nodes no longer stack opaque backgrounds over a translucent consumer surface. Complete — every chat root gets `data-copilotkit-root`, and tooltip/dropdown/toggle keep their own backgrounds (utility layer beats the base rule anyway).

Before merge, drop the unrelated churn so the diff is just the fix: the `import type` reflows in `CopilotPopupView`/`CopilotSidebarView`, and the `["--sidebar-width" as string]`→`"--sidebar-width"` cast change. Also drop the `.changeset/` file (`.changeset/` was removed from the repo in `5afa55f06`).

PR #4963, opened 2026-05-21
fix(react-core): show welcome screen on empty thread regardless of threadId source

jerelvelarde:
## Why this matters

Any CopilotKit consumer that mints its own `threadId` on mount — to drive a custom thread drawer, to dodge thread-locked errors on rotation, to sync with a BFF before the user types — currently lands on a half-built first-load experience: suggestion chips float at the top with no greeting, no scaffolding, no welcome state. The welcome screen is suppressed because `hasExplicitThreadId` flips true whenever a threadId prop is passed, even if the thread has zero messages.

Premint-on-mount is the standard pattern for any app doing its own thread management. For a canonical starter, gating the welcome screen on thread origin (rather than on message emptiness) means every developer doing real thread handling has to either work around the gate or accept the half-built look.

image

## How it's implemented

Single-line semantic fix in `packages/react-core/src/v2/components/chat/CopilotChatView.tsx`: the welcome-screen gate drops the `!hasExplicitThreadId` clause. The screen now renders whenever `isEmpty && !welcomeScreenDisabled && !isConnecting` — which matches what the welcome screen is actually *about*: having no messages to display, not the origin of the thread identifier.

## Migration

Consumers who relied on hiding the welcome screen for premint flows can drop their `hasExplicitThreadId={false}` workaround. Consumers who explicitly want to suppress the welcome screen should use `welcomeScreenDisabled` (or equivalent) — which has always been the right escape hatch.

Behavior changes only for the narrow case of "empty thread + explicit threadId" — every other state remains identical.

Vue counterpart `packages/vue/src/v2/components/chat/CopilotChatView.vue:193-199` has the identical `!props.hasExplicitThreadId` welcome-screen gate. Same bug, not fixed by this PR. To be addressed in a follow-up — kept out of this scope to keep the React fix isolated and reviewable.

### DIFF
--- a/.changeset/welcome-screen-on-empty-thread.md
+++ b/.changeset/welcome-screen-on-empty-thread.md
@@ -1,0 +1,5 @@
+---
+"@copilotkit/react-core": patch
+---
+
+Show CopilotChat welcome screen on empty threads regardless of whether threadId was pre-minted. The welcome-screen gate is now purely about message emptiness.
--- a/packages/react-core/src/v2/components/chat/CopilotChatView.tsx
+++ b/packages/react-core/src/v2/components/chat/CopilotChatView.tsx
@@ -94,11 +94,13 @@ export type CopilotChatViewProps = WithSlots<
      */
     isConnecting?: boolean;
     /**
-     * When `true`, the caller has explicitly picked a thread (via `threadId`
-     * prop or `CopilotChatConfigurationProvider`). Suppresses the welcome
-     * screen unconditionally — a caller-managed thread targets a specific
-     * conversation and should render its messages (or an empty panel during
-     * connect) rather than a generic "start a new chat" greeting.
+     * Indicates the caller explicitly picked a thread (via `threadId` prop or
+     * `CopilotChatConfigurationProvider`). Forwarded for compatibility with
+     * downstream consumers reading the flag; no longer gates the welcome
+     * screen, which is now driven purely by message emptiness so empty threads
+     * with a caller-minted UUID still render the welcome state. Use
+     * `welcomeScreen={false}` (or the equivalent slot value) to suppress the
+     * welcome screen explicitly.
      */
     hasExplicitThreadId?: boolean;
     /**
@@ -309,13 +311,14 @@ export function CopilotChatView({
   const isEmpty = messages.length === 0;
   // Type assertion needed because TypeScript doesn't fully propagate `| boolean` through WithSlots
   const welcomeScreenDisabled = (welcomeScreen as unknown) === false;
-  // Suppress the welcome screen (1) while the initial connect is in flight
-  // and (2) whenever the caller has picked a specific thread. The caller-
-  // managed case targets a conversation directly, so the generic welcome
-  // greeting is never the right thing to show — even for a thread that
-  // happens to have no messages yet.
+  // The welcome screen is semantically about having no messages, not about
+  // thread origin. Suppress it only while the initial connect is in flight
+  // (so a resumed thread doesn't flash the generic greeting before its
+  // bootstrap messages arrive) or when the caller explicitly disabled it.
+  // Empty threads with a caller-minted UUID (premint-on-mount, custom thread
+  // drawers, lock-recovery rotation, BFF sync) still get the welcome state.
   const shouldShowWelcomeScreen =
-    isEmpty && !welcomeScreenDisabled && !isConnecting && !hasExplicitThreadId;
+    isEmpty && !welcomeScreenDisabled && !isConnecting;
 
   if (shouldShowWelcomeScreen) {
     // Create a separate input for welcome screen with static positioning and disclaimer visible
--- a/packages/react-core/src/v2/components/chat/__tests__/CopilotChat.welcomeGate.test.tsx
+++ b/packages/react-core/src/v2/components/chat/__tests__/CopilotChat.welcomeGate.test.tsx
@@ -42,16 +42,23 @@ function renderWithKit(ui: React.ReactNode, agent: TrackingAgent) {
 }
 
 /**
- * Regression coverage for fix/welcome-not-showing-at-all.
+ * Regression coverage for fix/welcome-not-showing-at-all and the follow-up
+ * fix that decouples welcome-screen gating from thread origin.
  *
- * The underlying bug: the v1 <CopilotKit> wrapper pipes a ThreadsProvider-
- * minted UUID through to CopilotChatConfigurationProvider as `threadId`.
- * CopilotChat previously treated any non-empty providedThreadId as "caller
- * supplied a real backend thread" and (a) fired /connect (→ 404 for an
- * auto-minted UUID) and (b) suppressed the welcome screen forever. The
- * fix threads an `hasExplicitThreadId` signal through the provider chain;
- * these tests pin the contract that /connect and welcome-screen gating
- * now follow that signal rather than `!!threadId`.
+ * History: the v1 <CopilotKit> wrapper pipes a ThreadsProvider-minted UUID
+ * through to CopilotChatConfigurationProvider as `threadId`. CopilotChat
+ * previously treated any non-empty providedThreadId as "caller supplied a
+ * real backend thread" and (a) fired /connect (→ 404 for an auto-minted UUID)
+ * and (b) suppressed the welcome screen forever. The first fix threaded an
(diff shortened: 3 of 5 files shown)


jerelvelarde:
## Automated review

**Verdict: ship-able with two follow-ups worth noting.** The semantic fix is correct and tightly scoped — `messages.length === 0 && !isConnecting && !welcomeScreenDisabled` is the right gate, the rAF-deferred `isConnecting` release continues to cover the "thread loading history" case, and the JSDoc update on `hasExplicitThreadId` accurately reflects the new contract (forwarded for compatibility, no longer gates welcome). Tests are updated coherently with the new semantics. Two callouts below.

### What the PR does

Single-line change in `CopilotChatView.tsx` (line 320–321): drops `!hasExplicitThreadId` from the welcome-screen gate. The screen now renders whenever `isEmpty && !welcomeScreenDisabled && !isConnecting`. The `hasExplicitThreadId` prop is preserved (still gates `/connect` routing in `CopilotChat.tsx` and prevents welcome-screen flash during thread switch via `isConnecting`). Three existing tests are flipped to assert the new contract, and one JSDoc comment in `CopilotChatConfigurationProvider.test.tsx` is reworded.

### Semantic correctness

- The "messages.length === 0 false-positive on a thread loading history" concern is **adequately handled** by the existing `isConnecting` gate. `CopilotChat.tsx` keeps `isConnecting = hasExplicitThreadId && lastConnectedThreadId !== resolvedThreadId` true through the connect lifecycle, and the rAF in the `finally` block (line 263) defers `setLastConnectedThreadId` one frame past the final React commit. So a thread loading history will not flash the welcome screen — `isConnecting` stays true while messages are en route, and by the time it flips false the messages have committed.
- The narrow window this PR opens is the intended one: an explicitly-picked thread that genuinely has zero messages (premint-on-mount, fresh empty thread on the backend) now shows the welcome screen. That matches the stated intent.

### Test correctness

- The `waitFor(() => expect(screen.getByTestId("copilot-welcome-screen")).toBeDefined())` pattern in `CopilotChat.welcomeGate.test.tsx` is **functionally correct** — `getByTestId` throws when the element is missing, so `waitFor` will retry until the rAF flushes and the screen renders. The `.toBeDefined()` matcher is technically redundant (the throw is what gates the wait), but it doesn't break the test. The idiomatic alternative is `expect(screen.queryByTestId(...)).not.toBeNull()` — style only, not blocking.
- The thread-switch test (lines 156–202) preserves the critical mid-switch assertion at line 189 (`expect(screen.queryByTestId("copilot-welcome-screen")).toBeNull()`) using a **synchronous** `queryByTestId` immediately after the rerender — this still pins the no-flash-during-switch contract. Good.
- `CopilotChatView.connectingGate.test.tsx` is rendered without a `CopilotChat` wrapper, so `isConnecting` is driven purely by props. The "still shows the welcome screen when hasExplicitThreadId=true" test is synchronous and correct.

### `hasExplicitThreadId` consumers — still behave correctly

Audited all four consumers of the flag:

1. **`CopilotChatConfigurationProvider.tsx`** (lines 113–119, 170) — unchanged. Still resolves `hasExplicitThreadId` from prop > inherited parent. Surface preserved.
2. **`CopilotChat.tsx`** (line 216, 224, 284) — `isConnecting` derivation and the `/connect` skip both keep reading the flag. Behavior unchanged for routing.
3. **v1 bridge `copilotkit.tsx`** (line 764) — still pipes `isThreadIdExplicit` down. The bridge regression test at `src/components/copilot-provider/__tests__/v1-explicit-threadid-bridge.test.tsx` covers this surface and was not touched by this PR — still valid.
4. **`CopilotChatView.tsx`** — flag is destructured (line ~110) but only the welcome-screen branch reads it (now removed). No other view-level consumer; the prop is purely forward-compatible from this layer.

### Vue parity gap (worth a follow-up)

`packages/vue/src/v2/components/chat/CopilotChatView.vue` lines 193–199 still have the **identical** old gate:
```ts
const shouldShowWelcomeScreen = computed(
  () =>
    props.messages.length === 0 &&
    props.welcomeScreen !== false &&
    !props.isConnecting &&
    !props.hasExplicitThreadId,
);
```
The Vue welcome-gate test at `packages/vue/src/v2/components/chat/__tests__/CopilotChatView.connectingGate.test.ts` line 48 still asserts "suppresses the welcome screen when hasExplicitThreadId=true". Same bug, same fix — would be worth a parallel Vue PR or a follow-up. Not a blocker for this PR, but mentionable so the Vue port doesn't drift.

### CI `format` failure — not an actual format violation

Worth flagging: the user-facing description called this a "prettier nit," but the failing job is `static / quality / format` which runs **oxfmt** (not prettier), and the failure log shows a **checkout-step error** rather than a format violation:

```
##[error]A branch or tag with the name 'jerel/fix-welcome-screen-on-premint-thread' could not be found
```

The checkout failed before any formatter ran. Local validation confirms zero format issues on all four files:

- `npx prettier --check ...` → "All matched files use Prettier code style!"
- `npx oxfmt@0.36 --check ...` → "All matched files use the correct format. Finished in 17ms on 4 files."

A retry of the `static / quality` workflow should clear this. If oxfmt does flag anything on retry, the workflow auto-fixes for same-repo PRs but **skips the auto-commit for fork PRs** (this PR is from `jerelvelarde/CopilotKit`) — see `.github/workflows/static_quality.yml` lines 137–151. So if a real format issue surfaces post-retry, the author would need to run `npx oxfmt@0.36 --write` locally and push.

### Minor

- `messages.length === 0` is used directly; might be slightly more readable as the existing `isEmpty` named binding (line 311 already has `const isEmpty = messages.length === 0;`). The PR did keep `isEmpty` in the new combined expression — good.
- The JSDoc rewrite on `hasExplicitThreadId` is accurate but verbose. The "Use `welcomeScreen={false}` (or the equivalent slot value)" mention is useful for migrators.

deepakbatham572:
Hi @jerelvelarde , @timcharper  , @ddemaree  , @mnutt  , @dfdumaresq  when this PR will be merge ?

BenTaylorDev:
Traced the connect path: this doesn't reintroduce the `/connect` 404 (connect gating still keys on `hasExplicitThreadId`), and there's no welcome-screen flash on a resuming thread — messages are applied via `connectAgent`'s stream before `setLastConnectedThreadId` releases `isConnecting` (`CopilotChat.tsx:372`→`395`). Behavior is sound.

Note this reverses the earlier deliberate "suppress welcome on explicit threads" decision — reasonable given the premint-on-mount rationale and the `welcomeScreen={false}` opt-out.

One cleanup: `hasExplicitThreadId` is now unused in `CopilotChatView` (only destructured, never read after the gate change) — remove it, or if a downstream consumer really reads it, say where. Also drop the `.changeset/` file (`.changeset/` was removed from the repo in `5afa55f06`).

BenTaylorDev:
Thanks for this, and sorry for the slow follow-up. I'm closing this one — not because the diagnosis was wrong, but because the ground has moved under it since May. Three reasons:

**1. It's red for a real reason.** `unit (20.x)` fails on `CopilotChat.e2e.test.tsx > Suggestions Flow > should stream suggestion titles token by token`. The dumped DOM shows `data-testid="copilot-welcome-screen"` occupying the tree where the suggestion pills should be — dropping the gate lets the welcome screen swallow streamed suggestions on an empty thread. #4960 ran the same suite the same day against the same base and passed, so this is the change rather than a flake.

**2. `main` now covers the premint case with the flag intact.** `CopilotChatConfigurationProvider` has `startNewThread()`, documented as minting "a newly minted UUID with `hasExplicitThreadId=false`, so the welcome screen shows" (`CopilotChatConfigurationProvider.tsx:144`), with a test at `__tests__/CopilotChatConfigurationProvider.test.tsx:680` ("non-explicit set … leaves hasExplicitThreadId=false (welcome path)"). Apps that mint their own UUID get the welcome state by passing `hasExplicitThreadId={false}`, rather than by removing the gate.

**3. Removing the React gate would split the frameworks.** Vue (`CopilotChatView.vue:198`) and Angular (`copilot-chat-view.ts:325`) both still gate the welcome screen on `!hasExplicitThreadId`, and the flag has since become load-bearing elsewhere in React — `use-agent.tsx:457` uses it for threadId resolution. A one-framework behavior change isn't something we can land.

If you're hitting a case that `hasExplicitThreadId={false}` / `startNewThread()` doesn't cover, please open an issue with a repro — that would be a genuine gap and worth fixing. Appreciate the time you put into this.

PR #4965, opened 2026-05-21
fix(react-core): reserve scrollbar gutter on chat scroll container

jerelvelarde:
## Why this matters

In narrow chat layouts — sidebar panels, drawer chats, two-pane apps with a chat column under ~500px — the framework's always-visible 6px scrollbar runs directly through the bottom-right corner of the absolutely-positioned input pill. The scrollbar thumb and the pill's rounded edge occupy the same pixels.

It's a small visual defect with a large reach: any consumer who embeds the chat in anything narrower than a typical full-page layout sees it. For a canonical starter, this is one of the first polish papercuts a developer notices when they move beyond the full-screen demo.

(not the error but the overlap)

image

## How it's implemented

Single-line CSS rule in `packages/react-core/src/v2/styles/globals.css`: `scrollbar-gutter: stable` on the chat scroll container, scoped to a new `data-copilotkit-scroll-container` attribute. The attribute is added to the four scroll surfaces inside `CopilotChatView.tsx`'s `ScrollView` (the `StickToBottom.Content` pin-to-bottom case, the pin-to-send wrapper, the pre-mount fallback, and the `none` mode wrapper) so the rule reaches every autoScroll mode.

`scrollbar-gutter: stable` is the modern CSS primitive for exactly this case — it reserves space for the scrollbar whether the bar is currently visible or not. No JS, no WeakMap-of-timers, no opt-in flag. Works across Chromium, Firefox, and Safari (16.4+).

Scrollbar appearance (width, color, thumb style) is unchanged. Only the gutter reservation is new.

Note: `scrollbar-gutter: stable` reserves the gutter whether or not the scrollbar is currently visible. In wide layouts, this means a ~6px right-side inset is now reserved on the chat scroll container even when no scrollbar would render. Visually negligible in the default theme; consumers who explicitly want zero gutter can override with `scrollbar-gutter: auto`.

## Migration

Pure addition with no opt-out required for the default case. Apps that explicitly want zero gutter reservation (rare — would mean they want the overlap) can override locally with `scrollbar-gutter: auto` on `[data-copilotkit-scroll-container]`.

Vue/Angular counterparts have the same scrollbar-overlap risk in narrow chat layouts — both use `overflow-y-scroll` on the chat scroll container with an absolutely-positioned input pill, and neither `globals.css` reserves a gutter. To be addressed in a follow-up PR — left out of this scope to keep the React fix isolated. Confirmed:

- `packages/vue/src/v2/components/chat/CopilotChatView.vue` (scroll container line 589, absolute input wrapper line 694)
- `packages/angular/src/lib/components/chat/copilot-chat-view-scroll-view.ts` (scroll containers lines 51, 64, 134)
- `packages/angular/src/lib/components/chat/copilot-chat-view-input-container.ts` (absolute input wrapper line 76)
- `packages/vue/src/styles/globals.css` and `packages/angular/src/styles/globals.css` (neither sets `scrollbar-gutter`)

### DIFF
--- a/.changeset/fix-scrollbar-gutter.md
+++ b/.changeset/fix-scrollbar-gutter.md
@@ -1,0 +1,5 @@
+---
+"@copilotkit/react-core": patch
+---
+
+Reserve scrollbar gutter on CopilotChat scroll container via `scrollbar-gutter: stable`, so the always-visible scrollbar no longer overlaps the absolutely-positioned input pill in narrow layouts.
--- a/packages/react-core/src/v2/components/chat/CopilotChatView.tsx
+++ b/packages/react-core/src/v2/components/chat/CopilotChatView.tsx
@@ -471,6 +471,7 @@ export namespace CopilotChatView {
       <ScrollElementContext.Provider value={scrollEl}>
         <>
           <StickToBottom.Content
+            data-copilotkit-scroll-container
             className="cpk:overflow-y-auto cpk:overflow-x-hidden"
             style={{ flex: "1 1 0%", minHeight: 0 }}
           >
@@ -560,6 +561,7 @@ export namespace CopilotChatView {
         >
           <div
             ref={nonAutoScrollRefCallback}
+            data-copilotkit-scroll-container
             className="cpk:flex-1 cpk:min-h-0 cpk:overflow-y-auto cpk:overflow-x-hidden"
             {...props}
           >
@@ -685,7 +687,10 @@ export namespace CopilotChatView {
 
     if (!hasMounted) {
       return (
-        <div className="cpk:h-full cpk:max-h-full cpk:flex cpk:flex-col cpk:min-h-0 cpk:overflow-y-auto cpk:overflow-x-hidden">
+        <div
+          data-copilotkit-scroll-container
+          className="cpk:h-full cpk:max-h-full cpk:flex cpk:flex-col cpk:min-h-0 cpk:overflow-y-auto cpk:overflow-x-hidden"
+        >
           <div className="cpk:px-4 cpk:sm:px-0 cpk:[div[data-sidebar-chat]_&]:px-8 cpk:[div[data-popup-chat]_&]:px-6">
             {children}
           </div>
@@ -702,6 +707,7 @@ export namespace CopilotChatView {
         <ScrollElementContext.Provider value={nonAutoScrollEl}>
           <div
             ref={nonAutoScrollRefCallback}
+            data-copilotkit-scroll-container
             className={cn(
               "cpk:h-full cpk:max-h-full cpk:flex cpk:flex-col cpk:min-h-0 cpk:overflow-y-auto cpk:overflow-x-hidden cpk:relative",
               className,
--- a/packages/react-core/src/v2/styles/globals.css
+++ b/packages/react-core/src/v2/styles/globals.css
@@ -201,6 +201,14 @@
   scrollbar-color: oklch(0.5 0 0) transparent;
 }
 
+/* Reserve scrollbar gutter on chat scroll containers so the always-visible
+ * scrollbar can't cut through the absolutely-positioned input pill in narrow
+ * layouts. `stable` keeps the gutter reserved whether or not the scrollbar is
+ * currently visible, so content doesn't reflow when overflow toggles. */
+[data-copilotkit] [data-copilotkit-scroll-container] {
+  scrollbar-gutter: stable;
+}
+
 [data-copilotkit] .prose input[type="checkbox"] {
   appearance: none;
   background-color: #fff;


jerelvelarde:
## Automated review

**Verdict:** Looks good — narrow scope, correct CSS primitive for the job, complete coverage of the four runtime scroll surfaces. One small behavior-change note worth flagging in the PR body, no blockers.

### Scope and correctness

The diff is 15 additions / 1 deletion across two files:

- `packages/react-core/src/v2/styles/globals.css` — one new rule `[data-copilotkit] [data-copilotkit-scroll-container] { scrollbar-gutter: stable; }` with a multi-line comment explaining the why.
- `packages/react-core/src/v2/components/chat/CopilotChatView.tsx` — adds `data-copilotkit-scroll-container` to four scroll surfaces (lines 474, 564, 691, 710).

`scrollbar-gutter: stable` is the right primitive — it's the W3C-spec mechanism for exactly this case (reserve gutter even when bar isn't visible, avoid content reflow when overflow toggles). No JS, no opt-in flag, no perf cost.

### Scroll-surface coverage (verified)

`grep -n "overflow" packages/react-core/src/v2/components/chat/CopilotChatView.tsx` returns four elements with `overflow-y-auto`:

| Line | Element                                      | Mode               | Has attribute |
|------|----------------------------------------------|--------------------|---------------|
| 475  | `StickToBottom.Content`                      | pin-to-bottom      | yes (474)     |
| 565  | `PinToSendScrollContainer` inner `<div>`     | pin-to-send        | yes (564)     |
| 692  | Pre-mount fallback `<div>`                   | hydration          | yes (691)     |
| 712  | `none`-mode wrapper `<div>`                  | manual             | yes (710)     |

Coverage is complete. The only other `overflow-y-auto` in `react-core/src/v2/components/chat/` is the mention-list popover at `CopilotChatInput.tsx:972` (`max-h-64 overflow-y-auto`) and the dropdown content at `ui/dropdown-menu.tsx:46`. Both are bounded popovers, not chat scroll surfaces — correctly excluded. No false negatives, no false positives.

### Browser support

`scrollbar-gutter: stable` requires Chromium 94+ (Sep 2021), Firefox 97+ (Feb 2022), Safari 16.4+ (Mar 2023).

- No `.browserslistrc` or `browserslist` field in `packages/react-core/package.json`.
- The file already uses `oklch()` for scrollbar-thumb colors at line 191 (Safari 15.4+, Chrome 111+, Firefox 113+) — a higher floor than `scrollbar-gutter: stable`. So this PR doesn't raise the de-facto browser support bar at all; in older browsers the property is silently ignored, which is the desired progressive-enhancement behavior (the input-pill papercut returns, nothing else breaks).

### Visual impact in non-narrow layouts

Worth being explicit: `stable` reserves the 6px gutter whether the scrollbar is currently visible or not. In wide layouts where content doesn't overflow, that's a 6px inset on the right that wasn't there before. This is the correct trade-off (and matches the comment's "doesn't reflow when overflow toggles" rationale), but the PR body's "Scrollbar appearance is unchanged" framing is slightly understated — the gutter reservation itself *is* a visible change in non-overflowing wide layouts. Not a blocker, just worth a sentence in the PR description so reviewers don't expect zero visual delta.

If a no-gutter-when-not-overflowing variant is ever desired, `stable both-edges` or `auto` could be exposed as a token, but `stable` is the right default.

### Attribute naming consistency

Consistent with the `data-copilotkit-<role>` pattern that PR #4962 (still open) introduces with `data-copilotkit-root`. Both PRs nest scoping under `[data-copilotkit]` parent selector, matching the surrounding rule style in `globals.css`. If both land, the namespace stays coherent: `data-copilotkit` (descendants for variable scoping) / `data-copilotkit-root` (top-level chat container) / `data-copilotkit-scroll-container` (scroll surfaces).

### Other notes

- The descendant combinator `[data-copilotkit] [data-copilotkit-scroll-container]` requires the scroll container to be inside a `[data-copilotkit]` ancestor. All four sites render inside `CopilotChatView`'s `<div data-copilotkit>` wrapper at line 388, so scoping holds in every runtime path.
- Tests in `CopilotChatView.slots.e2e.test.tsx` query `.overflow-y-auto` — unaffected since className strings are unchanged.
- No slot API, prop, or layout-class change. Theming surface unchanged for consumers.
- CSS comment is well-written and motivates the `stable` choice over `auto`. Keep it.

### Suggested (optional, non-blocking)

- Add one sentence to the PR body acknowledging the 6px gutter is now reserved in wide layouts even when the scrollbar isn't visible — sets expectations for visual-regression-test runs and for screenshot-based downstream consumers.
- Consider a Playwright/Vitest visual snapshot for the narrow-layout case (≤500px chat column) as a regression guard, since the whole motivation is a layout-width-dependent papercut.

BenTaylorDev:
Approach is sound and complete (all 4 chat scroll containers in `CopilotChatView` are tagged; the slash-menu popover is correctly left out). It's currently CONFLICTING, but only mechanically: the attribute was added next to a class that changed on `main` (`sm:px-0`→`@3xl:px-0`), and the globals.css insertion point moved. Please rebase onto current `main` and re-anchor the hunks. Also drop the `.changeset/` file (`.changeset/` was removed from the repo in `5afa55f06`).

PR #4969, opened 2026-05-21
fix(react-core): propagate resolved threadId in legacy chat

### LINKED ISSUE #4943: existing-thread hydration fails in legacy CopilotPopup when /connect replays historical runs
### ♻️ Reproduction Steps

1. Use the legacy CopilotPopup / useCopilotChatInternal path with an explicit threadId supplied by the app, for example a persisted cookie-backed thread id.
2. Open a conversation, send messages, and keep using the same explicit thread id across reloads.
3. Ensure the runtime's /connect path replays historical events for that thread, including at least one earlier run that ended in RUN_ERROR or multiple past runs in the same replay stream.
4. Reload the page and let CopilotKit reconnect and hydrate the existing thread.

Observed while tracing 1.57.1:
- the legacy chat path does not consistently copy the resolved threadId onto the AG-UI agent before connect, run, and reload
- after forcing the same thread id through, HTTP connectAgent still fails during replay because it goes through the default AG-UI verifyEvents lifecycle

### ✅ Expected Behavior

When an app provides an explicit threadId, CopilotKit should reuse that thread consistently for connect, run, and reload flows.

On reload, /connect should hydrate the existing conversation history successfully, even if the replayed event stream contains events from multiple past runs or includes an earlier RUN_ERROR.

### ❌ Actual Behavior

Two related failures occur in the legacy popup flow.

1. The resolved threadId from CopilotKit is not always copied onto the underlying AG-UI agent before connect, run, and reload, so downstream runtime requests can use a new thread id instead of the caller-provided one.

2. Even when the explicit thread id is propagated correctly, HTTP connectAgent fails on hydration if the replay stream contains an earlier RUN_ERROR followed by a later RUN_STARTED. The stream is treated like a single fresh AG-UI run and AG-UI throws:

Cannot send event type 'RUN_STARTED': The run has already errored with 'RUN_ERROR'. No further events can be sent.

This breaks agent_connect_failed handling and prevents message hydration on reload.

### Screenshots

_No response_

### 𝌚 CopilotKit Version

```shell
@copilotkit/react-core 1.57.1
@copilotkit/react-ui 1.57.1
@copilotkit/runtime 1.57.1
transitive @copilotkit/core 1.57.1

App also had:
@ag-ui/client 0.0.42 (direct)
Transitive CopilotKit dependencies resolved @ag-ui 0.0.53 internally
```

### 📄 Logs (Optional)

```shell
Browser/devtools:
Cannot send event type 'RUN_STARTED': The run has already errored with 'RUN_ERROR'. No further events can be sent.

User-facing effect:
- agent_connect_failed on reload
- existing thread does not hydrate prior chat messages

Relevant source areas:
- packages/core/src/agent.ts
- packages/react-core/src/hooks/use-copilot-chat_internal.ts
```

anujshah3:
## What does this PR do?

This PR now scopes the fix to the legacy React chat path only.

Specifically, it:
- ensures the legacy `use-copilot-chat_internal` flow assigns the resolved `threadId` onto the underlying agent before connect, reload, and follow-up run paths
- adds focused regression coverage for thread-id propagation in the legacy React flow

## Why was the core-side change removed?

The original PR also included a core-side change to make proxied HTTP `connectAgent()` tolerant of replay and hydration streams that can include prior run events.

That core issue was relevant to the behavior I saw on `@copilotkit/core` `1.57.1`, but after rechecking against `1.57.3`, that core-side behavior is already resolved in the published package. Because of that, this PR no longer proposes the `packages/core/src/agent.ts` change and is limited to the remaining React-side gap.

## Motivation

The remaining issue is in the legacy React chat path: when reconnecting to an existing thread, the resolved `threadId` was not consistently assigned to the agent before connect and follow-up run paths, which could break thread continuity during hydration/reconnect flows.

## Related PRs and Issues

- Fixes https://github.com/CopilotKit/CopilotKit/issues/4943

## Checklist

### DIFF
--- a/packages/core/src/__tests__/proxied-runtime-transport.test.ts
+++ b/packages/core/src/__tests__/proxied-runtime-transport.test.ts
@@ -130,6 +130,22 @@ describe("ProxiedCopilotRuntimeAgent transport integration", () => {
         expect(headers.get("accept")).toBe("text/event-stream");
       });
 
+      it("hydrates replay streams that include a prior run error", async () => {
+        const agentId = "connect-agent";
+        const agent = new ProxiedCopilotRuntimeAgent({
+          runtimeUrl,
+          agentId,
+          headers: { Authorization: "Bearer test-token" },
+          transport,
+        });
+
+        fetchMock.mockResolvedValueOnce(createReplaySseResponse());
+
+        await expect(agent.connectAgent({})).resolves.toMatchObject({
+          newMessages: expect.any(Array),
+        });
+      });
+
       it("sends stop requests with the expected payload", () => {
         const agentId = "stop-agent";
         const threadId = "thread-123";
@@ -348,6 +364,42 @@ function createSseResponse(): Response {
           type: "RUN_FINISHED",
           threadId: "test-thread",
           runId: "test-run",
+          result: { newMessages: [] },
+        },
+      ];
+      const payload = events
+        .map((event) => `data: ${JSON.stringify(event)}\n\n`)
+        .join("");
+      controller.enqueue(encoder.encode(payload));
+      controller.close();
+    },
+  });
+
+  return new Response(stream, {
+    status: 200,
+    headers: { "content-type": "text/event-stream" },
+  });
+}
+
+function createReplaySseResponse(): Response {
+  const stream = new ReadableStream({
+    start(controller) {
+      const events = [
+        {
+          type: "RUN_ERROR",
+          threadId: "test-thread",
+          runId: "failed-run",
+          message: "previous run failed",
+        },
+        {
+          type: "RUN_STARTED",
+          threadId: "test-thread",
+          runId: "next-run",
+        },
+        {
+          type: "RUN_FINISHED",
+          threadId: "test-thread",
+          runId: "next-run",
           result: { newMessages: [] },
         },
       ];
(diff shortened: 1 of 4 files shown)


anujshah3:
Since the build is failing on main, wasn't able to complete the build locally. But I tested the changed on my project and it does work as expected

BenTaylorDev:
Thanks for this. The underlying issue is real, but this fix has since been superseded on `main` and unfortunately can't be merged as-is:

- The resolved-threadId propagation already landed centrally via #5041 (`d1506ec66`) in `use-agent.tsx` — `agent.threadId = configThreadId` runs in an effect before connect/reload/append fire, and the legacy `useCopilotChatInternal` hook consumes that same `useAgent`, so the legacy `CopilotPopup` path is already covered.
- More importantly, this PR assigns `resolvedThreadId = existingConfig?.threadId ?? threadId` **unconditionally**, whereas #5041 deliberately gates on `hasExplicitThreadId` — precisely to prevent a `ThreadsProvider`-minted *placeholder* UUID from overwriting the agent's stable UUID. So merging this would reintroduce the exact overwrite that guard exists to stop.

Closing as superseded. Worth a 30-second confirm on current `main` (open an existing thread in legacy `CopilotPopup`, check `/agent/connect` carries the resolved threadId), after which #4943 can likely be closed too. Appreciate the contribution.

PR #4971, opened 2026-05-21
fix: gate thread requests on runtime capabilities

MikeRyanDev:
## Root cause

`useThreads` and the web inspector treated a connected runtime as if thread list/inspect endpoints were always available. In non-Intelligence setups, that could trigger `/threads` and inspector requests against runtimes that did not expose compatible thread endpoints, producing noisy benign 404s.

## Behavior change

- Adds `threadEndpoints` capability metadata to runtime info and carries it through shared/core types.
- Reports thread list/inspect support only for multi-route runtimes backed by Intelligence or the local `InMemoryAgentRunner`; single-route runtime info reports thread endpoints as unavailable.
- Gates `useThreads` so it does not fetch when the runtime does not advertise thread list support, and returns an explicit unavailable-endpoints error instead.
- Gates web inspector owned thread stores and thread detail fetches on advertised list/inspect support.

## Docs

- Clarifies managed Intelligence thread metadata/history versus self-managed runner persistence.
- Documents that `AgentRunner`, `SqliteAgentRunner`, or a custom runner can persist and replay chat history, but does not automatically provide the managed `useThreads` list/mutation/realtime contract unless compatible REST thread endpoints are exposed.
- Updates inspector/event-inspector messaging to avoid implying thread browsing is Intelligence-only while still distinguishing durable managed history from local in-memory development support.

## Verification

Publish-pass checks:

- `gh --version`
- `gh auth status`
- `git diff --check`
- `git diff --cached --check`
- Commit hook ran `lint-fix` with warnings only, then package checks via Nx:
  - `nx run-many -t test --projects=packages/**`
  - `nx run-many -t publint,attw --projects=packages/**`

Implementation-agent reported checks:

- web-inspector tests passed
- core tests passed
- react-core tests passed
- runtime tests passed
- sqlite-runner tests passed
- `git diff --check` passed

Known limitation:

- Full `run-many check-types` was attempted by the implementation agent and failed due pre-existing dependency/typecheck issues and runtime OOM, not this FAC-1 change.

### DIFF
--- a/packages/react-core/src/v2/hooks/use-threads.tsx
+++ b/packages/react-core/src/v2/hooks/use-threads.tsx
@@ -7,9 +7,8 @@ import {
   ɵselectThreadsIsLoading,
   ɵselectHasNextPage,
   ɵselectIsFetchingNextPage,
-  type ɵThreadRuntimeContext,
-  type ɵThreadStore,
 } from "@copilotkit/core";
+import type { ɵThreadRuntimeContext, ɵThreadStore } from "@copilotkit/core";
 import {
   useCallback,
   useEffect,
@@ -213,6 +212,12 @@ export function useThreads({
       ),
     );
   }, [copilotkit.headers]);
+  const runtimeStatus = copilotkit.runtimeConnectionStatus;
+  const threadEndpointsSupported = copilotkit.threadEndpoints?.list === true;
+  const threadEndpointsUnavailable =
+    !!copilotkit.runtimeUrl &&
+    runtimeStatus === CopilotKitCoreRuntimeConnectionStatus.Connected &&
+    !threadEndpointsSupported;
   const runtimeError = useMemo(() => {
     if (copilotkit.runtimeUrl) {
       return null;
@@ -220,6 +225,15 @@ export function useThreads({
 
     return new Error("Runtime URL is not configured");
   }, [copilotkit.runtimeUrl]);
+  const threadEndpointsError = useMemo(() => {
+    if (!threadEndpointsUnavailable) {
+      return null;
+    }
+
+    return new Error(
+      "Thread endpoints are not available on this CopilotKit runtime",
+    );
+  }, [threadEndpointsUnavailable]);
 
   // Tracks whether we've dispatched the first real context to the store.
   // The store itself starts with `isLoading: false`, so before we dispatch
@@ -229,10 +243,16 @@ export function useThreads({
   // the first fetch is in flight (at which point the store's own
   // isLoading takes over).
   const [hasDispatchedContext, setHasDispatchedContext] = useState(false);
-  const preConnectLoading = !!copilotkit.runtimeUrl && !hasDispatchedContext;
+  const preConnectLoading =
+    !!copilotkit.runtimeUrl &&
+    !threadEndpointsUnavailable &&
+    !hasDispatchedContext;
 
-  const isLoading = runtimeError ? false : preConnectLoading || storeIsLoading;
-  const error = runtimeError ?? storeError;
+  const isLoading =
+    runtimeError || threadEndpointsError
+      ? false
+      : preConnectLoading || storeIsLoading;
+  const error = runtimeError ?? threadEndpointsError ?? storeError;
 
   useEffect(() => {
     store.start();
@@ -252,7 +272,6 @@ export function useThreads({
   // leave the previously-dispatched context in place — any in-flight
   // realtime subscription or cached thread list stays usable while the
   // runtime recovers, and we don't re-trigger a fetch storm on transitions.
-  const runtimeStatus = copilotkit.runtimeConnectionStatus;
   useEffect(() => {
     copilotkit.registerThreadStore(agentId, store);
     return () => {
@@ -263,12 +282,19 @@ export function useThreads({
   useEffect(() => {
     if (!copilotkit.runtimeUrl) {
       store.setContext(null);
+      setHasDispatchedContext(false);
       return;
     }
 
     // Wait for /info to land so we can include `wsUrl` in the initial
     // context and avoid a redundant second list fetch.
     if (runtimeStatus !== CopilotKitCoreRuntimeConnectionStatus.Connected) {
+      return;
+    }
+
+    if (!threadEndpointsSupported) {
+      store.setContext(null);
+      setHasDispatchedContext(false);
       return;
     }
 
@@ -289,6 +315,7 @@ export function useThreads({
     runtimeStatus,
     headersKey,
     copilotkit.intelligence?.wsUrl,
+    threadEndpointsSupported,
     agentId,
     includeArchived,
     limit,
--- a/packages/react-core/src/v2/hooks/use-threads.tsx
+++ b/packages/react-core/src/v2/hooks/use-threads.tsx
@@ -213,9 +213,10 @@ export function useThreads({
     );
   }, [copilotkit.headers]);
   const runtimeStatus = copilotkit.runtimeConnectionStatus;
-  const threadListEndpointSupported = copilotkit.threadEndpoints?.list === true;
+  const threadListEndpointSupported =
+    copilotkit.threadEndpoints?.list !== false;
   const threadMutationsSupported =
-    copilotkit.threadEndpoints?.mutations === true;
+    copilotkit.threadEndpoints?.mutations !== false;
   const threadEndpointsUnavailable =
     !!copilotkit.runtimeUrl &&
     runtimeStatus === CopilotKitCoreRuntimeConnectionStatus.Connected &&
--- a/packages/react-core/src/v2/hooks/use-threads.tsx
+++ b/packages/react-core/src/v2/hooks/use-threads.tsx
@@ -213,10 +213,9 @@ export function useThreads({
     );
   }, [copilotkit.headers]);
   const runtimeStatus = copilotkit.runtimeConnectionStatus;
-  const threadListEndpointSupported =
-    copilotkit.threadEndpoints?.list !== false;
+  const threadListEndpointSupported = copilotkit.threadEndpoints?.list === true;
   const threadMutationsSupported =
-    copilotkit.threadEndpoints?.mutations !== false;
+    copilotkit.threadEndpoints?.mutations === true;
   const threadEndpointsUnavailable =
     !!copilotkit.runtimeUrl &&
     runtimeStatus === CopilotKitCoreRuntimeConnectionStatus.Connected &&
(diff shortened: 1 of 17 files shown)


packages/react-core/src/v2/hooks/use-threads.tsx
```diff
@@ -213,13 +212,28 @@ export function useThreads({
       ),
     );
   }, [copilotkit.headers]);
+  const runtimeStatus = copilotkit.runtimeConnectionStatus;
```
marthakelly, line 215: When a user upgrades the client SDK but not the server, `threadEndpoints` is `undefined` in `/info`, so `useThreads` now errors where it previously would have tried (and possibly succeeded). Consider treating `undefined` as "unknown, try anyway" or making the error message hint at a version mismatch.

packages/runtime/src/v2/runtime/handlers/get-runtime-info.ts
```diff
@@ -99,3 +106,22 @@ export async function handleGetRuntimeInfo({
     );
   }
 }
+
+function resolveThreadEndpointInfo(
+  runtime: CopilotRuntimeLike,
+  threadEndpointsEnabled: boolean,
+): RuntimeInfo["threadEndpoints"] {
+  const hasRestThreadBackend =
+    isIntelligenceRuntime(runtime) ||
```
marthakelly, line 136: `instanceof` can silently return `false` across duplicate module copies (hoisting, symlinks, version mismatches in the monorepo). The codebase already has a cleaner pattern with `isIntelligenceRuntime()` using property-based checks. A similar approach here (e.g. a `supportsThreadPersistence` flag on the runner) would be more robust and easier to extend.

packages/web-inspector/src/index.ts
```diff
@@ -701,6 +702,7 @@ export class ɵCpkThreadDetails extends LitElement {
   thread: ɵThread | null = null;
   runtimeUrl = "";
   headers: Record<string, string> = {};
+  threadInspectionAvailable = true;
```
marthakelly: If a future render path forgets to pass this prop, the component will attempt the very 404s this PR eliminates. A default of `false` would be safer.

packages/react-core/src/v2/hooks/use-threads.tsx
```diff
@@ -213,13 +212,28 @@ export function useThreads({
       ),
     );
   }, [copilotkit.headers]);
```
marthakelly, line 214: The type distinguishes `list`/`inspect` from `mutations`, but the hook only checks `.list` and unconditionally exposes `renameThread`/`archiveThread`/`deleteThread`. For InMemory runners those will fail server-side with no client-side signal.

packages/runtime/src/v2/runtime/handlers/get-runtime-info.ts
```diff
@@ -99,3 +106,22 @@ export async function handleGetRuntimeInfo({
...
+    runtime.runner instanceof InMemoryAgentRunner;
+  const restEndpointsAvailable = threadEndpointsEnabled && hasRestThreadBackend;
```
marthakelly, line 138: Only the InMemory branch of this function is tested. The Intelligence branch is the only one that sets `mutations: true` and `realtimeMetadata: true`, so a regression here would silently downgrade Intelligence users to read-only.

packages/react-core/src/v2/hooks/use-threads.tsx
MikeRyanDev: In a101c8332, missing capability metadata now preserves legacy/unknown behavior: undefined threadEndpoints keeps the old try-the-request path, and only explicit list: false blocks listing. I also added/updated regression coverage for undefined capabilities.

packages/react-core/src/v2/hooks/use-threads.tsx
MikeRyanDev: In a101c8332, rename/archive/delete now gate on threadEndpoints.mutations and reject client-side when mutations is false, without calling the mutation endpoints. Added tests for the gated behavior.

packages/runtime/src/v2/runtime/handlers/get-runtime-info.ts
MikeRyanDev: In a101c8332, runtime info now reports thread support from runner capability metadata/property-based detection. InMemory declares support directly, avoiding duplicate-module instanceof fragility.

packages/runtime/src/v2/runtime/handlers/get-runtime-info.ts
MikeRyanDev: In a101c8332, I added regression coverage proving the Intelligence multi-route path reports list/inspect/mutations/realtimeMetadata as true.

packages/web-inspector/src/index.ts
MikeRyanDev: In a101c8332, threadInspectionAvailable now defaults to false, and the regression test verifies an omitted prop does not fetch messages/events/state.

marthakelly:
# Review: PR #4971 — fix: gate thread requests on runtime capabilities

## Summary

This PR adds a `threadEndpoints` capability object to the runtime `/info` response, then gates `useThreads` fetches, mutation calls, and web inspector thread stores/detail fetches on those advertised capabilities. It also replaces fragile `instanceof InMemoryAgentRunner` checks with a `supportsLocalThreadEndpoints()` type guard. The overall quality is high. No correctness bugs were found. The main review concern is an inconsistency in how `undefined` (legacy runtime) is interpreted between `useThreads` and the web inspector, plus two missing test branches in `resolveThreadEndpointInfo`.

---

## Findings

### Correctness

None found. The state machine in `useThreads`, the lifecycle in the web inspector, the type guard, and the single-route vs multi-route distinction are all correct. Backward compatibility with legacy runtimes (no `threadEndpoints` field) is properly handled.

### Design Quality

**`packages/react-core/src/v2/hooks/use-threads.tsx:217` vs `packages/web-inspector/src/index.ts:2590`** — `useThreads` checks `copilotkit.threadEndpoints?.list !== false` (treats `undefined` as supported for legacy compat), while the web inspector checks `core.threadEndpoints?.list === true` (treats `undefined` as unsupported). These two components serve the same user-facing concern (thread list display) but use opposite polarity for the same `undefined` value. Consider a shared helper like `isThreadListAvailable(endpoints)` that encodes the legacy fallback rule in one place, to prevent a future contributor from picking the wrong polarity when adding a third consumer. At minimum, a comment at both call sites explaining the intentional divergence would help.

**`packages/shared/src/utils/types.ts:34`** — `ThreadEndpointRuntimeInfo` has four independent booleans, but only three combinations are valid today: all-false, list+inspect only, or all-true. `resolveThreadEndpointInfo` always sets `list === inspect` and `mutations === realtimeMetadata`, so these are really two independent bits, not four. A discriminated union (e.g. `{ level: "none" | "readonly" | "full" }`) would make impossible states unrepresentable. Counter-argument: flat booleans are easier to evolve if capabilities genuinely become independent later. Worth discussing before the wire format solidifies.

**`packages/runtime/src/v2/runtime/handlers/get-runtime-info.ts:25`** — `threadEndpointsEnabled` is a routing topology question ("are thread URLs mounted?") leaking into a domain handler as a boolean parameter. The fetch-handler knows single vs multi-route mode, but instead of letting the info handler derive this from context, the caller threads a boolean through `dispatchRoute`. The default value `true` is also risky: if someone calls `handleGetRuntimeInfo` from a new integration path and forgets the parameter, the default silently claims thread endpoints exist.

**`packages/react-core/src/v2/hooks/use-threads.tsx:366`** — `fetchMoreThreads` is not gated, unlike the three mutation methods. If the thread list endpoint is unavailable, calling `fetchMoreThreads` hits the store (which likely no-ops), while calling `renameThread` gets an immediate descriptive rejection. The asymmetric failure modes could surprise a consumer.

### Simplification

**`packages/react-core/src/v2/hooks/use-threads.tsx:330-365`** — `renameThread`, `archiveThread`, and `deleteThread` each repeat the identical 4-line "if threadMutationsError, reject" guard before delegating to the store. A single `guardMutation` wrapper would collapse ~30 lines into ~10 and ensure a fourth mutation doesn't forget the guard.

**`packages/react-core/src/v2/hooks/use-threads.tsx:220-241`** — `threadEndpointsUnavailable` is a double-negative (true when `threadListEndpointSupported` is false), and the three separate `useMemo` error objects (`runtimeError`, `threadEndpointsError`, `threadMutationsError`) follow the same pattern. Inverting the name to something like `threadListReady` and merging the first two memos would reduce the negation layers.

**`packages/web-inspector/src/index.ts:1461,1499,1546`** — All three fetch methods (`fetchMessages`, `fetchEvents`, `fetchState`) now share an identical `if (!this.runtimeUrl || !this.threadInspectionAvailable)` early-return guard. A private `canFetchThreadDetail()` predicate would make the precondition self-documenting and ensure future guard additions update in one place.

### Cast & Type Quality

No `as any` casts anywhere in the diff. Clean.

**`packages/runtime/src/v2/runtime/__tests__/get-runtime-info.test.ts` (Intelligence test)** — The override bag includes Intelligence-specific fields (`intelligence`, `identifyUser`, `generateThreadNames`, `lockTtlSeconds`) but is cast as `as Partial<CopilotRuntimeLike>`. This launders the structural mismatch. If the Intelligence runtime shape changes, this test won't produce a compile error. A `as Partial<CopilotIntelligenceRuntimeLike>` or dedicated helper would be more truthful.

**`packages/runtime/src/v2/runtime/handlers/get-runtime-info.ts:109`** — `resolveThreadEndpointInfo` returns `RuntimeInfo["threadEndpoints"]` which resolves to `ThreadEndpointRuntimeInfo | undefined`, but the function always returns a concrete object. Annotating the return type as `ThreadEndpointRuntimeInfo` (non-optional) would be more precise.

### Test Coverage

**Consider adding before merge:**

1. **`resolveThreadEndpointInfo` with `threadEndpointsEnabled: false`** — Both new tests in `get-runtime-info.test.ts` pass `threadEndpointsEnabled: true`. The `false` path (single-route mode) is only covered indirectly through `fetch-handler.test.ts`. A direct unit test calling `handleGetRuntimeInfo({ ..., threadEndpointsEnabled: false })` with a runner that supports local thread endpoints would isolate this branch and prevent a regression that silently leaks thread capabilities to single-route runtimes.

2. **`resolveThreadEndpointInfo` with a plain runner and `threadEndpointsEnabled: true`** — No test covers a multi-route runtime where the runner is a vanilla `AgentRunner` (not InMemory, not Intelligence). This is the path for production runtimes with custom runners. A missing test here means a custom runner could accidentally get thread endpoints advertised.

**Worth adding:**

- `supportsLocalThreadEndpoints()` type guard: a dedicated unit test for the negative case (runner without the flag returns false).
- Web inspector `ensureOwnedThreadStore`: no test for the `threadEndpoints.list !== true` early return.
- `useThreads` transition test: supported -> unsupported thread endpoints. The cleanup path (`setHasDispatchedContext(false)`, `store.setContext(null)`) exists in the code but is untested.

---

## Highest-Priority Findings

1. **Inconsistent `undefined` interpretation** (`use-threads.tsx:217` vs `web-inspector/src/index.ts:2590`). A legacy runtime that doesn't send `threadEndpoints` will have threads fetched by `useThreads` but blocked by the inspector. A shared helper or normalization at the source would fix this.

2. **Two missing test branches for `resolveThreadEndpointInfo`**. The `threadEndpointsEnabled: false` path and the "plain runner with endpoints enabled" path are both untested at the unit level. These are the core decision branches of the PR's purpose.

3. **Triplicated mutation guard** (`use-threads.tsx:330-365`). Identical reject-if-unsupported code in all three mutation callbacks. Low risk today, but maintenance surface for the next person adding a mutation.

marthakelly (dismissed):

samjulien (approved):

PR #4983, opened 2026-05-22
fix: add Sentinel CI workflow for workflow security scanning

jpr5:
## Summary

Adds Sentinel ([jpr5/sentinel](https://github.com/jpr5/sentinel)) as a GitHub Action that scans `.github/workflows/*.yml` for CI/CD security issues on every PR and push-to-main.

Part of the [org-wide sentinel rollout](https://www.notion.so/copilotkit/3683aa381852818bacd8e14eb7233c22). Currently in **warn-only mode** (`fail-on-findings: false`) — findings surface as PR annotations without blocking merges. Will be promoted to blocking after 7 days of clean runs.

Rule set: 32 rules covering shell injection, dangerous triggers, hardcoded secrets, github-script injection, ai-config injection, and others.

## Test plan

- [ ] CI green on this PR
- [ ] Sentinel action runs and reports findings as annotations (if any) without failing the check
- [ ] If clean for 7 days, promote to `fail-on-findings: true` and add as required check

### DIFF
--- a/.github/workflows/publish-commit.yml
+++ b/.github/workflows/publish-commit.yml
@@ -30,9 +30,10 @@ jobs:
       pull-requests: write
 
     steps:
-      # persist-credentials required: pkg-pr-new uses repo token to post snapshot comments on the PR
       - name: Checkout code
         uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+        with:
+          persist-credentials: true
 
       - name: Install pnpm
         uses: pnpm/action-setup@0e279bb959325dab635dd2c09392533439d90093 # v6.0.8
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -63,12 +63,12 @@ jobs:
           echo "scope=$SCOPE" >> $GITHUB_OUTPUT
           echo "Detected scope: $SCOPE"
 
-      # No token/credential persistence: the publish job sets up its own
-      # `git config insteadOf` with secrets.GITHUB_TOKEN before pushing tags,
-      # so this checkout doesn't need write access. Critically, the
-      # subsequent `Upload workspace` step packs the entire checkout
-      # (including .git/config) into an artifact — persisting credentials
-      # here would leak a workflow-scoped token to anyone with actions:read.
+      # No token/credential persistence: the publish job uses per-step
+      # GIT_CONFIG_* env vars (scoped to the tag-push step only) instead of
+      # persisting credentials in .git/config. Critically, the subsequent
+      # `Upload workspace` step packs the entire checkout (including
+      # .git/config) into an artifact — persisting credentials here would
+      # leak a workflow-scoped token to anyone with actions:read.
       - name: Checkout Repo
         uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
         with:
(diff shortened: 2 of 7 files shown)


jpr5:
Force-pushed with SHA-pinned actions to satisfy zizmor's `unpinned-uses` rule. Per the org-wide sentinel rollout, the floating-tag-for-easy-updates model is being replaced with SHA-pin + dependabot. Spec updated. Rest of the PR unchanged.

jpr5:
Bumped to v1.3.1 SHA — picks up the warn-only fix ([jpr5/sentinel#34](https://github.com/jpr5/sentinel/pull/34)). The previous failing scan check should now pass since fail-on-findings: false is finally honored end-to-end.

jpr5:
Bundled the security fixes for the findings sentinel surfaced. Promoted to blocking mode (fail-on-findings: true). Scan now clean.

Each fix is self-contained:
- **stable-release.yml**: env-block refactor moves `inputs.dry_run/bump/scope` from inline `${{ }}` to `env:` block + shell variables. Remaining `inputs.*` in `peter-evans/create-pull-request` `with:` params are action API parameters (not shell-executed) -- excepted via policy.
- **publish-release.yml**: scoped `GIT_CONFIG_*` env vars on the tag-push step replace the standalone `Configure git credentials` step that persisted the token in `.git/config` for the entire job.
- **publish-commit.yml**: explicit `persist-credentials: true` + policy exception (pkg-pr-new needs the token for PR comments).
- **showcase_build_check.yml**: policy exception for ai-config-injection (no AI tools run -- Docker builds only).
- **.sentinel-ci.yml**: new policy file with 3 documented exceptions.

BenTaylorDev:
Closing this out — workflow security scanning landed, and each of the hardening changes carried here was independently resolved. Walking through them so nothing is closed on a hand-wave.

**The scanner.** `edf10769ac` added `.github/workflows/security_zizmor.yml`, running at `min-severity: low` as a blocking check, backed by a documented `.github/zizmor.yml`. That covers the same classes this PR targeted — template injection, dangerous triggers, unpinned refs, excessive permissions, credential persistence — and it blocks, where Sentinel was warn-only pending its 7-day soak.

**Template injection in the release workflows.** This PR routed `inputs.bump` / `inputs.scope` / `inputs.dry_run` through `env:`. On main those are `type: choice` with closed option sets (`monorepo|angular|channels`, `patch|minor|major`) and `type: boolean`, so GitHub constrains the values before they reach the shell. `.github/zizmor.yml` sets `template-injection: ignore: []` — any real occurrence blocks — and the check is green, so this isn't a live finding.

**Credential persistence in `publish-release.yml`.** This is the one worth being explicit about, since `git config --local url…insteadOf` is still there at lines 326 and 762. The leak vector was the uploaded workspace artifact, and that is closed at the checkout: the build job sets `persist-credentials: false`, and the `insteadOf` setup runs in the publish job *after* the upload. `.github/zizmor.yml` documents that reasoning under `artipacked` and deliberately leaves the workflow unsuppressed.

**`publish-commit.yml`.** The persisted token is suppressed with the same justification given here — pkg-pr-new needs it to post snapshot comments.

**depot pin.** Moot: main is on `98e78ad` / v1.18.0 with an accurate version comment.

So the substance of the review landed even though the branch didn't. Thanks @jpr5 — the `artipacked` and template-injection reasoning now written into `.github/zizmor.yml` is the durable version of what this PR was arguing for.

Issue #4819, opened 2026-05-14
🐛 Bug: processAgentResult has no circuit breaker > recursive follow-ups can loop indefinitely


Jszigeti:
### ♻️ Reproduction Steps

1. Register a frontend tool (useFrontendTool) that returns a valid result, e.g. a search tool
2. Send a message that triggers the tool
3. The tool executes successfully, processAgentResult sets needsFollowUp = true
4. processAgentResult calls runAgent recursively
5. If the LLM decides to call the same tool again (or the backend returns a RunError per #2416),
   the cycle repeats: tool > followUp > runAgent > tool > followUp > runAgent > ...
6. There is no depth limit, no circuit breaker, no duplicate detection. The loop runs indefinitely.

This also manifests when:
- The backend crashes after receiving tool results (RunError → infinite retry, see #2416)
- inputProcessors reprocess tool messages (#3044)
- The LLM generates repeated tool calls for any reason

### ✅ Expected Behavior

processAgentResult should have a built-in safeguard against infinite recursive follow-ups.

At minimum:
- Consecutive identical follow-ups (same tool calls with same arguments) should be capped after N repetitions (e.g. 3)
- An absolute depth safety cap should prevent runaway recursion regardless of the pattern
- A console.warn should be emitted when the breaker trips, for debuggability

This would protect against all infinite loop scenarios, not just specific triggers, without requiring users to set followUp: false (which breaks multi-step workflows) or implement their own circuit breakers.

### ❌ Actual Behavior

In packages/core/src/core/run-handler.ts, processAgentResult (line ~440) does:

```typescript
if (needsFollowUp && !this._runAbortController?.signal.aborted) {
  await this._internal.waitForPendingFrameworkUpdates();
  return await this.runAgent({ agent });
}
```

There is no limit on recursion depth, no detection of repeated tool call patterns, and no safety cap. Any scenario that keeps needsFollowUp = true will loop forever, consuming API quota and potentially DOS'ing the backend.

The only current workaround is setting followUp: false on tools, which is too blunt: it prevents legitimate multi-step agent workflows (e.g. search > fill form > confirm > update > send email).

### 𝌚 CopilotKit Version

```shell
@copilotkit/core@1.56.2
@copilotkit/react-core@1.56.2


Also verified on main branch, the code is unchanged.
```

### 📄 Logs (Optional)

```shell
[...]

PR #5003, opened 2026-05-26
fix: circuit breaker - prevent infinite recursive follow-ups in run handler

deepshekhardas:
## Description

Prevents infinite recursive follow-up loops in the RunHandler by adding a circuit breaker pattern with max depth limit.

### \
un-handler.ts\
- Added \MAX_RUN_DEPTH=20\ constant to cap recursive follow-up depth
- Added \_runDepth\ counter incremented on each \
unAgent()\ call
- Added \_runAbortController\ to support cancellation of in-progress runs
- \bortCurrentRun()\ exposed to allow \stopAgent()\ to cancel runs
- Check \_runAbortController.signal.aborted\ before scheduling follow-ups
- Reset \_runDepth\ on abort, error, or completion

### \core.ts\
- \stopAgent()\ now calls \	his.runHandler.abortCurrentRun()\ to cancel any in-progress run
- Added \waitForPendingFrameworkUpdates\ to \CopilotKitCoreFriendsAccess\ interface

### \
eact-core.ts\
- Implemented \waitForPendingFrameworkUpdates\ as \wait new Promise(resolve => setTimeout(resolve, 0))\ to flush React state updates before follow-up


BenTaylorDev:
Closing as superseded. The abort-controller, `abortCurrentRun()`/`stopAgent` cancellation, and `waitForPendingFrameworkUpdates` machinery in this PR have since landed on `main` independently, so this diff no longer applies cleanly. The one remaining novel piece — a recursion depth cap — is handled more surgically in #5158 (and the common identical-loop case in #5818). Thanks for kicking this off @deepshekhardas; it stays tracked under #4819.

Issue #2066, opened 2025-06-28
🐛 Bug: latest agui endpoint for langgraph streams fail to supress messages, and does not register any emit


inno-odyssean:
### ♻️ Reproduction Steps

```
<Tabs groupId="lg-deployment-type" items={["Local (LangGraph Studio)", "Self hosted (FastAPI)", "LangGraph Platform"]}>
  <Tab value="Local (LangGraph Studio)">
    ```ts
    import { 
      CopilotRuntime,
      LangGraphAgent  // [!code highlight]
      // ...
    } from "@copilotkit/runtime";
    // ...
    const runtime = new CopilotRuntime({
      agents: { // [!code highlight:8]
          // New agent entry per agent you wish to use
          'sample_agent': new LangGraphAgent({
              deploymentUrl: '<your-api-url>',
              graphId: 'sample_agent',
              langsmithApiKey: '<your-langsmith-api-key>' // Optional
          }),
      },
    });
    // ...
    ```
  </Tab>
```


do this example from https://github.com/CopilotKit/CopilotKit/blob/main/docs/snippets/self-hosting-remote-endpoints.mdx?plain=1

in my backend agent I suppress the messages using

```
export function createInternalLLMConfig(config?: RunnableConfig): RunnableConfig | undefined {
  if (!config) return undefined;
  
  return copilotkitCustomizeConfig(
    config,
    {
      emitMessages: false,
      emitToolCalls: false,
    }
  );
}
```

```
      const integrationResult = await integrationModel.invoke(integrationPrompt, 
       createInternalLLMConfig(config)
      );
```

```
      const formattedPrompt = await template.invoke({
        sectionsContent,
        originalInstructions: state.instructions || "Generate a comprehensive analysis report",
        dissectedQuery: state.dissected_query || state.instructions || "No enhanced query available",
        sectionPlanReasoning: state.section_plan_reasoning || "No section plan reasoning available"
      }, createInternalLLMConfig(config));
```




### ✅ Expected Behavior

The messages get supressed and only the streams and manual emit are received

### ❌ Actual Behavior

no manual emits are received at all, but llm messages to chat are received.  If I swap to this configuration:

```
    remoteEndpoints: [
        langGraphPlatformEndpoint({
            deploymentUrl: "",
            agents: [
                {
                    name: "",
                    description: ""
                    assistantId: ""
                }
            ],
            langsmithApiKey: process.env.LANGSMITH_API_KEY,
        }),
    ],
```

[...]

PR #5004, opened 2026-05-26
fix: langgraph agent state - preserve existing copilotkit keys in merge_state and fix double-dispatch

deepshekhardas:
## Description

Fixes agent state not being readable from langgraph tools due to copilotkit keys being overwritten in merge_state. Also fixes double-dispatch of custom events in Python SDK.

### JS: \packages/v1/runtime/.../langgraph/agent.ts\
- Read \existingCopilotkit\ from \guiMergedState.copilotkit\ before overwriting
- Spread \...existingCopilotkit\ into the returned \copilotkit\ object

### Python: \sdk-python/copilotkit/langgraph_agui_agent.py\
- \langgraph_default_merge_state\: Read \existing_copilotkit\ from merged state and spread it before setting actions/context
- \_dispatch_event\: Return \None\ for ManuallyEmitMessage and ManuallyEmitToolCall after dispatching individual events, instead of calling \super()._dispatch_event(event)\ which caused double-dispatch
- Added \
un()\ override to filter out \None\ values
- Handle both dict and object forms for \
aw_event.metadata\ (fixes issue #2066)
- Added \dict_repr()\ override for proper serialization


BenTaylorDev:
Closing this PR as not mergeable in its current form — the substantive fix (`langgraph_agui_agent.py` double-dispatch guard + state-merge preservation) is bundled with edits to the retired `packages/v1/runtime` tree. If the `sdk-python` change is still needed on current `main`, a fresh PR scoped to just `sdk-python/copilotkit/langgraph_agui_agent.py` would be reviewable (worth re-verifying against main first, since the dispatch path was reworked). Thanks @deepshekhardas.

Issue #5000, opened 2026-05-24
🐛 Bug: useAgent: agent.subscribe() throws "Cannot read properties of undefined (reading 'subscribers')" on mount


codeonym:
### ♻️ Reproduction Steps

```tsx
import { useAgent } from "@copilotkit/react-core/v2";
import { useEffect } from "react";

function useAgentSubscriber(agentId: string) {
  const { agent } = useAgent({ agentId });

  useEffect(() => {

    // Throws on first render:
    // TypeError: Cannot read properties of undefined (reading 'subscribers')
    const sub = agent.subscribe({
      onRunFinalized: (params) => console.log(params),
    });

    return () => sub.unsubscribe();
  }, [agent]);
}
```

### ✅ Expected Behavior

`agent.subscribe()` should be safe to call as soon as `useAgent()` returns the agent. If initialization is **async**, the hook should expose an `isReady` flag so consumers can guard accordingly:

```tsx
const { agent, isReady } = useAgent({ agentId });

useEffect(() => {
  if (!isReady) return;

  const sub = agent.subscribe({
    onRunFinalized: (params) => console.log(params),
  });

  return () => sub.unsubscribe();
}, [agent, isReady]);
```

### ❌ Actual Behavior

Calling agent.subscribe() inside a useEffect immediately after useAgent() returns throws at runtime:

```
TypeError: Cannot read properties of undefined (reading 'subscribers')
    at AbstractAgent.subscribe
```
The hook always returns an AbstractAgent instance — never null — so there is no signal to know when the agent is fully initialized and safe to call subscribe() on.

### 𝌚 CopilotKit Version

```shell
├── @copilotkit/react-core@1.57.1
├── @copilotkit/react-ui@1.57.1
├── @copilotkit/runtime@1.57.1
```

### 📄 Logs (Optional)

```shell

```

codeonym:
The current type declaration says agent is always an AbstractAgent:

```tsx
declare function useAgent({
  agentId,
  updates,
  throttleMs
}?: UseAgentProps): {
  agent: AbstractAgent;
};
```

PR #5005, opened 2026-05-26
fix: use-agent - return provisional agent on Error status instead of throwing

deepshekhardas:
## Description

When the runtime connection status is \Error\, \useAgent()\ was falling through to the 'unknown agent' error path and throwing. This caused crashes on transient network issues or auth token refreshes.

### \use-agent.tsx\
- Added \CopilotKitCoreRuntimeConnectionStatus.Error\ to the provisional agent check alongside \Disconnected\ and \Connecting\
- On Error status, return a \ProxiedCopilotRuntimeAgent\ instead of throwing, allowing the UI to gracefully reconnect when the runtime recovers


BenTaylorDev:
Closing in favor of #5424 — that PR is the canonical fix for #5000 (`agent.subscribe()` on a detached agent) on the current flat package tree (`packages/core/src/agent.ts` + `packages/react-core/src/v2/hooks/`). This PR targets the retired `packages/v2/` tree and won't apply. Let's land it in #5424.

PR #5007, opened 2026-05-26
fix(core): guard proxied agent subscriptions

Tushar-Khandelwal-2004:
## Summary

-> Fixes issue #5000 where provisional agents created via `useAgent()` could crash when `agent.subscribe()` was called immediately after mount.
-> The root cause was that `ProxiedCopilotRuntimeAgent.subscribe()` could access AG-UI subscriber bookkeeping before the internal `subscribers` array had been initialized. This change ensures subscriber storage is always normalized before subscriptions are forwarded, making `subscribe()` safe to call as soon as the agent instance is available.

## Changes

-> Normalize subscriber storage during `ProxiedCopilotRuntimeAgent` initialization.
-> Override `subscribe()` to defensively restore missing subscriber state before delegating to AG-UI.
-> Add guards around intelligence-mode subscriber forwarding when `this.subscribers` is undefined.
-> Add a regression test covering missing subscriber initialization and validating subscribe/unsubscribe behavior.

## Testing

-> Added regression coverage in:
`src/__tests__/proxied-runtime-transport.test.ts`

-> Verified:
    -> `@copilotkit/core:test`
    -> `@copilotkit/core:build --excludeTaskDependencies`

### DIFF
--- a/packages/core/src/__tests__/proxied-runtime-transport.test.ts
+++ b/packages/core/src/__tests__/proxied-runtime-transport.test.ts
@@ -235,6 +235,26 @@ describe("ProxiedCopilotRuntimeAgent capabilities", () => {
   });
 });
 
+describe("ProxiedCopilotRuntimeAgent subscribers", () => {
+  it("repairs missing subscriber storage before subscribing", () => {
+    const agent = new ProxiedCopilotRuntimeAgent({
+      runtimeUrl: "http://localhost:3000",
+      agentId: "test-agent",
+      description: "Test agent",
+    });
+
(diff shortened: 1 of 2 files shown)


BenTaylorDev:
Thanks @Tushar-Khandelwal-2004. Closing this one — `AbstractAgent`'s constructor always initializes `subscribers = []` before any subclass/proxied code runs, so `subscribe()` can't hit an undefined `subscribers` on a freshly-constructed agent; the guard isn't reachable in practice. The underlying ask from #5000 (a safe readiness check) is already covered by the `isReady` flag now exported from `useAgent` on main. Appreciate the contribution.

PR #5025, opened 2026-05-26
fix: two-arg z.record for Zod 4 compatibility in MCP Apps schemas

### LINKED ISSUE #4295: MCPAppsActivityContentSchema uses deprecated single-arg z.record (Zod 4)
## Summary

`MCPAppsActivityContentSchema` in `@copilotkit/react-core` defines `toolInput` with the single-argument form of `z.record()`, which Zod 4 [explicitly removed](https://zod.dev/v4/changelog#drops-single-argument-usage). The schema appears valid until a non-empty `toolInput` reaches it, at which point Zod 4 throws `Cannot read properties of undefined (reading '_zod')` while trying to walk the nonexistent `valueType`.

## Affected version

- `@copilotkit/react-core@1.56.3` (likely earlier 1.56.x too)
- ships with `zod@^4.x` resolved (`zod@4.3.6` in my install)

## Where

`@copilotkit/react-core/dist/copilotkit-By2G6-Zx.cjs` around line 1630:

\`\`\`js
const MCPAppsActivityContentSchema = zod.z.object({
  result: zod.z.object({
    content: zod.z.array(zod.z.any()).optional(),
    structuredContent: zod.z.any().optional(),
    isError: zod.z.boolean().optional()
  }),
  resourceUri: zod.z.string(),
  serverHash: zod.z.string(),
  serverId: zod.z.string().optional(),
  toolInput: zod.z.record(zod.z.unknown()).optional()  // <-- single-arg, removed in Zod 4
});
\`\`\`

## Reproduction

\`\`\`js
import { z } from "zod";
const schema = z.object({
  toolInput: z.record(z.unknown()).optional(),
});
schema.safeParse({ toolInput: {} });             // OK (no values to recurse into)
schema.safeParse({ toolInput: { code: "x" } });  // throws: Cannot read properties of undefined (reading '_zod')
\`\`\`

The empty-record case happens to pass, which is why MCP Apps tools that take no arguments (e.g. server-side dashboards rendered from a tool with no inputs) work fine — the bug is silent until a tool with a non-empty input is invoked.

## Suggested fix

Two-arg form per the Zod v4 migration guide:

\`\`\`js
toolInput: zod.z.record(zod.z.string(), zod.z.unknown()).optional()
\`\`\`

## Symptom seen by users

`useRenderActivityMessage` (`use-render-activity-message.tsx`) calls `renderer.content.safeParse(message.content)`. The throw bubbles up, the renderer returns `null`, and the iframe never mounts (or stays stuck in its initial state). For us this was visible in MCP Apps generative-UI flows where the tool's argument is the Prefab Python source string.

## References

- Zod 4 changelog — drops single-argument usage: https://zod.dev/v4/changelog#drops-single-argument-usage
- Same root cause hitting another project: omar-dulaimi/prisma-zod-generator#232

contextablemark:
## Summary

`MCPAppsActivityContentSchema` (and a couple of sibling schemas) used the single-argument `z.record(valueType)` form for `toolInput`. **Zod 4 made the key schema mandatory**, so the single-arg form is a **compile-time error (`TS2554: Expected 2-3 arguments, but got 1`)** when built against Zod 4. Since `@copilotkit/react-core` declares `zod: ">=3.0.0"`, a downstream consumer that resolves Zod 4 is affected. Fixes #4295.

This PR switches to the two-argument `z.record(z.string(), z.unknown())` form (valid and identical under both Zod 3 and Zod 4) and adds a lint guard so it can't regress.

### Note on the issue's framing
The original issue (and the Linear ticket) described this as breaking *runtime validation for non-empty records*. I verified against real Zod 4.4.3 in isolation: **runtime `safeParse` is unaffected** under both Zod majors — the incompatibility is purely **type-level** (`tsc`). That distinction drove the test/lint decisions below.

## Changes
- **Schemas** — `toolInput` now uses `z.record(z.string(), z.unknown())` in:
  - `packages/react-core/.../MCPAppsActivityRenderer.tsx`
  - `packages/vue/.../MCPAppsActivityRenderer.ts`
- **Sibling fix** — same single-arg form corrected in `packages/react-core/.../defineToolCallRenderer.test.tsx` (a `metadata` schema).
- **Field-contract test** — added to `MCPAppsActivityRenderer.e2e.test.tsx`, asserting `toolInput` round-trips mixed value types. (Honest scope: a runtime test cannot reproduce a type-level break under either Zod major — see below.)
- **Lint guard** — new `copilotkit/no-single-arg-zod-record` oxlint rule (with autofix) in the existing `packages/react-ui/oxlint-rules/` plugin, enabled as `error` for `packages/**` in `.oxlintrc.json`. Because the break is type-level and the workspace lockfile pins Zod 3, this static rule is the only thing that actually prevents regressions here.

### Lint rule scope
Scoped to `packages/**` (published surface that declares Zod 4-compatible peers). `examples/` and `showcase/` still contain the single-arg form in ~80 spots, but those are unpublished apps with pinned Zod; migrating them is out of scope for this fix and would be a large mechanical change. `packages/**` is clean after this PR (0 lint errors).

## Test plan
- [x] `nx run @copilotkit/react-core:test` — affected files pass (28/28 incl. new test)
- [x] Full `packages/**` suite + `publint`/`attw` green (pre-commit hook)
- [x] `oxlint packages/` — 0 errors; new rule fires + autofixes on a probe, clean on the fixed code
- [x] Isolated Zod 4.4.3 repro confirms: single-arg → `TS2554`; two-arg → compiles; runtime parsing unaffected for both


onsclom (approved):

packages/react-ui/oxlint-rules/no-single-arg-zod-record.mjs
```diff
@@ -0,0 +1,73 @@
+/**
```
marthakelly, line 1: Love the lint rule approach here, especially since the bug is type-level and no runtime test can actually catch a regression. Since this is the real safety net, worth adding a few RuleTester cases? The infra is already in the repo (oxlint/plugins-dev). Something like: valid two-arg form doesn't false-positive, invalid single-arg fires + autofix output is correct, SpreadElement reports without a fix.

PR #5049, opened 2026-05-27
fix(showcase/pocketbase): register CORS via routerPre and lock down baseline.createRule

jpr5:
## Summary

Two related fixes to `showcase/pocketbase` that share the same rebuild/deploy cycle.

### 1. CORS middleware was a silent no-op

`showcase/pocketbase/pb_hooks/main.pb.js` registered the CORS middleware via `onBeforeServe((e) => e.router.use(...))`. PB 0.22's JSVM excludes `onBeforeServe` from auto-binding — see `plugins/jsvm/binds.go` in v0.22.21 `excludeHooks`. The hook was never registered; CORS for the dashboard was effectively unwired.

The supported module-scope binding is `routerPre`, which internally wraps `OnBeforeServe + Router.Pre`. Switching to it gets the middleware actually registered.

One subtle gotcha forced an additional change: `routerPre` re-compiles the middleware function in a fresh executor VM via `m.String()` (see `wrapMiddlewares` in `binds.go`). Module-scope `const`s and helper functions defined outside the middleware body are NOT in that executor's scope — referencing them from the handler throws `ReferenceError` at request time and PB returns a generic 400. The CORS allowlist + helpers therefore had to be inlined into the middleware closure. `$os.getenv` survives because it's a JSVM global binding, not a module-scope identifier. The duplication is documented in the file.

### 2. `baseline.createRule` was publicly writable

`1777700000_create_baseline.js` shipped the `baseline` collection with `createRule: ""`. In PB rule semantics, empty-string means "any caller can create" — i.e. anyone in the world could `POST /api/collections/baseline/records` from a browser. The dashboard never creates baseline rows from the browser; status-writer + offline tooling do, both with admin auth.

Added `1779909019_baseline_createRule_lockdown.js` to set `createRule = null` (admin-only). Down restores the legacy empty-string rule for rollback compatibility.

## Local Docker verification

Built `showcase/pocketbase/Dockerfile` and ran with `PB_CORS_ORIGINS=https://example.test`:

- No `ReferenceError: onBeforeServe is not defined` (or any other) on hook load.
- `OPTIONS /api/health` from `https://example.test` → `204 No Content` + `Access-Control-Allow-Origin: https://example.test` + `Vary: Origin` + the expected methods/headers/max-age.
- `OPTIONS /api/health` from `https://evil.test` → `204 No Content` with NO `Access-Control-Allow-Origin` header.
- Migration `1779909019_baseline_createRule_lockdown.js` recorded in `_migrations`; sqlite dump confirms `baseline.createRule IS NULL`.

For the actual GET response: PB 0.22's built-in echo CORS middleware (running in the `Use()` phase, after our `Pre()`) overwrites `Access-Control-Allow-Origin` to `*`. We can't prevent this from a `Pre()` hook — but the OPTIONS short-circuit returns before that middleware runs, so preflight uses our echoed origin. The actual response getting `*` is acceptable here: the dashboard uses bearer tokens (no cookies / no `credentials: 'include'`).

## Test plan

- [ ] CI green on `showcase_build` for `showcase-pocketbase` image.
- [ ] After merge + CI build, pin Railway prod + staging `showcase-pocketbase` services to the new image digest (currently runs `:latest` — see PR comment).
- [ ] Verify dashboard at `https://dashboard.showcase.copilotkit.ai` still loads against the deployed PB (CORS now actually works end-to-end).
- [ ] Verify `POST /api/collections/baseline/records` from a non-admin caller returns 403/401 against the deployed PB.

### DIFF
--- a/showcase/pocketbase/pb_hooks/main.pb.js
+++ b/showcase/pocketbase/pb_hooks/main.pb.js
@@ -24,66 +24,75 @@
 // additively. An operator can add staging / preview origins without a
 // code change; removing the prod default still requires editing this
 // file (env var cannot retire it).
-
-// Built-in prod origin baked in so a fresh Railway volume is immediately
-// reachable from the production dashboard without requiring operator
-// env-var configuration. If prod ever moves off
-// dashboard.showcase.copilotkit.ai, change this default — env var alone
-// won't retire the stale origin.
-const CORS_DEFAULT_ORIGINS = ["https://dashboard.showcase.copilotkit.ai"];
-
-function corsAllowedOrigins() {
-  const extra = ($os.getenv("PB_CORS_ORIGINS") || "")
-    .split(",")
-    .map(function (s) {
-      return s.trim();
-    })
-    .filter(function (s) {
-      return s.length > 0;
-    });
-  return CORS_DEFAULT_ORIGINS.concat(extra);
-}
-
-function originAllowed(origin, allowlist) {
-  if (!origin) return false;
-  for (let i = 0; i < allowlist.length; i++) {
-    if (allowlist[i] === origin) return true;
-  }
-  return false;
-}
-
-// Middleware registered at Pre so CORS runs before auth short-circuits
-// OPTIONS preflights. PB 0.22 exposes `e.router.use(...)` on the echo
-// router inside `onBeforeServe`; the callback receives the echo context
-// as its argument.
-onBeforeServe((e) => {
-  e.router.use((next) => {
-    return (c) => {
-      const origin = c.request().header.get("Origin");
-      const allowlist = corsAllowedOrigins();
-      if (originAllowed(origin, allowlist)) {
-        c.response().header().set("Access-Control-Allow-Origin", origin);
-        c.response().header().set("Vary", "Origin");
-        c.response()
-          .header()
-          .set(
-            "Access-Control-Allow-Methods",
-            "GET, POST, PATCH, DELETE, OPTIONS",
-          );
-        c.response()
-          .header()
-          .set(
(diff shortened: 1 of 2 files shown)


jpr5:
**Post-merge follow-up (Railway):** after this PR merges and CI builds the new `showcase-pocketbase` image, Railway prod + staging both need an explicit `serviceInstanceUpdate` to pin the new digest.

`showcase-pocketbase` currently runs `ghcr.io/copilotkit/showcase-pocketbase:latest` on Railway. Per our production hosting spec, prod should never reference mutable tags — both because of supply-chain blast radius (any subsequent push to `:latest` lands instantly) and because Railway's redeploy-on-push behavior on a mutable tag is non-deterministic across services.

The mechanical step after CI publishes the new image:

1. Read the new digest from GHCR (`gh api /orgs/CopilotKit/packages/container/showcase-pocketbase/versions --jq '.[0].name'`, or the showcase deploy workflow log).
2. `serviceInstanceUpdate` on the `showcase-pocketbase` service in both the prod and staging environments to set the image source to `ghcr.io/copilotkit/showcase-pocketbase@sha256:<digest>`.
3. Wait for both to redeploy and confirm CORS + the `baseline.createRule = null` lockdown took effect against the live URLs.

Both env IDs and the Railway project ID are in the showcase-harness Railway reference. Tagging this here so it doesn't get lost when the PR merges.

BenTaylorDev:
Closing as superseded — the CORS wiring and createRule lockdown both shipped via later PocketBase work (`showcase/pocketbase/pb_hooks/main.pb.js` onBeforeServe + `pb_migrations/*lockdown*`). This PR's routerPre approach didn't land, but its goals are covered.

PR #5051, opened 2026-05-27
feat(ci): bundle-size tracking + ES-compat checks (OSS-123, OSS-121)

davidmckayv:
## Summary

Adds CI **instrumentation** for bundle health (no source/runtime behavior change shipped to consumers):

- **OSS-123 — bundle size tracking.** `static_bundle_size.yml` uses `preactjs/compressed-size-action` (SHA-pinned) to post a per-file **gzip** diff comment on every PR over the 9 in-scope packages. No hard-fail (Phase 1). `.size-limit.json` for the 4 bundled packages (local `pnpm size`). Plus a `copilotchat-headline` job that writes a **relative** "total bundled JS" regression signal to the job summary (esbuild-based; not a production figure — see below).
- **OSS-121 — ES-compat checks.** `static_compat.yml` runs `es-check` (hard-fail) across the same 9 packages via per-package `compat-check` scripts (es2022 ESM/CJS; es2018 UMD, **es2020 for react-core UMD** which uses dynamic `import()`); wired as an nx `compat-check` target.
- `.browserslistrc`, lefthook/oxfmt guards (keep formatters off `package.json`), and `.claude/docs/{bundle-size,browser-compat}.md`.

## Honest notes / deviations from the tickets
- Tool: `preactjs/compressed-size-action` instead of the ticket's `andresz1/size-limit-action` (the latter has unresolved CVEs). Consequence: **gzip-only** (the action reports one algorithm per run), not gzip+brotli.
- `compat-check` targets **es2022** (build output), not the ticket's proposed es2020.
- Only **4 of 9** packages get `.size-limit.json` (`@size-limit/file` only measures the barrel for the 5 unbundled packages).
- The react-core "headline number" is a **relative regression signal**, not the production Vite/Next figure. The faithful production measurement is done in the OSS-122 PR (Phase 0).

## Test plan
- [x] `compat-check` smoke: injected ES2020 syntax into a built es2018 UMD → `compat-check` went **red**; reverted → green. (Confirms the check fires.)
- [x] bundle-size smoke (local): padding react-core grew the measured `dist/index.mjs` by ~31.7 kB gzip; reverted to baseline. (Confirms delta detection; the PR-comment rendering will show on this PR's CI run.)
- [x] `pre-push-quality`: formatter, lint, tests, build, lockfile all green.
- [ ] Confirm `static_bundle_size` posts its comment and `static_compat` passes on this PR's CI.


dev-docs/browser-compat.md
tylerslaton, line 1: I think we can move these to dev-docs instead. I say that because, as I understand it, these are the first doc we're adding to this folder right now. Would be ideal to keep in one place. WDYT?

.github/workflows/static_bundle_size.yml
```diff
@@ -0,0 +1,93 @@
...
+
+permissions:
+  contents: read
+
+env:
+  NODE_OPTIONS: "--max-old-space-size=4096"
+  NX_VERBOSE_LOGGING: true
+
+jobs:
+  bundle-size:
```
tylerslaton, line 34: nit: can we add a comment to the top of this jobs describing what this is intended to do and how?

.github/workflows/static_bundle_size.yml
```diff
@@ -0,0 +1,93 @@
...
+      - name: Use Node.js 20
+        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
+        with:
+          node-version: 20.x
+          # setup-node built-in cache is fork-safe (fork PRs can't write to base repo cache)
+          cache: "pnpm"
+          cache-dependency-path: "**/pnpm-lock.yaml"
+
+      - name: Measure compressed bundle sizes
+        uses: preactjs/compressed-size-action@66325aad6443cb7cf89c4bfcd414aea2367cda94 # 2.9.1
```
tylerslaton, line 63: Just checking - we use the hash here for security purposes, right?

.github/workflows/static_bundle_size.yml
```diff
@@ -0,0 +1,93 @@
...
+          # Use the root `build` script (present on both this branch and the base
+          # branch) so compressed-size-action can build both sides for comparison.
+          # The `pattern` below restricts measurement to the 9 in-scope packages.
+          build-script: build
+          pattern: "packages/{core,shared,react-core,react-ui,react-textarea,runtime-client-gql,web-inspector,voice,a2ui-renderer}/dist/**/*.{mjs,js,cjs}"
+
+  # Relative regression signal for what an app importing { CopilotChat } bundles.
+  # Writes the total gzipped JS to the job summary (not the PR comment). This is
+  # an esbuild measurement for consistency across PRs — NOT a production Vite/Next
+  # figure (see .claude/docs/bundle-size.md and OSS-122 Phase 0).
```
tylerslaton: We don't want to include references to tickets or claude plans in comments like this.

.github/workflows/static_bundle_size.yml
```diff
@@ -0,0 +1,93 @@
...
+  copilotchat-headline:
```
tylerslaton: The name confused me slightly. What do you think of something like `copilotchat-import-size`? What do you think?

packages/react-core/scripts/measure-copilotchat.mjs
tylerslaton, line 1: Can we have a unit test that validates this script? Really easy for scripts to regress overtime if we ever come back and change stuff.

packages/react-core/.size-limit.json
```diff
@@ -0,0 +1,17 @@
+[
+  {
+    "name": "@copilotkit/react-core (v1 root)",
+    "path": "dist/index.mjs",
+    "gzip": true
+  },
```
tylerslaton: No need to add any tooling for v1, we only care about v2.

tylerslaton (changes requested):
Looking good! Had some comments I want to make sure we resolve for maintenance and agent operation in the future.

dev-docs/browser-compat.md
davidmckayv: Done in 7cde7a405 — moved both `.claude/docs/bundle-size.md` and `.claude/docs/browser-compat.md` into `dev-docs/` and removed the `.claude/docs/` folder. Cross-refs in the workflow now point at `dev-docs/bundle-size.md`.

.github/workflows/static_bundle_size.yml
davidmckayv: Added in 7cde7a405 — top-of-job comment explains the action, the pattern glob, the comment-on-PR behavior, and the Phase-1 no-hard-fail framing.

.github/workflows/static_bundle_size.yml
davidmckayv: Yes — every `uses:` is pinned to a full 40-char commit SHA precisely for supply-chain reasons. The trailing `# 2.9.1` is just a human-readable tag pointer; the SHA is the source of truth. zizmor's `ref-version-mismatch` audit cross-checks the SHA against the tag (which is why we corrected `# v2.9.1` → `# 2.9.1` — the actual tag has no `v` prefix).

.github/workflows/static_bundle_size.yml
davidmckayv: Done in 7cde7a405 — rewrote the job preamble. It now describes what the job does without naming OSS-122/Phase 0 or `.claude/docs/`, and links only to `dev-docs/bundle-size.md`.

.github/workflows/static_bundle_size.yml
davidmckayv: Renamed to `copilotchat-import-size` in 7cde7a405.

packages/react-core/scripts/measure-copilotchat.mjs
davidmckayv: Done in 7cde7a405 — refactored the script to export `measureBundle({ entryModule, pkgRoot })` with an `import.meta.url` CLI guard so it's importable from tests. Added `scripts/__tests__/measure-copilotchat.test.mjs` (two `node:test` cases against a tiny fixture: positive gzip total and determinism across two runs) plus the fixture file. Wired `test:scripts` into the package `test` script so CI runs it as part of `pnpm test`.

packages/react-core/.size-limit.json
davidmckayv: Done in 7cde7a405 — the v1-root entry is gone; `.size-limit.json` now only carries the v2 entries (`v2` and `v2/headless`).

davidmckayv:
@tylerslaton ready for re-review when you have a sec — all 7 review threads resolved (docs → `dev-docs/`, job comment added, ticket/claude refs removed, renamed to `copilotchat-import-size`, `measureBundle()` now has a node:test unit test wired into CI, v1 `.size-limit.json` entry dropped) and CI is green (41 pass / 0 fail).

.github/workflows/static_bundle_size.yml
```diff
@@ -0,0 +1,99 @@
+name: static / bundle-size
+
+on:
+  pull_request:
```
tylerslaton, line 4: nit: This workflow says the bundle-size action posts a PR comment on every PR, but `pull_request` runs from forks get a read-only token, and `preactjs/compressed-size-action` only prints the report to logs in that case. Since this repo accepts external PRs, we should either document that limitation or use a safe comment relay pattern if the PR comment is required.

dev-docs/bundle-size.md
```diff
@@ -0,0 +1,85 @@
...
+
+Key facts:
+
+- Reports by **file path**, not by named entry — it does not read `.size-limit.json` at all.
+- The action runs `build-script: build` (the root `build` script — `nx run-many -t build` over all `packages/**`) on both the PR branch and the base branch, then measures only the files matched by the `pattern` glob. The root `build` script is used (rather than a bundle-size-specific one) because the action must build the base branch too, and `build` exists on every branch. No separate build step is needed before the workflow triggers — the action handles both builds.
+- PR comments show paths like `packages/react-core/dist/index.mjs (+1.2 kB gzip)`.
+
+### The CopilotChat regression signal (job summary, not the PR comment)
+
+The `copilotchat-headline` job in `static_bundle_size.yml` measures what an app
```
tylerslaton: nit: This still says `copilotchat-headline`, but the workflow job was renamed to `copilotchat-import-size`.

tylerslaton (dismissed):
Approved with nits.

.github/workflows/static_bundle_size.yml
davidmckayv: Done in c6ca283e9 — documented the fork limitation rather than adding a relay (the report is informational with no hard-fail in Phase 1). Added a note to the `bundle-size` job preamble in the workflow and a callout in `dev-docs/bundle-size.md`: fork `pull_request` runs get a read-only token, so the action prints the report to logs instead of commenting; the measurement still runs. If the comment ever becomes required, the note points at the `pull_request_target` + `workflow_run` relay pattern.

dev-docs/bundle-size.md
davidmckayv: Fixed in c6ca283e9 — `dev-docs/bundle-size.md` now says `copilotchat-import-size` to match the renamed job. Confirmed no other `copilotchat-headline` references remain in docs or workflows.

tylerslaton (approved):

PR #5052, opened 2026-05-27
feat(react-core): reduce streamdown bundle impact (OSS-122)

davidmckayv:
## Summary

Cuts streamdown's contribution to CopilotChat's **initial** bundle. **Stacked on #5051** (base is the OSS-123 branch so this diff shows only the OSS-122 commits; it will retarget to `main` once #5051 merges).

- **streamdown `^1.3.0` → `^2.5.0`.** v2 CDN-loads Shiki language grammars instead of bundling them, and ships code/math/mermaid as **opt-in plugins**.
- **Lazy-load.** Both `<Streamdown>` call sites render via `LazyStreamdown` (`React.lazy` + `Suspense`, `<pre>` fallback). The lazy target `StreamdownWithPlugins` imports streamdown + `@streamdown/{code,math,mermaid}`, so all that machinery lives in a **lazy chunk** — builds that never render chat ship 0 bytes of it.
- **KaTeX gating.** `useKatexStyles` now injects KaTeX CSS only when the message contains math.
- **Functionality preserved (v1 parity):** code highlighting, block + inline (`$…$`) math, and on-demand mermaid all work — verified in a real browser (Playwright against Storybook).
- **Security:** pnpm overrides `mermaid: >=11.15.0` (patches 4 XSS/DoS advisories that streamdown v2's mermaid@11.12.2 had) and `shiki: 3.22.0` (dedupes a `@streamdown/code` vs `streamdown` version skew that caused a type error).
- Docs: `.claude/docs/streamdown-cdn.md` (lazy/CDN loading, CSP implications, security overrides).

## Reduction — faithful Next 15 measurement (OSS-122 Phase 0)
In a real `next build`, **none** of streamdown/katex/mermaid/shiki appear in the initial/shared JS — all lazy; Shiki's `oniguruma` engine isn't bundled at all. Initial/shared ≈ **156 kB gzip** vs Notion's v1 baseline of **386 kB main initial** (which had this machinery eager). The ~4 MB streamdown drag is out of the initial paint.

> Note: the OSS-123 "total bundled JS" regression signal reads ~3 MB once functionality is restored, because the machinery re-enters as **lazy** chunks. The win is initial-load, not total-bundled.

## Test plan
- [x] 1176 react-core unit tests pass (incl. new LazyStreamdown/Suspense + `containsMath` + single-`$` math regression tests).
- [x] Browser (Playwright): markdown, code blocks, inline + block math render; lazy Suspense fallback works; no streamdown/plugin JS errors.
- [x] Real Next 15 production build succeeds; heavy libs confirmed lazy-only.
- [x] CR (4 agents) + `pre-push-quality` green; security audit shows 0 streamdown/mermaid advisories.
- [ ] Follow-up (separate): dedupe the duplicate lazy chunk under bundlers that don't (real Next/webpack dedupe); add a `CopilotChatReasoningMessage` test (pre-existing gap).


onsclom (dismissed):

dev-docs/streamdown-cdn.md
tylerslaton, line 1: Can we move this into the dev-docs folder? We don't put stuff in this docs folder today.

tylerslaton:
Had an agent give this a look and it found some issues I'd like to double check. The agent reproduced several times locally and want your input.

---

I think there are two package-output regressions that should be fixed before merge.

1. **CJS consumers will fail when chat renders.** `StreamdownWithPlugins.tsx` imports `streamdown` and `@streamdown/{code,math,mermaid}` statically. The generated CJS lazy chunk now contains:

```js
require("streamdown");
require("@streamdown/code");
require("@streamdown/math");
require("@streamdown/mermaid");
```

Those packages are ESM-only and their `exports` only expose `import`, not `require`. I confirmed this in the PR worktree with `node -e "require('@streamdown/math')"`, which throws `ERR_PACKAGE_PATH_NOT_EXPORTED`. So a CommonJS consumer of `@copilotkit/react-core/v2` can import the package, but the `React.lazy` chunk fails as soon as `LazyStreamdown` resolves. We should either bundle/noExternal these ESM-only deps into the CJS lazy chunk or preserve dynamic `import()` for the CJS path.

2. **The UMD build now has unresolved streamdown plugin globals.** The build emits warnings for missing globals for `@streamdown/code`, `@streamdown/math`, and `@streamdown/mermaid`, and `dist/v2/index.umd.js` reads guessed globals like `global._streamdown_code`, `global._streamdown_math`, and `global._streamdown_mermaid`. Since `package.json` still advertises the UMD bundle via `unpkg`/`jsdelivr`, CDN users would need globals that those packages do not appear to publish. We should either bundle these plugin deps into UMD or provide explicit supported globals/loading docs.

Validation I reran:

```bash
pnpm --filter @copilotkit/react-core build
node -e "require('@streamdown/math')"
```

The build succeeds, but it exposes the generated CJS `require()` calls and UMD missing-global warnings above.

tylerslaton (changes requested):

dev-docs/streamdown-cdn.md
davidmckayv: Done in 8f079c707 — moved to `dev-docs/streamdown-cdn.md` (the `.claude/docs/` folder is gone here too).

davidmckayv:
On the package-output regressions your agent flagged:

**UMD globals — fixed in 8f079c707.** Defined `StreamdownCode` / `StreamdownMath` / `StreamdownMermaid` globals in the v2 UMD entry's `outputOptions.globals`, same model v1 had with `streamdown: "streamdown"`. The UMD wrapper now reads `global.StreamdownCode` etc. instead of guessed `global._streamdown_*` names (the `_streamdown_*` strings that remain in the bundle are internal factory-param aliases inside the IIFE, not consumer-facing global reads — verified by grep on the emitted UMD).

**CJS require() of ESM-only packages — attempted, OOM'd CI, reverted in 6cb46e7ef.** The fix was `tsdown noExternal` for streamdown + `@streamdown/{code,math,mermaid}` on the ESM/CJS entry — correct in principle (inlines them into the lazy chunk so the CJS variant has no bad `require()`). Locally it built fine, but on CI it pushed `react-core:build` past the runner's default Node heap (exit 130, SIGINT/OOM), cascading into every downstream check. I reverted it and documented the caveat in `dev-docs/streamdown-cdn.md` ("CJS consumer caveat").

In practice the broken `require()` path only fires inside a Node-CJS runtime that actually evaluates the chat's `React.lazy` — modern browser-bundler consumers (Next/Vite/webpack) resolve via the ESM dist and never see it; the CJS dist is consumed by SSR/RSC/tooling paths that don't trigger the lazy load. The proper fix is either (a) bump `NODE_OPTIONS=--max-old-space-size` on the build workflow so `noExternal` can land, or (b) route the CJS path through `await import()` instead of `require()`. Happy to do (a) in this PR if you'd rather take the inline approach now; otherwise it's a follow-up.

davidmckayv:
@tylerslaton ready for re-review when you have a sec — dev-docs move done, and on the two regressions your agent found: **UMD globals fixed** (`StreamdownCode`/`Math`/`Mermaid` defined in the v2 UMD entry). The **CJS `require()` fix** I attempted via `noExternal` but it OOM'd the build runner, so I reverted it and documented the caveat in `dev-docs/streamdown-cdn.md` — happy to land it properly with a build-heap bump if you'd rather not defer. CI green (15 pass / 0 fail), onsclom approved.

tylerslaton (changes requested):
Reading the comment about the CJS issue - we really can't regress on this. We have a lot of downstream consumers in enterprises right now that rely on older Node and React versions where this is important. I still can't quite get a sense if this is a regression from our change or if it has always been this way. Could you help me understand that? If it has always been this way, let's merge and then do a follow-up to fix it.

<details><summary>Agent review</summary>
<p>

Thanks for digging into this and for fixing the dev-docs move + UMD globals. I buy that the initial-load win is the right shape here.

I don’t want to merge with the CJS failure path knowingly left in the published package though. The docs caveat helps explain it, but it doesn’t change that `@copilotkit/react-core` still publishes a CJS build whose lazy chunk can emit `require("streamdown")` / plugin `require()` calls for ESM-only packages and throw `ERR_PACKAGE_PATH_NOT_EXPORTED` if that path is hit.

Even if the common Next/Vite/webpack path is fine, this is still a known runtime trap in an artifact we ship. I’d rather fix that here than make users discover it later.

Can you land one of the real fixes before merge?

- preserve `await import()` for the CJS lazy path, or
- inline/noExternal the streamdown deps and bump the build heap enough for CI

The UMD globals fix looks good to me; this is just about the remaining CJS issue.


</p>
</details>

davidmckayv:
@tylerslaton — landed fix (a): the CJS `require("streamdown")` trap is gone.

`streamdown` + `@streamdown/{code,math,mermaid}` are now loaded via dynamic `import()` inside the `React.lazy` factory in `LazyStreamdown.tsx` instead of a static import. tsdown/rolldown preserves a dynamic `import()` as a **native `import()`** in the CJS output (a static import is what got emitted as `require()`), and native `import()` resolves an ESM-only package correctly from a Node CJS runtime. The imports stay code-split, so the initial-load win is unchanged — and this avoids the `noExternal` inlining that OOM'd the build.

Verified on the built output:
- `dist/**/*.cjs` contains **no** `require("streamdown")` / `require("@streamdown/…")` — only the dynamic `import(...)` form.
- `streamdown` is not eager in either CJS entry (`dist/index.cjs`, `dist/v2/index.cjs`).
- All 94 react-core test files pass, including the LazyStreamdown markdown + single-$ math + fallback guards.

`StreamdownWithPlugins.tsx` was folded into the lazy factory (its only consumer). Updated the `tsdown.config.ts` note and rewrote the `dev-docs/streamdown-cdn.md` 'CJS consumer caveat' into a 'CJS safety' section describing the dynamic-import approach. Ready for re-review — this is the only commit on the PR now (#5051 merged, so it's retargeted to main).

BenTaylorDev:
Closing in favor of #5147 — the direction shifted from reducing the streamdown bundle (this PR, OSS-122) to removing the bundled dependency entirely with a pluggable renderer (#5147). They collide on the same files; #5147 is the more recent/comprehensive approach. Reopen if #5147 stalls and the bundle-reduction path is preferred.

PR #5081, opened 2026-05-28
fix(shell-docs): fix Tab double-escaping that hid multi-word tab content

onsclom:
## Problem

On [docs.copilotkit.ai/built-in-agent/build-with-agents](https://docs.copilotkit.ai/built-in-agent/build-with-agents), **"JSON Configuration File"** shows a blank panel.

image

## Why

Our `Tab` wrapper called `escapeValue()` before passing to Fumadocs's `FumadocsTab` — which also calls it internally. For 3-word labels, the double-escape produces different values for the trigger vs the content panel, so Radix hides the content.

```
"JSON Configuration File"
  → trigger (1 pass):   json-configuration file    ← space
  → content (2 passes): json-configuration-file    ← hyphen (mismatch → hidden)
```

## Fix

Remove `escapeValue()` from the `Tab` wrapper — Fumadocs already handles it. The `Tabs` `defaultValue` still needs pre-escaping since `FumadocsTabs` doesn't apply it internally.

### DIFF
--- a/showcase/shell-docs/src/components/docs-tabs.tsx
+++ b/showcase/shell-docs/src/components/docs-tabs.tsx
@@ -6,9 +6,12 @@
 //     `value.toLowerCase().replace(/\s/, "-")` to derive radix-tabs
 //     `value` keys. Our authored MDX uses the literal label as the
 //     <Tab value="...">, e.g. `<Tab value="JavaScript">` against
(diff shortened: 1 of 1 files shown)


tylerslaton (approved):
Nice 😄

PR #5084, opened 2026-05-28
ci(format): make main self-healing for formatting violations

onsclom:
## Why

Three PRs (#5018, #5020, #5021) merged to main before their format CI job could run, leaving unformatted files that broke the on-push `oxfmt --check .` gate.

Root cause: the `PROTECT_OUR_MAIN` ruleset only requires a review approval — no CI check is required to pass before merging. PRs merged within 1–3 seconds of CI triggering, before the format job started.

## What

**1. Fix on-push format check → auto-commit instead of fail** (`.github/workflows/static_quality.yml`)

When a push to main has formatting violations, the format job now runs `oxfmt --write` + `ruff format` and commits the fixes back to main automatically, instead of failing. Main becomes self-healing: a fast-merged PR that slips through with violations gets a bot follow-up commit within ~30 seconds.

The PR path is unchanged — authors still see violations and the auto-fixer still commits to the PR branch.

**2. Formatting fixes** (`showcase/shell-dashboard/`)

Applies `oxfmt --write .` to the three files that were blocking CI. No logic changes.

## Tradeoffs

- A violation that slips through will briefly leave main in an unformatted state until the bot commit lands
- The bot commit triggers a second CI run (concurrency group cancels the first if still running)
- `oxfmt --write .` scans the whole repo on every push to main regardless of what changed

If tighter enforcement is needed, adding `format` as a required status check in the [PROTECT_OUR_MAIN ruleset](https://github.com/CopilotKit/CopilotKit/rules/14521235) would block merges until the format job (and its auto-fix) completes.

### DIFF
--- a/.github/workflows/static_quality.yml
+++ b/.github/workflows/static_quality.yml
@@ -103,7 +103,7 @@ jobs:
           echo "PR-changed format candidates: $count"
           cat .pr-format-files.existing.txt
 
-      - name: Run formatter (check on push, fix on PR)
+      - name: Run formatter (fix on PR, fix-and-commit on push)
         run: |
           if [ "${{ github.event_name }}" = "pull_request" ]; then
             if [ "${{ steps.changed.outputs.count }}" = "0" ]; then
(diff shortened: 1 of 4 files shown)


onsclom:
## CR Round 1 — 7-agent review findings

**Bucket (a) — mandatory fixes (all in new push-path code):**

1. **No `git pull --rebase` before `git push`** — concurrent main merges cause a non-fast-forward failure, silently leaving main unformatted with a red CI run.

2. **`git add -A` too broad** — stages all untracked files in the working tree, not just formatter output. The PR path carefully uses a scoped file list; the push path should too.

3. **No post-write `--check` verification on push path** — if `oxfmt` fails mid-write, an incomplete fix gets committed silently. The PR path runs `--check` after `--write`; the push path does not.

4. **Push-path formatter lacks path scoping** — `oxfmt --write .` + `ruff format .` runs on the whole tree including `docs/`, `examples/`, and lockfiles. The PR path explicitly excludes lockfiles and those paths. The push path should use the same exclusion set.

**Bucket (b) — trivial / fixed alongside (a):**
- `git diff --name-only` → `git status --porcelain` for dirty-tree detection (catches staged changes too)

**Bucket (c) — pre-existing, deferred (not this PR's subject):**
- Dashboard bugs: `healthStats` gray counted as green, `depthDistribution` NaN for D6, `parityStats` unsafe cast, `arePropsEqual` missing health/smoke dimensions, operator error surfaces invisible — all in code this PR only reformatted.
- Pre-existing workflow: `grep -vxFf` empty-file behavior, oxfmt/ruff error-swallowing in PR path, `set -euo pipefail` missing, commitlint revert-commit gap.

**Bucket (d) — real issue, different PR:**
- Integration tests target deprecated `ComposedCell` instead of production `UnifiedCell` — subject: showcase-dashboard integration test coverage sweep.

Fixing bucket (a) + (b) now.

BenTaylorDev:
The self-healing goal here was achieved via a different design — the `format` job is now PR-only with an auto-fix-and-commit to the PR branch (rather than a push-to-main self-heal), and the shell-dashboard formatting has since landed independently. Closing as superseded.

PR #5099, opened 2026-05-29
fix(react-core): scope streamdown markdown styles

### LINKED ISSUE #5072: 🐛 Bug: Default Streamdown markdown classes are unstyled in non-Tailwind consumers due to cpk: prefix
### ♻️ Reproduction Steps

1. Create a fresh React app without Tailwind CSS (e.g. `npm create vite@latest my-app -- --template react-ts`)
2. Install CopilotKit: `npm install @copilotkit/react-core @copilotkit/react-ui`
3. Import the v2 stylesheet as documented: `import '@copilotkit/react-core/v2/styles.css'`
4. Render `<CopilotChat />` and send a message that produces markdown formatting — e.g. ask the agent to reply with `**bold text**`, a `# Heading`, and `- a bulleted list`
5. Inspect the rendered output in DevTools

### ✅ Expected Behavior

Bold text renders bold, headings have visible size hierarchy, and lists show their markers — using whatever styling CopilotKit ships as its default look.

The documented setup (install + import styles.css) should produce a working chat UI out of the box for all consumers, not just those who happen to have Tailwind installed.

### ❌ Actual Behavior

Markdown formatting is silently broken:

- **Bold** renders at normal weight — Streamdown emits `<span class="font-semibold">` but no `.font-semibold` rule exists in the bundled CSS
- Headings have no size/weight differentiation — Streamdown emits `<h1 class="mt-6 mb-2 font-semibold text-3xl">` but none of those utilities resolve
- Lists have no markers — `list-inside list-decimal/disc` don't resolve
- Inline code has no background — `bg-muted px-1.5 py-0.5 font-mono text-sm rounded` don't resolve
- Tables have no borders/padding — `bg-muted/80`, `divide-y divide-border`, `border-border border-b`, etc. don't resolve

Root cause:

1. CopilotKit's bundled Tailwind build uses the `cpk:` prefix (introduced to fix #1857 / #366 — CSS clashing with consumer Tailwind setups). So the generated rules are `.cpk\:font-semibold`, `.cpk\:bg-muted`, etc.
2. Streamdown (CopilotKit's default markdown renderer, via `CopilotChatAssistantMessage.MarkdownRenderer`) emits **unprefixed** Tailwind utility classes.
3. The unprefixed classes Streamdown emits don't match any rule in CopilotKit's prefixed CSS, so they render as no-ops.

Streamdown's README expects the consumer to add `@source '../node_modules/streamdown/dist/*.js'` to their own Tailwind globals — but CopilotKit is presented as a batteries-included integration that doesn't require the consumer to set up Tailwind.

The `cpk:prose` wrapper styles still apply because they cascade by element type (`.cpk\:prose :where(h1)`, etc.), but those alone don't cover everything Streamdown emits (no rules for `font-semibold`, `bg-muted`, list markers via `list-inside`, etc.) — only loosely styled headings and paragraph spacing come through.

Possible fixes (in rough order of effort):
- Ship an additional CSS file that includes the unprefixed utilities Streamdown emits (or merge them into the existing styles.css)
- Build/wrap Streamdown so its internal classes use the `cpk:` prefix
- Document loudly that non-Tailwind consumers need to either (a) install Tailwind and scan streamdown source, or (b) provide a custom `markdownRenderer` slot

### Screenshots

_No response_

### 𝌚 CopilotKit Version

```shell
1.56.3
```

### 📄 Logs (Optional)

```shell

```

Tushar-Khandelwal-2004:
## Summary

Fixes #5072 

Fix unstyled Streamdown markdown in non-Tailwind apps that import `@copilotkit/react-core/v2/styles.css`.

Streamdown renders plain utility class names, but CopilotKit’s CSS bundle is generated with the `cpk:` Tailwind prefix. That left default markdown elements like headings, lists, strong text, tables, blockquotes, inline code, and code blocks without matching styles unless the host app also shipped unprefixed Tailwind utilities.

## Changes

-> Add scoped fallback styles for Streamdown’s stable `data-streamdown` elements under `[data-copilotkit]`.
-> Use CopilotKit theme variables and prefixed Tailwind utilities so styles stay inside the CopilotKit surface.
-> Remove the ineffective Streamdown `@source` scan, since it only generated prefixed utilities that Streamdown does not emit.
-> Add a regression test to ensure the scoped Streamdown selectors remain present.

## Testing

-> `node node_modules\nx\bin\nx.js run @copilotkit/react-core:build:css --excludeTaskDependencies`
-> `node node_modules\nx\bin\nx.js run @copilotkit/react-core:test --excludeTaskDependencies --skip-nx-cache -- src/v2/styles/__tests__/streamdown-styles.test.ts`

### DIFF
--- a/packages/react-core/src/v2/styles/__tests__/streamdown-styles.test.ts
+++ b/packages/react-core/src/v2/styles/__tests__/streamdown-styles.test.ts
@@ -1,0 +1,23 @@
+import { readFileSync } from "node:fs";
+import { describe, expect, it } from "vitest";
+
+const globalsCss = readFileSync("src/v2/styles/globals.css", {
+  encoding: "utf8",
+});
+
+describe("Streamdown styles", () => {
+  it("ships scoped fallback styles for default markdown elements", () => {
+    expect(globalsCss).toContain(
+      '[data-copilotkit] [data-streamdown="strong"]',
+    );
+    expect(globalsCss).toContain(
+      '[data-copilotkit] [data-streamdown="unordered-list"]',
+    );
+    expect(globalsCss).toContain(
+      '[data-copilotkit] [data-streamdown="heading-1"]',
+    );
+    expect(globalsCss).toContain(
+      '[data-copilotkit] [data-streamdown="code-block"]',
+    );
+  });
+});
--- a/packages/react-core/src/v2/styles/__tests__/streamdown-styles.test.ts
+++ b/packages/react-core/src/v2/styles/__tests__/streamdown-styles.test.ts
@@ -19,5 +19,14 @@ describe("Streamdown styles", () => {
     expect(globalsCss).toContain(
       '[data-copilotkit] [data-streamdown="code-block"]',
     );
+    expect(globalsCss).toContain(
+      '[data-copilotkit] [data-streamdown="mermaid-block"]',
+    );
+    expect(globalsCss).toContain(
+      '[data-copilotkit] [data-streamdown="subscript"]',
+    );
+    expect(globalsCss).toContain(
+      '[data-copilotkit] [data-streamdown="superscript"]',
+    );
   });
 });
--- a/packages/react-core/src/v2/styles/globals.css
+++ b/packages/react-core/src/v2/styles/globals.css
@@ -297,6 +297,15 @@
   [data-copilotkit] [data-streamdown="code-block-download-button"] {
     @apply cpk:cursor-pointer cpk:p-1 cpk:text-muted-foreground cpk:transition-all cpk:hover:text-foreground cpk:disabled:cursor-not-allowed cpk:disabled:opacity-50;
   }
+
+  [data-copilotkit] [data-streamdown="mermaid-block"] {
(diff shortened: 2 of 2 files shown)


onsclom (approved):
Thank you for the great solution and tests in this PR! Was able to repro the problem locally and see this fixes most cases.

I pushed up three more selectors that 1.6.11 also emits: mermaid-block, subscript, superscript, plus the matching test assertions.

Heads up that some Streamdown chrome (loading spinner, code block language dropdown, mermaid fullscreen modal, error states) uses unprefixed Tailwind classes without data-streamdown attributes, so attribute selectors can't reach those. We'll do a follow-up with a build-scan approach: a second Tailwind pass over Streamdown's chunk files that scopes the output under [data-copilotkit]. Shouldn't block this PR.

Thanks again!

Tushar-Khandelwal-2004:
Thanks a lot for reviewing and for adding the missing selectors/tests. Glad the repro confirmed the fix for the main cases.

tylerslaton (approved):

PR #5132, opened 2026-05-31
docs(examples): add Vest MCP integration starter

itairegev:
## What

Adds `examples/integrations/vest-mcp` — a minimal Next.js starter that connects a CopilotKit `BuiltInAgent` to the hosted [Vest](https://www.getvest.ai) MCP server via `MCPAppsMiddleware`, so agents can recommend AI/SaaS tools and surface live cashback offers instead of guessing from training data.

This is the reference integration offered in #5117.

## Shape

Vest is a **public, hosted** streamable-HTTP MCP server, so unlike the [`mcp-apps`](../mcp-apps) starter there is no local MCP server to spawn — the middleware just points at `https://mcp.getvest.ai/mcp`:

```ts
new MCPAppsMiddleware({
  mcpServers: [{ type: "http", url: "https://mcp.getvest.ai/mcp", serverId: "vest" }],
});
```

The structure mirrors the existing `mcp-apps` example (same configs, `BuiltInAgent` + `ExperimentalEmptyAdapter` + `copilotRuntimeNextJSAppRouterEndpoint`), minus the local-server scaffolding.

## Notes

- Tool names in the README/agent prompt (`search_tools`, `get_cashback_estimate`, `get_stack`, …) are taken from the **live** `tools/list` on the endpoint, not assumed.
- `@copilotkit/*` pinned to `1.59.2`; an `overrides` entry pins `@ag-ui/client` to `0.0.53` so `@ag-ui/mcp-apps-middleware` and `@copilotkit/runtime` resolve the same `AbstractAgent` type (without it, `agent.use(middleware)` fails to typecheck).

## Verification

- `npx tsc --noEmit` — passes
- `npm run build` (Next.js production build) — passes; `/api/copilotkit` builds as a dynamic route

## Heads-up for maintainers

CONTRIBUTING asks contributors to reach out before significant work — opening this as the concrete artifact behind #5117 to make the discussion easy to evaluate. Happy to adjust the location/scope (or close) per your direction.


hansmbakker:
This does not add value, but it is a marketing issue.

BenTaylorDev:
Thanks for putting this together — the starter is cleanly built, and I appreciate that you read the tool names off a live `tools/list` rather than assuming them.

I'm closing it, and the reason isn't the code.

Vest's model is affiliate commission: tool vendors pay Vest for referrals, and Vest shares that back as cashback. That's a legitimate business, but it means an agent wired to this MCP server recommends products on a commission basis. Shipping that under `examples/integrations/` would put CopilotKit's name behind monetized recommendations as a reference pattern, and would set a precedent we'd have to extend to any affiliate network that asked. That's not a line I want our examples directory to cross, independent of how good the integration is.

A few smaller things I noticed, none of which are the reason:

- The body cites #5117 as where this integration was requested, but that number doesn't resolve to an issue or PR in this repo.
- 12,438 of the 12,800 added lines are `package-lock.json`. The example itself is roughly 360 lines. A committed npm lockfile inside a pnpm workspace is something we'd want to sort out regardless.
- It pins `@copilotkit/*` to `1.59.2` against a current `1.69.3`, and adds an `overrides` entry pinning `@ag-ui/client` to `0.0.53` — that override is a known source of duplicate `@ag-ui/client` copies in a consumer's tree.

The gap you were aiming at is real, though. Our `mcp-apps` starter only demonstrates spawning a local MCP server, and we have nothing showing the hosted streamable-HTTP shape — which is the more common production case and needs no local scaffolding. A version of this example built against a neutral public MCP server, with no commercial relationship to the recommendations it surfaces, is something I'd review and would genuinely like to have.

Thanks for the work, and sorry for the outcome.

PR #5138, opened 2026-06-01
fix: handle string/list emit_tool_calls in copilotkit_customize_config

### LINKED ISSUE #5120: 🐛 Bug:  emit_tool_calls parameter of copilotkit_customize_config does not work when specify it  a string or list of strings to emit only specific tool calls
### ♻️ Reproduction Steps

1.  I just want to use emit_tool_calls as a whitelist to specify which tool calling can emit tool events to frontend, and I configured it as following example:

   from copilotkit.langgraph import copilotkit_customize_config
        config = copilotkit_customize_config(
            config,
            emit_tool_calls=["draft_email_structured"],  # only show this one
        )
        # ---- Invoke LLM ----
        log_llm_input(logger, llm_messages, tag="supervisor")
        response = await llm.bind_tools(all_tools).ainvoke(llm_messages,config)
        log_llm_output(logger, response, tag="supervisor")

2.  Then when llm try call list_polish_operations , it still emit events to frontend;
3.  And I found that if I only set   emit_tool_calls=False, the frontend "useDefaultRenderTool" can still capture the tool call result eventually, seem it is caused by  event "MESSAGES_SNAPSHOT" to sync all backend status to frontend finally.
4. config = copilotkit_customize_config(
            config,
            emit_tool_calls=False, 
        )
5. ...

### ✅ Expected Behavior

expect that the "whitelists"  emit_tool_calls=["draft_email_structured"] could filter out all other tool call events.

### ❌ Actual Behavior

useDefaultRenderTool could capture all events finally, even I set  emit_tool_calls=["draft_email_structured"] or  emit_tool_calls=False

### Screenshots

_No response_

### 𝌚 CopilotKit Version

```shell
1.57.1
```

### 📄 Logs (Optional)

```shell

```

hashwnath:
## Summary

Fixes #5120

- The `emit_tool_calls` parameter of `copilotkit_customize_config` accepts `bool | str | list[str]`, but the AG-UI event dispatching logic only checked for `False`, silently ignoring string/list whitelist values.
- Added `_should_emit_tool_call` (Python) and `shouldEmitToolCall` (TypeScript) methods that properly handle all three types -- matching the existing v1 `RemoteLangGraphEventSource.shouldEmitToolCall` implementation.
- Tool call names are tracked by `tool_call_id` so that `ToolCallArgs` and `ToolCallEnd` events (which don't carry the tool name) can be filtered correctly.

### Files changed
- `sdk-python/copilotkit/langgraph_agui_agent.py` -- Python AG-UI agent filtering
- `packages/runtime/src/lib/runtime/agent-integrations/langgraph/agent.ts` -- TypeScript AG-UI agent filtering
- `sdk-python/tests/test_emit_tool_calls.py` -- 18 unit tests for the filtering logic

## Test plan

- [x] 18 pytest tests covering bool/string/list filtering, ID-based lookup, and unknown-name fallback
- [ ] Manual verification with a LangGraph agent using `emit_tool_calls=["specific_tool"]`

### DIFF
--- a/packages/runtime/src/lib/runtime/agent-integrations/langgraph/__tests__/dispatch-event-filtering.test.ts
+++ b/packages/runtime/src/lib/runtime/agent-integrations/langgraph/__tests__/dispatch-event-filtering.test.ts
@@ -343,3 +343,115 @@ describe("langGraphDefaultMergeState", () => {
     expect(result.copilotkit.actions[0].name).toBe("SharedTool");
   });
 });
+
+// ---------- emit-tool-calls: string/list filtering ----------
+
+function makeToolEvent(
+  type: EventType,
+  metadata: Record<string, any>,
+  toolCallId = "tc-1",
+  toolCallName?: string,
+) {
+  return {
+    type,
+    toolCallId,
+    ...(toolCallName ? { toolCallName } : {}),
+    ...(type === EventType.TOOL_CALL_START ? { parentMessageId: "msg-1" } : {}),
+    rawEvent: { metadata },
+  };
+}
+
+describe("emit-tool-calls string/list filtering", () => {
+  it("emits only the named tool when emit-tool-calls is a string", () => {
+    const { agent } = createAgent();
+    const meta = { "copilotkit:emit-tool-calls": "draft_email" };
+
+    const matchResult = agent.dispatchEvent(
+      makeToolEvent(EventType.TOOL_CALL_START, meta, "tc-1", "draft_email") as any,
+    );
+    expect(matchResult).toBe(true);
+
+    const noMatchResult = agent.dispatchEvent(
+      makeToolEvent(EventType.TOOL_CALL_START, meta, "tc-2", "list_items") as any,
+    );
+    expect(noMatchResult).toBe(false);
+  });
+
+  it("emits only listed tools when emit-tool-calls is an array", () => {
+    const { agent } = createAgent();
+    const meta = { "copilotkit:emit-tool-calls": ["draft_email", "send_email"] };
+
+    expect(
+      agent.dispatchEvent(
+        makeToolEvent(EventType.TOOL_CALL_START, meta, "tc-1", "draft_email") as any,
+      ),
+    ).toBe(true);
+
+    expect(
+      agent.dispatchEvent(
+        makeToolEvent(EventType.TOOL_CALL_START, meta, "tc-2", "send_email") as any,
+      ),
+    ).toBe(true);
+
+    expect(
+      agent.dispatchEvent(
+        makeToolEvent(EventType.TOOL_CALL_START, meta, "tc-3", "delete_item") as any,
+      ),
+    ).toBe(false);
+  });
+
+  it("resolves name via toolCallNames map for TOOL_CALL_ARGS events", () => {
+    const { agent } = createAgent();
+    const meta = { "copilotkit:emit-tool-calls": ["search"] };
+
+    // First dispatch START to populate the name map
+    agent.dispatchEvent(
+      makeToolEvent(EventType.TOOL_CALL_START, meta, "tc-10", "search") as any,
+    );
+
+    // TOOL_CALL_ARGS has toolCallId but not toolCallName — should resolve via map
+    const argsResult = agent.dispatchEvent({
+      type: EventType.TOOL_CALL_ARGS,
+      toolCallId: "tc-10",
+      delta: '{"query":"test"}',
+      rawEvent: { metadata: meta },
+    } as any);
+    expect(argsResult).toBe(true);
+
+    // A different id that was never tracked — unknown name defaults to emit
+    const unknownResult = agent.dispatchEvent({
+      type: EventType.TOOL_CALL_ARGS,
+      toolCallId: "tc-unknown",
+      delta: "{}",
+      rawEvent: { metadata: meta },
+    } as any);
+    expect(unknownResult).toBe(true);
+  });
+
+  it("cleans up toolCallNames on TOOL_CALL_END to prevent memory leak", () => {
+    const { agent } = createAgent();
+    const meta = { "copilotkit:emit-tool-calls": true };
+
+    // START populates the map
+    agent.dispatchEvent(
+      makeToolEvent(EventType.TOOL_CALL_START, meta, "tc-leak", "my_tool") as any,
+    );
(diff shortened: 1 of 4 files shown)


ranst91 (changes requested):
Thanks for the contribution! Few things stand out:

1. Memory leak: `toolCallNames` / `_tool_call_names` is never cleared. No delete on TOOL_CALL_END, no reset between runs. On a long-lived agent it grows unbounded. Suggest popping the entry on TOOL_CALL_END, or clearing on RUN_STARTED.

2. Python and TS diverge on missing names. Python uses `hasattr(event, 'tool_call_name')`, which is true even when the value is `None`, so the `elif` id-lookup never runs and the event emits unconditionally. TS handles this correctly with `??`. Suggest matching the TS behavior:
   ```python
   tool_call_name = getattr(event, 'tool_call_name', None)
   if tool_call_name is None:
       tool_call_name = self._tool_call_names.get(getattr(event, 'tool_call_id', None))
   ```

3. The tests exercise a standalone copy of `_should_emit_tool_call` with a different signature, not the real method. It's never imported, so the logic can drift while the tests stay green. The `None`-name case above wouldn't be caught.

4. No TS tests, and the manual-verification box is unchecked. The TS path is where the logic diverges from Python, so it's the most worth covering.

5. Minor: the triple `as unknown as` cast in TS could be a typed narrow on `event.type`. Worth confirming the "matches v1" claim holds for empty-list semantics.

Highest priority is 1 and 2.

hashwnath:
Hey, pushed fixes for all 5 memory leak, the None fallback, tests against the real method, TS coverage, and the casts in 91f6368. Thanks

BenTaylorDev (changes requested):
Thanks for circling back on this — items 2 (Python `None` handling) and 4 (TS tests) are properly fixed, and importing the real `should_emit_tool_call` is the right move for item 3. Three things still block, one of which is new and not your fault.

I reviewed this by merging the branch onto current `main` and running it, since the branch is stale enough that the GitHub diff doesn't reflect what would actually land.

**1. `toolCallNames` doesn't survive `clone()` — crashes every tool call on the v2 runtime**

This one appeared underneath you: the v2 runtime clones the registered agent per request (`fetch-handler.ts:553` → `handle-run.ts:36` → `cloneAgentForRequest` in `v2/runtime/handlers/shared/agent-utils.ts:78`). Both `AbstractAgent.clone()` and `@ag-ui/langgraph`'s override build the copy with `Object.create(getPrototypeOf(this))` and assign an explicit allowlist of fields — the constructor never runs, so `toolCallNames` is `undefined` on the clone.

Driving the real helper with the registration shape from `examples/integrations/langgraph-python/src/agent.ts`:

```
TypeError: Cannot read properties of undefined (reading 'set')
 ❯ LangGraphAgent.dispatchEvent src/lib/runtime/agent-integrations/langgraph/agent.ts:156:28
       this.toolCallNames.set(toolCallId, toolCallName);
```

The event metadata in that repro was `{}` — no `copilotkit_customize_config` anywhere. So this throws on the first tool call for every v2 LangGraph user regardless of whether they use `emit_tool_calls`.

Lazy-init (`this.toolCallNames ??= new Map()`) is the smallest fix. Scoping the map to `activeRun` instead of the instance would fix this and item 2 together, and is probably where this wants to end up.

**2. The leak cleanup misses the suppressed path**

The `delete` sits after the filter, but the filter returns first:

```ts
if (!this.shouldEmitToolCall(emitToolCalls, resolvedName)) {
  return false;          // <-- skips the cleanup below
}
...
if (event.type === EventType.TOOL_CALL_END) { this.toolCallNames.delete(toolCallId); }
```

Same shape in Python (`return None` at line 296, `pop` at 301). So entries are only reclaimed for tool calls that are *emitted* — never for the ones filtered out, which is the whole point of the flag. Running 200 suppressed + 200 emitted full START→ARGS→END lifecycles through the real `dispatchEvent` / `_dispatch_event`:

```
TS:     suppressed retained=200, emitted retained=0
Python: suppressed retained=200, emitted retained=0
```

Worth confirming the premise, since it's what makes this reachable: `@ag-ui/langgraph` attaches the *same* stream event as `rawEvent` to START, ARGS and END (`agent.py:1317/1326/1335`), so `copilotkit:emit-tool-calls` is present on END and the filter really does swallow the cleanup.

The existing TS test can't catch this because it uses `"copilotkit:emit-tool-calls": true`, the one config where cleanup runs. Clearing on `RUN_STARTED` handles both this and tool calls orphaned by an aborted run, without the ordering constraint.

**3. `tests/test_emit_tool_calls.py` doesn't collect on its own**

```
$ pytest tests/test_emit_tool_calls.py
E   AttributeError: Mock object has no attribute '__annotations__'
    Interrupted: 1 error during collection
```

The `sys.modules.setdefault(..., MagicMock())` block poisons `langchain.agents.middleware` for the whole session. Full-suite runs are green (232 passed) only because alphabetically-earlier files import the real modules first, making `setdefault` a no-op — force this file first and collection of `test_agui_agent.py` and `test_copilotkit_lg_middleware.py` breaks too. Dropping the block makes the file pass standalone (22 passed).

The premise behind it is also out of date: `sdk-python/tests/test_emit_filtering.py` already instantiates a real `LangGraphAGUIAgent(name=..., graph=MagicMock())` and drives `_dispatch_event` with no module mocking. Adding these cases there instead would remove the need for the extracted module-level function, and would have caught item 2.

**Smaller things**

- The "matches v1" framing isn't quite right. v1's `shouldEmitToolCall` (`packages/runtime/src/agents/langgraph/event-source.ts:34`) has no unknown-name branch — an unresolvable name falls through to `includes()`/`===` and is suppressed. The new `if (!toolCallName) return true` fails *open* on a suppression feature. Defensible, but it's a deliberate divergence and reads better as a comment than as parity.
- `event as unknown as ToolCallStartEvent` plus the second inline structural cast could be one narrow on `event.type`.
- The branch conflicts with `main` in `langgraph_agui_agent.py`'s `__init__` (your `_tool_call_names` vs. a newly added `_copilotkit_runtime_payload`) — trivial, adjacent lines.

The underlying fix is right and the bug is still live on `main` (`agent.ts:135`, `langgraph_agui_agent.py:242`), with all three read sites correctly accounted for. It's mainly the instance-state handling that needs another pass. The thing I'd most want to see asserted is that the map is empty after a run in which *every* tool call was filtered — through `_dispatch_event`, not the extracted helper.

PR #5152, opened 2026-06-01
feat(showcase): spec-driven, fail-closed D6 measurement (LGP green)

jpr5:
## Summary

Replaces the showcase D6 measurement from the harness's generic conversation-runner **heuristic** (which counted `[data-testid="copilot-assistant-message"]` DOM nodes) with **running each integration's own Playwright e2e specs**, rolled into PocketBase **fail-closed**. A green D6 cell now means the feature's real e2e spec passed — not "a node appeared."

### Why
The heuristic produced verified false signals in both directions:
- **False green:** a rollup short-circuit greened cells with no passing run (the "484 green cells" no run ever produced).
- **False red:** `gen-ui-interrupt` rendered correctly but the interrupt element is a *sibling* of the counted node → `waitForAssistantSettled` timed out → "assistant did not respond" on a working feature.

### What's in this PR (the engine + LGP)
- **Foundation:** spec→cell mapping (strict 1:1 + completeness/uniqueness CI guard), Playwright JSON-reporter parser, skip-list loader.
- **Fail-closed rollup:** a cell greens **only** from an explicit PASS row. Spec ran+failed → RED; missing/errored/0-test/no-row → UNKNOWN; declared skip → SKIPPED. Never green otherwise.
- **Driver cutover:** `d6-all-pills.ts` now flows run → parse → rollup → write; no node-counting in the pass/fail path. (`conversation-runner.ts` retained pending a later retirement task.)
- **2 specs promoted into LGP gold:** `hitl`, `interrupt-headless` (LGP had the routes+QA but no specs) — both genuinely pass.

### Validation (GATE A, live)
Drove the real driver against the live stack: **LGP = 40 spec files / 193 cases → 191 passed, 0 failed, 2 skipped; 40/40 cells GREEN; aggregate green.** Fail-closed proven live: a forced break → that cell RED + un-run cells UNKNOWN; a 0-test run → all UNKNOWN. Green cannot be manufactured.

### Scope / expected dashboard behavior
The driver runs all 18 integrations' suites. **LGP is green; the others show honest red/unknown** (fail-closed) until Phase 2 aligns them. This is intentional — green spreads as each integration is closed.

### Plan
[D6 Measurement Migration](https://www.notion.so/3723aa38185281af9197e2b17a27ad8e) (Notion, Plans/Proposals).

### Known follow-up (for CR)
The dead `launcher` arg + unused `E2eFullDriverDeps` fields (left after the cutover) should be removed once the cascade across all 4 driver call sites is owned. Deferred — not behavior-affecting.

## Test plan
- [x] Foundation helper suite (mapping/parser/skip-list/rollup) TDD red-green; harness vitest 1741 passing; tsc + oxfmt + oxlint clean; harness build clean.
- [x] GATE A: LGP 40/40 green through the new path, live; fail-closed verified live.
- [ ] Post-merge: staging dashboard shows real LGP green (Playwright-verified) before production promote.


BenTaylorDev:
Closing as superseded by #5872, which is your own re-shape of this work.

Re-checked against current `main` before deciding:

**The premise is half-resolved.** The false-green half is still live — `aggregateGreen = failed.length === 0` remains at `showcase/harness/src/probes/drivers/d6-all-pills.ts:1572`. The false-red half is already fixed by different work: the count-baseline `waitForAssistantSettled` gate was deleted in Phase 5 of the bubble-race-fix plan, and the live settle primitive is now the three-conjunct `waitForTurnComplete` (SSE + DOM + TEXT). So the `gen-ui-interrupt` sibling-node timeout this PR was partly motivated by no longer reproduces.

**The branch has gone stale against `main`.** It conflicts across 30 files, including the ones this change is built on: `d6-all-pills.ts` (rewritten by 12 commits since the branch point), `browser-pool.ts`, `types/index.ts`, `writers/status-writer.ts`, `shell-dashboard/src/lib/cell-model.ts`, `live-status.ts`.

**Three pieces have been overtaken specifically:**

- `probe_runs.triggered` optional already landed as `showcase/pocketbase/pb_migrations/1779989800_probe_runs_triggered_optional.js` — functionally the same ALTER.
- Both new migrations reuse prefixes `main` now occupies: `1779989300_*` (`create_resource_snapshots.js`) and `1779989400_*` (`create_probe_jobs.js`).
- The strict 1:1 spec→cell mapping would fail its own completeness guard — `a2ui-recovery.spec.ts` landed after the branch point and is unmapped.

**#5872 is the better vehicle for the same engine** — same `pw-json-reporter.ts` / `d6-rollup.ts` / `cli/e2e.ts` / mapping guard, but N:M rollup instead of strict 1:1, an empty `spec-driven-slugs.json` so the cutover is a proven no-op, and LGP deferred to a Phase 1 PR behind a live A/B equivalence gate. Nothing here is lost; it lands in a safer shape.

The fail-closed problem is still real and still unfixed on `main`, so #5872 is worth picking back up.

Issue #3044, opened 2026-01-21
🐛 Bug:"UseFrontendTool" has fallen into a loop.


LengYXin:
### ♻️ Reproduction Steps

I discovered a strange bug. All I had to do was add 'inputProcessors' to the Agent, and then call 'useFrontendTool', which would lead to an infinite loop /api/copilotkit interface call and the call of 'useFrontendTool'. The code only required two modifications as follows. This is the mastra template created by the official command.

page.tsx
```tsx
  useFrontendTool({
    name: 'test_bid_wizard',
    description: 'A test bid wizard tool.',
    parameters: [],
    render: () => <div className="h-14 bg-amber-500">test_bid_wizard</div>,
  });
```
```ts
import type {
  ProcessInputArgs,
  ProcessInputResult,
  Processor
} from '@mastra/core/processors';

class ExampleInputProcessor implements Processor {
  readonly id = 'example-input-processor';
  readonly name = 'Example Input Processor';

  async processInput({ messages, requestContext }: ProcessInputArgs): Promise<ProcessInputResult> {
    console.log("🚀 ~ ExampleInputProcessor ~ processInput ~ messages:", messages)
    // 返回原始消息，继续后续处理
    return messages;
  }
}

export const weatherAgent = new Agent({
 // ....
  inputProcessors: [new ExampleInputProcessor()],
  // ....
});

```

### ✅ Expected Behavior

![Image](https://github.com/user-attachments/assets/b3b90d94-0cd2-4ada-9150-f38d21c1de35)

### ❌ Actual Behavior

<img width="1033" height="1091" alt="Image" src="https://github.com/user-attachments/assets/4044ea6c-0b59-44e3-aa76-89299b813c9a" />

<img width="961" height="554" alt="Image" src="https://github.com/user-attachments/assets/fda4cc32-f6a2-4a71-9082-ad3aafa7fa97" />

### 𝌚 CopilotKit Version

```shell
├── @copilotkit/react-core@1.51.2
├── @copilotkit/react-ui@1.51.2
├── @copilotkit/runtime@1.51.2

├── @ag-ui/mastra@0.2.1-beta.2 invalid: "beta" from the root project
├── @mastra/client-js@1.0.1
├── @mastra/core@1.0.4
├── @mastra/libsql@1.0.0
├── @mastra/memory@1.0.0
├── mastra@1.0.1
```

### 📄 Logs (Optional)

```shell

```

BenTaylorDev:
[...]

Issue #2416, opened 2025-09-04
🐛 Bug: FE tools repeat indefinitely if toolresult run doesn't terminate with success.


tylerslaton:
### ♻️ Reproduction Steps

We have a custom AG_UI backend agent with a simple `Copilotkit FE Action` setup like this:

```ts
  useCopilotAction({
    available: "remote",
    name: "show_stock_timeseries",
    description: "Display a stock price timeseries visualization",
    parameters: [
      {
        name: "symbol",
        type: "string",
        required: true,
      },
    ],

    // Anything returned in `handler` will be passed back to the Agent as a Tool Response
    handler: async ({ symbol }) => {
      if (!(await validateSymbol(symbol))) {
        return {
          success: false,
          error: `${symbol} is not a valid ticker symbol.  You must provide a valid US stock symbol`,
        };
      }
      return { success: true };
    },
    render: ({ status, args }) => {
      return <SymbolChart symbol={args.symbol!} />;
    },
  });
```

After the FE Action renders and returns the tool result to our backend, if our backend fails for whatever reason and responds with a `RunError`, the FE Action will just continuously repeat itself as fast as our service can respond.

This effectively DOS's our service as it consumed our bedrock throttle limits very quickly.

### ✅ Expected Behavior

When a RunError is returned, it should display and halt execution.

### ❌ Actual Behavior

CopilotKit continuously repeats request

### 𝌚 CopilotKit Version

```shell
1.10.3
```

#### 1. Initial request.  User input triggers a FE action.  Backend finishes run with a toolcall. (`PyanticAI DeferredTool`):

```shell
INFO:src.agents.advice.advice_agent: Received AG-UI request:
{
  "thread_id": "ae230a89-d8e0-49fa-b213-7b9b28b0038c",
  "run_id": "4e7d30ff-68f6-461f-8693-3c506e6fe707",
  "state": {},
  "messages": [
    {
      "id": "ck-6aaca01d-a96a-4031-8173-ce816bba6b47",
      "role": "user",
      "content": "show me the apple stock visualization",
      "name": null
    }
  ],
  "tools": [
    {
      "name": "show_stock_timeseries",
      "description": "Display a stock price timeseries visualization",
      "parameters": {
        "type": "object",
        "properties": {
          "symbol": {
            "type": "string"
          }
        },
        "required": [
          "symbol"
        ]
      }
    }
  ],
  "context": [],
  "forwarded_props": {
    "config": {},
[...]

PR #5158, opened 2026-06-02
fix(core): add circuit breaker to prevent infinite follow-up loops

### LINKED ISSUE #4819: 🐛 Bug: processAgentResult has no circuit breaker > recursive follow-ups can loop indefinitely

18062706139fcz:
## Summary

Fixes #4819.

`processAgentResult` recurses into `runAgent` whenever `needsFollowUp` is true, with **no depth limit, no circuit breaker, no duplicate detection**. Any scenario that keeps `needsFollowUp = true` loops forever:

- The LLM repeatedly calling the same tool
- The backend returning a `RunError` after receiving tool results (#2416)
- Input processors reprocessing tool messages (#3044)

This silently consumes API quota and can DOS the backend. The only existing workaround — setting `followUp: false` on a tool — is too blunt and breaks legitimate multi-step workflows.

## Change

- Add an absolute safety cap `MAX_FOLLOW_UP_DEPTH = 100`, reusing the existing `_runDepth` counter (incremented in `runAgent`, decremented in its `finally`).
- In `processAgentResult`, when the depth cap is reached: stop recursing, emit a `logger.warn` (for debuggability), and return the current result normally (already-inserted tool messages are preserved).
- The cap is deliberately high so legitimate multi-step workflows (search → fill form → confirm → update → send email) are never affected; it only trips on runaway recursion.

This protects against **all** infinite-loop scenarios in the issue, not just specific triggers, without requiring users to set `followUp: false`.

## Test

- Added `core-full.test.ts` TEST 9b: an agent that always returns a fresh tool call (so `needsFollowUp` stays true). Asserts the run terminates, `runAgent` is capped at ≤100 calls, and a warning is emitted.
- `nx run @copilotkit/core:test` → 40 files / 431 tests pass.
- `nx run @copilotkit/core:build` → success.

### DIFF
--- a/packages/core/src/__tests__/core-full.test.ts
+++ b/packages/core/src/__tests__/core-full.test.ts
@@ -241,6 +241,49 @@ describe("CopilotKitCore.runAgent - Full Test Suite", () => {
       }
     });
 
+    it("TEST 9b: should stop recursive follow-ups at the depth limit instead of looping forever", async () => {
+      const warnSpy = vi
+        .spyOn(console, "warn")
+        .mockImplementation(() => undefined);
+
+      // A tool that always requests a follow-up. The agent always returns a
+      // fresh tool call for it, so without a circuit breaker processAgentResult
+      // would recurse into runAgent indefinitely.
+      const tool = createTool({
+        name: "loopingTool",
+        handler: vi.fn(async () => "Result"),
+        followUp: true,
+      });
+      copilotKitCore.addTool(tool);
+
+      const agent = new MockAgent({
+        newMessages: [createToolCallMessage("loopingTool")],
+      });
+      copilotKitCore.addAgent__unsafe_dev_only({
+        id: "test",
+        agent: agent as any,
+      });
+
+      // Every run keeps emitting a brand-new tool call so needsFollowUp stays true.
+      agent.runAgentCallback = () => {
+        agent.setNewMessages([createToolCallMessage("loopingTool")]);
+      };
+
+      const result = await copilotKitCore.runAgent({ agent: agent as any });
+
+      // It must terminate rather than hang, capped by the absolute depth limit (100).
+      expect(agent.runAgentCalls.length).toBeGreaterThan(1);
+      expect(agent.runAgentCalls.length).toBeLessThanOrEqual(100);
+      expect(result).toBeDefined();
+
+      // A warning must be emitted when the breaker trips, for debuggability.
+      expect(warnSpy).toHaveBeenCalledWith(
+        expect.stringContaining("Follow-up depth limit"),
+      );
+
+      warnSpy.mockRestore();
+    });
+
     it("TEST 10: should handle concurrent tool calls", async () => {
       const delays = [50, 30, 70];
       const tools = delays.map((delay, i) =>
--- a/packages/core/src/core/run-handler.ts
+++ b/packages/core/src/core/run-handler.ts
@@ -69,6 +69,20 @@ interface ExecuteToolHandlerResult {
 }
 
 /**
+ * Absolute safety cap on the depth of recursive follow-up runs triggered by
+ * `processAgentResult`. Without this, any scenario that keeps
+ * `needsFollowUp = true` (e.g. the LLM repeatedly calling the same tool, a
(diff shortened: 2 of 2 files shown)


BenTaylorDev (approved):
**Approving.** Adversarially reviewed against current `main`:

- **Termination is guaranteed for every loop shape.** `_runDepth` increments on each `runAgent` entry (`run-handler.ts:409`) and decrements in its `finally` (`:440`), so it grows one frame per nested follow-up *regardless of tool-call content* — the `>= MAX_FOLLOW_UP_DEPTH` check bounds identical, alternating, and varying-args loops alike. This is why the absolute cap is the right primitive: it can't be evaded by changing arguments.
- **Degrades gracefully.** On trip it doesn't throw or recurse — it falls through, preserves the already-inserted tool messages, returns the current result, and emits a clear `logger.warn`. Correct behavior for a runaway rather than a hang.
- **Genuinely tested.** The new case drives a real `MockAgent` that keeps emitting fresh tool calls through the real `addTool`/`runAgent` path, asserts the run count stays within the cap, and confirms the warning fired (`logger` is `console` in `@copilotkit/shared`, so the `console.warn` assertion is real). Nothing mocks away the recursion.
- Merges clean, and the bug is still live on `main` (no cap at the follow-up site today).

One non-blocking note for later: `_runDepth` is a single per-`RunHandler` counter, so concurrent runs on one core would share it — but `_runAbortController` already is too, so this isn't a regression this PR introduces.

BenTaylorDev:
Ran `@copilotkit/core`'s checks locally with this change rebased onto current `main` (the fork's CI didn't auto-trigger):

- **Tests:** 583/583 pass across 53 files, including the new depth-limit test (`core-full.test.ts` TEST 9b).
- **Type-check** (`tsc --noEmit`): clean.
- **Lint** (`oxlint`): 0 errors; the only warnings are pre-existing and unrelated to this change.

No regressions. `@copilotkit/core` is the only package this PR touches.

PR #5161, opened 2026-06-02
feat(sdk): auto-inject A2UI tool in CopilotKitMiddleware

ranst91:
Prebuilt agents get dynamic A2UI with no extra wiring — adding the middleware is enough. When the frontend registers an A2UI catalog (surfaced by the runtime into state["ag-ui"].a2ui_schema), the middleware infers the agent's own model, advertises the generate_a2ui tool in the model-call hook, and executes it in the tool-call hook. No catalog → the tool is never advertised.

Covers both @copilotkit/sdk-js and the copilotkit Python SDK. Bumps the A2UI tool-factory dependency to where get_a2ui_tools ships (@ag-ui/langgraph 0.0.35, ag-ui-langgraph >=0.0.37).


showcase/integrations/langgraph-typescript/src/agent/package.json
```diff
@@ -7,7 +7,7 @@
     "start": "node --import tsx liveness.mjs"
   },
   "dependencies": {
-    "@copilotkit/sdk-js": "1.59.2",
+    "@copilotkit/sdk-js": "1.59.3-alpha.1",
```
AlemTuzlak: Is this right?

AlemTuzlak (dismissed):

showcase/integrations/langgraph-typescript/src/agent/package.json
ranst91: Yes. Once merged I'll release a full version and update to non alpha in a different PR

tylerslaton (changes requested):
Today users can set injectA2UITool: false when they do not want CopilotKit to add an A2UI tool. This PR bypasses that setting because the SDK middleware now adds `generate_a2ui` anyway.
 
That can break existing apps where the backend already defines its own `generate_a2ui` tool. In that case the model may see two tools with the same name, which is exactly what some existing route comments say they were trying to avoid.

My ask here:
1. Add an explicit on/off option for this new middleware behavior at the creation of it.

    ```python
    # in python
    CopilotKitMiddleware(injectA2UI=False)
    ```

2. Respect existing `injectA2UITool: false` definitions from the user's runtime.
3. Do not auto-add `generate_a2ui` if the agent already has a tool with that name.

ranst91:
> Today users can set injectA2UITool: false when they do not want CopilotKit to add an A2UI tool. This PR bypasses that setting because the SDK middleware now adds `generate_a2ui` anyway.
> 
> That can break existing apps where the backend already defines its own `generate_a2ui` tool. In that case the model may see two tools with the same name, which is exactly what some existing route comments say they were trying to avoid.
> 
> My ask here:
> 
> 1. Add an explicit on/off option for this new middleware behavior at the creation of it.
>    ```python
>    # in python
>    CopilotKitMiddleware(injectA2UI=False)
>    ```
> 2. Respect existing `injectA2UITool: false` definitions from the user's runtime.
> 3. Do not auto-add `generate_a2ui` if the agent already has a tool with that name.

With the recent changes, ask number 2 (and thus, also number 3) are addressed.
As for number 1, we're trying to keep the backend clean, therefore the rule of thumb here is "everything adjustable with a2ui, you do through runtime", so I did not include 1

tylerslaton (dismissed):

tylerslaton (approved):

PR #5177, opened 2026-06-03
feat(react-native): upgrade react-native-streamdown and add react-native-enriched-markdown peer deps

hryhoriiK97:
## What does this PR do?

Upgrades `react-native-streamdown` to `>=0.2.0` and explicitly adds `react-native-enriched-markdown` (`>=0.6.0`), `react-native-worklets` (`>=0.8.3`), and `remend` (`^1.3.0`) as optional peer dependencies of `@copilotkit/react-native`.

> I am a maintainer of both `react-native-streamdown` and `react-native-enriched-markdown` at [Software Mansion](https://swmansion.com/), so I can confirm these version constraints are correct and the `flavor` prop API is stable.

`react-native-streamdown@0.2.0` introduces significantly improved streaming support for GitHub Flavored Markdown — including tables — rendered correctly even as tokens arrive token-by-token from the LLM. You can see it in action here: https://x.com/swmansion/status/2059286828797055112?s=20

**Changes:**

- **`packages/react-native/package.json`** — bumps `react-native-streamdown` peer dep from `>=0.1.0` to `>=0.2.0`; surfaces `react-native-enriched-markdown`, `react-native-worklets`, and `remend` as explicit optional peer deps (they were previously hidden transitive deps of `react-native-streamdown`)
- **`Markdown.tsx`** — adds a `flavor` prop to `CopilotMarkdown` (default: `"github"`) and passes it through to `StreamdownText`, enabling GitHub Flavored Markdown table rendering. Requires `react-native-enriched-markdown >=0.6.0`
- **`react-native-streamdown.ts` (mock)** — updates the test stub to include the `flavor` prop
- **`Markdown.test.tsx`** — adds test coverage for the `flavor` prop default and override
- **`USAGE.md`** — updated install command, added peer dep table, Bundle Mode setup callout, and `flavor` prop usage example
- **`docs/content/docs/(root)/react-native.mdx`** — updated install step with all peer deps, removed stale "Known limitations" entries (pre-built UI components and markdown rendering now exist), updated the "What's included" table

## Related PRs and Issues

- N/A

## Checklist

### DIFF
--- a/docs/content/docs/(root)/react-native.mdx
+++ b/docs/content/docs/(root)/react-native.mdx
@@ -30,25 +30,29 @@ import { Tabs, Tab } from "fumadocs-ui/components/tabs";
         ```
     </Step>
     <Step>
-        ### Install CopilotKit
+        ### Install CopilotKit and peer dependencies
 
         <Tabs groupId="package-manager" items={['npm', 'pnpm', 'yarn']}>
             <Tab value="npm">
                 ```bash
-                npm install @copilotkit/react-native
+                npm install @copilotkit/react-native react-native-streamdown react-native-enriched-markdown react-native-worklets remend
                 ```
             </Tab>
             <Tab value="pnpm">
                 ```bash
-                pnpm add @copilotkit/react-native
+                pnpm add @copilotkit/react-native react-native-streamdown react-native-enriched-markdown react-native-worklets remend
                 ```
             </Tab>
             <Tab value="yarn">
                 ```bash
-                yarn add @copilotkit/react-native
+                yarn add @copilotkit/react-native react-native-streamdown react-native-enriched-markdown react-native-worklets remend
                 ```
             </Tab>
         </Tabs>
+
+        <Callout type="warn" title="Bundle Mode setup required">
+          `react-native-streamdown` processes markdown on a background worklet thread using [Bundle Mode](https://docs.swmansion.com/react-native-worklets/docs/bundleMode/setup/) from `react-native-worklets`. You must complete the Bundle Mode Babel + Metro configuration before the markdown components will work. See the [react-native-streamdown README](https://github.com/software-mansion-labs/react-native-streamdown) for full setup instructions.
+        </Callout>
     </Step>
     <Step>
         ### Add polyfills
--- a/packages/react-native/USAGE.md
+++ b/packages/react-native/USAGE.md
@@ -5,10 +5,12 @@
 Install all required peer dependencies:
 
 ```bash
-npm install react react-native @gorhom/bottom-sheet react-native-gesture-handler react-native-reanimated react-native-streamdown
+npm install react react-native @gorhom/bottom-sheet react-native-gesture-handler react-native-reanimated react-native-streamdown react-native-enriched-markdown react-native-worklets remend
 ```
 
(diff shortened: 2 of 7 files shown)


BenTaylorDev:
Thanks for this, and for the upstream context — the `flavor` API claim checks
out. I verified it against the published tarballs rather than the types:
`flavor` is not in `react-native-streamdown`'s own declarations, but
`StreamdownTextProps extends EnrichedMarkdownTextProps` and `StreamdownText`
spreads `...enrichedMarkdownProps` through, so the prop reaches
`EnrichedMarkdownText` as documented. In `react-native-enriched-markdown` it
is stable from 0.4.0 through 1.0.2, typed `'commonmark' | 'github'`, default
`'commonmark'`. So the feature is sound and we want it.

Four things need fixing before this can land.

## 1. The streamdown bump is not in the diff

The description says the peer dep goes from `>=0.1.0` to `>=0.2.0`, but
`packages/react-native/package.json` on this branch still reads
`"react-native-streamdown": ">=0.1.0"` — that line is diff context, not a
change. Meanwhile `Markdown.tsx` passes `flavor` unconditionally, so a
consumer on 0.1.x satisfies the declared range and gets a prop that release
does not forward.

## 2. The version floors disagree with each other and with the registry

| declared here | reality today |
| --- | --- |
| `react-native-enriched-markdown >=0.4.0` | the JSDoc in `Markdown.tsx` and both doc files say `>=0.6.0`; `flavor` actually exists from 0.4.0, so the prose is the wrong half |
| `react-native-worklets >=0.8.3` | `react-native-streamdown@0.3.0` requires `>=0.10.0` |
| `remend ^1.3.0` | streamdown pins exactly `1.3.0`; `^1.3.0` admits the published 1.3.1 |
| target 0.2.0 | latest is 0.3.0, whose peers require `react-native-enriched-markdown >=1.0.0` |

As the upstream maintainer you are better placed than me to pick the target.
Whichever version it is, the peer ranges, the JSDoc, and the docs should all
name the same one.

## 3. The docs edit targets a deleted tree

`docs/content/docs/(root)/react-native.mdx` no longer exists — `docs` is now a
symlink and that tree was removed in `33a79485b6`. Docs are authored in
`showcase/shell-docs/src/content/`, so that hunk needs re-targeting to:

- `showcase/shell-docs/src/content/docs/frontends/react-native.mdx` — the
  install step, and the peer-dependency table around line 321
- `showcase/shell-docs/src/content/reference/react-native/components/CopilotMarkdown.mdx`
  — where the new `flavor` prop belongs; this PR does not touch it yet

`packages/react-native/USAGE.md` is already correct.

## 4. Please call out the default change explicitly

`'github'` is a container-based renderer, not a flag on the existing one, and
`AssistantMessage.tsx:41` renders `<CopilotMarkdown content={content} />` with
no flavor. So `flavor = "github"` switches the renderer for every assistant
message in the React Native chat. That looks like the intent, but it is a
visual behavior change for existing apps and I would like a screenshot of a
normal message thread (not just a table) on device before merging. The added
tests assert pass-through against the local `vi.mock`, which matches the
existing `streamingAnimation` tests but cannot catch a layout regression.

## Also

`pnpm-lock.yaml` will need regenerating after the rebase. Worth a look while
you are there: the current hunk adds resolved entries for these packages and
rekeys about a dozen peer-suffixed entries including `expo` and
`react-native-reanimated`, even though no workspace package depends on them —
they are optional peers only. I have not worked out what produced that, so it
may just be an artifact of the install that will not reappear.

PR #5078, opened 2026-05-28
feat(react-core): per-turn persistent intelligence indicator + tag finished state

lukasmoschitz:
## Summary

Refactors the `IntelligenceIndicator` along three axes:

1. **Per-turn persistence** — every agent turn that used Intelligence keeps its own indicator in chat history, instead of a single "latest only" indicator that supersedes prior turns.
2. **Subtle, premium visual** — the in-progress pill is unchanged, but the finished state collapses into an italic muted-gray "metadata line" (icon + past-tense label, no border, no background) that reads as a footnote attached to the assistant message, not as a status chip. The pill→tag transition is a "shrink-and-settle" morph rather than an opacity cross-fade.
3. **Multi-step-turn stability** — the in-turn flicker fix (see its own section below).

> **Scope / split:** this PR is the **indicator UI work**. The underlying `CopilotKitCore` runtime fix that stopped the whole transcript from blanking on a mid-session `/info` re-sync has been **split into a companion PR — #5178** — so it can be reviewed independently by core owners. The two have no compile coupling and can merge in any order.

## Behavior changes

| | Before | After |
|---|---|---|
| **Gate** | Latest matching-assistant across all of `agent.messages` (older turns disappeared) | **First** bash-using assistant *within its turn* (turns bounded by user messages). All turns' indicators coexist in history. |
| **History replay** | Brief `hidden` flash on mount before settling | `useState` lazy init → mounts directly into finished. No flash. |
| **Wrapper testid** | `cpk-intelligence-pill-{id}` | `cpk-intelligence-indicator-{id}` (the wrapper hosts both pill + tag modes; `data-status="in-progress"\|"finished"` distinguishes them). |

## Multi-step-turn stability (flicker fix)

When an assistant turn issued **multiple** `knowledge_base_shell` calls, the indicator visibly **flickered** — the spinner jumped / reset position as the turn progressed.

- **Cause:** the indicator was emitted per assistant message and keyed by **message id**, with the per-turn anchor being the *last* matching message. As more matching tool-calls streamed in, the anchor handed off from one message to the next, **remounting** the component (and resetting its local phase state).
[...]

PR #5178, opened 2026-06-03
fix(core): make setRuntimeTransport idempotent on the requested transport mode

lukasmoschitz:
## Summary

`setRuntimeTransport` was not idempotent on the **requested** transport mode. Auto-detect resolves the requested `"auto"` to a concrete transport (`"rest"`/`"single"`) and writes that back to `_runtimeTransport`; the guard then compared against that **resolved** value. So re-applying the same requested mode — which the provider effect does on **every render** — compared unequal and **re-ran the entire `/info` handshake**, rebuilding the runtime agents mid-session.

When that re-sync lands during a turn, `useAgent` hands the UI a freshly-rebuilt (empty) agent for a render, **blanking the whole transcript** (and any per-message UI bound to it — e.g. the intelligence indicator) until it replays.

## Fix

Track the **requested** mode separately (`_requestedTransport`) and guard on it. Re-applying an unchanged requested transport — including `"auto"` after auto-detect has resolved it — is now a no-op, so no redundant `/info` re-sync fires.

## Tests

`packages/core/src/__tests__/agent-registry-resync.test.ts`: re-applying `"auto"` after auto-detect resolves it does **not** refetch `/info`.

## Scope / risk

Touches only `packages/core` (`core/agent-registry.ts` + the test). **No public API change.** Genuine transport *changes* still re-sync exactly as before; only redundant re-applications of the same requested mode are skipped.

> **Verified live:** this idempotency guard *alone* eliminates the mid-turn re-sync in the e-commerce intelligence demo — instrumentation shows no `/info` handshake fires during a turn, and the transcript/indicator no longer flickers. (An earlier draft also made the re-sync itself non-destructive as defense-in-depth; that was dropped because, with this guard, no mid-turn re-sync occurs for it to protect against.)

## Related

Same fix, two bases (only one needs to merge per target):
- **#5178** → `mme/learn-from-user-activity` (so the feature line gets it without waiting on a `main` → `mme` sync)
- **#5179** → `main` (the canonical home — this is a general runtime-stability fix, independent of the indicator)

### DIFF
--- a/packages/core/src/__tests__/agent-registry-resync.test.ts
+++ b/packages/core/src/__tests__/agent-registry-resync.test.ts
@@ -1,0 +1,64 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { CopilotKitCore } from "../core";
+import { waitForCondition } from "./test-utils";
+
+/**
+ * `setRuntimeTransport` must be idempotent on the requested transport MODE.
+ * Auto-detect resolves "auto" → a concrete transport ("rest"/"single") and
+ * writes that back, so a guard comparing the resolved value would treat a
+ * re-applied "auto" (which the provider effect does on every render) as a
+ * change and re-run the entire `/info` handshake — needlessly rebuilding the
+ * runtime agents mid-session. Guard on the requested mode instead.
+ */
+describe("CopilotKitCore runtime re-sync", () => {
+  const originalFetch = global.fetch;
+  const originalWindow = (global as unknown as { window?: unknown }).window;
+
+  beforeEach(() => {
+    vi.restoreAllMocks();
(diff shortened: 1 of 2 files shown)


lukasmoschitz:
Companion PR: #5078 — the intelligence indicator UI work (per-turn persistent indicator + slot + in-turn flicker fix). **This** PR is the underlying `CopilotKitCore` runtime fix that addresses the root cause (transcript blanking on a mid-session `/info` re-sync). No compile coupling between them — either merge order is fine.

lukasmoschitz:
Closing — superseded by #5179, which lands the **identical** fix against `main` (its canonical home, since it's a general runtime-stability fix). These were the same patch on two bases; consolidating on the main-targeted PR.

PR #5193, opened 2026-06-03
feat(react-core): robust per-agent send serialization for CopilotChat

jpr5:
## Summary

Replaces the bare-`await` send path in `CopilotChat` with a **per-agent chained send queue** backed by a new **synchronously-exposed run-completion handle** on core. This is the robust follow-up to the minimal gen-ui cell fix and eliminates the four event-timing races at their root.

### Root cause

`packages/core/src/intelligence-agent.ts` assigns `activeRunCompletionPromise` only AFTER `await onInitialize`. A caller that wants to serialize against a run's full lifecycle therefore has no synchronous handle to gate on — a second send can add its message and kick a new run while the prior run's event pipeline is still mid-flight.

### Part A — core (additive, no signature changes)

- Split `RunHandler.runAgent` into a synchronous public wrapper + private `_runAgentInner`, capturing the inner run promise BEFORE any `await`.
- For top-level runs only (`_runDepth === 0`), register a normalized completion promise in a `WeakMap<AbstractAgent, Promise<void>>`; expose `runCompletion(agent)`.
- Surface `runAgentCompletion(agent)` on `CopilotKitCore`. The handle resolves (never rejects) on success OR failure, including a pre-RUN_STARTED rejection.

### Part B — react-core

- `sendChainsRef = useRef<WeakMap<AbstractAgent, Promise<void>>>`. `enqueueSend(doSend)` captures the agent at enqueue, chains off the prior send (swallowing rejections so the chain never breaks), runs `doSend()` (the `addMessage`), then holds the slot on `runAgentCompletion(agent) ?? runAgent(...)` until the run genuinely finishes.
- `onSubmitInput` and `handleSelectSuggestion` both route through `enqueueSend`. Agent change mid-send is handled by WeakMap keying (new agent = fresh chain).

Because the queue gates on completion, by the time a later run reaches `await agent.detachActiveRun()` the prior run has fully settled → detach is a clean no-op (the seam closes structurally).

### Tests (TDD red→green)

New `MockRunLifecycleAgent` + `Deferred` in `test-helpers.tsx` with controllable `gateRunStarted`/`gateCompletion`, `failBeforeStart()`, and call-order/concurrency recording — the Subject-based `MockStepwiseAgent` cannot reproduce these races.

- **core** `run-handler-completion.test.ts` (6 tests): completion defined synchronously after `runAgent()` before any await; settles on reject; not registered for follow-up runs; per-agent WeakMap.
- **react-core** `CopilotChat.sendQueue.e2e.test.tsx` (6 tests): race1 (settles-before-RUN_STARTED), race2 (3rd send between start+completion), race3 (agent change → fresh chain), race4 (reject before RUN_STARTED), abort-releases-queue, plus a suggestion-path variant.

Red-green evidence: with the queue stashed, race1/race2/race4/abort/suggestion fail with overlapping runs (`maxConcurrentRuns` = 2–3); all pass with the queue. The core suite (6/6) is RED with `runAgentCompletion is not a function` before the impl. race3 is a per-agent-keying correctness guard (a naive single-chain queue would fail it).

## Relationship to the minimal cell-fix PR

> **This PR SUPERSEDES the minimal gen-ui cell fix's bare-`await` serialization** and must be reconciled at integration: **keep** the minimal fix's `canSend` Enter-gate and its demo sync-resolve / guards; **replace only** the bare-`await` serialization with this completion-gated queue.

## Test plan

- [x] `nx run @copilotkit/core:test` — 436/436 pass (incl. 6 new completion tests)
- [x] `CopilotChat.sendQueue.e2e.test.tsx` — 6/6 pass; clean RED with queue stashed
- [x] `tsc --noEmit` (core + react-core) — no new errors in changed files (pre-existing phoenix/web-inspector/zod errors unrelated)
- [ ] CI green

### DIFF
--- a/packages/core/src/core/__tests__/run-handler-completion.test.ts
+++ b/packages/core/src/core/__tests__/run-handler-completion.test.ts
@@ -1,0 +1,173 @@
+import { describe, it, expect, beforeEach, vi, afterEach } from "vitest";
+import { CopilotKitCore } from "../core";
+import { MockAgent } from "../../__tests__/test-utils";
+
+/**
+ * Tests for the synchronously-exposed run-completion promise.
+ *
+ * The root cause these tests guard against: in `intelligence-agent.ts`,
+ * `activeRunCompletionPromise` is only assigned AFTER `await onInitialize`,
+ * so any caller that wants to observe a run's completion at the moment it
+ * is started (before RUN_STARTED) has no synchronous handle. The send queue
+ * in CopilotChat needs exactly such a handle to serialize sends.
+ *
+ * `RunHandler.runAgent` must therefore capture the inner run promise BEFORE
+ * any `await`, register a normalized completion promise in a per-agent
+ * WeakMap, and expose it via `runCompletion(agent)`. `CopilotKitCore`
+ * surfaces it as `runAgentCompletion(agent)`.
+ */
+
+/**
+ * A controllable agent whose `runAgent` resolves only when the test calls
+ * `resolveRun()` (or rejects via `rejectRun()`). Lets us assert that the
+ * completion promise is observable synchronously and settles correctly.
+ */
+class ControllableAgent extends MockAgent {
+  public runStarted = false;
+  private _resolve?: (value: { newMessages: [] }) => void;
+  private _reject?: (err: unknown) => void;
+
+  override async runAgent(): Promise<{ newMessages: [] }> {
+    this.runStarted = true;
+    return new Promise<{ newMessages: [] }>((resolve, reject) => {
+      this._resolve = resolve;
+      this._reject = reject;
+    });
+  }
+
+  resolveRun(): void {
+    this._resolve?.({ newMessages: [] });
+  }
+
+  rejectRun(err: unknown): void {
+    this._reject?.(err);
+  }
+
+  override async detachActiveRun(): Promise<void> {}
+}
+
+describe("RunHandler run-completion promise (synchronous exposure)", () => {
+  let core: CopilotKitCore;
+
+  beforeEach(() => {
+    core = new CopilotKitCore({});
+    vi.clearAllMocks();
+  });
+
+  afterEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  it("exposes a completion promise synchronously, BEFORE any await in runAgent", () => {
+    const agent = new ControllableAgent({ agentId: "test" });
+    core.addAgent__unsafe_dev_only({ id: "test", agent: agent as any });
+
+    // Kick off the run but do NOT await it. The promise handle must be
+    // registered synchronously by the time runAgent() returns its promise,
+    // which is the same microtask. We immediately query the completion.
+    void core.runAgent({ agent: agent as any });
+
+    const completion = core.runAgentCompletion(agent as any);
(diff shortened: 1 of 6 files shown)


BenTaylorDev:
Closing this as superseded.

The narrower fixes landed instead, in #5195 (`cd0e14536a`) and #5213 (`32551f2bdb`), within hours of this PR opening. Together they are what `main` runs today:

- `packages/react-core/src/v2/components/chat/CopilotChat.tsx:684` — `waitForActiveRunToSettle()`, called from `onSubmitInput` (line 730) and `handleSelectSuggestion` (line 829).
- `packages/core/src/intelligence-agent.ts:110,124` — the typed `RunCompletionAware` contract and the `isRunCompletionAware` guard.

This branch is stale and now conflicts in both load-bearing files, `packages/core/src/core/run-handler.ts` and `CopilotChat.tsx`. Part A is not applicable as authored: it defines `runAgent(params)`, and `main`'s signature is now `runAgent(params, continuationHandoff?)`. Part B is not salvageable either, because `onSubmitInput` has since grown attachment guards, an optimistic clear-and-restore of the composer text, and `localRunId` wake tracking.

**The race described here is still open on `main`.** I reproduced it against the pinned `@ag-ui/client@0.0.57` using main's exact guard logic:

```
[A] async onRunInitialized, 2nd send 30ms later
  at 2nd send: isRunning=true completionPromise=undefined detach$=undefined
  maxConcurrentRuns=2 -> RACE
[B] CONTROL: sync onRunInitialized, 2nd send next macrotask
  at 2nd send: isRunning=true completionPromise=set detach$=set
  maxConcurrentRuns=1 -> serialized
```

In 0.0.57, both `AbstractAgent.runAgent` and `AbstractAgent.connectAgent` set `isRunning = true` synchronously, then assign `activeRunDetach$` and `activeRunCompletionPromise` only after `await this.onInitialize(...)`. Inside that window both of our guards do nothing. `CopilotChat.tsx:692` requires `agent.isRunning && activeRunCompletionPromise`, and `detachActiveRun()` returns early when `activeRunDetach$` is undefined, so the pre-run detach at `run-handler.ts:529` is also a no-op. The second run then orphans the first run's pipeline.

The window opens whenever anything in run initialization is genuinely async. That is a documented pattern, not an edge case: our AG-UI subscriber docs show `onRunInitialized: async ({ messages, state }) => { const enrichedState = await loadUserPreferences(); ... }` at `showcase/shell-docs/src/content/ag-ui/sdk/js/client/subscriber.mdx:366`. Every `onRunInitialized` inside this repo is synchronous, which is why the control above serializes and why we have no report of this yet.

Closing the window is a small change against today's `main`, not a rebase of this branch. One note for whoever picks it up: the `_runCompletions` WeakMap here registers only top-level `RunHandler.runAgent` calls at `_runDepth === 0`, so `ɵconnectWithoutEventVerification` (`packages/core/src/utils/connect-replay.ts:48`, called from `agent.ts` and `intelligence-agent.ts`) would stay uncovered. It has the same assign-after-await shape at lines 80 and 84.

Tracked as a fresh issue so the finding is not lost with this branch.

BenTaylorDev:
Filed as #6937 with the reproduction and the two entry points that need covering.

PR #5151, opened 2026-06-01
feat(integrations): Intelligence threads — north-star foundation + batch 1 (7 examples on 1.59.1) [ENT-679]

BenTaylorDev:
## Summary

**Threads/Intelligence rollout across the integration examples ([ENT-679](https://linear.app/copilotkit/issue/ENT-679)) — the shared foundation + Batch 1.** Makes the **`langgraph-python` north-star** Intelligence-ready, ships the framework-agnostic activation overlay + parity wiring, and integrates the **six Batch-1 examples**, standardizing everything on **`@copilotkit/* 1.59.1`**.

> Originally two stacked PRs; the batch-1 work was folded into this branch so it's reviewed as one unit.

## Per-example status
| Example | agent | threads | demo | `next build` | verified |
|---|---|---|---|---|---|
| langgraph-python (north-star) | `default` | ✅ | todo (CopilotChat) | ✅ | bumped 1.56.5→1.59.1 |
| langgraph-js | `default` | ✅ | todo (shared) | ✅ | build |
| mastra | `weatherAgent` | ✅ | proverbs/weather/moon | ✅ | ✅ **full E2E** |
| adk | `my_agent` | ✅ | proverbs (`useRenderTool` weather) | ✅ | build |
| agno | `agno_agent` | ✅ | proverbs (local state) + tools | ✅ | build |
| ms-agent-framework-python | `my_agent` | ✅ | proverbs/weather/moon | ✅ | build |
| ms-agent-framework-dotnet | `my_agent` | ✅ | proverbs/weather/moon | ✅ | build |

## What's in it

**Foundation (north-star + shared):**
- **Env-gated runtime wiring** (`route.ts`): license → `CopilotKitIntelligence` + `identifyUser` + `licenseToken`; otherwise `InMemoryAgentRunner`. Marker-fenced for opt-out; multi-route endpoint.
- **Threads frontend**: `threads-drawer/**` refactored into a **token-driven themeable base** + locked-state tease; `next.config` derives `NEXT_PUBLIC_COPILOTKIT_THREADS_ENABLED` from the license (SSR-safe client mount). **Without a license the panel renders locked.**
- **`_intelligence/` activation overlay**: docker-compose (postgres + redis + composite `0.2.0`), init-db, `.env.intelligence`, an idempotent **`provision-user`** one-shot (`demo-user`), and **`CHECK_ORIGIN=false`** so the local realtime-gateway accepts the app's websocket origin (live thread updates).
[...]

PR #5218, opened 2026-06-04
fix(examples): restore starter @chat usability (default agent key + useSingleEndpoint=false)

jpr5:
## Summary

Re-applies the starter `@chat` usability fix that was collaterally reverted, **fully decoupled from the Intelligence-threads feature.**

- **PR #5210** standardized starter agent registry keys to `"default"`.
- **PR #5151 (ENT-679)** had added `useSingleEndpoint={false}` to the starter `<CopilotKit>` layouts.
- Together these two halves fixed the **"@chat: No assistant response"** blocker. Without them, the client auto-detects single-endpoint transport and POSTs to a route that 404s instead of the per-agent endpoint keyed `default`.
- The ENT-679 revert (**#5217**, commit `3721e7b36`) backed out **both** halves as collateral, so starters on `main` are broken again (`adk` key `my_agent`, `agno` `agno_agent`, etc., and no `useSingleEndpoint={false}`).

This PR re-applies **only** the usability fix. It does **not** reintroduce any Intelligence/threads scaffolding — no threads-drawer, no `_intelligence` overlay, no `CopilotKitIntelligence` runtime block / PATCH+DELETE handlers, no ghcr composite images, no `package.json`/version changes.

## Per-file changes (per affected starter)

- `route.ts` (or `mastra/src/mastra/index.ts`): agents registry key → `default`
- `layout.tsx`: add `useSingleEndpoint={false}` to `<CopilotKit>`, drop the `agent=` prop so it falls back to `"default"`
- `page.tsx`: `useAgent({ name|agentId: ... })` literal → `"default"`

Agent **semantic** names/exports are intentionally unchanged (e.g. mastra `weatherAgent` export, the Python backend's internal `my_agent` variable). Only the runtime registry KEY and the frontend agentId become `"default"`, exactly as #5210 did.

## Affected starters

adk, agno, crewai-crews, llamaindex, ms-agent-framework-dotnet, ms-agent-framework-python, pydantic-ai, mastra.

(`agno`'s `page.tsx` is unchanged because it has no `useAgent` reference; `langgraph-*` were already on the `default` key and are untouched. `langgraph-python-threads` is untouched.)

## Test plan

- [ ] CI `smoke-starter` matrix — the `@chat` legs must go green (this is the authoritative verification; the smoke harness boots each starter and exercises the chat round-trip)

### DIFF
--- a/examples/integrations/adk/src/app/api/copilotkit/[[...slug]]/route.ts
+++ b/examples/integrations/adk/src/app/api/copilotkit/[[...slug]]/route.ts
@@ -8,7 +8,7 @@ import { handle } from "hono/vercel";
 
 const runtime = new CopilotRuntime({
   agents: {
-    my_agent: new HttpAgent({
+    default: new HttpAgent({
       url: process.env.AGENT_URL || "http://localhost:8000/",
     }),
   },
--- a/examples/integrations/adk/src/app/layout.tsx
+++ b/examples/integrations/adk/src/app/layout.tsx
@@ -17,7 +17,11 @@ export default function RootLayout({
   return (
     <html lang="en">
       <body className={"antialiased"}>
-        <CopilotKit runtimeUrl="/api/copilotkit" agent="my_agent">
+        {/* Force REST transport so runtime-info + threads both hit the multi-route endpoint (auto-detect races the lazily-compiled API route in next dev). */}
+        <CopilotKit
+          runtimeUrl="/api/copilotkit"
+          useSingleEndpoint={false}
+        >
           {children}
         </CopilotKit>
       </body>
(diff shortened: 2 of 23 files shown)


jpr5:
Closing: re-applying the @chat fix on the post-revert starters does NOT work — they were rolled back to pre-1.59 react-core (1.55.2/1.56.x) by the ENT-679 revert (#5217), and on those versions the fix is inert/crashes (mastra throws a client-side React error; @chat smoke fails on all 8 touched starters locally + CI). The @chat usability fix requires 1.59.x. Superseded by a bundled float-forward-to-1.59.x + @chat effort (decoupled from Intelligence threads).

PR #5237, opened 2026-06-04
fix(showcase/harness): prevent browser-pool /tmp profile-dir accumulation (durable root-cause fix for recurring wedge)

jpr5:
## Root cause (the recurring BrowserPool wedge, ~4th occurrence)

The long-lived harness container runs a **single Node process** that launches chromium repeatedly under sustained d6 cron load — init fill, hygiene recycles, and crash-recovery relaunches. Every `chromium.launch()` Playwright performs `mkdtemp`s **two** ephemeral dirs under the OS temp dir:

- a `playwright_chromium*dev_profile-*` **userDataDir**, and
- a `playwright-artifacts-*` dir.

Because the harness passes `--disable-dev-shm-usage` (which is **also** a Playwright default), chromium routes its shared-memory / cache I/O into that profile dir under `/tmp` instead of `/dev/shm` — concentrating all per-browser pressure on `/tmp`.

Playwright removes those dirs in the child process's `close` handler, but the removal is **best-effort**: `processLauncher`'s `cleanup()` calls `removeFolders()` and **logs-and-swallows** any failure (ENOSPC, EBUSY, a hung close on a wedged chromium, or the Node process being signalled). Over **hours**, the orphaned `playwright_*` dirs accumulate **monotonically** in the container's `/tmp` until the filesystem exhausts inodes/space. At that point chromium can no longer create its profile dir and dies **instantly on startup**, so **every** `chromium.launch()` throws:

```
browserType.launch: Target page, context or browser has been closed
```

The browser set empties, `startSelfHeal()` relaunches into the **same exhausted `/tmp`**, and the pool wedges **permanently** — only a container restart (which gives a fresh `/tmp`) clears it. This matches the recurring staging signature exactly (every launch throws "Target page … has been closed"; `self-heal-launch-failed` repeating; only a restart recovers).

### Why this is the mechanism (evidence)

- **Code**: the production launcher pins no `userDataDir`, so Playwright `mkdtemp`s a fresh profile + artifacts dir **per launch**; cleanup is tied to the child `close` handler and `removeFolders` failures are swallowed.
- **Empirical**: confirmed locally that each `chromium.launch({args:["--no-sandbox","--disable-dev-shm-usage"]})` creates **2** `playwright*` temp dirs and that graceful `browser.close()` removes them — while a failed/swallowed cleanup leaves them orphaned.
- The other candidate mechanisms are already mitigated and have **different** error signatures: PID/thread exhaustion (`pthread_create: Resource temporarily unavailable`) is handled by the Dockerfile `ulimit -u` raise + launch-stagger + relaunch backoff; `/dev/shm` exhaustion is redirected by `--disable-dev-shm-usage`. Neither produces "Target page … has been closed" on every launch.

## Prevention (this PR — distinct from the circuit-breaker)

PR #5234 adds a circuit-breaker that **escapes** the wedge after the fact (purges `/tmp/playwright_*` only **after** N consecutive self-heal failures, once the fleet is already 0/N). This PR **stops the accumulation** so the wedge never happens:

- Proactively **sweep** stale `${tmpdir()}/playwright_*` (and `playwright-artifacts-*`) dirs on a **cadence** — once at `init()` and on **every recycle relaunch** (crash recovery **and** hygiene). Each recycle is a chromium replacement, exactly where orphans are produced.
- The sweep is **age-gated** (`staleProfileDirTtlMs`, default **5 min**): a dir for a launch in flight or a live browser is touched continuously (mtime ~now), so it is never older than the TTL and is **never** swept. The sweep therefore cannot delete a dir still in use.
- Fire-and-forget (never on a launch's critical path) and best-effort (failures logged, never fatal). Tunable via `BROWSER_POOL_STALE_PROFILE_DIR_TTL_MS` and an injectable `sweepStaleProfileDirs`.

Net effect: `/tmp` stays bounded at ~`2 * browserCount` live dirs instead of growing `2`-per-launch toward exhaustion.

## Red-green tests

- Sweep fires **once at init()** (carries the configured TTL).
- Sweep fires on **every recycle** (crash + hygiene).
- **Bounded accumulation**: across 30 recycle cycles `/tmp` stays bounded at ~`2 * browserCount` instead of growing `2`-per-cycle (a no-sweep run would leave ~64 entries).
- **Age gate (real FS)**: `defaultSweepStaleProfileDirs` removes an hour-old orphan but **never** a fresh (in-use) dir, and **never** a non-`playwright` dir.

All four **fail without the wiring** (proven by stashing the source change), pass with it.

## Test plan

- [x] `pnpm exec tsc --noEmit` — clean
- [x] `vitest run browser-pool.test.ts` — 52 passed (48 existing + 4 new)
- [x] Full harness suite — 1772 passed; the **only** failure is the known/unrelated `starter_smoke.yml "Unrecognized key: discovery"` in `probe-loader.test.ts` (not touched by this PR)
- [x] `prettier --check` — clean
- [ ] CI green

Does **not** touch PR #5234's files/branch; the two are complementary (prevention + escape).

### DIFF
--- a/showcase/harness/src/probes/helpers/browser-pool.test.ts
+++ b/showcase/harness/src/probes/helpers/browser-pool.test.ts
@@ -1,6 +1,9 @@
 import { describe, it, expect } from "vitest";
+import { promises as fsp } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
 import type { Browser, BrowserContext } from "playwright";
-import { BrowserPool } from "./browser-pool.js";
+import { BrowserPool, defaultSweepStaleProfileDirs } from "./browser-pool.js";
 import type { LaunchBrowser } from "./browser-pool.js";
 
 /**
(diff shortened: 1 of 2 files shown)


jpr5:
Folded into #5234 (the unified durable browser-pool fix).

Both PRs edited the same two files (`browser-pool.ts` + `browser-pool.test.ts`) and both implemented an age-gated `/tmp` profile-dir purge — merging them separately would conflict and duplicate the purge logic. So the proactive root-cause prevention from this PR has been integrated into #5234 and unified onto a SINGLE canonical purge helper:

- `defaultPurgeProfileDirs(minAgeMs, dir?)` is now the one purge helper, taking an age threshold as its first arg.
- The PROACTIVE sweep (this PR's cadence — init + every recycle relaunch) calls it with the conservative `staleProfileDirTtlMs` (5 min).
- The breaker's hard-recovery (#5234) calls the SAME helper with the aggressive `PURGE_MIN_AGE_MS` (60 s).

Same hardened (symlink-safe, realpath-checked) implementation, different age — no duplicate logic. All 4 prevention tests from this PR were ported and pass alongside #5234's existing breaker/purge/alarm coverage (62 tests green). Per one-PR-for-related-work, the full root-cause → prevention → escape → alerting story now lives in #5234.

Closing in favor of #5234.

PR #5246, opened 2026-06-04
fix(react-native): preserve polyfill side effects in the build

habasefa:
## What does this PR do?

Fixes a startup crash in `@copilotkit/react-native` where the polyfills (crypto, streams, encoding, dom, location) silently never ran, so `crypto.getRandomValues` was missing and `uuid` threw `crypto.getRandomValues() not supported` at app startup.

**Root cause.** `@copilotkit/react-native/polyfills` aggregates the individual polyfill modules via bare `import "./polyfills/x"` side-effect imports. The package's `sideEffects` field only listed `./dist/**` paths, so at build time the bundler evaluated the `src/**` modules as side-effect-free and tree-shook those bare imports out of `dist/polyfills.mjs` — leaving only `installStreamingFetch()` (which survived because it's a called function, not a bare import). None of the other polyfills ran.

**Fix.** Add the matching `./src/**` globs to `sideEffects` so the bundler preserves the polyfill imports. `dist/polyfills.mjs` now imports and runs all five polyfill modules.

```diff
 "sideEffects": [
+  "./src/index.ts",
+  "./src/polyfills.ts",
+  "./src/polyfills/**/*",
+  "./src/streaming-fetch.ts",
   "./dist/index.*",
   "./dist/polyfills.*",
   "./dist/polyfills/**/*"
 ]
```

**Verification.**
- Built `dist/polyfills.mjs` now contains all five `import "./polyfills/*"` statements (previously zero).
- Full RN test suite passes (245 tests).
- Verified on a physical device (Samsung Galaxy S21 Ultra, Android 14, bare React Native): the app boots and the crypto polyfill installs at startup — no `crypto.getRandomValues() not supported`.

A changeset is included (`patch`).

## Related PRs and Issues

- (none)

## Checklist

### DIFF
--- a/.changeset/rn-polyfills-side-effects.md
+++ b/.changeset/rn-polyfills-side-effects.md
@@ -1,0 +1,19 @@
+---
+"@copilotkit/react-native": patch
+---
+
+fix(react-native): preserve polyfill side effects in the build so crypto/streams/etc. actually run
+
+`@copilotkit/react-native/polyfills` aggregates the individual polyfill modules
+(`crypto`, `streams`, `encoding`, `dom`, `location`) via bare `import "./polyfills/x"`
+side-effect imports. The package's `sideEffects` field only listed `./dist/**`
+paths, so at build time the bundler evaluated the `src/**` modules as
+side-effect-free and tree-shook those bare imports out of `dist/polyfills.mjs` —
+leaving only `installStreamingFetch()`. As a result none of the other polyfills
+ran, and `crypto.getRandomValues` was never installed, so `uuid` (used by
+`randomUUID`/`randomId` and pulled in by the provider) threw
+`crypto.getRandomValues() not supported` at startup in a release/Hermes runtime.
+
+Add the corresponding `src/**` paths to `sideEffects` so the bundler preserves the
+polyfill imports. `dist/polyfills.mjs` now imports all five polyfill modules and
+they execute on import as intended.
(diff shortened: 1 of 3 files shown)


BenTaylorDev:
Thank you for this — closing because `main` already covers it, from a change that landed after you opened this.

`packages/react-native/package.json` on `main` now declares:

```
"sideEffects": [
  "./src/index.*",
  "./src/headless.*",
  "./src/polyfills.*",
  "./src/polyfills/**/*",
  "./dist/index.*",
  "./dist/headless.*",
  "./dist/polyfills.*",
  "./dist/polyfills/**/*"
]
```

The glob forms cover every `src/` entry this PR adds, plus `./src/headless.*`, which this PR doesn't have. The bundle-test idea landed too: `test` runs `vitest run && pnpm run test:scripts`, and `scripts/__tests__/` now holds 16 test files.

The one line not already covered is `./src/streaming-fetch.ts` — and I don't think that one belongs. `sideEffects` marks modules whose mere import has an effect, and `streaming-fetch.ts` documents itself as "Call `installStreamingFetch()` once at app startup after polyfills." It's an explicit-call module, so listing it would suppress tree-shaking without protecting anything. If you've seen it dropped from a bundle in a way that broke something, say so and I'll take another look — that would mean I've misread how it's consumed.

The underlying instinct was right, and the polyfill side-effect handling did need shoring up. Thanks for flagging it.

PR #5247, opened 2026-06-04
fix(react-native): lazy-load optional expo attachment peers

habasefa:
## What does this PR do?

Makes the optional Expo attachment peers truly optional in `@copilotkit/react-native`, so bare React Native apps (without Expo) can use the package.

**Root cause.** `useAttachments` statically imported `expo-document-picker` and `expo-file-system` at module top level. The hook is re-exported from the package root and used internally by `CopilotChat`, so every consumer of the package loaded those imports — even consumers that never use attachments. Although both packages are declared as optional peers (`peerDependenciesMeta`), the static import made them effectively mandatory, breaking bare RN apps two ways:

1. **Bundle time** — Metro could not resolve `expo-document-picker` / `expo-file-system` when they weren't installed.
2. **Runtime** — even when JS resolved, the Expo modules need native autolinking; without it the app crashed at startup in `expo-modules-core` (`Cannot read property 'EventEmitter' of undefined`).

**Fix.** Load the modules lazily via dynamic `import()` inside the attachment callbacks (`openPicker`, `processFiles`), so non-attachment consumers and bare RN apps no longer pull in Expo. The specifiers are built from variables to keep them out of Metro's static analysis, so resolution happens only when an attachment API is actually used. Attachment behavior is unchanged; modules load on first use.

**Verification.**
- Built `dist` contains no static top-level Expo import; the dynamic `import()` calls are preserved.
- Full RN test suite passes (245 tests).
- Verified on a physical device (Samsung Galaxy S21 Ultra, Android 14, bare React Native with no Expo installed): the app bundles and boots with no `Unable to resolve expo-document-picker` / `expo-file-system` error.

A changeset is included (`patch`).

## Related PRs and Issues

- (none)

## Checklist

### DIFF
--- a/.changeset/rn-optional-expo-attachments.md
+++ b/.changeset/rn-optional-expo-attachments.md
@@ -1,0 +1,24 @@
+---
+"@copilotkit/react-native": patch
+---
+
+fix(react-native): load expo attachment peers lazily so bare RN apps work without Expo
+
+`useAttachments` imported `expo-document-picker` and `expo-file-system` at the top
+level of the module, and `useAttachments` is re-exported from the package root and
+used internally by `CopilotChat`. Because those imports were static, importing
+_anything_ from `@copilotkit/react-native` (e.g. the documented quick-start
+`CopilotKitProvider` / `useAgent`) pulled both Expo modules into the consumer's
+bundle — even though they are declared as **optional** peer dependencies
+(`peerDependenciesMeta`). In a bare React Native app without Expo this broke two
+ways: Metro failed to resolve the modules at bundle time, and when they were
+present as JS-only it crashed at startup in `expo-modules-core`
+(`Cannot read property 'EventEmitter' of undefined`) since there is no native
+autolinking.
+
+The two modules are now imported lazily, inside the attachment callbacks that
+actually use them (`openPicker`, `processFiles`), via dynamic `import()` with a
+non-statically-analyzable specifier. Consumers that never use attachments — and
+bare RN apps without Expo — can now import the rest of the package without pulling
+in Expo or crashing. Attachment users are unaffected; the modules load on first
+use, matching the existing optional-peer contract.
--- a/packages/react-native/package.json
+++ b/packages/react-native/package.json
@@ -66,7 +66,8 @@
   "scripts": {
     "build": "tsdown",
     "dev": "tsdown --watch",
-    "test": "vitest run",
+    "test": "vitest run && pnpm run test:bundle",
+    "test:bundle": "node --test scripts/__tests__/*.test.mjs",
     "check-types": "tsc --noEmit",
     "publint": "publint .",
     "attw": "attw --pack . --profile node16 --ignore-rules internal-resolution-error",
--- a/packages/react-native/scripts/__tests__/attachments-optional-expo.test.mjs
+++ b/packages/react-native/scripts/__tests__/attachments-optional-expo.test.mjs
@@ -1,0 +1,89 @@
+// Standalone Node test (not vitest) — it drives real tsdown builds, which
+// don't work inside the package-wide jsdom vitest environment. Same approach
+// as react-core's scripts/__tests__/measure-copilotchat.test.mjs.
+//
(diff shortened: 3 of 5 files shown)


BenTaylorDev:
Thanks for this — the problem is real, and it is still open on main for the
prebuilt-UI path, so the fix is wanted. The laziness half of your change works.
But I measured the bundle-time half and it does not do what the description
says, so this needs one more pass.

## The template-literal trick does not hide anything from Metro

Metro's `collectDependencies` runs Babel `.evaluate()` on the `import()`
argument, so a same-module `const` folds right back to the literal and the
dependency is recorded anyway. What decides whether a bare RN app can bundle
is not visibility, it is the `isOptional` flag: `@react-native/metro-config`
sets `allowOptionalDependencies: true` (`dist/index.js:77`), and Metro only
swallows a failed resolution for a dependency it has marked optional.

I drove `metro/src/ModuleGraph/worker/collectDependencies.js` (metro 0.84.6,
the line RN 0.85.2 pins) against this branch's actual `use-attachments.ts`:

| variant | recorded | `isOptional` |
| --- | --- | --- |
| this PR (template literal, no `.catch()`) | yes | **false** |
| literal specifier, no `.catch()` | yes | false |
| literal specifier **+ `.catch()` on the loader** | yes | **true** |

The template literal changes nothing, and as it stands both deps are still
non-optional, so Metro will not swallow the resolution failure.

## The fix

A `.catch()` on each loader is enough, at any nesting depth — the module-scope
loaders can stay exactly where they are:

```ts
import type * as DocumentPickerTypes from "expo-document-picker";
import type * as FileSystemTypes from "expo-file-system";

const loadDocumentPicker = (): Promise<typeof DocumentPickerTypes> =>
  import("expo-document-picker").catch(() => {
    throw new Error("expo-document-picker is required to open the file picker");
  });

const loadFileSystem = (): Promise<typeof FileSystemTypes> =>
  import("expo-file-system").catch(() => {
    throw new Error("expo-file-system is required to read attachments");
  });
```

I measured this variant too: both deps come back `isOptional=true`. The
type-only imports are erased before Metro sees them (no extra dependency
recorded), and they let you drop `Promise<any>` and the hand-written
`asset: { uri: string; ... }` annotation, so the Expo calls keep their types.
A rejected promise still lands in your existing `try`/`catch`, so the
graceful-failure behavior your tests assert is unchanged.

## Why your tests stayed green

`attachments-optional-expo.test.mjs` greps for a literal specifier sitting
directly inside `from` / `require(` / `import(`. I built this code shape with
the package's own tsdown config: the bundler does not fold the template
literal, so the output keeps `const FILE_SYSTEM_MODULE = "expo-file-system"`
alongside a template-literal `import()` call. Both of your assertions pass on
that output — no static reference matched, bare string present — while Metro
folds the same code and records the dependency. The comment says the pattern
covers "a constant-folded template literal", but a regex cannot see one.

Worth replacing that test with one that reads `isOptional` off
`collectDependencies` directly. A bundle grep cannot answer this question.

`use-attachments-missing-expo.test.ts` is fine and worth keeping — it verifies
the load counters stay at 0 until a callback runs, and that both failure paths
degrade gracefully. It runs in vitest, so it resolves at Node runtime and
simply cannot observe bundle-time behavior.

## One question

Your device verification (bare RN on the S21 Ultra, no Expo, bundles and
boots) does not fit the measurement above. What was in that app's
`metro.config.js` — any `resolver.blockList`, `extraNodeModules`, or stub
entries for the two Expo modules? I would like to understand what the app was
actually exercising before dismissing the report.

## Also

- Please drop `.changeset/rn-optional-expo-attachments.md` — version bumps go
  through the release workflow scopes here.
- `"test": "vitest run && pnpm run test:bundle"` puts two real tsdown builds
  into the unit lane, which is where react-native runs under nx-affected.
  That is a maintainer call, so flagging rather than blocking.

For context on scope: main has since gained
`@copilotkit/react-native/headless`, which covers consumers who do not need
the chat UI. `CopilotChat.tsx:222` still calls `useAttachments`, so the
prebuilt-UI path this PR targets is genuinely still broken.

PR #5248, opened 2026-06-04
feat(shell-docs): recommend create + skills onboard commands in landing heroes

onsclom:
## What

Replaces the landing-page CTA with **three entry points**, framed by situation, and renders the **identical action block on the home hero and every framework landing hero**:

| | action |
|---|---|
| **New project** | `npx copilotkit create` |
| **Existing project** | `npx copilotkit skills onboard` |
| **Guided walkthrough** | **Quickstart** button (preserved from the previous hero) |

- **Unified `<HeroStartActions>` block**: two equal-weight command cards plus a quickstart row beneath, shared verbatim by the home hero and the framework landing heroes (per review: the two surfaces previously diverged).
- **Quickstart preserved** in its original accent treatment. On the home hero it is the framework-picker dropdown (`<HeroQuickstartDropdown>`, restored); on framework pages it links straight to that framework's quickstart guide. The home hero also keeps the "Learn more about building with agents" link in the same row.
- **Framework landing heroes** (e.g. `/langgraph-typescript`): the create command **pre-fills the framework** via the CLI's `--framework` flag (e.g. `--framework langgraph-js`).

**Framework-flag mapping**: docs slug to CLI `--framework` value, verified against the CLI's `AGENT_FRAMEWORKS` enum (`langgraph-typescript`→`langgraph-js`, `langgraph-python`→`langgraph-py`, `google-adk`→`adk`, `strands`→`aws-strands-py`, `ms-agent-dotnet`→`microsoft-agent-framework-dotnet`, identical for `mastra`/`pydantic-ai`/`llamaindex`/`agno`/`ag2`). Slugs with **no** 1:1 CLI template fall back to a bare `npx copilotkit create`, notably `crewai-crews` (the CLI ships *CrewAI Flows*, not Crews), plus `langgraph-fastapi`, `claude-sdk-*`, `langroid`, `spring-ai`, `agent-spec`, `deepagents`. `skills onboard` has no framework flag, so it is identical everywhere. Frameworks with bespoke setup (`a2a` `git clone`, `ms-agent-dotnet`) keep the pre-cards layout: quickstart button plus their own copy-command chip.

**Responsive, with all text always visible.** Commands **wrap, never truncate**:

- Wraps happen at spaces only; every token is non-breaking, so `--framework` can never split into a dangling `-` at a line edge.
- `text-wrap: balance` splits multi-line commands evenly, typically right at the flag boundary (`npx copilotkit@latest create` / `--framework langgraph-js`).
- The block caps at 740px with 12px mono, the narrowest cap where both home commands fit one line with enough headroom to survive platform mono-font width differences.
- Cards sit two-up from `sm` and stack below it; the grid (`min-w-0`, `items-stretch`) keeps long commands inside their track and the card pair equal-height.

## Screenshots

**Home**: two cards, quickstart dropdown, learn-more link

home

**Home, quickstart dropdown open** (framework picker preserved)

home dropdown

**Framework landing (LangGraph)**: same block, framework pre-filled, create command balanced across two lines, quickstart links to the guide

framework

**Worst case (Microsoft Agent Framework, Python)**: longest CLI flag value, three balanced lines, fully readable

long command

**Bespoke setup (A2A)**: quickstart button plus own command chip (pre-cards layout preserved)

a2a

**Mobile (375px)**: cards stack, quickstart goes full-width

| home | framework |
|---|---|
| home mobile | framework mobile |

## Telemetry

Both hero copy buttons are now explicitly instrumented: each click captures **`hero_command_copied`** (`command_id`: `create` | `onboard`, full `command` string, `clipboard_blocked`), so create-vs-onboard funnels are queryable per landing page. The pre-existing global `cli_command_copied` (fired by `CopyTracker` on any clipboard copy) still fires for volume metrics; the new event uses a different name so that funnel is not double-counted. Validated locally against a live PostHog client: each click POSTs both events (plus `$autocapture`) to `/ingest/e` with HTTP 200.

## Notes

- Both cards equal weight; accent only on hover. Copy rows copy on click with `aria-live` feedback plus a clipboard-blocked fallback; cursor is `pointer`.
- Removes `agent-start-prompt.tsx` and `hero-command-copy.tsx`. `hero-quickstart-dropdown.tsx` is back (restored unchanged after review feedback).


tylerslaton (changes requested):
Looking really great! 

I have a few pieces of feedback that I think will let us ship this.

1. It would seem we lost the `quickstart` button in the transformation, I do think we should have all three. Could we see an iteration that considers this? Ideally we preserve the old quickstart button
2. Minor nit, not a blocker but can we unify the layout for both the landing page + framework specific landing pages? Right now they're slightly different which I think is slightly confusing.

onsclom:
Both addressed in 2502533cd:

1. **Quickstart restored**: original accent button, all three actions present. Home gets the framework dropdown back; framework pages link straight to their quickstart guide.
2. **Layouts unified**: home and framework heroes now render the identical `<HeroStartActions>` block. The stacked framework variant is gone; long commands wrap balanced at token boundaries instead.

PR description has fresh screenshots, including the worst-case long command and mobile.

tylerslaton (approved):

PR #5267, opened 2026-06-05
Implement config validation and normalization

krataratha:
Added a function to validate and normalize Doctest configuration.

## What does this PR do?

## Related PRs and Issues

## Checklist

### DIFF
--- a/scripts/doc-tests/extract.ts
+++ b/scripts/doc-tests/extract.ts
@@ -33,7 +33,29 @@ interface DoctestConfig {
 }
 
 // ---------------------------------------------------------------------------
-// Config
+// Config Feature: Validator and Normalizer
+// ---------------------------------------------------------------------------
+
+/**
+ * Validates and normalizes the Doctest configuration object,
+ * ensuring all language keys have at least an empty dependencies array.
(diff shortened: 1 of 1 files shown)


BenTaylorDev:
Thanks for opening this, but I'm closing it — `normalizeConfig` is never called.

The symbol appears exactly twice in `scripts/doc-tests/extract.ts` as this PR leaves it: the function definition, and a commented-out usage example (`// const safeConfig = normalizeConfig(rawConfig);`). Nothing in the file or the wider script invokes it, so the config the doctest runner actually reads is unchanged. The PR adds a function and a comment, and no behavior.

Two smaller things, in case you want to pick this up again:

- It doesn't validate. The title says validation and normalization, but the implementation coerces a missing or non-array `deps` to `[]` and never rejects malformed input. If the goal is catching a bad doctest config early, that wants an explicit error, not a silent default.
- There are no tests, and `scripts/doc-tests/` has a test setup that would take one.

If you hit a real problem with doctest config handling, I'd welcome a PR that wires the normalizer into the read path and covers it with a test — that's the version of this change that would do something.

PR #5276, opened 2026-06-05
fix(showcase): pin aimock base to 1.28.0 (full-depth fixture recursion)

jpr5:
## Summary

Pin the showcase aimock image base to `ghcr.io/copilotkit/aimock:1.28.0` (was bare `:latest`).

The repo bakes nested `d4/` and `d6/` fixture directories into the image, but bare `:latest`'s fixture loader had shallow recursion and silently skipped the nested dirs. 1.28.0 has full-depth recursion — verified to load all 7165 fixtures including d4/d6.

> **Note:** This PR previously also added a shared "Hello" greeting fixture, which collided with the pre-existing scoped "Hello" fixtures in the d4 chat.json files and failed the Validate Showcase shared-vs-scoped collision check. That fixture has been reverted; the net change here is the Dockerfile pin only. The starter-smoke chat rung is greened separately in #5277 by repointing the probe at the existing collision-free "hello world" shared fixture.

## Proof

- Validate Showcase collision validator: **733/733 pass** (the reverted "Hello" fixture no longer collides).
- Net diff vs `main`: `showcase/aimock/Dockerfile` only (1 line).

## Test plan

- [ ] CI `Validate Showcase` green
- [ ] Post-deploy: aimock image loads full d4/d6 fixture set

### DIFF
--- a/showcase/aimock/Dockerfile
+++ b/showcase/aimock/Dockerfile
@@ -1,4 +1,4 @@
-FROM ghcr.io/copilotkit/aimock:latest
+FROM ghcr.io/copilotkit/aimock:1.28.0
 
 # Depth-organized fixture directories
 COPY shared/ /fixtures/shared/
--- a/showcase/aimock/shared/common.json
+++ b/showcase/aimock/shared/common.json
@@ -5,6 +5,11 @@
   },
   "fixtures": [
     {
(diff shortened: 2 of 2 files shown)


jpr5:
Note: this 1.28.0 base pin is NOT the chat-rung fix — root-cause is the harness probe missing X-AIMock-Context (staging aimock is healthy + loads d4/d6). Keeping this open purely as the prod determinism-hygiene pin per the staging→prod plan; not merging into healthy staging now to avoid an unnecessary aimock rebuild.

jpr5:
Closing as obsolete. The 1.28.0 pin never landed — `main:showcase/aimock/Dockerfile` is still `:latest`, which now resolves well past 1.28.0. Landing this today would pin to an *older* image (a downgrade), and the full-depth fixture-recursion fix it targeted has long been in `:latest`.

PR #5293, opened 2026-06-06
fix(showcase): thread real shell host server-side so dashboard links resolve

jpr5:
## Summary

The dashboard page tree is entirely `"use client"`, so `FeatureGrid`'s `resolveShellUrl()` read the CLIENT runtime config, which returns the `https://ssr-placeholder.invalid/` sentinel during SSR (no `window`). That dead sentinel was baked into all 1184 Demo/Code `<a href>` values and never re-derived after hydration, so every link pointed at a non-resolving host (NXDOMAIN / chrome-error on click).

This converts the route file (`app/page.tsx`) into a server component that reads `SHELL_URL` at request time via `getRuntimeConfig()` and threads the real `shellUrl` down into the renamed client `DashboardPage`, which forwards it to `FeatureGrid` as a prop. `FeatureGrid` now builds anchors from the real host during SSR, so the sentinel never reaches the DOM, links work pre-hydration (crawlers/no-JS), and the previously-dead links resolve. The trailing-slash normalization from the prior commit is retained.

## Verification

Proven in a prior session via a local browser run (`next start` with real `SHELL_URL`): SSR HTML has 0 `ssr-placeholder.invalid` and 1184 real-host anchors; post-hydration all 1184 anchors are the real host; clicking a Demo link navigates to a real showcase page (not chrome-error).


jpr5:
Superseded: every load-bearing element of this PR already landed on main (server-component page.tsx threading shellUrl, urlsFor trailing-slash strip, resolveShellUrl + prop threading in feature-grid.tsx, dashboard-page.tsx, and the SSR-anchor tests) — with enhancements (PROD_INVALID_SHELL_URL guard, degraded/now props). Rebasing this branch onto main yields a 0-byte diff. Closing as already-landed.

PR #5295, opened 2026-06-06
fix(showcase): fail verify-deploy when dashboard config carries an env-unset sentinel

jpr5:
## Summary

The dashboard deploy probe previously only asserted HTTP 200, so a deploy with an unset `SHELL_URL` (or `POCKETBASE_URL`) on the Railway service shipped green even though the injected runtime config fell back to a sentinel — leaving every Demo/Code/hover-preview link (`shellUrl`) or Status-tab live-reader (`pocketbaseUrl`) dead. This exact failure hit staging.

After a green baseline, `probeDashboard` now fetches `/`, parses the injected `window.__SHOWCASE_CONFIG__`, and fails loud when `shellUrl` is `about:blank#shell-url-missing` or `pocketbaseUrl` is `http://pocketbase.invalid` (the `PROD_INVALID_*` sentinels in shell-dashboard runtime-config). When the config block is absent from a render it does not false-fail; a present-but-malformed block fails explicitly.

## Verification

Proven in a prior session.

### DIFF
--- a/showcase/scripts/__tests__/verify-deploy.drivers.test.ts
+++ b/showcase/scripts/__tests__/verify-deploy.drivers.test.ts
@@ -657,10 +657,16 @@ describe.each(DRIVER_CASES)(
         });
       });
       const healthUrls = seen.filter((u) => !u.includes("/graphql/v2"));
-      expect(healthUrls).toHaveLength(1);
-      expect(healthUrls[0]).toBe(
-        `https://${entry.domains.prod}${expectedHealthPath}`,
-      );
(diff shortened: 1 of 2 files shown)


jpr5:
Superseded: this fix already landed on main via 64461e170 ("fail verify-deploy on env-unset config sentinel + robust config extractor"), a hardened re-implementation (tag-boundary config extractor, fail-loud on format drift, additional tests). Rebasing this branch onto main yields an empty diff. Closing as already-merged-in-superior-form.

Issue #5282, opened 2026-06-05
🐛 Bug: Headers not being sent on /threads requests


Simon-Gregory-LG:
### ♻️ Reproduction Steps

We are enforcing CRSF checks through a backend proxy and using BFF

1. Set headers on CopilotKit
``` html
<CopilotKit
  runtimeUrl={AGENT_URL}
  agent={AGENT_ID}
  useSingleEndpoint={false}
  headers={{ 'X-CSRF': 1 }} >
```
2. When the CopilotKitRuntime requests are made the /info and other request succeed but the /threads ones fail to pass the header and get rejected by the backend proxy.

### ✅ Expected Behavior

The X-CSRF: 1 header should be included on the /threads requests

### ❌ Actual Behavior

However the /threads request is not passing the X-CSRF (or any other headers) defined. This causes the backend proxy to reject the request with 403 - failing the proxy header check (this would be the same with Authorization header).

### Screenshots

<img width="769" height="112" alt="Image" src="https://github.com/user-attachments/assets/f6832491-7c87-44d9-b6d9-e6e88995478d" />

### 𝌚 CopilotKit Version

```shell
+-- @copilotkit/react-core@1.57.1
+-- @copilotkit/react-ui@1.57.1
+-- @copilotkit/runtime@1.57.1
```

### 📄 Logs (Optional)

```shell

```

BenTaylorDev:
Hello, and thanks for the well-structured bug report.

The fix in #5300 should resolve this issue.

PR #5300, opened 2026-06-06
fix(react-core): refresh thread headers on provider updates

sjh9714:
## What does this PR do?

Fixes #5282.

Provider header updates already reach the core instance via `setHeaders`, but React consumers of `useCopilotKit()` were only re-rendered for runtime connection status changes. That left `useThreads()` with stale context after the provider `headers` prop changed, so subsequent `/threads` requests could miss headers such as `X-CSRF`.

This subscribes the React context hook to `onHeadersChanged` and adds a provider-level regression test that verifies `/threads` is refetched with the updated header.

## Related PRs and Issues

- Fixes https://github.com/CopilotKit/CopilotKit/issues/5282

## Testing

- `pnpm install --frozen-lockfile`
- `pnpm --dir packages/react-core exec vitest run src/v2/hooks/__tests__/use-threads-provider-headers.e2e.test.tsx` (failed before the production change, passed after)
- `pnpm --dir packages/react-core exec vitest run src/v2/hooks/__tests__/use-threads.test.tsx src/v2/hooks/__tests__/use-threads-provider-headers.e2e.test.tsx`
- `pnpm exec oxlint packages/react-core/src/v2/context.ts packages/react-core/src/v2/hooks/__tests__/use-threads-provider-headers.e2e.test.tsx`
- `pnpm exec oxfmt --check packages/react-core/src/v2/context.ts packages/react-core/src/v2/hooks/__tests__/use-threads-provider-headers.e2e.test.tsx`
- `pnpm nx run @copilotkit/react-core:build`
- `pnpm nx run @copilotkit/react-core:test`
- `git commit -m "fix(react-core): refresh thread headers on provider updates"` (pre-commit ran `pnpm run test && pnpm run check:packages`; commit-msg ran `pnpm commitlint --edit`)

Also checked: `pnpm nx run @copilotkit/react-core:check-types` currently fails before checking this package because dependency package type/build errors are reported in `@copilotkit/shared` and `@copilotkit/a2ui-renderer`; I did not change those packages.

Note: I used Codex while preparing this change, reviewed the final diff, and ran the listed checks locally.

## Checklist

- [x] I have read the [Contribution Guide](https://github.com/copilotkit/copilotkit/blob/master/CONTRIBUTING.md)
- [ ] If the PR changes or adds functionality, I have updated the relevant documentation (not applicable: internal header propagation bug fix)
- [x] "Allow edits by maintainers" is checked (lets us help iterate on your PR directly - faster turnaround for everyone)

PR #5302, opened 2026-06-06
test(core): cover thread request headers

Tushar-Khandelwal-2004:
## Summary

Add regression coverage for configured CopilotKit headers on `/threads` requests.

Issue #5282 reported that headers like `X-CSRF` were not being sent on `/threads` requests when using multi-endpoint transport (`useSingleEndpoint={false}`), causing backend proxy CSRF/auth checks to reject the requests. Current main already forwards thread headers through the core thread store, so this PR locks that behavior down with focused tests.

## Changes

- Add a core thread-store regression test asserting configured headers are forwarded to all thread REST calls:
  - list threads
  - subscribe credentials
  - next page
  - rename
  - archive
  - delete
- Add a React `useThreads` regression test asserting provider headers reach thread list and subscribe requests.

## Testing

- `pnpm nx run @copilotkit/core:test --run src/__tests__/threads.test.ts`
- `pnpm -C packages/react-core exec vitest run src/v2/hooks/__tests__/use-threads.test.tsx`

### DIFF
--- a/packages/core/src/__tests__/threads.test.ts
+++ b/packages/core/src/__tests__/threads.test.ts
@@ -359,6 +359,63 @@ describe("thread store", () => {
     });
   });
 
+  it("forwards configured headers to all thread REST requests", async () => {
+    const fetchMock = vi
+      .fn()
+      .mockResolvedValueOnce({
+        ok: true,
+        json: async () => ({
+          threads: sampleThreads,
+          joinCode: "jc-1",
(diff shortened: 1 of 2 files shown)


BenTaylorDev:
Hello, and thanks for the PR.  

We had a related PR (https://github.com/CopilotKit/CopilotKit/pull/5300) at around the same time that hits the issue more directly, and we are going to accept that in place of this one.

PR #5308, opened 2026-06-07
fix(sdk-python): restore frontend tool history faithfully in LangGraph middleware

### LINKED ISSUE #4759: 🐛 Bug: LangGraph Middleware lies to agent about FE tools execution
### 🐛 Bug explained

The Middleware removes frontend tool calls in [here](https://github.com/CopilotKit/CopilotKit/blob/main/sdk-python/copilotkit/copilotkit_lg_middleware.py#L482)

```python
# Create updated AIMessage with only backend tool calls
        updated_ai_message = AIMessage(
            content=last_message.content,
            tool_calls=backend_tool_calls,
            id=last_message.id,
        )
```

This triggers a tool node execution for the backend tool. 

Then the LLM is called again in the same turn (tool calling loop), but the LLM **DOES NOT SEE that it called a FE tool** as well.

So it runs the risk of calling the tool again.

### 🔨 Possible fixes

I think either the messages history should be fixed before calling the LLM again or [before this line](https://github.com/CopilotKit/CopilotKit/blob/main/sdk-python/copilotkit/copilotkit_lg_middleware.py#L482) add a placeholder AI FE tool request + AI ToolMessage response, before adding the messages for BE tools.

### ♻️ Reproduction Code

Run a single round interaction, where agent requests 1FE tool + 1BE tool. 

Bug: On the second LLM call, the messages do not show that the LLM requested the FE tool on the first call.

<details><summary>Expand for code</summary>

```python
from langchain.agents import create_agent
from langchain_core.messages import AIMessage, HumanMessage
from langchain_core.tools import tool
from langgraph.checkpoint.memory import MemorySaver
from langchain_core.language_models.chat_models import BaseChatModel
from langchain_core.messages import AIMessageChunk, BaseMessage
from langchain_core.messages.tool import ToolCallChunk
from langchain_core.outputs import ChatGenerationChunk, ChatResult
from pydantic import Field, PrivateAttr

import json
from typing import Any

class MockChatModel(BaseChatModel):
    """Mock LLM that yields pre-configured `AIMessage`s in order.
    `responses` is a list of either single `AIMessage`s or lists of them
    (one list = one stream of chunks).
    """

    responses: list[Any] = Field(default_factory=list)
    _call_count: int = PrivateAttr(default=0)
    _call_args_list: list = PrivateAttr(default_factory=list)

    @property
    def _llm_type(self) -> str:
        return "mock"

    @property
    def call_count(self) -> int:
        return self._call_count

    @property
    def stream_call_args_list(self) -> list:
        return self._call_args_list

    def _generate(self, messages: list[BaseMessage], stop=None, run_manager=None, **kwargs) -> ChatResult:  # noqa: ARG002
        pass

    def _stream(self, messages: list[BaseMessage], stop=None, run_manager=None, **kwargs):  # noqa: ARG002
        self._call_args_list.append(list(messages))
        response = self.responses[self._call_count]
        self._call_count += 1
        chunks = response if isinstance(response, list) else [response]
        for msg in chunks:
            tool_call_chunks = []
            for call_chunk in (msg.tool_calls or []):
                args = call_chunk["args"]
                tool_call_chunks.append(ToolCallChunk(
                    name=call_chunk["name"],
                    args=json.dumps(args) if isinstance(args, dict) else args,
                    id=call_chunk["id"],
                    index=0,
                ))
            yield ChatGenerationChunk(
                message=AIMessageChunk(
                    content=msg.content,
                    tool_call_chunks=tool_call_chunks,
                    id=getattr(msg, "id", None),
                )
            )

    def bind_tools(self, tools: Any, **kwargs: Any) -> "MockChatModel":  # noqa: ARG002
        return self

_FRONTEND_TOOLS_FOR_TOPOLOGY_TESTS = [
    {
        "name": "_frontend_tool",
        "description": "frontend tool",
        "parameters": {
            "type": "object",
            "properties": {"item": {"type": "object"}},
        },
    },
]


async def _run_via_stream_events(graph, state: dict, config: dict) -> None:
    """Drive the graph via `astream_events` exactly the way `ag_ui_langgraph`
    does.
    """
    async for _ in graph.astream_events(
        input=state,
        config=config,
        subgraphs=True,
        version="v2",
    ):
        pass


# Single backend tool used to flip create_agent into the
# "install model-loop edges" branch. It changes the compiled graph topology.
@tool
def _backend_tool(x: str) -> str:
    """No-op backend tool used to materialise the `tools` node + loop edges."""
    return f"echoed: {x}"



def _build_graph_with_tools(model: MockChatModel, tools: list, checkpointer: MemorySaver):
    return create_agent(
        model=model,
        tools=tools,
        middleware=[CopilotKitMiddleware()],
        checkpointer=checkpointer,
    )

# Printing utils
def _dump(label: str, messages: list) -> None:
        print(f"\n--- {label} ---")
        for i, m in enumerate(messages):
            tcs = getattr(m, "tool_calls", None) or []
            print(
                f"  [{i}] {type(m).__name__} id={getattr(m, 'id', None)!r} "
                f"content={m.content!r} tool_calls={[tc.get('name') for tc in tcs]}"
            )


async def test_in_turn_fe_tool_missing_when_a_backend_tool_is_registered():
    """With a backend tool registered, the conditional `after_model → model`
    edge is installed and the model is invoked multiple times within a
    single `agent.astream_events()`. The mock emits a mixed `tool_calls` set
    (frontend + backend); after `after_model` strips the frontend call,
    `tools_condition` sees the surviving backend call, ToolNode runs, and
    control loops back to the model. On the next iteration the AIMessage
    no longer carries the frontend call,
    which can easily trigger another wrong invocation of FE tools.
    """
    cp = MemorySaver()
    config = {"configurable": {"thread_id": "backend-tools"}}

    model = MockChatModel(responses=[
        # Iteration 1: frontend + backend tool_calls. after_model strips
        # the frontend; ToolNode runs the backend; loop back to model.
        AIMessage(id="ai_iter1", content="", tool_calls=[
            {"name": "_frontend_tool", "args": {"item": {"name": "a"}}, "id": "call_fe1"},
            {"name": "_backend_tool", "args": {"x": "ping"}, "id": "call_be1"},
        ]),
        # Iteration 2:  Respond user (model "didn't see" its own frontend call.)
        AIMessage(id="ai_iter2", content="done"),

    ])
    graph = _build_graph_with_tools(model, [_backend_tool], cp)

    await _run_via_stream_events(
        graph,
        {
            "messages": [HumanMessage(id="h1", content="add and process")],
            "copilotkit": {"actions": _FRONTEND_TOOLS_FOR_TOPOLOGY_TESTS},
        },
        config,
    )

    # Helpful prints
    _dump("tool-loop-iter-1 model input history", model.stream_call_args_list[0])
    _dump("tool-loop-iter-2 model input history", model.stream_call_args_list[1])
    state = await graph.aget_state(config)
    _dump("checkpoint messages after round 1", state.values.get("messages", []))

    # Model called twice, with 2nd call after backend tool execution.
    assert model.call_count == 2


    # The actual bug condition: on the second model invocation the AIMessage
    # On the loop-back (iter-2) the model's input history must reflect what
    # `CopilotKitMiddleware.after_model` left behind: the frontend tool_call
    # was stripped, the backend tool_call survived (and ToolNode answered it).
    iter2_input = model.stream_call_args_list[1]
    iter2_tool_call_names = {
        tc.get("name")
        for m in iter2_input
        if isinstance(m, AIMessage)
        for tc in (getattr(m, "tool_calls", None) or [])
    }
    assert "_frontend_tool" not in iter2_tool_call_names
    assert "_backend_tool" in iter2_tool_call_names


if __name__ == "__main__":
    import asyncio
    asyncio.run(test_in_turn_fe_tool_missing_when_a_backend_tool_is_registered())
```

</details>

<details><summary>Expand for output</summary>

```bash
--- tool-loop-iter-1 model input history ---
  [0] HumanMessage id='h1' content='add and process' tool_calls=[]

--- tool-loop-iter-2 model input history ---
  [0] HumanMessage id='h1' content='add and process' tool_calls=[]
  [1] AIMessage id='ai_iter1' content='' tool_calls=['_backend_tool']
  [2] ToolMessage id='8a381f0d-e9e1-4071-b1c6-601837cd2000' content='echoed: ping' tool_calls=[]

--- checkpoint messages after round 1 ---
  [0] HumanMessage id='h1' content='add and process' tool_calls=[]
  [1] AIMessage id='ai_iter1' content='' tool_calls=['_backend_tool', '_frontend_tool']
  [2] ToolMessage id='8a381f0d-e9e1-4071-b1c6-601837cd2000' content='echoed: ping' tool_calls=[]
  [3] AIMessage id='ai_iter2' content='done' tool_calls=[]
```

</details>

### ✅ Expected Behavior

Model should see that it called FE tools during the tool loop iterations.

### ❌ Actual Behavior

Model does not see the FE tools as stripped by the Middleware, see output in ♻️ Reproduction Code.

### 𝌚 CopilotKit Version

In `main`: commit  [4414188](https://github.com/CopilotKit/CopilotKit/commit/44141889108895f774bef2410b59f9a4976e6ca6)

mvanhorn:
## Description

Fixes the LangGraph middleware "lying" to the agent about frontend tool execution (#4759). When a model turn calls a `useFrontendTool` handler, the middleware strips the FE tool calls so the backend `ToolNode` only runs calls it actually made, then rehydrates them afterward. Previously the rehydration could leave the restored AI message and its tool history inconsistent, so providers that require every tool call to have a matching result would break, and the agent saw an incorrect picture of what executed.

## Changes

- `_restore_intercepted_tool_call_history` rebuilds the intercepted FE tool calls with synthetic `ToolMessage` results only for the restored model history (so providers that require every `tool_call` to have a result stay valid), and preserves the original AI message's `name`, `additional_kwargs`, and `response_metadata` on restore.
- Tracks `original_tool_calls` in the CopilotKit private state so the original call set is restored exactly.

## Testing

`uv run pytest tests/test_copilotkit_lg_middleware.py` — 68 passed, including new cases asserting the restored AI message keeps its tool-call ids, name, `additional_kwargs`, and `response_metadata`, and that tool results line up with their calls.

Closes #4759

AI was used for assistance.

### DIFF
--- a/sdk-python/copilotkit/copilotkit_lg_middleware.py
+++ b/sdk-python/copilotkit/copilotkit_lg_middleware.py
@@ -59,6 +59,7 @@ _a2ui_tools_by_thread: dict[str, Any] = {}
 # checkpointer). Collisions across concurrent context-less runs are an
 # acceptable edge — the deployed path always carries a thread id.
 _DEFAULT_THREAD_KEY = "__copilotkit_a2ui_default__"
+_FRONTEND_TOOL_RESULT_CONTENT = json.dumps({"status": "forwarded_to_frontend"})
 
 
 def _current_thread_id() -> "str | None":
@@ -446,6 +447,10 @@ class CopilotKitMiddleware(AgentMiddleware[StateSchema, Any]):
     ) -> ModelResponse:
         _extract_forwarded_headers_from_config()
         _ensure_httpx_hook(request.model)
+        self._restore_intercepted_tool_call_history(
+            request.messages,
+            request.state.get("copilotkit", {}),
+        )
         request = self._apply_state_note(request)
 
         a2ui_tool = self._maybe_build_a2ui_tool(request)
@@ -641,6 +646,139 @@ class CopilotKitMiddleware(AgentMiddleware[StateSchema, Any]):
 
         return messages
 
+    @staticmethod
+    def _frontend_tool_result_message(tool_call: dict[str, Any]) -> ToolMessage | None:
+        tool_call_id = tool_call.get("id")
+        if not tool_call_id:
+            return None
+
+        return ToolMessage(
+            content=_FRONTEND_TOOL_RESULT_CONTENT,
+            tool_call_id=tool_call_id,
+            name=tool_call.get("name"),
+            id=f"copilotkit-fe-tool-result-{tool_call_id}",
+        )
+
+    @staticmethod
+    def _copy_ai_message_with_tool_calls(
+        message: AIMessage,
+        tool_calls: list[dict[str, Any]],
+    ) -> AIMessage:
+        if hasattr(message, "model_copy"):
+            return message.model_copy(update={"tool_calls": tool_calls})
+        return message.copy(update={"tool_calls": tool_calls})
+
+    @classmethod
+    def _restore_intercepted_tool_call_history(
+        cls,
+        messages: list,
+        copilotkit_state: dict[str, Any],
+    ) -> bool:
+        """Rehydrate FE tool calls with synthetic results before model history.
+
+        ``after_model`` strips FE calls so the backend ToolNode only executes
+        backend calls. Once backend ToolMessages have been appended, the next
+        LLM call still needs to see the full assistant turn, including the FE
+        calls it already made. Synthetic ToolMessages make that restored
+        history valid for providers that require every tool call to be
+        answered.
+        """
+        intercepted_tool_calls = copilotkit_state.get("intercepted_tool_calls")
+        original_message_id = copilotkit_state.get("original_ai_message_id")
+        original_tool_calls = copilotkit_state.get("original_tool_calls")
+
+        if not intercepted_tool_calls or not original_message_id:
+            return False
+
+        intercepted_by_id = {
+            call.get("id"): call
+            for call in intercepted_tool_calls
+            if isinstance(call, dict) and call.get("id")
+        }
+        if not intercepted_by_id:
+            return False
+
+        for idx, msg in enumerate(messages):
+            if not isinstance(msg, AIMessage) or msg.id != original_message_id:
+                continue
+
+            changed = False
+            existing_tool_calls = list(getattr(msg, "tool_calls", None) or [])
+            existing_tool_call_ids = {
+                call.get("id")
+                for call in existing_tool_calls
+                if isinstance(call, dict) and call.get("id")
+            }
+            missing_tool_calls = [
+                call
+                for call in intercepted_tool_calls
+                if (
+                    isinstance(call, dict)
+                    and call.get("id") not in existing_tool_call_ids
+                )
+            ]
+
+            if original_tool_calls:
+                full_tool_calls = list(original_tool_calls)
+            else:
+                full_tool_calls = [*existing_tool_calls, *missing_tool_calls]
+
+            if missing_tool_calls or full_tool_calls != existing_tool_calls:
+                messages[idx] = cls._copy_ai_message_with_tool_calls(
+                    msg,
+                    full_tool_calls,
+                )
+                changed = True
+
+            tool_messages_end = idx + 1
+            while (
+                tool_messages_end < len(messages)
+                and isinstance(messages[tool_messages_end], ToolMessage)
+            ):
+                tool_messages_end += 1
+
+            existing_tool_messages = messages[idx + 1 : tool_messages_end]
+            tool_messages_by_id: dict[str, ToolMessage] = {}
+            for tool_message in existing_tool_messages:
+                tool_call_id = getattr(tool_message, "tool_call_id", None)
+                if tool_call_id and tool_call_id not in tool_messages_by_id:
+                    tool_messages_by_id[tool_call_id] = tool_message
+
+            ordered_tool_messages = []
+            used_tool_message_ids = set()
+            for tool_call in full_tool_calls:
+                if not isinstance(tool_call, dict):
+                    continue
+
+                tool_call_id = tool_call.get("id")
+                if not tool_call_id:
+                    continue
+
+                tool_message = tool_messages_by_id.get(tool_call_id)
+                if tool_message is None and tool_call_id in intercepted_by_id:
+                    tool_message = cls._frontend_tool_result_message(tool_call)
+                    if tool_message is not None:
+                        changed = True
+
+                if tool_message is not None:
+                    ordered_tool_messages.append(tool_message)
+                    used_tool_message_ids.add(tool_call_id)
+
+            ordered_tool_messages.extend(
+                tool_message
+                for tool_message in existing_tool_messages
+                if getattr(tool_message, "tool_call_id", None)
+                not in used_tool_message_ids
+            )
+
+            if ordered_tool_messages != existing_tool_messages:
+                messages[idx + 1 : tool_messages_end] = ordered_tool_messages
+                changed = True
+
+            return changed
+
+        return False
+
     async def awrap_model_call(
         self,
         request: ModelRequest,
@@ -648,6 +786,10 @@ class CopilotKitMiddleware(AgentMiddleware[StateSchema, Any]):
     ) -> ModelResponse:
         _extract_forwarded_headers_from_config()
         _ensure_httpx_hook(request.model)
+        self._restore_intercepted_tool_call_history(
+            request.messages,
+            request.state.get("copilotkit", {}),
+        )
         self._fix_messages_for_bedrock(request.messages)
         request = self._apply_state_note(request)
 
@@ -872,11 +1014,12 @@ class CopilotKitMiddleware(AgentMiddleware[StateSchema, Any]):
         if not frontend_tool_calls:
             return None
 
-        # Create updated AIMessage with only backend tool calls
-        updated_ai_message = AIMessage(
-            content=last_message.content,
-            tool_calls=backend_tool_calls,
-            id=last_message.id,
+        # Keep only backend calls for the ToolNode. The full assistant turn is
+        # restored with synthetic FE ToolMessages before the next model call
+        # and before the agent exits.
+        updated_ai_message = self._copy_ai_message_with_tool_calls(
+            last_message,
+            backend_tool_calls,
         )
 
         return {
@@ -884,6 +1027,7 @@ class CopilotKitMiddleware(AgentMiddleware[StateSchema, Any]):
             "copilotkit": {
                 "intercepted_tool_calls": frontend_tool_calls,
                 "original_ai_message_id": last_message.id,
+                "original_tool_calls": tool_calls,
             },
         }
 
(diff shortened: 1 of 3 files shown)


BenTaylorDev (changes requested):
Thanks for this — the core diagnosis is right and the **mid-loop restore** half is a genuine fix: replaying the intercepted frontend calls into `wrap`/`awrap` so the model sees its full assistant turn addresses a real, still-present bug (#4759). Two concerns on the `after_agent` half before it can land:

1. **The synthetic `forwarded_to_frontend` ToolMessage collides with the existing orphan-handoff contract.** On `main`, `after_agent` deliberately leaves the frontend tool call as an *orphan* (no result) — that's the slot the real frontend result fills, and `_fix_messages_for_bedrock` (`copilotkit_lg_middleware.py:563`) + langchain's `patch_orphan_tool_calls` reposition/dedup it by matching `_INTERRUPTED_PAT` (`:585`, the "…was interrupted before completion." placeholder). This PR persists a synthetic `{"status":"forwarded_to_frontend"}` result whose content **doesn't** match that pattern, so it's treated as a second real result and the reposition path is skipped — on the Bedrock path the real frontend result can end up non-adjacent and then get stripped as unanswered.
2. **Sync/async asymmetry.** `awrap_model_call` runs `_restore_intercepted_tool_call_history` *and* `_fix_messages_for_bedrock` (`:789`, `:793`), but the sync `wrap_model_call` (`:450`) only runs the restore — so the synthetic messages it injects never go through the Bedrock dedup, risking a duplicate `tool_call_id` reaching the provider.

The tests hand-build message lists and never exercise the real round-trip (checkpoint persist → `add_messages` merge → `patch_orphan_tool_calls` → real FE result reintegration → dedup), so they don't catch either. Could you (a) confirm the real frontend-result round-trip still keeps the result adjacent given the synthetic no longer matches `_INTERRUPTED_PAT` — or better, avoid persisting a synthetic result and keep the orphan-handoff contract intact — and (b) run the sync path through `_fix_messages_for_bedrock` too? The mid-loop restore looks good on its own.

mvanhorn:
Both concerns addressed in 9f847a2e.

1. The synthetic `forwarded_to_frontend` result is gone from `after_agent`. You were right that it collided with the orphan-handoff contract: it did not match `_INTERRUPTED_PAT`, so `_fix_messages_for_bedrock` and `patch_orphan_tool_calls` read it as a real second result and skipped the reposition path. `_restore_intercepted_tool_call_history` now takes `synthesize_frontend_tool_results`, which `after_agent` passes as `False`, so the frontend call stays an orphan for the real result to fill. Any synthetic result already sitting in history is dropped by its `copilotkit-fe-tool-result-<id>` id, so a mid-loop restore followed by `after_agent` does not leave one behind.

2. The sync/async asymmetry is closed: `wrap_model_call` now runs `_fix_messages_for_bedrock` after the restore, matching `awrap_model_call`.

The mid-loop restore path is unchanged, since that half was the part you said was a genuine fix.

Two tests changed with it: `test_after_agent_restores_intercepted_tool_calls_on_original_message` now asserts no ToolMessages survive, and the duplicate-history test asserts only the backend result remains.

Verification, with a caveat. `tests/test_copilotkit_lg_middleware.py` gives 61 passed, 7 failed. The 7 failures are all in `TestAutoA2UI` and are identical on clean HEAD, so they are not from this change; they come from a newer `langchain_core` rejecting a `MagicMock` in the `tool` decorator. The sdk is a Poetry project and I installed the deps into a plain venv rather than through Poetry, so that version skew is mine, not yours.

BenTaylorDev (changes requested):
Thanks for the quick turnaround — concern (2) is fully addressed: `wrap_model_call` now runs `_fix_messages_for_bedrock` after the restore, and the restore-then-fix ordering is correct. The mid-loop half is confirmed working; against a real `create_agent` + AG-UI adapter run, the second model call now sees `ai-1` with `['be-1','fe-1']` and both answered, where `main` shows only `['be-1']`. That's a genuine fix for the intra-turn half of #4759.

Concern (1) isn't resolved yet, though. The synthetic result still lands in the persisted checkpoint — I see `ToolMessage id='copilotkit-fe-tool-result-fe-1' content='{"status": "forwarded_to_frontend"}'` in the checkpoint after turn 1. Two reasons: `_restore_intercepted_tool_call_history` mutates `request.messages` in place, and that list is the state channel's, so the synthetic is written into state; and `after_agent` removes it by omitting it from the returned list, which `add_messages` cannot honor — dropping a message needs a `RemoveMessage`. So `synthesize_frontend_tool_results=False` and the id-filter don't actually keep the FE call orphaned.

To be fair: at the currently-resolved versions this is not a live regression — turn 2 behaves identically to `main` — because the `patch_orphan_tool_calls` placeholder the adapter's repair path keys off (`_ORPHAN_TOOL_MSG_RE`) is no longer produced by langchain/langgraph at all. But persisting a synthetic tool result into user checkpoints is durable state we can't easily walk back, and it silently defeats that repair path wherever it is active.

The smallest fix is to keep the rehydration request-scoped rather than state-scoped — restore into a copy, e.g. `request = request.override(messages=list(request.messages))` before the restore. I verified that shape keeps the mid-loop benefit (model still sees `['be-1','fe-1']`, both answered) while leaving the checkpoint's FE call orphaned and synthetic-free. It also makes the `synthesize_frontend_tool_results` flag and the id-filter unnecessary.

Two notes: I'd soften "Closes #4759" to the intra-turn half — the cross-turn path is still broken, but that's pre-existing (`_fix_messages_for_bedrock` mutating in place permanently erases the FE tool call from persisted state across turns), and worth a separate issue. And your 7 failures were the plain-venv skew, not a real break: installing via poetry the way CI does, `main` is 210 passed / 11 skipped and this branch is 214 passed / 11 skipped, all green.

mvanhorn:
Addressed the remaining review point. Sync and async model restoration now operate on copied message lists, so synthetic results no longer leak into checkpoint state. Dropped the obsolete synthesis flag and the result-ID filter that was not actually filtering, and `after_agent` now restores the frontend call as an orphan. Added sync and async regression coverage for request-scoped restoration.

`ruff format --check` and `ruff check` both pass. The focused pytest run could not collect locally because the Poetry/LangChain deps are not installed on my side.

mvanhorn:
Pushed tests for the checkpoint round-trip you asked about.

`after_agent` still restores tool_calls only and does not persist a synthetic `forwarded_to_frontend` result, so the real frontend result can fill the orphan. wrap/awrap copy messages before mid-loop restore so that synthetic result does not land in checkpoint state. New tests cover placeholder replacement after `patch_orphan_tool_calls`.

BenTaylorDev (approved):
Re-reviewed at `49bf4535a`. Concern (1) is resolved, and I verified it live rather than from the tests — a real `create_agent` + `InMemorySaver` two-turn run:

```
-- model call #2 (mid-loop restore) --  AI(ai-1) tool_calls=['be-1','fe-1']; TM(be-1); TM(fe-1 synthetic)
-- checkpoint after turn 1 --           AI(ai-1) tool_calls=['be-1','fe-1']; TM(be-1) only
   synthetic 'forwarded_to_frontend' in checkpoint?  False
   fe-1 left orphaned?                              True
```

`request.override(messages=list(request.messages))` + `model_copy` keeps the rehydration request-scoped exactly as suggested, and `after_agent` persists tool_calls only — the orphan-handoff contract is intact and no synthetic result reaches durable state. Suite is 217 passed / 11 skipped on the merged tree vs 210 / 11 on `main`, and CI is green across the full 3.10–3.14 matrix. Nice work on the `original_tool_calls` tracking too — restoring the exact original call set is better than the append-order reconstruction it replaces.

Two follow-ups, neither blocking:

1. The sync `_fix_messages_for_bedrock(request.messages)` I asked for in round 2 turns out to cost more than it buys. It mutates message *objects* in place (`msg.tool_calls = …` at `copilotkit_lg_middleware.py:595`), so the list copy doesn't shield state — on turn 2 the sync path now drops the real FE result from the request and permanently reduces `ai-1` to `['be-1']` in the checkpoint, where `main`'s sync path keeps both. That matches what the async path already does on `main`, so it's parity rather than a new class of bug, but the duplicate-`tool_call_id` risk that motivated my ask is gone now that nothing synthetic is persisted — so that one line can simply come out. My mistake for requesting it; sorry for the churn.
2. Worth softening `Closes #4759` to the intra-turn half. The cross-turn path is still broken by `_fix_messages_for_bedrock` corrupting persisted state, which is pre-existing and deserves its own issue.

Minor: `request.state.get("copilotkit", {})` (`:606`, `:950`) are the only two `request.state` reads in the file without the `or {}` guard every other site uses.

PR #5314, opened 2026-06-08
feat: update all dependencies to use latest a2ui implementation features

ranst91:
Upgrade of dependencies so the latest changes (mainly a2ui stuff) from ag-ui are reflected here


contextablemark (approved):
Looks good

PR #5318, opened 2026-06-08
feat(runtime): decouple agent display name from registry key in /info

### LINKED ISSUE #4775: 🚀 Feature Request: agentId != name
### Pre-flight Checklist

### Problem or Motivation

The `info` request handler forces `agentId` to be the same as the human friendly agent name.

```typescript
export async function handleGetRuntimeInfo({
  runtime,
  request,
}: HandleGetRuntimeInfoParameters) {
  try {
    const agents = await resolveAgents(runtime.agents, request);

    const agentEntries = await Promise.all(
      Object.entries(agents).map(async ([name, agent]) => {
```

### 🎤 Tell us your idea

It would be preferable if the agentId could be different from the displayed name.

sikandar100:
## What does this PR do?

Decouples an agent's human-facing display name from its routing key in the runtime's `/info` endpoint.

Today, the dictionary key passed to `CopilotRuntime({ agents: {...} })` serves two roles: it's the stable routing identifier (`agentId`) *and* the display name returned by `/info`. That forces developers to choose between a clean routing key (e.g. `"customer-support-v1"`) and a friendly UI label (e.g. `"Customer Support Specialist"`) — they can't have both.

This PR adds a small, opt-in mechanism: if an agent subclass declares a public `name` field, `/info` returns that as the display name. If not, behavior is identical to today (the registry key is used). No existing agents need to change.

### Example

```ts
import { BuiltInAgent, CopilotRuntime } from "@copilotkit/runtime/v2";

class CustomerSupportAgent extends BuiltInAgent {
  name = "Customer Support Specialist";
}

  agents: {
    "customer-support-v1": new CustomerSupportAgent({ model: "openai:gpt-5.4-mini" }),
  },
});
```

The routing key stays `"customer-support-v1"`; `/info` reports `"Customer Support Specialist"` as the agent's name.

## Changes

Four commits:

1. **`feat(runtime)`** — adds a `resolveAgentDisplayName` helper in `get-runtime-info.ts`. Reads `agent.name` when present, non-empty, and a string (trimmed); falls back to the registry key otherwise.
2. **`test(runtime)`** — 7 new test cases in `get-runtime-info.test.ts` covering: regression guard (no `name` → key fallback), happy path (`name` set → returned), whitespace trim, empty-string fallback, whitespace-only fallback, non-string fallback, and registry-key-independent display name.
3. **`docs(runtime)`** — adds a "Display names" section to `copilot-runtime.mdx` with a worked example.
4. **`docs(runtime)`** — notes that the same pattern works for `BuiltInAgent`, `HttpAgent`, `LangGraphAgent`, `LlamaIndexAgent`, `AgnoAgent`, and any other `AbstractAgent` subclass.

## Backward compatibility

Fully backward compatible:

- The `agents:` config shape on `CopilotRuntime` is unchanged.
- Agents without a `name` field behave exactly as before (display name equals registry key).
- The internal routing identity (`agentId`) is unchanged.
- Thread filtering, history lookup, and analytics are unchanged.
- The `AgentDescription` type in the public API is unchanged.

## Related PRs and Issues

Closes #4775.

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective and that the feature works
- [x] New and existing unit tests pass locally with my changes

### DIFF
--- a/packages/runtime/src/v2/runtime/__tests__/get-runtime-info.test.ts
+++ b/packages/runtime/src/v2/runtime/__tests__/get-runtime-info.test.ts
@@ -332,4 +332,168 @@ describe("handleGetRuntimeInfo", () => {
       message: "Failed to get agents",
     });
   });
+
+  describe("agent display name (decoupled from registry key)", () => {
+    /**
+     * Covers the behavior added for issue #4775: an agent that exposes a
+     * public `name` field provides its own human-facing display name in
+     * the /info response, while the registry-key (used for routing) stays
+     * the dictionary key. Agents without a `name` keep today's behavior
+     * (display name = registry key) — backward compatible.
+     */
+    const mockRequest = new Request("https://example.com/info");
+
+    it("falls back to the registry key when the agent does not expose a name (regression guard)", async () => {
+      const mockAgent = {
+        description: "support agent",
+        constructor: { name: "SupportAgent" },
+      };
+      const runtime = new CopilotRuntime({
+        agents: {
+          "support-v1": mockAgent as AbstractAgent,
+        },
+      });
+
+      const response = await handleGetRuntimeInfo({
+        runtime,
+        request: mockRequest,
+      });
+      const data = await response.json();
+
+      expect(data.agents["support-v1"].name).toBe("support-v1");
+    });
+
+    it("uses agent.name as the display name when the agent exposes one", async () => {
+      const mockAgent = {
+        name: "Customer Support Specialist",
+        description: "support agent",
+        constructor: { name: "SupportAgent" },
+      };
+      const runtime = new CopilotRuntime({
+        agents: {
+          "support-v1": mockAgent as unknown as AbstractAgent,
+        },
+      });
+
+      const response = await handleGetRuntimeInfo({
+        runtime,
+        request: mockRequest,
+      });
+      const data = await response.json();
+
+      expect(data.agents["support-v1"].name).toBe(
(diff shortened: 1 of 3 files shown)


BenTaylorDev (changes requested):
This is a solid change — the `/info` display-name decoupling is correct and genuinely opt-in: `resolveAgentDisplayName` falls back to the registry key, base `AbstractAgent` has no `name` field and no server subclass declares one, and routing keys off the agent id rather than `description.name`, so nothing silently changes. It's also orthogonal to the related #6120 (no file overlap).

The one thing blocking merge is a stale-branch conflict. `git merge-tree` shows a conflict confined to the **import block** in `get-runtime-info.ts` — `main` independently grew its imports (`isA2UIEnabled`, `ThreadEndpointRuntimeInfo`, `supportsLocalThreadEndpoints`) while this branch reflowed the `RuntimeLicenseStatus` import. Your load-bearing changes (the `resolveAgentDisplayName` helper + the one-line call) apply cleanly on top.

Could you rebase onto current `main`? Keep main's import block and re-apply your helper + call, then re-run `@copilotkit/runtime` tests. Once it's rebased this looks good to go — I held off approving since the rebase would dismiss it anyway.

PR #5319, opened 2026-06-08
fix(showcase): render AgentCore deploy pages with framework-scoped tabs

ranst91:
## Problem

`/strands/deploy-agentcore` and `/langgraph/deploy-agentcore` rendered only their title — the body was **empty**. The page MDX is just a shell:

```mdx
---
title: AWS AgentCore
...
---
<Content framework="strands" />
```

`<Content>` resolved to a **dead stub** in the MDX component registry (`({children}) => <div>{children}</div>`) that rendered nothing — even though the body content was authored all along in the shared partial `src/content/snippets/integrations/agentcore/index.mdx`. The stub's comment wrongly claimed the content "was never authored".

A secondary issue: the partial pre-selects the command tabs by framework, but the loader discarded that context, so both LangGraph **and** Strands tabs showed on every page.

## Fix

- **`mdx-registry.tsx`** — replace the dead `Content` stub with a dedicated component that renders the agentcore partial via `PartialLoader` and threads the page's `framework` into MDX scope.
- **`mdx-registry-loader.tsx`** — `PartialLoader` accepts an optional `scope`, forwarded to `MDXRemote` `options.scope`. (next-mdx-remote binds scope as **module-level identifiers**, not as the rendered component's `props` — `props.framework` could never resolve.)
- **`agentcore/index.mdx`** — reference `{framework}` (bare scope var) instead of `props.framework`, so `<AgentCoreCommandTabs>` collapses to the single relevant framework per page.

## Verification (local dev server, :3003)

| Page | Result |
|---|---|
| `/strands/deploy-agentcore` | 200, full content, **2× Strands tab / 0 LangGraph** |
| `/langgraph/deploy-agentcore` | 200, full content, **2× LangGraph tab / 0 Strands** |

Typecheck clean on touched files. New frameworks get the same behavior automatically via `<Content framework="…" />` — no per-page wiring.


contextablemark (approved):
looks good

PR #5327, opened 2026-06-08
docs: simplify README agent-skills section

onsclom:
Replaces the "Install as a Claude Code plugin" section and its skill-inventory table with our recommended one-line install:

```bash
npx copilotkit@latest skills install
```

**Why:**

- The hand-maintained skill table goes stale fast. It already listed 9 skills and 6 lifecycle slugs that no longer exist (there are 11 skills now).
- `npx copilotkit@latest skills install` is our new recommended way to install the skills. It works across Claude Code, Codex, Cursor, Gemini, and others, and always pulls the latest.

**Two commits, reviewed separately:**

1. `style: format README with oxfmt`. README has failed `oxfmt` on main since 1e70dae97 widened the lint-fix glob to cover markdown but never backfilled existing files. This runs the formatter once so the file conforms, which also keeps the content change below from being buried under whole-file reformat noise. Render-neutral: only emphasis markers and table column padding, which GitHub renders identically.
2. `docs: simplify README agent-skills section`. The actual content change. Surgical, one section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,19 +80,19 @@ https://github.com/user-attachments/assets/72b7b4f3-b6e7-460c-a932-5746fe3c8db3
 - **Generative UI** – Allows agents to generate and update UI components dynamically at runtime based on user intent and agent state.
 - **Shared State** – A synchronized state layer that both agents and UI components can read from and write to in real time.
 - **Human-in-the-Loop** – Lets agents pause execution to request user input, confirmation, or edits before continuing.
-- **Self-Learning** *(early access)* – Agents that continuously improve from user feedback via in-context reinforcement learning (CLHF).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -242,38 +242,17 @@ Here are a few useful resources to help you get started:
 
 - Want to contribute but not sure how? [Join our Discord](https://discord.gg/6dffbvGU3D) and we'll help you out!
 
-## Install as a Claude Code plugin
+## Agent skills
```
samjulien: Can you rebase from main, move this up after Quick Start, and make the header "Agent Skills" (title case instead of sentence case)?

samjulien (changes requested):

samjulien (approved):

PR #5331, opened 2026-06-08
build: exclude root README from oxfmt lint-fix hook

onsclom:
Excludes the root `README.md` from the `oxfmt` lint-fix pre-commit hook.

**Why:** The README is hand-tuned marketing prose with emoji-aligned tables. `oxfmt` re-pads those tables (it miscounts emoji width) and rewrites emphasis markers. `main` already fails `oxfmt` on this file, so editing one section reformats the whole file and forces a `--no-verify` commit. Excluding it keeps README edits surgical.

**Verified:** with only `README.md` staged, the lint-fix hook now reports `skip — no files for inspection` and leaves the file untouched.

One line of config (plus a comment):

```yaml
exclude:
  - "README.md"
  - "docs/**"
  ...
```

### DIFF
--- a/lefthook.yml
+++ b/lefthook.yml
@@ -56,6 +56,12 @@ pre-commit:
       # oxfmt exits 2 with "Expected at least one target file" and blocks
       # the commit.
       exclude:
+        # Root README is hand-tuned marketing prose with emoji-aligned tables
+        # that oxfmt re-pads (it miscounts emoji display width) and emphasis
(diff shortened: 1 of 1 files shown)


onsclom:
Closing in favor of folding the README formatting into #5327 as a separate commit. Excluding README from the formatter would have reversed the original intent of 1e70dae97 (which deliberately brought markdown under oxfmt); the better fix is to backfill the formatting that was never run, so the file conforms and future edits stay surgical.

PR #5345, opened 2026-06-09
docs(skills): fix outdated APIs in copilotkit-setup skill

onsclom:
Fixes 4 accuracy issues surfaced by a docs gardener agent reviewing the `copilotkit-setup` skill. Docs-only; no runtime code touched.

| # | Issue | Fix | Proof (source) |
|---|-------|-----|----------------|
| 1 | `licenseKey` is not a valid `CopilotKitProvider` prop | Use `publicLicenseKey` (alias of `publicApiKey`) | [`publicApiKey` L102](https://github.com/CopilotKit/CopilotKit/blob/01f89396a494cd44437979730ded6d9dcbbc02ab/packages/react-core/src/v2/providers/CopilotKitProvider.tsx#L102), [`publicLicenseKey` L106](https://github.com/CopilotKit/CopilotKit/blob/01f89396a494cd44437979730ded6d9dcbbc02ab/packages/react-core/src/v2/providers/CopilotKitProvider.tsx#L106) |
| 2 | `createCopilotEndpoint` / `createCopilotEndpointSingleRoute` are `@deprecated` | Use `createCopilotHonoHandler` with `mode: "single-route"` | [`createCopilotEndpoint` deprecated alias, hono.ts L52-55](https://github.com/CopilotKit/CopilotKit/blob/01f89396a494cd44437979730ded6d9dcbbc02ab/packages/runtime/src/v2/runtime/endpoints/hono.ts#L52-L55), [single-route deprecated, hono-single.ts L24](https://github.com/CopilotKit/CopilotKit/blob/01f89396a494cd44437979730ded6d9dcbbc02ab/packages/runtime/src/v2/runtime/endpoints/hono-single.ts#L24) |
| 3 | `createCopilotEndpointExpress` / `...SingleRouteExpress` are `@deprecated` | Use `createCopilotExpressHandler` with `mode: "single-route"` | [`createCopilotEndpointExpress` deprecated alias, express.ts L88-91](https://github.com/CopilotKit/CopilotKit/blob/01f89396a494cd44437979730ded6d9dcbbc02ab/packages/runtime/src/v2/runtime/endpoints/express.ts#L88-L91), [single-route deprecated, express-single.ts L41](https://github.com/CopilotKit/CopilotKit/blob/01f89396a494cd44437979730ded6d9dcbbc02ab/packages/runtime/src/v2/runtime/endpoints/express-single.ts#L41) |
| 4 | `headers` prop type incomplete | Document `Record<string, string> \| (() => Record<string, string>)`, evaluated per-request | [headers union type, CopilotKitProvider.tsx L94](https://github.com/CopilotKit/CopilotKit/blob/01f89396a494cd44437979730ded6d9dcbbc02ab/packages/react-core/src/v2/providers/CopilotKitProvider.tsx#L94) |

### Notes for review
- Both deprecated factory families collapse into a single factory each (`createCopilotHonoHandler` / `createCopilotExpressHandler`), differentiated by `mode` (`"multi-route"` is the default). The deprecated names still work, so existing user code is unaffected; only the recommended docs path changed.
- `createCopilotExpressHandler` defaults to `cors: true` ([express.ts L91](https://github.com/CopilotKit/CopilotKit/blob/01f89396a494cd44437979730ded6d9dcbbc02ab/packages/runtime/src/v2/runtime/endpoints/express.ts#L91)), matching the old `...SingleRouteExpress` wrapper, so the single-route example keeps identical CORS behavior without an extra option.
- Quick Reference table rewritten with a `Mode` column; a one-line note keeps the deprecated names discoverable.
- Skill `version` bumped from 1.1.1 to 1.1.2.

### DIFF
--- a/skills/copilotkit-setup/SKILL.md
+++ b/skills/copilotkit-setup/SKILL.md
@@ -222,7 +222,9 @@ npm install hono @hono/node-server
 
 ### Step 3: Set up the frontend provider
 
-Wrap your application with `CopilotKitProvider` from `@copilotkit/react-core/v2`.
+Wrap your application with `CopilotKit` from `@copilotkit/react-core/v2`.
+
+> **Which provider component?** Always use `CopilotKit` imported from `@copilotkit/react-core/v2`. It is the compatibility bridge across v1 and v2 and a strict superset of the other provider APIs. Do **not** use `CopilotKit` from the package root (`@copilotkit/react-core`, legacy v1) or `CopilotKitProvider` from `/v2` (a subset of the functionality).
 
 **Important:** Import the stylesheet in your root layout:
 
@@ -237,15 +239,15 @@ In `src/app/page.tsx` (or a client component):
 ```tsx
 "use client";
 
-import { CopilotKitProvider, CopilotChat } from "@copilotkit/react-core/v2";
+import { CopilotKit, CopilotChat } from "@copilotkit/react-core/v2";
 
 export default function Home() {
   return (
-    <CopilotKitProvider runtimeUrl="/api/copilotkit">
+    <CopilotKit runtimeUrl="/api/copilotkit">
       <div style={{ height: "100vh" }}>
         <CopilotChat />
       </div>
-    </CopilotKitProvider>
+    </CopilotKit>
   );
 }
 ```
@@ -255,17 +257,14 @@ export default function Home() {
 When the runtime runs on a separate server (e.g., Express on port 4000):
 
 ```tsx
-<CopilotKitProvider
-  runtimeUrl="http://localhost:4000/api/copilotkit"
-  useSingleEndpoint
->
+<CopilotKit runtimeUrl="http://localhost:4000/api/copilotkit" useSingleEndpoint>
   {children}
-</CopilotKitProvider>
+</CopilotKit>
 ```
 
 Set `useSingleEndpoint` when the backend uses single-route endpoints (`createCopilotHonoHandler` or `createCopilotExpressHandler` with `mode: "single-route"`).
 
-#### CopilotKitProvider key props
+#### CopilotKit key props
 
 | Prop                | Type                                                       | Description                                                                                                          |
 | ------------------- | ---------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------- |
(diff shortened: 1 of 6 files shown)


tylerslaton (changes requested):
Quick preface - I know this is confusing. 

The TL;DR is to just only use:

```ts
import { CopilotKit } from "@copilotkit/react-core/v2
```

And not either of these:

```ts
import { CopilotKit } from "@copilotkit/react-core
import { CopilotKitProvider } from "@copilotkit/react-core/v2
```

---

One thing the agent got wrong in these skills: we actually don’t want to recommend ‎`CopilotKitProvider`.

‎`CopilotKit` from ‎`v2` is the compatibility bridge across both ‎`v1` and ‎`v2`, and is a strict superset of the older ‎`CopilotKit` / ‎`CopilotKitProvider` APIs. If we point users at ‎`CopilotKitProvider`, they only get a subset of the functionality, so the agent should always prefer ‎`CopilotKit` from ‎`v2`.

This *also* impacts the truth to `licenseKey` not being valid. It is present in `CopilotKit` from `v2`. Let's tackle that in this PR so its all in one place.

onsclom:
> Quick preface - I know this is confusing.
> 
> The TL;DR is to just only use:
> 
> ```ts
> import { CopilotKit } from "@copilotkit/react-core/v2
> ```
> 
> And not either of these:
> 
> ```ts
> import { CopilotKit } from "@copilotkit/react-core
> import { CopilotKitProvider } from "@copilotkit/react-core/v2
> ```
> 
> One thing the agent got wrong in these skills: we actually don’t want to recommend ‎`CopilotKitProvider`.
> 
> ‎`CopilotKit` from ‎`v2` is the compatibility bridge across both ‎`v1` and ‎`v2`, and is a strict superset of the older ‎`CopilotKit` / ‎`CopilotKitProvider` APIs. If we point users at ‎`CopilotKitProvider`, they only get a subset of the functionality, so the agent should always prefer ‎`CopilotKit` from ‎`v2`.
> 
> This _also_ impacts the truth to `licenseKey` not being valid. It is present in `CopilotKit` from `v2`. Let's tackle that in this PR so its all in one place.

Switched everything to CopilotKit from /v2 in ba99cbef8. One note: CopilotKit from /v2 doesn't have a licenseKey prop. Its props only define [publicApiKey](https://github.com/CopilotKit/CopilotKit/blob/5994e7be6/packages/react-core/src/components/copilot-provider/copilotkit-props.tsx#L20-L32)[ / ](https://github.com/CopilotKit/CopilotKit/blob/5994e7be6/packages/react-core/src/components/copilot-provider/copilotkit-props.tsx#L20-L32)[publicLicenseKey](https://github.com/CopilotKit/CopilotKit/blob/5994e7be6/packages/react-core/src/components/copilot-provider/copilotkit-props.tsx#L20-L32), and the component only [reads those two](https://github.com/CopilotKit/CopilotKit/blob/5994e7be6/packages/react-core/src/components/copilot-provider/copilotkit.tsx#L82), so I kept the publicLicenseKey rename. (Bare licenseKey only exists on the [Angular package](https://github.com/CopilotKit/CopilotKit/blob/5994e7be6/packages/angular/src/lib/config.ts#L14).)

tylerslaton (approved):

PR #5351, opened 2026-06-09
docs(skills): recommend CopilotKit from /v2 everywhere and scrub CopilotCloud mentions

onsclom:
Follow-up to #5345.

## What changed

- All skills now recommend `CopilotKit` from `@copilotkit/react-core/v2` instead of `CopilotKitProvider` (~140 mentions, 32 files). Safe rename: `CopilotKitProps extends Omit<CopilotKitProviderProps, "children">`, so every prop carries over.
- CopilotCloud / Copilot Cloud / CopilotKit Cloud mentions replaced with CopilotKit Intelligence (or deleted). Grep returns zero hits.
- Bonus: react-core's provider-setup.md claimed `publicApiKey` was canonical. Inverted to match #5345 (`publicLicenseKey` canonical, `publicApiKey` deprecated alias).

## Review notes

- Half the diff is generated: react-core, runtime, and a2ui-renderer live in `packages/*/skills/` and are mirrored into `skills/` by `pnpm sync:plugin-skills`. Review the `packages/` side; the `skills/` side is a copy. The sync also re-pins `.claude-plugin` versions to 1.59.5 (same as #5347).
- Remaining `CopilotKitProvider` mentions are intentional: literal file paths, "do not use" notes, and eval patterns that accept legacy code.
- copilotkit-upgrade tables distinguish old vs new by import path (root = v1, `/v2` = target) since the component name is unchanged.
- Left alone: endpoint URLs (`api.cloud.copilotkit.ai`), real identifiers (`MissingPublicApiKeyError`), generic infra terms (Cloudflare, Google Cloud).


tylerslaton (approved):
🙏🏻

PR #5321, opened 2026-06-08
Feat/ng a2ui exp

wolfmanfx:
## What does this PR do?

This PR expands the Angular package with first-class support for A2UI and Open Generative UI, while improving the chat experience and demo coverage.

- Adds Angular A2UI tool and activity renderers, including built-in renderer registration, surface syncing, custom catalog/theme support (using LIT), progress UI, and action bridging back to the active agent.
- Adds Open Generative UI support through the `generateSandboxedUi` tool, streamed activity rendering, sandboxed iframe previews, sandbox functions, and default design/tool guidance.
- Extends `CopilotChat` with attachments, suggestions, reasoning messages, transcription support, updated input/toolbar behavior, and richer public exports.
- Adds activity-message renderer configuration and support for passing the active agent into tool renderers.
- Updates the Angular demo, storybook examples, and demo server to exercise the new A2UI/Open Generative UI flows, including provider selection/OpenRouter support.
- Adds tests for A2UI web components/renderers, Open Generative UI rendering, partial HTML processing, chat view behavior, input handling, and related utilities.
- Updates Angular package/build configuration for the newer Angular toolchain and scoped generated styles.

## Testing

We run the unit tests as well manually tested with openrouter.

## Checklist

- [ x ] I have read the [Contribution Guide](https://github.com/copilotkit/copilotkit/blob/master/CONTRIBUTING.md)
- [ x ] If the PR changes or adds functionality, I have updated the relevant documentation
- [ x ] "Allow edits by maintainers" is checked (lets us help iterate on your PR directly — faster turnaround for everyone)

PR from Rainer Hahnekamp & Murat Sari (https://Soverius.ai)

PR #5356, opened 2026-06-10
fix: repair check-types rot surfaced by Angular A2UI PR

onsclom:
## What does this PR do?

Fixes the typecheck failures [reported on #5321](https://github.com/CopilotKit/CopilotKit/pull/5321#issuecomment-4669012390), where errors appeared in files that PR never touched. CI does not run `check-types` anywhere, so type errors accumulated silently in several packages.

### angular

`tsconfig.json` paths pointed at core and shared sources, so Angular's typecheck re-checked those packages under Angular's compiler settings. Paths now resolve to built `dist` declarations first, with a src fallback. Same pattern as the Vue package.

### a2ui-renderer

`A2UIViewer.tsx` and `theme/viewer-theme.ts` were dead code left over from the A2UI 0.8 to 0.9 migration. Nothing imports them, and they import `@a2ui/lit`, which is no longer a dependency. Deleted, along with the unused `@a2ui/lit` entries in `tsdown.config.ts`. Also underscored two unused type params on deprecated aliases.

### shared

`@copilotkit/license-verifier` 0.3.0 dropped `LicenseContextValue` and `LicenseMode`, but shared still re-exported them. Both types are now defined in shared itself, using the original definitions from license-verifier 0.2.0. Also fixed a string-index type error in `telemetry-client.ts`.

- `@copilotkit/core` check-types is still broken (about 230 errors). Separate, larger fix.
- CI should run `check-types` for the packages that now pass, to prevent this from recurring.

## Testing

- `check-types` passes for shared, a2ui-renderer, and angular (all failed before)
- Build and tests pass for shared, a2ui-renderer, angular, and react-core
- Repo grep confirms nothing references the deleted files

cc @rainerhahnekamp @wolfmanfx


packages/shared/src/index.ts
```diff
@@ -24,16 +24,40 @@ export const COPILOTKIT_VERSION = packageJson.version;
...
+  status: LicenseStatus | null;
+  /** Convenience: the license payload if valid, null otherwise. */
+  license: LicensePayload | null;
+  /** Whether a specific feature is licensed. Returns true if no licensing is active (no token). */
+  checkFeature: (feature: string) => boolean;
+  /** Get a numeric feature limit. Returns null if not applicable. */
+  getLimit: (feature: string) => number | null;
+}
+
+export type LicenseMode = "online" | "offline";
```
MikeRyanDev: This is a code smell to me. I suspect this means we've got dead code to remove. Happy to merge as-is if you want to clean up in a follow on PR.

packages/a2ui-renderer/src/theme/viewer-theme.ts
```diff
@@ -1,443 +0,0 @@
-/*
```
MikeRyanDev, line 1: We don't need this anymore?

packages/a2ui-renderer/src/A2UIViewer.tsx
```diff
@@ -1,213 +0,0 @@
-"use client";
```
MikeRyanDev, line 1: We don't need this anymore?

MikeRyanDev (changes requested):

packages/a2ui-renderer/src/theme/viewer-theme.ts
onsclom: Yes, nothing needs these
- Nothing in the repo references either file except each other (A2UIViewer imports viewer-theme)
- A2UIViewer was never in the public surface: src/index.ts doesn't export it and package.json exports only the root entry
- Both import @a2ui/lit, which isn't in a2ui-renderer's dependencies anymore

packages/a2ui-renderer/src/A2UIViewer.tsx
onsclom: https://github.com/CopilotKit/CopilotKit/pull/5356#discussion_r3390561798

packages/shared/src/index.ts
onsclom: You are right! `LicenseMode` was dead (zero consumers, and the old re-export already failed to compile), so I deleted it in this PR. The trickier smell, the whole context machinery is inert today. `createLicenseContextValue` only accepts null, so `checkFeature()` always returns true and the gates in CopilotChat/Sidebar/Popup never fire. Real license warnings come from `/info` status instead.

Removing it touches ~300 lines across 4 packages and public exports like `useLicenseContext`, so I'll do that in a follow-up PR. Quick question for that: rip it out, or wire it to `/info` if per-feature gating is still planned?

MikeRyanDev: Per-feature gating is still planned, so wire it to `/info`

MikeRyanDev (approved):

rainerhahnekamp:
Thanks, we'll be rebasing ours (#5321) and hope that it will be ready then as well.

PR #5360, opened 2026-06-10
fix(runtime): coalesce consecutive same-role Anthropic messages before dispatch

### LINKED ISSUE #2910: 🐛 Bug: When using Anthropic next assistant message is appended to previous one
### ♻️ Reproduction Steps

1. integrate copiliot kit
2. create an agent with Anthropic provider
3. chat with the model
...

### ✅ Expected Behavior

should return messages one after the other, assistant -> user -> assistant -> user

### ❌ Actual Behavior

https://github.com/user-attachments/assets/1fb47dbe-51fe-422f-af59-3890a1127209

Image

### 𝌚 CopilotKit Version

```shell
@copilotkit/react-core 1.50.1
@copilotkit/react-ui 1.50.1
@copilotkit/runtime 1.50.1
```

### 📄 Logs (Optional)

```shell

```

rodboev:
## Summary

In multi-turn conversations using the Anthropic adapter, a model turn that contains both assistant text and a tool call can be replayed as two consecutive `{role: "assistant"}` entries in the Anthropic payload. Anthropic expects one message object per turn with alternating roles, so that split payload can blur turn boundaries on the next request. This PR coalesces same-role Anthropic messages before dispatch so one assistant turn stays one assistant message.

## Root cause

[`convertMessageToAnthropicMessage`](https://github.com/CopilotKit/CopilotKit/blob/005aebbededbfdd7978b3c1ee221580b74dd088d/packages/runtime/src/service-adapters/anthropic/utils.ts#L142-L219)
maps each CopilotKit message independently. A mixed assistant turn, a `TextMessage(role=assistant)` followed by an `ActionExecutionMessage`, therefore becomes two separate assistant entries. The Anthropic Messages API requires all content blocks for one assistant turn to be sent in a single message object: https://docs.anthropic.com/en/api/messages

[`AnthropicAdapter.process()`](https://github.com/CopilotKit/CopilotKit/blob/005aebbededbfdd7978b3c1ee221580b74dd088d/packages/runtime/src/service-adapters/anthropic/anthropic-adapter.ts#L282-L366)
already deduplicates `tool_result` blocks on the user side, but it previously forwarded the mapped assistant-side payload without a coalescing pass. Rebasing onto current `main` also exposed test-fixture drift in the two new regression tests, so the final branch updates those fixtures to the current object-form `TextMessage` constructor and removes one stale `ActionInput` field while keeping the production fix unchanged.

## Changes

- `packages/runtime/src/service-adapters/anthropic/utils.ts`: add
  `coalesceConsecutiveSameRoleMessages(messages)` to merge adjacent equal-role
  Anthropic messages by concatenating their `content` arrays.
- `packages/runtime/src/service-adapters/anthropic/anthropic-adapter.ts`: call
  `coalesceConsecutiveSameRoleMessages` before `limitMessagesToTokenCount`.
- `packages/runtime/tests/service-adapters/anthropic/anthropic-adapter.test.ts`:
  add same-role regression coverage, align the new fixtures with the current
  object-form `TextMessage` constructor, and remove the stale `parameters` field
  from the action fixture.
- `.changeset/coalesce-anthropic-same-role-messages.md`: patch changeset for
  `@copilotkit/runtime`.

## Scope

The `tool_result` allowlist deduplication stays unchanged. Token trimming and the orphan-removal post-processor still run after coalescing, so `tool_use` and `tool_result` pairing remains intact. The OpenAI, Google, LangChain, and Groq adapters are unaffected.

The reported regression and the explicit regression coverage are assistant-side. The coalescing helper itself is role-agnostic for adjacent equal-role array content, but this PR does not add a separate user-side regression case.

## Related PRs and Issues

Prior attempted fix: #2864, which addressed unrelated message callbacks rather than the adapter payload.

## Test plan

- [x] `pnpm -C packages/runtime exec vitest run tests/service-adapters/anthropic/anthropic-adapter.test.ts`
  10/10 tests pass, including the two same-role coalescing cases.
- [x] `pnpm -C packages/runtime exec vitest run tests/service-adapters/anthropic/utils-token-trimming.test.ts`
  9/9 tests pass.

## Upstream

Closes #2910.
Reported by @alonronin.

### DIFF
--- a/.changeset/coalesce-anthropic-same-role-messages.md
+++ b/.changeset/coalesce-anthropic-same-role-messages.md
@@ -1,0 +1,9 @@
+---
+"@copilotkit/runtime": patch
+---
+
+fix(runtime): coalesce consecutive same-role Anthropic messages before dispatch
+
+Fixes multi-turn conversations with Anthropic where a text + tool_use turn produced
+two consecutive assistant messages in the API payload, violating role-alternation and
+causing subsequent assistant responses to appear appended to the previous message.
--- a/packages/runtime/src/service-adapters/anthropic/anthropic-adapter.ts
+++ b/packages/runtime/src/service-adapters/anthropic/anthropic-adapter.ts
@@ -33,6 +33,7 @@ import {
 import {
   convertActionInputToAnthropicTool,
   convertMessageToAnthropicMessage,
+  coalesceConsecutiveSameRoleMessages,
   limitMessagesToTokenCount,
 } from "./utils";
 
@@ -326,9 +327,15 @@ export class AnthropicAdapter implements CopilotServiceAdapter {
         return true;
       }) as Anthropic.Messages.MessageParam[];
 
+    // Coalesce consecutive same-role messages so the Anthropic API receives alternating turns.
+    // A mixed text+tool_use response is stored as separate TextMessage / ActionExecutionMessage
+    // objects; they must be merged into one assistant message before dispatch.
+    const coalescedMessages =
+      coalesceConsecutiveSameRoleMessages(anthropicMessages);
+
     // Apply token limits
     const limitedMessages = limitMessagesToTokenCount(
-      anthropicMessages,
+      coalescedMessages,
       tools,
       model,
       this.maxInputTokens,
@@ -345,7 +352,7 @@ export class AnthropicAdapter implements CopilotServiceAdapter {
     );
 
     // We'll check if we need a fallback response after seeing what Anthropic returns
-    // We skip grouping by role since we've already ensured uniqueness of tool_results
+    // Role alternation is now enforced by coalesceConsecutiveSameRoleMessages above.
 
     let toolChoice: any = forwardedParameters?.toolChoice;
     if (forwardedParameters?.toolChoice === "function") {
--- a/packages/runtime/src/service-adapters/anthropic/utils.ts
+++ b/packages/runtime/src/service-adapters/anthropic/utils.ts
@@ -217,3 +217,35 @@ export function convertMessageToAnthropicMessage(
     };
   }
 }
+
+/**
+ * Anthropic's Messages API requires alternating user/assistant roles.
+ * CopilotKit stores a mixed text+tool_use response as separate TextMessage and
+ * ActionExecutionMessage objects, which map to two consecutive assistant-role entries.
+ * This function merges consecutive same-role messages by concatenating their content
+ * arrays so the payload satisfies Anthropic's role-alternation constraint.
+ */
+export function coalesceConsecutiveSameRoleMessages(
+  messages: Anthropic.Messages.MessageParam[],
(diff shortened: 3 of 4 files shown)


BenTaylorDev (approved):
**Approving.** Verified against current `main`: the Anthropic adapter still sends consecutive same-role messages (`anthropic-adapter.ts` still carries the "we skip grouping by role" note), which breaks Anthropic's alternating-role requirement — the reported #2910 case (assistant text preamble + assistant tool_use → two consecutive assistant messages). The fix coalesces correctly: `convertMessageToAnthropicMessage` always returns array `content`, so the merge guard holds; `tool_use` (assistant) and `tool_result` (user) are never merged across roles; and it's scoped to Anthropic only (`coalesceConsecutiveSameRoleMessages` is called nowhere but `anthropic-adapter.ts:334`), so no cross-provider regression. Tests assert the real outgoing payload.

BenTaylorDev (changes requested):
Following up on my approval — running the Anthropic adapter suite locally against current `main`, the two **new** same-role-coalescing tests fail, so the PR doesn't pass its own suite as merged. Converting to request-changes.

Both fail with the captured Anthropic `messages` coming back empty:
- `merges TextMessage(assistant) + ActionExecutionMessage into a single Anthropic assistant message` → `expected [] to have a length of 1`
- `does not merge alternating user/assistant messages (no regression)` → `expected [] to have a length of 3`

The 8 pre-existing anthropic tests pass, and the coalesce call is correctly wired (`anthropic-adapter.ts:334`), so this looks like the two new tests' setup/capture needs reconciling with the current `process()` flow rather than a problem with the coalescing logic itself. Could you rebase onto current `main` and confirm the two coalescing tests go green? The fix's approach is sound (Anthropic-scoped, array-content guard holds) — it just needs its own tests passing against current main before merge. Sorry for the back-and-forth.

rodboev:
Thanks for catching this. The rebase exposed test-fixture drift from current `main`, while the Anthropic coalescing path itself remained sound.

1. In `packages/runtime/tests/service-adapters/anthropic/anthropic-adapter.test.ts`, I updated both same-role test fixtures to use the current object-form `TextMessage` constructor with `Role`, and removed the obsolete `parameters` field from the action fixture.
2. I left the production call in `packages/runtime/src/service-adapters/anthropic/anthropic-adapter.ts` and the array-guarded merge in `packages/runtime/src/service-adapters/anthropic/utils.ts` unchanged.
3. The rebased Anthropic adapter test file now passes all 10 tests, including the text plus `tool_use` merge and alternating-role non-regression cases.

BenTaylorDev (approved):
**Re-approving** — the two coalescing tests pass now. Verified on the updated branch against current `main`: Anthropic adapter suite 10/10, lint 0 errors, no type errors in the changed files, and CI is green. The coalescing is correct and Anthropic-scoped; the fixtures now align with the current message types.

Issue #4739, opened 2026-05-08
🐛 Bug: useAgent fails to save messages under the specified thread ID


7codo:
### ♻️ Reproduction Steps

- I use CopilotSidebar, and it works correctly — the messages are saved under the thread ID that I set.
- However, when I use useAgent, the messages are not saved under the thread ID that I set.
- I am using Next.js, FastAPI, and PostgreSQL for the database."

page.tsx
```typescript
'use client'
import { CopilotSidebar } from "@copilotkit/react-core/v2";
import { useState, useCallback } from "react";
import { useAgent } from "@copilotkit/react-core/v2";
import { useCopilotKit } from "@copilotkit/react-core/v2";


export default function Page() {
  const { agent } = useAgent({ agentId: "sample_agent" });
  const { copilotkit } = useCopilotKit();
  const [input, setInput] = useState("");

  const sendMessage = useCallback(async () => {
    agent.addMessage({
      id: crypto.randomUUID(),
      role: "user",
      content: input,
    });
    setInput("");
    await copilotkit.runAgent({ agent });
  }, [input, agent, copilotkit]);

  return (
    <main>
      <h1>Your App</h1>
      <div className="flex flex-col h-full">
        <div className="flex-1 overflow-y-auto p-4 space-y-4">
          {agent.messages.map((msg) => (
            <div
              key={msg.id}
              className={msg.role === "user" ? "ml-auto bg-blue-100 rounded-lg p-3 max-w-md" : "bg-gray-100 rounded-lg p-3 max-w-md"}
            >
              <p className="text-sm font-medium">{msg.role}</p>
              <p>{msg.content}</p>
            </div>
          ))}
          {agent.isRunning && <div className="text-gray-400">Thinking...</div>}
        </div>
      </div>
      <form
        className="border-t p-4 flex gap-2"
        onSubmit={(e) => { e.preventDefault(); sendMessage(); }}
      >
        <input
          value={input}
          onChange={(e) => setInput(e.target.value)}
          placeholder="Type a message..."
          className="flex-1 border rounded-lg px-3 py-2"
        />
        <button type="submit" disabled={agent.isRunning}>Send</button>
      </form>
    </main>
  );
}
```

layout.tsx
```typescript
import { CopilotKit } from "@copilotkit/react-core";
import "@copilotkit/react-ui/v2/styles.css";
import './globals.css';

// ...

export default function RootLayout({ children }: { children: React.ReactNode }) {
  return (
[...]

PR #5361, opened 2026-06-10
feat(react-core): accept threadId in useAgent options

rodboev:
## Summary

Adds optional `threadId` support to the `useAgent` hook, enabling headless callers to associate an agent with a specific thread without manually patching `agent.threadId` after each render. When a caller provides a `threadId` prop to `useAgent`, the agent's thread is pinned to that value, routing all runs and message persistence to the specified thread.

## Root cause

`useAgent({ agentId })` in headless mode (without a wrapping `CopilotChat`/`CopilotSidebar` component) ignores any desired thread ID. The agent's `AbstractAgent.threadId` is auto-minted as a random UUID at construction and never synced to a caller-supplied value. Users must manually set `agent.threadId = desiredId` before each `copilotkit.runAgent()` call — a pattern the docs do not describe and UI components handle automatically via `CopilotChatConfigurationProvider`.

### 1. Add `threadId` prop to `UseAgentProps` (packages/react-core/src/v2/hooks/use-agent.tsx)

- Added optional `threadId?: string` field to the props interface with JSDoc explaining headless usage.
- Updated the function signature to destructure `threadId`.

### 2. Apply caller-supplied threadId in a new useEffect

- Inserted a new effect after the headers-sync effect that sets `agent.threadId = threadId` when the prop is provided.
- Placed before the existing chat-config sync so the provider value can override if both are present (design intent: explicit provider config wins).

### 3. Add test coverage (packages/react-core/src/v2/hooks/__tests__/use-agent.e2e.test.tsx)

New test suite "threadId prop" with two tests:
- **"sets agent.threadId from the prop on mount"**: Verifies that when `useAgent({ threadId: "thread-abc" })` is called, `agent.threadId` equals the supplied value after render.
- **"updates agent.threadId when the prop changes"**: Verifies that when the threadId prop is updated via state change, `agent.threadId` reflects the new value.

Tests use a custom render setup without a config-provided threadId to isolate the prop's behavior.

### 4. Write changeset (`.changeset/use-agent-thread-id-option.md`)

Changesets entry for `@copilotkit/react-core` as a minor version bump (new optional public API).

## What is NOT changed

- The existing chat-config threadId sync (lines 234–240) is unchanged.
- When both a prop threadId and a `CopilotChatConfigurationProvider`-sourced thread ID are present, the provider value wins (it runs last in effect order).
- `CopilotChat`, `CopilotSidebar`, `CopilotChatConfigurationProvider`, and `ThreadsProvider` are untouched.
- No behavior changes to existing components or v1 bridges.
- Python SDK and LangGraph parity are not in scope.

## Test plan

- [x] **use-agent.e2e.test.tsx**: 4/4 tests pass. Tests verify that the `threadId` prop is applied to the agent on mount and updated when the prop changes (no `runAgent` calls needed; direct inspection of `agent.threadId` is sufficient).
- [x] **use-agent-stability.test.tsx**: 4/4 tests pass. Existing stability tests remain unaffected.
- [x] Code formatted with oxfmt and linted with oxlint (zero errors, zero warnings in modified test file).

## Notes

- The implementation mirrors the existing chat-config threadId sync pattern but at the hook's prop level, providing the headless equivalent.
- Effect order ensures provider-level config takes precedence when both paths are active (e.g., nested `useAgent` inside a `CopilotChat` subtree).

### DIFF
--- a/.changeset/use-agent-thread-id-option.md
+++ b/.changeset/use-agent-thread-id-option.md
@@ -1,0 +1,8 @@
+---
+"@copilotkit/react-core": minor
+---
+
+feat(react-core): accept threadId in useAgent options
+
+Headless callers can now pin the agent to a specific thread without
+manually setting agent.threadId after each render.
--- a/packages/react-core/src/v2/hooks/__tests__/use-agent.e2e.test.tsx
+++ b/packages/react-core/src/v2/hooks/__tests__/use-agent.e2e.test.tsx
@@ -1,8 +1,14 @@
-import React from "react";
-import { screen, fireEvent, waitFor } from "@testing-library/react";
+import React, { useState } from "react";
+import {
+  screen,
+  fireEvent,
+  waitFor,
+  act,
+  render,
+} from "@testing-library/react";
 import { describe, it, expect } from "vitest";
-import { type BaseEvent, type RunAgentInput } from "@ag-ui/client";
-import { Observable } from "rxjs";
+import type { BaseEvent, RunAgentInput } from "@ag-ui/client";
+import type { Observable } from "rxjs";
 import {
   MockStepwiseAgent,
   renderWithCopilotKit,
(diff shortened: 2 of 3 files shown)


rodboev:
Closing this because current main already contains the fix for the reported #4739 path via #5042. I missed that prior merged fix when opening this PR.

This branch adds a new `threadId` option to `useAgent`, but the issue evidence is about `<CopilotKit threadId=...>` not propagating into headless `useAgent`, and that provider-sync path is already handled on main. Keeping this open would ask maintainers to review a new API surface without a still-live bug target, so I’m withdrawing it.

PR #5365, opened 2026-06-10
fix: repair check-types across all packages and gate it in CI

onsclom:
## What this does

Follow-up to #5356. CI never ran `check-types`, so type errors piled up silently across the monorepo. #5356 fixed shared, a2ui-renderer, and angular. This PR fixes every remaining package and adds a CI gate so it cannot happen again.

Almost all of the diff is mechanical type repair: type annotations, `import type` splits, casts, ambient declarations, and tsconfig module-resolution bumps. The sections below call out the parts that are not purely mechanical so review can focus there.

## Where to focus your review

These are the only changes with runtime or public-API impact. Everything else is type-level and behavior-preserving.

1. **License gating wired to /info** (shared, react-core, vue, react-native). Behavioral change, details below.
2. **react-native bug fix**: a `catch` binding referenced by `TypeError.cause` had been lint-stripped and is restored, plus a `{ cause }` is now attached to a parse error.
3. **runtime `Schema` passthrough**: a type-only cap on `tool()` schema inference. Runtime behavior is unchanged; it only stops tsc from blowing past 8 GB.
4. **New public type export**: `Anchor` from web-inspector, consumed by react-core's `inspectorDefaultAnchor`.
5. **New CI job**, see the CI section.

## License context wired to /info

Follow-up to @MikeRyanDev's review on #5356 (wire the inert license machinery to /info). `createLicenseContextValue` now takes the server-reported license status instead of a hardwired null. `checkFeature()` returns false only when the runtime reports `expired` or `invalid`, and fails open otherwise (null, none, expiring, and valid all keep features on). The React, Vue, and React Native providers feed the status they already track from /info. New tests cover the gating in shared and react-core. Per-feature data is not in /info yet, so gating is uniform across features.

## Type fixes by package

- **core** (392 errors): bundler module resolution (matches the tsdown build and the vue package), strict-mode and AG-UI drift in sources and tests, added `@types/phoenix`.
- **react-core** (255): component, hook, and test drift (now-private `activeRunCompletionPromise`, RunFinished outcome union, slot prop types, StandardSchema variance), ambient declarations for `katex` CSS and the react-markdown JSX namespace, plus the `Anchor` export.
- **runtime** (393, previously hidden behind an OOM crash): BuiltInAgentConfiguration narrowing, AI SDK v6 and AG-UI drift, the UserMessage attachment migration, OpenAI v4/v5 surface changes, and the `Schema` passthrough noted above.
- **react-native** (23): provider/core bridge types, RN 0.85 drift, and the `catch`/`cause` bug fix noted above.
- **smaller fixes**: runtime-client-gql (UserMessage image narrowing, pinned `types`), web-inspector (nodenext import extensions, RunFinished narrowing), react-textarea (es2023 lib for `toReversed`), react-ui, voice, agentcore-runner, sqlite-runner (one-liners), sdk-js (bundler resolution plus the missing `@standard-schema/spec` dev dep), examples/v2/node (drop stale node10 resolution overrides).

## CI

New `check-types` job in `static_quality.yml`: run graphql codegen, then `nx run-many -t check-types --parallel=1` with a 12 GB heap. The runtime typecheck alone peaks near 10 GB and takes about 5 minutes on a 16 GB runner, so it runs serially.

## Testing

- `nx run-many -t check-types` passes for all 24 projects locally and in CI.
- All package test suites pass: core 431, react-core 1271, runtime 1523, vue 1001, react-native 246, runtime-client-gql 129, plus the smaller packages.
- Lockfile diff is only the two added type-only dev dependencies (`@types/phoenix`, `@standard-schema/spec`).

- react-core's built `.d.mts` files emit extensionless relative imports, which silently degrade consumer types to `any` under `skipLibCheck` (found while fixing react-native).
- The AI SDK x zod type-instantiation cost in runtime deserves a real fix (40M instantiations); trace data available.
- `@ai-sdk/anthropic`'s new `authToken` setting is not forwarded by the Anthropic adapter.
- `MCPClientProvider.tools()` has diverged from AI SDK v6's `ToolSet` typing.


MikeRyanDev (dismissed):
Blocking on the new typecheck gate: this PR still does not typecheck.

`pnpm nx run @copilotkit/core:check-types` fails at the current head (`783df784`) with:

```text
packages/core/src/__tests__/core-a2ui-scoping.test.ts(27,14): error TS2790: The operand of a 'delete' operator must be optional.
```

The test file is unchanged from `main`; the failure is exposed by this PR's `packages/core/tsconfig.json` change that pins `"types": ["node"]`. GitHub's new `check-types` job is also red, so the added CI gate is currently failing on the PR it is introduced in.

There is also a separate red `Validate Showcase` check: `__tests__/generate-registry.test.ts` fails with `data drift not restored: .../showcase/shell/src/data/catalog.json`. The isolated failing file passed locally for me, so that may be suite-level flake/race rather than this PR's touched surface, but the PR is still blocked by the failed check.

Validation I ran:
- `git diff --check origin/main...HEAD` passed
- `pnpm install --frozen-lockfile` passed
- `pnpm nx run @copilotkit/core:check-types` failed as above
- `pnpm nx run @copilotkit/shared:test -- --run src/__tests__/license-context.test.ts` passed, 7/7
- `pnpm nx run @copilotkit/showcase-scripts:test -- --run __tests__/generate-registry.test.ts` passed locally, 5/5

onsclom:
Fixed the `check-types` failure. Thanks for catching it @MikeRyanDev.

Root cause was exactly as you diagnosed: pinning `"types": ["node"]` in `packages/core/tsconfig.json` made `fetch` a required member of `globalThis`. The test's cleanup cast intersected with `typeof globalThis`, so the intersection kept `fetch` non-optional and `delete` tripped TS2790:

```ts
// before
delete (global as typeof globalThis & { fetch?: typeof fetch }).fetch;
// after — drop the globalThis intersection so the operand is genuinely optional
delete (global as unknown as { fetch?: typeof fetch }).fetch;
```

This mirrors the `window` cleanup pattern already used two lines below. Folded the one-line fix into the `fix(core): repair check-types …` commit (via autosquash) so the tsconfig pin and the test repair it necessitates stay in the same commit — history is otherwise unchanged.

Verified:
- `nx run @copilotkit/core:check-types` ✓
- `nx run-many -t check-types --parallel=1` ✓ all 24 projects (the gate this PR introduces)
- `@copilotkit/core:test` on the touched file ✓ 4/4

The separate `Validate Showcase` (`generate-registry.test.ts` data-drift) red you flagged as likely suite-level race is untouched here — happy to look if it persists on the new run.

tylerslaton (approved):

PR #5372, opened 2026-06-10
fix(sdk-python): emit tool call events for middleware-intercepted SDK Actions

### LINKED ISSUE #4342: CopilotKitMiddleware does not execute SDK Action tool calls when tools=[] is passed to langchain.agents.create_agent (copilotkit 0.1.87)
## Summary

When using `langchain.agents.create_agent` with `CopilotKitMiddleware` and `tools=[]`, SDK Actions registered via `CopilotKitRemoteEndpoint` are surfaced to the model and invoked by the LLM (the assistant message contains a valid `tool_calls=[...]`), but the runtime never executes the action handler. No AG-UI `TOOL_CALL_START`/`TOOL_CALL_ARGS`/`TOOL_CALL_END` events are emitted, no `role=tool` message appears in the final `MESSAGES_SNAPSHOT`, and the conversation terminates after the LLM's preamble.

The same SDK Actions execute correctly when driven by a custom `chat_node` that does explicit `bind_tools` + a tool-execution loop (and lets the middleware inject Actions into `state["copilotkit"]["actions"]` for discoverability only).

## Versions

- `copilotkit==0.1.87` (latest on PyPI as of 2026-04-28)
- `langchain==1.2.10`
- `langchain-core==1.3.2`
- `langchain-anthropic==1.4.1`
- `ag-ui-protocol==0.1.18`
- Python 3.11; FastAPI runtime; Anthropic Claude as the model.

## Reproduction

```python
from langchain.agents import create_agent
from langchain.agents.middleware.types import AgentState
from langgraph.checkpoint.memory import MemorySaver
from copilotkit.langchain import CopilotKitMiddleware
from copilotkit import CopilotKitRemoteEndpoint, Action

# 1. Define an SDK Action
async def search_tenants_handler(query: str) -> dict:
    return {"type": "tenantSearch", "totalCount": 5, "groups": [...]}

action = Action(
    name="searchTenants",
    description="Search the landlord's tenants",
    parameters=[{"name": "query", "type": "string", "required": True}],
    handler=search_tenants_handler,
)

# 2. Build the agent with tools=[] and middleware
agent = create_agent(
    model=...,
    tools=[],
    middleware=[CopilotKitMiddleware()],
    state_schema=AgentState,
    checkpointer=MemorySaver(),
)

# 3. Register the agent + action via CopilotKitRemoteEndpoint
endpoint = CopilotKitRemoteEndpoint(
    agents=[CopilotKitLangGraphAgent(name="default", graph=agent)],
    actions=[action],
)

# 4. Send a user prompt that should invoke `searchTenants`
# (e.g. "who are my tenants?")
```

## Expected behavior

1. Middleware injects the SDK Action(s) into the model call.
2. LLM emits a `tool_call` for `searchTenants`.
3. Runtime invokes `search_tenants_handler`.
4. AG-UI lifecycle emits `TOOL_CALL_START` → `TOOL_CALL_ARGS` → `TOOL_CALL_END`.
5. A `role=tool` message containing the handler return value is appended to state.
6. LLM is re-invoked, consumes the tool result, and emits a final natural-language answer.

## Actual behavior

The LLM emits a valid `tool_call` for `searchTenants` in its assistant message, but the runtime never executes the handler:

- AG-UI `TOOL_CALL_*` event count: **0/0/0** (vs **1/2/1** on a custom-`chat_node` baseline path that exercises the same Action).
- Final `MESSAGES_SNAPSHOT` is missing the `role=tool` message entirely.
- Final assistant message contains `tool_calls=[{name: "searchTenants", ...}]` but no second LLM turn occurs.
- Conversation ends after the LLM's preamble (e.g. "I'll search for all your tenants.").

### Concrete event evidence

For an identical user prompt ("who are my tenants?"), comparing the middleware path against a custom `chat_node` (explicit `bind_tools` + tool-execution loop) on the same model and same SDK Actions:

| Event type | Custom `chat_node` | `create_agent` + `CopilotKitMiddleware` |
|---|---:|---:|
| `RUN_STARTED` | 1 | 1 |
| `STEP_STARTED` | 1 | 4 |
| `TOOL_CALL_START` | **1** | **0** |
| `TOOL_CALL_ARGS` | **2** | **0** |
| `TOOL_CALL_END` | **1** | **0** |
| `TEXT_MESSAGE_START` / `_CONTENT` / `_END` | 1 / 2 / 1 | 1 / 2 / 1 |
| `STATE_SNAPSHOT` | 2 | 4 |
| `STEP_FINISHED` | 1 | 4 |
| `MESSAGES_SNAPSHOT` | 1 | 1 |
| `RUN_FINISHED` | 1 | 1 |
| **Total non-RAW events** | **15** | **19** |

Final `MESSAGES_SNAPSHOT` on the middleware path contains exactly 3 messages — `role=user`, `role=system` (App Context), `role=assistant` with `tool_calls=[searchTenants(...)]` — and no `role=tool` follow-up. The custom `chat_node` baseline produces 4 messages including the `role=tool` JSON payload that the frontend `render` handler needs.

## Workarounds tried

- **`tools=[my_tool]` directly** (without middleware tool injection): collides at request time with `anthropic.BadRequestError: 'tools: Tool names must be unique.'` because `CopilotKitMiddleware` re-injects the same SDK Action by name on every model call.
- **`tools=[my_tool]` + omitting middleware**: works for tool execution, but loses CopilotKit state propagation (e.g. `useCopilotReadable`, frontend-managed actions).
- **`system_prompt=` on `create_agent`**: collides with middleware-injected state messages — `ValueError: Received multiple non-consecutive system messages.`

The current production workaround we use is a custom `chat_node` with explicit `bind_tools` + a tool-execution loop, with `CopilotKitMiddleware` left to inject Actions into `state["copilotkit"]["actions"]` for discoverability only. This works but bypasses the apparent intent of the middleware.

## Question for maintainers

Is `CopilotKitMiddleware.awrap_model_call` supposed to drive Action handler execution (i.e., is the middleware itself supposed to be the tool-execution loop when `tools=[]` is passed to `create_agent`), or is it only supposed to annotate / inject Actions onto the model call, with execution remaining the responsibility of an explicit tool-execution loop in the agent graph?

If the latter, what is the recommended tool-execution shape when integrating SDK Actions with `langchain.agents.create_agent`? The current public examples appear to assume `create_agent` will handle execution, but with `tools=[]` no LangChain tool node is wired up to consume the LLM's `tool_calls`.

If the former, this looks like a bug at copilotkit 0.1.87 — the `tool_calls` emitted by the LLM are not picked up and executed by the middleware path.

Happy to share additional event traces, runtime logs, or a minimal reproducible repository if helpful.

rodboev:
## Summary

Frontend SDK Actions intercepted by `CopilotKitMiddleware` could disappear before the client received a tool-call lifecycle, leaving the handler uncalled and the conversation without a tool result or follow-up answer. The bridge now consumes the middleware's authoritative intercepted-call state and publishes each call through the AG-UI adapter at most once. The focused intercepted file has 8 tests and the manual AG-UI file has 30 tests; the combined run is 38 tests, with the supported-version matrix remaining CI-owned.

## Root cause

`CopilotKitMiddleware.after_model()` removes frontend calls before LangChain's backend tool node and records those exact calls for final-message restoration. The previous adapter fallback ignored that state and reconstructed frontend ownership from `on_chat_model_end` plus the run's action catalog. That duplicated classification, failed when an action carried `function: None`, and never ran on Python 3.10 because the async LangChain path emitted no model callbacks there.

## Changes

- `sdk-python/copilotkit/langgraph_agui_agent.py`
  - publish middleware-classified calls from the existing chain-state event path
  - reuse one lifecycle materializer and the parent adapter's per-run ID ledger
  - remove raw model-end/action-schema reconstruction and the duplicate logger assignment
- `sdk-python/tests/test_intercepted_tool_call_events.py`
  - exercise the real `create_agent(..., tools=[])` middleware and AG-UI path in streaming and non-streaming modes
  - cover exactly-once identity, Python 3.10 restoration, multiple IDs, metadata opt-out, backend negative space, and malformed action/state shapes

## What is not changed

- backend tools still use LangChain's normal tool node
- `after_agent` still restores intercepted calls into the final assistant snapshot
- tool-call emission metadata still controls streamed lifecycle visibility
- explicit `copilotkit_emit_tool_call()` calls keep their existing caller-controlled semantics

## Related PRs and Issues

Closes #4342.

## Test plan

- [x] Focused intercepted SDK Action regression file, 8 passed on Python 3.12. Covers the real `create_agent(..., tools=[])` path, streaming, restoration, mixed ownership, chain metadata suppression, malformed state, and action-shape isolation.
- [x] Existing manual tool-call materialization file, 30 passed on Python 3.12.
- [x] Combined focused selection, 38 passed on Python 3.12, counted once rather than adding a second 38-test total.
- [x] Ruff check and format check on both changed files, both passed.
- [ ] Python SDK CI matrix green on Python 3.10, 3.11, 3.12, 3.13, and 3.14

### DIFF
--- a/sdk-python/copilotkit/copilotkit_lg_middleware.py
+++ b/sdk-python/copilotkit/copilotkit_lg_middleware.py
@@ -946,30 +946,8 @@ class CopilotKitMiddleware(AgentMiddleware[StateSchema, Any]):
         state: StateSchema,
         runtime: Runtime[Any],
     ) -> dict[str, Any] | None:
-        # Get the base interception result from the sync implementation
-        result = self.after_model(state, runtime)
-        if result is None:
-            return None
-
-        # Emit custom events for intercepted tool calls so the AG-UI event stream
-        # can pick them up and emit corresponding TOOL_CALL_* events to the frontend.
-        intercepted = (result.get("copilotkit") or {}).get("intercepted_tool_calls", [])
-        if intercepted:
-            from langchain_core.callbacks.manager import adispatch_custom_event
-            from langgraph.config import get_config
-
-            config = get_config()
-            for call in intercepted:
-                await adispatch_custom_event(
-                    "copilotkit_manually_emit_tool_call",
-                    {
-                        "name": call.get("name", ""),
-                        "args": call.get("args", {}),
-                        "id": call.get("id", ""),
-                    },
-                    config=config,
-                )
-        return result
+        # Delegate to sync implementation
+        return self.after_model(state, runtime)
 
     # Restore frontend tool calls to AIMessage before agent exits
     def after_agent(
--- a/sdk-python/copilotkit/copilotkit_lg_middleware.py
+++ b/sdk-python/copilotkit/copilotkit_lg_middleware.py
@@ -946,8 +946,30 @@ class CopilotKitMiddleware(AgentMiddleware[StateSchema, Any]):
         state: StateSchema,
         runtime: Runtime[Any],
     ) -> dict[str, Any] | None:
-        # Delegate to sync implementation
-        return self.after_model(state, runtime)
+        # Get the base interception result from the sync implementation
+        result = self.after_model(state, runtime)
+        if result is None:
+            return None
+
+        # Emit custom events for intercepted tool calls so the AG-UI event stream
+        # can pick them up and emit corresponding TOOL_CALL_* events to the frontend.
+        intercepted = (result.get("copilotkit") or {}).get("intercepted_tool_calls", [])
+        if intercepted:
+            from langchain_core.callbacks.manager import adispatch_custom_event
+            from langgraph.config import get_config
+
+            config = get_config()
+            for call in intercepted:
+                await adispatch_custom_event(
+                    "copilotkit_manually_emit_tool_call",
+                    {
+                        "name": call.get("name", ""),
+                        "args": call.get("args", {}),
+                        "id": call.get("id", ""),
+                    },
+                    config=config,
+                )
+        return result
 
     # Restore frontend tool calls to AIMessage before agent exits
     def after_agent(
--- a/sdk-python/copilotkit/langgraph_agui_agent.py
+++ b/sdk-python/copilotkit/langgraph_agui_agent.py
@@ -44,6 +44,7 @@ class LangGraphEventTypes(Enum):
     """LangGraph event types"""
 
     OnChatModelStream = "on_chat_model_stream"
+    OnChatModelEnd = "on_chat_model_end"
     OnCustomEvent = "on_custom_event"
 
 
@@ -283,7 +284,9 @@ class LangGraphAGUIAgent(LangGraphAgent):
             for emitted_event in self._intercepted_frontend_tool_call_events(event):
                 yield self._dispatch_event(emitted_event)
 
-    def _intercepted_frontend_tool_call_events(self, event: Any) -> list[ToolCallEvents]:
+    def _intercepted_frontend_tool_call_events(
+        self, event: Any
+    ) -> list[ToolCallEvents]:
         frontend_tool_names = self._frontend_tool_names()
         if not frontend_tool_names:
             return []
--- a/sdk-python/copilotkit/langgraph_agui_agent.py
+++ b/sdk-python/copilotkit/langgraph_agui_agent.py
@@ -29,8 +29,6 @@ except ImportError:
     # Langchain >= 1.0.0
     from langchain_core.messages import BaseMessage
 
-logger = logging.getLogger(__name__)
-
 
 class CustomEventNames(Enum):
     """Custom event names for CopilotKit"""
@@ -44,7 +42,6 @@ class LangGraphEventTypes(Enum):
     """LangGraph event types"""
 
     OnChatModelStream = "on_chat_model_stream"
-    OnChatModelEnd = "on_chat_model_end"
     OnCustomEvent = "on_custom_event"
 
 
@@ -115,86 +112,11 @@ class LangGraphAGUIAgent(LangGraphAgent):
                 return super()._dispatch_event(event)
 
             if custom_event.name == CustomEventNames.ManuallyEmitToolCall.value:
-                value = custom_event.value
-                if not isinstance(value, dict):
-                    raise CopilotKitMisuseError(
-                        f"ManuallyEmitToolCall event 'value' must be a dict, got {type(value).__name__}"
-                    )
-
-                tool_call_id = value.get("id")
-                tool_call_name = value.get("name")
-                tool_call_args = value.get("args")
-
-                if not isinstance(tool_call_id, str) or not tool_call_id.strip():
-                    raise CopilotKitMisuseError(
-                        f"ManuallyEmitToolCall event missing valid 'id': got {type(tool_call_id).__name__}"
-                    )
-                if not isinstance(tool_call_name, str) or not tool_call_name.strip():
-                    raise CopilotKitMisuseError(
-                        f"ManuallyEmitToolCall event missing valid 'name': got {type(tool_call_name).__name__}"
-                    )
-                if tool_call_args is None:
-                    raise CopilotKitMisuseError(
-                        f"ManuallyEmitToolCall event missing 'args' for tool_call_id={tool_call_id}"
-                    )
-
-                try:
-                    delta = (
-                        tool_call_args
-                        if isinstance(tool_call_args, str)
-                        else json.dumps(tool_call_args)
-                    )
-                except (TypeError, ValueError) as e:
-                    raise CopilotKitMisuseError(
-                        f"ManuallyEmitToolCall 'args' is not JSON-serializable for tool_call_id={tool_call_id}: {e}"
-                    ) from e
-
-                dispatched_start = False
-                end_dispatched = False
-                try:
-                    super()._dispatch_event(
-                        ToolCallStartEvent(
-                            type=EventType.TOOL_CALL_START,
-                            tool_call_id=tool_call_id,
-                            tool_call_name=tool_call_name,
-                            parent_message_id=tool_call_id,
-                            raw_event=event,
-                        )
-                    )
-                    dispatched_start = True
-                    super()._dispatch_event(
-                        ToolCallArgsEvent(
-                            type=EventType.TOOL_CALL_ARGS,
-                            tool_call_id=tool_call_id,
-                            delta=delta,
-                            raw_event=event,
-                        )
-                    )
-                    super()._dispatch_event(
-                        ToolCallEndEvent(
-                            type=EventType.TOOL_CALL_END,
-                            tool_call_id=tool_call_id,
-                            raw_event=event,
-                        )
-                    )
-                    end_dispatched = True
-                except Exception:
-                    if dispatched_start and not end_dispatched:
-                        try:
-                            super()._dispatch_event(
-                                ToolCallEndEvent(
-                                    type=EventType.TOOL_CALL_END,
-                                    tool_call_id=tool_call_id,
-                                    raw_event=event,
-                                )
-                            )
-                        except Exception:
-                            logger.error(
-                                "Failed to emit compensating TOOL_CALL_END for %s",
-                                tool_call_id,
-                                exc_info=True,
-                            )
-                    raise
+                self._materialize_tool_call_events(
+                    custom_event.value,
+                    event,
+                    parent_message_id=None,
+                )
                 return super()._dispatch_event(event)
 
             if custom_event.name == CustomEventNames.ManuallyEmitState.value:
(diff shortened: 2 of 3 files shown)


BenTaylorDev (changes requested):
Thanks for this. Before it can land I'd like to resolve a likely double-emission and get the mechanism actually exercised:

- **Double emission.** The PR emits `copilotkit_manually_emit_tool_call` for each intercepted call in `aafter_model`, but `after_agent` still *restores* those same tool calls back onto the AIMessage (`copilotkit_lg_middleware.py:950`, `tool_calls=[*existing_tool_calls, *intercepted_tool_calls]`). Nothing here demonstrates that `ag_ui_langgraph` doesn't *also* emit those calls — from model-stream tokens or the final messages snapshot of the restored AIMessage. If it does, the frontend receives the same `tool_call_id` twice → double execution and/or broken `tool_call`↔`ToolMessage` pairing. Please confirm exactly-once emission end-to-end.
- **Tests don't hit the real path.** `test_intercepted_tool_call_events.py` patches `langgraph.config.get_config` and `adispatch_custom_event` and calls `aafter_model` on a hand-built dict — there's no `create_agent`, no `LangGraphAGUIAgent`, no ag_ui streaming. It proves "aafter_model calls adispatch with these args," not that the events reach the frontend as a single TOOL_CALL_* triple on the real `create_agent(tools=[])` path.
- **Repro on current `main` unconfirmed.** #4342 was filed against copilotkit 0.1.87 (v1); `main` is a different `create_agent`-middleware architecture. Worth confirming the bug still reproduces there before investing further.

The fix is plausibly right, but the load-bearing "exactly one emission" property is demonstrated nowhere — please add a real end-to-end run (capture the AG-UI event stream for an intercepted SDK Action) and confirm no duplicate.

rodboev:
Thanks, I tightened the proof rather than change the mechanism.

1. I reran the real `create_agent(..., tools=[])` + `LangGraphAGUIAgent.run` path with a deterministic fake model. Current `main` emits zero `TOOL_CALL_*` events for the intercepted SDK Action, this branch emits exactly one `TOOL_CALL_START` / `TOOL_CALL_ARGS` / `TOOL_CALL_END` triple.
2. `sdk-python/tests/test_intercepted_tool_call_events.py` now exercises that full middleware + AG-UI path instead of patching `aafter_model` and asserting on `adispatch_custom_event`.
3. The final `MessagesSnapshot` still contains the restored tool call from `after_agent`, and the run still emits only one triple for that `tool_call_id`, so the restore step is not introducing a duplicate stream emission.

BenTaylorDev (changes requested):
Much better test — it now drives the real `create_agent` + `CopilotKitMiddleware` + `LangGraphAGUIAgent.run()` path and counts at `_dispatch_event`. But it only proves exactly-once for a *non-streaming* model: the fake model implements only `_generate` (no `_astream`), so the streaming path never fires. For a streaming model (what OpenAI/Anthropic use in `create_agent`), the manual-emit handler (`ag_ui_langgraph agent.py:1315-1336`) emits START/ARGS/END **unconditionally** — it doesn't consult `streamed_tool_call_ids` the way the `OnToolEnd` paths guard with `already_streamed` — so the same `tool_call_id` gets emitted twice (double execution / broken pairing). Please add a streaming fake-model test (implement `_astream` yielding `tool_call_chunks`) and skip the manual emit for ids already in `streamed_tool_call_ids`.

rodboev:
Thanks, you're right that the previous harness only exercised `_generate`.

I've added an `_astream` fake to `test_intercepted_tool_call_events.py` that drives `tool_call_chunks` through the same `create_agent(..., tools=[])` and `LangGraphAGUIAgent.run()` path, guarded the CopilotKit manual handler in `langgraph_agui_agent.py` with `streamed_tool_call_ids`, and kept the non-streaming manual fallback plus final snapshot restoration covered in both modes.

BenTaylorDev (changes requested):
Almost there — the streaming coverage is exactly right on Python 3.11–3.14, but CI is red on **3.10**: the two `*_in_both_modes` tests fail during the `streaming=True` branch with

```
RuntimeError: Called get_config outside of a runnable context
  ("Python 3.11 or later required to use this in an async context")
```

That's a LangGraph limitation — its async streaming path calls `get_config()`, which only works in an async context on Python 3.11+. The non-streaming half (and the other 201 tests) pass fine on 3.10. Since LangGraph async streaming isn't available on 3.10 at all, the streamed double-emit this guard prevents can't occur there either, so gating the streaming case to 3.11+ costs no real coverage. Something like:

```python
import sys
STREAMING_MODES = [False] + ([True] if sys.version_info >= (3, 11) else [])

@pytest.mark.parametrize("streaming", STREAMING_MODES)
def test_intercepted_sdk_action_emits_single_tool_call_triple(streaming): ...
```

(or a `@pytest.mark.skipif(sys.version_info < (3, 11), ...)` on a split-out streaming test). Once 3.10 is green this is good to go.

rodboev:
Thanks. I took a rebuild route here instead of version-gating the 3.10 streaming assertions.

The problem turned out to be the branch's middleware transport, not the coverage shape. Streaming models already emitted the frontend SDK Action lifecycle from `tool_call_chunks`, and the middleware fallback then had to dedupe that second producer. On Python 3.10 the same transport also depended on `get_config()` inside `aafter_model()`, which is what made the shard red.

I've rebased onto current `main` and moved the non-streaming fallback into `LangGraphAGUIAgent` itself. The adapter now uses the final `on_chat_model_end` `AIMessage.tool_calls` payload for intercepted frontend SDK Actions, while the existing streamed path remains the owner for `_astream`. That removes the second producer entirely, so there is no middleware-side duplicate guard and no middleware-side `get_config()` lookup left in the production path.

Current diff is down to:

1. `sdk-python/copilotkit/langgraph_agui_agent.py`, adapter-owned non-streaming fallback on `on_chat_model_end`
2. `sdk-python/tests/test_intercepted_tool_call_events.py`, real end-to-end proof for both `_generate` and `_astream`

Focused rerun on the rebuilt head:

1. `cd sdk-python && pytest tests/test_intercepted_tool_call_events.py tests/test_agui_agent.py -q`, 32 passed
2. `cd sdk-python && ruff check copilotkit/langgraph_agui_agent.py tests/test_intercepted_tool_call_events.py tests/test_agui_agent.py`
3. `cd sdk-python && ruff format --check copilotkit/langgraph_agui_agent.py tests/test_intercepted_tool_call_events.py tests/test_agui_agent.py`

BenTaylorDev (changes requested):
The rebuild is the right architecture and the mechanism now holds up under scrutiny — but the 3.10 shard is still red, so one more pass.

## What's now solid

I ran the full `sdk-python` suite in a clean venv on every CI Python (identical resolved deps in each: langchain 1.3.15, langchain-core 1.5.4, langgraph 1.2.11, ag-ui-langgraph 0.0.42). 3.11/3.12/3.13/3.14 are all `212 passed, 11 skipped`, which CI has now confirmed independently.

I also confirmed the two things I was skeptical about:

- The `streaming` field on `BoundFakeToolModel` really does select the path — I traced `_generate` vs `_astream` per iteration, so `*_in_both_modes` now covers two genuinely distinct paths. That was my open concern from the last round and it's addressed.
- The exactly-once property isn't luck. `streamed_tool_call_ids` is recorded upstream at `ag_ui_langgraph/agent.py:1214` on every streamed tool-call start *regardless* of `should_emit_tool_calls`, and `already_streamed` there gates only the Start/Args/End triple (RESULT still emits). So reading that set from the adapter is consistent with its documented contract, and the dedupe holds even when tool-call emission is opted out via metadata. Probing confirms the streaming START comes from `on_chat_model_stream` with the new fallback emitting zero events, and the non-streaming START comes from `on_chat_model_end`.

And the fix is load-bearing: swapping `main`'s `langgraph_agui_agent.py` back in makes both new tests fail on 3.11.

## Blocking: 3.10

CI is red on `test (3.10)` while 3.11–3.14 pass:

```
FAILED tests/test_intercepted_tool_call_events.py::test_intercepted_sdk_action_emits_single_tool_call_triple_in_both_modes - assert 0 == 1
FAILED tests/test_intercepted_tool_call_events.py::test_after_agent_restores_tool_call_in_both_modes - assert 0 == 3
2 failed, 210 passed, 11 skipped
```

Not the old `get_config` `RuntimeError` — this is silent zero emission, and it fails on the first (`streaming=False`) iteration, so it isn't streaming-specific.

I isolated the cause with CopilotKit out of the picture entirely — a bare `create_agent(model=fake, tools=[])` graph driven by `astream_events(version="v2")`:

```
py3.10: chat_model events = []            # both streaming=False and streaming=True
py3.11: ['on_chat_model_start', 'on_chat_model_end']
```

On 3.10, `astream_events` emits no `on_chat_model_*` events at all — langchain-core needs 3.11+ for async callback context propagation (the same root cause as the earlier `"Python 3.11 or later required to use this in an async context"`). So `on_chat_model_end` never fires on 3.10, and no change at the adapter layer can make it fire. Worth noting the interception itself is fine there: the final `MessagesSnapshot` still restores `tc-1` with the expected name and args on 3.10; only the streamed lifecycle triple is absent.

This is what I was asking for with a version gate — not on the *fix* (the fix is right), on the *assertion*, because the AG-UI streaming lifecycle is simply unavailable on 3.10 for the whole `create_agent` path. Something like:

```python
pytestmark = pytest.mark.skipif(
    sys.version_info < (3, 11),
    reason="langchain astream_events emits no on_chat_model_* events on 3.10 "
           "(async callback context propagation requires 3.11+), so the AG-UI "
           "tool-call lifecycle cannot be observed there",
)
```

If you'd rather keep 3.10 coverage of the part that does work, gate only the lifecycle assertions and leave the interception/restoration assertion ungated.

## Non-blocking

1. `copilotkit/langgraph_agui_agent.py:32` adds a second `logger = logging.getLogger(__name__)`; `main` already defines it at line 24. Drop the new one.
2. `_frontend_tool_names()` does `tool.get("function", {}).get("name")`, which raises `AttributeError` on an entry carrying `function: None`. `_serialize_copilotkit_runtime_payload` never produces a `function` key, so either drop that branch or use `(tool.get("function") or {})`.
3. Worth a one-line comment in `_intercepted_frontend_tool_call_events` noting that exactly-once depends on `ag_ui_langgraph` recording streamed ids (`agent.py:1214`). It's the load-bearing cross-package assumption, and an upstream rename would silently reintroduce double emission.
4. Release-note thought, not a change request: anyone who worked around #4342 by calling `copilotkit_emit_tool_call` themselves will now get two triples, since the manual-emit branch in `_dispatch_event` doesn't consult `streamed_tool_call_ids`.

Fix the 3.10 gate and the duplicate `logger` and I think this lands.

rodboev:
Thanks for isolating the Python 3.10 callback boundary and checking both model modes.

The follow-up now:

1. Publishes the middleware's authoritative intercepted-call state through the adapter's chain-state path, so frontend ownership isn't reconstructed from the action catalog or `on_chat_model_end`.
2. Keeps streamed IDs as the per-run exactly-once ledger and applies the existing `copilotkit:emit-tool-calls` filtering to the intercepted fallback.
3. Preserves the public manual-emission path, removes the duplicate logger, and removes the obsolete action-catalog lookup that could fail on `function: None`.
4. Adds real-run assertions for yielded lifecycle events, restoration, metadata suppression, malformed state, and both streaming modes.

The focused intercepted tests pass 8/8, the manual AG-UI tests pass 30/30, and the combined selection passes 38 tests. The Python 3.10 through 3.14 matrix remains unchecked until CI completes.

BenTaylorDev (approved):
Re-reviewed at `d9667bc`. This addresses everything from the last round. Approving.

**3.10 is fixed.** Moving the bridge from `on_chat_model_end` to the `on_chain_end` state path was the right call — it no longer depends on chat-model callbacks langchain-core can't emit on 3.10. Full matrix using the workflow's exact commands (`poetry lock && poetry install --with dev && poetry run python -m pytest tests/`), one clean env per version — `218 passed, 11 skipped` on 3.10, 3.11, 3.12, 3.13 and 3.14. CI now agrees: all five shards green, 32/33 checks success (1 skipped). The duplicate `logger` and the `function: None` `_frontend_tool_names()` branch are gone with the old architecture.

**Mechanism verified.** Tool-lifecycle events on the real `create_agent` path, main vs this branch, on the locked deps:

```
py3.10  main  streaming=False []                 streaming=True []
py3.10  PR    streaming=False [START,ARGS,END]   streaming=True [START,ARGS,END]
py3.12  main  streaming=False []                 streaming=True [START,ARGS,END]
py3.12  PR    streaming=False [START,ARGS,END]   streaming=True [START,ARGS,END]
```

The final `MESSAGES_SNAPSHOT` carries the tool call in every one of those cells, so the gap was purely the lifecycle a frontend handler fires on — missing for non-streaming models everywhere, and for *all* models on 3.10. Now exactly one triple in every case.

Exactly-once rests on `ag_ui_langgraph` recording streamed IDs at `agent.py:1214` (0.0.42) — it adds regardless of `should_emit_tool_calls`, gates only the triple at `:1382`/`:1436`, and discards before `TOOL_CALL_RESULT` at `:1408`/`:1462`. Unchanged in 0.0.43. And since `test_intercepted_sdk_action_streaming_emits_once` drives the real path, an upstream rename of that key surfaces as a doubled triple rather than silently.

**Heads-up, not blocking: `ag-ui-langgraph 0.0.43` breaks 4 of these tests.** It reads `self.emit_raw_events` in `_dispatch_event` — an attribute that doesn't exist in 0.0.42 — and `_bridge_agent()` builds its agent with `object.__new__(LangGraphAGUIAgent)`, so the four `_run_state_event` tests die with `AttributeError`. CI is unaffected (`poetry lock` keeps the committed lock at 0.0.42), but it fires the moment anyone re-locks, and `object.__new__` on a third-party base will keep breaking on future upstream attributes. Constructing a real agent fixes it — verified 218/218 on all five:

```python
def _bridge_agent():
    agent = LangGraphAGUIAgent(
        name="bridge",
        graph=create_agent(
            model=BoundFakeToolModel(responses=[AIMessage(content="", id="ai-1")]),
            tools=[],
            middleware=[CopilotKitMiddleware()],
        ),
    )
    agent.active_run = {"streamed_tool_call_ids": {"streamed"}}
    return agent
```

**Other non-blocking notes**

1. `_materialize_tool_call_events` carries three orthogonal flags (`dispatch`, `dispatch_via_adapter`, `dispatched_events` out-param) and runs twice per call — once to build `valid_calls`, then again to emit. A `_validate` / `_emit_triple` split drops both the flags and the second pass.
2. `(getattr(self, "active_run", None) or {}).setdefault(...)` dedupes into a throwaway dict if `active_run` is ever missing — double emission rather than a loud failure. An early return reads safer.
3. `_bridge_agent()` still sets `_copilotkit_runtime_payload = {"actions": [{"function": None}]}`, and `test_bridge_ignores_runtime_action_catalog_shapes` is left over from the catalog-reconstruction design — the bridge never reads the catalog now.
4. Release note, not a change request: anyone who worked around #4342 by calling `copilotkit_emit_tool_call()` with the model's own tool_call_id will now get two triples, since the manual-emit branch doesn't consult `streamed_tool_call_ids`.

Note this needs a python SDK release to reach anyone hit by #4342 — the fix is in `copilotkit`, and `sdk-python/pyproject.toml` is untouched here (correctly, since version bumps land as their own release commit).

Nice work sticking with this through four rounds.

PR #5373, opened 2026-06-10
fix(sdk-python): bridge copilotkit context into LangGraph subgraphs

### LINKED ISSUE #3886: 🐛 Bug: CopilotKitMiddleware not work if create a lanchain agent as a node of langgraph
### ♻️ Reproduction Steps

Thanks for your great works!

I found that CopilotKitMiddleware not work if create a lanchain agent as a node of langgraph;  in CopilotKitMiddleware follow line always return empty:

frontend_tools = request.state.get("copilotkit", {}).get("actions", [])

attached the source code to reproduce the issue, thanks!

### ✅ Expected Behavior

CopilotKitMiddleware  can work with langgraph agent node

### ❌ Actual Behavior

 in CopilotKitMiddleware follow line always return empty:

frontend_tools = request.state.get("copilotkit", {}).get("actions", [])

### Screenshots

_No response_

### 𝌚 CopilotKit Version

```shell
all versions
```

### 📄 Logs (Optional)

```shell

```

[frontend.txt](https://github.com/user-attachments/files/26704252/frontend.txt)
[backend.txt](https://github.com/user-attachments/files/26704253/backend.txt)

rodboev:
## Summary

`CopilotKitMiddleware` loses frontend actions and app context when a middleware-wrapped agent runs as a LangGraph subgraph. Parent graphs propagate run context to child nodes, but `state["copilotkit"]` stays empty there, so the child misses both frontend-tool injection and the `App Context:` note.

## Root cause

`LangGraphAGUIAgent.langgraph_default_merge_state(...)` materializes CopilotKit data into state for top-level runs, but child LangGraph nodes inherit runtime context rather than arbitrary parent state. The missing piece was earlier in the write path: by the time the base agent calls `get_stream_kwargs(...)`, it has already replaced the original AG-UI request with merged graph state, so serializing from that argument drops the frontend tools and app context on the real subgraph path.

The fix moves ownership to the run entry: `LangGraphAGUIAgent.run(...)` now captures the frontend tools and app context from the real `RunAgentInput`, and `get_stream_kwargs(...)` forwards that captured payload into LangGraph using `context["copilotkit"]` when the graph supports runtime context, or `config["configurable"]["copilotkit"]` on older graph shapes. `CopilotKitMiddleware` still uses one shared lookup order everywhere it reads CopilotKit data: `state["copilotkit"]`, then `runtime.context["copilotkit"]`, then LangGraph config fallbacks. That same carrier now drives the app-context-note helper path too.

## Changes

- `sdk-python/copilotkit/langgraph_agui_agent.py`: capture frontend actions and app context from the original `RunAgentInput` in `run(...)`, then write the captured payload into the LangGraph runtime carrier that the current graph shape supports before the subgraph starts.
- `sdk-python/copilotkit/copilotkit_lg_middleware.py`: unify frontend-tool, app-context, and interception lookups around the same carrier order, and cover the current-`main` app-context helper path.
- `sdk-python/tests/test_agui_agent.py`: add focused coverage for both the runtime-context path and the configurable fallback path on graphs without `context` support.
- `sdk-python/tests/test_copilotkit_lg_middleware.py`: replace the manually seeded subgraph proof with a real `LangGraphAGUIAgent.run(...)` regression that reaches the child middleware through the production path.

## What is NOT changed

- Top-level agents still prefer `state["copilotkit"]`; the runtime bridge only fills the subgraph gap.
- Header propagation behavior is unchanged.
- No JS LangGraph executor changes are part of this PR.

## Related PRs and Issues

Closes #3886.

## Test plan

- [x] `cd sdk-python && pytest tests/test_agui_agent.py tests/test_copilotkit_lg_middleware.py -q` - 108/108 pass. Covers the AGUI run-entry bridge, the configurable fallback for graphs without `context` support, frontend-tool injection, app-context note injection, and a real `create_agent` subgraph invocation through `LangGraphAGUIAgent.run(...)`.
- [x] `ruff check sdk-python/copilotkit/copilotkit_lg_middleware.py sdk-python/copilotkit/langgraph_agui_agent.py sdk-python/tests/test_agui_agent.py sdk-python/tests/test_copilotkit_lg_middleware.py` - pass. Covers Python lint on every touched file.
- [x] `ruff format --check sdk-python/copilotkit/copilotkit_lg_middleware.py sdk-python/copilotkit/langgraph_agui_agent.py sdk-python/tests/test_agui_agent.py sdk-python/tests/test_copilotkit_lg_middleware.py` - pass. Covers repo formatting on every touched file.

### DIFF
--- a/sdk-python/copilotkit/copilotkit_lg_middleware.py
+++ b/sdk-python/copilotkit/copilotkit_lg_middleware.py
@@ -274,6 +274,25 @@ class CopilotKitMiddleware(AgentMiddleware[StateSchema, Any]):
     def name(self) -> str:
         return "CopilotKitMiddleware"
 
+    @staticmethod
+    def _get_copilotkit_context(state: dict) -> dict:
+        """Read copilotkit context from state, falling back to LangGraph config.
+
+        When the agent runs as a subgraph, the parent may not propagate the
+        copilotkit state key. LangGraph always propagates config.configurable
+        to subgraph nodes, so we fall back to reading from there.
+        """
+        ck = state.get("copilotkit") or {}
+        if ck.get("actions") or ck.get("context"):
+            return ck
+        try:
+            from langgraph.config import get_config
+
+            cfg = get_config() or {}
+            return (cfg.get("configurable") or {}).get("copilotkit") or ck
+        except Exception:  # noqa: BLE001 - no active context / older langgraph
+            return ck
+
     # ------------------------------------------------------------------
     # State-to-prompt surfacing
     # ------------------------------------------------------------------
@@ -539,7 +558,9 @@ class CopilotKitMiddleware(AgentMiddleware[StateSchema, Any]):
         request = self._apply_app_context_note(request)
 
         a2ui_tool = self._maybe_build_a2ui_tool(request)
-        frontend_tools = request.state.get("copilotkit", {}).get("actions", [])
+        frontend_tools = self._get_copilotkit_context(request.state or {}).get(
+            "actions", []
+        )
         if a2ui_tool is not None:
             # Our generate_a2ui replaces the runtime's render tool — don't
             # advertise both. Drop the render tool the A2UI middleware injected.
@@ -743,7 +764,9 @@ class CopilotKitMiddleware(AgentMiddleware[StateSchema, Any]):
         request = self._apply_app_context_note(request)
 
         a2ui_tool = self._maybe_build_a2ui_tool(request)
-        frontend_tools = request.state.get("copilotkit", {}).get("actions", [])
+        frontend_tools = self._get_copilotkit_context(request.state or {}).get(
+            "actions", []
+        )
         if a2ui_tool is not None:
             # Our generate_a2ui replaces the runtime's render tool — don't
             # advertise both. Drop the render tool the A2UI middleware injected.
@@ -820,7 +843,7 @@ class CopilotKitMiddleware(AgentMiddleware[StateSchema, Any]):
         state: StateSchema,
         runtime: Runtime[Any],
     ) -> dict[str, Any] | None:
-        frontend_tools = state.get("copilotkit", {}).get("actions", [])
+        frontend_tools = self._get_copilotkit_context(state).get("actions", [])
         if not frontend_tools:
             return None
 
--- a/sdk-python/copilotkit/copilotkit_lg_middleware.py
+++ b/sdk-python/copilotkit/copilotkit_lg_middleware.py
@@ -275,21 +275,54 @@ class CopilotKitMiddleware(AgentMiddleware[StateSchema, Any]):
         return "CopilotKitMiddleware"
 
     @staticmethod
-    def _get_copilotkit_context(state: dict) -> dict:
-        """Read copilotkit context from state, falling back to LangGraph config.
+    def _has_copilotkit_payload(candidate: Any) -> bool:
+        return isinstance(candidate, dict) and (
+            bool(candidate.get("actions")) or bool(candidate.get("context"))
+        )
+
+    @staticmethod
+    def _copilotkit_from_runtime_context(runtime_context: Any) -> dict[str, Any]:
+        if not isinstance(runtime_context, dict):
+            return {}
+        nested = runtime_context.get("copilotkit")
+        if CopilotKitMiddleware._has_copilotkit_payload(nested):
+            return nested
+        if CopilotKitMiddleware._has_copilotkit_payload(runtime_context):
+            return runtime_context
+        return {}
+
+    @staticmethod
+    def _get_copilotkit_context(
+        state: dict,
+        runtime_context: Any = None,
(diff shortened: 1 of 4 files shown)


BenTaylorDev (changes requested):
Thanks for tackling the subgraph case — #3886 is a real gap (a langchain agent used as a langgraph node has empty `state["copilotkit"]`). My concern is that the fix reads from a location that nothing populates, so it may be a no-op in production:

- The fallback reads `config["configurable"]["copilotkit"]`, but I can't find anywhere in the SDK that writes `copilotkit` into `config.configurable` — `langgraph_default_merge_state` puts it into **state**, and the invocation config never adds it. LangGraph does propagate `config.configurable` parent→subgraph, but the copilotkit dict lives in state, which isn't bridged into config. So at a subgraph node `get_config()["configurable"]["copilotkit"]` is empty unless the user manually threads it — the same manual step this is meant to remove. Both new tests `monkeypatch` `get_config` to return copilotkit, which mocks the exact source production doesn't fill.
- **Please verify with a real subgraph** (no mocks): reproduce #3886 with the PR applied and confirm `get_config()["configurable"]["copilotkit"]` actually contains the actions/context at the node. If it doesn't, the real fix is to *populate* `config.configurable.copilotkit` at the top-level invocation (or otherwise bridge state→subgraph), not just read it.
- **Completeness:** since this branch was opened, `main` added `_apply_app_context_note` / `_build_app_context_note` (reads `state["copilotkit"]["context"]`), which the PR's helper doesn't cover — so even if the mechanism worked, the app-context half of the issue would remain broken for subgraph nodes.

The state-precedence path is correctly preserved (no regression for normal agents) and the guard is safe — the blocker is proving the config fallback actually recovers the context at runtime.

rodboev:
Thanks, you were right that the config-only fallback was not enough.

1. `LangGraphAGUIAgent.get_stream_kwargs(...)` now writes a nested `context["copilotkit"]` payload before the run enters LangGraph, so child subgraphs receive the frontend actions and app context without manual threading.
2. `CopilotKitMiddleware` now reads the same carrier order everywhere it needs CopilotKit data: `state["copilotkit"]` first, then `runtime.context["copilotkit"]`, then LangGraph config fallbacks. That covers the newer app-context-note path too.
3. I replaced the mocked `get_config()` proof with a real subgraph test that builds a parent `StateGraph`, mounts `create_agent(..., middleware=[CopilotKitMiddleware()])` as a child node, invokes it with `context={"copilotkit": ...}`, and asserts that the child model sees the frontend tool plus the injected `App Context:` note.

BenTaylorDev (changes requested):
Good progress — the read side is now genuinely real (LangGraph does propagate run-scoped `runtime.context` into subgraph nodes, nicely proven), and the app-context site (`_build_app_context_note`) is covered too. But the write side doesn't populate it on the real path: the `get_stream_kwargs` override serializes from `input` via `getattr(input, "tools"/"context")`, yet the library calls `get_stream_kwargs(input=stream_input)` where `stream_input` is a merged **state dict** (not the AG-UI `RunAgentInput`) — so `getattr(dict, "tools")` is `None`, the payload is empty, and the subgraph still gets nothing. The two tests pass only because each half is fed a favorable manual input (one calls `get_stream_kwargs` with a `RunAgentInput` directly; the other pre-populates `invoke(context=...)`) — the real join is never exercised. Please serialize from the actual `RunAgentInput` (capture it in the `run()` override), and add one end-to-end test through the real library path with no manually-supplied context.

rodboev:
Thanks, you're right. I proved the read side, but I wasn't carrying the original AG-UI request through the point where the base agent swaps it for merged graph state.

I've updated it in two parts:

1. `LangGraphAGUIAgent.run(...)` now captures the frontend tools and app context from the real `RunAgentInput`, and `get_stream_kwargs(...)` merges that captured payload into LangGraph context while leaving the base stream input alone.
2. The regression now goes through `LangGraphAGUIAgent.run(...)` into the real `StateGraph` + `create_agent(...)` subgraph path, with no manually supplied LangGraph context, and verifies that the child middleware sees both the frontend tool and the app-context note.

I also removed the direct `get_stream_kwargs(...)` test that passed a favorable `RunAgentInput`, since it skipped the transformation that broke the production path.

BenTaylorDev (approved):
Verified locally against `langchain 1.2.15` / `ag-ui-langgraph 0.0.42`: full suite green (210 passed / 11 skipped), and all three new tests are genuine regression tests — reverting `get_stream_kwargs` to serialize from the stream input makes each fail, including `test_real_subgraph_context_bridge_reaches_child_agent`, which drives the real `StateGraph` + `create_agent` subgraph with no manually-supplied context and confirms the child middleware sees both the frontend tool and the app-context note. Capturing the `RunAgentInput` payload in `run()` (and clearing it in `finally`) is exactly the right seam, and the `supports_context` fallback to `config.configurable` covers older LangGraph. Thanks for chasing the real join — this is complete now.

PR #5374, opened 2026-06-10
fix(core): preserve tool result ordering for parallel frontend tool calls

### LINKED ISSUE #2809: 🐛 Bug: Agent executes only the first useFrontendTool when multiple tool calls are generated in a single request (OpenAI + ADK + LiteLLM)
### 📝 **Description**

When multiple `useFrontendTool` hooks are registered and the agent receives a user message containing **multiple actionable tasks**, only the **first** tool call is executed.
All subsequent tool calls are ignored.

This issue occurs **only with OpenAI models (e.g., GPT-4o)** when used through **ADK + LiteLLM**.
The model generates multiple `tool_call_id`s, but the backend does **not** return tool result messages for each of them—resulting in only the first frontend tool being executed.

Gemini models **do not exhibit this problem**, indicating that the issue lies in how OpenAI tool calls are handled through LiteLLM/ADK.

---

# ♻️ **Reproduction Steps**

### 1️⃣ **Frontend: Define two `useFrontendTool` hooks in `page.tsx`**

```typescript
"use client";

import { useCoAgent, useCopilotAction, useFrontendTool } from "@copilotkit/react-core";
import { CopilotKitCSSProperties, CopilotSidebar } from "@copilotkit/react-ui";
import { useState } from "react";

export default function CopilotKitPage() {
  const [themeColor, setThemeColor] = useState("#6366f1");

  useFrontendTool({
    name: "greetUser",
    parameters: [{
      name: "name",
      description: "Greet user with alert when any name is provided.",
      required: true,
    }],
    handler({ name }) {
      alert(`hello, ${name}!`);
      return {
        status: "success",
        message: `Greeted user with name ${name}`,
      };
    },
  });

  useFrontendTool({
    name: "setThemeColor",
    parameters: [{
      name: "themeColor",
      description: "The theme color to set. Make sure to pick nice colors.",
      required: true,
    }],
    handler({ themeColor }) {
      setThemeColor(themeColor);
      return {
        status: "success",
        message: `Background changed to ${themeColor}`,
      };
    },
  });

  return (
    <main style={{ "--copilot-kit-primary-color": themeColor } as CopilotKitCSSProperties}>
      <YourMainContent themeColor={themeColor} />
      <CopilotSidebar
        clickOutsideToClose={false}
        defaultOpen={true}
        labels={{
          title: "Popup Assistant",
          initial: """👋 Hi there! You're chatting with an agent that comes with a few tools.\n\nTry:\n- **Frontend Tools**: Set the theme to orange, Greet Alex"""
        }}
      />
    </main>
  );
}

function YourMainContent({ themeColor }: { themeColor: string }) {
  return (
    <div
      style={{ backgroundColor: themeColor }}
      className="h-screen w-screen flex justify-center items-center flex-col transition-colors duration-300"
    ></div>
  );
}
```

---

### 2️⃣ **Backend: ADK Agent Server (LiteLLM + OpenAI)**

```python
# 

"""Shared State feature."""

# from __future__ import annotations

from dotenv import load_dotenv
load_dotenv()
import json
from fastapi import FastAPI
from ag_ui_adk import ADKAgent, add_adk_fastapi_endpoint
# ADK imports
from google.adk.agents import LlmAgent
from google.genai import types

from typing import List, Optional

import litellm
from google.adk.models.lite_llm import LiteLlm
import asyncio
import os
import logging

litellm.ssl_verify = False

import logging
import sys

# Create handlers
console_handler = logging.StreamHandler(sys.stdout)
file_handler = logging.FileHandler("app.log", mode="a")  # saves logs to app.log

# Common log format
log_format = logging.Formatter('%(asctime)s - %(levelname)s - %(name)s - %(message)s')

console_handler.setFormatter(log_format)
file_handler.setFormatter(log_format)

# Configure root logger
logging.basicConfig(
    level=logging.DEBUG,
    handlers=[console_handler, file_handler]  # send logs to both console and file
)

# Example log
logging.debug("Logging initialized with file logging enabled.")


lite_llm_model = LiteLlm(
        model=f"azure/{os.getenv("AZURE_OPENAI_DEPLOYMENT_NAME")}",
        api_key=os.getenv("AZURE_OPENAI_API_KEY"),
        api_base=os.getenv("AZURE_OPENAI_ENDPOINT"),
        api_version=os.getenv("AZURE_OPENAI_API_VERSION"),
    )
# lite_llm_model = LiteLlm(
#         model=f"openai/{os.getenv("OPENAI_MODEL_NAME")}",
#         api_key=os.getenv("OPENAI_API_KEY"),
#     )
# lite_llm_model = LiteLlm(
#         model=f"gemini/{os.getenv("GEMINI_MODEL_NAME")}",
#         api_key=os.getenv("GEMINI_API_KEY"),
#     )

agent = LlmAgent(
        name="OrchesratorAgent",
        model=lite_llm_model,
        instruction=f"""
           You are an orchestrator agent with access to multiple agents and tools, call appropriate agent and tools according to user queries to complete the task.
        """,
    )

# Create ADK middleware agent instance
adk_agent = ADKAgent(
    adk_agent=agent,
    # app_name="orchestrator_app",
    app_name="agents",
    user_id="demo_user",

)

# Create FastAPI app
app = FastAPI(title="ADK Middleware Proverbs Agent")

# Add the ADK endpoint
# add_adk_fastapi_endpoint(app, adk_agent, path="/orchestrator")
add_adk_fastapi_endpoint(app, adk_agent, path="/agent")

if __name__ == "__main__":
    import os
    import uvicorn

    if not os.getenv("GOOGLE_API_KEY"):
        print("⚠️  Warning: GOOGLE_API_KEY environment variable not set!")
        print("   Set it with: export GOOGLE_API_KEY='your-key-here'")
        print("   Get a key from: https://makersuite.google.com/app/apikey")
        print()

    port = int(os.getenv("PORT", 9000))
    uvicorn.run(app, host="0.0.0.0", port=port)
```

---

### 3️⃣ **Send a message to the agent**

```
1. Greet Alex.
2. Change the theme to orange.
```

---

### 4️⃣ **Observed Result**

* Only **the first tool** (`greetUser`) runs.
* **Theme does NOT change**.

---

### 5️⃣ **Additional Example**

```
1. Change the theme to orange.
2. Greet Alex.
```

👉 Same issue: only the **first** tool executes.

---

### 📷 Screenshots

Image

Image

---

# ✅ **Expected Behavior**

Both tasks should execute:

* User is greeted via alert
* Theme color updates to **orange**

---

# ❌ **Actual Behavior**

Only the **first** task executes:

* Greeting alert appears
* **Theme color does not change**

---

# 💠 **CopilotKit Version**

```shell
@copilotkit/react-core@1.10.6
@copilotkit/react-ui@1.10.6
@copilotkit/runtime@1.10.6
```

---

# 🤖 **LLM Used**

* **OpenAI GPT-4o** (❌ reproduces the issue)
* **Gemini models** (✅ works correctly)

⚠️ Issue appears specific to **OpenAI tool call handling through LiteLLM + ADK**.

---

# 📄 **Logs (Optional)**

```shell
<logs go here>
```

rodboev:
## Summary
When an assistant message contains multiple parallel tool calls targeting frontend tools (`useFrontendTool`), the tool result messages are spliced into `agent.messages` in reverse order. This violates the ordering contract that some LLM providers (OpenAI via ADK + LiteLLM) enforce: tool results must appear in the same order as their corresponding `tool_call_id`s. The mismatch causes the provider to silently discard unmatched results during the follow-up run, making it appear as though only the first frontend tool executed.

## Root cause
`executeSpecificTool` and `executeWildcardTool` both used the fixed insertion point `messageIndex + 1`, where `messageIndex` is the assistant message's position. Because the assistant message's index doesn't shift when a tool result is inserted after it, each successive splice pushed previously inserted results to the right, reversing the final order.

## Changes
- `packages/core/src/core/run-handler.ts`: In both `executeSpecificTool` and `executeWildcardTool`, advance the insertion index past any tool-result messages already present after the parent assistant message before splicing.
- `packages/core/src/__tests__/core-parallel-tool-order.test.ts`: New regression coverage for parallel tool-call ordering, including the follow-up path, with type-safe assistant and tool-message assertions on current `main`.
- `.changeset/fix-parallel-tool-order.md`: Patch changeset for `@copilotkit/core`.

## What is NOT changed
- Sequential execution of tool handlers is preserved; only the splice insertion position changes.
- `runTool()` (programmatic API) is unaffected; it creates a single tool call per invocation.
- The `newMessages.findIndex` guard that skips already-answered tool calls is unchanged.
- Follow-up recursion logic is unchanged.

## Related PRs and Issues
Closes #2809.

## Test plan
- [x] `npx nx run @copilotkit/core:build`
- [x] `pnpm -C packages/core exec vitest run src/__tests__/core-parallel-tool-order.test.ts` — 3/3 pass
- [x] `pnpm -C packages/core exec vitest run` — 586/586 pass
- [x] `pnpm -C packages/core exec tsc --noEmit`
- [ ] CI green

### DIFF
--- a/.changeset/fix-parallel-tool-order.md
+++ b/.changeset/fix-parallel-tool-order.md
@@ -1,0 +1,5 @@
+---
+"@copilotkit/core": patch
+---
+
+fix(core): preserve tool result ordering for parallel frontend tool calls
--- a/packages/core/src/__tests__/core-parallel-tool-order.test.ts
+++ b/packages/core/src/__tests__/core-parallel-tool-order.test.ts
@@ -1,0 +1,207 @@
+import { describe, it, expect, beforeEach, vi, afterEach } from "vitest";
+import { CopilotKitCore } from "../core";
+import {
+  MockAgent,
+  createAssistantMessage,
+  createTool,
+  createMultipleToolCallsMessage,
+} from "./test-utils";
+
+describe("CopilotKitCore.runAgent - Parallel Tool Order", () => {
+  let copilotKitCore: CopilotKitCore;
+
+  beforeEach(() => {
+    copilotKitCore = new CopilotKitCore({});
+    vi.clearAllMocks();
+  });
+
+  afterEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  it("should preserve tool result ordering for parallel tool calls with followUp: false", async () => {
+    const toolA = createTool({
+      name: "toolA",
+      handler: vi.fn(async () => "Result A"),
+      followUp: false,
+    });
+    const toolB = createTool({
+      name: "toolB",
+      handler: vi.fn(async () => "Result B"),
+      followUp: false,
+    });
+
+    copilotKitCore.addTool(toolA);
+    copilotKitCore.addTool(toolB);
+
+    // Create a message with two parallel tool calls in order: [A, B]
+    const message = createMultipleToolCallsMessage([
+      { name: "toolA" },
+      { name: "toolB" },
+    ]);
+
+    const agent = new MockAgent({ newMessages: [message] });
+    copilotKitCore.addAgent__unsafe_dev_only({
+      id: "test",
+      agent: agent as any,
+    });
+
+    await copilotKitCore.runAgent({ agent: agent as any });
+
+    // Verify both handlers were called
+    expect(toolA.handler).toHaveBeenCalled();
+    expect(toolB.handler).toHaveBeenCalled();
+
+    // The message structure should be: [assistant, toolResultA, toolResultB, ...]
+    // Find the assistant message
+    const assistantIndex = agent.messages.findIndex(
+      (m) => m.id === message.id,
+    );
+    expect(assistantIndex).toBeGreaterThanOrEqual(0);
+
+    // The next two messages should be tool results in order
+    const firstToolResult = agent.messages[assistantIndex + 1];
+    const secondToolResult = agent.messages[assistantIndex + 2];
+
+    expect(firstToolResult).toBeDefined();
+    expect(secondToolResult).toBeDefined();
+    expect(firstToolResult?.role).toBe("tool");
+    expect(secondToolResult?.role).toBe("tool");
+
+    // Verify the order: first result corresponds to toolA, second to toolB
+    const toolCallIds = message.toolCalls?.map((tc) => tc.id) || [];
+    expect(firstToolResult?.toolCallId).toBe(toolCallIds[0]);
+    expect(secondToolResult?.toolCallId).toBe(toolCallIds[1]);
+  });
+
+  it("should preserve tool result ordering with one tool having followUp: true", async () => {
+    const toolA = createTool({
+      name: "toolA",
+      handler: vi.fn(async () => "Result A"),
+      followUp: false,
+    });
+    const toolB = createTool({
+      name: "toolB",
+      handler: vi.fn(async () => "Result B"),
+      followUp: true, // This will trigger a follow-up
+    });
+
+    copilotKitCore.addTool(toolA);
+    copilotKitCore.addTool(toolB);
+
+    // Create a message with two parallel tool calls in order: [A, B]
+    const message = createMultipleToolCallsMessage([
+      { name: "toolA" },
+      { name: "toolB" },
+    ]);
+
+    const followUpMessage = createAssistantMessage({
+      content: "Follow-up response",
+    });
+
+    const agent = new MockAgent({ newMessages: [message] });
+    copilotKitCore.addAgent__unsafe_dev_only({
+      id: "test",
+      agent: agent as any,
+    });
+
+    let callCount = 0;
+    agent.runAgentCallback = () => {
+      callCount++;
+      if (callCount === 2) {
+        agent.setNewMessages([followUpMessage]);
+      }
+    };
+
+    await copilotKitCore.runAgent({ agent: agent as any });
+
+    // Verify both handlers were called
+    expect(toolA.handler).toHaveBeenCalled();
+    expect(toolB.handler).toHaveBeenCalled();
+
+    // Verify the follow-up was triggered
+    expect(agent.runAgentCalls.length).toBeGreaterThan(1);
+
+    // Find the assistant message from the first call
+    const assistantIndex = agent.messages.findIndex(
+      (m) => m.id === message.id,
+    );
+    expect(assistantIndex).toBeGreaterThanOrEqual(0);
+
+    // The next two messages should be tool results in order
+    const firstToolResult = agent.messages[assistantIndex + 1];
+    const secondToolResult = agent.messages[assistantIndex + 2];
+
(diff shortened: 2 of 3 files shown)


BenTaylorDev (approved):
**Approving.** Verified against current `main`: each parallel frontend-tool result was splicing at `messageIndex + 1` while the assistant index stayed fixed, reversing result order (A,B,C → C,B,A) at both sites (`run-handler.ts:721` and `:859`, both still live on `main`). The fix advances `insertAt` past the contiguous tool-result block before splicing, preserving call order with `tool_call`↔`ToolMessage` pairing intact, and it's applied at both sites. The test drives the real `runAgent`→`executeSpecificTool` splice path (only the transport is mocked).

BenTaylorDev (changes requested):
Following up on my approval — I ran the `@copilotkit/core` checks locally with this on current `main`, and it doesn't pass type-check, so I'm converting this to request-changes. Tests pass (586/586) and lint is clean, but `tsc --noEmit` fails (exit 2):

- **Source** — `run-handler.ts:722` and `:871` (the fix itself): `TS2532: Object is possibly 'undefined'`. `agent.messages[insertAt].role` isn't narrowed by the `insertAt < agent.messages.length` guard (TS doesn't infer definedness from a length check under the indexed-access rules). Use optional chaining: `agent.messages[insertAt]?.role === "tool"`.
- **Test** — `core-parallel-tool-order.test.ts` (lines ~72/139/202): `.toolCalls` / `.toolCallId` are accessed on the `Message` union without narrowing (TS2339), and `tc` is implicitly `any` (TS7006). Narrow by `role` (or assert the assistant/tool message type) and type the `tc` param.

The ordering fix itself is correct — I verified the reversed-insertion behavior and that advancing `insertAt` past the contiguous tool block restores call order at both sites. It just needs these type errors cleared so `check-types` is green. Apologies for the churn on my end.

rodboev:
Thanks for checking this on current `main`.

1. `packages/core/src/core/run-handler.ts` now uses `agent.messages[insertAt]?.role === "tool"` in both insertion scans, so the ordering fix stays the same and the indexed read is type-safe in both execution paths.
2. `packages/core/src/__tests__/core-parallel-tool-order.test.ts` now narrows the assistant message before reading `toolCalls` and narrows each inserted result before reading `toolCallId`, which clears the union errors without weakening the ordering assertions.
3. The branch is rebased onto current `main`, the core build passes, the full `@copilotkit/core` suite still passes, and `tsc --noEmit` is clean.

BenTaylorDev (approved):
**Re-approving** — the type errors are fixed. Verified on the updated branch against current `main`: `@copilotkit/core` tests 586/586, `tsc --noEmit` clean, lint 0 errors — and CI agrees (unit ×3, check-types, build, oxlint, format all green). The remaining red `build-check`/Vercel jobs are fork-token/registry failures (`permission_denied: Invalid token`), not code. The ordering fix is correct across both splice sites.

PR #5375, opened 2026-06-10
fix(core): prevent runAgent hang after frontend restart with stale connect state

### LINKED ISSUE #3009: 🐛 Bug: useAgent().runAgent() hangs indefinitely after frontend restart when backend thread state already exists
### ♻️ Reproduction Steps

1. Create a CopilotKit + LangGraph setup using `useAgent({ agentId: "research_assistant_agent" })` on the frontend and a LangGraph-backed Copilot Runtime on the backend.
2. Start a conversation so that a backend thread state is created and persisted.
3. Refresh or restart the frontend application without clearing backend thread state.
4. Call:

```ts
const { agent } = useAgent({ agentId: "research_assistant_agent" });

await agent.setState({ ...existingState, command });
await agent.runAgent();
```

5. Observe that `runAgent()` never resolves and no further network or SSE activity occurs.

This happens consistently after a frontend reload when the backend already has an existing thread state for the same threadId.

### ✅ Expected Behavior

`agent.runAgent()` should execute the LangGraph agent normally and return once the command finishes, even if the frontend was restarted and the backend thread state already exists.

### ❌ Actual Behavior

`agent.runAgent()` hangs indefinitely and never resolves.

No network request is sent to the runtime after the call, and no errors appear in the console. The UI remains stuck in a “running” state forever.

To work around this, I had to bypass `useAgent().runAgent()` entirely and call a custom backend HTTP endpoint that directly invokes the LangGraph graph.

Workaround hook example:

```ts
// Direct HTTP call replacing agent.runAgent()
const response = await fetch(
  `/projects/research-assistant/${threadId}/command`,
  {
    method: "POST",
    headers: { "Content-Type": "application/json" },
    credentials: "include",
    body: JSON.stringify({ cmd, payload }),
  }
);
```

This workaround reliably executes the agent, confirming the backend graph is functional and the hang is isolated to `useAgent().runAgent()` after frontend restart.

### 𝌚 CopilotKit Version

```shell
1.51.2
```

### 📄 Logs (Optional)

```shell

```

rodboev:
## Summary
After a page refresh, calling `agent.runAgent()` (via `useAgent()`) hangs indefinitely when `CopilotChat` has an active `connectAgent` SSE pipeline for the same thread. No HTTP request reaches the runtime, no error appears, and the UI stays in a "running" state forever. The hang is in `detachActiveRun()` awaiting a completion promise that the stale connect pipeline never resolves.

## Root cause
`ProxiedCopilotRuntimeAgent.detachActiveRun()` (`packages/core/src/agent.ts:180-185`) delegates to `super.detachActiveRun()` (AG-UI `HttpAgent`), which signals the active pipeline's takeUntil Subject and then `await`s an internal `activeRunCompletionPromise`. If the connect pipeline's Observable finalize block doesn't run (stale SSE connection, half-open socket after reload, or the abort signal failing to propagate through the RxJS chain), the promise never resolves. Since the `await` has no timeout, `detachActiveRun()` blocks forever, which in turn blocks `runAgent()`.

The reporter's workaround (bypassing `agent.runAgent()` with a direct HTTP POST to the backend) confirms the backend graph is functional; the hang is isolated to the client-side agent lifecycle.

## Changes
- `packages/core/src/agent.ts`: Add a 5-second timeout guard to `ProxiedCopilotRuntimeAgent.detachActiveRun()`. If the delegate or base-class detach doesn't resolve within 5 seconds, force-continue and log a warning. Reset `this.isRunning = false` as a safety net after the timeout to prevent the agent from being permanently stuck. (~+25 lines net)
- `packages/core/src/__tests__/core-detach-timeout.test.ts`: New test verifying that `detachActiveRun()` resolves within the timeout when the completion promise never settles. Covers: timeout fires, warning logged, `isRunning` reset. (~+60 lines)
- `.changeset/fix-runagent-hang.md`: Patch changeset for `@copilotkit/core`.

## What is NOT changed
- `RunHandler.runAgent()` and `RunHandler.connectAgent()` are unchanged; the fix is at the agent instance layer, benefiting both `copilotkit.runAgent()` and direct `agent.runAgent()` callers.
- The `CopilotChat` connect effect and its cleanup logic are unchanged.
- The AG-UI client's `AbstractAgent.runAgent()` and `HttpAgent.detachActiveRun()` are not modified; the fix wraps them at the CopilotKit proxy layer.

## Related PRs and Issues
Closes #3009.

## Test plan
- [x] `pnpm -C packages/core exec vitest run src/__tests__/core-detach-timeout.test.ts` — 5/5 pass
- [ ] CI green

### DIFF
--- a/.changeset/fix-runagent-hang.md
+++ b/.changeset/fix-runagent-hang.md
@@ -1,0 +1,5 @@
+---
+"@copilotkit/core": patch
+---
+
+fix(core): prevent runAgent hang after frontend restart with stale connect state
--- a/packages/core/src/__tests__/core-detach-timeout.test.ts
+++ b/packages/core/src/__tests__/core-detach-timeout.test.ts
@@ -1,0 +1,183 @@
+import { describe, it, expect, beforeEach, afterEach, vi } from "vitest";
+import { ProxiedCopilotRuntimeAgent } from "../agent";
+
+describe("ProxiedCopilotRuntimeAgent - detachActiveRun timeout guard", () => {
+  let agent: ProxiedCopilotRuntimeAgent;
+  let consoleWarnSpy: ReturnType<typeof vi.spyOn>;
+
+  beforeEach(() => {
+    agent = new ProxiedCopilotRuntimeAgent({
+      agentId: "test-agent",
+      runtimeUrl: "http://localhost:8000",
+    });
+    consoleWarnSpy = vi.spyOn(console, "warn").mockImplementation(() => {});
+  });
+
+  afterEach(() => {
+    vi.restoreAllMocks();
+    vi.clearAllTimers();
+  });
+
+  it("should timeout and continue when delegate.detachActiveRun() never resolves", async () => {
+    vi.useFakeTimers();
+
+    // Mock delegate that returns a never-resolving promise
+    const mockDelegate = {
+      detachActiveRun: vi.fn(
+        () =>
+          new Promise<void>(() => {
+            // Never resolves
+          }),
+      ),
+    };
+    agent["delegate"] = mockDelegate as any;
+
+    // Mock super.detachActiveRun to return a resolving promise
+    const originalDetach = Object.getPrototypeOf(
+      Object.getPrototypeOf(agent),
+    ).detachActiveRun;
+    vi.spyOn(
+      Object.getPrototypeOf(Object.getPrototypeOf(agent)),
+      "detachActiveRun",
+      "get",
+    ).mockReturnValue(
+      vi.fn(() => Promise.resolve()),
+    );
+
+    const detachPromise = agent.detachActiveRun();
+
+    // Fast-forward through the 5s timeout
+    await vi.advanceTimersByTimeAsync(5_000);
+
+    // Should complete without hanging
+    await expect(detachPromise).resolves.toBeUndefined();
+
+    // Should have warned about the timeout
+    expect(consoleWarnSpy).toHaveBeenCalledWith(
+      expect.stringContaining("delegate.detachActiveRun()"),
+    );
+    expect(consoleWarnSpy).toHaveBeenCalledWith(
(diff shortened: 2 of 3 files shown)


rodboev:
I’m pausing this one for maintainer guidance before changing the branch. The issue reports a `runAgent()` hang after reload, and this PR adds a defensive timeout around `detachActiveRun()`.

While re-auditing, I found related lifecycle work: #3516 fixed a previous `detachActiveRun` deadlock by updating `@ag-ui/client`, and #5380 is an open maintainer PR changing adjacent detach behavior for in-flight runs during thread switching. This branch still tests and merges cleanly, but I do not want to push a broad timeout guard if the preferred fix is in the chat/thread cleanup path.

Would you prefer this timeout guard, a narrower repro tied to the current main path, or closing this in favor of the lifecycle work?

BenTaylorDev:
Thanks for this. The indefinite hang here was the `connectAgent` → `ConnectNotImplementedError` cleanup deadlock, and it's since been resolved upstream: `@ag-ui/client ≥0.0.42` always runs the finalize block (`ConnectNotImplementedError → EMPTY`), so the detach completion promise resolves reliably. Current `main` pins `0.0.57` and deliberately `await`s `detachActiveRun()` again — the rationale is documented right in the code at `run-handler.ts:380-384`. So the restart hang from #3009 shouldn't reproduce on current `main`.

The PR's 5-second timeout guards a condition the dependency bump already eliminated, and it carries some risk of prematurely tearing down a legitimately slow in-flight connect; its test also mocks a completion promise that never settles, whereas on current `main` that promise does settle. It also adds a `.changeset/` file (the repo has since moved off changesets).

Closing as superseded — if you can still reproduce the hang on latest `main` (frontend restart with existing backend thread state) on `@ag-ui/client 0.0.57`, please reopen and we'll dig into the current detach path.

PR #5380, opened 2026-06-11
fix(react-core): detach in-flight runs when switching from non-explicit threads

tylerslaton:
## Bug

In the v2 `CopilotChat` connect effect, the non-explicit-thread guard returned early **before** the effect's cleanup was registered:

```tsx
agent.threadId = resolvedThreadId;
if (!hasExplicitThreadId) return; // ← the cleanup below never registers
// ...
return () => {
  detached = true;
  connectAbortController.abort();
  void agent.detachActiveRun().catch(() => {});
};
```

`useAgent` hands every chat the same shared agent instance per `agentId` (`copilotkit.getAgent(agentId)`), with `threadId` mutated in place. So when a run is still streaming on a non-explicit (fresh/auto-minted) thread and the user switches threads (via the `threadId` prop or `CopilotChatConfigurationProvider`), the in-flight run is never detached:

- the old run keeps applying events to the shared agent after `agent.threadId` is repointed at the new thread, and
- its stream collides with the new thread's connect replay, surfacing as repeated `AGUIError: Cannot send 'RUN_STARTED' while a run is still active. The previous run must be finished with 'RUN_FINISHED' before starting a new run.`

Symptoms are timing-dependent: the old stream keeps rendering, the new thread's replay fails, or both.

**Repro** (verified 2026-06-10 against published 1.59.5, Intelligence mode): fresh `CopilotChat` (no explicit `threadId`) → send a message that streams for several seconds → while streaming, set an explicit `threadId` (switch threads) → console fills with the AGUIError above.

## Fix

Register the cleanup unconditionally; gate only the `connect()` call (and its `AbortController` wiring) on `hasExplicitThreadId`. The cleanup now detaches the in-flight run **before** the new effect body repoints the shared agent — and it also covers the two paths where nothing else would ever detach:

- switching to **another non-explicit** thread (new-chat-while-streaming): non-explicit destinations never call `connectAgent`, so the core-side defensive detach never runs;
- **unmount** mid-run.

`CopilotKitCore.connectAgent` already detaches defensively on the core side (`run-handler.ts`, covered by `core-connect-thread-switch.test.ts`), so no core change is needed — but that detach only fires for explicit destinations, and only after `agent.threadId` has already been mutated underneath the in-flight run. The effect cleanup is the only teardown that runs for every outgoing thread.

## Tests

Two regression tests in `CopilotChat.threadSwitchDetach.test.tsx`, written first and verified to fail against the buggy code:

1. **non-explicit → explicit switch mid-run** — the first detach after the switch must arrive while the agent still points at the outgoing thread (cleanup ordering), and the new thread still connects. Failed before the fix with `expected 'explicit-thread-b' to be 'mock-thread-id'`: the only detach came from `connectAgent`'s internal defense, after the agent was repointed.
2. **non-explicit → non-explicit switch mid-run** — the run must be detached even though no connect is issued, and `/connect` stays skipped. Failed before the fix: no detach at all.

Verification:
- `nx run @copilotkit/react-core:test` — 1273/1273 pass
- `nx run @copilotkit/react-core:build` — clean
- `check-types` — fails identically on `main` (pre-existing, 268 errors incl. a2ui-renderer's missing `@a2ui/lit`; not a CI gate). The new test file introduces no new error class — only the same pre-existing `MockStepwiseAgent`/`AbstractAgent` `activeRunCompletionPromise` privacy conflict every sibling test file already exhibits.

## Context

Found while building the control-room demo (#5378). Its `ThreadsPanel.switchThread` detaches in-flight runs before switching as a workaround — that can stay (harmless double-detach), but with this fix other consumers don't need it.

### DIFF
--- a/packages/react-core/src/v2/components/chat/CopilotChat.tsx
+++ b/packages/react-core/src/v2/components/chat/CopilotChat.tsx
@@ -231,13 +231,6 @@ export function CopilotChat({
     // ship the same SDK-generated threadId that the chat UI is rendering.
     agent.threadId = resolvedThreadId;
 
-    // When the caller hasn't picked a specific thread, resolvedThreadId is a
-    // UUID minted locally (either in this CopilotChat or in a wrapping
-    // ThreadsProvider). The backend has never seen it, so /connect would
-    // always 404 — skip the call. A real thread is only created once the
-    // user runs the agent for the first time.
-    if (!hasExplicitThreadId) return;
-
     let detached = false;
 
     // Create a fresh AbortController so we can cancel the HTTP request on cleanup.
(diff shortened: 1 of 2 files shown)


BenTaylorDev:
Checked with Tyler - closed as defunct.

Issue #5369, opened 2026-06-10
Runtime info flattens per-agent A2UI scoping — a2ui catalog/schema/guidelines context (~30KB) is injected into every agent's runs


contextablemark:
## Summary

When a `CopilotRuntime` is configured with an `a2ui` block scoped to specific agents (`a2ui: { agents: [...] }`), the runtime-info response reduces that scoping to an endpoint-wide boolean. The client then injects the full A2UI catalog context — catalog capabilities, the complete component JSON Schema, generation guidelines, and design guidelines (~30KB) — into **every** agent run served by that endpoint, including agents that have nothing to do with A2UI.

Observed on the **1.59.5** release line (`@copilotkit/runtime`, `@copilotkit/core`, `@copilotkit/react-core`), which is current `latest`.

## The inconsistency

The runtime's **run path respects the per-agent scoping** — `dist/v2/runtime/handlers/shared/agent-utils.cjs`:

```js
if (runtime.a2ui) {
    const { agents: targetAgents, ...a2uiOptions } = runtime.a2ui;
    if ((!targetAgents || targetAgents.includes(agentId)) && typeof agent.use === "function")
        agent.use(new A2UIMiddleware(a2uiOptions));   // applied ONLY to agents in the list
}
```

…but the **info response flattens it** — `dist/v2/runtime/handlers/get-runtime-info.cjs`:

```js
a2uiEnabled: !!runtime.a2ui,   // per-agent scoping in runtime.a2ui.agents is discarded
```

## Full chain to the leak

1. `@copilotkit/runtime` advertises `a2uiEnabled: !!runtime.a2ui` (above).
2. `@copilotkit/core` stores it globally — `AgentRegistry`: `this._a2uiEnabled = runtimeInfoResponse.a2uiEnabled ?? false;`
3. `@copilotkit/react-core`'s `CopilotKitProvider` auto-mounts `A2UICatalogContext` whenever that flag is true:
   ```js
   runtimeA2UIEnabled && jsx(A2UICatalogContext, { catalog: a2ui?.catalog, includeSchema: a2ui?.includeSchema })
   ```
4. `A2UICatalogContext` calls `copilotkit.addContext(...)` with the four entries. `addContext` is the **global** context store, so they're attached to **every** `runAgent` — for every agent on the endpoint.

There is also no escape hatch: the provider's `a2ui` prop only configures `theme` / `catalog` / `includeSchema`; there is no `enabled: false`.

## Impact

[...]

PR #5385, opened 2026-06-11
fix(runtime): add a2ui enabled opt-out via shared isA2UIEnabled predicate

ranst91:
## Summary

Follow-up to #5383 (per-agent A2UI scoping). Two small, backwards-compatible runtime changes:

1. **One shared `isA2UIEnabled()` predicate.** Before, "is a2ui on?" was computed independently in two places — `if (runtime.a2ui)` in the run path (`agent-utils.ts`) and `!!runtime.a2ui` in the `/info` response (`get-runtime-info.ts`). Two separate checks can drift apart, and that drift between the run path and the info path is the root of #5369. Both now call the same predicate, so they can't disagree.

2. **An explicit `enabled` off-switch** on the runtime `a2ui` config. Previously, once `a2ui` was configured there was no way to turn it off short of deleting the whole block. Now `a2ui: { enabled: false }` disables it while keeping the rest of the config (e.g. `schema`/`catalog`) in place.

```ts
export function isA2UIEnabled(a2ui) {
  return !!a2ui && a2ui.enabled !== false;
}
```

## Backwards compatibility

No breaking changes. `enabled` is a new optional field; every existing config behaves identically:

- omitted `a2ui` → off (unchanged)
- `a2ui: {}` → on (unchanged)
- `a2ui: { injectA2UITool: true }` / `{ agents: [...] }` → on (unchanged)
- only an explicit `a2ui: { enabled: false }` is new behavior — and it's an opt-out nobody could have relied on.

Enablement stays entirely server-side; no client/provider API change.

## Scope

This does **not** fix the #5369 leak — #5383 does that via per-agent scoping. This is complementary hardening: it adds the missing off-switch and removes the run-path/info-path divergence that allowed the leak to exist in the first place.

## Tests

- `get-runtime-info`: reports `a2uiEnabled: false` when `a2ui: { enabled: false }`.
- run path (`handle-run`): does not apply `A2UIMiddleware` when `a2ui.enabled` is `false`.

https://claude.ai/code/session_01TYohiEJyhsU3mJS4jabdv6

---
_Generated by [Claude Code](https://claude.ai/code/session_01TYohiEJyhsU3mJS4jabdv6)_


contextablemark (approved):
looks good

PR #5386, opened 2026-06-11
fix(react-core): stabilize pin-to-send scrolling

### LINKED ISSUE #5355: 🐛 Bug: Scroll to bottom button not appearing and user message shifts after streaming finishes in "pin to send" autoscroll mode
### ♻️ Reproduction Steps

We are using `<CopilotPopup />` at the base level to reproduce:

```
<CopilotPopup 
  scrollView={{
  	autoScroll: 'pin-to-send',
  }}
/>
```

We are seeing the same errors in the base level reproduction as we see using our full overrides:

```
<CopilotPopup
	header={ResolvedHeader}
	input={ResolvedInput}
	scrollView={{
		scrollToBottomButton: ScrollButton,
        autoScroll: 'pin-to-send',
	}}
	toggleButton={hideToggle ? HideToggleOverride : undefined}
	messageView={ResolvedMessageView}
	suggestionView={{
		suggestion: SuggestionPill,
		container: suggestionContainerCss,
	}}
	welcomeScreen={false}
	width={isResizable ? width : defaultWidth}
	height={isResizable ? height : defaultHeight}
	id={id}
	{...dataProps}
/>
```

To reproduce, chat with the agent in pin to send autoscroll mode. We see that the scroll to bottom button does not appear and that the user messages shifts abruptly after the agent finishes streaming.

Reproduction with CPK chat with overriden slots:

https://github.com/user-attachments/assets/6b0020f5-c251-47a6-bf8b-e18d54c20480

Reproduction with base CPK chat:

https://github.com/user-attachments/assets/fdb3ab45-c656-48d1-b79e-02fd2c24d76f

### ✅ Expected Behavior

Scroll to bottom button should be present in pin to send mode and chat message should remain in the same position after streaming

### Root cause analysis

For bottom scroll button not appearing, I believe the reason is that the useEffect in ScrollView that attaches scroll listeners has [scrollRef, mode] as dependencies. scrollRef is a stable useRef object, so the effect only runs on the initial render when scrollRef.current is still null (because the hasMounted guard renders a placeholder on the first pass, before the ref callback div mounts). The effect early-returns and never re-runs, leaving showScrollButton permanently false. (https://github.com/CopilotKit/CopilotKit/blob/c4a5f9d57ccddfd0d70cdc0f051bc09fc4be6152/packages/react-core/src/v2/components/chat/CopilotChatView.tsx#L694) 

I'm unsure what the root cause for the flash of content for user message is. Looking for help from CPK team on this!

westkite1201:
## What does this PR do?

Fixes `pin-to-send` scrolling in the v2 chat view.

- Re-attaches the non-autoscroll scroll listener after the real scroll element mounts by depending on `nonAutoScrollEl`, not the stable `scrollRef` object.
- Lets the `usePinToSend` spacer adjust in both directions as content below the pinned user message changes, so the user message stays anchored after streaming finishes and layout height changes.
- Adds regression coverage for the scroll-to-bottom button and spacer adjustment behavior.

## Related PRs and Issues

Fixes #5355

## Tests

- `corepack pnpm -C packages/react-core exec vitest run src/v2/hooks/__tests__/use-pin-to-send.test.tsx src/v2/components/chat/__tests__/CopilotChatView.pinToSend.test.tsx`
- `corepack pnpm exec oxfmt --check packages/react-core/src/v2/components/chat/CopilotChatView.tsx packages/react-core/src/v2/hooks/use-pin-to-send.ts packages/react-core/src/v2/hooks/__tests__/use-pin-to-send.test.tsx packages/react-core/src/v2/components/chat/__tests__/CopilotChatView.pinToSend.test.tsx`
- `git diff --check`

Attempted:

- `corepack pnpm -C packages/react-core run check-types`
  - This failed in the local workspace on existing/type-resolution issues outside this diff, including `react-markdown` JSX namespace errors, missing `@copilotkit/runtime-client-gql` declarations, and existing e2e mock `AbstractAgent` private member mismatches.

## Checklist

- [x] If the PR changes or adds functionality, I have updated the relevant documentation (N/A: bug fix only, no API/docs change)

### DIFF
--- a/packages/react-core/src/v2/components/chat/CopilotChatView.tsx
+++ b/packages/react-core/src/v2/components/chat/CopilotChatView.tsx
@@ -668,7 +668,7 @@ export namespace CopilotChatView {
     useEffect(() => {
       if (mode === "pin-to-bottom") return; // Skip for autoscroll mode
 
-      const scrollElement = scrollRef.current;
+      const scrollElement = nonAutoScrollEl;
       if (!scrollElement) return;
 
       const checkScroll = () => {
@@ -691,7 +691,7 @@ export namespace CopilotChatView {
         scrollElement.removeEventListener("scroll", checkScroll);
         resizeObserver.disconnect();
       };
-    }, [scrollRef, mode]);
+    }, [nonAutoScrollEl, mode]);
 
     if (!hasMounted) {
       return (
--- a/packages/react-core/src/v2/components/chat/__tests__/CopilotChatView.pinToSend.test.tsx
+++ b/packages/react-core/src/v2/components/chat/__tests__/CopilotChatView.pinToSend.test.tsx
@@ -1,5 +1,5 @@
-import { describe, it, expect, vi, beforeEach } from "vitest";
-import { render } from "@testing-library/react";
+import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
+import { fireEvent, render, waitFor } from "@testing-library/react";
 import React from "react";
 import { CopilotKitProvider } from "../../../providers/CopilotKitProvider";
 import { CopilotChatConfigurationProvider } from "../../../providers/CopilotChatConfigurationProvider";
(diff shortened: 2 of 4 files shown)


tylerslaton:
Thank you for the contribution @westkite1201!

tylerslaton (approved):

PR #5387, opened 2026-06-11
fix(react-core): preserve markdown code block newlines

### LINKED ISSUE #3330: 🐛 Bug: Markdown code block rendering doesn't preserve newlines
### ♻️ Reproduction Steps

1. Start the v2 storybook: `pnpm storybook:react`
2. Open the story on `CopilotChatAssistantMessage` that tests all markdown features: `open http://localhost:6006/?path=/story/ui-copilotchatassistantmessage--test-all-markdown-features`
3. Observe non-preserved newlines in code blocks
...

### ✅ Expected Behavior

Code blocks rendered with markdown preserve newlines.

### ❌ Actual Behavior

Code blocks are illegible, because newlines are removed.

This happened after upgrading from v1.51.4 to v1.52.1. I haven't tested with v1.52.0, but I _suspect_ the bug is there too, and that it maybe stems from the tailwind style scoping and style generation pipeline?

### Screenshots

Image

### 𝌚 CopilotKit Version

```shell
1.52.1
```

### 📄 Logs (Optional)

```shell

```

westkite1201:
## What does this PR do?

Fixes markdown code blocks in the v2 React chat UI so newlines are preserved when Streamdown renders fenced code blocks.

This adds `cpk:whitespace-pre` to the existing `[data-streamdown="code-block"] > pre` rule and includes a small regression test that keeps the CSS rule from dropping the whitespace preservation utility again.

I also checked for existing open and closed PRs by issue number and root-cause terms (`#3330`, code block, newlines, `whitespace-pre`, Streamdown) before opening this PR and did not find an overlapping PR.

Testing:

- `corepack pnpm -C packages/react-core exec vitest run src/v2/styles/__tests__/globals-css.test.ts`
- `corepack pnpm -C packages/react-core run build:css`
- `corepack pnpm exec oxfmt --check packages/react-core/src/v2/styles/__tests__/globals-css.test.ts packages/react-core/src/v2/styles/globals.css`
- `git diff --check`

## Related PRs and Issues

Fixes #3330

## Checklist

- [x] If the PR changes or adds functionality, I have updated the relevant documentation (not applicable: scoped visual bug fix)

### DIFF
--- a/packages/react-core/src/v2/styles/__tests__/globals-css.test.ts
+++ b/packages/react-core/src/v2/styles/__tests__/globals-css.test.ts
@@ -1,0 +1,17 @@
+import { readFileSync } from "node:fs";
+import { fileURLToPath } from "node:url";
+import { dirname, join } from "node:path";
+import { describe, expect, it } from "vitest";
+
+const currentDir = dirname(fileURLToPath(import.meta.url));
+const globalsCss = readFileSync(join(currentDir, "../globals.css"), "utf8");
+
+describe("v2 globals.css", () => {
+  it("preserves newlines inside streamdown code blocks", () => {
+    const codeBlockPreRule = globalsCss.match(
+      /\[data-copilotkit\]\s+div\[data-streamdown="code-block"\]\s+>\s+pre\s*\{[^}]*\}/,
+    );
+
+    expect(codeBlockPreRule?.[0]).toContain("cpk:whitespace-pre");
+  });
+});
--- a/packages/react-core/src/v2/styles/globals.css
+++ b/packages/react-core/src/v2/styles/globals.css
@@ -177,7 +177,7 @@
 }
 
 [data-copilotkit] div[data-streamdown="code-block"] > pre {
-  @apply cpk:mt-0 cpk:mb-0;
+  @apply cpk:mt-0 cpk:mb-0 cpk:whitespace-pre;
 }
 
 /* Minimal scrollbar for WebKit browsers (Chrome, Safari, Edge) */


BenTaylorDev (changes requested):
Thanks — this is on the right, live selector (`div[data-streamdown="code-block"] > pre` still matches in streamdown 1.6.11), it merges cleanly, it's strictly additive (can't regress prose whitespace), and it fits the pattern maintainers have been using to re-declare streamdown elements under the `cpk:` prefix. My concern is whether it actually fixes the *reported* symptom:

- Streamdown renders each code line as a `display:block` span with **no newline characters between the spans** — the line breaks come from that `block` utility, not from `\n`. CopilotKit's build is `@import "tailwindcss" prefix(cpk)` scanning react-core source only (not `node_modules/streamdown`, which ships no CSS), so streamdown's unprefixed `block` class isn't emitted and the per-line spans have no `data-streamdown` attribute to re-declare. In that state `white-space: pre` on the `<pre>` preserves intra-line indentation but can't restore inter-line breaks (there are no newline chars to preserve).
- The test only `readFileSync`s `globals.css` and asserts it contains `cpk:whitespace-pre` — it never renders Streamdown, so green CI isn't evidence the newlines are actually preserved.

Could you render a multi-line fenced code block in the v2 chat with this applied (in a consumer relying on CopilotKit's packaged CSS, not one whose own Tailwind scans `node_modules/streamdown`) and confirm both line breaks *and* indentation survive? If lines still collapse, the fix is broader than this one-liner (supplying streamdown's line-span styles under the prefix). Also note commenters report the code-block *header* is broken too — separate from this change.

BenTaylorDev:
Re-reviewed this with a live render rather than by inspection, and the result is conclusive: the change has no effect.

I built the packaged CSS from both `main` and this branch and rendered streamdown 1.6.11's actual code-block DOM in a browser. Computed `white-space` on the `<pre>`, the line-span `display`, and the number of rendered line rows are identical before and after — a 3-line fenced block still collapses onto one row in both.

Two independent reasons:

1. **The declaration is redundant.** The UA stylesheet already applies `white-space: pre` to `<pre>`, and nothing in CopilotKit's packaged CSS overrides it — Tailwind v4's preflight only sets font properties on `pre` (`tailwindcss/preflight.css:109-127`), and the built `dist/v2/index.css` has no competing rule. The compiled delta is exactly `+white-space:pre`, which changes no computed value.

2. **There are no newlines to preserve.** Rendering `Streamdown` through the repo's own test harness, `pre.textContent` comes back as ``"function greet(name) {  return `hi ${name}`;}"`` — no `\n` anywhere. Streamdown splits the source on `\n` and emits one `<span class="block …">` per line; the line breaks come entirely from that unprefixed `block` class, which CopilotKit's `@import "tailwindcss" prefix(cpk)` build never emits (the compiled CSS contains `.cpk\:block` only).

The rule that does fix it, verified in the same harness (3 rows, indentation intact):

```css
[data-copilotkit] [data-streamdown="code-block-body"] > code > span {
  @apply cpk:block;
}
```

`whitespace-pre` isn't needed alongside it.

Also worth knowing: `main` has since gained a scoped streamdown style block covering `code-block`, `code-block-header` and `code-block-body`, and `streamdown-styles.test.ts` is now the established home for assertions of this kind — so the new `globals-css.test.ts` would be a third file doing the same string matching.

Closing this, since the one-line change doesn't move the behavior and the branch has drifted a long way from current `main`. #3330 stays open. Thanks for the careful write-up, and for checking for overlapping PRs before opening — that part was exactly right.

PR #5388, opened 2026-06-11
Add Angular contributors as code owners of the Angular package

mme:
## What

Adds `wolfmanfx` (Murat Sari) and `rainerhahnekamp` (Rainer Hahnekamp) as code owners of the Angular package paths in `.github/CODEOWNERS`:

- `packages/angular/`
- `packages/v2/angular/`

Existing core-dev owners are retained on both lines, so this is purely additive.

## Notes

- **Access is granted separately.** This CODEOWNERS entry only assigns review ownership; it does not grant repo access. The two contributors still need to be added as repo collaborators via GitHub settings for the entry to take effect (GitHub permissions are repo-level, not folder-scoped).
- **Branch protection is currently disabled on `main`**, so this assigns reviewers but does not gate merges until protection requiring code-owner review is enabled.


BenTaylorDev:
Pushed a small commit here (`54989f3`) removing the `packages/v2/angular/` line — the repo uses a flat package layout so there is no `packages/v2/` directory, and that entry would fail CODEOWNERS path validation without ever matching. The `packages/angular/` line is retained and is the correct path.

Also confirmed both `@wolfmanfx` and `@rainerhahnekamp` already have write access, so the ownership takes effect once this merges. Just needs an approving review to unblock.

tylerslaton (approved):

PR #5402, opened 2026-06-11
fix(runtime): recognize nested MCP Apps resource metadata

### LINKED ISSUE #5382: 🐛 Bug: MCPAppsMiddleware does not render FastMCP / prefab-ui app tools — no standard MCP UI rendering protocol
## Summary

`MCPAppsMiddleware` only understands the `@modelcontextprotocol/ext-apps` rendering 
protocol (`_meta.ui.resourceUri`). It cannot render tools from FastMCP servers that 
return `prefab-ui` iframe payloads. This means any Python-native MCP server built 
with `fastmcp[apps]` produces no UI in CopilotKit — the LLM falls back to describing 
the tool result as plain text.

## Environment

- `@copilotkitnext/runtime` (CopilotKit)
- `@ag-ui/mcp-apps-middleware`
- `fastmcp` 3.2.x with `prefab-ui`
- AG-UI / ADK agent backend (Python)

## Steps to Reproduce

1. Create a FastMCP server with a `prefab-ui` app tool:
   ```python
   from fastmcp import FastMCP, FastMCPApp
   from prefab_ui.app import PrefabApp
   from prefab_ui.components import Column, Heading, Form, Input, Button

   app = FastMCPApp("BugReporter")

   @app.tool()
   def submit_bug(title: str, description: str) -> dict:
       return {"saved": True}

   @app.ui()
   def bug_reporter() -> PrefabApp:
       """Open the bug reporter form."""
       with Column(gap=4, css_class="p-6") as view:
           Heading("Bug Reporter")
           with Form(on_submit=CallTool(submit_bug)):
               Input(name="title", label="Title", required=True)
               Input(name="description", label="Description")
               Button("Submit")
       return PrefabApp(view=view, state={})

   mcp = FastMCP("BugReporterServer", providers=[app])
   mcp.run(transport="http", host="0.0.0.0", port=8087)


2. Add the FastMCP server to MCPAppsMiddleware in router.ts:
```js
agent.use(new MCPAppsMiddleware({
  mcpServers: [
    { type: 'http', url: 'http://localhost:8087/mcp', serverId: 'fastmcp' },
  ],
}));
```

3. Ask the LLM to call bug_reporter in the CopilotKit chat.

### ✅ Expected Behavior

The interactive form renders as a UI component inside the chat, similar to how
@modelcontextprotocol/ext-apps tools render HTML resources.

### ❌ Actual Behavior

The LLM receives the raw prefab-ui structured payload but CopilotKit has no
renderer for it. The LLM describes the form fields as a markdown table instead
of rendering any interactive UI.

### Screenshots

_No response_

### 𝌚 CopilotKit Version

```shell
│ ├── @copilotkit/runtime@1.57.1
│ ├── @copilotkit/react-core@1.57.1
│ ├── @copilotkit/react-ui@1.57.1
```

### 📄 Logs (Optional)

```shell

```

rodboev:
## Summary
- Adds support for the nested `_meta.ui.resourceUri` shape used by FastMCP Prefab app tools, alongside the existing flat `_meta["ui/resourceUri"]` key
- Inlines `MCPAppsMiddleware` into the runtime package so the resource-URI extraction logic (`getUIResourceUri`) can be patched without waiting for an upstream `@ag-ui/mcp-apps-middleware` release
- Pending UI tool calls now produce `TOOL_CALL_RESULT` + `ACTIVITY_SNAPSHOT` events with the full `structuredContent` preserved for the iframe renderer

Closes #5382

## Test plan
- [x] `npx nx run @copilotkit/runtime-client-gql:graphql-codegen`
- [x] `pnpm exec oxlint packages/runtime/src/v2/runtime/core/runtime.ts packages/runtime/src/v2/runtime/handlers/shared/mcp-apps-middleware.ts packages/runtime/src/v2/runtime/handlers/shared/agent-utils.ts packages/runtime/src/v2/runtime/__tests__/mcp-apps-middleware-integration.test.ts`
- [x] `pnpm -C packages/runtime exec vitest run src/v2/runtime/__tests__/mcp-apps-middleware-integration.test.ts`
- [x] `pnpm run build`
- [ ] Manual: connect a FastMCP `prefab-ui` server and verify the form renders in the CopilotKit chat

### DIFF
--- a/.changeset/nested-mcp-apps-resource-uri.md
+++ b/.changeset/nested-mcp-apps-resource-uri.md
@@ -1,0 +1,7 @@
+---
+"@copilotkit/runtime": patch
+---
+
+fix(runtime): recognize nested MCP Apps resource metadata
+
+FastMCP Prefab app tools using `_meta.ui.resourceUri` now produce MCP Apps activities with their structured tool result preserved for the iframe renderer.
--- a/examples/integrations/agentcore/infra-cdk/lambdas/copilotkit-runtime/src/runtime.ts
+++ b/examples/integrations/agentcore/infra-cdk/lambdas/copilotkit-runtime/src/runtime.ts
@@ -4,12 +4,12 @@
  */
 import { EventType, HttpAgent } from "@ag-ui/client";
 import type { BaseEvent } from "@ag-ui/client";
-import { MCPAppsMiddleware } from "@ag-ui/mcp-apps-middleware";
 import {
   CopilotKitIntelligence,
   CopilotRuntime,
   createCopilotEndpoint,
   InMemoryAgentRunner,
+  MCPAppsMiddleware,
 } from "@copilotkit/runtime/v2";
 import { concatMap, of } from "rxjs";
 import { randomUUID } from "node:crypto";
--- a/examples/integrations/mcp-apps/app/api/copilotkit/[[...slug]]/route.ts
+++ b/examples/integrations/mcp-apps/app/api/copilotkit/[[...slug]]/route.ts
@@ -3,9 +3,9 @@ import {
   CopilotRuntime,
   InMemoryAgentRunner,
   createCopilotEndpoint,
+  BuiltInAgent,
+  MCPAppsMiddleware,
 } from "@copilotkit/runtime/v2";
-import { BuiltInAgent } from "@copilotkit/runtime/v2";
-import { MCPAppsMiddleware } from "@ag-ui/mcp-apps-middleware";
 import { handle } from "hono/vercel";
 
 const middlewares = [
--- a/examples/showcases/generative-ui-playground/src/app/api/copilotkit/[[...slug]]/route.ts
+++ b/examples/showcases/generative-ui-playground/src/app/api/copilotkit/[[...slug]]/route.ts
@@ -8,14 +8,13 @@
  * The frontend uses the `agent` prop on CopilotKitProvider to select which agent to use.
  */
 
+import { BasicAgent, MCPAppsMiddleware } from "@copilotkit/runtime/v2";
 import {
   CopilotRuntime,
   createCopilotEndpoint,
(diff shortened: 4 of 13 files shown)


BenTaylorDev:
Thanks for digging into this. There's a real observation in here — upstream `@ag-ui/mcp-apps-middleware@0.0.3` only matches the flat `_meta["ui/resourceUri"]` and misses a nested `_meta.ui.resourceUri`. But I don't think this PR should land, for a few reasons:

- **Wrong home for the change.** It forks the ~556-line `@ag-ui/mcp-apps-middleware` into `packages/runtime/.../shared/mcp-apps-middleware.ts` and repoints the runtime + example `route.ts` files at the local copy. AG-UI middleware is authored upstream in `ag-ui-protocol/ag-ui`, not vendored in this repo — an in-runtime fork will silently miss future upstream MCP-Apps fixes. If the nested-`resourceUri` recognition is worthwhile, it belongs upstream with a version bump.
- **Retired-tree debris.** It adds `.changeset/nested-mcp-apps-resource-uri.md`, but the `.changeset/` tree was deleted from `main` (we're on conventional-commit-driven releases now) — that file can't merge.
- **Likely doesn't fix the reported issue.** #5382 is about FastMCP `prefab-ui`, which (per the report) returns inline structured UI payloads in the tool result, not a fetchable HTML resource. This PR only recognizes a nested `resourceUri` *string*, and the frontend renderers all `resources/read` a URL — so a prefab-ui tool with inline content still has nothing to render. The integration test invents the `_meta.ui.resourceUri` shape via a mock rather than a real FastMCP payload, so green CI doesn't demonstrate the fix.

Closing here — but if you can share a real FastMCP `prefab-ui` `_meta`/result payload, that'd help us scope the right fix (upstream recognition + a structured-content renderer). Genuinely appreciate the investigation.

PR #5404, opened 2026-06-12
feat(shell-docs): serve Built-in Agent docs at the docs root

tylerslaton:
## Summary

Built-in Agent docs now render at bare root URLs, such as `/quickstart`, instead of under `/built-in-agent`. Legacy `/built-in-agent/*` and `/integrations/built-in-agent/*` paths continue to redirect to canonical root or backend URLs.

## Why

The Built-in Agent is the default docs surface, so public docs URLs should not expose it as a sub-slug. Existing links still need to keep working, and root-page navigation must stay stable regardless of a visitor's stored framework selection.

## How

- Root docs resolve Built-in Agent authored pages first while preserving reserved routes like `/ag-ui` and framework-prefixed docs.
- Redirect rules in `next.config.ts` and `seo-redirects.ts` point retired Built-in Agent and unselected paths directly at canonical destinations, with regression coverage that prevents redirect destinations from targeting `/built-in-agent`.
- The sidebar/framework provider treats bare URLs as the default Built-in Agent surface instead of letting a stored framework value rewrite root-page chrome.
- MDX link and search-result href rewriting strip retired Built-in Agent prefixes on root-rendered pages while preserving explicit cross-framework links.
- Stale docs links, sitemap, `llms.txt`, markdown exports, and OG resolution now align with the root-served Built-in Agent surface.
- Showcase generated-data tests now serialize shared fixture restoration to avoid concurrent drift in CI.

## Verification

- GitHub checks are green on `f655013dd2576a46dc17b5901eb4dc501cb21028`.
- `npm --prefix showcase/shell-docs run test -- src/lib/__tests__/search-hrefs.test.ts src/lib/__tests__/docs-link-rewrite.test.ts src/lib/__tests__/seo-redirects.test.ts src/lib/__tests__/next-config-redirects.test.ts src/components/__tests__/docs-landing-next.test.tsx 'src/app/[framework]/[[...slug]]/__tests__/framework-root-shell-layout.test.ts'`
- `npm --prefix showcase/shell-docs run test -- --exclude src/app/__tests__/public-assets.test.ts`
- `npm --prefix showcase/shell-docs run typecheck`
- `npm --prefix showcase/shell-docs run lint`
- `npm --prefix showcase/shell-docs run build`
- `npm --prefix showcase/scripts run test`
- `pnpm exec nx run @copilotkit/bot-slack:build --skip-nx-cache --verbose`


samjulien (changes requested):
I think we should fix two edge cases before merging:

1. The `.md/.mdx` export route does not fully mirror the page resolver. In `app/llms-mdx/[[...slug]]/route.ts`, generated frameworks still resolve root MDX before framework override MDX. For `/langgraph-python/quickstart.mdx`, that means we can return the root `quickstart.mdx` routing shim instead of the actual LangGraph quickstart, even though the rendered page route special-cases `quickstart` to prefer `integrations/<framework>/quickstart.mdx`.

2. `resolveDocsHref()` can still framework-prefix legacy redirect aliases. Links such as `/integrations/langgraph/...` or `/ag-ui-protocol` should stay absolute so redirect/canonical rules can handle them, but on a framework-scoped page they can become paths like `/mastra/integrations/langgraph/...` or `/mastra/ag-ui-protocol`.

PR #5418, opened 2026-06-12
fix(angular): emit fresh messages array so OnPush views render on addMessage

### LINKED ISSUE #5416: 🐛 Bug: Messages added via addMessage don't render until the run finishes — AgentStore re-sets the messages signal with the same array reference
♻️ Reproduction Steps
  app-level addMessage:
  
    // main.ts
    import { bootstrapApplication } from '@angular/platform-browser';
    import { Component } from '@angular/core';
    import { provideCopilotKit, CopilotChat } from '@copilotkitnext/angular';
    import { HttpAgent } from '@ag-ui/client';
    import '@copilotkitnext/angular/styles.css';

    @Component({
      selector: 'app-root',
      standalone: true,
      imports: [CopilotChat],
      template: `<copilot-chat [agentId]="'default'" style="display:block;height:100dvh"></copilot-chat>`,
    })
    class App {}

    bootstrapApplication(App, {
      providers: [
        provideCopilotKit({
          agents: { default: new HttpAgent({ url: 'https://<any-ag-ui-endpoint>' }) },
        }),
      ],
    });

  Steps:
  1. Type a message and hit send.
  2. Your own user bubble does not appear on send. It only shows up once the assistant's
     reply arrives — both bubbles flush together at the end of the run.

  The delay is the stock CopilotChat.submitInput() path:

      const userMessage = { id: randomUUID(), role: "user", content: value };
      agent.addMessage(userMessage);
      this.cdr.markForCheck();
      await this.copilotKit.core.runAgent({ agent });

  addMessage is AbstractAgent's base method (push in place), so this reproduces with the
  default component and a plain HttpAgent. The delay is on the user's OWN message and is
  independent of what the agent replies.

  ✅  Expected Behavior
  A message added through agent.addMessage(...) — including the user's own message on
  submit — should render in <copilot-chat> as soon as it's added, not be deferred until the
  run finishes.

  ❌  Actual Behavior
  The message isn't shown until the messages array reference changes, which currently only
  happens when the run pipeline reassigns this.messages to a fresh array (e.g. at run
  completion). No console errors.

  The cause is in AgentStore (packages/angular/src/lib/agent.ts). The store mirrors the
  agent's messages into a signal:

      readonly #messages = signal<Message[]>([]);
      ...
      onMessagesChanged: () => {
        this.#messages.set(abstractAgent.messages);
      }

  AbstractAgent.addMessage in @ag-ui/client mutates the array in place and then notifies
  with the same reference:

      addMessage(m) { this.messages.push(m); /* ...onMessagesChanged({ messages: this.messages }) */ }

  So onMessagesChanged calls set() with the exact array reference the signal already holds.
  Signals compare with Object.is by default, so set(sameRef) is a no-op for change
  propagation: the signal never notifies, the OnPush <copilot-chat> view is never marked
  dirty, and it doesn't re-render. It only updates later because the run pipeline does
  this.messages = <new array>, a new identity — which is why everything appears at run end.

  Forcing change detection doesn't help: the descendant OnPush message view stays clean
  because the signal itself never notified. submitInput() already calls this.cdr.markForCheck()
  right after addMessage and the bubble still doesn't show, so this isn't a missing-tick
  problem — the array reference is the only lever.

  Fix that works for me — copy into a new array so the reference changes:

      onMessagesChanged: () => {
        this.#messages.set([...abstractAgent.messages]);
      }

  A shallow copy is enough — only the array identity needs to change; the Message objects
  can stay referentially stable for trackBy. update() would be equivalent here but buys
  nothing, since the new value is a full replace from abstractAgent.messages, not derived
  from the previous signal value.

  Still present on main: packages/angular/src/lib/agent.ts, AgentStore constructor,
  onMessagesChanged is still this.#messages.set(abstractAgent.messages).

  Screenshots
  N/A — timing issue, nothing visual beyond the delayed bubble.

  𝌚CopilotKit Version
  @copilotkitnext/angular  1.54.3
  @ag-ui/client            0.0.52
  @ag-ui/core              0.0.52
  @angular/core            21.2.15  (mechanism is version-independent — Object.is signal equality, same on v16+)
  node                     v22.20.0

  📄 Logs
  No errors. The SSE stream itself is well-formed:
  RUN_STARTED -> TEXT_MESSAGE_START -> TEXT_MESSAGE_CONTENT (deltas) -> ... -> RUN_FINISHED.

kitos13:
Fixes #5416

## Problem

`AgentStore` mirrored the agent's messages into a signal via `this.#messages.set(abstractAgent.messages)`. `AbstractAgent.addMessage` pushes **in place** and notifies with the **same array reference**. Angular signals compare with `Object.is`, so `set(sameRef)` is a no-op: the signal never notifies, the OnPush `<copilot-chat>` view is never marked dirty, and a freshly added message — including the user's own bubble on submit — does not render until the run pipeline reassigns `messages` to a new array at run completion.

## Fix

```ts
this.#messages.set([...abstractAgent.messages]);
```

A shallow copy gives the array a new identity, so the signal notifies and OnPush views re-render immediately. The `Message` objects stay referentially stable, so `trackBy` still avoids re-creating existing bubbles.

`onStateChanged` is intentionally left unchanged: `AbstractAgent` reassigns `this.state` to a freshly cloned reference on every change (`setState(e) { this.state = clone(e) }`) and never mutates state in place, so that signal already notifies correctly. Scoping the fix to messages keeps it minimal.

## Tests

Added two regression tests in `agent.spec.ts`, both exercising the in-place mutation path that the existing tests never hit (they assign a fresh array each emit):

1. **Reference guard** — the store must not hand back the agent's live `messages` array.
2. **OnPush re-render** — a count rendered in an OnPush *descendant* (a root `detectChanges()` would force-check the root and hide the bug; only a child whose dirty flag depends on the signal exposes it). Two in-place pushes; the second is the same-reference no-op the fix repairs.

Both tests fail without the fix (`'1' to be '2' // Object.is equality`) and pass with it.

## Verification

- `agent.spec.ts`: 6/6 pass
- full `@copilotkitnext/angular` suite: 50/50 pass
- `tsc --noEmit`: clean


onsclom (approved):
Looks great! Thanks for the great fix and regression test here.

kitos13:
Hi @onsclom, how does the package publish process works, when can I expect a new package with this fix included?

PR #5424, opened 2026-06-12
fix(core): keep agent.subscribe callable when detached from the agent instance

### LINKED ISSUE #5000: 🐛 Bug: useAgent: agent.subscribe() throws "Cannot read properties of undefined (reading 'subscribers')" on mount

tylerslaton:
## What does this PR do?

Triage + fix for #5000 (`agent.subscribe()` throws `TypeError: Cannot read properties of undefined (reading 'subscribers')` on mount).

### Triage

The issue's literal repro — `agent.subscribe({...})` called as a method inside a `useEffect` on first render — **does not reproduce**: verified empirically against the published `@copilotkit/react-core@1.57.1` packages with the runtime in every connection state (the provisional `ProxiedCopilotRuntimeAgent` is fully constructed and subscribable synchronously). The `isReady` flag requested in the issue is based on a misdiagnosis — there is no async-initialization race.

The reported error *is* reproduced (byte-for-byte, including the `AbstractAgent.subscribe` stack frame) when `subscribe` is invoked **detached** from the agent:

```ts
const { subscribe } = agent;            // or: useSyncExternalStore(agent.subscribe, ...)
subscribe({ onRunFinalized: ... });     // this === undefined → TypeError
```

`AbstractAgent.subscribe` is a prototype method, and these callback-reference shapes are idiomatic React, so the footgun is easy to hit and the resulting error is hostile.

### Fix

Bind `subscribe` in the `ProxiedCopilotRuntimeAgent` constructor. Every agent CopilotKit itself constructs (provisional agents from the React/Vue/Angular hooks, runtime-discovered agents, `registerProxiedAgent` results, clones) keeps `subscribe` callable without its receiver. Subclass overrides still win — `this.subscribe` resolves through the prototype chain at construction time.

Red/green: both new test files were written first and failed with the exact reported TypeError; the hook-level test also pins the issue's literal repro (direct call on first render) as a regression guard.

### Verification

- `@copilotkit/core`: full suite + new `agent-detached-subscribe.test.ts` ✅
- `@copilotkit/react-core`: 1275/1275 incl. new `use-agent-subscribe.test.tsx` ✅
- `@copilotkit/vue`: 1001/1001 ✅
- `oxfmt --check` / `oxlint` clean; pre-commit gate (tests + publint + attw across 19 projects) green ✅
- `core:check-types` has a pre-existing 393-error baseline on main (extension-less-import TS2835s across every test file); stash-baseline comparison confirms the production change adds zero errors — the one new entry is the new test file's package-conventional import style. The enforced type gate (`core:build` dts emission) passes.

Upstream follow-up: the same footgun exists for any raw `AbstractAgent`/`HttpAgent` the user constructs themselves — the comprehensive fix belongs in `ag-ui`'s `AbstractAgent`.

## Related PRs and Issues

- Fixes #5000

## Checklist

- [ ] If the PR changes or adds functionality, I have updated the relevant documentation (no doc change needed — docs only show receiver-attached `agent.subscribe(...)` calls, which always worked)

### DIFF
--- a/packages/core/src/__tests__/agent-detached-subscribe.test.ts
+++ b/packages/core/src/__tests__/agent-detached-subscribe.test.ts
@@ -1,0 +1,43 @@
+import { describe, it, expect, vi } from "vitest";
+import { ProxiedCopilotRuntimeAgent } from "../agent";
+
+/**
+ * Issue #5000: `agent.subscribe()` is a prototype method, so holding a
+ * detached reference to it (destructuring, passing it as a callback, e.g.
+ * `useSyncExternalStore(agent.subscribe, ...)`) invokes it with
+ * `this === undefined` and throws
+ * "Cannot read properties of undefined (reading 'subscribers')".
+ *
+ * These React patterns are common enough that the agents CopilotKit hands
+ * out must keep `subscribe` callable without its receiver.
+ */
+describe("ProxiedCopilotRuntimeAgent detached subscribe (issue #5000)", () => {
+  function createAgent() {
+    return new ProxiedCopilotRuntimeAgent({
+      runtimeUrl: "https://runtime.example/api",
+      agentId: "test-agent",
+      transport: "rest",
+      runtimeMode: "pending",
+    });
+  }
+
+  it("registers the subscriber when subscribe is invoked detached from the agent", () => {
+    const agent = createAgent();
+    const { subscribe } = agent;
+    const onRunFinalized = vi.fn();
+
+    expect(() => subscribe({ onRunFinalized })).not.toThrow();
+    expect(agent.subscribers).toContainEqual({ onRunFinalized });
+  });
+
+  it("unsubscribe returned by a detached subscribe call removes the subscriber", () => {
+    const agent = createAgent();
+    const { subscribe } = agent;
+    const subscriber = { onRunFinalized: vi.fn() };
+
+    const subscription = subscribe(subscriber);
+    subscription.unsubscribe();
+
+    expect(agent.subscribers).not.toContain(subscriber);
+  });
+});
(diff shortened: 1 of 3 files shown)


mme (approved):

BenTaylorDev:
Checked with Tyler - closed as defunct.

PR #5436, opened 2026-06-13
feat(angular): add React v2 run-control parity to CopilotChat (#5428)

### LINKED ISSUE #5428: 🚀 Feature: @copilotkitnext/angular — Stop button + serialize next send (parity with React v2)
## Summary

The Angular package (`@copilotkitnext/angular`) is missing the run-control behavior that the React v2 components (`@copilotkit/react-core`, `src/v2`) already implement. While an agent run is in flight, the React v2 chat:

- Turns the **Send** button into a **Stop** button; clicking it aborts the run (`core.stopAgent({ agent })`).
- **Enter on an empty composer** stops the run; **Enter with text** sends a new message even mid-run.
- **Serializes** that new send behind the active run — it awaits the in-flight run's completion (`activeRunCompletionPromise`) before dispatching, rather than pre-empting it.
- Keeps the textbox **enabled** while running.

Angular currently has none of this: there is no stop affordance, Enter always sends, the send button merely disables, and `CopilotChat.submitInput` calls `runAgent` immediately while a run is in flight — which pre-empts/aborts the in-flight run.

## Why serialize instead of pre-empt

Pre-empting an in-flight run is the consecutive-interrupt regression fixed for React in #5195: aborting an interrupt **resume** leaves the agent graph paused, so the next message re-interrupts the same paused graph instead of starting a fresh turn. Awaiting the active run's completion serializes the two turns so the resume finishes before the new message starts a fresh run.

## Proposed scope (Angular, parity with React v2)

1. Stop button replaces Send while running (click → `core.stopAgent({ agent })`, fallback `agent.abortRun()`).
2. Enter-on-empty-composer stops; Enter-with-text sends (queued behind the active run).
3. `CopilotChat.submitInput` awaits the in-flight run's completion before adding the new message and running.
4. Textbox stays enabled while running.

Tests will mirror the existing React v2 coverage (`CopilotChatInput.test.tsx`, `CopilotChat.e2e.test.tsx`).

## Affected files

- `packages/angular/src/lib/chat-state.ts`
- `packages/angular/src/lib/components/chat/copilot-chat.ts`
- `packages/angular/src/lib/components/chat/copilot-chat-input.ts`

## Environment

- `@copilotkitnext/angular` (v2 line)

jdgajjar:
Mirrors the full React v2 CopilotChat run-control surface:

## Stop button while a run is active
- CopilotChatInput now imports the Square (stop) icon from lucide-angular.
- When isProcessing() is true the send button renders a Square icon and routes clicks to triggerStop() instead of send(), regardless of composer contents.  This matches React v2 handleSendButtonClick().
- A new stop output() is emitted so parent templates can react directly.

## Enter-on-empty-input stops the run
- handleKeyDown now mirrors React v2 exactly: if (isProcessing && !canSend)  →  stop
    else                           →  send
  A non-empty composer always sends even mid-run (unblocks consecutive
  interrupt pills, #5195).  Stop is only reachable via Enter when the
  composer is empty — the genuine 'cancel' affordance.

## Serialization of new sends behind activeRunCompletionPromise
- CopilotChat.submitInput now calls waitForActiveRunToSettle() before dispatching the user message.  This awaits agent.activeRunCompletionPromise (via the isRunCompletionAware() type guard from @copilotkit/core) so an interrupt-resume can finish before the next runAgent() call begins. Agents that don't implement RunCompletionAware degrade safely (await is skipped).

## stopCurrentRun with stopAgent / abortRun fallback
- CopilotChat.stopCurrentRun() calls core.stopAgent({ agent }) first (which also aborts in-flight tool execution via the run-handler's abort controller) and falls back to agent.abortRun() on failure, mirroring React v2.

## Keep input editable while running
- CopilotChatInput.textAreaContext now sets disabled = mode === 'processing' (transcription in progress) instead of also disabling on isRunning. The textarea stays editable mid-run so users can compose the next message while the agent is still streaming.

## ChatState interface extended
- Added optional isRunning: Signal<boolean> and stopCurrentRun?(): void to the abstract ChatState so CopilotChatInput can consume them without a direct dependency on CopilotChat.  Both are optional for backwards compatibility with custom ChatState implementations.

## Angular tests
- Expanded copilot-chat-input.component.spec.ts with:
    * isProcessing / canSend / canStop computed signal assertions
    * Enter-vs-button routing contract (mirrors React v2 test suite):
        - Enter with sendable text mid-run → SEND (not stop)
        - Enter with empty composer mid-run → STOP - Button click mid-run → STOP regardless of composer contents - Shift+Enter → newline (no send/stop)
    * stopCurrentRun wiring and graceful degradation
    * Textarea disabled state (enabled while running, disabled in processing)
    * waitForActiveRunToSettle serialization proof (unit-level)

Closes #5428

## What does this PR do?

## Related PRs and Issues

## Checklist

### DIFF
--- a/packages/angular/src/lib/chat-state.ts
+++ b/packages/angular/src/lib/chat-state.ts
@@ -1,11 +1,24 @@
-import { inject, Injectable, WritableSignal } from "@angular/core";
+import { inject, Injectable, Signal, WritableSignal } from "@angular/core";
 
 @Injectable()
 export abstract class ChatState {
   abstract readonly inputValue: WritableSignal<string>;
 
-  abstract submitInput(value: string): void;
+  abstract submitInput(value: string): void | Promise<void>;
   abstract changeInput(value: string): void;
+
+  /**
+   * Whether the agent is currently running.
+   * Optional — components that don't manage a run fall back to `undefined`.
+   */
+  readonly isRunning?: Signal<boolean>;
+
+  /**
+   * Stop the currently active run.
+   * Optional — only provided when the chat is wired to a run-aware agent
+   * (e.g. `CopilotChat`).
+   */
+  stopCurrentRun?(): void;
 }
 
 export function injectChatState(): ChatState {
--- a/packages/angular/src/lib/chat-state.ts
+++ b/packages/angular/src/lib/chat-state.ts
@@ -14,6 +14,19 @@ export abstract class ChatState {
   readonly isRunning?: Signal<boolean>;
 
   /**
+   * Whether a stop action is currently meaningful.
+   *
+   * Mirrors React v2 `shouldAllowStop = agent.isRunning && hasMessages`:
+   * the stop affordance is only active when a run is in flight AND the thread
+   * already has at least one message (so clicking Stop on the welcome screen
+   * before any message is sent does nothing).
+   *
+   * Optional — when absent, `CopilotChatInput` falls back to checking whether
+   * `stopCurrentRun` is defined (the pre-#5428 behaviour).
+   */
+  readonly canStopRun?: Signal<boolean>;
+
+  /**
    * Stop the currently active run.
    * Optional — only provided when the chat is wired to a run-aware agent
    * (e.g. `CopilotChat`).
--- a/packages/angular/src/lib/components/chat/__tests__/copilot-chat-input.component.spec.ts
+++ b/packages/angular/src/lib/components/chat/__tests__/copilot-chat-input.component.spec.ts
@@ -20,12 +20,13 @@ class ChatStateStub extends ChatState {
 }
 
 // ---------------------------------------------------------------------------
-// Stub that exposes isRunning + stopCurrentRun (mirrors CopilotChat)
+// Stub that exposes isRunning + stopCurrentRun + canStopRun (mirrors CopilotChat)
 // ---------------------------------------------------------------------------
 @Injectable()
 class RunAwareChatStateStub extends ChatState {
   inputValue = signal("");
   readonly isRunning = signal<boolean>(false);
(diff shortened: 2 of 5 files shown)


jdgajjar:
This PR aims to follow the existing React v2 implementation as closely as possible rather than introducing new behavior.

The primary goal is to bring Angular parity with:

* Stop button support
* Run cancellation behavior
* Send serialization via `activeRunCompletionPromise`
* Editable composer while streaming

I've also added tests covering the new run-control behavior. Happy to adjust the implementation if there are preferred Angular patterns or existing abstractions I should use instead.

BenTaylorDev:
Thanks for this, and apologies it sat — the run-control parity you're adding (Stop button + serialize-next-send) is genuinely still missing from the Angular package, so **issue #5428 stays open**. Unfortunately this branch can't be merged or rebased as-is: it's ~2,200 commits behind, and the Angular package has been substantially rewritten since it was opened. Specifically:

- The package scope was renamed `@copilotkitnext/angular` → `@copilotkit/angular`.
- The stop icon here imports `Square` from `lucide-angular`, but `lucide-angular` was removed — icons now come from the internal `../icons/copilot-icon`.
- `ChatState` was rewritten (new `attachmentsEnabled`/`attachmentsUploading` signals; `submitInput` now returns `void`), which conflicts with the changes here.
- All four edited files hard-conflict against current `main`.

So I'm closing this in favor of a fresh implementation against the current `@copilotkit/angular` (stop must be reimplemented on the internal icon system, and `chat-state.ts` reconciled with the new attachments signals). Really appreciate the work — the design intent is right, it just needs to be re-targeted at the current tree. I'll keep #5428 open to track it.

PR #5440, opened 2026-06-14
fix(core): execute frontend tools when backend returns placeholder result

### LINKED ISSUE #3442: 🐛 Bug: useHumanIntheLoop not executing frontend tool after confirmation

rodboev:
## Summary

Fixes the remote-agent HITL case where a frontend tool is skipped because the backend returns the placeholder result `Forwarded to client`.

The recovery is now marker-based. `processAgentResult` only removes an existing tool result when the content normalizes to that exact placeholder and there is an executable frontend handler on the current path. Every other backend result stays authoritative and blocks frontend re-execution.

## Implementation

- `packages/core/src/core/run-handler.ts` keeps the current `executeFrontendTools` and follow-up flow, then replaces the old existence-only check with exact placeholder detection at https://github.com/CopilotKit/CopilotKit/blob/7449f6d790dbe80ab8e0950f66b04a9bf4acde2c/packages/core/src/core/run-handler.ts#L484-L599
- the placeholder is removed from both `newMessages` and `agent.messages` before the existing specific or wildcard execution path inserts the real result
- `.changeset/fix-frontend-tool-placeholder.md` is dropped because current `main` no longer carries a `.changeset` tree

## Regression coverage

- exact placeholder replacement for a named frontend tool: https://github.com/CopilotKit/CopilotKit/blob/7449f6d790dbe80ab8e0950f66b04a9bf4acde2c/packages/core/src/__tests__/core-frontend-tool-placeholder.test.ts#L22-L51
- placeholder passthrough when the tool has no handler: https://github.com/CopilotKit/CopilotKit/blob/7449f6d790dbe80ab8e0950f66b04a9bf4acde2c/packages/core/src/__tests__/core-frontend-tool-placeholder.test.ts#L53-L82
- genuine backend-result preservation for a named frontend tool: https://github.com/CopilotKit/CopilotKit/blob/7449f6d790dbe80ab8e0950f66b04a9bf4acde2c/packages/core/src/__tests__/core-frontend-tool-placeholder.test.ts#L85-L109
- existing edge-case guard for a non-placeholder result: https://github.com/CopilotKit/CopilotKit/blob/7449f6d790dbe80ab8e0950f66b04a9bf4acde2c/packages/core/src/__tests__/core-edge-cases.test.ts#L23-L54
- existing wildcard execution still works with no pre-existing result: https://github.com/CopilotKit/CopilotKit/blob/7449f6d790dbe80ab8e0950f66b04a9bf4acde2c/packages/core/src/__tests__/core-frontend-tool-placeholder.test.ts#L111-L154
- wildcard placeholder replacement also stays on the exact marker path: https://github.com/CopilotKit/CopilotKit/blob/7449f6d790dbe80ab8e0950f66b04a9bf4acde2c/packages/core/src/__tests__/core-frontend-tool-placeholder.test.ts#L156-L187
- the recursive follow-up starts with the assistant tool call intact and exactly one real tool message for the same `toolCallId`: https://github.com/CopilotKit/CopilotKit/blob/7449f6d790dbe80ab8e0950f66b04a9bf4acde2c/packages/core/src/__tests__/core-frontend-tool-placeholder.test.ts#L189-L250

Closes #3442.

## Validation

- [x] `pnpm exec oxfmt --check packages/core/src/core/run-handler.ts packages/core/src/__tests__/core-edge-cases.test.ts packages/core/src/__tests__/core-frontend-tool-placeholder.test.ts`
- [x] `pnpm exec oxlint packages/core/src/core/run-handler.ts packages/core/src/__tests__/core-edge-cases.test.ts packages/core/src/__tests__/core-frontend-tool-placeholder.test.ts`
- [x] `pnpm -C packages/shared exec tsdown`
- [x] `pnpm -C packages/core exec tsdown`
- [x] `pnpm -C packages/core exec vitest run src/__tests__/core-frontend-tool-placeholder.test.ts src/__tests__/core-edge-cases.test.ts` 17/17
- [x] `pnpm -C packages/core exec vitest run` 54 files, 590 tests
- [ ] CI green on the rebased head

### DIFF
--- a/.changeset/fix-frontend-tool-placeholder.md
+++ b/.changeset/fix-frontend-tool-placeholder.md
@@ -1,0 +1,7 @@
+---
+"@copilotkit/core": patch
+---
+
+fix(core): execute frontend tools when backend returns placeholder result
+
+Frontend tools registered via `useFrontendTool` now correctly execute after HITL approval with remote agents (Strands, LangGraph, CrewAI) instead of being silently skipped by the backend's placeholder result.
--- a/.changeset/fix-frontend-tool-placeholder.md
+++ b/.changeset/fix-frontend-tool-placeholder.md
@@ -1,7 +1,0 @@
----
-"@copilotkit/core": patch
----
-
-fix(core): execute frontend tools when backend returns placeholder result
-
-Frontend tools registered via `useFrontendTool` now correctly execute after HITL approval with remote agents (Strands, LangGraph, CrewAI) instead of being silently skipped by the backend's placeholder result.
--- a/packages/core/src/__tests__/core-edge-cases.test.ts
+++ b/packages/core/src/__tests__/core-edge-cases.test.ts
@@ -20,10 +20,10 @@ describe("CopilotKitCore.runAgent - Edge Cases", () => {
     vi.restoreAllMocks();
   });
 
-  it("should skip tool call when result already exists in newMessages", async () => {
+  it("should replace placeholder and execute handler when tool has a handler and result already exists", async () => {
     const tool = createTool({
       name: "alreadyProcessedTool",
-      handler: vi.fn(async () => "Should not be called"),
+      handler: vi.fn(async () => "Real result"),
     });
     copilotKitCore.addTool(tool);
 
@@ -51,7 +51,8 @@ describe("CopilotKitCore.runAgent - Edge Cases", () => {
 
     await copilotKitCore.runAgent({ agent: agent as any });
 
-    expect(tool.handler).not.toHaveBeenCalled();
+    // With the placeholder fix, a tool with a handler replaces the placeholder
+    expect(tool.handler).toHaveBeenCalled();
   });
 
   it("should handle empty tool function name", async () => {
--- a/packages/core/src/__tests__/core-edge-cases.test.ts
+++ b/packages/core/src/__tests__/core-edge-cases.test.ts
@@ -20,7 +20,7 @@ describe("CopilotKitCore.runAgent - Edge Cases", () => {
     vi.restoreAllMocks();
   });
 
-  it("should replace placeholder and execute handler when tool has a handler and result already exists", async () => {
+  it("should preserve non-placeholder results when tool has a handler and result already exists", async () => {
     const tool = createTool({
       name: "alreadyProcessedTool",
       handler: vi.fn(async () => "Real result"),
@@ -51,8 +51,13 @@ describe("CopilotKitCore.runAgent - Edge Cases", () => {
 
     await copilotKitCore.runAgent({ agent: agent as any });
 
-    // With the placeholder fix, a tool with a handler replaces the placeholder
-    expect(tool.handler).toHaveBeenCalled();
+    expect(tool.handler).not.toHaveBeenCalled();
+
+    const toolMessages = agent.messages.filter(
+      (m) => m.role === "tool" && m.toolCallId === toolCallId,
+    );
+    expect(toolMessages).toHaveLength(1);
+    expect((toolMessages[0] as any).content).toBe("Already processed");
   });
 
   it("should handle empty tool function name", async () => {
--- a/packages/core/src/__tests__/core-frontend-tool-placeholder.test.ts
+++ b/packages/core/src/__tests__/core-frontend-tool-placeholder.test.ts
@@ -1,0 +1,129 @@
+import { describe, it, expect, beforeEach, vi, afterEach } from "vitest";
+import { CopilotKitCore } from "../core";
+import {
+  MockAgent,
+  createToolCallMessage,
+  createToolResultMessage,
+  createTool,
+} from "./test-utils";
+
+describe("CopilotKitCore - Frontend Tool Placeholder (remote agent HITL)", () => {
+  let copilotKitCore: CopilotKitCore;
+
+  beforeEach(() => {
+    copilotKitCore = new CopilotKitCore({});
+    vi.clearAllMocks();
+  });
+
+  afterEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  it("should execute handler and replace placeholder when backend returns a placeholder result", async () => {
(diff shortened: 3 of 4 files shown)


BenTaylorDev (changes requested):
Thanks — #3442 is a real, still-live bug (the `newMessages.findIndex(m => m.role === "tool" && m.toolCallId === id) === -1` guard skips the registered frontend handler when a backend placeholder result is present). Two things before it can land:

- **The "placeholder detection" isn't content-based.** The condition is effectively "a frontend tool with a handler is registered AND a tool result already exists → discard it and re-run the handler" — it never inspects the result to confirm it's actually a placeholder. So any tool that's both registered on the frontend *and* legitimately executed by the backend would have its genuine result spliced out and re-executed client-side (double-execution / semantic change). Please key the recovery on an actual placeholder marker, not just "handler + result exists."
- **Un-mergeable + debris.** It hard-conflicts against the rewritten `processAgentResult` (now gated by `executeFrontendTools`, driving recursive follow-ups with the `MAX_FOLLOW_UP_DEPTH` breaker, inserting the real result into `agent.messages`), and it adds a `.changeset/` file to a repo that no longer uses changesets. Re-target onto current `main` and drop the changeset.

After re-targeting, please confirm that removing the placeholder leaves exactly one tool message per `toolCallId` going into the follow-up run (no orphaned tool_call, no duplicate ToolMessage).

rodboev:
Thanks for the detailed review. I reworked the fix around the current `processAgentResult` flow.

1. Placeholder recovery now keys on the exact `Forwarded to client` marker and only reruns when there is an executable frontend handler on the current path. Genuine backend results stay untouched and do not trigger frontend re-execution.
2. The placeholder is removed from both the returned messages and the agent message history before the existing specific or wildcard execution path inserts the real result.
3. The follow-up regression now checks the second run directly and asserts that the assistant tool call is still present and there is exactly one real tool message for that `toolCallId`.
4. The branch is rebased onto current `main`, keeps the existing `executeFrontendTools` and follow-up depth protections, and drops the stale `.changeset` file.

Local validation on the rebased head is green:

- `pnpm exec oxfmt --check packages/core/src/core/run-handler.ts packages/core/src/__tests__/core-edge-cases.test.ts packages/core/src/__tests__/core-frontend-tool-placeholder.test.ts`
- `pnpm exec oxlint packages/core/src/core/run-handler.ts packages/core/src/__tests__/core-edge-cases.test.ts packages/core/src/__tests__/core-frontend-tool-placeholder.test.ts`
- `pnpm -C packages/shared exec tsdown`
- `pnpm -C packages/core exec tsdown`
- `pnpm -C packages/core exec vitest run src/__tests__/core-frontend-tool-placeholder.test.ts src/__tests__/core-edge-cases.test.ts`
- `pnpm -C packages/core exec vitest run`

BenTaylorDev (approved):
**Re-approving** — the placeholder recovery is now content-based (`isFrontendPlaceholderResult` matches the exact `"Forwarded to client"` marker from the issue's own trace), so a genuine backend result is preserved rather than re-executed, and the `.changeset` file is gone. Verified locally on the updated branch against current `main`: `@copilotkit/core` tests 593/593, `tsc --noEmit` clean, lint 0 errors. One non-blocking note: recovery keys on that exact literal string, so a differently-worded placeholder wouldn't be recovered — fine for closing #3442, worth noting as a known limitation.

PR #5441, opened 2026-06-14

deepshekhardas:
Port of #5433


BenTaylorDev:
Closing as a duplicate of #5433, which is the canonical version of this fix (identical diff, and it links the issue it resolves). Let's keep the discussion there.

PR #5435, opened 2026-06-13
fix(react-core): allow self-managed agents without runtime

Tushar-Khandelwal-2004:
## Summary

Fixes #5417 by allowing the public `<CopilotKit />` wrapper to accept `selfManagedAgents` without requiring `runtimeUrl`, `publicApiKey`, or `publicLicenseKey`.

## Changes

-> Updated the legacy wrapper validation so local/self-managed agents count as a valid CopilotKit configuration source.
-> Added regression coverage for the reported `/v2` usage path: `<CopilotKit selfManagedAgents={{ testAgent }} agent="testAgent">`.
-> Verified the supplied self-managed agent is registered and resolved without needing a dummy runtime URL.

## Testing

-> `pnpm -C packages/react-core exec vitest run src/components/copilot-provider/__tests__/v1-explicit-threadid-bridge.test.tsx --reporter verbose`

## Note

I saw the issue discussion also mentions a docs update. This PR focuses on the code path from the reported repro, where the `/v2` `CopilotKit` export still goes through wrapper validation before the v2 provider can use `selfManagedAgents`.

Happy to adjust or follow up with docs if preferred.

PR #5442, opened 2026-06-14
fix(react-core): allow self-managed agents without runtime

deepshekhardas:
Port of #5435


BenTaylorDev:
Closing as a duplicate of #5435 (this PR is a byte-identical port), which is itself now superseded — the `selfManagedAgents`-without-runtime fix already landed on `main` in `9907519989`, including a regression test. Thanks @deepshekhardas; no action needed here.

PR #5443, opened 2026-06-14
fix(runtime-client-gql): guard abort errors without message

deepshekhardas:
Port of #5437


BenTaylorDev:
Closing as a duplicate of #5437, which is the canonical version of this fix (identical diff, and it links the issue it resolves). Let's keep the discussion there.

PR #5444, opened 2026-06-14
fix(angular): serialize sends while agent is running

### LINKED ISSUE #5428: 🚀 Feature: @copilotkitnext/angular — Stop button + serialize next send (parity with React v2)

Tushar-Khandelwal-2004:
## Summary

Fix Angular CopilotChat run-control parity with React v2.

This updates the Angular chat input so an active agent run shows a stop affordance, keeps the textarea editable while streaming, and routes run-time actions correctly:
-> clicking the running button stops the active run
-> pressing Enter with empty input stops the active run
-> pressing Enter with text sends the next message instead of stopping

The parent `CopilotChat` now also serializes new sends behind the active run completion promise before adding the next user message or calling `runAgent`, preventing a new turn from pre-empting an interrupt resume.

## Changes

-> Added optional run-control APIs to `ChatState`
-> Wired Angular `CopilotChatInput` to render/send/stop like React v2
-> Added `CopilotChat.stopCurrentRun()` with `core.stopAgent()` and `abortRun()` fallback
-> Added active-run send serialization in `CopilotChat.submitInput()`
-> Added input behavior tests and parent-level `CopilotChat` tests for serialization and stop fallback

## Testing

-> `oxlint` passed on all touched files
-> `git diff --check` passed
-> `tsc -p packages/angular/tsconfig.spec.json --noEmit` reports unrelated existing Angular spec errors, but no errors from this patch
-> Vitest/Nx Angular test execution hung locally on Windows, including existing specs

## Related Issue

Closes #5428

### DIFF
--- a/packages/angular/src/lib/chat-state.ts
+++ b/packages/angular/src/lib/chat-state.ts
@@ -1,11 +1,14 @@
-import { inject, Injectable, WritableSignal } from "@angular/core";
+import { inject, Injectable, Signal, WritableSignal } from "@angular/core";
 
 @Injectable()
 export abstract class ChatState {
   abstract readonly inputValue: WritableSignal<string>;
 
-  abstract submitInput(value: string): void;
+  abstract submitInput(value: string): void | Promise<void>;
   abstract changeInput(value: string): void;
+
+  readonly isRunning?: Signal<boolean>;
+  stopCurrentRun?(): void;
 }
 
 export function injectChatState(): ChatState {
--- a/packages/angular/src/lib/components/chat/__tests__/copilot-chat-input.component.spec.ts
+++ b/packages/angular/src/lib/components/chat/__tests__/copilot-chat-input.component.spec.ts
@@ -4,6 +4,7 @@ import {
   runInInjectionContext,
   signal,
 } from "@angular/core";
+import type { Signal } from "@angular/core";
 import { TestBed } from "@angular/core/testing";
 import { beforeEach, describe, expect, it, vi } from "vitest";
 import { CopilotChatInput } from "../copilot-chat-input";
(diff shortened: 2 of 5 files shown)


BenTaylorDev:
Thanks @Tushar-Khandelwal-2004. We're consolidating the Angular run-serialization work (#5428) onto #5436, which covers the same five files and additionally uses the public `isRunCompletionAware` guard from `@copilotkit/core` plus a `hasMessages` stop-guard (so Stop doesn't appear on the empty welcome screen — matching React v2's `shouldAllowStop`). Closing in favor of #5436. Appreciate the contribution.

PR #5448, opened 2026-06-15
fix(inspector): persist announcement popout dismissal via X control

BenTaylorDev:
## Problem

Users reported that the announcement banner popping out of the inspector's floating icon **does not stay closed after dismissing it**.

Confirmed in a real browser: the popout reappears on every reload. Root cause — the popout preview bubble (`renderAnnouncementPreview`) had **no dismiss control of its own**. Its only hide paths (`handleAnnouncementPreviewClick`, `openInspector`) clear an in-memory flag without persisting. On every mount `fetchAnnouncement()` recomputes `showAnnouncementPreview` from the stored timestamp — which only the *in-window* banner X (`markAnnouncementSeen`) ever wrote. So unless the user opened the inspector and dismissed the inner banner, the popout came back on every load.

## Fix

Add an X / dismiss control directly to the popout bubble:
- Click (or Enter/Space) → `markAnnouncementSeen()`, which **persists** the announcement timestamp to `localStorage`.
- `event.stopPropagation()` so the X does not bubble up to the preview body / floating button (i.e. it dismisses without opening the inspector).
- Implemented as a `role="button"` span (not a `<button>`): the popout renders *inside* the floating `<button>`, so a nested `<button>` would be invalid HTML. Includes keyboard support and `:focus-visible` styling.

Body-click behavior is unchanged: clicking the bubble itself still opens the inspector (engagement), and intentionally does **not** persist, so the in-window banner still surfaces the announcement.

## Verification

- **Live browser:** popout shows X → click → `localStorage` persists `{"timestamp":...}`, inspector stays closed → **reload keeps it gone**.
- **Unit tests (3 new, full suite 32/32 green):** X persists timestamp & hides bubble; X does not open the inspector; body-click still opens *without* persisting.
- Format (`oxfmt`), lint (`oxlint`), and build all clean; pre-commit full-suite passed.

### DIFF
--- a/packages/web-inspector/src/__tests__/web-inspector.spec.ts
+++ b/packages/web-inspector/src/__tests__/web-inspector.spec.ts
@@ -1,9 +1,7 @@
 import { WebInspectorElement, ɵCpkThreadDetails } from "../index";
-import {
-  CopilotKitCore,
-  CopilotKitCoreRuntimeConnectionStatus,
-  type CopilotKitCoreSubscriber,
-} from "@copilotkit/core";
+import type { CopilotKitCore } from "@copilotkit/core";
+import { CopilotKitCoreRuntimeConnectionStatus } from "@copilotkit/core";
+import type { CopilotKitCoreSubscriber } from "@copilotkit/core";
 import type { AbstractAgent, AgentSubscriber } from "@ag-ui/client";
 import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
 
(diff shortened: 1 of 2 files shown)


tylerslaton:
I’d request one change before approving this.

The timestamp behavior looks correct: dismissing stores only the current announcement timestamp, so a future announcement with a new timestamp should still show.

The issue is the new dismiss control in the popout. The PR avoids nesting a real `<button>` inside the floating inspector `<button>`, but the replacement is still a focusable interactive element inside that same native button:

```html
<span
  class="announcement-preview__dismiss"
  role="button"
  tabindex="0"
>
```

That still violates the HTML button content model: a `<button>` cannot contain interactive descendants, and it also cannot contain descendants with `tabindex` specified.

I’d fix this by rendering the announcement preview as a sibling/overlay next to the floating inspector button instead of as a child of it. Then the dismiss affordance can be a normal `<button type="button">`, which keeps the DOM valid and makes the accessibility behavior much cleaner.

BenTaylorDev:
Good catch — fixed in a01d846.

The preview now renders as a **sibling** of the floating button inside a `position: relative` wrapper (`.console-button-wrapper`), rather than as a child of the `<button>`. With it out of the button's content model, the dismiss affordance is back to a real `<button type="button">` — no `role="button"`, no `tabindex`, no manual Enter/Space keydown shim (native button handling covers it). `stopPropagation` is kept so the X doesn't bubble to the preview body, whose click opens the inspector.

Verified live in a browser:
- DOM: preview is no longer inside the floating button; dismiss is `<button type="button">` with no `tabindex`/`role`.
- Positioning is byte-for-byte the same (the wrapper has the button's footprint, so the absolute-positioned bubble anchors to the same edge).
- Behavior unchanged: X persists the timestamp + hides the bubble without opening the inspector; body-click opens without persisting. web-inspector unit suite 32/32 green.

tylerslaton (approved):

PR #5450, opened 2026-06-15
ci(test-unit): scope build+test to nx affected packages

BenTaylorDev:
## What & why

`test / unit` ran `pnpm build` + `pnpm test` over **every** package, **3× across the Node 20/22/24 matrix**, regardless of what the PR touched. This scopes it to `nx affected`. Nx Cloud is intentionally off here, so this is Nx-Cloud-free.

## Change (`test_unit.yml`)
- `fetch-depth: 0` so affected has a merge-base to diff against.
- Derive `NX_BASE`/`NX_HEAD`: PR → merge-base with the base branch tip (`FETCH_HEAD`); push → `github.event.before` with a `HEAD~1` fallback.
- Compute the package list with **`nx show projects --affected --projects='packages/**'`** and feed it to `run-many`.
  - ⚠️ The `nx affected` **run** form ignores `--projects` and pulls in downstream `examples`/`storybook`; only `nx show projects` honours the scope. Hence the show-projects → run-many split.
- `workflow_dispatch` still runs **all** packages (manual/full run).
- GitHub context passed via `env:` (not inline `${{ }}` in run blocks) to satisfy zizmor template-injection; `NX_VERBOSE_LOGGING` forced off for the JSON-parsing step.

## Measured impact (CI)
| metric | OLD (build+test all) | NEW (affected) | Δ |
|---|---|---|---|
| wall-clock (longest leg) | 4m25s | 2m23s | −46% |
| runner-min (3 legs) | 12.0m | 5.75m | −52% |

Floor shown (this PR touches only workflow files → empty affected → skip). Saving scales with blast radius; most PRs are narrow.

## Dropped from this PR
- **e2e-dojo build-once dedup**: cut runner-minutes ~8% but **regressed wall-clock ~12%** (the shared `build-cpk` job became a serial prefix every suite waits on). Reverted — wall-clock is the priority. Investigating a non-serializing way to cut dojo wall-clock separately.
- **Vitest transform/env tax**: no safe config win (`--no-isolate` breaks 315 tests, `--pool=threads` is a wash, `deps.optimizer` is 3×+ faster but breaks 45 files via `vi.mock`). Needs a test-infra refactor; `nx affected` already stops these suites running on every PR.


tylerslaton:
Can we validate one more case before merging?

The current CI run for this PR did not run the new package build/test step because no packages were selected.

Please either:
1. run this workflow manually on the branch so it tests all packages, or
2. update the workflow so changes to `test_unit.yml` run all packages automatically.

That would confirm the new build/test command works when packages are selected.

BenTaylorDev:
Went with **option 2**, @tylerslaton: the workflow now forces a full all-packages run whenever `test_unit.yml` itself changes in the diff range (same path as a manual `workflow_dispatch`), so the `run-many` build/test command gets exercised on the very PR that touches it. `nx affected` can't surface a workflow edit as an affected package, hence the explicit self-edit check.

This push therefore selects **all** packages, so this CI run validates the build/test path with packages selected.

Note: I rebuilt the branch as a single clean commit on top of current `main` (the old history carried a build-once dojo experiment + its revert that collided with #5452 on rebase). Net diff is unchanged — `test_unit.yml` only — and it's now cleanly rebaseable. Rebase auto-merge is enabled.

tylerslaton (approved):

PR #5457, opened 2026-06-15
fix(examples): bump integrations to @copilotkit 1.60.1 for @ag-ui fetch fix

mxmzb:
## Summary

Fixes the integration examples whose chat never responds and logs this on **every agent run**:

```
TypeError: Failed to execute 'fetch' on 'Window': Illegal invocation
  (agent_run_failed_event){ source: onRunFailed, agentId: default }
```

The fix is a dependency bump that lands every integration demo on `@copilotkit/* 1.60.1` + `@ag-ui/client 0.0.57`.

## Root cause

`@ag-ui/client@0.0.56` (and `0.0.55`/`0.0.53`) `HttpAgent` stores the global `fetch` **unbound** and later calls it as a method:

```ts
// buggy
this.fetch = config.fetch ?? fetch;
run(input) { return runHttpRequest(() => this.fetch(this.url, this.requestInit(input))); }
```

`this.fetch(...)` invokes native `fetch` with the agent instance as the receiver instead of `window`. Native `fetch` is brand-checked and throws `Illegal invocation`. It fires on the client run path (`ProxiedCopilotRuntimeAgent.run` → `#runViaHttp`) in **both dev and prod** — not environment-specific.

Fixed upstream in `@ag-ui/client@0.0.57`:

```ts
this.fetch = config.fetch ?? ((url, requestInit) => fetch(url, requestInit));
```

`@copilotkit/* 1.60.1` (published) bumped its pin to `@ag-ui/client@0.0.57`. Examples still on `1.60.0` ship the buggy client — which is why deployments fail.

## What changed (all `examples/integrations/*`)

- **`@copilotkit/* 1.60.0 → 1.60.1`** across every integration demo. The `_parity` manifest requires all demos to track the north-star (`langgraph-python`) `@copilotkit` version, so they move in lockstep — a partial bump fails `parity-check`.
- **Starter-fleet apps force a single `@ag-ui` tree via `overrides`** (`@ag-ui/client|core|encoder|proto`). Bumped that single-tree pin **`0.0.53`/`0.0.55` → `0.0.57`** so they actually carry the fetch fix instead of a `1.60.1 ↔ 0.0.55` skew. Adapter packages (`@ag-ui/crewai`, `@ag-ui/mastra`, `@ag-ui/llamaindex`, `@ag-ui/a2a`, middlewares) are intentionally left at their own versions.
- Lockfiles regenerated; all resolve `@ag-ui/client 0.0.57`.

## Verification

- [x] Reproduced `Illegal invocation` locally on `1.60.0` (dev + prod build)
- [x] Verified `langgraph-python` on `1.60.1`: agent run completes, todo created, **0 console errors**
- [x] All lockfiles resolve `@ag-ui/client 0.0.57`
- [x] `pnpm parity:check` passes locally (0 errors)
- [ ] CI `parity-check` + per-integration `smoke-starter` green
- [ ] Redeploy affected examples (e.g. Render) and confirm chat works


tylerslaton (dismissed):
Shouldn't this apply to all integrations?

tylerslaton (approved):

PR #5464, opened 2026-06-16
feat(showcase): make showcase_promote.yml dep-aware so single-slug promotes include aimock + harness + dashboard + per-integration container

jpr5:
## Gap
Promoting an integration alone (e.g. `service=langgraph-typescript`) currently leaves stale `showcase-aimock`, `showcase-harness`, and `shell-dashboard` in prod. We hit this with today's prod-vs-staging drift after aimock-only promote at 00:46Z 2026-06-15.

## Fix
- Add optional `promoteDeps?: string[]` to `ServiceEntry` in `railway-envs.ts` (SSOT).
- Every integration declares its runtime deps: `[aimock, harness, dashboard]` (SSOT keys, one level deep).
- `showcase_promote.yml` `resolve-targets` reads the SSOT JSON and unions the deps into `services_csv` with deps-before-parent ordering so promote-fleet bounces infra before the integration that consumes it.
- `service=all` is unchanged (still promotes every `probe.prod == true` entry).
- Infra leaves (aimock/harness/dashboard/shell-*/webhooks) carry no deps — one-way graph; reverse would re-promote the world on every aimock bump.

## Dep graph (one-level)
All 19 agent integrations carry `promoteDeps: [aimock, harness, dashboard]`:

```
showcase-ag2, showcase-agno, showcase-built-in-agent,
showcase-claude-sdk-python, showcase-claude-sdk-typescript,
showcase-crewai-crews, showcase-google-adk,
showcase-langgraph-fastapi, showcase-langgraph-python,
showcase-langgraph-typescript, showcase-langroid,
showcase-llamaindex, showcase-mastra,
showcase-ms-agent-dotnet, showcase-ms-agent-harness-dotnet,
showcase-ms-agent-python, showcase-pydantic-ai,
showcase-spring-ai, showcase-strands
    -> [aimock, harness, dashboard]
```

## Validation guards (new `assertPromoteDepsValid`, runs at module load)
- every dep resolves to an SSOT key (own-property lookup, prototype-safe)
- every dep is `prod` AND `prod.probe !== false`
- no self-reference
- one-level depth (rejects cycles — dep cannot list parent in its own `promoteDeps`)
- no duplicate entries

## Tests
- Real SERVICES map passes `assertPromoteDepsValid` (positive case)
- Every `probeDriver:"agent"` integration with a prod env lists all of aimock+harness+dashboard
- Leaves (aimock/harness/dashboard) carry zero `promoteDeps`
- Synthetic cycle rejected
- Synthetic dangling dep rejected
- Self-reference rejected
- Non-prod-eligible dep (`probe === false`) rejected
- Dep with no `prod` env rejected
- Empty / undefined `promoteDeps` accepted

## Local verify
- `npx vitest run railway-envs.test.ts` -> 84 passed
- `npx tsx emit-railway-envs-json.ts` -> clean regen
- `yamllint -d relaxed` -> warnings only (line-length, pre-existing)
- `bats showcase/scripts/__tests__/promote-fleet.bats` -> 12/12 ok
- `actionlint` not installed locally; CI will run it.
- Cannot dry-run `workflow_dispatch` locally; verified by CI.

### DIFF
--- a/.github/workflows/showcase_promote.yml
+++ b/.github/workflows/showcase_promote.yml
@@ -152,7 +152,32 @@ jobs:
               echo "::error::Ambiguous service '$INPUT' matches multiple SSOT entries: $LIST"
               exit 1
             fi
-            CSV="$MATCHES"
+            # Dep-aware expansion: integration promotes must also bounce the
+            # shared infra they run against (aimock + harness + dashboard at
+            # minimum) or prod drifts off staging. SSOT carries the dep set
+            # on each integration (railway-envs.ts → ServiceEntry.promoteDeps);
+            # the generated JSON exposes it under `.promoteDeps`. Here we read
+            # the dep list from the resolved entry, re-validate each dep is
+            # prod-probe-eligible (defense-in-depth against a runtime-edited
+            # generated JSON — the SSOT validator already enforces this at
+            # module load), append the parent, and dedupe while PRESERVING
+            # order (deps-before-parent so promote-fleet.sh bounces infra
+            # before the integration that consumes it; `unique` would re-sort
+            # alphabetically and drop that ordering).
+            CSV=$(jq -r --arg s "$MATCHES" '
+              ( .services[] | select(.name == $s) ) as $svc
+              | ($svc.promoteDeps // []) as $deps
+              | [ $deps[] as $d
+                  | (.services[] | select(.name == $d) | select(.probe.prod == true) | .name)
+                ] as $okDeps
+              | ($okDeps + [$s])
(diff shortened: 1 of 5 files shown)


jpr5:
Closing as obsolete. Superseded on main by the `runtimeDeps` + transitive `computePromoteClosure()` implementation and the bats-tested `resolve-promote-targets.sh` (19 commits after this PR opened) — strictly more capable than this PR's inline-jq approach. Every anchor this diff patches has been replaced (mergeState DIRTY).

PR #5469, opened 2026-06-16
docs(react-core): clarify system_prompt requirements for useFrontendTool

### LINKED ISSUE #4950: 📚 Documentation: Clarify when useFrontendTool requires system-prompt guidance (sales dashboard month-range fetch)
## Summary

When using `useFrontendTool` to fetch domain data (e.g. reload a sales dashboard by month range) alongside `useAgentContext` (readable snapshot of the same UI), the LangGraph agent does not reliably call the frontend tool unless we also duplicate usage rules in the agent `system_prompt`.

The official CopilotKit + LangGraph demo registers frontend tools like `toggleTheme` / `enableAppMode` with only a short `description`, and does not document equivalent prompt rules for those tools—while other flows (e.g. charts) *do* say "call `query_data` first" in `system_prompt`.

We would like clarity on the **recommended pattern** for mandatory, data-correctness frontend tools.

## Our setup

- **Frontend (Vue host + React CopilotKit v2):** `useFrontendTool` named `fetch_sales_overview_by_month_range` calls the host app's API and updates cards/charts; returns a summary to the agent.
- **Context:** `useAgentContext` exposes the current sales dashboard snapshot (including `current_year`, `today`, and data for the *currently displayed* month range).
- **Agent:** LangGraph `create_agent` with `CopilotKitMiddleware()` only (no Python tools for this fetch). Custom `system_prompt` instructs: when the user asks about a specific month/range, **must** call the frontend tool first and must not guess revenue or order counts.

Without the `system_prompt` rules, the model often answers from stale `useAgentContext` data instead of calling the frontend tool.

## Official sample behavior (for comparison)

From the template repo:

- `useFrontendTool` examples (`toggleTheme`, `enableAppMode`) rely on short descriptions; `system_prompt` does not list each frontend tool by name.
- **But** `system_prompt` *does* include tool guidance for backend / multi-step flows, e.g. "Charts: call `query_data` first, then render with the chart component" and "Todos: enable app mode first, then manage todos."
- Suggestion pills also explicitly mention tools (e.g. "use the toggleTheme tool").

So it is unclear whether **mandatory frontend data pipelines** should be documented like backend `query_data`, or whether `useFrontendTool` `description` + context alone should be sufficient.

## Questions for maintainers

1. For a frontend tool that **must** run before answering (fresh data + update UI), is the recommended approach:
   - (a) explicit rules in agent `system_prompt`,
   - (b) only a detailed `useFrontendTool` `description`,
   - (c) move data fetch to a **backend** LangChain tool and use frontend only for UI sync,
   - or (d) something else (e.g. tool choice middleware)?

2. When `useAgentContext` already contains dashboard data, how do you recommend avoiding the model answering from stale context instead of calling a refresh frontend tool?

3. Should the docs include a "data fetch via frontend tool" example (similar in strictness to the chart + `query_data` flow), including date/parameter inference rules?

## Minimal reproduction (conceptual)

1. Register `useFrontendTool` that fetches sales data by `start_month` / `end_month`.
2. Register `useAgentContext` with the current sales dashboard JSON (different month than the user asks).
3. Agent with generic system prompt + `CopilotKitMiddleware()` only.
4. User asks: "How much revenue did we have in January?" (no year → should use `current_year` from context).
5. **Observed:** model often answers from context without calling the tool.
6. **Workaround:** add mandatory tool-call rules to `system_prompt` → more reliable, but feels duplicated with tool `description`.

## Environment

- `@copilotkit/react-core` v2
- LangGraph agent + `CopilotKitMiddleware`
- Model: `gpt-4.1-mini` (also reproducible with other models)

## What would help us

- Documented best practice for **mandatory** vs **optional** frontend tools.
- Guidance on `useAgentContext` + `useFrontendTool` interaction (staleness / when to refresh).
- Optional: sample that mirrors "query_data first" but with a frontend fetch tool, if that is a supported pattern.

Thanks!

nuthalapativarun:
## What does this PR do?
Adds documentation explaining that `useFrontendTool` may require explicit `system_prompt` guidance in LangGraph agents to be reliably called. Includes a working code example showing both a well-described `useFrontendTool` registration and the matching system prompt instruction. Addresses the confusion reported in #4950 where users implement tools with only a `description` and find the agent doesn't call them.

Changes:
- `docs/integrations/langgraph/frontend-tools.mdx` — new section "Ensuring your agent reliably calls frontend tools" with a `Callout`, Python+TypeScript code examples, and a decision table (description vs. system_prompt)
- `reference/hooks/useFrontendTool.mdx` — added "LangGraph agents: description vs. system_prompt" subsection linking to the full example
- `docs/troubleshooting/common-issues.mdx` — expanded the "tool listed but agent never calls it" bullet to cross-link the new guide

## Related PRs and Issues
- Closes #4950

## Checklist

### DIFF
--- a/showcase/shell-docs/src/content/docs/integrations/langgraph/frontend-tools.mdx
+++ b/showcase/shell-docs/src/content/docs/integrations/langgraph/frontend-tools.mdx
@@ -123,3 +123,106 @@ Use frontend tools when you need your agent to interact with client-side primiti
         <video src="https://cdn.copilotkit.ai/docs/copilotkit/images/frontend-actions-demo.mp4" className="rounded-lg shadow-xl" loop playsInline controls autoPlay muted />
     </Step>
 </Steps>
+
+## Ensuring your agent reliably calls frontend tools
+
+For simple, low-stakes tools (like toggling a theme or enabling a UI mode), a clear `description` is usually enough for the agent to discover and call the tool at the right time. However, for tools where correct behavior is critical — such as fetching fresh data before answering, or updating UI state before replying — LangGraph agents may still answer from cached context or skip the tool call entirely.
+
+The root cause is that LLMs weigh tool descriptions against everything else in the context. If the agent has recent data in `useAgentContext`, it may decide that data is "close enough" and skip the fetch. Stale-data bugs are especially common for domain-specific data pipelines where the agent must call a frontend tool to get up-to-date values.
+
+**Recommended pattern:** supplement the tool `description` with an explicit instruction in the agent's `system_prompt`.
+
+<Callout type="info" title="Frontend tool descriptions vs. system_prompt guidance">
+The `description` field on `useFrontendTool` tells the model *what* the tool does. The agent's `system_prompt` is where you tell it *when* to call it — especially for mandatory steps like refreshing data before answering a question.
+
+If you find your LangGraph agent answering from stale context instead of calling the tool, add an explicit instruction to `system_prompt` as shown below.
+</Callout>
+
+The following example shows a data-fetch frontend tool paired with a matching `system_prompt` rule that ensures the agent always calls the tool before answering questions about live data.
+
+<Tabs groupId="language_langgraph_agent" items={['Python', 'TypeScript']} default="Python" persist>
+    <Tab value="Python">
+        ```python title="agent.py"
+        from langgraph.prebuilt import create_react_agent
+        from copilotkit import CopilotKitMiddleware, CopilotKitState
+
+        graph = create_react_agent(
+            model="openai:gpt-4o",
+            tools=[],  # backend tools go here
+            middleware=[CopilotKitMiddleware()],
+            state_schema=CopilotKitState,
+            # [!code highlight:7]
+            # Tell the agent WHEN it must call the frontend tool.
+            # A short description alone may not be enough — make the rule explicit.
+            prompt=(
+                "You are a helpful sales assistant.\n\n"
+                "IMPORTANT: When the user asks about sales data for a specific month or date range, "
+                "you MUST call the `fetch_sales_by_month_range` frontend tool first to retrieve "
+                "up-to-date figures. Never answer from memory or previously seen context values."
+            ),
+        )
+        ```
+    </Tab>
+    <Tab value="TypeScript">
+        ```typescript title="agent-js/src/agent.ts"
+        import { createReactAgent } from "@langchain/langgraph/prebuilt";
+        import { copilotkitMiddleware, CopilotKitStateSchema } from "@copilotkit/sdk-js/langgraph";
+
+        export const graph = createReactAgent({
+          model: "openai:gpt-4o",
+          tools: [],  // backend tools go here
+          stateSchema: CopilotKitStateSchema,
+          checkpointSaver: undefined,
+          // [!code highlight:6]
+          // Tell the agent WHEN it must call the frontend tool.
+          // A short description alone may not be enough — make the rule explicit.
+          prompt:
+            "You are a helpful sales assistant.\n\n" +
+            "IMPORTANT: When the user asks about sales data for a specific month or date range, " +
+            "you MUST call the `fetch_sales_by_month_range` frontend tool first to retrieve " +
+            "up-to-date figures. Never answer from memory or previously seen context values.",
+        });
+        ```
+    </Tab>
+</Tabs>
+
+And register the matching frontend tool on the client:
+
+```tsx title="page.tsx"
+import { z } from "zod";
+import { useFrontendTool } from "@copilotkit/react-core/v2";
+
+useFrontendTool({
+  name: "fetch_sales_by_month_range",
+  // [!code highlight:3]
+  // Keep the description accurate so the model understands the tool's contract,
+  // but rely on system_prompt to enforce WHEN it is called.
+  description:
+    "Fetches sales overview data for the given month range from the live API and updates the dashboard. " +
+    "Returns a summary of revenue and order counts for the requested period.",
+  parameters: z.object({
+    start_month: z.string().describe("Start month in YYYY-MM format"),
+    end_month: z.string().describe("End month in YYYY-MM format"),
+  }),
+  handler: async ({ start_month, end_month }) => {
+    const data = await fetchSalesOverview({ start_month, end_month });
+    updateDashboard(data);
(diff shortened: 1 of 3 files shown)


nuthalapativarun:
Just bumping this — the system_prompt docs clarification for useFrontendTool should be a quick read whenever someone has a few minutes. The Vercel check failure is just the auth-gated preview deploy for fork PRs, not an actual build issue. Thanks!

nuthalapativarun:
Hi team — following up on PR #5469 (useFrontendTool system_prompt docs, closes #4950). Mergeable, Vercel CI failures are auth-gated fork-PR false positives. 2nd bump. Happy to address any feedback. Thanks!

BenTaylorDev:
Thanks @nuthalapativarun — content looks reasonable. Two items:

1. **Dangling cross-link:** the new section links to `/langgraph/advanced/emit-api-hook-mapping`, which only exists in your #5483 — merging this alone would produce a broken link. These two are coupled; let's land them together (or drop the link until #5483 is in).
2. Please double-check the `create_react_agent(middleware=[CopilotKitMiddleware()], prompt=...)` / `state_schema=CopilotKitState` example against the current Python SDK signature, since the branch is a ways behind main.

BenTaylorDev (changes requested):
Thanks — the thesis is valuable and (nice) it's correctly authored against the current `showcase/shell-docs/src/content/` tree and merges clean. But the code examples don't match the shipped API and wouldn't run as written:

- They use `from langgraph.prebuilt import create_react_agent` with a `middleware=` kwarg — but in the pinned langchain 1.x, `middleware=` belongs to `create_agent` (`langchain.agents`), not `create_react_agent`; as written this likely raises `TypeError`. Every example on main uses `create_agent`/`createAgent`.
- They pass `prompt=`/`prompt:` for the system prompt, which both contradicts the PR's own headline ("add it to `system_prompt`") and isn't the current param (`system_prompt=`/`systemPrompt:`).
- The **TypeScript example imports `copilotkitMiddleware` but never attaches it** (no `middleware:` array) — so frontend tools wouldn't actually reach the agent in that snippet.
- The added "Imperatively emitting tool calls" section links to `/langgraph/advanced/emit-api-hook-mapping`, which doesn't exist on main (dead link), and is scope creep beyond #4950 (system_prompt guidance).

Please align the examples with the current pinned API (`create_agent`/`createAgent` + `system_prompt=`/`systemPrompt:`, matching `main.py`/`agent.ts`), actually wire `copilotkitMiddleware` into the TS `middleware:` array, and either drop the emit section or fix its link. The prose is close — it's the examples that need to be made accurate and runnable.

nuthalapativarun:
Thanks for the detailed review! Pushed a fix that:

- Swaps `create_react_agent(..., middleware=..., prompt=...)` for `create_agent`/`createAgent` with `system_prompt`/`systemPrompt`, matching the current pinned langchain 1.x API (verified against the pattern already used in `agent-app-context.mdx`).
- Wires `copilotkitMiddleware` into the TypeScript example's `middleware` array — it was imported but never attached, so frontend tools wouldn't have reached the agent.
- Points the "Imperatively emitting tool calls" link at `/integrations/langgraph/advanced/emit-api-hook-mapping`, matching the actual route of the new page added in #5483 (also just updated) so the cross-link isn't dangling once both land.

Since you flagged these two as coupled, happy to have them merge together.

BenTaylorDev (changes requested):
Thanks — the API examples are now correct in both languages (`create_agent`/`createAgent`, `system_prompt`/`systemPrompt`, and the TS example actually attaches `copilotkitMiddleware`), matching the in-repo reference agents, and the docs are correctly under `showcase/shell-docs`. One item remains: the "Imperatively emitting tool calls from agent code" section still links to `/integrations/langgraph/advanced/emit-api-hook-mapping`, which doesn't exist on `main` (you fixed the URL prefix, but the target page isn't there). That page is the one #5483 creates — so either drop this section (it's outside the system_prompt scope of this PR) or sequence it to land after #5483 so the link resolves. Everything else looks good.

BenTaylorDev (approved):
Re-reviewed against current `main`. **The one item I left open has resolved itself** — #5483 merged about nine minutes after my last review, so `/integrations/langgraph/advanced/emit-api-hook-mapping` now exists on `main`, the link text matches that page's title verbatim, and the page does cover `copilotkit_emit_tool_call` and `copilotkit_customize_config` as promised. That was the sequencing option I offered, so nothing is left for you to do here. Sorry for the wait — the blocker expired the same afternoon I wrote it.

Everything else I checked holds up:

- **API accuracy** — I verified the snippets against the pinned wheels, not just by eye. `create_agent(model=, tools=, middleware=, state_schema=, system_prompt=)` matches `langchain/agents/factory.py` exactly; `CopilotKitMiddleware` and `CopilotKitState` are real top-level `copilotkit` exports; the Python example matches `examples/integrations/langgraph-python/agent/main.py` line for line, imports included. The TypeScript example matches `examples/integrations/langgraph-js/agent/src/agent.ts`, and `copilotkitMiddleware` is genuinely wired into `middleware:` now. `stateSchema: CopilotKitStateSchema` is valid — it's a real `StateSchema` instance.
- **Placement** — correctly authored under `showcase/shell-docs/src/content/`, and it merges into current `main` clean despite the branch being stale.
- **Conventions** — `<Tabs groupId="language_langgraph_agent" … persist>`, `<Callout>`, and the `[!code highlight:N]` transformer all match what the surrounding pages already do. The anchor the two cross-links point at matches the new heading.

Two nits, neither blocking, both fine to leave:

- `model="openai:gpt-4o"` — these docs otherwise use `openai:gpt-5.4`.
- The `[!code highlight:7]` / `:6` ranges each stop one line short, so the closing line of each block isn't highlighted.

One note for anyone reading this later, *not* a request: the `# Works the same for "create_react_agent" or similar options` comment isn't actually true — in the pinned `langgraph`, `create_react_agent` takes neither `middleware` nor `system_prompt` and raises on both. That comment already ships on `main` in two other files, so you were following the house style and this PR isn't the place to fix it; I'll file the tree-wide cleanup separately.

The prose is genuinely useful — the "description tells the model *what*, `system_prompt` tells it *when*" framing plus the decision table is the right shape for #4950. Thanks for sticking with this through two rounds.

I bless the code with a glad heart 🙏

BenTaylorDev:
Note: 2 failing tests are caused by the stale main base and are fixed on main, and these changes are docs-only and so wouldn't run afoul of them.

PR #5482, opened 2026-06-16
fix(shared): lazy-load segment analytics for browser bundles

### LINKED ISSUE #4151: 🐛 Bug: Vite browser build warns about Node.js modules being resolved
### ♻️ Reproduction Steps

1. Open this Vite 7 playground: https://stackblitz.com/edit/vitejs-vite-3fxmdguz?file=src%2Fmain.tsx
2. Run `npm run build`

Alternative:

1. Create a new Vite 7 project
2. Install `@copilotkit/react-core@1.56.2`
3. Import something from the package and use it in the code so it doesn't get removed in the bundling process
    ```ts
    import { useAgent } from '@copilotkit/react-core/v2';
    console.log(useAgent);
    ```
4. Create a production build using Vite

### ✅ Expected Behavior

CopilotKit browser-facing packages should not cause Vite to resolve Node-only dependencies during a browser build.

If telemetry is not explicitly used, packages such as @segment/analytics-node, node-fetch, and Node built-ins like stream should not be pulled into the browser build graph, and no externalized-module warnings should appear.

### ❌ Actual Behavior

The build succeeds, but Vite prints the following warnings:

> [plugin vite:resolve] Module "stream" has been externalized for browser compatibility, imported by "/home/projects/vitejs-vite-3fxmdguz/node_modules/node-fetch/lib/index.mjs". See https://vite.dev/guide/troubleshooting.html#module-externalized-for-browser-compatibility for more details.
[plugin vite:resolve] Module "http" has been externalized for browser compatibility, imported by "/home/projects/vitejs-vite-3fxmdguz/node_modules/node-fetch/lib/index.mjs". See https://vite.dev/guide/troubleshooting.html#module-externalized-for-browser-compatibility for more details.
[plugin vite:resolve] Module "url" has been externalized for browser compatibility, imported by "/home/projects/vitejs-vite-3fxmdguz/node_modules/node-fetch/lib/index.mjs". See https://vite.dev/guide/troubleshooting.html#module-externalized-for-browser-compatibility for more details.
[plugin vite:resolve] Module "https" has been externalized for browser compatibility, imported by "/home/projects/vitejs-vite-3fxmdguz/node_modules/node-fetch/lib/index.mjs". See https://vite.dev/guide/troubleshooting.html#module-externalized-for-browser-compatibility for more details.
[plugin vite:resolve] Module "zlib" has been externalized for browser compatibility, imported by "/home/projects/vitejs-vite-3fxmdguz/node_modules/node-fetch/lib/index.mjs". See https://vite.dev/guide/troubleshooting.html#module-externalized-for-browser-compatibility for more details.

```
❯ npm ls node-fetch
vite-react-typescript-starter@0.0.0 /home/projects/vitejs-vite-3fxmdguz
`-- @copilotkit/react-core@1.56.2
  `-- @copilotkit/shared@1.56.2
    `-- @segment/analytics-node@2.3.0
      `-- node-fetch@2.7.0
```

From inspecting the installed package, it looks like @copilotkit/shared eagerly imports its telemetry client from the main entrypoint, which pulls in @segment/analytics-node and then node-fetch. That seems to make Vite resolve a Node-only dependency chain in a browser-targeted build.

Interestingly, these warnings were not emitted when using Vite 8 with Rolldown.

### Screenshots

_No response_

### 𝌚 CopilotKit Version

```shell
@copilotkit/react-core@1.56.2
```

### 📄 Logs (Optional)

```shell

```

nuthalapativarun:
## What does this PR do?
Changes the top-level `import { Analytics } from "@segment/analytics-node"` in `telemetry-client.ts` to a lazy dynamic import inside the initialization method. This prevents Node.js built-in modules (`stream`, `http`, `url`, `https`, `zlib`) from being pulled into browser Vite builds, eliminating the "module externalized for browser compatibility" warnings reported in #4151.

The `Analytics` class is now loaded via an `import("@segment/analytics-node").then(...)` call in the constructor, so the import only resolves in Node.js environments. In browser builds, the dynamic import fails gracefully and telemetry falls back to the lambda sink only.

## Related PRs and Issues
- Closes #4151

## Checklist

### DIFF
--- a/packages/shared/src/telemetry/telemetry-client.ts
+++ b/packages/shared/src/telemetry/telemetry-client.ts
@@ -1,4 +1,3 @@
-import { Analytics } from "@segment/analytics-node";
 import type { AnalyticsEvents } from "./events";
 import { flattenObject } from "./utils";
 import { v4 as uuidv4 } from "uuid";
@@ -23,7 +22,8 @@ export function isTelemetryDisabled(): boolean {
 }
 
 export class TelemetryClient {
-  segment: Analytics | undefined;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  segment: any | undefined;
   globalProperties: Record<string, any> = {};
   cloudConfiguration: { publicApiKey: string; baseUrl: string } | null = null;
   // EIP / Intelligence license token (Ed25519-signed JWT). The lambda
@@ -73,9 +73,18 @@ export class TelemetryClient {
       process.env.COPILOTKIT_SEGMENT_WRITE_KEY ||
       "n7XAZtQCGS2v1vvBy3LgBCv2h3Y8whja";
 
-    this.segment = new Analytics({
-      writeKey,
-    });
+    // Lazy dynamic import so that @segment/analytics-node (which pulls in
+    // Node.js built-ins like stream, http, url, https, zlib) is never
+    // evaluated at module parse time in browser / Vite builds. If the
+    // import fails (e.g. in a browser context), Segment is silently skipped.
+    import("@segment/analytics-node")
+      .then(({ Analytics }) => {
+        this.segment = new Analytics({ writeKey });
+      })
+      .catch(() => {
+        // Browser environment or missing package — telemetry will use
+        // the lambda sink only; Segment track() calls are no-ops.
+      });
 
     this.setGlobalProperties({
       "copilotkit.package.name": packageName,
--- a/packages/shared/src/telemetry/telemetry-client.ts
+++ b/packages/shared/src/telemetry/telemetry-client.ts
@@ -1,3 +1,4 @@
+import type { Analytics } from "@segment/analytics-node";
 import type { AnalyticsEvents } from "./events";
 import { flattenObject } from "./utils";
 import { v4 as uuidv4 } from "uuid";
@@ -22,8 +23,14 @@ export function isTelemetryDisabled(): boolean {
 }
 
 export class TelemetryClient {
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  segment: any | undefined;
+  segment: Analytics | undefined;
+  // Resolves once the lazy segment import has settled (loaded or failed).
+  // `capture()` awaits this before checking `this.segment` so a fast-exit
+  // process (e.g. a short-lived CLI) can't fire capture() before the import
+  // resolves and silently drop the event. Defaults to an already-resolved
+  // promise so capture() is a no-op await when telemetry is disabled (the
+  // constructor returns early without ever assigning a real promise).
+  private segmentReady: Promise<void> = Promise.resolve();
   globalProperties: Record<string, any> = {};
   cloudConfiguration: { publicApiKey: string; baseUrl: string } | null = null;
   // EIP / Intelligence license token (Ed25519-signed JWT). The lambda
@@ -75,9 +82,17 @@ export class TelemetryClient {
 
     // Lazy dynamic import so that @segment/analytics-node (which pulls in
     // Node.js built-ins like stream, http, url, https, zlib) is never
-    // evaluated at module parse time in browser / Vite builds. If the
-    // import fails (e.g. in a browser context), Segment is silently skipped.
-    import("@segment/analytics-node")
+    // evaluated at module parse time in browser / Vite builds. The
+    // `@vite-ignore` comment stops Vite's optimizeDeps esbuild scanner from
+    // statically discovering and pre-bundling this specifier — without it,
+    // the scanner still walks into @segment/analytics-node's Node built-in
+    // requires at dev-server start and reproduces the "externalized for
+    // browser compatibility" warning this change is meant to eliminate.
+    // If the import fails (e.g. in a browser context), Segment is silently
+    // skipped and telemetry falls back to the lambda sink only. capture()
+    // awaits `segmentReady` before checking `this.segment`, so callers never
+    // race the import.
+    this.segmentReady = import(/* @vite-ignore */ "@segment/analytics-node")
       .then(({ Analytics }) => {
         this.segment = new Analytics({ writeKey });
       })
@@ -149,6 +164,11 @@ export class TelemetryClient {
       packageVersion: this.packageVersion,
       licenseToken: this.licenseToken ?? undefined,
     });
+
+    // Wait for the lazy segment import to settle before checking
+    // `this.segment`, so a capture() firing right after construction can't
+    // race the import and silently drop the event.
+    await this.segmentReady;
 
     if (this.segment) {
       this.segment.track({


nuthalapativarun:
Friendly nudge on this one — lazy-loading segment analytics should help trim the browser bundle size a bit. Noticed the Vercel preview check shows red, but that's just the auth gate on fork PRs, the actual build is fine. Let me know if you'd like any changes.

nuthalapativarun:
Hi team — following up on PR #5482 (lazy-load segment analytics, closes #4151). Mergeable. 2nd bump. Happy to adjust if needed. Thanks!

gslaszlo:
Wow, +1 for this PR, loading Node.js built-in modules for browser build is just really annoying and needs workarounds. This simple fix solves the problem, can't wait to be merged! Thanks!

BenTaylorDev:
Thanks @nuthalapativarun — the top-level `@segment/analytics-node` import at `telemetry-client.ts:1` is a real problem. One blocking concern on the mechanism, plus two smaller ones:

1. **Does deferring the import actually remove the warnings?** Rollup/Vite statically analyze a static-specifier dynamic `import("@segment/analytics-node")` and still walk/externalize its Node built-ins (`stream`, `http`, `zlib`…), so the "externalized for browser compatibility" warnings tend to just move into the lazy chunk rather than disappear. Could you attach a before/after browser (Vite) build against this branch showing the warnings genuinely go away? If they don't, this wouldn't close #4151.
2. **Type loss:** `segment: Analytics` → `segment: any` drops type-checking on `this.segment.track(...)` — please keep the type (e.g. a lazily-typed field).
3. **Race:** assignment is now async, so a fast-exit CLI could `capture()` before the import resolves and silently drop the event — worth guarding.

BenTaylorDev (changes requested):
Thanks — the goal is right (keep the Node-only `@segment/analytics-node` out of browser bundles), and it's the only static import site, so completeness is fine. Two things to sort out before merge:

1. **The mechanism may not actually silence the warning — needs build evidence.** Vite's `optimizeDeps` esbuild scanner follows dynamic `import()` calls with static string literals, so `@segment/analytics-node` (and its `stream`/`http`/`url`/`https`/`zlib` requires) is still likely discovered and pre-bundled at dev time — the exact path that emits the "externalized for browser compatibility" warning #4151 reports. Dynamic import may drop the production rollup warning while leaving the dev-server one. Could you attach a before/after from an actual Vite build **and** dev server reproducing #4151? If the dev warning persists, the durable fix is probably a `browser` conditional export / stub rather than a runtime dynamic import.
2. **Async-init drops early events.** On `main`, `this.segment` is assigned synchronously in the constructor; with the dynamic import it's `undefined` until the promise resolves, and `capture()` guards `if (this.segment)` (`telemetry-client.ts:144`). For short-lived Node processes (CLI one-shots), any `capture()` firing before the import resolves is silently dropped from the Segment sink. Please preserve the Node-path behavior (e.g. await/queue events until the client is ready).

Also minor: `segment: Analytics | undefined` was downgraded to `any` — worth keeping the type.

nuthalapativarun:
Thanks for digging into the mechanism! Pushed a fix addressing all three points:

1. **Vite scanner discovery**: added `/* @vite-ignore */` to the dynamic `import()` so Vite's optimizeDeps esbuild scanner doesn't statically discover and pre-bundle `@segment/analytics-node` at dev-server start — that static discovery is what was walking into its Node built-in requires and reproducing the warning. I don't have a scaffolded Vite app in this environment to capture a literal before/after dev-server log, so I'd appreciate a check on your end that the warning is gone; happy to iterate further if it isn't (e.g. a `browser` conditional export/stub as you suggested).
2. **Type loss**: restored `segment: Analytics | undefined` via a type-only import (`import type { Analytics } from "@segment/analytics-node"`), which is erased at build time so it doesn't reintroduce the eager runtime import.
3. **Race condition**: `capture()` now awaits a `segmentReady` promise (set to the dynamic import's promise) before checking `this.segment`, instead of checking it synchronously. This closes the "fast-exit process drops the event" gap you flagged. Verified against the existing `telemetry-client.test.ts` suite, which mocks `@segment/analytics-node` — all 192 tests pass (5 of them were actually failing against my first pass at this fix before the `segmentReady` await, which confirmed the race was real).

BenTaylorDev (changes requested):
The typing is restored (`import type` + `Analytics | undefined`) and the async-init race is genuinely fixed (`segmentReady` promise awaited before the `if (this.segment)` check) — both good. The Vite item is still open: the before/after evidence the last review asked for isn't here, and there's a specific risk — `@copilotkit/shared` is built with tsdown/rolldown, so the inline `/* @vite-ignore */` must survive verbatim into the emitted `dist/telemetry/telemetry-client.mjs` for downstream Vite consumers to benefit. Could you paste that built `.mjs` line showing the comment intact inside the `import()`, plus a downstream Vite dev/build before/after showing the "externalized for browser compatibility" warning gone? That's the load-bearing proof.

gslaszlo:
Hello guys, an chance on this get moving forward? Really would love this fixed, so I could remove ugly polyfills in vite. Thank you very much for your work! 🙏

BenTaylorDev:
Thanks for the follow-up @nuthalapativarun, and thanks for being straight about not having a Vite app to measure in. I ran that measurement here, and the result is that the lazy import does not remove the warnings.

## Method

I built `@copilotkit/shared` with tsdown twice from your branch head (`78252c8`). The two builds differ in exactly one file, `packages/shared/src/telemetry/telemetry-client.ts`: your version, and the version at this branch's merge base (`d305dc6`). I then pointed a minimal Vite 7.3.2 app at each build and ran `vite build`.

## Result

| consumer | variant | warnings | modules transformed | total asset bytes |
| --- | --- | --- | --- | --- |
| `import { TelemetryClient }` | base | 4 | 663 | 445,970 |
| `import { TelemetryClient }` | this PR | 4 | 663 | 446,186 |
| `import { logger }` | base | 4 | 663 | 742 |
| `import { logger }` | this PR | 4 | 663 | 742 |

The four warnings are the same four from #4151 (`stream`, `http`, `https`, `zlib`, all via `node-fetch`), and they are byte-identical across variants.

To make sure the probe was sensitive and not just blind, I ran an isolated fixture with a control:

- static `import f from "node-fetch"` gives 4 warnings
- `import(/* @vite-ignore */ "node-fetch")` gives the same 4 warnings
- no import at all gives 0 warnings

## Why the mechanism cannot work

The warnings come from `vite:resolve` walking the module graph. That happens before tree-shaking, and Vite resolves the target of a dynamic import too, because it has to emit a chunk for it. A dynamic import defers *evaluation*, but it keeps the *graph edge*, so the resolve step still reaches `node-fetch`.

I also tested the dev-server path your comment describes, because that claim deserved a check rather than an assumption. `vite optimize --force` pre-bundles `@segment/analytics-node` identically in both variants. So `@vite-ignore` does not change the optimizeDeps scanner here either.

There is also no bundle win. For a consumer that does not touch telemetry, output is 742 bytes in both variants, because `@segment/analytics-node` was already tree-shaken out on main. For a consumer that does touch it, the payload is the same size and simply moves behind a chunk that the constructor fetches right away.

## What does fix it

Cutting the static edge out of the browser-facing entry. Measured on the same setup: drop `export * from "./telemetry-client"` from `packages/shared/src/telemetry/index.ts` and move `isTelemetryDisabled` into its own module. That gives 448 modules and zero warnings, while `lambdaClient` and `isTelemetryDisabled` stay on the root export. This is the same reasoning already written into `packages/shared/src/index.ts` for `license-verifier`, where only types are re-exported so Node's `crypto` stays out of client bundles.

We will take that one ourselves, because it moves `TelemetryClient` off the root export of a published package, and that is a breaking change we need to sequence with a deprecation.

So I am closing this PR, and #4151 stays open until the export split lands. Two things in your latest push were right on their own terms and I want to say so: restoring the type through a type-only import, and closing the capture-before-import race. They just ride on a premise that does not hold. Thank you for the time you put into this, and sorry it did not land.

PR #5483, opened 2026-06-16
docs(langgraph): document emit API to v2 hook mapping end-to-end

### LINKED ISSUE #3301: 📚 Documentation: # Improve docs for integrating LangGraph SDK emit APIs with React v2 UI hooks (`useComponent`, `useFrontendTool`, etc.)

Hi CopilotKit team, thanks for the work on the project.

I’d like to request better documentation for how backend LangGraph SDK APIs and frontend React v2 hooks are intended to work together, especially for:

- `copilotkit_emit_tool_call`
- `copilotkit_customize_config(..., emit_tool_calls=...)`
- `useComponent`
- `useFrontendTool`
- Etc.

Current docs I checked:

- https://docs.copilotkit.ai/reference/v1/sdk/python/LangGraph
- https://docs.copilotkit.ai/reference/v2
- https://docs.copilotkit.ai/langgraph/

The issue is that it’s hard to find an end-to-end explanation of expected behavior across backend + frontend. In practice, developers have to infer semantics from trial/error.

## What would help a lot

1. End-to-end matrix/table of behavior
   - Persisted tool calls in `messages` vs emitted tool calls via `copilotkit_emit_tool_call`
   - Which frontend mechanisms react to each (`useComponent`, `useFrontendTool`, custom events/subscriptions)
   - Whether behavior is transient vs persisted in chat history

2. Explicit lifecycle docs for each UI hook
   - What triggers `useComponent` rendering exactly
   - What triggers `useFrontendTool.handler` exactly
   - Whether either works from emitted events alone, persisted messages alone, or both

3. One minimal reproducible example per pattern
   - Persisted tool call -> inline chat component
   - Emit-only tool call -> transient UI update
   - Custom event -> side-panel/state update

4. Clear guidance on recommended production patterns
   - When to use emitted events vs persisted messages
   - How to avoid duplicate rendering or message pollution
   - Tradeoffs for thread persistence/debugging

5. Version-specific notes
   - If behavior differs between v1/v2 docs, please call this out explicitly.

## Why this matters

It is currently difficult to adopt CopilotKit professionally because teams need predictable integration behavior across backend and frontend, and this is hard to piece together from existing docs.

Thanks again for considering this.

nuthalapativarun:
## What does this PR do?
Adds end-to-end documentation showing how LangGraph SDK backend emit APIs (`copilotkit_emit_tool_call`, `copilotkit_customize_config`) connect to React v2 frontend hooks (`useFrontendTool`, `useComponent`). Includes a mapping table and a concrete code example demonstrating the full wiring from a LangGraph node to a frontend tool handler. Addresses the documentation gap reported in #3301.

## Related PRs and Issues
- Closes #3301

## Checklist

### DIFF
--- a/showcase/shell-docs/src/content/docs/integrations/langgraph/advanced/emit-api-hook-mapping.mdx
+++ b/showcase/shell-docs/src/content/docs/integrations/langgraph/advanced/emit-api-hook-mapping.mdx
@@ -1,0 +1,211 @@
+---
+title: "Connecting Backend Emit APIs to Frontend Hooks"
+icon: "lucide/ArrowLeftRight"
+description: "End-to-end guide: how LangGraph SDK emit functions map to React v2 frontend hooks."
+---
+
+## What is this?
+
+CopilotKit's Python SDK provides **emit functions** that your LangGraph nodes can call to push tool calls and control streaming behavior. On the frontend, React v2 provides **hooks** that register handlers and renderers for those emitted events.
+
+This page shows exactly how the two sides connect so you can build predictable, end-to-end integrations without guessing.
+
+## The mapping at a glance
+
+| Backend (Python SDK) | Frontend (React v2) | What it does |
+|---|---|---|
+| `copilotkit_emit_tool_call(config, name, args)` | `useFrontendTool` | Agent manually fires a tool call; the registered handler runs in the browser |
+| LLM calls a frontend tool normally (via `bind_tools`) | `useFrontendTool` | Agent lets the LLM decide when to call the tool; same handler, same result |
+| LLM calls a frontend tool (via `bind_tools`) | `useComponent` | Agent calls a tool and the registered component is rendered inline in the chat |
+| `copilotkit_customize_config(config, emit_tool_calls=["ToolName"])` | `useFrontendTool` / `useComponent` | Controls which tool calls are streamed to the frontend; use this to filter noise |
+| `copilotkit_customize_config(config, emit_tool_calls=False)` | — | Suppresses all tool call streaming for that LLM invocation |
+
+### Persisted vs. emitted tool calls
+
+<Callout type="info" title="Key concept: persisted vs. transient">
+- **Persisted tool calls** are LLM-generated tool call messages stored in `state["messages"]`. They appear in chat history and are replayed on page refresh.
+- **Emitted tool calls** (`copilotkit_emit_tool_call`) are transient events that fire a frontend handler immediately. They are **not** stored in chat history unless you explicitly add a message to state.
+
+`useComponent` and `useFrontendTool` both respond to persisted tool calls triggered by the LLM. `copilotkit_emit_tool_call` lets you fire the same frontend mechanism imperatively, without waiting for the LLM.
+</Callout>
+
+## End-to-end example: `copilotkit_emit_tool_call` → `useFrontendTool`
+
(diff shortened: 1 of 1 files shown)


nuthalapativarun:
Checking in on this one — tried to give the emit API to v2 hook mapping for langgraph a thorough end-to-end writeup since that mapping trips people up. The failing Vercel check is just the fork auth gate, not a real failure. Happy to adjust if needed.

nuthalapativarun:
Hi team — following up on PR #5483 (langgraph emit API docs, closes #3301). Mergeable. 2nd bump. Thanks for your time!

BenTaylorDev:
Thanks @nuthalapativarun — the new `emit-api-hook-mapping` page is a good addition. Two fixes needed:

1. **Please make this docs-only.** The diff also bundles a ~9-line `isReady` change to `use-agent.tsx` that's unrelated to this docs page (and duplicates the now-merged #6041) — please drop that so the PR is just the docs.
2. The example hardcodes model `"gpt-5.4"`, which isn't a real model — please swap in a real one.

Also note this is coupled with #5469 (its new section links here), so we'll land the pair together.

BenTaylorDev (changes requested):
Nice page — it's correctly authored under `showcase/shell-docs/`, and the emit-API mapping checks out (`copilotkit_emit_tool_call` / `copilotkit_customize_config` signatures and the `useFrontendTool`/`useComponent` hooks all match current source; no equivalent page exists yet). Two fixes before merge:

- **Drop the bundled `use-agent.tsx` change.** The PR also edits `packages/react-core/src/v2/hooks/use-agent.tsx` to add an `isReady` field — but `isReady` already exists on `main` with a fuller implementation, so this reintroduces a conflicting, inferior duplicate (it's the source of the merge conflict) and is unrelated to the docs. The `.mdx` never references it.
- **Fix the "See also" links.** They use the `/langgraph/...` prefix, but the docs route these under `/integrations/langgraph/...` — so `frontend-tools`, the generative-UI display-only page, and `advanced/disabling-state-streaming` all 404 as written. The `/reference/v1/sdk/python/LangGraph` link has no target page at all — drop or replace it.

With the source edit removed and the links corrected, this is good to go.

nuthalapativarun:
Thanks for catching both! Pushed a fix:

- Rebased onto current `main` and dropped the `use-agent.tsx` `isReady` commit entirely — this branch is now docs-only, one commit. `isReady` already ships on main via #6041 as you noted.
- Fixed the "See also" links to the `/integrations/langgraph/...` prefix (`frontend-tools`, `generative-ui/your-components/display-only`, `advanced/disabling-state-streaming`), and dropped the dead `/reference/v1/sdk/python/LangGraph` link since it has no valid target.
- Also swapped the placeholder `"gpt-5.4"` model string for a real one (`gpt-4o`) in both code samples.

This should also unblock #5469, which links to this page's now-correct route.

BenTaylorDev (approved):
**Re-approving** — both blockers are fixed: the conflicting `use-agent.tsx` edit is dropped (the PR is now just the `.mdx`), and the "See also" links are corrected to `/integrations/langgraph/...` (all three targets exist on `main`) with the invented python-reference link removed. Content matches the current emit-API signatures. One cosmetic nit (non-blocking): the at-a-glance table shows `copilotkit_emit_tool_call(config, name, args)` positionally while `name`/`args` are keyword-only — the step-by-step example below it uses the correct form.

PR #5485, opened 2026-06-16
fix(react-ui): display tool call icons in flex row

### LINKED ISSUE #4777: 🐛 Bug: Tool messages design issue
### ♻️ Reproduction Steps

1. Begin with default fresh installation (1.57.0)
2. Use an Ollama LLM instead of the default.
3. Send a prompt that triggers query_data tool.

### ✅ Expected Behavior

Chat window displays the used tool information in good design.

### ❌ Actual Behavior

As you see, tool information icons displayed vertically.

Image

### Screenshots

_No response_

### 𝌚 CopilotKit Version

```shell
@copilotkit/a2ui-renderer@1.57.0
@copilotkit/react-core@1.57.0
@copilotkit/react-ui@1.57.0
@copilotkit/runtime@1.57.0
```

### 📄 Logs (Optional)

```shell

```

nuthalapativarun:
## What does this PR do?

Fixes the tool call icon container in `WildcardToolCallRender` to use `display: flex` (with `align-items: center`) via inline styles as a guaranteed fallback, so the chevron, status dot, and tool name render horizontally inline instead of stacking vertically.

**Root cause**: The header row and inner icon container in `WildcardToolCallRender` used only CPK Tailwind utility classes (`cpk:flex`, `cpk:items-center`) for their flex layout. When the v2 stylesheet (`@copilotkit/react-ui/v2/styles.css`) was not imported — for example when a project only imports the v1 `@copilotkit/react-ui/styles.css` — those classes were not generated and the browser fell back to `display: block`, stacking the icons vertically.

The fix adds inline `style={{ display: "flex", alignItems: "center", ... }}` to the two critical containers so the layout is guaranteed regardless of whether the CPK CSS is loaded. The existing Tailwind classes are kept for theme-consistent spacing overrides when the CSS is present.

## Related PRs and Issues
- Closes #4777

## Checklist

### DIFF
--- a/packages/react-core/src/v2/components/WildcardToolCallRender.tsx
+++ b/packages/react-core/src/v2/components/WildcardToolCallRender.tsx
@@ -24,9 +24,10 @@ export const WildcardToolCallRender = defineToolCallRenderer({
         <div className="cpk:rounded-xl cpk:border cpk:border-zinc-200/60 cpk:dark:border-zinc-800/60 cpk:bg-white/70 cpk:dark:bg-zinc-900/50 cpk:shadow-sm cpk:backdrop-blur cpk:p-4">
           <div
             className="cpk:flex cpk:items-center cpk:justify-between cpk:gap-3 cpk:cursor-pointer"
+            style={{ display: "flex", alignItems: "center", justifyContent: "space-between", gap: "12px", cursor: "pointer" }}
             onClick={() => setIsExpanded(!isExpanded)}
           >
-            <div className="cpk:flex cpk:items-center cpk:gap-2 cpk:min-w-0">
+            <div className="cpk:flex cpk:items-center cpk:gap-2 cpk:min-w-0" style={{ display: "flex", alignItems: "center", gap: "8px", minWidth: 0 }}>
               <svg
                 className={`cpk:h-4 cpk:w-4 cpk:text-zinc-500 cpk:dark:text-zinc-400 cpk:transition-transform ${
                   isExpanded ? "cpk:rotate-90" : ""
--- a/packages/react-core/src/v2/components/WildcardToolCallRender.tsx
+++ b/packages/react-core/src/v2/components/WildcardToolCallRender.tsx
@@ -24,10 +24,9 @@ export const WildcardToolCallRender = defineToolCallRenderer({
         <div className="cpk:rounded-xl cpk:border cpk:border-zinc-200/60 cpk:dark:border-zinc-800/60 cpk:bg-white/70 cpk:dark:bg-zinc-900/50 cpk:shadow-sm cpk:backdrop-blur cpk:p-4">
           <div
             className="cpk:flex cpk:items-center cpk:justify-between cpk:gap-3 cpk:cursor-pointer"
-            style={{ display: "flex", alignItems: "center", justifyContent: "space-between", gap: "12px", cursor: "pointer" }}
             onClick={() => setIsExpanded(!isExpanded)}
           >
-            <div className="cpk:flex cpk:items-center cpk:gap-2 cpk:min-w-0" style={{ display: "flex", alignItems: "center", gap: "8px", minWidth: 0 }}>
+            <div className="cpk:flex cpk:items-center cpk:gap-2 cpk:min-w-0">
               <svg
                 className={`cpk:h-4 cpk:w-4 cpk:text-zinc-500 cpk:dark:text-zinc-400 cpk:transition-transform ${
                   isExpanded ? "cpk:rotate-90" : ""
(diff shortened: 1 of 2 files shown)


nuthalapativarun:
Gentle bump — this fixes the tool call icons getting squashed instead of laying out in a flex row. Vercel shows a failure but that's only the preview-deploy auth gate on forked PRs, not an actual issue with the change. Would appreciate a look when you have time.

nuthalapativarun:
Hi team — following up on PR #5485 (tool call icons flex row, closes #4777). Mergeable. 2nd bump. Thanks!

BenTaylorDev:
Thanks @nuthalapativarun — the diagnosis here is correct (v2 tool-card components rely on `cpk:flex` classes that aren't loaded in a v1-styled app), but the fix is on the wrong component. `WildcardToolCallRender` isn't actually on the render path — it has no runtime consumers. The component that renders the built-in wildcard tool card is `DefaultToolCallRenderer` in `packages/react-core/src/v2/hooks/use-default-render-tool.tsx` (wired as the `name:"*"` renderer via `useDefaultRenderTool`), and it has the same missing-inline-fallback issue. Could you move the flex fallback there instead?

Also worth considering whether the real fix is ensuring the v2 component CSS loads in v1 installs, rather than inline-patching individual cards.

BenTaylorDev (changes requested):
Thanks — and the good news is there's no drift here (the file is unchanged on `main`, merges clean). But I think this is aimed at the wrong surface, so I'd like to sort that out before it goes further:

- **Surface mismatch.** This patches the **v2** renderer `packages/react-core/src/v2/components/WildcardToolCallRender.tsx`, but issue #4777 is reported against the **v1 react-ui** stack (`@copilotkit/react-ui@1.57.0`; Ollama + `query_data`). `git grep` finds *zero* references to `WildcardToolCallRender` in `packages/react-ui/src` — the v1 render path is separate (`RenderMessage.tsx` → `AssistantMessage` with `generativeUI`). So the component you're fixing isn't the one that produced the reported screenshot.
- **Partial cosmetic patch.** Even on its own terms, the change adds inline `display:flex` fallbacks to only 2 of ~10 `cpk:`-styled elements; the status badge, card chrome, and expanded grid get none — so without the v2 stylesheet the component still renders broken. That reads more like papering over a missing `@copilotkit/react-ui/v2/styles.css` import than fixing a layout bug.

**Could you reproduce #4777 as reported (v1 react-ui stack) and confirm which component actually renders the stacked icons?** If it's the v1 `AssistantMessage`/`generativeUI` path (as the version block and grep suggest), this PR doesn't address the issue and should be re-targeted (or closed). If it genuinely is the v2 renderer, it needs to cover all the styled elements — or the real fix is to require/document the v2 stylesheet import.

nuthalapativarun:
Thanks for tracing this to the actual render path! Pushed a fix that moves the inline `display:flex` fallback from `WildcardToolCallRender` (confirmed no runtime consumers, reverted that no-op change) to `DefaultToolCallRenderer` in `use-default-render-tool.tsx`, which is the component actually wired as the `name:"*"` renderer via `useDefaultRenderTool`. Added the same fallback to its header `<button>` and inner icon-row `<div>`, which carry the same `cpk:flex` classes with no inline fallback.

Didn't extend this to a full v2-stylesheet-import requirement/doc change — figured the inline fallback (matching the pattern the original PR already used) was the more contained fix, but happy to discuss that direction if you'd still prefer it.

BenTaylorDev:
Re-reviewed this, and I owe you an apology: my 2026-07-20 redirect was wrong. You did exactly what I asked — moved the fallback to `DefaultToolCallRenderer` — but that isn't the component in the issue either. I sent you to the wrong place twice, and this PR should be closed as a result. Sorry for the wasted round trip.

Here's what actually renders the screenshot in #4777. It shows a native `<details>` disclosure triangle, a check, a **wrench**, the tool name, a ChevronDown, and expanded rows reading `prompt:"…"` / `state:{2 anahtar}` (Chrome page-translated "keys"). That is `ToolReasoning` in `examples/integrations/*/src/components/tool-rendering.tsx:37-70` — the *example app's own* component, including its `formatValue` producing `` `{${Object.keys(value).length} keys}` ``. `DefaultToolCallRenderer` has no wrench, no check, no `<details>`, and renders a "Done"/"Running" text badge, none of which appear in the screenshot.

It also can't be on that path: the example registers its own render at `examples/integrations/langgraph-python/src/hooks/use-generative-ui-examples.tsx:63-69`, and `use-default-render-tool.tsx:140-142` only falls back to `DefaultToolCallRenderer` when no user `render` is supplied. So the patched component is bypassed in the reported setup.

Two other things worth recording for anyone who picks this up later:

1. **The fallback doesn't hold up even in the no-CSS case.** I rendered the header row in a browser with the `cpk:` utilities absent. The chevron `<svg>` carries no `width`/`height` attributes, so it has no intrinsic size — it measures **1097×1097** with the inline flex applied and 1184×1184 without it, and the status dot `<span>` collapses to **0×0** and disappears. Two `display:flex` declarations don't produce a usable card here; intrinsic `width`/`height` on the `<svg>` plus explicit dot sizing would be the actual shape of that fix.
2. **The inline gaps override the theme rather than backstopping it.** Inline styles beat class rules, so `gap: "10px"`/`"8px"` win over `cpk:gap-2.5`/`cpk:gap-2` whenever the stylesheet *is* loaded, pinning rem-based spacing to px and breaking root-font-size scaling. That inverts the description in the PR body.

I've posted the real root cause on #4777 (the reporter's project isn't compiling Tailwind at all — that's why the native `<details>` marker shows and the lucide icons sit at their default 24px). Closing this one; thanks for the persistence, and again, sorry it was my bad trace that cost you the second pass.

PR #5486, opened 2026-06-16
docs(shell-docs): emit canonical versioned URLs for reference pages in llms-full.txt

### LINKED ISSUE #3385: 📚 Documentation: The llms-full.txt file is out of sync with the online documentation.

I noticed that the content in:

- https://docs.copilotkit.ai/llms-full.txt

is not in sync with the online docs at:

- https://docs.copilotkit.ai/

I ran into this while upgrading from 1.10.6 to 1.53.0 and looking for “Migrate to V2” guidance.

In the online docs, migration info references @copilotkit/react-core/v2, but in llms-full.txt there seems to be no mention of @copilotkit/react-core/v2 at all.

This causes issues for workflows/tools that rely on llms-full.txt as a complete source of documentation.

Could you clarify:

1. Is llms-full.txt expected to be fully synchronized with the online docs?
2. If yes, could you update/regenerate it to include the V2 migration content?
3. If not, is there a recommended machine-readable source that is guaranteed to stay current?

nuthalapativarun:
## What does this PR do?

`getAllLlmPages()` in `llm-text.ts` previously walked `src/content/reference/` directly and emitted reference pages at `reference/<slug>` (e.g. `reference/hooks/useCopilotAction`). The live site serves those pages at their versioned canonical URL — `/reference/v2/hooks/useCopilotAction` for the current v2 API — so `llms-full.txt` contained non-canonical source URLs that diverged from what users see in the browser.

**Root cause:** The v2 API reference lives at the _root_ of `src/content/reference/` (no `v2/` subfolder), so a bare filesystem walk cannot distinguish v2 from older SDK versions. It emits `reference/hooks/foo` instead of the correct `reference/v2/hooks/foo`.

**Fix:** Replace step 3 with an enumeration via `loadReferenceVersionItems` (which already knows the canonical URL per version) and `resolveReferencePage` (which resolves the content file path). This matches the URL scheme used by the `/reference/[...slug]` route handler.

**Result:**
- v2 hooks/components now appear at `reference/v2/hooks/...` in `llms-full.txt`
- v1, react-native, core, and bot pages appear at their correct versioned prefixes
- Version root index pages (`reference/v2`, `reference/v1`, ...) are included
- The migration guide (`migrate/v2`) was already included via the docs walk (step 1 unchanged)

## Related PRs and Issues
- Closes #3385

## Checklist
- I have read the Contribution Guide
- If the PR changes or adds functionality, I have updated the relevant documentation
- "Allow edits by maintainers" is checked

### DIFF
--- a/showcase/shell-docs/src/lib/llm-text.ts
+++ b/showcase/shell-docs/src/lib/llm-text.ts
@@ -37,9 +37,13 @@ import matter from "gray-matter";
 import { CONTENT_DIR, inlineSnippets, loadDoc } from "./docs-render";
 import { getDocsFolder, getIntegrations } from "./registry";
 import {
+  REFERENCE_VERSIONS,
+  loadReferenceVersionItems,
+  resolveReferencePage,
+} from "./reference-items";
+import {
   AG_UI_CONTENT_DIR,
   DOCS_CONTENT_DIR,
-  REFERENCE_CONTENT_DIR,
   walkMdx,
 } from "./sitemap-helpers";
 import demoContent from "@/data/demo-content.json";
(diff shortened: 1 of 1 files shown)


nuthalapativarun:
Following up on this — it makes llms-full.txt emit canonical versioned URLs for the reference pages instead of relative ones, which should help downstream consumers. The red Vercel check is just the fork-PR auth gate, not a real build problem. Thanks for taking a look eventually!

nuthalapativarun:
Hi team — following up on PR #5486 (canonical versioned URLs in llms-full.txt). Mergeable. 2nd bump. Happy to adjust if needed. Thanks!

BenTaylorDev (approved):
**Approving.** Reviewed the diff and verified it end-to-end by building both the old and new reference-URL sets against the real content tree.

The fix is correctly targeted: it replaces the bare `walkMdx(REFERENCE_CONTENT_DIR)` step-3 walk with enumeration via `loadReferenceVersionItems` + `resolveReferencePage`, so `llms-full.txt` emits the same canonical versioned URLs the `/reference/[...slug]` route serves.

Verified behavior:
- **25 v2 pages relocated** from bare `reference/hooks/…` / `reference/components/…` to canonical `reference/v2/…`, plus the `reference` root → `reference/v2`.
- **1 duplicate eliminated** — the old walk emitted both `reference/hooks/useCapabilities` and `reference/v2/hooks/useCapabilities`; the new logic collapses to the single canonical v2 URL (net −1, 173 → 172).
- **No dangling URLs** — every emitted reference URL resolves to a real content file; the PR's set is exactly the canonical set derived from `referenceHref()` per version (nothing extra, nothing missing).
- **All other SDKs untouched** — v1, angular, vue, channels, core, react-native were already versioned correctly and are identical before/after.
- **Content-read path is safe** — `readSource` reads reference pages via `page.filePath` (set to the resolved real file), and the new `loadSlug` `contentSlug` matches the file's physical-relative path, so snippet inlining is unaffected.
- Dropping the bare `/reference` URL is correct, not a regression — `referenceStaticParams` never generates a version-less `/reference` route.

Single-file change, merges cleanly into `main`. Closes #3385. LGTM.

PR #5487, opened 2026-06-16
fix(react-ui): wrap CopilotKit CSS selectors in :where() to reduce specificity

### LINKED ISSUE #263: When overriding CSS properties, sometimes the overridden css property does not take precedence

nuthalapativarun:
## What does this PR do?
Wraps CopilotKit namespace CSS selectors (`.copilotKit*`, `.cpk-*`) in the `:where()` pseudo-class, which has zero specificity. This allows users to override CopilotKit styles with their own selectors without needing `!important`. Previously, compound selectors like `.copilotKitButton .icon` (specificity 0,2,0) would win over a user's single-class selectors (0,1,0). With `:where()`, the specificity drops to the element/combinator level only. Fixes #263.

## Related PRs and Issues
- Closes #263

## Checklist

### DIFF
--- a/packages/react-core/src/v2/styles/globals.css
+++ b/packages/react-core/src/v2/styles/globals.css
@@ -396,7 +396,7 @@
  * historical-replay turn renders directly in its settled state with no
  * animation — which is the desired behavior.
  */
-.cpk-intelligence-indicator {
+:where(.cpk-intelligence-indicator) {
   position: relative;
   display: inline-flex;
   align-items: center;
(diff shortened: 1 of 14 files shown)


nuthalapativarun:
Just circling back on this — wrapping the CopilotKit CSS selectors in :where() should fix the specificity conflicts folks have been hitting when overriding styles. The Vercel failure is only the auth-gated preview for fork PRs, the actual build passes. Happy to tweak if you want a different approach.

nuthalapativarun:
Hi team — following up on PR #5487 (CSS selector specificity fix, closes #263). Mergeable. 2nd bump. Thanks!

BenTaylorDev:
Thanks for looking at #263. We're going with the more targeted approach in #5805 — it lowers specificity only on the sidebar-descendant selectors (keeping a floor above host element styles) and adds a regression test, a smaller blast radius than wrapping all selectors in `:where()` (which would let host resets / Tailwind preflight bleed into component internals). Closing in favor of #5805.

Heads-up that #263 likely isn't fully closed by either approach — a `@layer` cascade layer is probably the complete fix — so we'll keep the issue open.

PR #5488, opened 2026-06-16
feat(bot-slack): native streaming + pending feedback in channels

AlemTuzlak:
## What

Make the native Slack experience work **everywhere**, not just inside the assistant pane — and show **pending feedback before the first token** in channels, the way Linear's Slack bot does.

When a thread uses native streaming but isn't a pane thread (e.g. the bot is mentioned in a regular channel):
- the legacy `:hourglass: thinking…` placeholder and `:wrench: Calling …` tool rows are gone, **and**
- the native streaming bubble is **opened eagerly at run start** (`chat.startStream`), so Slack renders its streaming indicator immediately — that's the "pending feedback" the user sees while the agent reasons / calls tools before the first token. The reply then continues in that same bubble.

## Why

`assistant.threads.setStatus` (the "is thinking…" composer indicator) is **assistant-pane-only** per Slack's API — there's no equivalent surface in a channel. But `chat.startStream` "creates a placeholder message that updates in real-time" and works in channels with `recipient_user_id` + `recipient_team_id` (both already wired). Opening it eagerly is the native way to show progress in a channel without reintroducing legacy chrome.

## Behavior

In a **channel with native streaming** (this PR): run start opens `chat.startStream` so the streaming bubble appears immediately; the bubble keeps showing through tool calls; the reply adopts that same bubble. **Pane** mode (unchanged) uses `setStatus`. **Legacy / flat DM** (unchanged) uses the `:hourglass:` placeholder + `:wrench:` rows + `chat.update`.

`captureToolCall` still runs in all modes, so the run-loop's tool dispatch is unaffected. Pane mode still wins when both could apply. A run that produces **no** text reply discards the empty eagerly-opened bubble (`chat.delete`) so nothing is left behind. Priming and discarding are degrade-never-throw (a `startStream` failure falls over to the legacy transport).

## Changes

- `packages/bot-slack/src/native-stream.ts` — `NativeMessageStream.prime()` (idempotent eager open, fails over to legacy on error); optional `NativeStreamTransport.discardStream`; `finish()` deletes an empty primed bubble instead of finalizing an empty message.
- `packages/bot-slack/src/event-renderer.ts` — `nativeNoStatus` channel mode: prime an `eagerStream` on `RUN_STARTED`, adopt it for the first text message (no second `startStream`), discard it on `RUN_FINISHED` / `markInterrupted` (and defensively before re-priming) when no reply arrived.
- `packages/bot-slack/src/adapter.ts` — `nativeTransport` provides `discardStream` via `chat.delete`.
- Tests — `native-stream.test.ts` (prime/adopt/discard/failover) and `event-renderer.test.ts` (eager open, reply adopts the bubble, empty run discards it, no legacy chrome, tool calls still captured).

## Verification

- `nx run @copilotkit/bot-slack:test` — 207 passed.
- `nx run @copilotkit/bot-slack:build` — clean.
- Code review (pr-review-toolkit) — no material issues.

**E2E (user-owned):** mention the bot in a regular channel → the native streaming bubble appears immediately (before the first token), then fills in; no `:hourglass:`/`:wrench:` rows. A run that produces no text → no leftover empty bubble. Pane behavior unchanged. *If the empty bubble renders blank rather than a visible loading affordance, the one-line follow-up is to seed `prime()` with an initial `markdown_text` status.*

## Follow-up (release)

Needs a `@copilotkit/bot-slack@0.0.3` publish + `examples/slack` bump + Railway redeploy to reach the deployed bot. `0.0.3` should also fix bot-slack pinning `@copilotkit/bot@0.0.1` exactly → `~0.0.2`.

### DIFF
--- a/packages/bot-slack/src/__tests__/event-renderer.test.ts
+++ b/packages/bot-slack/src/__tests__/event-renderer.test.ts
@@ -519,3 +519,87 @@ describe("createRunRenderer — assistant pane status mode", () => {
     expect(f.statuses.at(-1)?.status).toBe("");
   });
 });
+
+describe("createRunRenderer — native streaming in a channel (no status)", () => {
+  // A non-pane thread (e.g. a channel mention) with native streaming: Slack
+  // has no composer-status surface here, so the renderer posts NO "thinking…"
+  // placeholder and NO `:wrench:` tool rows — the native stream is the only
+  // surface. Tool calls are still captured for the run-loop.
+  function makeNativeTransport() {
+    let counter = 0;
+    const appended: string[] = [];
+    const transport = {
+      startStream: vi.fn(async () => {
+        counter++;
+        return `ns${counter}`;
+      }),
+      appendStream: vi.fn(async (_ts: string, md: string) => {
+        appended.push(md);
+      }),
+      stopStream: vi.fn(async (_ts: string) => {}),
+    };
+    return { transport, appended };
+  }
+
+  it("posts no placeholder on run start", async () => {
+    const f = makeFakeClient();
+    const nt = makeNativeTransport();
+    const { subscriber: sub } = createRunRenderer({
+      client: f.client,
+      target: { channel: "C1", threadTs: "100.0" },
+      nativeStreaming: { transport: nt.transport },
+    });
+    await sub.onRunStartedEvent!({} as never);
+    expect(f.posts).toHaveLength(0);
+  });
+
+  it("posts no :wrench: rows for tool calls, but still captures them", async () => {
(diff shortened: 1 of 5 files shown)


AlemTuzlak:
Closing — wrong approach. This tried to replace the thinking placeholder with an eagerly-opened native stream, but for card answers (no streamed text) that left an undeletable empty "Thinking…" bubble (chat.delete → cant_delete_message). Native streaming + the placeholder already work in channel threads on main; the only real gap was noisy per-tool rows, addressed in the new PR by setting showToolStatus: false on the example.

PR #5493, opened 2026-06-16
docs(react-ui): clarify all modalities in attachments prop examples

nuthalapativarun:
## Summary

The `attachments` prop supports images, audio, video, and documents — but the JSDoc example in `Chat.tsx` only showed `image/*,application/pdf`, and the docs configuration example used `accept: image/*`, silently teaching users to restrict themselves to images.

**Before (Chat.tsx JSDoc):**
```tsx
accept: image/*,application/pdf,
```

**After:**
```tsx
accept: image/*,audio/*,video/*,application/pdf,
```

The docs configuration example now also clarifies that omitting `accept` defaults to `*/*` (all files), and the shown value includes all four supported modalities.

## Changes

- `packages/react-ui/src/components/chat/Chat.tsx` — updated JSDoc example to show all modalities; added note that default `accept` is `*/*`
- `showcase/shell-docs/src/content/docs/multimodal-attachments.mdx` — updated configuration example to show `image/*,audio/*,video/*,application/pdf` and note that omitting `accept` allows all types

### DIFF
--- a/packages/react-ui/src/components/chat/Chat.tsx
+++ b/packages/react-ui/src/components/chat/Chat.tsx
@@ -236,16 +236,19 @@ export interface CopilotChatProps {
    * Configuration for file attachments in the chat input.
    * Enables users to attach images, audio, video, and documents.
    *
+   * Supports images, audio, video, and documents. Omit `accept` to allow all
+   * file types (default: `"*\/*"`), or restrict with a MIME filter.
+   *
    * @example
    * ```tsx
    * <CopilotChat
    *   attachments={{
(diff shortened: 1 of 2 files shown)


nuthalapativarun:
Quick follow-up — this just clarifies the attachments prop docs to cover all the modalities, since the examples were only showing one. Vercel's check is red only because of the fork auth gate, the actual docs build is fine. Whenever you have a moment to glance at it, thanks!

onsclom (approved):
Thank you for catching and fixing this!

PR #5494, opened 2026-06-16
feat(runtime): add url field to ImageMessageInput

### LINKED ISSUE #1967: 🚀 Feature Request: ImageMessage should support url as input
### 🎤 Tell us your idea

ref code: https://github.com/vercel/ai/blob/main/packages/openai/src/convert-to-openai-chat-messages.ts#L66-L81

nuthalapativarun:
## Summary

Adds `url` field to `ImageMessageInput` so images can be sent by URL instead of requiring base64 `bytes` + `format`.

**Before:** `imageMessage` only accepted `{ format, bytes }` — every image had to be base64-encoded by the client before sending.

**After:** `imageMessage` accepts `{ url }` as an alternative:

```graphql
mutation {
  generateCopilotResponse(data: {
    messages: [{
      id: img-1
      imageMessage: {
        role: user
        url: https://example.com/diagram.png
      }
    }]
  }) { ... }
}
```

Both OpenAI and Anthropic service adapters route URL images through the native URL source type (no base64 encoding needed):

- **OpenAI**: `{ type: image_url, image_url: { url } }`
- **Anthropic**: `{ type: image, source: { type: url, url } }`

## Files changed

- `packages/runtime/src/graphql/inputs/message.input.ts` — add `url?: string`, make `format`/`bytes` nullable
- `packages/runtime/src/graphql/types/converted/index.ts` — add `url?: string` to `ImageMessage`
- `packages/runtime/src/service-adapters/conversion.ts` — pass `url` through conversion
- `packages/runtime/src/service-adapters/openai/utils.ts` — use URL directly when present
- `packages/runtime/src/service-adapters/anthropic/utils.ts` — use Anthropic URL source when present
- `packages/runtime/src/graphql/message-conversion/agui-to-gql.ts` — accept url-only image messages
- `packages/runtime/src/graphql/message-conversion/gql-to-agui.ts` — pass url through; skip format validation for url-only path
- `packages/runtime/src/graphql/message-conversion/roundtrip-conversion.test.ts` — add URL roundtrip test

## Test plan

- 1528 tests pass (1527 existing + 1 new URL roundtrip test)
- Build passes (tsdown, no errors)
- Lint passes (0 new warnings, 0 errors)

Closes #1967

### DIFF
--- a/packages/runtime/src/graphql/inputs/message.input.ts
+++ b/packages/runtime/src/graphql/inputs/message.input.ts
@@ -29,7 +29,7 @@ export class ActionExecutionMessageInput {
     nullable: true,
     deprecationReason: "This field will be removed in a future version",
   })
-  scope?: String;
+  scope?: string;
 }
 
 @InputType()
@@ -76,11 +76,14 @@ export class AgentStateMessageInput {
 
 @InputType()
 export class ImageMessageInput {
-  @Field(() => String)
-  format: string;
+  @Field(() => String, { nullable: true })
+  format?: string;
 
-  @Field(() => String)
-  bytes: string;
+  @Field(() => String, { nullable: true })
+  bytes?: string;
+
+  @Field(() => String, { nullable: true })
+  url?: string;
 
   @Field(() => String, { nullable: true })
   parentMessageId?: string;
--- a/packages/runtime/src/graphql/message-conversion/agui-to-gql.ts
+++ b/packages/runtime/src/graphql/message-conversion/agui-to-gql.ts
@@ -30,14 +30,15 @@ function hasImageProperty(
     return false;
   }
 
-  const image: { format?: string; bytes?: string } | undefined =
-    "image" in message ? message.image : undefined;
+  const image: { format?: string; bytes?: string; url?: string } | undefined =
+    "image" in message ? (message as any).image : undefined;
   if (image === undefined) {
     return false;
   }
 
-  const isMalformed = image.format === undefined || image.bytes === undefined;
-  if (isMalformed) {
+  const hasBase64 = image.format !== undefined && image.bytes !== undefined;
+  const hasUrl = image.url !== undefined;
+  if (!hasBase64 && !hasUrl) {
     return false;
   }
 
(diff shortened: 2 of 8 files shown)


nuthalapativarun:
Checking in on this — adds a url field to ImageMessageInput so images can be referenced by URL instead of requiring inline data. The Vercel preview failure is just the auth gate for fork PRs; the real build succeeds. Appreciate a review whenever convenient.

BenTaylorDev:
Thanks for this @nuthalapativarun — the `url` field is genuinely missing and the feature is useful. A few things to sort before it can land:

1. **The `as any` casts hide a real type gap.** `gql-to-agui.ts` (`image: { url } as any`) and `agui-to-gql.ts` (`(message.image as any).url`) are needed because the AG-UI `ImageMessage` type has no `url` field — so `url` only really exists in the GQL layer and is force-cast across the AG-UI boundary. Can you either thread a properly-typed field through (or confirm AG-UI's type supports it) rather than casting? As-is, downstream correctness is unverified.
2. **Schema change.** Flipping `format`/`bytes` from required to nullable on the GraphQL `ImageMessageInput` is a public-schema change, not just an additive field — worth calling out explicitly so we can confirm it's acceptable for persisted-query/schema consumers.
3. **Scope.** The diff also carries unrelated churn (`scope` type tweak, `catch (e)`→`catch`, `import`→`import type` in conversion/anthropic utils). Please split those out so the PR is just the feature.
4. **Coverage.** The test only exercises gql→agui `url`; it'd be good to assert the OpenAI `image_url.url` and Anthropic `source:{type:'url'}` adapter output too (and note Google/Bedrock don't handle image messages at all, so url images silently drop there — same as base64 today, but worth a doc note).

BenTaylorDev:
Thanks — URL image input is a good capability to have, and it's actually already on `main` through a more general mechanism: `@ag-ui/core`'s `InputContentSource` has a `type: "url"` variant, and the attachments layer consumes it directly (`getSourceUrl` at `packages/shared/src/attachments/utils.ts:146`). So #1967 is covered on the supported path.

This PR instead adds `url` to the older base64 `ImageMessageInput` / `ImageData` GraphQL path, which is explicitly deprecated in favor of `InputContentSource` (`packages/shared/src/types/message.ts` marks `ImageData` `@deprecated @since 1.56.0` with a migration guide). To make it fit that path it threads `url` through the AG-UI leg via `as any` casts, plus some unrelated reflow churn.

Since URL images already work via `InputContentSource`/attachments, I'm closing this as superseded rather than extending the deprecated base64 type. If there's a gap in the attachments path for your use case, a fresh issue against that mechanism would be the way to go — thanks for picking this up.

PR #5504, opened 2026-06-16
docs(shell-docs): update useRenderToolCall reference + schema-free tool-render opt-out

NathanTarbert:
## What

Updates the v2 `useRenderToolCall` reference page and makes opting a tool out of the default rendering a single, schema-free call.

## Why

The reference page had drifted out of sync with the hook implementation (`packages/react-core/src/v2/hooks/use-render-tool-call.tsx`) — most notably `toolCallId` in the render props, added after the doc was last touched. While documenting how to opt out of rendering, the natural example (`useRenderTool({ name: "...", render: () => <></> })`) only type-checked for the wildcard `"*"`; a named tool required a `parameters` schema, forcing a throwaway `z.any()`. This PR re-aligns the doc and removes that rough edge.

### Docs (`showcase/shell-docs`)
- `useRenderToolCall.mdx`:
  - Document `toolCallId` in the render-prop shape (previously undocumented).
  - Describe agentId-scoped lookup priority: agent-specific → unscoped → wildcard `"*"` → built-in `DefaultToolCallRenderer`.
  - Note args are parsed with `partialJSONParse` (streaming), not strict `JSON.parse`.
  - Correct `toolCall` prop to `toolCall.function.name` / `toolCall.function.arguments`.
  - Rewrite the Status Resolution table to match real logic (`toolMessage` presence + provider executing set).
  - New **"Disable default tool rendering"** section, ordered least→most specific: wildcard first (all tools), then a **"For specific tools"** subsection. Both use a schema-free `useRenderTool` call; dropped the old `useFrontendTool` handler/schema boilerplate.
- `useRenderTool.mdx`: document the render-only (no-schema) named overload.

### react-core
- Make `parameters` optional on the named `useRenderTool` overload, mirroring the existing wildcard path; `defineToolCallRenderer` defaults the args schema to `z.any()` when none is given.
- This lets `useRenderTool({ name: "myTool", render: () => <></> }, [])` type-check with no Zod schema. Typed `parameters` behavior is unchanged.
- Added a test for the named render-only registration.

## Verification

- `@copilotkit/react-core` tests pass (1280) including the new case; `build` (tsc) passes.
- Opt-out snippets type-checked in-package (`tsc`): wildcard, specific-name (no schema), and named-with-schema all compile.
- `oxlint` (shell-docs) passes — 0 errors.
- Previewed locally at `/reference/hooks/useRenderToolCall`.

### DIFF
--- a/showcase/shell-docs/src/content/reference/hooks/useRenderToolCall.mdx
+++ b/showcase/shell-docs/src/content/reference/hooks/useRenderToolCall.mdx
@@ -176,7 +176,7 @@ function App() {
 
 ### Opting Out of Rendering
 
-Because the hook falls back to the built-in `DefaultToolCallRenderer` when no renderer matches, tool calls paint UI out-of-the-box. To suppress that — render nothing for a tool call — register a renderer whose `render` returns `null`. A registered renderer (even one returning `null`) takes priority over the default, so no fallback UI is shown.
+Because the hook falls back to the built-in `DefaultToolCallRenderer` when no renderer matches, tool calls paint UI out-of-the-box. To suppress that — render nothing for a tool call — register a renderer whose `render` produces no UI (return `null`, or an empty fragment `<></>`). A registered renderer takes priority over the default, so no fallback UI is shown.
 
 Opt out of a **single tool** by name:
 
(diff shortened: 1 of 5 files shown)


showcase/shell-docs/src/content/reference/hooks/useRenderToolCall.mdx
```diff
@@ -168,14 +168,54 @@ function App() {
...
+        /* ... */
+      },
+      render: () => null,
+    },
+    [],
+  );
+
+  return <ChatInterface />;
+}
+```
```
tylerslaton: I recommend moving all of this a sub-section below called

```
#### Opting out of rendering for a specific tool
```

And move everything below this to be above this as a parent section.

The reason for this is you're going for least specific to most specific - this logically tracks for folks.

End state would be...

(Updating the headers a bit here too)
```
### Disable default tool rendering
...

#### For specific tools
...
```

showcase/shell-docs/src/content/reference/hooks/useRenderToolCall.mdx
```diff
@@ -168,14 +168,54 @@ function App() {
...
+```tsx
+function App() {
+  // Run the tool, but render nothing for it.
+  useFrontendTool(
+    {
+      name: "searchDatabase",
+      parameters: z.object({ query: z.string() }),
+      handler: async ({ query }) => {
+        /* ... */
+      },
```
tylerslaton: None of this is needed for what we're trying to show here, we can remove it I'd say.

showcase/shell-docs/src/content/reference/hooks/useRenderToolCall.mdx
NathanTarbert: Done. Renamed to **Disable default tool rendering** and reordered least→most specific: wildcard (all tools) first, then a **For specific tools** subsection (`#### For specific tools`).

showcase/shell-docs/src/content/reference/hooks/useRenderToolCall.mdx
NathanTarbert: Removed. Both examples are now a single schema-free `useRenderTool({ name, render: () => <></> }, [])` call — no handler/parameters boilerplate. To make the specific-tool case work without a throwaway `z.any()`, I made `parameters` optional on the named `useRenderTool` overload (mirrors the wildcard path; defaults to `z.any()`). Covered by a new test.

tylerslaton (changes requested):
Thanks Nathan, I'm not sure why we made changes to the react-core package. Can we remove those? Then we can merge.

showcase/shell-docs/src/content/reference/hooks/useRenderToolCall.mdx
NathanTarbert: Updated

showcase/shell-docs/src/content/reference/hooks/useRenderToolCall.mdx
NathanTarbert: Updated

NathanTarbert:
Done - reverted all the react-core changes, this is docs-only now (no `packages/` diff). 

The wildcard opt-out stays clean: `useRenderTool({ name: "*", render: () => <></> }, [])`. Since the named `useRenderTool` overload still requires a schema, the per-tool example passes a pass-through `parameters: z.any()`. Let me know if you if you need any other changes.

tylerslaton (approved):

PR #5508, opened 2026-06-16
showcase(claude): add SDK demo parity

tylerslaton:
## Summary

- Productizes the Claude SDK Python and TypeScript showcase demos with LangGraph-parity frontends.
- Wires the Claude demo backends through the official Claude Agent SDK/AG-UI adapter paths using `claude-sonnet-4.6`.
- Keeps Claude integration docs hidden for this PR and excludes generated/authored docs artifacts from scope.

## Why

The goal is to bring the productized LangGraph demo surface to Claude Agents SDKs without publishing integration docs in this pass. This keeps the PR focused on local showcase demos, runtime behavior, fixtures, and validation support.

## How

- Ported the demo frontend surfaces and local shell-dojo support for Claude SDK Python/TypeScript.
- Added official Claude SDK adapter/backend wiring plus real-Claude local compose support.
- Updated Claude aimock fixtures and validation ratchets for the expanded demo set.
- Set both Claude manifests to `docs_mode: hidden` and removed docs setup/snippet artifacts from the PR scope.


jpr5 (approved):
Approving.

Reviewed the full diff — logic is correct and well-structured, model normalization (`claude-sonnet-4.6` → `claude-sonnet-4-6`) is applied consistently at every API boundary, the OpenAI→Anthropic migration in `generate_a2ui` is clean (correct `parameters`→`input_schema` / `tool_choice` remap), and `mcp_apps_agent.py` now preserves `tool_use`/`tool_result` as structured turns instead of flattening them. Change is scoped entirely to `showcase/integrations/claude-sdk-*` with no V1/V2 production package touched. CI fully green.

Confirmations run locally:
- `pnpm test aimock-fixtures` → 823 passed. The bumped collision ceilings (297, 142) are correct as committed.
- Checked the `openai` dependency I'd initially flagged as dead — it's NOT dead. It backs the `copilotkit-voice` transcription route in both packages (`new OpenAI({ apiKey })`). Retracting that; the dependency is required.

Non-blocking follow-up (fine as a later PR): the hand-rolled partial-JSON string parsers (`_decode_partial_json_string` / `partialJsonStringProperty`) and `sanitize_unresolved_tool_uses` power live shared-state streaming but have no direct unit tests. Logic traced as sound; worth covering directly at some point.

PR #5542, opened 2026-06-18
fix(core): allow clearing headers via setHeaders with null/undefined

### LINKED ISSUE #5535: 🐛 Bug: Cannot clear Authorization header using setHeaders with null value
### ♻️ Reproduction Steps

Reference: `provider-setup.md` reference in `react-core` agent skill

Create `AuthTokenSync` component following the skill.

`copilotkit.setHeaders(...)` requires non-null values for its headers dictionary (otherwise typescript shows errors). There is no way to clear the `Authorization` header when the user wants to logout.

### ✅ Expected Behavior

The copilotkit react sdk allows clearing the Authorization header

### ❌ Actual Behavior

It is not possible to clear the authorization header. I can set it to an empty string, but then the header is not removed.

### Screenshots

_No response_

### 𝌚 CopilotKit Version

```shell
1.60.1
```

### 📄 Logs (Optional)

```shell

```

onsclom:
## Summary

Fixes #5535.

`CopilotKitCore.setHeaders` was typed `Record<string, string>`, so there was no type-safe way to clear a header like `Authorization` on logout. `null` was a TS error, and an empty string leaves the header present with a blank value.

## Change

- Widen `setHeaders` to `Record<string, string | null | undefined>` and drop any `null`/`undefined` entry. A shared `normalizeHeaders` helper enforces the same string-only invariant at both write paths (the constructor and `setHeaders`).
- `setHeaders` stays a full overwrite, so clearing one header while keeping the rest uses the spread pattern:
  ```ts
  copilotkit.setHeaders({ ...copilotkit.headers, Authorization: token ? `Bearer ${token}` : null });
  ```
- Update the `react-core` `AuthTokenSync` skill example to show the logout/clear path, and warn that a header must not be managed via both the `headers` prop and imperative `setHeaders` (the provider re-applies its prop-derived headers as a full overwrite whenever its inputs change).

## Tests

Added `packages/core` coverage: drop `null`/`undefined` keys, empty-string preservation, overwrite-not-merge semantics, single-header clear via spread, `onHeadersChanged` notification, and propagation to local and remote (`ProxiedCopilotRuntimeAgent`) agents.

## Notes

- No API break: `Record<string, string>` is assignable to the widened type, so existing callers are unaffected.
- The second commit syncs the plugin manifest version (`plugin.json` + `marketplace.json` `plugins[0].version`) to `1.60.2` via `pnpm sync:plugin-skills`. This drift pre-existed on `main` and surfaced in CI only because this PR touches skill files; it is unrelated to the fix.

### DIFF
--- a/packages/core/src/__tests__/core-headers.test.ts
+++ b/packages/core/src/__tests__/core-headers.test.ts
@@ -1,5 +1,6 @@
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import { CopilotKitCore } from "../core";
+import { ProxiedCopilotRuntimeAgent } from "../agent";
 import { HttpAgent } from "@ag-ui/client";
 import { waitForCondition } from "./test-utils";
 
@@ -29,6 +30,8 @@ describe("CopilotKitCore headers", () => {
 
   it("includes provided headers when fetching runtime info", async () => {
     const fetchMock = vi.fn().mockResolvedValue({
+      ok: true,
+      status: 200,
       json: vi.fn().mockResolvedValue({ version: "1.0.0", agents: {} }),
     });
     global.fetch = fetchMock as unknown as typeof fetch;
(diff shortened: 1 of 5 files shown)


tylerslaton (changes requested):
Thanks for the fix. The core change looks correct to me: `setHeaders` normalizes before storing, keeps the internal `Record<string, string>` invariant, reapplies the normalized map to local/remote `HttpAgent` instances, and the new tests cover the clear/spread/overwrite cases.

I am requesting changes because the live API reference is still stale for the public behavior this PR adds. `showcase/shell-docs/src/content/reference/core/classes/CopilotKitCore.mdx` still documents:

```ts
setHeaders(headers: Record<string, string>): void
```

But this PR widens the actual API to:

```ts
setHeaders(headers: Record<string, string | null | undefined>): void
```

The skill docs now tell users that `Authorization: null` is the supported clearing path, so the reference docs should match that before this lands. Please update the reference signature and add a short note that `null`/`undefined` drops the header key while an empty string remains a valid header value.

Once that doc mismatch is fixed, I would approve.

onsclom:
Thanks @tylerslaton — good catch. Updated the live reference at `showcase/shell-docs/src/content/reference/core/classes/CopilotKitCore.mdx` to the widened signature:

```ts
setHeaders(headers: Record<string, string | null | undefined>): void
```

and added a note that `null`/`undefined` drops the header key (the supported clearing path) while an empty string remains a valid header value. Pushed in 9287a0593.

tylerslaton (dismissed):

NathanTarbert:
Hi @onsclom, would you mind resolving the conflicts?

tylerslaton (approved):

PR #5586, opened 2026-06-19
docs(frontends): improve picker and language for non-react frontends

tylerslaton:
## Summary
- Fix frontend selection, logo, and docs navigation behavior in shell docs
- Update link rewriting, search href generation, SEO redirects, and sitemap handling for frontend docs paths
- Refresh related docs pages and tests across framework and frontend routes

## Testing
- Added and updated unit tests for frontend options, link rewriting, search hrefs, SEO redirects, and framework shell layout
- Added route-level coverage for the llms-mdx endpoint

### DIFF
--- a/showcase/scripts/generate-search-index.ts
+++ b/showcase/scripts/generate-search-index.ts
@@ -36,6 +36,17 @@ interface SearchEntry {
   href: string;
 }
 
+const FRONTEND_SEARCH_PAGES = [
+  { id: "vue", name: "Vue", guidanceTitle: "Docs status" },
+  { id: "react-native", name: "React Native", guidanceTitle: "Docs status" },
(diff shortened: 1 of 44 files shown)


samjulien (changes requested):
I found one redirect issue here: `/frontends/react/using-these-docs` now redirects to `/using-these-docs`, which 404s. The new `/react/using-these-docs` route appears to do the same thing via `reactRootPath()`.

Since `frontendPathForBackend()` already treats React `using-these-docs` as canonicalizing to `/`, I think this path should special-case that slug to `/` rather than preserving it as `/using-these-docs`.

I verified locally:
- `/frontends/react/using-these-docs` -> `/using-these-docs` -> 404
- `/react/using-these-docs` -> `/using-these-docs` -> 404

samjulien (approved):

PR #5592, opened 2026-06-19
fix: resolve useAgent agentId from chat configuration (#5533)

### LINKED ISSUE #5533: 🐛 Bug: CopilotKit React setup does not find non-default agent
### ♻️ Reproduction Steps

1. Create copilot runtime exposing an agent with a non-default name, e.g.
   ```
     agents: {
       TravelBookingAgent: new HttpAgent({ url: agentUrl }),
     },
   });
   ```
2. Configure CopilotKit frontend following documentation
   ```tsx
   <CopilotKit runtimeUrl="/copilotkit" useSingleEndpoint={false} onError={handleCopilotError}>
     <CopilotChat agentId="TravelBookingAgent" />
   </CopilotKit>
   ```

### ✅ Expected Behavior

Chat uses configured agent

### ❌ Actual Behavior

Frontend breaks and with error in browser console:
> Uncaught Error: useAgent: Agent 'default' not found after runtime sync (runtimeUrl=/copilotkit). Known agents: [TravelBookingAgent] Verify your runtime /info and/or agents__unsafe_dev_only.

### Screenshots

_No response_

### 𝌚 CopilotKit Version

```shell
"@copilotkit/react-core": "^1.60.1",
```

### 📄 Logs (Optional)

```shell

```

NathanTarbert:
## Summary

Fixes #5533

When a runtime registers an agent under a **non-default** name (e.g. `agents: { TravelBookingAgent }`) and the frontend renders `<CopilotChat agentId="TravelBookingAgent" />` without an `agent` prop on `<CopilotKit>`, the app throws after runtime sync:

> useAgent: Agent 'default' not found after runtime sync (runtimeUrl=…). Known agents: [TravelBookingAgent]

## Root cause

`useAgent()` (`packages/react-core/src/v2/hooks/use-agent.tsx`) resolved its `agentId` only from its own prop, falling back straight to `DEFAULT_AGENT_ID`. It never consulted the surrounding `CopilotChatConfigurationProvider` — even though `CopilotChat` installs that provider around its subtree with the resolved (non-default) agentId.

`CopilotChat` resolves its *own* `useAgent` call correctly, so the chat works. But any **descendant** that calls `useAgent()` without re-passing `agentId` (a custom message/tool-render component, a sibling hook) silently resolves to `'default'`. Once `/info` sync lands and the registry holds only the non-default agent, that consumer throws — which is why the thrown id is `'default'`, not `'TravelBookingAgent'`.

## Fix

Resolve `agentId` in `useAgent` with the same precedence `CopilotChat` already uses:

```ts
const resolvedAgentId = agentId ?? chatConfig?.agentId ?? DEFAULT_AGENT_ID;
```

The hook already imported and called `useCopilotChatConfiguration` (for `threadId`); the call is hoisted and reused — no duplicate hook call. An explicit `agentId` prop still wins; with no chat config it still falls back to `DEFAULT_AGENT_ID`. No changes to core or the providers.

## Tests added

`packages/react-core/src/v2/hooks/__tests__/use-agent-nondefault-agentid.test.tsx` — a `useAgent()` consumer inside a chat configured for `TravelBookingAgent` (runtime synced to `agents:{TravelBookingAgent}`) must not throw `Agent 'default' not found`, and must inherit `TravelBookingAgent`. Fails before the fix, passes after.

## Checklist

- [x] Failing test written and confirmed failing before the fix
- [x] Fix applied, test passes
- [x] Full `@copilotkit/react-core` suite passes (1291 passed)
- [x] Build succeeds (`nx build @copilotkit/react-core`)
- [x] Formatter passes (`pnpm format`)


packages/react-core/src/v2/hooks/use-agent.tsx
contextablemark, line 30:
```suggestion
   /**
   * Agent to bind to. Resolution precedence: this property, then the surrounding
   * chat configuration's agentId, then the global default.
   */
    agentId?: string;
```

contextablemark (changes requested):
A few issues for documentation accuracy - one I've attached as a suggestion to use-agent.tsx, the others are : 

[use-capabilities.tsx:5](https://github.com/CopilotKit/CopilotKit/blob/6d9397c42ea3823a71fb26272c1961cdafc8d00b/packages/react-core/src/v2/hooks/use-capabilities.tsx#L5)

Now: * Returns the capabilities declared by the given agent (or the default agent).
Fix: * Returns the capabilities declared by the given agent (or the agent resolved from the surrounding chat configuration, falling back to the default agent).

[use-capabilities.tsx:11](https://github.com/CopilotKit/CopilotKit/blob/6d9397c42ea3823a71fb26272c1961cdafc8d00b/packages/react-core/src/v2/hooks/use-capabilities.tsx#L11)

Now: * @param agentId - Optional agent ID. If omitted, uses the default agent.
Fix: * @param agentId - Optional agent ID. If omitted, inherits the surrounding chat configuration's agent, falling back to the default agent.

contextablemark:
Follow-up: the Vue `useAgent` (`packages/vue/src/v2/hooks/use-agent.ts`) has the identical #5533 bug — it's a separate implementation, so this React-core fix doesn't flow into it. Tracked separately in #5656 (kept out of this PR so it can ship with its own Vue reproduction test). `@copilotkit/react-native` re-exports react-core's hook, so it's covered by this PR transitively.

contextablemark (approved):
Approving. The core `use-agent.tsx` fix (resolve `agentId ?? chatConfig?.agentId ?? DEFAULT_AGENT_ID`, mirroring `CopilotChat`'s precedence) is clean — hooks stay unconditional, the memo dep array correctly swaps to `resolvedAgentId`, and an explicit prop still wins.

Nice catch extending this to `CopilotListeners`: that's the path that actually crashes the reporter's repro (no `agent` prop). Since the listener mounts as a sibling of `{children}` under the top-level chat config seeded with `"default"`, the chat-config fallback alone can't reach it — the first-registered-agent fallback is what closes it. The `useMemo` keeps it rules-of-hooks-clean, the deps (`configAgentId`, `copilotkit.agents`) cover the `getAgent` read, and the connecting-phase timing holds (empty agents → provisional, no throw; repopulates after sync).

Two non-blocking notes for the record:
- "first registered agent" (`Object.keys(...)[0]`) is a heuristic — fine for a best-effort PredictState subscription that must not crash, just a conscious semantic choice when multiple non-default agents are registered.
- The `use-capabilities` / `UseAgentProps.agentId` JSDoc updates are a welcome touch and keep the contract accurate.

FYI for tracking, not this PR: Vue ships its own `useAgent` (`packages/vue/src/v2/hooks/use-agent.ts`) with the same `agentId ?? DEFAULT_AGENT_ID` gap (never reads `chatConfig.agentId`), tracked separately in #5656. `react-native` re-exports the react-core hook, so it's covered transitively here.

NathanTarbert:
Thanks @contextablemark — all three doc items are addressed in `ca63ae24`:

- `use-agent.tsx` `UseAgentProps.agentId` — added the JSDoc with the resolution precedence (your suggestion).
- `use-capabilities.tsx:5` — updated to note the agent resolves from the surrounding chat configuration, falling back to the default.
- `use-capabilities.tsx:11` — updated the `@param agentId` description to match.

Separately, a later commit (`589fc51`) handles the actual crash the reporter hit (a provider-level `CopilotListeners` resolving to `default` outside the `<CopilotChat>` subtree), with an integration test mirroring their repro. When you have a moment, mind re-reviewing?

contextablemark (dismissed):
Sorry @NathanTarbert ... meant to approve earlier.  Some merge conflicts have crept in during the interim. Let me know if I need to reapprove

contextablemark (dismissed):
Re-approved

contextablemark (approved):
Looks good

PR #5594, opened 2026-06-19
docs: polish cookbook navigation and partner logos

tylerslaton:
Adding logos and reworking the cookbook page.

Screenshot 2026-06-19 at 4 42 13 PM


contextablemark (changes requested):
The cookbook nav + partner logos look good and render correctly locally. One blocker before merge.

## 🔴 Failing unit test: `globals-css.test.ts`

`showcase/shell-docs/src/app/__tests__/globals-css.test.ts:63` fails (`1 of 162` in the shell-docs vitest suite, as of `67509cc`).

The `style: auto-fix formatting` pass ran Prettier, which wrapped the long selector in `globals.css` across three lines:

```css
.shell-docs-sidebar-cookbook
  [data-radix-scroll-area-viewport]
  > div:first-child {
  padding-top: 0 !important;
  padding-bottom: 1.5rem !important;
}
```

…but the test still asserts the **single-line** selector via `toContain`:

```js
".shell-docs-sidebar-cookbook [data-radix-scroll-area-viewport] > div:first-child {\n  padding-top: 0 !important;\n  padding-bottom: 1.5rem !important;"
```

so the substring no longer matches.

**Possible fixes:**
- Quick: update the expected string to the wrapped form Prettier produces.
- Preferred: make the assertion whitespace-tolerant (normalize runs of whitespace before `toContain`) so the next `style: auto-fix` run can't re-break it. Re-flattening the CSS to one line would just get re-wrapped on the next Prettier pass.

## ⚠️ This suite isn't covered by CI — the failure is silent

Worth flagging because it explains why checks are green despite the above: the `showcase/shell-docs/**/__tests__` vitest suite is **not run by any workflow**.

- `test_unit.yml` has `paths-ignore: showcase/**` and only runs Nx `packages/**` projects — shell-docs is a standalone npm app outside the pnpm/Nx workspace.
- `test_integration-docs.yml` fires on `showcase/shell-docs/src/content/**` but runs the doc-model-name validator + `scripts/doc-tests`, not the vitest unit suite.
- `showcase_validate.yml`'s `vitest run` is scoped to `showcase/scripts`; the shell-docs Dockerfile only does `npm ci` + `npm run build`.

So the failing test (and the new `docs-render.test.ts` coverage added here) only run if someone does `npm test` in `showcase/shell-docs` locally. Fixing the assertion still matters so the suite stays trustworthy. Adding a CI job to run these on `showcase/shell-docs/**` PRs is a separate follow-up, not a blocker for this PR.

contextablemark (approved):
Re-reviewed at `790dfd0a7`. The blocker is resolved — approving.

- `globals-css.test.ts` now passes; full shell-docs vitest suite is green (**162/162**, 31 files).
- The fix uses the preferred approach: a `normalizeWhitespace()` helper that collapses whitespace before asserting, so a future `style: auto-fix` (Prettier re-wrap) can't re-break it. 👍

Cookbook nav + partner logos render correctly locally; typecheck clean, lint warnings-only (all pre-existing).

Non-blocking follow-up (unchanged from prior review): the `showcase/shell-docs/**/__tests__` vitest suite still isn't executed by any CI workflow, so it's only verified locally. Worth a separate ticket to wire a CI job on `showcase/shell-docs/**` PRs so this coverage actually guards the docs shell.

PR #5596, opened 2026-06-20
fix(web-inspector): forward core headers on owned thread store requests

### LINKED ISSUE #5581: 🐛 Bug: `enableInspector=true` causes `/threads` requests to ignore configured headers
### ♻️ Reproduction Steps

Following the fix in #5300, headers are now correctly included in `/threads` requests when using `useThreads()` directly.

However, when `enableInspector={true}` is enabled, the `/threads` requests triggered by the inspector **do not include the headers defined in `CopilotKit`**, leading to failures in environments that enforce CSRF or authorisation checks.

1. Configure `CopilotKit` with custom headers:

```jsx
<CopilotKit
  runtimeUrl={AGENT_URL}
  agent={AGENT_ID}
  useSingleEndpoint={false}
  enableInspector={true}
  headers={{ "X-CSRF": "1" }}
>
```

2. Open the application and trigger the inspector (or observe its initialisation).

3. Inspect the network requests for `/threads`.

### ✅ Expected Behavior

All `/threads` requests (including those initiated by the inspector) should include headers provided via `CopilotKit`, e.g.:

```
X-CSRF: 1
```

### ❌ Actual Behavior

* `/threads` requests triggered by the inspector:
  * ❌ Do **not include any custom headers**
* This leads to:
  * ✅ HTTP 200 in non-strict local environments (but still missing headers)
  * ❌ HTTP 403 in environments with enforced CSRF/Auth validation

## 🔍 Possible Root Cause
While debugging, I noticed the web-inspector code path appears to initialise the thread store with empty headers:

```ts
store.setContext({
  runtimeUrl: core.runtimeUrl,
  headers: {},
  agentId,
});
```

### Screenshots

Image

### 𝌚 CopilotKit Version

```shell
@copilotkit/react-core: 1.60.2
@copilotkit/react-ui: 1.60.2
@copilotkit/runtime: 1.60.2
```

### 📄 Logs (Optional)

```shell

```

serhiizghama:
## Problem

Fixes #5581.

When `enableInspector={true}` and `useThreads()` is **not** mounted, the inspector creates its own thread store per agent (`ensureOwnedThreadStore`). That store initialized its context with empty headers:

```ts
store.setContext({
  runtimeUrl: core.runtimeUrl,
  headers: {}, // ← ignores headers configured on <CopilotKit>
  agentId,
});
```

So the inspector's `/threads` requests omitted the headers configured on `<CopilotKit>` (e.g. `X-CSRF`, `Authorization`). In environments that enforce CSRF/auth checks this returns **HTTP 403**; in lax local envs it 200s but still sends no headers. This is the inspector-side counterpart to the `useThreads()` fix in #5300.

## Solution

1. Source the headers from `core.headers` when the owned store's context is created.
2. Add an `onHeadersChanged` subscriber that re-applies headers to all owned stores when the host updates them at runtime (e.g. a refreshed auth/CSRF token via `core.setHeaders`). This mirrors `useThreads()`, which re-dispatches the context whenever `core.headers` change, so the owned stores' requests stay authorized.

Headers are spread (`{ ...core.headers }`) to match the existing pattern in `use-threads.tsx` and produce a fresh mutable object. Stores registered by `useThreads()` are untouched — only inspector-owned stores are affected.

## Testing

Added two regression tests in `packages/web-inspector/src/__tests__/web-inspector.spec.ts` (stub `globalThis.fetch`, drive the owned store via the agents-changed path):

- the owned store's `/threads` request carries `core.headers`;
- an `onHeadersChanged` update re-applies the new headers on the next request.

`pnpm --filter @copilotkit/web-inspector test` → 34 passed. Verified the first test fails when the fix is reverted. Lint (oxlint) and formatting (oxfmt) clean on the changed files.

### DIFF
--- a/packages/web-inspector/src/__tests__/web-inspector.spec.ts
+++ b/packages/web-inspector/src/__tests__/web-inspector.spec.ts
@@ -575,3 +575,146 @@ describe("WebInspectorElement announcement preview dismissal", () => {
     expect(a.hasUnseenAnnouncement).toBe(true);
   });
 });
+
+// --- Owned thread store header forwarding (issue #5581) ---
+//
+// When useThreads() isn't mounted, the inspector creates its own thread store
+// per agent (ensureOwnedThreadStore). That store's /threads requests must carry
+// the headers configured on <CopilotKit> (e.g. X-CSRF / auth), otherwise the
+// requests 403 in environments that enforce CSRF/auth checks.
+
+type HeaderMockCore = {
+  agents: Record<string, AbstractAgent>;
+  context: Record<string, unknown>;
+  properties: Record<string, unknown>;
+  runtimeConnectionStatus: CopilotKitCoreRuntimeConnectionStatus;
+  runtimeUrl: string;
+  headers: Record<string, string>;
+  subscribe: (subscriber: CopilotKitCoreSubscriber) => {
+    unsubscribe: () => void;
+  };
+  getThreadStores: () => Record<string, never>;
+  getThreadStore: (agentId: string) => undefined;
+  registerThreadStore: (agentId: string, store: unknown) => void;
+  unregisterThreadStore: (agentId: string) => void;
+};
+
+function createHeaderMockCore(
+  agents: Record<string, AbstractAgent>,
+  headers: Record<string, string>,
+) {
+  const subscribers = new Set<CopilotKitCoreSubscriber>();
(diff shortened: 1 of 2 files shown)


onsclom (approved):

onsclom:
thanks, great fix!

PR #5597, opened 2026-06-20
docs(cookbook): add Angular + Google ADK production recipe

davidmckayv:
## What

A new Cookbook recipe — **Build an agentic app on Angular + Google ADK** — covering the non-obvious production gotchas when you wire an Angular frontend to a Google ADK agent over AG-UI, with optional CopilotKit Intelligence threads and memory. Written in the existing cookbook house style (symptom → cause → fix callouts), and it links out to the Angular and ADK quickstarts rather than re-teaching setup.

## Why

The Angular + ADK combination has a handful of correctness issues that only surface in a real, multi-user, governed app, and they aren't covered by the per-piece quickstarts:

- One agent, one store (a second composer must not create its own store)
- Scope the user via the **run body**, not an HTTP header (a header lags by one in-session switch)
- Never reconfigure the runtime mid-submit (it recreates the agent store and drops the message)
- Governance is server-side; the per-request allow-list rides the run body
- Choose a capable model, and degrade gracefully when the Intelligence platform is absent

The recipe is **model-flexible**: it frames ADK as running Gemini by default but supporting any model ADK supports, consistent with the ADK quickstart.

## Changes

- **New** `cookbook/angular-adk-agentic-app.mdx` — the recipe
- `cookbook/meta.json` — register in nav
- `cookbook/index.mdx` — add the index card (uses `/logos/google-adk.svg`)
- `src/lib/sidebar-icon.tsx` — add `custom/google-adk` sidebar icon entry
- `frontends/angular.mdx` — bidirectional cross-link into the recipe
- `src/lib/__tests__/docs-render.test.ts` — update the cookbook nav test for the new page

## Test plan

Run from `showcase/shell-docs`:

- `npm run typecheck` — passes
- `npm run lint` — passes (only pre-existing warnings; none in changed files)
- `npm run test` — the cookbook-nav test passes. Two `public-assets.test.ts` cases fail **only locally** because git-LFS PNGs are unmaterialized in a fresh worktree (the tests assert assets are not LFS pointer stubs); they are unrelated to this change and pass in CI.
- `npm run build` — passes; all 211 static pages generate, including `/cookbook` and `/cookbook/[...slug]`.

Verified in the browser at `localhost:3003`: recipe renders with correct callout styling, code highlighting, populated TOC, sidebar entry, and index card.

image


showcase/shell-docs/src/content/docs/cookbook/angular-adk-agentic-app.mdx
```diff
@@ -0,0 +1,200 @@
...
+  providers: [
+    provideCopilotKit({
+      runtimeUrl: "/api/copilotkit",
+      properties: { userId: currentUserId }, // arrives on the agent as forwardedProps
+    }),
+  ],
+};
+```
+
+On the agent side, the AG-UI adapter copies the run's `state` onto the ADK session, so a tool reads the user id straight off `tool_context.state`:
```
contextablemark: This `forwardedProps` → `tool_context.state` bridge doesn't hold on the ADK middleware, and it's load-bearing for the whole section (the "single most common correctness bug").

The two preceding lines say `properties: { userId }` "arrives on the agent as forwardedProps" — correct: CopilotKit `properties` serialize to `RunAgentInput.forwardedProps`. But `ADKAgent` only mirrors **`state`** (and `context`, under `_ag_ui_context`) into the ADK session; it never reads `forwarded_props`:

- `adk_agent.py` builds session state from `dict(input.state)` only — there is no `input.forwarded_props` read anywhere in `src/` or `tests/` (the lone occurrence is `forwarded_props=None` on the synthetic input in the agents-state endpoint).
- Contrast `LangGraphAgent`, which *does* surface specific keys (e.g. `injectA2UITool`) out of `forwardedProps` explicitly — ADK has no equivalent.

So as written, `properties.userId` → `forwardedProps` → dropped, and `tool_context.state.get("userId")` returns `None`. The recommended pattern silently fails to scope the user — the exact bug class this section warns about.

Two ways to make it correct on ADK:
1. **Carry `userId` in the run's `state`** (or as a `context` entry) instead of `properties` — that's what actually lands in `tool_context.state`. In CopilotKit that means agent/shared state, not the `properties` config object.
2. Keep `properties`/`forwardedProps` on the client but add a small server-side extractor that writes the value into session state before the run (mirroring how LangGraph handles its forwardedProps keys).

The header-lags-by-one-switch argument is still valid either way — it's only the specific transport field in the code sample that's off. Worth reconciling the "arrives as forwardedProps" line with the "copies the run's state" line so they don't contradict. Same fix applies to the coding-agent prompt near the bottom (`read it in ADK tools from tool_context.state["userId"]`).

contextablemark (dismissed):
See comment

showcase/shell-docs/src/content/docs/cookbook/angular-adk-agentic-app.mdx
davidmckayv: Confirmed against `ag_ui_adk/adk_agent.py`: session state is built from `dict(input.state)` (`adk_agent.py:2670`) plus `input.context` under `_ag_ui_context` (`:2691`). `forwarded_props` is only read for the `injectA2UITool` flag (`:2437`) and in `a2ui_tool.py`, never mirrored into session state. So `properties` → `forwardedProps` → `tool_context.state.get("userId")` returns `None`, as you said.

Fixed in 1676e2b72:
- Client now carries the user id via `connectAgentContext({ description: "userId", value: userId })` (a `Signal<Context>` so it re-registers on an in-session switch).
- Agent reads it from `tool_context.state["_ag_ui_context"]`, with shared agent state (`tool_context.state["userId"]`) noted as the direct-key alternative.
- Removed the contradictory "arrives as forwardedProps" / "copies the run's state" lines, and updated the coding-agent prompt to match.

The header-lags-by-one-switch point is unchanged; only the transport field was wrong.

contextablemark (commented):
Content looks good, so removed the stale review.  But I just noticed that all of the SVG logos in https://github.com/CopilotKit/CopilotKit/tree/main/showcase/shell-docs/public/logos are not actually logos, but rather SVGs containing the first two letters of each, so you will probably want to do something about that (e.g., choosing either an ADK logo or Angular logo and using the actual PNG as was done with Daytona and Arcade, for example), before seeking design approval.

davidmckayv:
On the logo: confirmed `public/logos/google-adk.svg` was a placeholder (grey tile + the text "Go"). Replaced it in fc5c5b244 with the real ADK glyph, reusing the vector paths from the existing `AdkIcon` component (`src/components/icons/framework-icons.tsx`) on a light tile so it renders on both light and dark surfaces. Only the new recipe references that file, so nothing else is affected.

It's monochrome for now rather than a guess at official ADK brand colors — figured that's the right thing to settle at design approval. Card and sidebar both verified rendering the real mark locally.

davidmckayv:
Correction on the logo: swapped to the **official** ADK mark — `agent-development-kit.png` from [google/adk-python](https://github.com/google/adk-python/blob/main/assets/agent-development-kit.png) — committed as an LFS PNG like Daytona/Arcade (d351b0dfa). The card and sidebar now point at `/logos/google-adk.png`, and the placeholder `google-adk.svg` is reverted to its original state. Verified rendering in both surfaces locally.

tylerslaton (approved):

PR #5599, opened 2026-06-21
Accept ag_ui (underscore) state key alongside ag-ui in LangGraph

### LINKED ISSUE #5463: 🐛 Bug:  AG-UI State Key `ag-ui` vs `ag_ui` Inconsistency Causes A2UI Rendering Failures
## Description

When using CopilotKit with LangGraph, the AG-UI state key is inconsistently accessed as `ag-ui` (hyphenated) in the CopilotKit SDK, but in some scenarios, the state is stored with the key `ag_ui` (underscore). This mismatch causes A2UI components to fail rendering because the `a2ui_schema` and other AG-UI properties cannot be properly read from the state.

**Root Cause**: The LangGraph JSON schema validation and state merging logic converts hyphenated keys to underscore-separated keys in certain scenarios, but the CopilotKit middleware code only checks for `state["ag-ui"]` and never falls back to `state["ag_ui"]`.

### Prerequisites
- Python 3.10+
- LangGraph >= 1.1.0
- `ag-ui-langgraph` >= 0.0.40
- `copilotkit` >= 0.1.94

### Step 1: Set up a basic LangGraph agent with A2UI

```python
# agent.py
from langgraph.graph import StateGraph, END
from langchain_openai import ChatOpenAI
from typing import TypedDict, Annotated, List
from langchain_core.messages import BaseMessage

class AgentState(TypedDict):
    messages: Annotated[List[BaseMessage], "Messages in the conversation"]
    # Note: ag-ui state is injected by middleware

async def chat_node(state: AgentState):
    model = ChatOpenAI(model="gpt-4o-mini")
    response = await model.ainvoke(state["messages"])
    return {"messages": [response]}

workflow = StateGraph(AgentState)
workflow.add_node("chat", chat_node)
workflow.add_edge("START", "chat")
workflow.add_edge("chat", END)

agent = workflow.compile(
    checkpointer=MemorySaver(),
    middleware=[CopilotKitMiddleware()]
)
```

### Step 2: Check the state during execution

Add debug logging to observe the actual state keys:

```python
async def chat_node(state: AgentState):
    # Debug: Print all state keys
    print("=== Current State Keys ===")
    print(f"All keys: {list(state.keys())}")
    print(f"Has 'ag-ui': {'ag-ui' in state}")
    print(f"Has 'ag_ui': {'ag_ui' in state}")
    print(f"ag-ui value: {state.get('ag-ui', 'NOT FOUND')}")
    print(f"ag_ui value: {state.get('ag_ui', 'NOT FOUND')}")
    
    model = ChatOpenAI(model="gpt-4o-mini")
    response = await model.ainvoke(state["messages"])
    return {"messages": [response]}
```

### Step 3: Observe the issue

When running the agent through a CopilotKit endpoint, you may see output like:

```
=== Current State Keys ===
All keys: ['messages', 'ag_ui', 'copilotkit']
Has 'ag-ui': False
Has 'ag_ui': True
ag-ui value: NOT FOUND
ag_ui value: {'a2ui_schema': '{"catalogId": "...", "components": [...]}', 'inject_a2ui_tool': True, ...}
```

Notice that the state contains `ag_ui` (underscore) but CopilotKit middleware code looks for `ag-ui` (hyphen).

### Step 4: Verify A2UI rendering failure

The `generate_a2ui` tool will fail to find the catalog because:
- `_resolve_a2ui_catalog()` looks for `state["ag-ui"]["a2ui_schema"]` → returns `None`
- `_a2ui_inject_decision()` looks for `state["ag-ui"]["inject_a2ui_tool"]` → returns `None`

This results in:
1. A2UI tool not being injected
2. "Unknown component" errors when trying to render A2UI surfaces
3. `a2ui_schema` not passed to sub-agents

## Workaround

In your agent code, manually copy the state from `ag_ui` to `ag-ui`:

```python
async def chat_node(state: AgentState):
    # Workaround: Ensure both key formats are available
    if "ag_ui" in state and "ag-ui" not in state:
        state["ag-ui"] = state["ag_ui"]
    
    model = ChatOpenAI(model="gpt-4o-mini")
    response = await model.ainvoke(state["messages"])
    return {"messages": [response]}
```

## Environment

- **CopilotKit Version**: 0.1.94+
- **AG-UI LangGraph Version**: 0.0.40+
- **LangGraph Version**: 1.1.0+
- **Python Version**: 3.10+
- **OS**: [Your OS]

## Additional Context

This issue appears to be related to how LangGraph handles JSON schema property names during state merging. Hyphenated property names in JSON schemas may be converted to underscores internally by LangGraph, but the CopilotKit middleware expects the original hyphenated format.

The issue affects:
1. A2UI tool injection (`inject_a2ui_tool` flag)
2. A2UI catalog resolution (`a2ui_schema`)
3. Frontend tool forwarding (`tools`)
4. Context propagation (`context`)

All of these properties live under the `ag-ui`/`ag_ui` state key and are affected by this inconsistency.

## Expected Behavior

The CopilotKit middleware should handle both `ag-ui` (hyphen) and `ag_ui` (underscore) state keys interchangeably, since LangGraph may use either format depending on the context.

## Actual Behavior

The middleware only reads `state["ag-ui"]`, causing failures when the state is stored with `ag_ui` key.

### 1. `sdk-python/copilotkit/langgraph_agui_agent.py`

**Line ~280**:
```python
# Current code:
agui_properties = merged_state.get("ag-ui", {}) or merged_state

# Should be:
agui_properties = merged_state.get("ag-ui", {}) or merged_state.get("ag_ui", {}) or merged_state
```

### 2. `sdk-python/copilotkit/copilotkit_lg_middleware.py`

**`_resolve_a2ui_catalog()` method**:
```python
# Current code:
ag_ui = state.get("ag-ui") or {}

# Should be:
ag_ui = state.get("ag-ui") or state.get("ag_ui") or {}
```

**`_a2ui_inject_decision()` method**:
```python
# Current code:
return (state.get("ag-ui") or {}).get("inject_a2ui_tool")

# Should be:
ag_ui = state.get("ag-ui") or state.get("ag-ui") or {}
return ag_ui.get("inject_a2ui_tool")
```

### Screenshots

_No response_

### 𝌚 CopilotKit Version

```shell
- **CopilotKit Version**: 0.1.94+
- **AG-UI LangGraph Version**: 0.0.40+
- **LangGraph Version**: 1.1.0+
- **Python Version**: 3.10+
```

### 📄 Logs (Optional)

```shell

```

itxaiohanglover:
## What does this PR do?

LangGraph's JSON-schema validation can rewrite the hyphenated `ag-ui` state key to `ag_ui` (underscore). The CopilotKit LangGraph middleware only read `state["ag-ui"]`, so when the namespace arrived under `ag_ui` it was missed and A2UI components failed to render (no `generate_a2ui` tool injection, no catalog resolution, no `inject_a2ui_tool` flag read).

This adds an `ag_ui` fallback next to every `ag-ui` state read, in the three places identified in #5463:

- `sdk-python/copilotkit/langgraph_agui_agent.py` — `langgraph_default_merge_state` reads agui `tools`/`context`.
- `sdk-python/copilotkit/copilotkit_lg_middleware.py` — `_resolve_a2ui_catalog` reads `a2ui_schema`.
- `sdk-python/copilotkit/copilotkit_lg_middleware.py` — `_a2ui_inject_decision` reads `inject_a2ui_tool`.

The change is backward compatible: when `ag-ui` is present and truthy, behavior is unchanged; the underscore key is only consulted as a fallback.

## Related PRs and Issues

- Fixes #5463

## Checklist

### DIFF
--- a/sdk-python/copilotkit/copilotkit_lg_middleware.py
+++ b/sdk-python/copilotkit/copilotkit_lg_middleware.py
@@ -339,7 +339,7 @@ class CopilotKitMiddleware(AgentMiddleware[StateSchema, Any]):
         BYOC custom catalogs render their own components, not the basic one).
         """
         # AG-UI native path.
-        ag_ui = state.get("ag-ui") or {}
+        ag_ui = state.get("ag-ui") or state.get("ag_ui") or {}
         a2ui_schema = ag_ui.get("a2ui_schema")
         if a2ui_schema:
             catalog_id = None
@@ -385,7 +385,7 @@ class CopilotKitMiddleware(AgentMiddleware[StateSchema, Any]):
         means no signal at all (off, or no A2UI middleware in the pipeline), in
         which case we do not auto-inject.
         """
-        return (state.get("ag-ui") or {}).get("inject_a2ui_tool")
+        return (state.get("ag-ui") or state.get("ag_ui") or {}).get("inject_a2ui_tool")
 
     def _maybe_build_a2ui_tool(self, request: ModelRequest) -> Any | None:
         """Build a ``generate_a2ui`` tool bound to the agent's own model when
--- a/sdk-python/copilotkit/langgraph_agui_agent.py
+++ b/sdk-python/copilotkit/langgraph_agui_agent.py
@@ -273,7 +273,7 @@ class LangGraphAGUIAgent(LangGraphAgent):
         """Override to add CopilotKit actions to the state"""
         merged_state = super().langgraph_default_merge_state(state, messages, input)
         # Extract tools from the merged state and add them as CopilotKit actions
-        agui_properties = merged_state.get("ag-ui", {}) or merged_state
+        agui_properties = merged_state.get("ag-ui", {}) or merged_state.get("ag_ui", {}) or merged_state
 
         return {
             **merged_state,


BenTaylorDev (changes requested):
Thanks — the read-side approach is right: defending against both spellings at the read sites (rather than chasing the writer, which lives in the external `ag_ui_langgraph` adapter) is the correct architecture, and the `state.get("ag-ui") or state.get("ag_ui")` precedence is backward-compatible. Two things before it can land:

1. **Missing a fourth site the issue called out.** `_RESERVED_STATE_KEYS` (`copilotkit_lg_middleware.py:198-207`) still lists only `"ag-ui"`, not `"ag_ui"`. That frozenset is the `expose_state` filter — so in the exact scenario this PR targets (LangGraph having rewritten the key to `ag_ui`), with `expose_state` enabled the whole AG-UI plumbing dict (`tools`, `context`, `a2ui_schema`, `inject_a2ui_tool`) is no longer reserved and leaks into the LLM system prompt. The fix makes the code *consume* `ag_ui` while leaving it *un-reserved*, which half-handles the underscore case. Please add `"ag_ui"` to `_RESERVED_STATE_KEYS`.
2. **No test for the underscore path.** Every existing test uses the `ag-ui` hyphen key, so the new behavior has zero coverage and the load-bearing premise (LangGraph actually delivering state under `ag_ui`) is asserted but never exercised. Please add a test that drives the underscore key through `_resolve_a2ui_catalog`, `_a2ui_inject_decision`, and the state-note/reserved-keys path.

Also worth confirming for the record: can you point to where the underscore key actually shows up at runtime (a repro or a captured state)? That's the one thing green CI can't currently prove.

PR #5603, opened 2026-06-21
fix(web-inspector): handle empty tool result content

### LINKED ISSUE #5598: 🐛 Bug:
### ♻️ Reproduction Steps

Problem Symptom
When opening historical sessions in the CopilotKit Inspector debugging tool, an error is printed in the console:
```[intercept-console-error.ts:42] [CopilotKit Inspector] Failed to parse tool-call result content
{
  toolCallId: 'call_7b969ac708aa4e0c8d643574',
  raw: '',
  error: SyntaxError: Unexpected end of JSON input
    at JSON.parse (<anonymous>)
    at ɵCpkThreadDetails.…
}```


### ✅ Expected Behavior

Root Cause Analysis
```
// Buggy code
tc.result = JSON.parse(msg.content ?? "{}") as Record<string, unknown>;
```
The nullish coalescing operator (??) only handles null and undefined values.
The content field of certain tool result messages in historical sessions is an empty string "".
An empty string is not a nullish value, so the fallback value "{}" fails to take effect.
Executing JSON.parse("") throws the error: SyntaxError: Unexpected end of JSON input

### ❌ Actual Behavior

 Root Cause Analysis
```
// Buggy code
tc.result = JSON.parse(msg.content ?? "{}") as Record<string, unknown>;
```
The nullish coalescing operator (??) only handles null and undefined values.
The content field of certain tool result messages in historical sessions is an empty string "".
An empty string is not a nullish value, so the fallback value "{}" fails to take effect.
Executing JSON.parse("") throws the error: SyntaxError: Unexpected end of JSON input

### Screenshots

_No response_

### 𝌚 CopilotKit Version

```shell
web-inspector  1.60.0
```

### 📄 Logs (Optional)

```shell

```

Tushar-Khandelwal-2004:
## Summary

Fixes the web inspector crash when historical tool result messages have empty `content`.

Some stored thread tool-result messages can have `content: ""`. The inspector previously parsed tool result content with `JSON.parse(msg.content ?? "{}")`, which still passes the empty string into `JSON.parse` and throws `Unexpected end of JSON input`.

## Changes

-> Treat empty and whitespace-only tool result content as an empty result object
-> Keep malformed non-empty result content flowing through the existing parse-error fallback
-> Add regression coverage for empty and whitespace-only tool result messages in the thread conversation mapper

## Testing

-> `pnpm exec oxlint packages/web-inspector/src/index.ts packages/web-inspector/src/__tests__/web-inspector.spec.ts`
-> `pnpm -C packages/web-inspector exec vitest run src/__tests__/web-inspector.spec.ts --pool=threads --testNamePattern "maps empty tool result content" --reporter=verbose`

## Notes

-> `pnpm nx show project @copilotkit/web-inspector --json` timed out locally
-> `pnpm -C packages/web-inspector run check-types` is currently blocked by existing package-level TypeScript errors around explicit `.js` import extensions and one unrelated existing error

## Related Issue

Closes #5598

### DIFF
--- a/packages/web-inspector/src/__tests__/web-inspector.spec.ts
+++ b/packages/web-inspector/src/__tests__/web-inspector.spec.ts
@@ -411,6 +411,19 @@ type ThreadDetailsInternals = {
   timelineItemsFromEvents: (
     events: Array<Record<string, unknown>>,
   ) => Array<Record<string, unknown>>;
+  mapMessages: (
+    messages: Array<{
+      id: string;
+      role: string;
+      content?: string | null;
+      toolCallId?: string | null;
+      toolCalls?: Array<{
+        id: string;
+        name: string;
+        args: string;
+      }>;
+    }>,
+  ) => Array<Record<string, unknown>>;
 };
 
 function createThreadDetails(): {
@@ -453,7 +466,6 @@ function createDeferred<T>(): {
   });
   return { promise, resolve, reject };
 }
-
 describe("ɵCpkThreadDetails caching", () => {
   beforeEach(() => {
     document.body.innerHTML = "";
@@ -748,6 +760,57 @@ describe("ɵCpkThreadDetails caching", () => {
 
     expect(internals.renderState()).not.toBe(stateA);
     expect(internals.renderEvents()).not.toBe(eventsA);
+  });
+
+  it("maps empty tool arguments and results as empty objects", () => {
+    const { internals } = createThreadDetails();
+
+    const items = internals.mapMessages([
+      {
+        id: "assistant-1",
+        role: "assistant",
+        content: "",
+        toolCalls: [
+          {
+            id: "call-empty",
+            name: "lookupUser",
+            args: "",
+          },
+          {
+            id: "call-whitespace",
+            name: "lookupOrder",
+            args: "   ",
+          },
+        ],
+      },
+      {
+        id: "tool-empty",
+        role: "tool",
+        toolCallId: "call-empty",
+        content: "",
+      },
+      {
+        id: "tool-whitespace",
+        role: "tool",
+        toolCallId: "call-whitespace",
+        content: "   ",
+      },
+    ]);
+
+    expect(items).toMatchObject([
+      {
+        type: "tool_call",
+        toolCallId: "call-empty",
+        arguments: {},
+        result: {},
+      },
+      {
+        type: "tool_call",
+        toolCallId: "call-whitespace",
+        arguments: {},
+        result: {},
+      },
+    ]);
   });
 });
 
--- a/packages/web-inspector/src/index.ts
+++ b/packages/web-inspector/src/index.ts
@@ -2327,10 +2327,10 @@ export class CpkThreadInspector extends LitElement {
             let args: Record<string, unknown> = {};
             if (typeof tc.args === "string") {
               try {
-                args = JSON.parse(tc.args) as Record<string, unknown>;
+                args = this.parseToolCallContent(tc.args);
               } catch (err) {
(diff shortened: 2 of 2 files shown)


BenTaylorDev:
Thanks @Tushar-Khandelwal-2004 — the fix itself is correct and complete: `JSON.parse(msg.content ?? "{}")` doesn't catch empty-string content, and `parseToolCallResultContent` handling empty/whitespace as `{}` while letting malformed non-empty content flow to the existing catch is right. It's also the only affected site (the SSE `TOOL_CALL_RESULT` path is already guarded).

The blocker is just staleness — the branch is well behind main and currently conflicts in **both** files (`index.ts` and the spec, which has since been refactored). Could you rebase? After rebasing, please double-check the production edit still lands on the live `JSON.parse(msg.content ?? "{}")` site (the test-file conflict is the larger one since `settleThread` moved).

BenTaylorDev (changes requested):
Thanks for this. The fix itself is correct and — nicely — the new test drives the real `mapMessages` path with `""` and `"   "` rather than a mock, so the mechanism is genuinely exercised. Two things to sort out before merge:

1. **A design decision, not just code.** The immediately-adjacent tool-**args** parser (`index.ts` ~2330-2340) deliberately does the *opposite* of this, with an explicit comment: *"Inspector is a debugging surface — surface malformed payloads instead of silently substituting `{}`."* This PR makes empty tool-**result** content silently become `{}`. Whether an empty result is "a tool that returned nothing" (→ `{}` is right) or "a debugging surface should show it as empty/anomalous" is a real UX call that conflicts with the sibling parser's stated philosophy. Could you align the two — either justify why results should differ from args, or surface empties consistently?
2. **Symptom framing.** The current `JSON.parse(msg.content ?? "{}")` is already inside a try/catch, so an empty result today renders as a `__parseError` sentinel (a *misrender*), not the "crash / Unexpected end of JSON input" the description states. Worth correcting the framing so the fix is evaluated for what it actually does.

One logistical note: the branch is well behind `main` (the edited region is unchanged, so it rebases cleanly at the code level, but it does need a rebase before merge).

Tushar-Khandelwal-2004:
Thanks for the thoughtful review @BenTaylorDev I’ve rebased the branch onto current `main` and resolved both conflicts.

I also aligned the behavior for tool arguments and tool results: empty or whitespace-only content is now treated as an empty object in both cases, while malformed non-empty JSON still surfaces the existing parse-error sentinel in the inspector.

The regression test now exercises the real `mapMessages` path for empty and whitespace-only arguments and results. I’ve also updated the implementation framing accordingly: this fixes a misrender to the parse-error state, rather than a crash.

BenTaylorDev (approved):
**Re-approving.** The args-vs-results conflict is resolved by unifying both parsers on a shared `parseToolCallContent` (`index.ts:2392`): trim → `{}` if empty/whitespace, else `JSON.parse` (which still throws on genuinely malformed input → surfaced as the `__parseError` sentinel). It's applied at both the args (`:2330`) and result (`:2375`) sites, each still inside its `try/catch`, and the old contradictory comment is rewritten to the single rule. This also fixes the original mechanism — on `main` the result site's `?? "{}"` only caught null/undefined, so `""` fell into `JSON.parse("")` and misrendered; now `""` → `{}`. Verified locally on the updated branch against current `main` (web-inspector 97/97, lint 0, `tsc` clean) and CI is green.

Tushar-Khandelwal-2004:
Thanks for the review and merge @BenTaylorDev, appreciated you pushing on the args vs results consistency issue, made for a much cleaner fix.

PR #5609, opened 2026-06-22
Guard canvas plan auto-continue loops

### LINKED ISSUE #5606: 🐛 Bug: canvas/langgraph-python example self-loops chat_node on LLM-controlled `planStatus` with no iteration guard — graph crashes on degenerate LLM output
### ♻️ Reproduction Steps

The buggy logic is in [`examples/canvas/langgraph-python/agent/agent.py`](https://github.com/CopilotKit/CopilotKit/blob/ca12d09c36fe303cec9040471bbf4704ff7571f1/examples/canvas/langgraph-python/agent/agent.py), specifically the two `Command(goto="chat_node")` returns at lines [672](https://github.com/CopilotKit/CopilotKit/blob/ca12d09c36fe303cec9040471bbf4704ff7571f1/examples/canvas/langgraph-python/agent/agent.py#L672) and [705](https://github.com/CopilotKit/CopilotKit/blob/ca12d09c36fe303cec9040471bbf4704ff7571f1/examples/canvas/langgraph-python/agent/agent.py#L705). Both branches re-enter `chat_node` whenever `planStatus` is anything other than `"completed"`. There is no iteration counter or fallback exit — termination is entirely conditioned on the LLM eventually writing `planStatus="completed"` via a tool call.

Reproducing the failure mode doesn't require running the full example (it pulls `copilotkit`, openai client, etc.). The bug is in the control-flow shape, not in the node body — a faithful re-implementation with the same self-loop guard surfaces the failure:

```python
import asyncio
from typing import TypedDict
from langgraph.graph import StateGraph, START
from langgraph.types import Command

class S(TypedDict, total=False):
    planStatus: str
    loops: int

async def chat_node(state: S):
    n = state.get("loops", 0) + 1
    # Same self-loop guard as agent.py:670-693 — the only path out is
    # planStatus == "completed", which only the LLM can write
    if state.get("planStatus", "") != "completed":
        return Command(goto="chat_node", update={
            "planStatus": state.get("planStatus", ""),
            "loops": n,
        })
    return Command(goto="__end__", update={"loops": n})

g = StateGraph(S)
g.add_node("chat_node", chat_node)
g.add_edge(START, "chat_node")
app = g.compile()

# Simulate a turn where the LLM never writes planStatus="completed"
try:
    asyncio.run(app.ainvoke({"planStatus": "", "loops": 0}, {"recursion_limit": 25}))
except Exception as e:
    print(f"{type(e).__name__}: {e}")
```

Output:
```
GraphRecursionError: Recursion limit of 25 reached without hitting a stop condition.
```

A static check on the real source at HEAD `ca12d09c`:
- 2 `Command(goto="chat_node")` returns (lines 672, 705)
- both guarded only on `planStatus != "completed"`
- no `recursion_limit`, `max_iter`, or any iteration counter anywhere in `agent.py`

### ✅ Expected Behavior

Either:

- The self-loop has a maximum iteration counter that forces exit (and surfaces a user-friendly error / tool message) before hitting LangGraph's default `recursion_limit`, or
- The retry loop is structured around a deterministic state transition that the framework can always advance (not "wait for the LLM to call `complete_plan`"), or
- `chat_node` returns `Command(goto=END)` after N retries with a clear error explaining "the plan never reached completed status; please retry with clearer instructions."

In any version, a flaky / lazy / parser-failing LLM response should not crash the graph.

### ❌ Actual Behavior

When the LLM never writes `planStatus="completed"` (which can happen on:

- a degenerate response that omits the structured tool call,
- a parser failure on a malformed function-call argument,
- a model that decides "the plan is good as-is" without explicitly emitting `complete_plan`,
- or just plain model laziness),

the graph re-enters `chat_node` indefinitely until LangGraph raises `GraphRecursionError`. The user sees an opaque framework error instead of a useful "I couldn't complete the plan" agent response.

### Screenshots

_No response_

### 𝌚 CopilotKit Version

```shell
Repo HEAD: `ca12d09c36fe303cec9040471bbf4704ff7571f1` (current `main`, 2026-06-19)
Affected file: `examples/canvas/langgraph-python/agent/agent.py`
Verified on `langgraph==1.2.6`.
```

### 📄 Logs (Optional)

```shell

```

Tushar-Khandelwal-2004:
## Summary
Fixes #5606.

This adds a bounded guard around the canvas LangGraph Python example's automatic plan self-loops. If the model keeps re-entering `chat_node` without making tool progress or marking the plan complete, the graph now exits cleanly with a useful assistant message instead of eventually surfacing a LangGraph recursion error.

## Changes
- Track consecutive automatic plan continuations in agent state.
- Reset the counter when backend tool calls or frontend tool calls make progress.
- Mark unfinished plan steps failed after repeated retries so the UI gets a clear terminal state.
- Deterministically mark the plan completed when all steps are already complete but the final completion signal was not emitted.

## Testing
- `python -m py_compile examples/canvas/langgraph-python/agent/agent.py`
- `git diff --check`

### DIFF
--- a/examples/canvas/langgraph-python/agent/agent.py
+++ b/examples/canvas/langgraph-python/agent/agent.py
@@ -68,6 +68,7 @@ class AgentState(CopilotKitState):
     planSteps: List[Dict[str, Any]] = []
     currentStepIndex: int = -1
     planStatus: str = ""
+    planContinuationCount: int = 0
 
 
 def summarize_items_for_prompt(state: AgentState) -> str:
@@ -155,6 +156,8 @@ backend_tools = [
 # Extract tool names from backend_tools for comparison
 backend_tool_names = [tool.name for tool in backend_tools]
 
+MAX_AUTOMATIC_PLAN_CONTINUES = 8
+
 # Frontend tool allowlist to keep tool count under API limits and avoid noise
 FRONTEND_TOOL_ALLOWLIST = set(
     [
@@ -195,7 +198,7 @@ FRONTEND_TOOL_ALLOWLIST = set(
 
 async def chat_node(
     state: AgentState, config: RunnableConfig
-) -> Command[Literal["tool_node", "__end__"]]:
+) -> Command[Literal["chat_node", "tool_node", "__end__"]]:
     """
     Standard chat node based on the ReAct design pattern. It handles:
     - The model to use (and binds in CopilotKit actions and the tools defined above)
@@ -278,6 +281,10 @@ async def chat_node(
     plan_steps = state.get("planSteps", []) or []
     current_step_index = state.get("currentStepIndex", -1)
     plan_status = state.get("planStatus", "")
+    try:
+        plan_continue_count = int(state.get("planContinuationCount", 0) or 0)
+    except Exception:
+        plan_continue_count = 0
     field_schema = (
         "FIELD SCHEMA (authoritative):\n"
         "- project.data:\n"
@@ -441,6 +448,9 @@ async def chat_node(
                             "planSteps": state.get("planSteps", []),
                             "currentStepIndex": state.get("currentStepIndex", -1),
                             "planStatus": state.get("planStatus", ""),
+                            "planContinuationCount": state.get(
+                                "planContinuationCount", 0
+                            ),
                         },
                     )
     except Exception:
@@ -615,6 +625,7 @@ async def chat_node(
                 "currentStepIndex": state.get("currentStepIndex", -1),
                 "planStatus": state.get("planStatus", ""),
                 **plan_updates,
+                "planContinuationCount": 0,
                 # guidance for follow-up after tool execution
                 "__last_tool_guidance": "If a deletion tool reports success (deleted:ID), acknowledge deletion even if the item no longer exists afterwards.",
             },
@@ -660,6 +671,7 @@ async def chat_node(
                 "currentStepIndex": state.get("currentStepIndex", -1),
                 "planStatus": state.get("planStatus", ""),
                 **plan_updates,
+                "planContinuationCount": 0,
                 "__last_tool_guidance": (
                     "Frontend tool calls issued. Waiting for client tool results before continuing."
                 ),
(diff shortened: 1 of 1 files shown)


BenTaylorDev (changes requested):
Thanks for this, @Tushar-Khandelwal-2004 — the bug is real (I reproduced the unbounded `chat_node` self-loop on current `main` at lines 672/705), and the general shape of the fix is sound: the state field, the return-type widening, and the two terminal-`END` blocks all check out, and I confirmed there's no data loss (`plan_updates` only carries `planSteps`/`currentStepIndex`/`planStatus`, all propagated) and no scope/import issues.

One correctness gap is holding it back, though: **the guard only bounds *consecutive no-tool-call* continuations, so a tool-spinning model still hits `GraphRecursionError` — the exact crash this PR is meant to prevent.**

Tracing it:
1. `workflow.add_edge("tool_node", "chat_node")` — every tool execution returns to `chat_node`.
2. Both the backend-tool path and the frontend-tool path reset `planContinuationCount` to `0`.
3. So a lazy/degenerate model that emits *any* tool call each turn (e.g. repeatedly listing items) but never writes `planStatus="completed"` produces a `chat_node → tool_node → chat_node` cycle where the counter resets every round and never reaches the threshold.
4. Each round burns ~2 supersteps → after ~12 rounds LangGraph raises `GraphRecursionError` regardless.

This is the more realistic runaway, and it's what the issue's *expected behavior* calls out — option 1 asks specifically for *"a maximum iteration counter that forces exit before hitting LangGraph's default `recursion_limit`."* The current implementation is a *consecutive-no-progress* counter rather than a *total-iteration* one, so it covers the narrowest slice (zero tool calls) and leaves the tool-spinning case crashing.

Two ways to close it, either works:
- Add a **total** per-invocation entry cap on `chat_node` that isn't reset by tool calls, or
- Reset the counter only on *genuine plan advancement* (`currentStepIndex` increases, or a step transitions to `completed`) rather than on any tool call — that bounds both loop shapes.

Minor, non-blocking: the deterministic completion/failure paths set `planStatus` directly instead of going through the `complete_plan` tool the UI normally observes — worth a quick check that the canvas reflects the terminal state correctly.

Everything else looks good — just need the guard to actually bound the tool-calling loop and this is mergeable.

Tushar-Khandelwal-2004:
Thanks for the detailed review - I've pushed a follow-up in 18d0c82236.

`planContinuationCount` is now a total per-invocation cap: backend tool routes increment it instead of resetting it, and the guard runs before another `tool_node` route is returned. That bounds the `chat_node ? tool_node ? chat_node` loop and exits cleanly before LangGraph's recursion limit. The terminal path marks unfinished steps as failed; if every step is already complete but `complete_plan` was missed, it marks the plan completed instead.

I also checked the canvas: it renders terminal state from `planStatus`, so the deterministic `completed` / `failed` updates are reflected in the UI.

Validated with `python -m py_compile` and `git diff --check`.

BenTaylorDev (changes requested):
Thanks for the follow-up, @Tushar-Khandelwal-2004 — **the change request is resolved.** I ran the real `agent.py` (this branch vs `main`) under LangGraph with a scripted tool-spinning model, and the tool-cycle case that used to crash now exits cleanly: `main` → `GraphRecursionError` at 13 LLM calls, this branch → clean END at 8 with `planStatus='failed'`. The no-tool-call case from #5606 is fixed too, legitimate multi-step plans are unaffected (real work goes through frontend tools, which end the turn and reset the counter), and the 8-entry cap lands at ~15 supersteps, comfortably under the default limit of 25. `ruff format --check` is clean and I confirmed the canvas renders the terminal `failed`/`completed` state (`page.tsx:1506`, `page.tsx:1548`).

One thing to resolve before this can go in — placement, not design:

The guard sits *before* the `route_to_tool_node(response)` check, so when it fires it discards `response`. In the tool-spinning case that response carries a backend tool call which has **already been streamed to the client** — the CopilotKit LangGraph agent dispatches `TOOL_CALL_START/ARGS/END` off `on_chat_model_stream` unless `copilotkit:emit-tool-calls` is disabled, and this example never disables it. The result is an emitted tool call that never receives a `ToolMessage`. My run confirms it: `DANGLING backend tool_call streamed but never resolved: ['update_plan_progress']`.

That matters because the `4.1` pre-flight at `agent.py:437` only catches dangling *frontend* calls (`name not in backend_tool_names`), so on the user's next message that assistant message goes straight into `model_with_tools.ainvoke(...)` and OpenAI rejects it with *"an assistant message with 'tool_calls' must be followed by tool messages responding to each tool_call_id"* — poisoning the thread for every subsequent turn, which is worse than the one-off crash we're replacing.

The guard needs to reach its terminal `END` without leaving an emitted backend tool call unresolved. Whatever shape you pick, please confirm the tool-spinning case still terminates cleanly and that legitimate multi-step plans are unchanged.

Two optional notes, neither blocking:

- The guard requires `effective_steps` to be non-empty and `planStatus != "completed"`, so a model spinning backend tools with **no plan set** (or after the plan is marked complete) still hits `GraphRecursionError`. That's a plain ReAct loop rather than the plan self-loop #5606 describes, and it predates this PR — fine to leave, worth a comment if you'd rather note it.
- `planContinuationCount` isn't mirrored in the frontend `AgentState` (`src/lib/canvas/types.ts:69`, `src/lib/canvas/state.ts:11`). Harmless at runtime, just tidier for the shared-state shape.

PR #5627, opened 2026-06-23
Route research canvas frontend tool calls

### LINKED ISSUE #5607: 🐛 Bug: research-canvas TypeScript agent route() omits WriteReport/WriteResearchQuestion — graph silently terminates with the tool call unhandled
## ♻️ Reproduction Steps

The bug is in [`examples/v1/research-canvas/agents/typescript/src/agent.ts`](https://github.com/CopilotKit/CopilotKit/blob/ca12d09c36fe303cec9040471bbf4704ff7571f1/examples/v1/research-canvas/agents/typescript/src/agent.ts) lines 37-67. The `route()` function only recognizes `"Search"` and `"DeleteResources"` as tool names:

```ts
function route(state: AgentState) {
  const messages = state.messages || [];
  if (... && messages[len-1].constructor.name === "AIMessageChunk") {
    const aiMessage = messages[len-1] as AIMessage;
    if (aiMessage.tool_calls && aiMessage.tool_calls.length > 0 &&
        aiMessage.tool_calls[0].name === "Search") {
      return "search_node";
    } else if (... && aiMessage.tool_calls[0].name === "DeleteResources") {
      return "delete_node";
    }
  }
  if (... && messages[len-1].constructor.name === "ToolMessage") {
    return "chat_node";
  }
  return END;   // ← any unrecognized tool call falls through
}
```

The agent's prompt and the sibling **Python** implementation (in [`examples/v1/research-canvas/agents/python/src/lib/chat.py`](https://github.com/CopilotKit/CopilotKit/blob/ca12d09c36fe303cec9040471bbf4704ff7571f1/examples/v1/research-canvas/agents/python/src/lib/chat.py)) tell the LLM to use **`WriteReport`** and **`WriteResearchQuestion`** as well. The TS `route()` does not handle either — they fall through to `return END`.

A small AST diff confirms the divergence:

```python
import re, pathlib
REPO = pathlib.Path("CopilotKit_CopilotKit")
ts  = (REPO/"examples/v1/research-canvas/agents/typescript/src/agent.ts").read_text()
py  = (REPO/"examples/v1/research-canvas/agents/python/src/lib/chat.py").read_text()
ts_handled = set(re.findall(r'tool_calls\[0\]\.name === "([^"]+)"', ts))
py_known   = set(re.findall(r'^def (\w+)\(', py, re.M))
print("TS handles:", ts_handled)
print("Python expects:", py_known)
print("missing from TS:", py_known - ts_handled - {'chat_node', 'state'})
```

Output:
```
TS handles:           {'Search', 'DeleteResources'}
Python expects:       {'Search', 'DeleteResources', 'WriteReport', 'WriteResearchQuestion', 'search'}
missing from TS:      {'WriteReport', 'WriteResearchQuestion'}
```

## ✅ Expected Behavior

`route()` should either:

- Explicitly handle every tool name the LLM is prompted to use (and any frontend `useCopilotAction` registered through CopilotKit), or
- Recognize that an unknown tool call is still a tool call and route somewhere reasonable (e.g. back to `chat_node` to let CopilotKit's frontend handler consume it via the AG-UI channel) instead of returning `END`.

Either way, an LLM tool call should not be silently dropped on the floor.

## ❌ Actual Behavior

When the LLM emits `WriteReport(...)` or `WriteResearchQuestion(...)` (the two most common tool calls in this agent, given the prompt), `route()` falls through every branch and returns `END`. The compiled graph terminates cleanly, but:

- The `AIMessage` containing the tool_call is the last message in the channel.
- No `ToolMessage` is ever produced for that `tool_call_id`.
- The frontend's `useCopilotAction` handler (which would have run the `WriteReport` action) never fires because the run terminated server-side before reaching the AG-UI handoff.
- The user observes the agent "saying" it wrote a report, but no report appears in the canvas.

## Screenshots

n/a (control-flow bug).

## 𝌚 CopilotKit Version

Repo HEAD: `ca12d09c36fe303cec9040471bbf4704ff7571f1` (current `main`, 2026-06-19)
Affected file: `examples/v1/research-canvas/agents/typescript/src/agent.ts`

## 📄 Logs (Optional)

No errors — the graph terminates with status `ok`. The misbehavior is visible only as "report never showed up in the UI."

## Suggested fix

Add explicit handling for the frontend actions (and any future ones) or fall back to letting the unhandled tool call propagate:

```diff
 function route(state: AgentState) {
   const messages = state.messages || [];
   if (messages.length > 0 && messages[messages.length - 1].constructor.name === "AIMessageChunk") {
     const aiMessage = messages[messages.length - 1] as AIMessage;
-    if (aiMessage.tool_calls && aiMessage.tool_calls.length > 0 && aiMessage.tool_calls[0].name === "Search") {
+    const toolName = aiMessage.tool_calls?.[0]?.name;
+    if (toolName === "Search") {
       return "search_node";
-    } else if (aiMessage.tool_calls && aiMessage.tool_calls.length > 0 && aiMessage.tool_calls[0].name === "DeleteResources") {
+    } else if (toolName === "DeleteResources") {
       return "delete_node";
+    } else if (toolName) {
+      // Frontend actions (WriteReport, WriteResearchQuestion, etc.) — let
+      // chat_node loop again so CopilotKit's AG-UI handler picks them up.
+      return "chat_node";
     }
   }
   ...
 }
```

The Python sibling in `agents/python/src/lib/chat.py` handles `WriteReport` explicitly (line 114). Bringing the TS path to parity would close the gap.

Surfaced by a static parity check between the two language implementations of the same agent.

Tushar-Khandelwal-2004:
## Summary
Fixes #5607 .

The TypeScript research-canvas graph was ending whenever the model emitted a tool call that was not `Search` or `DeleteResources`. That meant frontend/state tools such as `WriteReport` and `WriteResearchQuestion` could be dropped before the agent handled them.

## Changes
- Extract the first AI tool call name once in the router.
- Keep existing routing for `Search` and `DeleteResources`.
- Route any other tool call back through `chat_node` so write/frontend tool calls are handled instead of falling through to `END`.

## Testing
- `git diff --check`
- TypeScript transpile/syntax check for `agent.ts`

### DIFF
--- a/examples/v1/research-canvas/agents/typescript/src/agent.ts
+++ b/examples/v1/research-canvas/agents/typescript/src/agent.ts
@@ -3,7 +3,7 @@
  * It defines the workflow graph and the entry point for the agent.
  */
 
-import { AIMessage, ToolMessage } from "@langchain/core/messages";
+import { AIMessage } from "@langchain/core/messages";
 import { StateGraph, END } from "@langchain/langgraph";
 import { MemorySaver } from "@langchain/langgraph";
 import { AgentState, AgentStateAnnotation } from "./state";
@@ -36,32 +36,27 @@ export const graph = workflow.compile({
 
 function route(state: AgentState) {
   const messages = state.messages || [];
+  const lastMessage = messages[messages.length - 1];
 
   if (
-    messages.length > 0 &&
-    messages[messages.length - 1].constructor.name === "AIMessageChunk"
+    lastMessage &&
+    ["AIMessage", "AIMessageChunk"].includes(lastMessage.constructor.name)
   ) {
-    const aiMessage = messages[messages.length - 1] as AIMessage;
+    const aiMessage = lastMessage as AIMessage;
+    const toolName = aiMessage.tool_calls?.[0]?.name;
 
-    if (
-      aiMessage.tool_calls &&
-      aiMessage.tool_calls.length > 0 &&
-      aiMessage.tool_calls[0].name === "Search"
-    ) {
+    if (toolName === "Search") {
       return "search_node";
-    } else if (
-      aiMessage.tool_calls &&
-      aiMessage.tool_calls.length > 0 &&
-      aiMessage.tool_calls[0].name === "DeleteResources"
-    ) {
+    } else if (toolName === "DeleteResources") {
       return "delete_node";
+    } else if (toolName) {
+      return "chat_node";
     }
   }
-  if (
-    messages.length > 0 &&
-    messages[messages.length - 1].constructor.name === "ToolMessage"
-  ) {
+
+  if (lastMessage && lastMessage.constructor.name === "ToolMessage") {
     return "chat_node";
   }
+
   return END;
 }


BenTaylorDev:
Thanks for the PR, and for digging into the routing. Closing after tracing the mechanism, though — the change is based on a misdiagnosis and would introduce a regression.

**`WriteReport`/`WriteResearchQuestion` aren't actually dropped.** `chat_node` (`chat.ts`) intercepts both tool calls *before* `route()` runs, returning the AI message plus a synthetic `ToolMessage` ("Report written." / "Research question written."). With the append reducer from `CopilotKitStateAnnotation.spec`, the last message is that `ToolMessage`, so current `route()` already hits its `ToolMessage → "chat_node"` branch. They never fall through to `END`.

**The Python reference matches current `main`, not this PR.** Python's `chat_node` handles those two tools in-node and defaults `goto="__end__"` for any *other* tool call — i.e. it ends the graph on an unrecognized tool call, exactly like the TS version does today.

**The core change risks a regression.** `else if (toolName) return "chat_node"` routes any unrecognized tool call (e.g. a genuine frontend `useCopilotAction`) back to `chat_node` instead of ending. Ending is correct there: it yields control to the frontend to execute the action and resume with a `ToolMessage`. Routing back to `chat_node` leaves a dangling `tool_call` with no matching `ToolMessage`, which triggers the provider's "tool_calls must be followed by tool messages" error (or a chat_node↔chat_node loop).

Happy to revisit if there's a concrete repro where a tool call is genuinely dropped.

PR #5629, opened 2026-06-23
fix(react-ui): sanitize raw HTML in Markdown renderer to prevent XSS

jpr5:
## Summary

While auditing `@copilotkit/react-ui`, we found that the legacy chat Markdown renderer (`packages/react-ui/src/components/chat/Markdown.tsx`) runs `react-markdown` with `rehype-raw` and **no HTML sanitizer**. `rehype-raw` re-enables parsing of raw HTML embedded in Markdown, and this renderer is the default path for every assistant message (`AssistantMessage.tsx`). Because assistant/agent output is attacker-influenceable (prompt injection via RAG/retrieved docs, tool/web-fetch results, or a compromised agent), raw HTML could reach the DOM in the host app's origin — a cross-site scripting sink (CWE-79).

Note: the v2 `react-core` path already renders via sanitizing Streamdown; only this legacy `react-ui` renderer was exposed.

## Fix

Add `rehype-sanitize` to the rehype plugin chain, positioned as the **terminal pass**:

```
[rehypeRaw, ...(rehypePlugins ?? []), rehypeSanitize]
```

Sanitize runs **last**, after any consumer-supplied `rehypePlugins`, so it cannot be bypassed by a downstream plugin re-introducing raw HTML. The default render path (no consumer plugins) is unchanged in behavior. Adds `rehype-sanitize` as a dependency and `react-dom` (caret range) as a devDependency for the SSR-based regression test.

## Red → green

A regression test installs a malicious consumer `rehypePlugins` entry that injects a raw `<script>` node into the HAST:
- **Before** (sanitize not terminal): payload survived to output — `…safe content</div><script>alert("pwned")</script>` — bypass reproduced.
- **After** (sanitize terminal): payload stripped; test passes.

The suite also covers `<script>/<style>/<base>/<form>/<iframe>`, event handlers, and `javascript:` URLs, and asserts legitimate Markdown/GFM features still render.

## Risk

**Medium — CVSS:3.1/AV:N/AC:L/PR:N/UI:R/S:U/C:L/I:L/A:N (4.8).** `react-ui` is a `"use client"` boundary, so the "arbitrary `<script>` under SSR/RSC" tier is non-default (host-misconfiguration-dependent); on the client path React already neutralizes `onerror`/`javascript:`/`<script>`. The realistic residual set this closes is `<base href>` (relative-URL/redirect hijack), `<style>`/CSS injection, DOM-clobbering, and `<form action>`. All `react-ui` 1.x consumers on the default chat are exposed today.

## Functionality retention

Full react-ui suite green (**53/53**). Verified preserved: `language-*` code classNames (syntax highlighting), GFM tables/strikethrough, the streaming cursor, and links. The default GitHub sanitize schema sufficed — no custom schema needed.

## Follow-ups (out of scope here)

- `remarkMath` is enabled with no `rehype-katex`, so `$$…$$` renders as a code-block widget rather than math — add KaTeX (+ schema) or drop `remarkMath`.
- `code` renderer robustness: it ignores react-markdown's `inline` prop and `String(children)` on multi-node arrays yields `"[object Object]"`.
- Custom components spread `{...props}` (leaking `node`), and `{...components}` lets callers override the hardened `<a>`.
- Add CSP guidance to docs (`base-uri 'self'`, `form-action 'self'`, `script-src`) as defense-in-depth.

## Test plan

- [x] `pnpm -F @copilotkit/react-ui test` — 53/53
- [x] `nx run @copilotkit/react-ui:build`
- [x] red→green on the consumer-plugin injection test

### DIFF
--- a/packages/react-ui/package.json
+++ b/packages/react-ui/package.json
@@ -63,6 +63,7 @@
     "react-markdown": "^10.1.0",
     "react-syntax-highlighter": "^15.6.1",
     "rehype-raw": "^7.0.0",
+    "rehype-sanitize": "^6.0.0",
     "remark-gfm": "^4.0.1",
     "remark-math": "^6.0.0"
   },
@@ -72,6 +73,7 @@
     "eslint": "^8.56.0",
     "postcss": "^8.4.20",
     "react": "^18.2.0",
+    "react-dom": "^19.2.0",
     "tailwind-config": "workspace:*",
     "tailwindcss": "^3.3.0",
     "tsconfig": "workspace:*",
--- a/packages/react-ui/src/components/chat/Markdown.tsx
+++ b/packages/react-ui/src/components/chat/Markdown.tsx
@@ -1,9 +1,12 @@
-import { FC, memo, useMemo } from "react";
-import ReactMarkdown, { Options, Components } from "react-markdown";
+import type { FC } from "react";
+import { memo, useMemo } from "react";
+import type { Options, Components } from "react-markdown";
+import ReactMarkdown from "react-markdown";
 import { CodeBlock } from "./CodeBlock";
 import remarkGfm from "remark-gfm";
 import remarkMath from "remark-math";
 import rehypeRaw from "rehype-raw";
+import rehypeSanitize from "rehype-sanitize";
 
 const defaultComponents: Components = {
   a({ children, ...props }) {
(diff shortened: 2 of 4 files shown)


jpr5:
@AlemTuzlak @tylerslaton Need your eyes here.  Think I found something.

ranst91 (approved):

PR #5631, opened 2026-06-23
fix(react-core): resolve renderAndWaitForResponse silent completion on abandonment

Akarshkushwaha:
## What does this PR do?

Resolves a bug where `renderAndWaitForResponse` silently completes with an empty result when the user abandons the form/interaction (e.g., closing the tab or navigating away). 

Changes introduced:
- **Component Unmount Signal**: `useHumanInTheLoop` now explicitly rejects its pending promise if the component unmounts, generating an "abandonment" error which is propagated properly to the backend.
- **Backend AbortSignal Integration**: The tool handler now listens to the global `RunHandler` `AbortSignal` and rejects the promise if the agent explicitly aborts the run.
- **Configurable Timeouts**: Added a new optional `timeout` configuration property to `ReactHumanInTheLoop` options, allowing forms to automatically time out and reject the promise if the user fails to respond within a given timeframe.
- **Updated Tests**: Updated the execution context persistence E2E tests in `use-human-in-the-loop.e2e.test.tsx` to align with the new expectation that unmounting the component aborts the pending tool call.

## Related PRs and Issues

- Fixes #5554

## Checklist

### DIFF
--- a/packages/react-core/src/v2/hooks/__tests__/use-human-in-the-loop.e2e.test.tsx
+++ b/packages/react-core/src/v2/hooks/__tests__/use-human-in-the-loop.e2e.test.tsx
@@ -1410,19 +1410,15 @@ describe("HITL Thread Reconnection Bug", () => {
       toggleRemount();
     });
 
-    // After remount, should STILL see executing status
-    // This is the key assertion: executingToolCallIds survives component remounts
-    // because it's tracked at the CopilotKitProvider level
+    // After remount, the unmount should have rejected the promise (abandonment),
+    // so the status becomes Complete (with an error result under the hood)
     await waitFor(() => {
       expect(screen.getByTestId("remount-status").textContent).toBe(
-        ToolCallStatus.Executing,
-      );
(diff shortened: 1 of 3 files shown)


lukasmoschitz:
Thanks @Akarshkushwaha, and sorry for the slow reply.

The abort half of this landed in 7d644f22fa back on 2026-06-22 — `use-human-in-the-loop.tsx` now has a pre-abort check, an `onAbort` listener registered with `{ once: true }`, and a `cleanupAbortRef` that removes it again. That last part is missing here, so main's version is the safer one.

What's left in this branch are two things we can't take:

- The `timeout` option isn't something we've asked for, and it adds public API surface to the hook.
- Rejecting on unmount is the opposite of intended behaviour. `use-human-in-the-loop.e2e.test.tsx:1436` asserts that `executingToolCallIds` **survives** component remounts — that's deliberate, and this PR flips the assertion.

Closing for that reason rather than because the report was wrong; the underlying issue was real and is fixed. Thanks for digging into it.

PR #5633, opened 2026-06-23
fix: reject pending human-in-the-loop promise on run abort (#5554)

NathanTarbert:
## Summary

Fixes #5554

When a backend agent calls a frontend tool via `renderAndWaitForResponse` (the `useHumanInTheLoop` hook) and the run is aborted (`stopAgent`/`abortRun`) while the form is still pending, the handler promise was settled only by `respond()` — so on abort it either hung forever or silently resolved to an empty string. The backend received an empty `tool_call_result` (no error), which downstream agent logic interpreted as "no input" — a silent state corruption.

## Root cause

`useHumanInTheLoop` (`packages/react-core/src/v2/hooks/use-human-in-the-loop.tsx`) created its handler promise capturing only `resolve`, and ignored the `AbortSignal` that the core `RunHandler` already passes to tool handlers (`packages/core/src/core/run-handler.ts`). An aborted run therefore never settled the pending promise; an `undefined` result is stringified to `""` (run-handler.ts) → silent empty tool result.

## Fix

Honor the existing `AbortSignal` in the HITL handler:
- If the signal is already aborted when the handler runs, reject immediately.
- Otherwise attach a one-shot `abort` listener that rejects the pending promise with an explicit `Error("Human-in-the-loop interaction aborted")`.
- `respond()` detaches the listener before resolving, so a normal response is unchanged and abort cannot fire after a normal resolve.

Core's existing catch path converts the rejection into an explicit error tool result instead of a silent `""`. Scoped to the one hook; no protocol change, no new timeout API, and **unmount is deliberately not touched** (to avoid regressing reconnect/remount-resume).

## Tests added

`packages/react-core/src/v2/hooks/__tests__/use-human-in-the-loop.e2e.test.tsx` — drives a HITL tool to the executing state, aborts the run without calling `respond()`, and asserts an explicit non-empty error surfaces (via `onToolExecutionEnd`) rather than a silent empty result. Fails before the fix (handler hangs), passes after.

## Checklist

- [x] Failing test written and confirmed failing before the fix
- [x] Fix applied, test passes
- [x] Full `@copilotkit/react-core` suite passes (1291 passed; reconnect/remount tests green)
- [x] Build succeeds (`nx build @copilotkit/react-core`)
- [x] Formatter passes


contextablemark (approved):
looks good.

PR #5640, opened 2026-06-23
fix(runtime): set telemetry license token for v2 SSE runtimes

BenTaylorDev:
## Problem

Self-hosted users on the v2 SSE runtime never get a `telemetry_id` on their runtime telemetry events — even with a license token configured.  The root cause is in `packages/runtime/src/v2/runtime/core/runtime.ts`:

- `CopilotIntelligenceRuntime` **did** call `telemetry.setLicenseToken(...)` in its constructor.
- `BaseCopilotRuntime` and `CopilotSseRuntime` **did not**.

`telemetry_id` is derived only inside `setLicenseToken` (`parseAndWarnTelemetryId`). If it's never called, every event the runtime emits is sent anonymously. SSE-mode handlers (`handle-connect`, `handle-run`, `sse-response`) all emit `oss.runtime.*` events through the shared telemetry singleton, so all of them went out unattributed for SSE users.

## Fix

Hoist the license-token resolution (`options.licenseToken ?? COPILOTKIT_LICENSE_TOKEN`) and the `telemetry.setLicenseToken` call **into `BaseCopilotRuntime`'s constructor**, so SSE and Intelligence runtimes attribute telemetry identically. The now-redundant duplicate is removed from `CopilotIntelligenceRuntime` (its `licenseChecker` stays). The v1 `CopilotRuntime` already set the token and is unchanged.

## Test coverage — every construction path into the endpoints

The token is set at construction time and all endpoints share one telemetry singleton, so covering every runtime construction path covers every endpoint.

- **`runtime-license-telemetry.test.ts`** — `CopilotSseRuntime` and `CopilotIntelligenceRuntime` (direct) + the `CopilotRuntime` shim (both SSE and Intelligence delegates), each across `{explicit option, COPILOTKIT_LICENSE_TOKEN fallback, none}`. Asserts the token is set **exactly once** (guards against a double-set after the hoist).
- **`sse-license-telemetry.integration.test.ts`** — end-to-end: an SSE runtime built with a license token forwards it all the way to `lambdaClient.send` on a real Express endpoint request.
- **`copilot-runtime-license-telemetry.test.ts`** — regression guard for the v1 `CopilotRuntime` path (already worked, previously untested — exactly the kind of untested path that let this gap appear).

The new tests are **red before the fix** (the SSE-path assertions fail) and green after.

## Verification

- New + existing v2 runtime suite: **770/770 pass**, no regressions.
- Lint: 0 new warnings. Build: green (types clean). Format: clean (oxfmt).

### DIFF
--- a/packages/runtime/src/lib/runtime/__tests__/copilot-runtime-license-telemetry.test.ts
+++ b/packages/runtime/src/lib/runtime/__tests__/copilot-runtime-license-telemetry.test.ts
@@ -1,0 +1,60 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+import { CopilotRuntime } from "../copilot-runtime";
+import telemetry from "../../telemetry-client";
+
+/**
+ * The v1 (GraphQL) CopilotRuntime is a separate endpoint path from the v2
+ * runtime, with its own constructor. It already forwards the license token to
+ * telemetry — this pins that behavior so the v1 path can't silently regress
+ * into anonymous telemetry the way the v2 SSE path did.
+ */
+describe("v1 CopilotRuntime — telemetry license token", () => {
+  // Real JWT shape with telemetry_id so the parser doesn't warn.
+  const TOKEN = `header.${Buffer.from('{"telemetry_id":"abc-123"}').toString(
+    "base64url",
+  )}.sig`;
+
+  let setLicenseTokenSpy: ReturnType<typeof vi.spyOn>;
+  let originalEnv: string | undefined;
+
+  beforeEach(() => {
+    setLicenseTokenSpy = vi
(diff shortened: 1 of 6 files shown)


marthakelly:
Highest-Priority Findings

  1. Store the resolved license token once as a protected readonly base field
  (runtime.ts:270-271/:335) — collapses the duplicated ?? process.env... without merging the two
  concerns, and structurally closes the "two sites must agree" invariant that's currently held
  together by a comment. This is the one finding a senior reviewer is most likely to raise.
  2. Integration-level env-fallback test — the only coverage gap with real user value, since
  env-var-only config is the exact scenario this PR exists to fix. Non-blocking but the most
  worthwhile of the "should add" list.

tylerslaton (approved):

PR #5646, opened 2026-06-23
fix(showcase): hide harness-only plumbing from generated doc snippets

NathanTarbert:
## Description

The showcase agents rely on a few test-harness-only helpers at runtime — `makeChatOpenAI`, `withForwardedHeaders`, the CVDIAG instrumentation — that exist so aimock can match fixtures during CI. Those helpers were being copied straight into the docs snippets generated from the showcase source, so anyone following the docs would hit imports like `./openai-headers` that don't exist outside the harness. QA flagged it on a LangGraph.js page, but it turned out to span 13 integrations across TypeScript, Python, and C#.

## The approach

Rather than teach the bundler to rewrite "X into Y" (which bakes one framework's assumptions into a shared tool), the substitution lives in the source via a small marker convention:

```
// @doc-replace
const model = makeChatOpenAI(config, { model: "gpt-5.4" });   // runs in CI, unchanged
// @doc-as
// const model = new ChatOpenAI({ model: "gpt-5.4" });        // what the docs show
// @doc-end
```

The `@doc-replace` block is the real code that still runs. The `@doc-as` block is a comment the bundler un-comments into the published snippet. An empty `@doc-as` just drops the block from the docs (useful for harness-only imports). It supports both `//` and `#` comments, and the bundler never guesses a replacement — each source owns its own.

## Scope

The `applyDocSwaps` transform plus markers across all 13 affected integrations: langgraph-typescript, built-in-agent, mastra, google-adk, claude-sdk-typescript/python, langgraph-python/fastapi, ag2, agno, strands-typescript, and both ms-agent-dotnet variants. Harness helper *definitions* aren't bundled, so they're left as-is.

## Runtime safety

Nothing about runtime behavior changes — the markers are comments and the live code is byte-identical, so the diff touches zero real code lines. Every per-integration build check on CI confirms it.

## Tests

Unit coverage for the transform (swaps, omits, both comment styles, indentation, malformed input throws), plus a guard test that regenerates the bundles and fails if any harness identifier shows up in a published snippet. That guard is the real safety net: if a future demo forgets the markers, CI catches it before the docs ship broken code.


showcase/scripts/doc-snippets-no-harness-plumbing.test.ts
```diff
@@ -0,0 +1,65 @@
...
+  "makeChatOpenAI",
+  "openai-headers",
+  "withForwardedHeaders",
+  "_header_forwarding",
+  "withCvdiagBackend",
+  "CVDIAG",
+  "x-aimock-context",
+  "x-diag-hops",
+  "x-diag-run-id",
+  "copilotkit_forwarded_headers",
```
contextablemark, line 30: The guard keys off a hardcoded identifier list, but a few harness-only identifiers this PR actively swaps out aren't in it — most notably `forwardingFetch` and the `header-forwarding` import path. Files that import `forwardingFetch` alone (e.g. `built-in-agent/src/lib/factory/*.ts`, `strands-typescript/src/agent/model-factory.ts`) contain neither `withForwardedHeaders` nor any other listed string, so a future forgotten marker there would ship an unresolvable `../header-forwarding` import and this guard would stay green.

Note `"header-forwarding"` (hyphen) is deliberately distinct from the existing `"openai-headers"` / `"_header_forwarding"` (underscore) entries — `includes()` won't match across them, which is exactly why it's currently uncovered; it matches both `@/lib/header-forwarding` and `../header-forwarding`. (`extractForwardedHeaders` is intentionally left out — it's a reader-facing local function kept in the snippets.)

```suggestion
  "copilotkit_forwarded_headers",
  // Header-forwarding / CVDIAG helpers that get swapped out of snippets.
  // A forwardingFetch-only file has no other listed identifier, so without
  // these a forgotten marker would ship a broken "../header-forwarding" import.
  "forwardingFetch",
  "header-forwarding",
  "diagOutboundHeaders",
```

NathanTarbert: Good catch — applied the suggestion and added `forwardingFetch`, `header-forwarding`, and `diagOutboundHeaders` to the forbidden list.

Turning on the stronger guard surfaced two real leaks it now correctly catches, both fixed in this branch:
- `strands-typescript/src/agent/tools.ts`: the `import { forwardingFetch } from "./header-forwarding.js"` was sitting outside the `@doc-replace` block (its *usage* was wrapped, the import wasn't), so it would have shipped a broken `./header-forwarding` import. Folded it into the existing drop block.
- `claude-sdk-typescript/src/agent_server.ts`: one `diagOutboundHeaders(forwardedHeaders)` call site (the a2ui secondary-call region) was unwrapped. Wrapped it to match the other three sites.

Guard is green across all bundled snippets again. Also merged latest `main` to clear the conflicts.

NathanTarbert:
Merged current `main` in and this is green again: `showcase/scripts` is 80 files / 2558 tests passing, `validate-parity` 21/21, oxlint clean. Net change against `main` is 140 files, all under `showcase/`.

Two things from the merge worth flagging rather than leaving in the diff for someone to find.

**There's an inconsistency in the `@doc-as` bodies that I did not want to resolve unilaterally.** Several mastra routes already annotated on this branch rewrite their snippet to the v1 `copilotRuntimeNextJSAppRouterEndpoint` API, while the live code uses v2 `createCopilotRuntimeHandler` — so those published snippets teach an older API than the code actually runs. For the four mastra routes newly annotated here I used the minimal swap instead: drop only the `withForwardedHeaders` wrapper and keep the modern handler. That means both conventions are currently in this PR. The minimal swap is the one I'd argue for, since a snippet that teaches a superseded API is a worse failure than one that shows a slightly plainer setup, but if the intent was deliberately to publish the v1 shape I'll convert the four to match rather than split the difference.

**`main` independently adopted two workarounds for this same problem** while the branch sat: hoisting harness imports above `@region`, and the `runChatNode(state, config, model)` refactor with a duplicated `chatNodeWithHeaders`. Those fix region snippets but not the whole-file `/code` view that `collectDemoFiles` also bundles, so they're complementary to doc-swap rather than redundant, and I kept main's structure and re-applied markers to it. Retiring them in favour of doc-swap alone would be a reasonable follow-up but is outside this PR.

@contextablemark, your original point about the guard matching a bare word is what caught 14 of the 15 new demos main added during the drift — two of them leaking only through prose comments naming `copilotkit_forwarded_headers` and `x-aimock-context`, which is the same false-positive class you raised. Those use doc-swap so the source keeps its full rationale and only the snippet is trimmed.

One note on the commit: the pre-commit hook's `lint-fix` runs repo-wide, so on a merge this size it auto-fixed and staged 11 files outside `showcase/` (`import type` splits, and a `.reverse()` → `.toReversed()` in `packages/channels-slack`). I reverted all 11 to main's content, so nothing outside the showcase is in this diff.

BenTaylorDev (changes requested):
The core mechanism is correct, and I verified it end to end. One unrelated change set needs to come out before this lands.

## The premise holds, and the fix works

I checked this independently of the PR's own tests. I rebuilt the `applyDocSwaps` transform, applied it to every file the bundlers actually collect (manifest `highlight:` paths plus everything under `src/app/demos/`), and grepped the result for harness identifiers on non-comment lines.

| Ref | Candidate files | Files still leaking harness plumbing |
|---|---|---|
| `origin/main` | 5604 | **85** |
| PR head `b83f140` | 5604 | **0** |

That is a wider net than the guard test in this PR, which greps the four output bundles for a fixed identifier list. The design choice is right too: the substitution is authored in the source, so the bundler never guesses a replacement. Unit coverage for the transform is thorough — 11 cases, including both comment styles and four malformed-input throws.

## 1. Blocking — 31 unrelated files, 5 of which break frontend parity

Only 2 of the 33 marker-free changed files belong to this PR: the two bundler call sites. The other 31 are a mechanical sweep that splits inline `type` imports into separate `import type` statements, plus three style tweaks (`?? {}` removal in `next.config.ts`, `.reverse()` -> `.toReversed()`, one prettier reflow).

Five of them land on shadcn primitives in one integration only:

```
showcase/integrations/langgraph-typescript/src/components/ui/{alert,button-group,carousel,input-group,tabs}.tsx
```

`alert.tsx` today is byte-identical across 12 integrations. This PR makes `langgraph-typescript` a singleton, which is the divergence that iron rule 2 in `showcase/AGENTS.md` exists to prevent:

```
682e8d36...  built-in-agent, claude-sdk-python, strands-typescript
9272d17e...  ag2, agno, claude-sdk-typescript, crewai-crews, langgraph-fastapi,
             langgraph-python, langgraph-typescript, langroid, llamaindex,
             mastra, spring-ai, strands
```

The 16 demo `.tsx` files under `ms-agent-dotnet/` and `mastra/` are also collected by `collectDemoFiles`, so that churn changes published snippets as a side effect.

Please drop all 31 from this PR. If the import-style sweep is wanted, it belongs in its own change applied to every integration at once.

## 2. Design note — the `@doc-as` bodies are unverified

The live code stays compiled and type-checked. The replacement text is a comment, so nothing checks it. `built-in-agent/src/app/api/copilotkit/route.ts:116` publishes three hand-written export lines that no tool validates.

I confirmed there is no compile step over the bundles: `verify-shell-docs.ts` parses `demo-content.json` for routes and links only, and `test_integration-docs.yml` triggers on `showcase/shell-docs/src/content/**`, so it never sees generated snippets.

The previous state was worse — valid code that was wrong for readers — so this is not a regression. It is worth a follow-up that type-checks the swapped output.

## 3. Two minor items

- `strands-typescript/src/agent/model-factory.ts:20` keeps `import { forwardingFetch } from "./header-forwarding.js"` outside the markers, while its only use at line 87 is swapped away. The file is in no `highlight:` list, so nothing ships today, and the guard catches it the moment someone adds it. Wrapping line 20 costs nothing.
- `extract-starter.ts` copies integration sources verbatim into user-facing starters and preserves `src/app/api/copilotkit/route.ts`. Extracted starters will now carry visible `// @doc-replace` comments. Calling `applyDocSwaps` there would fix both the markers and the harness code starters already ship.

PR #5647, opened 2026-06-23
ci: notify engineering when release PR is created

tylerslaton:
## Problem

Release PR creation still ran dependency lifecycle scripts, and newly created release PRs did not notify engineering.

## Why

The release scripts need installed workspace dependencies for tsx, but they do not need local hook installation. Engineering also needs a visible heads-up when a release PR is ready for CI and review.

## Fix

- Install with `pnpm install --frozen-lockfile --ignore-scripts`.
- Post a guarded `SLACK_WEBHOOK_ENGR` message to #engr only when create-pull-request reports a newly created PR.
- Log a warning instead of failing if the Slack webhook is unset.

### DIFF
--- a/.github/workflows/stable-release.yml
+++ b/.github/workflows/stable-release.yml
@@ -50,6 +50,8 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 15
     environment: npm
+    env:
+      SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK_ENGR }}
     steps:
       - name: Check for existing release PR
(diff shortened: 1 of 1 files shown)


BenTaylorDev:
Superseded — we now notify at **publish** time via `publish-release.yml` (the "🚀 … published to npm" post to #engr, same `SLACK_WEBHOOK_ENGR`), which covers the release-visibility need. This PR's notification fires earlier (release-PR-created), but the team settled on the publish-time signal, so closing.

The one unique piece here is the `--ignore-scripts` hardening on the release install (`stable-release.yml:89` is still plain `--frozen-lockfile` on main) — if we want that, happy to take it as a tiny standalone PR; otherwise it's marginal enough to drop.

PR #5665, opened 2026-06-24
fix: merge agent-specific headers instead of overwriting in applyHeadersToAgent

### LINKED ISSUE #5635: 🐛 Bug: `HttpAgent.headers` are silently overwritten when the agent is registered via `agents__unsafe_dev_only`
### ♻️ Reproduction Steps

1. Stand up a self-hosted AG-UI backend that requires an `Authorization` header (returns 401 without it).
2. Create an agent with a header and register it on the provider:

```jsx
import { HttpAgent } from "@ag-ui/client";
import { CopilotKitProvider } from "@copilotkit/react-core/v2";

const agent = new HttpAgent({
  url: "http://localhost:XXXX/foundry-agent",
  headers: { Authorization: "Bearer <token>" }, // also try mutating agent.headers afterwards
});

<CopilotKitProvider agents__unsafe_dev_only={{ foundry: agent }}>
  {/* ...chat... */}
</CopilotKitProvider>
```

3. Send a message and inspect the outgoing request (Network tab / backend logs).

**Observed:** no `Authorization` header on the request; backend returns 401.
**Workaround that works:** drop the per-agent header and pass it via the provider prop instead:

```jsx
<CopilotKitProvider
  headers={{ Authorization: "Bearer <token>" }}
  agents__unsafe_dev_only={{ foundry: agent }}
>
```

### ✅ Expected Behavior

One of:
- `applyHeadersToAgent` should **merge** the core headers into the agent's existing headers (`agent.headers = { ...agent.headers, ...core.headers }`) rather than replacing them, **or**
- the behavior should be clearly documented: "headers set on an `HttpAgent` passed to `agents__unsafe_dev_only` are ignored; use the `CopilotKitProvider` `headers` prop."

The silent overwrite with no warning is surprising and hard to debug (the request simply goes out without the header).

### ❌ Actual Behavior

`agent.headers` is replaced by `{ ...core.headers }` at registration, so headers configured directly on the agent are dropped without any warning. The only working channel is the provider-level `headers` prop.

### Screenshots

_No response_

### 𝌚 CopilotKit Version

```shell
├── @ag-ui/client@0.0.52
├── @copilotkit/react-core@1.56.2
├── @copilotkit/react-ui@1.56.2
```

### 📄 Logs (Optional)

```shell

```

deepshekhardas:
Fixes #5635

\pplyHeadersToAgent\ in \packages/core/src/core/agent-registry.ts\ was replacing \gent.headers\ entirely with core headers, silently dropping any headers configured directly on an \HttpAgent\ instance passed to \gents__unsafe_dev_only\.

Now merges agent-specific headers after core headers, so agent headers take precedence.

### DIFF
--- a/packages/core/src/core/agent-registry.ts
+++ b/packages/core/src/core/agent-registry.ts
@@ -319,6 +319,7 @@ export class AgentRegistry {
     if (agent instanceof HttpAgent) {
       agent.headers = {
         ...(this.core as unknown as CopilotKitCoreFriendsAccess).headers,
+        ...agent.headers,
       };
     }
   }


onsclom:
Fixed in https://github.com/CopilotKit/CopilotKit/pull/5637, so closing this. Thank you!

PR #5673, opened 2026-06-24
fix(showcase): green strands(+TS) D6 shared-state-read + multimodal (32→34/35)

ranst91:
## Summary

Greens two of the three failing D6 (e2e-full) cells for **both** the `strands` (Python) and `strands-typescript` integrations: **32/35 → 34/35** each.

## Fixes

- **shared-state-read** — the turn-2 fixture leg wrongly pinned `turnIndex: 0`, so the aimock matcher skipped it on turn 2 ("candidate fixture skipped by sequence/turn state") → 404 → turn-2 `sse-missing`. Dropped `turnIndex` to mirror the langgraph-python gold-standard fixture (whose turn-2 leg omits it).
- **multimodal** — `sample.png/pdf/wav` were committed as **git-LFS pointers**, so deploy/test environments without `git lfs pull` served the ~130-byte pointer text as the uploaded file → the agent run never started (`runsFinished=0`). Now shipped as regular binaries via a per-integration `.gitattributes` lfs-unset + real bytes, mirroring langgraph-python's existing convention ("must stay as regular binaries ... so deploy environments without `git lfs pull` serve the actual files").

Both fixes verified locally: `showcase/bin/showcase test {strands,strands-typescript} --d6 --direct --rebuild --isolate` → 34/35 each, shared-state-read + multimodal green.

## Known remaining red (out of scope here)

**gen-ui-declarative** (A2UI dynamic) stays red on both. The A2UI surface **paints correctly** (real dashboard data), but `generate_a2ui` never completes, so the run hangs "Running" (`runsFinished=0`). This reproduces on **real-LLM staging** too, so it is not a fixture/aimock artifact — it is a Strands A2UI-dynamic run-completion issue. The fixtures and frontend are byte-identical to langgraph-python's (which passes), pointing at the Strands adapter's auto-inject completion path. Tracked separately for an adapter-level fix.


jpr5 (approved):
Approved after a three-dimension review (diff-vs-claims/deps, LFS-binary conversion, fixture turnIndex): both fixes are correct and byte-faithful to the langgraph-python gold standard (matching blob SHAs), the .gitattributes LFS-unset correctly overrides the repo-root rule with no regression, and the turnIndex removal's root cause is confirmed with no over-broadened matching. The ag-ui-strands adapter bumps are in-scope support for the strands column. CI green.

PR #5676, opened 2026-06-24
fix(showcase/google-adk): all 39 in-matrix D6 cells green — restore post-tool re-invoke loop

ranst91:
## What

Fixes the 6 remaining red D6 cells for the **google-adk** showcase integration, bringing it to **39/39 on the D6 matrix** (under aimock replay, reproduced across two independent full-matrix runs, no in-matrix regression). See the scope/caveats section before reading this as "fully done."

## Root cause + changes

**1. `entrypoint.sh` — remove `ADK_DISABLE_PROGRESSIVE_SSE_STREAMING=1` (the big one).**
A/B tested in-container: with the flag set, ADK's non-progressive aggregation path ends the agentic loop after the first tool round — no post-tool LLM re-invoke. That broke every demo needing a second turn after a tool result:
- `subagents` (research → writing → critique never chained past research)
- `tool-rendering-reasoning-chain` (AAPL → MSFT)
- `shared-state-read-write` (the post-`set_notes` confirmation text)
- `tool-rendering-custom-catchall` (the post-tool narration)

The flag was added to dodge an intermittent "last event is partial" abort on tool-rendering; the in-callback `stop_on_terminal_text` guard (shared by every agent) is intended to cover that. See caveat 1 — that guard's sufficiency is verified under aimock only, not against real Gemini.

**2. `manifest.yaml` — un-skip-list `tool-rendering-reasoning-chain`.** It was marked `not_supported` (vacuous-green) only because of the loop gap; it passes now.

**3. `headless_complete_agent.py` — add `AGUIToolset()`.** Removing the flag unmasked a pre-existing gap: the frontend `highlight_note` tool was never injected/routed, so turn 3 dispatched it server-side and the backend registry rejected it (`Tool 'highlight_note' not found`). langgraph-python auto-injects frontend tools; ADK needs `AGUIToolset()` in the agent's `tools` list (every other frontend-tool google-adk agent has it).

**4. `aimock/d6/google-adk/gen-ui-interrupt.json` — re-key.** The alice pill 503'd: its emit leg was gated `hasToolResult:false`, but the earlier sales pill leaves a tool result in the thread (thread-global). Re-ordered each pill's narration (`toolCallId`) leg before its emit leg and dropped the `hasToolResult` gate, mirroring the reasoning-chain fixture pattern.

## Verification

`showcase test google-adk --d6 --isolate` (full per-pill matrix), run twice on separate isolated stacks:
- Both: `passed=39, failed=0, total=39, state=green`.
- The 6 previously-red cells all pass: shared-state-write, tool-rendering-custom-catchall, subagents, tool-rendering-reasoning-chain, gen-ui-interrupt (sales + alice), gen-ui-headless-complete.
- No regression: all previously-passing cells (incl. all `tool-rendering*` and A2UI: gen-ui-declarative, gen-ui-a2ui-fixed, beautiful-chat) stay green.

## Scope and caveats (read before merging)

1. **The flag removal is verified under aimock only, NOT against real Gemini.** `ADK_DISABLE_PROGRESSIVE_SSE_STREAMING=1` was originally added for a *real-Gemini intermittent* "last event is partial" abort. D6 is deterministic aimock replay and cannot reproduce that intermittent condition, so the "0 partial-aborts" result does not prove the `stop_on_terminal_text` guard is sufficient against real Gemini. There is a real (unmeasured) risk this re-introduces the partial-abort on real-LLM tool-rendering runs. Recommend a real-Gemini smoke of the tool-rendering demos before relying on this in a real-LLM context.
2. **"39/39" is the in-matrix set, not every feature.** Three features remain `not_supported` and are excluded from the matrix (not fixed): `interrupt-headless`, `reasoning-default-render`, `agentic-chat-reasoning`.
3. **Local verification only.** D6 is not CI-gated for google-adk; evidence is two isolate runs on one machine, not an independent CI signal.

### DIFF
--- a/showcase/aimock/d6/google-adk/gen-ui-interrupt.json
+++ b/showcase/aimock/d6/google-adk/gen-ui-interrupt.json
@@ -1,15 +1,26 @@
 {
   "_meta": {
-    "description": "D6 fixtures for google-adk / gen-ui-interrupt. Strategy-B scheduling flow (no LangGraph interrupt()): each pill is a 2-leg round-trip — leg 1 (hasToolResult:false) returns a `schedule_meeting` tool call routed to the frontend useHumanInTheLoop picker; leg 2 (toolCallId after the user picks a slot) returns the confirmation narration. toolCallIds (call_d5_schedule_sales_001 / _alice_001) match the shared d5 contract. Cribbed from the orphaned schedule_meeting entries previously parked in hitl-in-chat.json (that demo uses book_call).",
+    "description": "D6 fixtures for google-adk / gen-ui-interrupt. Strategy-B scheduling flow (no LangGraph interrupt()): each pill is a 2-leg round-trip — an emit leg returns a `schedule_meeting` tool call routed to the frontend useHumanInTheLoop picker; a narration leg (gated on that call's toolCallId) returns the confirmation after the user picks a slot. The narration (toolCallId) leg is ordered BEFORE the emit leg so it wins once the pill's own tool result is the last message, and the emit leg matches on userMessage+toolName only — NOT hasToolResult, which is thread-global and would wrongly fail the second pill once an earlier pill left a tool result in history (the alice pill 503'd that way). toolCallIds (call_d5_schedule_sales_001 / _alice_001) match the shared d5 contract.",
     "sourceFile": "d5-all.json",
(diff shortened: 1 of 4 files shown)


contextablemark (approved):
Looks good

PR #5680, opened 2026-06-24
Isolate MAF agent instructions per request

### LINKED ISSUE #5659: 🐛 Bug: MAF showcase agents mutate shared singleton's default_options["instructions"] per-request — concurrent users leak system prompts and useAgentContext
### ♻️ Reproduction Steps

The bug fires under any concurrent load against the singleton agents mounted at `/agent-config` and `/readonly-state-agent-context` in
`showcase/integrations/ms-agent-python/src/agent_server.py`. A self-contained Python reproducer (no LLM, no infra) — faithful port of
`AgentConfigFrameworkAgent.run` at [`agent_config_agent.py:106-130`] https://github.com/CopilotKit/CopilotKit/blob/2ea3e6670285/showcase/integrations/ms-agent-python/src/agents/agent_config_agent.py#L106-L130):

```python
import asyncio


class StubAgent:
    def __init__(self, instructions: str):
        # Mirror agent_framework.Agent: instructions live in default_options dict.
        self.default_options = {"instructions": instructions}


_LLM_OBSERVATIONS: list[tuple[str, str]] = []


async def _stub_llm_call(agent: StubAgent, request_id: str, delay_s: float) -> None:
    # Mirror the framework: read self.agent.default_options["instructions"]
    # at the moment a model call fires.
    await asyncio.sleep(delay_s)
    _LLM_OBSERVATIONS.append((request_id, agent.default_options["instructions"]))


class BuggyAgentConfigFrameworkAgent:
    """Faithful port of AgentConfigFrameworkAgent.run."""

    def __init__(self, agent: StubAgent):
        self.agent = agent

    async def run(self, request_id: str, props: dict, llm_delay_s: float) -> None:
        system_prompt = f"<{props['tone']}+{props['expertise']}>"
        options = self.agent.default_options
        previous_instructions = options.get("instructions")
        options["instructions"] = system_prompt
        try:
            await _stub_llm_call(self.agent, request_id, llm_delay_s)
        finally:
            if previous_instructions is None:
                options.pop("instructions", None)
            else:
                options["instructions"] = previous_instructions


async def main() -> None:
    # The agent_server.py pattern: ONE singleton agent instance, shared
    # across all concurrent requests routed to /agent-config.
    singleton = StubAgent(instructions="<neutral fallback>")
    wrapper = BuggyAgentConfigFrameworkAgent(singleton)

    # Two concurrent users with different config. A's LLM call resolves
    # faster (100ms) than B's (300ms) — realistic for ChatCompletions
    # with different prompt sizes or model providers.
    await asyncio.gather(
        wrapper.run("USER_A", {"tone": "professional", "expertise": "expert"}, 0.1),
        wrapper.run("USER_B", {"tone": "casual",       "expertise": "beginner"}, 0.3),
    )

    for rid, instr in _LLM_OBSERVATIONS:
        print(f"  {rid} saw: {instr!r}")
    print(f"\nFinal singleton.default_options['instructions'] = {singleton.default_options['instructions']!r}")


asyncio.run(main())
```

Save as `repro.py`, run `python repro.py`. Observed output:

```
  USER_A saw: '<casual+beginner>'
  USER_B saw: '<neutral fallback>'

Final singleton.default_options['instructions'] = '<professional+expert>'
```

Both users' system prompts are wrong, and the singleton's
`default_options["instructions"]` is left in a stale state instead of
the original `<neutral fallback>`.

The same race exists at
[`readonly_state_agent_context.py:79-94`](https://github.com/CopilotKit/CopilotKit/blob/2ea3e6670285/showcase/integrations/ms-agent-python/src/agents/readonly_state_agent_context.py#L79-L94) —
the only difference is what's being leaked
(`useAgentContext` payload instead of tone/expertise settings).

### ✅ Expected Behavior

Each concurrent request's LLM call should see the system prompt built from **that request's** `forwardedProps` (resp. `useAgentContext`
data). Two users hitting `/agent-config` simultaneously must not observe each other's tone/expertise/length configuration, and two users hitting `/readonly-state-agent-context` simultaneously must not observe each other's UI context payload.

### ❌ Actual Behavior

The "save previous / mutate / finally-restore previous" guard is unsound under concurrency. Trace, given a singleton `self.agent.default_options["instructions"] = "<fallback>"` and two overlapping requests A and B:

| time | event | `options["instructions"]` |
|---|---|---|
| T=0   | A.run enters; saves `previous_A = <fallback>` | `<fallback>` |
| T=0   | A sets to A's prompt | `<prof+expert>` |
| T=0   | A awaits LLM call (100ms) | `<prof+expert>` |
| T=0+ε | B.run enters; reads `previous_B = <prof+expert>` (A's value!) | `<prof+expert>` |
| T=0+ε | B sets to B's prompt | `<casual+beginner>` |
| T=0+ε | B awaits LLM call (300ms) | `<casual+beginner>` |
| T=100 | A's LLM call fires; reads `<casual+beginner>` — **A sees B's prompt** | `<casual+beginner>` |
| T=100 | A's `finally` restores to `previous_A = <fallback>` | `<fallback>` |
| T=300 | B's LLM call fires; reads `<fallback>` — **B sees fallback, not its own** | `<fallback>` |
| T=300 | B's `finally` restores to `previous_B = <prof+expert>` (A's stale value) | `<prof+expert>` |

After both finish, the singleton is left with `<prof+expert>` instead of `<fallback>` — every subsequent serial request that doesn't enter the bug branch will also pick up the wrong baseline.

The bug's blast radius depends on which agent it's in:

- **`agent_config_agent.py`**: cross-tenant leak of UI-facing tone config. Annoying for the user, not catastrophic.
- **`readonly_state_agent_context.py`**: the whole point of that demo is to thread per-user UI context (`useAgentContext` payload) into the
  system prompt for one specific user. Under concurrency that payload  leaks across users — which can include the page state of the *other* user's dashboard, sensitive scratchpad data, or anything an application has put in `useAgentContext`.

### Screenshots

_No response_

### 𝌚 CopilotKit Version

```shell
Files affected:

- `showcase/integrations/ms-agent-python/src/agents/agent_config_agent.py:98-130`
- `showcase/integrations/ms-agent-python/src/agents/readonly_state_agent_context.py:60-94`
```

### 📄 Logs (Optional)

```shell

```

Tushar-Khandelwal-2004:
## Summary
Fixes #5659.

The MS Agent Framework showcase mounts several agents as FastAPI singletons. Two of those agents were temporarily mutating the singleton agent's `default_options["instructions"]` per request, which can leak one user's generated system prompt or `useAgentContext` payload into another concurrent request.

## Changes
- Added a request-scoped instruction proxy that overlays `instructions` into per-run options instead of mutating the shared agent.
- Updated the Agent Config and readonly `useAgentContext` agents to use the request-scoped path.
- Added a regression test that exercises concurrent instruction overlays and verifies shared defaults stay unchanged.

## Testing
- `python -m pytest showcase/integrations/ms-agent-python/tests/python/test_request_scoped_instructions.py`
- `python -m py_compile ...`
- `git diff --check`

### DIFF
--- a/showcase/integrations/ms-agent-python/src/agents/_request_scoped_instructions.py
+++ b/showcase/integrations/ms-agent-python/src/agents/_request_scoped_instructions.py
@@ -1,0 +1,67 @@
+"""Run MS Agent Framework agents with request-local instructions.
+
+Agent Framework stores the agent system prompt in ``default_options``. These
+showcase agents are mounted as FastAPI singletons, so mutating that shared
+dictionary for one request can leak instructions into another concurrent
+request. This module keeps the shared agent immutable and overlays
+``instructions`` only on the individual ``run`` call.
+"""
+
+from __future__ import annotations
+
+from collections.abc import AsyncGenerator
+from typing import TYPE_CHECKING, Any
+
+if TYPE_CHECKING:
+    from ag_ui.core import BaseEvent
+    from agent_framework_ag_ui import AgentFrameworkAgent
+else:
+    BaseEvent = Any
+    AgentFrameworkAgent = Any
+
+
+class _InstructionScopedAgent:
+    """Proxy an Agent while injecting request-local run options."""
+
+    def __init__(self, agent: Any, instructions: str) -> None:
+        self._agent = agent
+        self._instructions = instructions
+
+        base_options = getattr(agent, "default_options", None)
+        self.default_options = (
+            dict(base_options) if isinstance(base_options, dict) else {}
+        )
+        self.default_options["instructions"] = instructions
+
+        self.id = getattr(agent, "id", None)
+        self.name = getattr(agent, "name", None)
+        self.description = getattr(agent, "description", None)
+        self.client = getattr(agent, "client", None)
+        self.mcp_tools = getattr(agent, "mcp_tools", None)
+
+    def __getattr__(self, name: str) -> Any:
+        return getattr(self._agent, name)
+
+    def run(self, messages: Any = None, *args: Any, **kwargs: Any) -> Any:
+        options = dict(kwargs.get("options") or {})
+        options["instructions"] = self._instructions
+        kwargs["options"] = options
+        return self._agent.run(messages, *args, **kwargs)
+
+
+async def run_with_request_instructions(
+    wrapper: AgentFrameworkAgent,
+    input_data: dict[str, Any],
+    instructions: str,
+) -> AsyncGenerator[BaseEvent, None]:
+    """Delegate to ``wrapper`` with instructions scoped to this run only."""
+    from agent_framework_ag_ui._agent_run import run_agent_stream  # type: ignore
+
+    scoped_agent = _InstructionScopedAgent(wrapper.agent, instructions)
+    async for event in run_agent_stream(
+        input_data,
+        scoped_agent,
+        wrapper.config,
+        pending_approvals=wrapper._pending_approvals,
+    ):
+        yield event
--- a/showcase/integrations/ms-agent-python/src/agents/_request_scoped_instructions.py
+++ b/showcase/integrations/ms-agent-python/src/agents/_request_scoped_instructions.py
@@ -10,6 +10,7 @@ request. This module keeps the shared agent immutable and overlays
 from __future__ import annotations
 
 from collections.abc import AsyncGenerator
+from copy import copy
 from typing import TYPE_CHECKING, Any
 
 if TYPE_CHECKING:
@@ -54,14 +55,21 @@ async def run_with_request_instructions(
     input_data: dict[str, Any],
     instructions: str,
 ) -> AsyncGenerator[BaseEvent, None]:
-    """Delegate to ``wrapper`` with instructions scoped to this run only."""
-    from agent_framework_ag_ui._agent_run import run_agent_stream  # type: ignore
+    """Run a request-local agent through the adapter's public API.
 
-    scoped_agent = _InstructionScopedAgent(wrapper.agent, instructions)
-    async for event in run_agent_stream(
-        input_data,
-        scoped_agent,
-        wrapper.config,
-        pending_approvals=wrapper._pending_approvals,
-    ):
+    ``AgentFrameworkAgent`` changed its public entry point from ``run_agent``
+    to ``run`` during its beta period. A shallow wrapper clone preserves its
+    configuration and internal state while substituting only the agent for this
+    invocation; the shared singleton remains untouched.
+    """
+    scoped_wrapper = copy(wrapper)
+    scoped_wrapper.agent = _InstructionScopedAgent(wrapper.agent, instructions)
+
+    run = getattr(scoped_wrapper, "run", None) or getattr(
+        scoped_wrapper, "run_agent", None
+    )
+    if run is None:
+        raise TypeError("AgentFrameworkAgent exposes neither run nor run_agent")
+
+    async for event in run(input_data):
         yield event
--- a/showcase/integrations/ms-agent-python/src/agents/agent_config_agent.py
+++ b/showcase/integrations/ms-agent-python/src/agents/agent_config_agent.py
@@ -7,8 +7,8 @@ per turn.
 The CopilotKit provider's ``properties`` prop is wired through the runtime as
 ``forwardedProps`` on each AG-UI run. Because Microsoft Agent Framework agents
 store their system prompt in ``default_options["instructions"]``, we subclass
-``AgentFrameworkAgent`` and intercept ``run`` to swap in a freshly-built
-instruction string for the duration of each invocation.
+``AgentFrameworkAgent`` and intercept ``run`` to pass a freshly-built
+instruction string as request-local run options.
 
 Invalid or missing values fall back to the corresponding ``DEFAULT_*``
 constant -- this function never raises so the demo can't deadlock on a bad
@@ -24,6 +24,7 @@ from typing import Any, Literal
 from ag_ui.core import BaseEvent
 from agent_framework import Agent, BaseChatClient
 from agent_framework_ag_ui import AgentFrameworkAgent
+from agents._request_scoped_instructions import run_with_request_instructions
 
 Tone = Literal["professional", "casual", "enthusiastic"]
 Expertise = Literal["beginner", "intermediate", "expert"]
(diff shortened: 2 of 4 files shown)


suzunn:
I like that this keeps the singleton `Agent` immutable and pushes `instructions` into the per-run options instead of trying to make the save/restore block concurrency-safe. One detail I would lock down before merging is option preservation: the regression currently proves the two instruction overlays do not leak, but it does not assert that caller-provided `options` survive the proxy hop. A small assertion around the captured calls for values like `store=True` and per-request metadata would protect against a future refactor accidentally replacing the run options wholesale while still passing the instruction-only assertions.

It may also be worth adding the same concurrent test through `run_with_request_instructions` with a stub wrapper, not only `_InstructionScopedAgent` directly. That would catch adapter-level changes in `run_agent_stream` that start reading `default_options` differently from `agent.run(..., options=...)`.

Tushar-Khandelwal-2004:
Thanks for the review! Pushed a fix for both points:

- **Options preservation**: added assertions on the captured call dicts (`calls_by_instruction`) in the existing test, so `store` and `metadata` are now checked alongside `instructions`. A refactor that replaces the options dict wholesale instead of merging into it will now fail this test.

- **Adapter-level coverage**: added a new test, `test_run_with_request_instructions_scopes_concurrent_runs`, that fakes out `agent_framework_ag_ui._agent_run.run_agent_stream` via `sys.modules` patching and runs two concurrent requests through `run_with_request_instructions` directly, not just through `_InstructionScopedAgent`. It also asserts `config` and `pending_approvals` get threaded through correctly.

Let me know if you'd like anything else tightened up!

BenTaylorDev:
Thanks @Tushar-Khandelwal-2004 — the concurrency bug is real (mutating the shared singleton's `options["instructions"]` per request leaks across concurrent runs). One thing to firm up before merge:

`run_with_request_instructions` bypasses `super().run()` and calls the private `agent_framework_ag_ui._agent_run.run_agent_stream(input_data, scoped_agent, wrapper.config, pending_approvals=...)`. The regression test monkeypatches `run_agent_stream` with a fake and swaps in a `SimpleNamespace` wrapper — so it never proves (a) that private API actually exists with that signature in the pinned `agent-framework-ag-ui` beta, or (b) that base `AgentFrameworkAgent.run` is a thin wrapper over it (if `run()` does anything extra — middleware, approvals — this proxy silently drops it).

Could you verify against the pinned SDK that `_agent_run.run_agent_stream` has exactly that signature and that base `run()` delegates identically — ideally by running the showcase rather than the mock? The `_InstructionScopedAgent` proxy itself looks fine; it's the behavioral test that's resting on a fake.

BenTaylorDev (changes requested):
Thanks for tackling this — the shared-singleton race is real and still live on `main`, and you've correctly covered **both** affected sites (agent_config + the readonly `useAgentContext` agent). Two issues need addressing before merge, neither visible in the (fully-mocked) tests.

**1. Prompt-semantics regression via concatenation.** `run_agent_stream` never reads `instructions` from `default_options`; it delegates to the real underlying `Agent`, and `agent_framework` core *concatenates* rather than replaces: `_merge_options` does `result["instructions"] = f"{result['instructions']}\n{value}"` (`agent_framework/_agents.py:150-152`). So the model receives the constructor instructions **plus** the per-request string — the readonly agent's `SYSTEM_PROMPT` ends up duplicated, and agent_config stacks fallback + per-request. Main's save/restore gave the model the per-request prompt *only*. Fix direction: construct the base agents instruction-less, or replace rather than overlay.

**2. Private-API version coupling.** The fix bypasses `super().run()` to call `agent_framework_ag_ui._agent_run.run_agent_stream(...)` and reads private `wrapper._pending_approvals`. Against the pinned floor (`agent-framework-ag-ui==1.0.0b251117`, what `>=1.0.0b251117` names) there is **no `_agent_run` module, no `run_agent_stream`, no `_pending_approvals`** — the import would `ModuleNotFoundError`; it only works on the current `1.0.0`. It also drops `approval_state_store`, which the real base `run` forwards (`_agent.py:145-150`) — harmless for these tool-less agents but a sign the reimplementation froze a stale copy of the base method.

The tests can't catch either: `test_run_with_request_instructions_scopes_concurrent_runs` monkeypatches a *fake* `run_agent_stream` and stubs the agent's `run`, so CI validates the assumed contract, not `_merge_options`.

**To move forward:** run the two agents live against the agent-framework version CI actually resolves, inspect the system prompt the model receives (confirm no duplicated `SYSTEM_PROMPT`), and add a non-mocked assertion on the final merged instruction string. Prefer overriding the public seam over reaching into private internals so it survives across agent-framework releases.

Tushar-Khandelwal-2004:
Thanks @BenTaylorDev pushed a follow-up that addresses both concerns.

* The request-scoped helper now shallow-copies the wrapper and calls its public `run()` API, with `run_agent()` compatibility for the beta floor. This preserves the adapter's own middleware and approval handling instead of reimplementing its internals.
* Both affected base agents are now instruction-less, so the request prompt is the **only** instruction merged into the model call.
* I added a non-mocked assertion against Agent Framework's real `_merge_options` behavior to verify that the final instruction is exactly the request prompt. The previous fallback + request case is also explicitly shown to concatenate.
* The focused suite passes (**4 tests**) against the stable SDK.
* I also verified the beta source: it exposes the public `run_agent()` API and does not have an `_agent_run` module.

BenTaylorDev (changes requested):
The concatenation fix (instruction-less base agents) and the move off the private `_agent_run`/`_pending_approvals` API are both right. But the new indirection introduces an infinite recursion: `scoped_wrapper = copy(wrapper)` keeps the subclass, so `getattr(scoped_wrapper, "run")` resolves to the *overriding* `AgentConfigFrameworkAgent.run` / `ReadonlyContextFrameworkAgent.run`, which calls back into `run_with_request_instructions` → copies again → `RecursionError`, never reaching the base adapter's orchestrator (the old `super().run()` avoided exactly this). The tests miss it because their wrappers are plain classes whose `run` doesn't re-enter the helper. Please invoke the *base* adapter's `run`/`run_agent` (via `super`/MRO, not `getattr` on the subclass copy), and add a test with a real `AgentFrameworkAgent` subclass that delegates through the helper to lock it out.

BenTaylorDev:
Thank you for the work here, @Tushar-Khandelwal-2004. I am closing this, and I want to be clear that it is not because anything is wrong with it.

The race you found was real and your diagnosis was right: mutating the shared singleton's `default_options["instructions"]` per request leaks prompts across concurrent runs. That bug has since been fixed on `main`, by a different approach that landed on 2026-09-01 while this PR was open. Both agents now inject a request-local system message into a copy of `input_data` rather than touching the singleton:

- `showcase/integrations/ms-agent-python/src/agents/agent_config_agent.py:120-141`
- `showcase/integrations/ms-agent-python/src/agents/readonly_state_agent_context.py:93-114`

Concurrency regression tests are in `tests/python/test_readonly_context_isolation.py`, covering both agents and asserting the singleton stays unmutated, which are the acceptance criteria @suzunn set out on #5659. Both files this PR touches now conflict with that fix, so there is no clean way to carry it forward.

On the substance, for the record: you covered both affected call sites, and in round two you correctly moved off the private `_agent_run` / `_pending_approvals` API once the version-coupling problem was raised. The recursion I flagged in round three was also real, and I re-verified it while closing this out. But you should not have been left working on that without being told the underlying bug had been solved elsewhere. That part is on us.

Two rounds of careful and responsive work, and the shipped fix matches the diagnosis you made first.

PR #5695, opened 2026-06-25
fix(vue): inherit chat configuration agent id in useAgent

### LINKED ISSUE #5656: fix(vue): useAgent should resolve agentId from chat configuration (Vue parity for #5533)
## Summary

The Vue `useAgent` hook (`packages/vue/src/v2/hooks/use-agent.ts`) carries the same bug as #5533, which was fixed for `@copilotkit/react-core` in #5592 but **not** for `@copilotkit/vue`. Vue has its own, separate `useAgent` implementation, so the React fix does not flow into it.

When a runtime registers an agent under a **non-default** name (e.g. `agents: { TravelBookingAgent }`) and a bare `useAgent()` consumer is rendered inside a `<CopilotChat agentId="TravelBookingAgent">` subtree without re-passing `agentId`, it resolves to `'default'` and throws after runtime sync:

> useAgent: Agent 'default' not found after runtime sync (runtimeUrl=…). Known agents: [TravelBookingAgent]

## Root cause

[`packages/vue/src/v2/hooks/use-agent.ts:130`](https://github.com/CopilotKit/CopilotKit/blob/main/packages/vue/src/v2/hooks/use-agent.ts#L130) resolves `agentId` straight to `DEFAULT_AGENT_ID`, never consulting the surrounding `CopilotChatConfigurationProvider` — even though the very next line already reads `threadId` from that same config:

```ts
const agentId = computed(() => toValue(props.agentId) ?? DEFAULT_AGENT_ID);
const chatConfig = useCopilotChatConfiguration();
const threadId = computed(() => toValue(props.threadId) ?? chatConfig.value?.threadId);
```

The Vue chat config provider already exposes a resolved `agentId` ([`CopilotChatConfigurationProvider.vue:81`](https://github.com/CopilotKit/CopilotKit/blob/main/packages/vue/src/v2/providers/CopilotChatConfigurationProvider.vue#L81): `agentId: resolvedAgentId.value`), so the fallback is available — it's just not consulted.

This mirrors the React root cause exactly: an explicit prop should win, then the surrounding chat configuration's `agentId`, then the global default.

## Proposed fix

Resolve `agentId` with the same precedence the Vue `CopilotChatConfigurationProvider` and the React `useAgent` already use:

```ts
const chatConfig = useCopilotChatConfiguration();
const agentId = computed(
  () => toValue(props.agentId) ?? chatConfig.value?.agentId ?? DEFAULT_AGENT_ID,
);
```

Note: `chatConfig` is currently declared *after* `agentId`. Vue's `computed` is lazy so referencing it inside the callback works regardless, but moving the `chatConfig` declaration above `agentId` reads cleaner.

## Tests

Add a failing-before/passing-after test mirroring the React reproduction added in #5592 (`packages/react-core/src/v2/hooks/__tests__/use-agent-nondefault-agentid.test.tsx`): a bare `useAgent()` consumer inside a chat configured for a non-default agent must not throw `Agent 'default' not found`, and must inherit the configured agent id. Also cover that an explicit `agentId` prop still wins over the surrounding config.

## Out of scope (already covered)

- `@copilotkit/react-core` — fixed in #5592.
- `@copilotkit/react-native` — has no separate `useAgent`; it re-exports `@copilotkit/react-core/v2/headless`, so the #5592 fix flows in transitively.

## References

- Original issue: #5533
- React fix: #5592

westkite1201:
## Summary

- Make Vue `useAgent` resolve agent IDs from explicit prop, then surrounding chat configuration, then `DEFAULT_AGENT_ID`.
- Add regression coverage for bare `useAgent()` inside a non-default configured chat and for explicit prop precedence.
- Update the Vue parity matrix note for the `useAgent` precedence behavior.

Fixes #5656

## Verification

- `pnpm nx run @copilotkit/vue:test -- src/v2/hooks/__tests__/use-agent.test.ts` (failed before the fix with `Received: "default"`, passes after)
- `pnpm --dir packages/vue exec eslint src/v2/hooks/use-agent.ts src/v2/hooks/__tests__/use-agent.test.ts`
- `pnpm nx run @copilotkit/vue:check-types`
- `pnpm nx run @copilotkit/vue:test` (94 files / 1007 tests)
- `lefthook pre-commit test-and-check-packages` (ran `@copilotkit/vue` test, build, publint, and attw)
- `git diff --check upstream/main...HEAD`

## Notes

- `pnpm nx run @copilotkit/vue:lint` still fails on pre-existing unrelated Vue lint errors across untouched files; the changed files pass eslint directly.


itxaiohanglover:
Good fix! The prop -> chat config -> default precedence makes sense. Test coverage with both scenarios (configured-only and explicit-over-configured) is solid. The PARITY.md update documenting the resolution order is a nice touch.

onsclom (approved):

onsclom:
Thank you for this great fix!

PR #5701, opened 2026-06-25
feat: CopilotDrawer — promote threads management to SDK (shadow-DOM web component + React vertical)

BenTaylorDev:
## CopilotDrawer — promote threads management to the SDK

Implements the hardened [CopilotDrawer design (Threads v2)](https://app.notion.com/p/3883aa3818528190b5d1f9b6ba26dca0). Promotes thread listing/selection/creation/rename/archive/delete out of the example apps and into the SDK as a framework-agnostic shadow-DOM web component plus a React vertical (Angular wrapper included; Vue/RN to follow).

> Note: assumes [CopilotKit#5624](https://github.com/CopilotKit/CopilotKit/pull/5624) is already merged (built on top of its `unarchiveThread` surface).

### What's in this PR (by area)

- **A — `@copilotkit/core`**: thread store with optimistic add/rename/archive/delete (delete rolls back on reject), session-guarded results (a superseded session can't leak an error/rollback), `startNewThread` (non-explicit), `refetchThreads`, `getServerState`/`getServerSnapshot`, per-store memoized selectors, and **non-fatal** realtime diagnostics (a channel-join or metadata-credential failure warns and keeps the already-loaded list rather than blanking it). Registers a `"threads"` feature for catalog gating.
- **B — `@copilotkit/web-components` (new package)**: the shadow-DOM `<copilotkit-drawer>` Lit element — self-contained styles with **build-time token sync** from react-core's `@theme` (checked-in `generated-tokens.ts`, drift-guarded), slot projection for custom rows (`slot="row:{id}"`), license/upsell gating, active/all filtering, and a mobile modal. The element owns view-state; the host owns domain state. Multi-format build: ESM/CJS externalize `lit` (peer dep); the **UMD bundle inlines `lit`** so the advertised `<script>`/CDN path is self-contained.
- **C/E/F — `@copilotkit/react-core`**: `<CopilotDrawer>` wrapper (custom-element interop, two-pronged license gate, scoped chat-input focus-return, registration-gated header launcher); `CopilotChatConfigurationProvider` gains `drawerOpen` + `<768px` modal/drawer mutual-exclusion + a **non-explicit active-thread setter** so a bare `<CopilotDrawer>` connects to the picked thread and shows the welcome screen on "+ New" with no host wiring; `useThreads` gains an `{enabled}` gate and a list-only error channel.
- **D — `@copilotkit/angular`**: `injectThreads` signal wrapper — reactive runtime context (re-dispatches when `wsUrl`/`threadEndpoints` arrive), enabled-gated single-slot registration, bound result methods, and error/loading parity with react-core.

### Quality

- Tests: core, web-components (46), react-core (1370), angular (106) — all green; `tsc --noEmit` clean across all four; full builds pass; oxfmt clean.
- **Code review:** 4 rounds of an unbiased 7-slot review fleet (15 agents/round) — 3 fix rounds + 1 confirming round that converged with zero load-bearing findings. The convergence lever was applied per the requester: the deep focus/overlay-accessibility rework, pagination wiring, and Angular parity gaps are deferred to the documented follow-ups below rather than blocking this PR.

### Follow-ups (intentionally deferred)

- **Accessibility (own PR + a11y test harness):** the mobile-modal Tab trap now runs at the host level and contains the backdrop (fixed in review); remaining a11y work — focus-restore to the opener on close, scroll-lock reference-counting, confirm-delete `alertdialog` focus management, the default-open-on-mobile scroll-lock, and a dedicated a11y test harness.
- **Pagination wiring:** the element exposes `hasMore`/fetch-more UI but nothing triggers it and the wrapper doesn't pass a `limit`; also clear a recovered next-page error and cancel an in-flight next-page on filter change.
- **Error surfacing:** wire `onError` to a transient toast and separate the mutation-error channel from the list-load error (the list no longer blanks on a failed mutation — the rollback restores the row — but mutation failures are currently silent beyond the rejected promise).
- **Angular parity:** mirror react-core's `startNewThread` config-error dismissal, add a list-only `listError`, and refresh intelligence/threadEndpoints in `updateRuntime` on a runtime-URL change.
- **Reach:** Vue + React Native wrappers; de-fork the ~18 example apps onto the SDK drawer; nested-provider active-thread sharing.
- **Minor:** migrate the react-core/angular wrappers to per-store `store.selectors`; honor-or-drop `ThreadEnvironment.fetch` (RxJS `fromFetch` ignores it; both wrappers pass `globalThis.fetch`, so no behavior change today); `threadId=""` controlled-predicate edge; reveal animation for slotted rows; empty-string-timestamp coalescing; `_visibleThreads` recompute; empty-name `aria-label`.


MikeRyanDev (changes requested):
Requesting changes for two blockers before this ships.

1. Required CI is red on the PR head. The failures are not isolated downstream noise: `build`, `compat-check`, `check-types`, `package-quality`, `bundle-size`, and all Node unit matrices ultimately report `@copilotkit/web-components:build` as the failed task, which then prevents `@copilotkit/react-core` and dependents from building/testing. Since this PR publishes `@copilotkit/web-components` and adds it as a `react-core` dependency, we should not merge until the new package builds cleanly in CI.

2. The mobile drawer is shipped as a modal dialog, but focus is not contained when focus is outside `.root`. In `packages/web-components/src/drawer/copilotkit-drawer.ts`, the backdrop button is rendered as a sibling before `.root` (around lines 335-353), while the Tab trap is attached only to `.root` (line 352) and computes focusables only under `.root` (lines 277-323). That means if focus lands on the backdrop or starts elsewhere in the shadow root, Tab can escape the modal instead of cycling within it. For a public mobile modal, please trap at the shadow-root/host level or make all focusable modal-owned nodes part of one trap, and add a regression test for backdrop/outside-root focus.

BenTaylorDev:
Thanks for the careful review — both blockers are addressed.

**1. CI red (`@copilotkit/web-components:build`).** Root cause: the UMD bundle inlines `lit` for a self-contained CDN `<script>`, which also pulls in lit's runtime deps (`@lit/reactive-element`, `lit-html`, `lit-element`, `@lit-labs/ssr-dom-shim`). tsdown flags that transitive bundling as "unintended" and **promotes the warning to a fatal error only under `CI=true`** — so it built locally but exited 1 in CI, failing `web-components:build` and everything downstream. Fixed in `fix(web-components): make the UMD lit-inlining CI-safe` by listing all lit packages in `noExternal` and setting `inlineOnly: false` to declare the bundling intentional. Verified with `CI=true` (build exits 0; `es-check es2018` passes on the UMD; `es2022` on the rest). I also reproduced the green build across `core`/`web-components`/`react-core`/`angular` locally under `CI=true`.

**2. Mobile modal focus trap.** You're right — the Tab trap was bound to `.root` while the backdrop renders as a sibling outside it, so Tab from the backdrop (or any out-of-root node) could escape. Fixed in `fix(web-components): contain mobile-modal focus including the backdrop`:
- Moved the Tab trap from the `.root` `@keydown` to the **host** keydown listener. `keydown` is composed, so it now fires regardless of where focus sits in the shadow root — including the backdrop.
- The backdrop is now part of the focusable cycle (`_composedFocusable` includes it first), so wrap-around contains it at both boundaries.
- Initial focus still lands on the first real control, not the backdrop scrim.
- Added a regression test — *"mobile modal traps Tab at both boundaries, including the out-of-root backdrop"* — that focuses the backdrop and asserts Shift+Tab wraps to the last control and Tab from the last wraps back to the backdrop (a keydown dispatched on the backdrop must reach the host trap).

As an extra check I loaded the built UMD bundle via a plain `<script>` in a browser and confirmed `<copilotkit-drawer>` mounts, renders rows, and emits its events — i.e. the CDN path works end to end.

The broader a11y items (focus-restore to the opener on close, reference-counted scroll-lock, confirm-delete `alertdialog` focus management, default-open-on-mobile) remain tracked as follow-ups in the PR description.

MikeRyanDev (approved):
Re-reviewed the latest head (225a3fc). The package build blocker is resolved: build, compat-check, check-types, package-quality, bundle-size, unit matrices, and dojo e2e checks are green. The mobile focus-trap blocker is also addressed by moving Tab handling to the host, including the backdrop in the focus cycle, preserving initial focus on an inner control, and adding a regression test for backdrop boundary trapping. Approving.

PR #5707, opened 2026-06-25
feat(drawer): CopilotDrawer host theming, mobile launcher, action icons + thread-management/tooltip fixes

BenTaylorDev:
## CopilotDrawer — SDK polish + thread-management fixes

Polish and fixes on top of the shipped `<copilotkit-drawer>` ([#5701](https://github.com/CopilotKit/CopilotKit/pull/5701)), surfaced by real de-fork pilots run against a locally-built SDK + the hosted Intelligence runtime: **langgraph-js** (rich canvas / inline `CopilotChat` / dark theme) and **mastra** (the proverbs CoAgents demo / `CopilotSidebar` / light theme / controlled→uncontrolled provider). Spans **`@copilotkit/web-components`** and **`@copilotkit/react-core`** — no example/de-fork changes here (those wait on a release that publishes web-components + a `react-core` bump that includes `CopilotDrawer`). Captured as Phase-5 acceptance requirements in the [design doc](https://app.notion.com/p/3883aa3818528190b5d1f9b6ba26dca0).

### `react-core` — thread management
- **Thread switching works under `<CopilotKit>`.** `CopilotChatConfigurationProvider` treated any `threadId` prop as controlled, so the auto-minted, non-explicit threadId that `<CopilotKit>`'s v1 bridge seeds blocked the drawer's imperative `setActiveThreadId` / `startNewThread` — thread switching and "+ New" were dead in every app wrapped in `<CopilotKit>`. A non-explicit (`hasExplicitThreadId={false}`) threadId is now overridable; a genuine caller-supplied threadId stays controlled.
- **"+ New" resets the chat.** `CopilotChat` clears messages when switching to a fresh non-explicit thread so the welcome screen shows instead of the prior thread's messages — including when a still-loading `/connect` is superseded by the switch (the aborted connect's stale snapshot is dropped instead of repopulating the view).
- **No upsell flash/stick.** `CopilotDrawer` treats a pending (`null`) license status as loading, not unlicensed, so the upgrade upsell no longer flashes (or sticks) before the runtime reports the license.
- **In-header drawer launcher is mobile-only.** `CopilotModalHeader.DrawerLauncher` toggles `drawerOpen`, which only drives the off-canvas *mobile* drawer; on desktop the drawer is a persistent in-flow panel that ignores `open`, so the launcher was a dead no-op there. It now renders only at ≤767px. On desktop the chat's own open/close (the toggle FAB + the header close, both lucide `X`) is the consistent pair.

### `web-components` — theming, launcher, icons, tooltips
- **Host-theme inheritance (incl. dark mode).** Each token now resolves `--cpk-drawer-* override → host theme var (`--background`/`--card`/`--foreground`/`--border`/…) → built-in light default`. The drawer follows the host app's light/dark theme by inheritance, instead of baking light-only values (it showed a light drawer on a dark app). Custom properties aren't reset by the `:host { all: initial }` guard, so the standalone fallback still holds.
- **Row actions are icon buttons with instant, on-brand tooltips.** Archive / Unarchive / Delete render inline lucide-style SVGs (`currentColor`). Native `title` is replaced by an instant CSS tooltip (matches the React components' Radix `delayDuration: 0`; `title`'s show-delay is browser-fixed at ~1.5s), styled to match the standard CopilotKit tooltip — a primary-colored bubble with an arrow, not a surface/bordered box — and positioned to the side so the list's scroll-overflow never clips it. The tooltip is suppressed while the delete-confirm dialog is open (the clicked trash button keeps `:focus-visible` otherwise, leaving its "Delete" tooltip floating over the dialog).
- **Clipped thread names get a tooltip.** A truncated name exposes its full text via the same instant primary-bubble tooltip as the row actions (not the native `title`), shown only when actually clipped. The name text moved to an inner span (so the outer can host the un-clipped bubble) and the hovered row is z-lifted (each row is its own stacking context from the entry-animation transform, which otherwise painted the bubble under later rows).
- **Archived rows are muted**, no longer struck through.
- **Refetch preserves the list.** A refetch (e.g. Active↔All toggle) keeps the known threads visible; the full loading state only renders on the initial empty fetch.
- **Empty `footer` region hides** (it rendered as a stray bordered box at the bottom). Driven by a `slotchange` listener, mirroring the `memories` region.
- **Self-owned mobile launcher.** When the drawer is a closed off-canvas mobile modal it renders its own floating "open threads" launcher, so it's always openable without the host wiring a header button. The icon is swappable via a `launcher-icon` **slot**, position is themeable via `--cpk-drawer-launcher-top` / `--cpk-drawer-launcher-left` (so a host can center it on its own header controls), and it exposes `part="launcher"`. The richer in-header `CopilotModalHeader.DrawerLauncher` remains an optional integration.

### Tests
- **react-core**: thread-control + non-explicit-override cases on `CopilotChatConfigurationProvider`, the "+ New" message reset on `CopilotChat`, license-pending gating on `CopilotDrawer`, and the mobile-only / desktop-hidden launcher on `CopilotModalHeader`. New regression tests verified red-green; chat + provider suites green, `tsc --noEmit` clean, build green. (The superseded-connect *race* guard is covered by live verification + its synchronous sibling test — an async-timing unit test would be flaky, which the repo treats as a bug.)
- **web-components**: element tests for the icon tooltips + confirm-dialog suppression, the clipped-name tooltip (`data-tooltip` + placeholder), archived styling, refetch-preserves-list, footer hide-when-empty, and mobile launcher render/open + desktop-absent. Drawer suite green, `tsc --noEmit` clean, build + `es-check` green under `CI=true`. (Truncation detection + tooltip stacking are layout/CSS — verified live, not in jsdom.)

### DIFF
--- a/packages/react-core/src/v2/components/chat/CopilotChat.tsx
+++ b/packages/react-core/src/v2/components/chat/CopilotChat.tsx
@@ -226,7 +226,19 @@ export function CopilotChat({
   const isConnecting =
     hasExplicitThreadId && lastConnectedThreadId !== resolvedThreadId;
 
+  // Tracks the threadId the connect effect last ran for, so it can tell a real
+  // thread SWITCH from an incidental re-render (agent identity change, etc.).
+  const previousThreadIdRef = useRef<string | null>(null);
+
+  // Latest explicitness, readable from an async connect that may resolve after
+  // the user has already switched threads (see the stale-connect guard below).
+  const hasExplicitThreadIdRef = useRef(hasExplicitThreadId);
+  hasExplicitThreadIdRef.current = hasExplicitThreadId;
+
   useEffect(() => {
+    const threadChanged = previousThreadIdRef.current !== resolvedThreadId;
+    previousThreadIdRef.current = resolvedThreadId;
+
     // Non-explicit threads skip /connect, but the first runAgent still has to
     // ship the same SDK-generated threadId that the chat UI is rendering.
     agent.threadId = resolvedThreadId;
@@ -236,7 +248,18 @@ export function CopilotChat({
     // ThreadsProvider). The backend has never seen it, so /connect would
     // always 404 — skip the call. A real thread is only created once the
     // user runs the agent for the first time.
-    if (!hasExplicitThreadId) return;
+    if (!hasExplicitThreadId) {
+      // Switching to a fresh, non-backend thread (e.g. startNewThread / the
+      // drawer's "+ New"): there are no messages to /connect for, so drop any
+      // messages carried over from the previously-viewed thread and fall back
+      // to the welcome screen. Guard on an actual threadId change so re-renders
+      // of the current thread (including its first run) never wipe an
+      // in-progress conversation.
+      if (threadChanged && agent.messages.length > 0) {
+        agent.setMessages([]);
+      }
+      return;
+    }
 
     let detached = false;
 
@@ -276,6 +299,14 @@ export function CopilotChat({
           raf(() => {
             if (!detached) setLastConnectedThreadId(resolvedThreadId);
           });
+        } else if (!hasExplicitThreadIdRef.current) {
+          // This connect was superseded (the user switched away while it was
+          // still loading). If the now-current thread is a fresh non-explicit
+          // one (e.g. the drawer's "+ New"), any snapshot this connect managed
+          // to apply is stale — clear it so the welcome screen shows instead of
+          // the abandoned thread's messages. A switch to ANOTHER explicit thread
+          // is left alone: that thread's own connect owns the message reset.
+          agentToConnect.setMessages([]);
         }
       }
     };
--- a/packages/react-core/src/v2/components/chat/CopilotDrawer.tsx
+++ b/packages/react-core/src/v2/components/chat/CopilotDrawer.tsx
@@ -193,11 +193,20 @@ export function CopilotDrawer({
   // license is configured, so it cannot by itself detect the no-license case.
   // We therefore also require a positive license-present signal from the
   // runtime-reported status. Only a "valid" or "expiring" license is treated
-  // as present; null/"none"/"unknown" (no/indeterminate license) and
-  // "expired"/"invalid" all gate the drawer to the upsell.
+  // as present; a resolved "none"/"expired"/"invalid" status gates the drawer
+  // to the upsell.
   const licensePresent = status === "valid" || status === "expiring";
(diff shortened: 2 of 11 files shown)


MikeRyanDev (changes requested):
Found one blocking issue in the clipped-name tooltip change.

In `packages/web-components/src/drawer/styles.ts`, the stacking fix is written as `.row.name-clipped:hover { z-index: 2; }`, but `_syncNameClipping()` toggles `name-clipped` on the `.row-name` element, not on the surrounding `li.row`. That selector never matches, so the intended z-lift does not run. Because each row has its own transform-created stacking context, clipped-name tooltips can still paint under later rows, which is the exact visual issue this PR says it fixes.

Please apply the clipped state to the row (or otherwise make the z-index rule target the actual row stacking context) and add coverage/assertion for that class/selector contract.

BenTaylorDev:
Fixed in edbf98034 — thanks @MikeRyanDev, good catch.

`_syncNameClipping()` was toggling `name-clipped` on `.row-name`, while the stacking rule in `styles.ts` is `.row.name-clipped:hover { z-index: 2 }` — a different element, so it never matched and the clipped-name tooltip stayed trapped under later rows (each `.row` is its own `transform` stacking context).

Fix: also stamp the flag on the owning `.row` via `name.closest('.row')`, so the z-lift rule matches. The tooltip bubble itself stays scoped to `.row-name.name-clipped:hover` so hovering a row-action never surfaces it.

Added a red-green contract test (`a clipped name marks BOTH the row-name and the owning row…`) that simulates a truncated text node and asserts `name-clipped` lands on **both** `.row-name` (tooltip) and `.row` (z-index) — and clears on both when the name fits. Verified it fails on the row assertion with the fix reverted, passes with it. Full web-components suite 57/57 green.

BenTaylorDev:
_(Note: the branch was force-rebased onto latest main between my comment and push, so the fix landed as `9223fc476` rather than `edbf98034` — same change, cherry-picked onto the updated tip.)_

MikeRyanDev (approved):
Re-reviewed after the clipped-name tooltip fix. The previous blocker is addressed: `name-clipped` is now stamped on both `.row-name` and the owning `.row`, and the contract test covers set/clear behavior.

Verified locally with:
- `pnpm exec nx run @copilotkit/web-components:test` (57 tests passed)
- `pnpm exec nx run @copilotkit/react-core:test` (107 files / 1375 Vitest tests passed, plus 2 node test script tests passed)

PR #5728, opened 2026-06-26
feat(examples): enrich AIMock fixtures so keyless demo suggestions return scripted replies

BenTaylorDev:
## Why

The 6 OpenAI drop-in integration examples enabled for the CLI's **keyless AIMock mock mode** (`langgraph-python`, `langgraph-js`, `mastra`, `llamaindex`, `agno`, `pydantic-ai`) shipped `fixtures/default.json` files that **did not cover the prompts each starter's own UI suggests**. AIMock matches `userMessage` as a **case-sensitive substring, first-match-wins**, so a keyless first-run user who clicked the demo's suggestion chips mostly fell through to the generic catch-all ("I only have scripted replies…") instead of getting a scripted demo reply.

Two root issues found (audit at `origin/main`):
- **`langgraph-python` was mis-keyed** — fixtures keyed on suggestion *titles* (`"Pie Chart"`, `"Toggle Theme"`, `"Task Manager"`) while the UI sends long *message* strings that don't contain those substrings → all 9 suggestions missed.
- The other 5 shipped only a generic `Hello` (+ a `weather` fixture in langgraph-js/mastra) → 0–1 of each starter's chips covered.

This PR re-keys/extends each example's fixtures so the surfaced suggestions return scripted replies. Ordering preserved: most-specific-first, `{}` catch-all last (unchanged text).

> Tracking: **ENT-1003** (CopilotKit/Intelligence). Sibling to ENT-989 (fixtures for *not-yet-enabled* frameworks). This PR covers the *already-enabled* 6.

## What changed (per template)

| Template | Suggestions now covered | Tool-call replies | Text replies |
|---|---|---|---|
| langgraph-python | 9/9 (3 re-keyed) | pie/bar chart (`query_data`→component), `scheduleTime`, `toggleTheme`, `manage_todos` | Search Flights, Excalidraw, Calculator, Sales-Dashboard A2UI step |
| langgraph-js | 9/9 | `query_data`, `search_flights`, `generate_a2ui`, `manage_todos` (schemas verified vs agent) | scheduleTime, Excalidraw, generateSandboxedUi, toggleTheme |
| mastra | 6/6 | `get-weather`, `setThemeColor`, `go_to_moon` (HITL) | 3 proverb chips (proverbs are agent shared-state, not a tool) |
| llamaindex | 3/3 | — | theme / proverb / weather |
| agno | 4/4 | — | weather / theme / stock / proverb |
| pydantic-ai | n/a (UI surfaces no chips) | — | best-effort free-typer fixtures: "what can you do" / "proverb" / "weather" |

## Honest caveats (best-effort; draft)

- **Tool-call shapes only where evidenced.** Where a suggestion drives A2UI streaming, an MCP app (Excalidraw), or a frontend-only tool (`generateSandboxedUi`, `scheduleTime` in some templates) whose call shape isn't defined in the template, I used an **on-topic text reply** rather than fabricating a tool envelope. Those replies beat the catch-all but won't trigger the live generative UI under mock mode — a follow-up could record the real shapes.
- **`pydantic-ai` surfaces no suggestion chips** in its UI, so there's nothing to key on; the real fix is a small UI change (add `suggestions` to `CopilotSidebar`), which is out of scope for a fixtures-only PR. The added fixtures are a fallback for free-typing users.
- **Not verified end-to-end here** (authored against `origin/main`, not run live). Each template's `docker-compose.test.yml` AIMock smoke should stay green; note its `@chat` test only sends `"Hello"` and asserts a non-empty reply, so it does **not** validate suggestion-chip coverage — extending it to assert a non-catch-all reply for a real suggestion would close that blind spot (also noted in ENT-1003).

## Test plan
- [ ] Per-template `docker-compose.test.yml` AIMock smoke still green.
- [ ] Manual: scaffold/run each keyless, click each suggestion chip, confirm a scripted reply (not the catch-all).

### DIFF
--- a/examples/integrations/agno/fixtures/default.json
+++ b/examples/integrations/agno/fixtures/default.json
@@ -7,6 +7,30 @@
       }
     },
     {
+      "match": { "userMessage": "weather in San Francisco" },
+      "response": {
+        "content": "The weather in San Francisco is: 70 degrees and Sunny."
+      }
+    },
+    {
+      "match": { "userMessage": "theme to green" },
+      "response": {
+        "content": "Done! I've changed the theme color to green (#22c55e). The page background and accent colors should now reflect the new theme."
+      }
+    },
+    {
+      "match": { "userMessage": "price of Apple stock" },
+      "response": {
+        "content": "Here is the latest price information for Apple Inc. (AAPL):\n\n| Symbol | Price   | Change  | Change % |\n|--------|---------|---------|----------|\n| AAPL   | $213.49 | +1.24   | +0.58%   |\n\n*Note: you are running against aimock — this is a scripted fixture value, not a live quote.*"
+      }
+    },
+    {
+      "match": { "userMessage": "proverb about AI" },
+      "response": {
+        "content": "Added a new proverb: *\"An AI that learns from its mistakes is wiser than a human who repeats them.\"*\n\nYou should see it appear in the Proverbs panel on the left."
+      }
+    },
+    {
       "match": {},
       "response": {
         "content": "You're currently running against aimock (a mock LLM server). This response is a catch-all for requests that don't match any test fixture. To use a real LLM: (1) Add your OPENAI_API_KEY to .env, (2) Remove or unset OPENAI_BASE_URL from your environment so requests go to OpenAI instead of aimock, (3) Restart with `pnpm dev`."
--- a/examples/integrations/langgraph-js/fixtures/default.json
+++ b/examples/integrations/langgraph-js/fixtures/default.json
@@ -123,7 +123,9 @@
     {
       "match": { "userMessage": "Toggle the app theme" },
       "response": {
-        "content": "I've toggled the app theme for you using the toggleTheme tool. The interface should now be in the alternate color scheme -- let me know if you'd like to switch back."
+        "toolCalls": [
+          { "name": "toggleTheme", "arguments": "{}", "id": "call_toggle_theme_001" }
+        ]
       }
     },
     {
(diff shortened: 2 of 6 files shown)


marthakelly:
Toggle Theme doesn't work in langgraph-js mock mode (langgraph-js/fixtures/default.json:124).

Behavioral bug: identical chip click, different outcome depending on template. The fix is one JSON object swap (text response → tool call). This is the only finding I'd say is worth blocking on, or at least addressing before the PR lands if toggle-theme is a showcase feature for langgraph-js.

marthakelly (dismissed):

BenTaylorDev:
Fixed in `bc24d62ce`. Confirmed the bug — the chip rendered the "I've toggled the theme" text path while emitting no tool call, so nothing toggled.

- **toggle-theme** (langgraph-js): swapped the text response for the `toggleTheme` tool call, mirroring langgraph-python (frontend tool, fire-and-forget — no terminating fixture needed). Verified aimock now returns the `toggleTheme` tool call for the chip.
- **scheduleTime** (same class, caught while in there): the langgraph-js "schedule a meeting" chip had the identical text-instead-of-tool-call divergence, and `scheduleTime` is a registered langgraph-js frontend tool (`reasonForScheduling`/`meetingDuration`). Swapped it to the tool call too. Because `scheduleTime` is Human-in-the-Loop (the picked time returns as a tool result and re-invokes the LLM), I also added a terminating `call_schedule_time_001` result fixture in **both** langgraph-js and langgraph-python (python lacked it → latent loop) so the result turn doesn't re-match the user message and re-call forever. Verified both terminate at 2 steps (tool call → terminating text).

Excalidraw / calculator chips stay text on purpose — those aren't registered tools in langgraph-js (MCP / `generateSandboxedUi`), so there's no tool call to emit.

Issue #175, opened 2024-01-23
Installing CopilotKit into a fresh app (via create-next-app) leads to IDE errors


mme:
**Describe the bug**
When installing CopilotKit into a fresh app created via create-next-app, VSCode displays errors in the imports.

This is the exact error:

```
Could not find a declaration file for module '@copilotkit/react-core'. '/private/tmp/my-app/node_modules/@copilotkit/react-core/dist/index.mjs' implicitly has an 'any' type.
  There are types at '/private/tmp/my-app/node_modules/@copilotkit/react-core/dist/index.d.ts', but this result could not be resolved when respecting package.json "exports". The '@copilotkit/react-core' library may need to update its package.json or typings.ts(7016)
```

**To Reproduce**
1. `yarn create next-app my-app --ts --eslint --no-tailwind --src-dir --app --import-alias="@/*" # create new app`
2. `cd my-app`
3. `yarn add @copilotkit/react-core @copilotkit/react-ui`
4. In page.tsx, add: `import { CopilotProvider } from "@copilotkit/react-core";`
5. The line appears red in VSCode

**Expected behavior**
No errors in VSCode, should just work.

**Screenshots**
<img width="574" alt="image" src="https://github.com/CopilotKit/CopilotKit/assets/247775/57166c15-52ab-41e4-9aa3-2dae37940f8b">

** IDE:**
 - VSCode

PR #5729, opened 2026-06-26
feat(bot): anonymous oss.bot.* usage telemetry (adoption funnel)

BenTaylorDev:
## What

Adds anonymous **`oss.bot.*`** usage telemetry to the `@copilotkit/bot` SDK — a **configured → started → agent_run** adoption funnel that answers "what are people doing with the bot packages?" (which adapters, which store backends, how they iterate on config, where they fail) — without knowing *who* the customer is.

Design spec: [Bot SDK Telemetry — oss.bot.* Adoption Funnel](https://app.notion.com/p/38b3aa3818528177a3abfda9ce35bdbc) · Plan: [Implementation Plan](https://app.notion.com/p/38b3aa38185281d5a09cfea114df8897)

## Events (all 100%, anonymous, metadata-only)

| Event | When | Notable props |
|-------|------|---------------|
| `oss.bot.configured` | `createBot()` returns (repeats per restart → iteration signal) | `platforms`, `store`, `toolsCount`, `hasComponents`, … |
| `oss.bot.started` | `bot.start()` connected ≥1 adapter | `platforms`, `startedCount`, `failedCount`, handler flags |
| `oss.bot.start_failed` | an adapter threw during start | `platform`, `errorClass` |
| `oss.bot.agent_run` | a successful agent run | `platform`, `durationMs`, `toolCallCount`, `iterations`, `interrupted` |
| `oss.bot.agent_run_failed` | an agent run errored | `platform`, `errorClass`, `stage` |

Catalog + property reference: `packages/bot/telemetry-events.json`.

## Design

- **Anonymous.** Bot deployments carry no `telemetry_id`/license/API key, so events are stitched by a persisted `anonymous_id` (durable `StateStore` → project-local cache file → per-process UUID) plus a per-`createBot` `bot_session_id`. Reuses `lambdaClient.send` from `@copilotkit/shared` (no new deps).
- **Zero-config.** Works out of the box; **no new env vars**. Reads only the pre-existing optional `COPILOTKIT_TELEMETRY_URL` (override) and `COPILOTKIT_TELEMETRY_DISABLED` / `DO_NOT_TRACK` (opt-out), and `NODE_ENV`/`VITEST` for the `environment` tag + test suppression.
- **Fire-and-forget.** `capture()` never throws into or blocks the host app; all dispatch/fetch/fs failures are swallowed.
- **PII guardrails.** Snapshots are scalars/counts only (never adapter/store option objects → no token/connection-string leakage); errors are mapped to a bounded `errorClass` category (`auth`/`network`/`timeout`/`validation`/`unknown`) — never a raw message or stack; no message text, user ids, or channel names anywhere.

**Sink side** (`oss.bot.` allow-list prefix + `anonymous_id` → PostHog `distinct_id`) shipped separately in oss-path-to-production PR #175 (already merged).

## Testing

**Unit (TDD, red→green per module):**
- `sanitize-error.test.ts` — category mapping + a "never leaks the message" secret-redaction assertion (2)
- `install-id.test.ts` — durable-store persistence, file persistence, unwritable-dir fallback (3)
- `bot-telemetry.test.ts` — global-props shape, disabled no-op, never-throws-on-send-reject, event-name set (4)
- `events-catalog.test.ts` — drift guard: catalog keys == emitted event names (1)
- `run-loop.test.ts` — new `{iterations, interrupted}` return value (interrupt + normal paths)
- `create-bot-telemetry.test.ts` — mocked-telemetry wiring: configured snapshot, started/start_failed (with `xoxb-SECRET` redaction assertion), agent_run

**End-to-end:**
- `e2e-telemetry.test.ts` — drives the **real `BotTelemetry`** through `createBot → start → runAgent` (only the network boundary is stubbed), asserts the `configured → started → agent_run` payloads carry `anonymous_id`/`bot_session_id` and **runs with an empty telemetry env** (proves zero-config; asserts no license token attached).

**Full suite (authoritative, on a real `pnpm install` in the integration worktree):**
- `tsc --noEmit` → 0 errors · `@copilotkit/bot` vitest → **26 files / 141 tests pass**

**In-session manual smoke (real wire bytes):** stubbed `globalThis.fetch`, drove a real bot through one turn, captured the actual serialized POSTs:
```
sink URL (no env var set → default): https://telemetry.copilotkit.ai/ingest
oss.bot.configured  {"properties":{"platforms":["fake"],"adapterCount":1,"store":"memory",...},
                     "global_properties":{"anonymous_id":"74b3c598-…","bot_session_id":"7cddd757-…","environment":"production"},
                     "package":{"name":"@copilotkit/bot","version":"0.0.3"},"ts":1782502088}
oss.bot.started     {"properties":{"platforms":["fake"],"startedCount":1,"failedCount":0,"hasMentionHandler":true,...}, "global_properties":{ same anonymous_id + bot_session_id }}
oss.bot.agent_run   {"properties":{"platform":"fake","durationMs":0,"toolCallCount":0,"iterations":1,"interrupted":false}, "global_properties":{ same anonymous_id + bot_session_id }}
```
All three share one `anonymous_id` + `bot_session_id` (funnel stitching), default sink URL with no env var set, no license/key anywhere.

**No new env vars (verified):** `git diff origin/main -- packages/bot/src/create-bot.ts packages/bot/src/thread.ts | grep process.env` → none; wiring code reads no env var directly.

**Code review:** one focused reviewer pass on the full diff — no blocker/high findings; PII, fire-and-forget, zero-config, and no-regression axes all confirmed clean. Two low/nit findings applied (dropped `TypeError`→`validation` mis-categorization; documented the interrupt→resume `agent_run` double-count in the catalog).

### DIFF
--- a/packages/bot/src/create-bot.ts
+++ b/packages/bot/src/create-bot.ts
@@ -40,6 +40,22 @@ import {
 import { Transcripts } from "./transcripts.js";
 import type { Identity, TranscriptsConfig } from "./transcripts.js";
 import type { StandardSchemaV1, InferSchemaOutput } from "./standard-schema.js";
+import { BotTelemetry } from "./telemetry/bot-telemetry.js";
+import { errorClass } from "./telemetry/sanitize-error.js";
+import { createRequire } from "node:module";
+
+const pkg = createRequire(import.meta.url)("../package.json") as {
+  name: string;
+  version: string;
+};
+
+function storeKind(s: StateStore): "memory" | "postgres" | "redis" | "custom" {
+  const n = s.constructor?.name;
+  if (n === "MemoryStore") return "memory";
+  if (n === "PostgresStore") return "postgres";
+  if (n === "RedisStore") return "redis";
+  return "custom";
+}
 
 /** Platforms whose tokens the emoji table can normalize. */
 const EMOJI_PLATFORMS: ReadonlySet<EmojiPlatform> = new Set([
@@ -249,6 +265,11 @@ export function createBot<
   }
 
   const backend: StateStore = cfg.adapter ?? new MemoryStore();
+  const telemetry = new BotTelemetry({
+    backend,
+    packageName: pkg.name,
+    packageVersion: pkg.version,
+  });
   const transcripts = new Transcripts(backend, cfg.transcripts ?? {});
   const registry = new ActionRegistry({
     store: opts.actionStore ?? kvActionStore(backend),
@@ -327,6 +348,7 @@ export function createBot<
       transcripts,
       userKey: extras?.userKey,
       message: extras?.message,
+      telemetry,
     };
     return new Thread(deps);
   }
@@ -710,14 +732,36 @@ export function createBot<
       const startResults = await Promise.allSettled(
         opts.adapters.map((a) => a.start(makeSink(a))),
       );
+      const startedPlatforms: string[] = [];
+      const failedPlatforms: string[] = [];
       startResults.forEach((r, i) => {
+        const platform = opts.adapters[i]!.platform;
         if (r.status === "rejected") {
+          failedPlatforms.push(platform);
           console.error(
-            `[bot] adapter "${opts.adapters[i]!.platform}" failed to start:`,
+            `[bot] adapter "${platform}" failed to start:`,
             r.reason,
           );
+          telemetry.capture("oss.bot.start_failed", {
+            platform,
+            errorClass: errorClass(r.reason),
+          });
+        } else {
+          startedPlatforms.push(platform);
         }
       });
+      if (startedPlatforms.length > 0) {
+        telemetry.capture("oss.bot.started", {
+          platforms: startedPlatforms,
+          startedCount: startedPlatforms.length,
+          failedCount: failedPlatforms.length,
+          hasMentionHandler: mentionHandlers.length > 0,
+          hasMessageHandler: messageHandlers.length > 0,
+          interruptHandlers: interruptHandlers.size,
+          commandsCount: commandHandlers.size,
+          toolsCount: toolMap.size,
+        });
+      }
       // Hand declared commands to adapters that register them up front (e.g.
       // Discord); adapters without `registerCommands` are skipped. Per-adapter
       // failures are isolated the same way as start().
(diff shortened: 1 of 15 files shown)


packages/bot/src/create-bot.ts
```diff
@@ -752,5 +796,17 @@ export function createBot<
       });
     },
   };
+  telemetry.capture("oss.bot.configured", {
+    platforms: opts.adapters.map((a) => a.platform),
```
tylerslaton: `platform` is a free-form adapter string, so sending it raw can leak custom tenant/project names and create high-cardinality telemetry. Can we normalize known platforms and bucket unknowns as `custom` before emitting?

packages/bot/src/thread.ts
```diff
@@ -394,21 +402,41 @@ export class Thread implements ThreadInterface {
...
-        const h = this.deps.interruptHandlers.get(interrupt.eventName);
-        if (h) await h({ payload: interrupt.value, thread: this });
-      },
-      initialResume,
+        errorClass: errorClass(err),
+        stage: "agent",
+      });
+      throw err;
+    }
+    this.deps.telemetry?.capture("oss.bot.agent_run", {
```
tylerslaton, line 471: This records a successful `agent_run` before transcript append and `renderer.finish?.()` run below. If either later await rejects, callers see a failed run after success telemetry was already sent. Can we move this after finalization, or capture those later failures with a separate stage?

tylerslaton (commented):
Left two small inline concerns around telemetry correctness/privacy.

packages/bot/src/thread.ts
BenTaylorDev: Fixed in 4d67b2bef. `oss.bot.agent_run` now fires only after the run loop AND the transcript-append + `renderer.finish?.()` steps all succeed — those steps are inside the try now. A throw in finalization emits `oss.bot.agent_run_failed` with `stage: "finalize"` (vs `stage: "agent"` for run-loop failures) instead of a success event being followed by a failure. Catalog updated to document the two stages.

packages/bot/src/create-bot.ts
BenTaylorDev: Good call — fixed in 4d67b2bef. Added `normalizePlatform()` (in `telemetry/sanitize-error.ts`): known platforms (`slack`/`discord`/`telegram`/`whatsapp`) pass through, everything else buckets to `custom`, so a custom adapter's free-form label cannot leak a tenant/project name and cardinality stays bounded. Applied to every platform emit site (configured/started/start_failed/agent_run/agent_run_failed); the human-facing `console.error` on start failure still logs the raw label for debugging. Catalog + a `normalizePlatform` unit test added.

BenTaylorDev: Follow-up: added a monorepo-invariant drift guard (`platform-allowlist-coverage.test.ts`) that scans the sibling `bot-*` adapter packages for their declared `readonly platform` literal and fails if any isn't in the allow-list — so adding a new adapter without updating `normalizePlatform` is caught in CI rather than silently bucketing to `custom`. It immediately surfaced the new `bot-teams` adapter (`platform: "teams"`) that landed on `main` during the rebase; `teams` is now in the allow-list.

tylerslaton (approved):

PR #5741, opened 2026-06-27
feat: sync Intelligence thread runs across devices

maxkorp:
## Summary
- add Intelligence thread run-activity metadata normalization and subscription support
- reconnect explicit native Intelligence chats on remote run activity without keeping /connect or /run streams long-lived
- harden replay cursor handling, passive /connect completion, and passive replay tool execution

## Verification
- `pnpm --dir packages/core exec vitest run src/__tests__/threads.test.ts src/__tests__/intelligence-agent.test.ts src/__tests__/core-run-agent-resume.test.ts src/__tests__/core-connect-thread-switch.test.ts src/__tests__/core-connect-passive-replay.test.ts src/__tests__/core-follow-up.test.ts`
- `pnpm --dir packages/react-core exec vitest run src/v2/components/chat/__tests__/CopilotChat.runActivityReconnect.test.tsx src/v2/components/chat/__tests__/CopilotChat.absentThreadConnect.test.tsx src/v2/components/chat/__tests__/copilot-chat-throttle.test.tsx src/v2/hooks/__tests__/use-human-in-the-loop.e2e.test.tsx`
- `pnpm --dir packages/core run check-types`
- `pnpm --dir packages/react-core run check-types`
- `pnpm --filter @copilotkit/core run build`
- `pnpm --filter @copilotkit/react-core run build`
- pre-commit package checks passed for affected package set

## Manual testing needed before ready
- two browser tabs/devices on the same explicit Intelligence thread: tab A sends a later run, tab B live-syncs through wake-triggered /connect replay
- old/new compatibility smoke tests: old client with new gateway, new client with old gateway

Draft until manual testing is complete.

### DIFF
--- a/packages/core/src/core/run-handler.ts
+++ b/packages/core/src/core/run-handler.ts
@@ -18,6 +18,11 @@ export interface CopilotKitCoreRunAgentParams {
   agent: AbstractAgent;
   forwardedProps?: Record<string, unknown>;
   /**
+   * Optional caller-supplied run identifier forwarded to the underlying AG-UI
+   * agent. When omitted, the agent creates one.
+   */
+  runId?: string;
+  /**
    * Per-interrupt responses addressing every open AG-UI interrupt from the
    * previous run. Forwarded to the agent as the standard `resume` array.
    */
@@ -303,6 +308,7 @@ export class RunHandler {
     agent,
     forwardedProps,
     resume,
+    runId,
   }: CopilotKitCoreRunAgentParams): Promise<RunAgentResult> {
     // Agent ID is guaranteed to be set by validateAndAssignAgentId
     if (agent.agentId) {
@@ -375,6 +381,7 @@ export class RunHandler {
             ...forwardedProps,
           },
           ...(resume !== undefined ? { resume } : {}),
+          ...(runId !== undefined ? { runId } : {}),
           tools: this.buildFrontendTools(agent.agentId),
           context: this._internal.getContextForAgent(agent.agentId),
         },
--- a/packages/core/src/core/run-handler.ts
+++ b/packages/core/src/core/run-handler.ts
@@ -307,7 +307,11 @@ export class RunHandler {
         this.createAgentErrorSubscriber(agent),
       );
 
-      return this.processAgentResult({ runAgentResult, agent });
+      return this.processAgentResult({
+        runAgentResult,
+        agent,
+        executeFrontendTools: false,
+      });
     } catch (error) {
       const connectError =
         error instanceof Error ? error : new Error(String(error));
@@ -448,9 +452,11 @@ export class RunHandler {
   private async processAgentResult({
     runAgentResult,
     agent,
+    executeFrontendTools = true,
   }: {
     runAgentResult: RunAgentResult;
     agent: AbstractAgent;
+    executeFrontendTools?: boolean;
   }): Promise<RunAgentResult> {
     const { newMessages } = runAgentResult;
     // Agent ID is guaranteed to be set by validateAndAssignAgentId
@@ -458,38 +464,22 @@ export class RunHandler {
 
     let needsFollowUp = false;
 
-    for (const message of newMessages) {
-      if (message.role === "assistant") {
-        for (const toolCall of message.toolCalls || []) {
-          if (
-            newMessages.findIndex(
-              (m) => m.role === "tool" && m.toolCallId === toolCall.id,
-            ) === -1
-          ) {
-            const tool = this.getTool({
-              toolName: toolCall.function.name,
-              agentId: agent.agentId,
-            });
-            if (tool) {
-              const followUp = await this.executeSpecificTool(
-                tool,
-                toolCall,
-                message,
-                agent,
-                agentId,
-              );
-              if (followUp) {
-                needsFollowUp = true;
-              }
-            } else {
-              // Wildcard fallback for undefined tools
-              const wildcardTool = this.getTool({
-                toolName: "*",
+    if (executeFrontendTools) {
+      for (const message of newMessages) {
+        if (message.role === "assistant") {
+          for (const toolCall of message.toolCalls || []) {
+            if (
+              newMessages.findIndex(
+                (m) => m.role === "tool" && m.toolCallId === toolCall.id,
+              ) === -1
+            ) {
+              const tool = this.getTool({
+                toolName: toolCall.function.name,
                 agentId: agent.agentId,
               });
-              if (wildcardTool) {
-                const followUp = await this.executeWildcardTool(
-                  wildcardTool,
+              if (tool) {
+                const followUp = await this.executeSpecificTool(
+                  tool,
                   toolCall,
                   message,
                   agent,
@@ -497,6 +487,24 @@ export class RunHandler {
                 );
                 if (followUp) {
                   needsFollowUp = true;
+                }
+              } else {
+                // Wildcard fallback for undefined tools
+                const wildcardTool = this.getTool({
+                  toolName: "*",
+                  agentId: agent.agentId,
+                });
+                if (wildcardTool) {
+                  const followUp = await this.executeWildcardTool(
+                    wildcardTool,
+                    toolCall,
+                    message,
+                    agent,
+                    agentId,
+                  );
+                  if (followUp) {
+                    needsFollowUp = true;
+                  }
                 }
               }
             }
--- a/packages/core/src/core/run-handler.ts
+++ b/packages/core/src/core/run-handler.ts
@@ -267,12 +267,17 @@ export class RunHandler {
       if (isFreshRestore) {
         agent.setMessages([]);
         agent.setState({});
-        const proxied = agent as { clearReplayCursor?: (id: string) => void };
-        if (
-          incomingThreadId &&
-          typeof proxied.clearReplayCursor === "function"
-        ) {
-          proxied.clearReplayCursor(incomingThreadId);
+        const cursorAware = agent as {
+          clearReconnectCursor?: (id: string) => void;
+          clearReplayCursor?: (id: string) => void;
+        };
+        if (incomingThreadId) {
+          if (typeof cursorAware.clearReplayCursor === "function") {
+            cursorAware.clearReplayCursor(incomingThreadId);
+          }
+          if (typeof cursorAware.clearReconnectCursor === "function") {
+            cursorAware.clearReconnectCursor(incomingThreadId);
+          }
         }
       }
 
--- a/packages/core/src/core/run-handler.ts
+++ b/packages/core/src/core/run-handler.ts
@@ -117,7 +117,9 @@ export class RunHandler {
    * downstream churn into duplicate `cpki_event_id` rows in the
    * inspector and intermittent "Message not found" toasts.
    */
-  private _lastConnectedThreadId: string | null = null;
+  private _lastConnectedThreadIdsByAgent = new Map<string, string | null>();
+  private _anonymousAgentIds = new WeakMap<AbstractAgent, string>();
+  private _nextAnonymousAgentId = 0;
 
   constructor(private core: CopilotKitCore) {}
 
@@ -136,6 +138,27 @@ export class RunHandler {
    */
   private get _internal(): CopilotKitCoreFriendsAccess {
     return this.core as unknown as CopilotKitCoreFriendsAccess;
+  }
+
+  /**
+   * Return a stable restore-tracking key for the logical agent being
+   * connected. Named agents share their last-thread marker across proxy
+   * instances; anonymous agents fall back to object identity.
+   */
+  private getConnectRestoreKey(agent: AbstractAgent): string {
+    if (agent.agentId) {
+      return `agent:${agent.agentId}`;
+    }
+
+    const existing = this._anonymousAgentIds.get(agent);
+    if (existing) {
+      return existing;
+    }
+
+    const anonymousId = `anonymous:${this._nextAnonymousAgentId}`;
+    this._nextAnonymousAgentId += 1;
+    this._anonymousAgentIds.set(agent, anonymousId);
+    return anonymousId;
   }
 
   /**
@@ -224,8 +247,11 @@ export class RunHandler {
   }: CopilotKitCoreConnectAgentParams): Promise<RunAgentResult> {
     try {
       const incomingThreadId = agent.threadId ?? null;
-      const isFreshRestore = incomingThreadId !== this._lastConnectedThreadId;
-      this._lastConnectedThreadId = incomingThreadId;
+      const restoreKey = this.getConnectRestoreKey(agent);
+      const isFreshRestore =
+        incomingThreadId !==
+        (this._lastConnectedThreadIdsByAgent.get(restoreKey) ?? null);
+      this._lastConnectedThreadIdsByAgent.set(restoreKey, incomingThreadId);
 
       // Detach any active run before connecting to avoid previous runs
       // interfering. This stays unconditional — both fresh restores and
--- a/packages/core/src/intelligence-agent.ts
+++ b/packages/core/src/intelligence-agent.ts
@@ -17,6 +17,7 @@ import {
   EMPTY,
   Subject,
   Notification,
+  combineLatest,
   defer,
   dematerialize,
   lastValueFrom,
@@ -614,14 +615,19 @@ export class IntelligenceAgent extends AbstractAgent {
         input.threadId,
         channel$,
         REPLAY_COMPLETE_EVENT,
-      ).pipe(ignoreElements(), share());
+      ).pipe(shareReplay({ bufferSize: 1, refCount: true }));
       const streamIdle$ = this.observeControlEvent$(
         input.threadId,
         channel$,
         STREAM_IDLE_EVENT,
       ).pipe(shareReplay({ bufferSize: 1, refCount: true }));
       const streamIdleCompletion$ =
-        options.streamMode === "connect" ? streamIdle$.pipe(take(1)) : EMPTY;
+        options.streamMode === "connect"
+          ? combineLatest([
+              replayComplete$.pipe(take(1)),
+              streamIdle$.pipe(take(1)),
+            ]).pipe(take(1))
+          : EMPTY;
       const threadCompleted$ = threadEvents$.pipe(
         ignoreElements(),
         endWith(null),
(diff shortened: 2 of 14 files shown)


maxkorp:
Merge order:

1. Merge/release this CopilotKit PR first.
2. Then merge CopilotKit/Intelligence#458 after Intelligence can consume a CopilotKit package version containing these client changes.

The rollout is additive under version skew, but full multi-device live-sync validation requires both sides.

packages/core/src/intelligence-agent.ts
```diff
@@ -599,14 +615,19 @@ export class IntelligenceAgent extends AbstractAgent {
...
+      ).pipe(shareReplay({ bufferSize: 1, refCount: true }));
       const streamIdle$ = this.observeControlEvent$(
         input.threadId,
         channel$,
         STREAM_IDLE_EVENT,
       ).pipe(shareReplay({ bufferSize: 1, refCount: true }));
       const streamIdleCompletion$ =
-        options.streamMode === "connect" ? streamIdle$.pipe(take(1)) : EMPTY;
+        options.streamMode === "connect"
+          ? combineLatest([
```
BenTaylorDev: **[HIGH / blocker] connect hangs forever if the gateway emits `stream_idle` without `replay_complete`.**

Completion in connect mode changed to require BOTH control events via `combineLatest([replayComplete$.take(1), streamIdle$.take(1)])`. This is the *only* terminal path: in connect mode `completeOnRunFinished` is false, so `threadEvents$` never completes and `threadCompleted$`'s `endWith(null)` never fires — and there's no timeout on `streamIdleCompletion$`.

So if `replay_complete` never arrives (an idle thread with nothing to replay, or an older/self-hosted gateway build), `connectAgent()`'s promise never resolves, the socket + channel are never torn down, and in `CopilotChat` `activeConnectCountRef` never decrements — permanently blocking the reconnect drain. The 6+ test edits that inject `replay_complete` before every `stream_idle` are the tell that the new code strictly requires it.

Either guarantee (gateway contract) that `replay_complete` always precedes `stream_idle` on connect — including empty threads — or add a fallback/timeout terminal.

packages/core/src/core/run-handler.ts
```diff
@@ -258,7 +294,11 @@ export class RunHandler {
         this.createAgentErrorSubscriber(agent),
       );
 
-      return this.processAgentResult({ runAgentResult, agent });
+      return this.processAgentResult({
+        runAgentResult,
+        agent,
+        executeFrontendTools: false,
```
BenTaylorDev, line 313: **[MED — product sign-off] passing `executeFrontendTools: false` drops the HITL respond-after-reload affordance.**

Passive replay no longer re-invokes frontend tool handlers, so a replayed pending HITL assistant tool call can't be completed from restored state. The `use-human-in-the-loop.e2e.test.tsx` assertion was correspondingly weakened (`Executing` → `/^(executing|inProgress)$/`, and the `expect(getByTestId("hitl-respond")).toBeDefined()` line deleted).

Not re-executing side-effecting tools during replay is the right call (see the new `core-connect-passive-replay.test.ts`), but losing the ability to finish an interrupted approval after a reload is a real UX change — worth an explicit product decision rather than a silently relaxed test.

packages/core/src/intelligence-agent.ts
```diff
@@ -783,6 +807,10 @@ export class IntelligenceAgent extends AbstractAgent {
       return;
     }
 
+    this.advanceLastSeenEventId(threadId, eventId);
+  }
+
+  private advanceLastSeenEventId(threadId: string, eventId: string): void {
```
BenTaylorDev, line 839: **[MED — gateway-contract dependent] `advanceLastSeenEventId` blindly overwrites, mixing two id-namespaces into one cursor.**

The per-thread cursor is written both from `ag_ui_event` metadata (`cpki_event_id` runner id, else `cpki_ingested`) and from control events (`latestEventId`/`latest_event_id` ingestion cursor), via an unconditional `Map.set` — despite the "advance" name (it isn't monotonic).

If a control event's payload lacks `latest_event_id` (→ `readControlEventId` returns null → no write) while the preceding events carried a `cpki_event_id`, the value sent as `last_seen_event_id` on the next connect is a runner id, not the ingestion cursor the gateway expects → mismatched resume (full replay, skipped range, or parse error). The design relies on control events always arriving last AND always carrying `latest_event_id`; neither is enforced here. A namespace-aware / monotonic set would remove the fragility.

packages/react-core/src/v2/components/chat/CopilotChat.tsx
```diff
@@ -327,6 +433,179 @@ export function CopilotChat({
...
+        if (hasActiveAgentRun()) {
+          scheduleAgentIdleDrain();
+          return;
+        }
+        startRunActivityReconnectRef.current?.(generation);
+      }, 10);
+    };
+
+    const connect = async () => {
+      activeConnectCountRef.current += 1;
```
BenTaylorDev, line 483: **[LOW–MED] the bootstrap connect effect and the wake reconnect share one `activeConnectCountRef`.**

Both this wake `connect()` and the bootstrap connect effect (line ~370) increment/decrement the same counter and both gate their pending-reconnect drain on `activeConnectCountRef.current === 0`. When a thread/agent re-render fires the bootstrap connect while a wake connect is in flight, `=== 0` no longer means "no wake connect running," so a queued reconnect can be dropped (neither `finally` re-arms it) or double-fire (both drain). The reconnect tests mostly pass, so it's usually guarded — but the counter should distinguish wake connects from bootstrap connects.

packages/react-core/src/v2/components/chat/CopilotChat.tsx
```diff
@@ -327,6 +433,179 @@ export function CopilotChat({
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [resolvedThreadId, agent, resolvedAgentId, hasExplicitThreadId]);
 
+  useEffect(() => {
+    if (!hasNativeIntelligenceRunActivity) return;
```
BenTaylorDev, line 437: **[ALTITUDE] this ~200-line wake-reconnect machine belongs in core, beside `runId`.**

Nine refs (`activeLocalRunIdsRef`, `recentlyLocalRunIdsRef`, `activeWakeRunIdsRef`, `recentlyWakeRunIdsRef`, `pendingWakeRunIdRef`, `activeConnectCountRef`, `pendingRunActivityReconnectRef`, generation counter, `startRunActivityReconnectRef`), a 30s TTL, and a 10ms self-rescheduling idle-poll exist solely to answer "is this `runId` one my own tab originated?" — which core already has the data to answer (it mints/forwards `runId`, and the notification carries it).

A core-side origination registry exposing `isLocallyOriginatedRun(runId)` / a pre-filtered remote-activity subscription would collapse this to one subscription + one `connectAgent()` call, and any other consumer (Angular drawer, headless) then avoids reimplementing the dedup. This also subsumes: the drain block copy-pasted 3× (connect finally / submit finally / suggestion finally — already drifting), the duplicated submit vs suggestion run wrappers, the byte-identical `rememberRecentlyLocalRunId`/`rememberRecentlyWakeRunId`, the 10ms busy-poll (the `finally`-block drain already covers this event-drivenly), and the unnamed `30_000`/`10` literals.

packages/core/src/core/run-handler.ts
```diff
@@ -133,6 +140,27 @@ export class RunHandler {
     return this.core as unknown as CopilotKitCoreFriendsAccess;
   }
 
+  /**
+   * Return a stable restore-tracking key for the logical agent being
+   * connected. Named agents share their last-thread marker across proxy
+   * instances; anonymous agents fall back to object identity.
+   */
+  private getConnectRestoreKey(agent: AbstractAgent): string {
```
BenTaylorDev, line 148: **[LOW] `_lastConnectedThreadIdsByAgent` never prunes anonymous-agent entries.**

Anonymous (no-`agentId`) agents get a monotonic `anonymous:N` string key that's stored strongly in a plain `Map` and never deleted. The WeakMap dedups per instance, but each new anonymous agent instance is a new key, so a long-lived session that creates fresh anonymous agents accumulates `Map` entries indefinitely. Named agents are bounded and fine.

BenTaylorDev (commented):
Review of the cross-device run-sync work. The rename/cross-file surface checks out (`clearReconnectCursor` exists on IntelligenceAgent + ProxiedAgent, `runId` is optional so no caller breaks, the `../../context` re-export is fine). One hard blocker plus a few decisions to make — details inline, summarized here.

**Blocker**
- **connect can hang forever** if the gateway emits `stream_idle` without `replay_complete`. The new `combineLatest([replayComplete$, streamIdle$])` is the *only* terminal path in connect mode (`completeOnRunFinished` is false, so `threadCompleted$` never fires) and there's no timeout. An idle thread with nothing to replay, or an older gateway build, would leave `connectAgent()` unresolved and the socket/channel un-torn-down — and it permanently blocks the new reconnect drain (`activeConnectCountRef` never decrements). Needs a gateway guarantee that `replay_complete` always precedes `stream_idle` (including empty threads), or a fallback/timeout.

**Decisions**
- **HITL respond-after-reload affordance is dropped**: passive replay no longer re-invokes frontend tool handlers (`executeFrontendTools: false`), and the HITL e2e test was correspondingly weakened (the `hitl-respond` assertion deleted). Not re-executing side-effecting tools on replay is correct; losing the ability to complete an interrupted approval from restored state is a product call, not a silent test relaxation.
- **cursor namespace mixing**: `advanceLastSeenEventId` blindly overwrites a single cursor that's fed from both runner ids (`cpki_event_id`) and ingestion cursors (`latest_event_id`/`cpki_ingested`). Gateway-contract dependent, but if a control event lacks `latest_event_id` while events carried `cpki_event_id`, the next reconnect sends a wrong-namespace cursor.

**Altitude**
- The ~200-line wake-reconnect machine (9 refs, 30s TTL, generation counter, 10ms poll) exists only to answer "is this `runId` one my own tab caused?" — which core already has the data to answer. A core-side origination registry (`isLocallyOriginatedRun(runId)` / pre-filtered remote-activity subscription) would collapse this to one subscription + one `connectAgent()`, and spares Angular/headless from reimplementing the dedup. This subsumes the duplicated drain block (3×), the duplicated submit/suggestion run wrappers, the identical `rememberRecently*` helpers, the busy-poll, and the unnamed `30_000`/`10` literals.

**Minor**
- `metadataCredentialsFailed` clears the request latch but nothing re-triggers the effect, so realtime sync stays down until an unrelated list refresh (known gap per the retry-on-refetch test).
- `_lastConnectedThreadIdsByAgent` never prunes `anonymous:N` entries → unbounded growth for no-`agentId` agents in long sessions.
- `wakeReconnectAbortController` is only wired to the agent when `agent instanceof HttpAgent`, but this feature targets `IntelligenceAgent` — so its `abort()` on cleanup is a no-op and cancellation relies entirely on `detachActiveRun()`; effectively dead code for the intended path.

Given it's a Draft with old/new-gateway and two-device flows still pending manual testing, that all tracks — the `replay_complete` hang is the one I'd resolve before it leaves Draft.

BenTaylorDev (commented):
Re-review at `e0553b0` (branch was force-rebased since my last pass at `0fcf37b`). The prior blocker and both structural findings are resolved — details, then the residuals.

### ✅ Addressed since last review
- **Connect-hang blocker** — `streamIdleCompletion$` now `merge`s the old `combineLatest([replayComplete$, streamIdle$])` with a `streamIdle$ + delay(100ms)` fallback, so `stream_idle` without `replay_complete` completes connect instead of hanging. New test covers it. 👍
- **Cursor namespace mixing** — `readDurableEventId` now funnels both event and control cursors and explicitly rejects `cpki_ingested*` ids. Tests updated.
- **Unbounded `anonymous:N` cursor growth** — run-handler now keys per-agent via a `WeakMap`.
- **HITL respond-after-reload** — the weakened assertion is now applied to both the React *and* Vue e2e tests, confirming it's a deliberate product call rather than a silent test relaxation.

### Residual findings (none are hard blockers)

**1. `packages/core/src/intelligence-agent.ts` (~L633) — the 100ms fallback can truncate replay** *(contract-dependent)*
The hang fix trades "hang forever" for "complete after `stream_idle` + 100ms." Under the intended ordering (`replay_complete` precedes `stream_idle`), `combineLatest` fires at 0ms and the fallback never wins — safe. But if the gateway ever emits `stream_idle` before `replay_complete`, or `replay_complete` lags `stream_idle` by >100ms on a non-empty thread, `threadEvents$.pipe(takeUntil(streamIdleCompletion$))` cuts the stream early and the reconnecting tab shows a **truncated** thread. Strict improvement over the old hang, but the 100ms heuristic is an unverified assumption against the gateway (Intelligence#458) — worth pinning down in that contract.

**2. `packages/core/src/intelligence-agent.ts` (~L849) — rejecting `cpki_ingested` can stall cursor advance** *(by-design)*
If a thread's events/control cursors only ever carry `cpki_ingested` (no durable `cpki_event_id`), the reconnect cursor never advances and every reconnect re-replays the full range (duplicate redelivery). Correct **iff** the gateway always stamps durable events with `cpki_event_id`; the tests encode the `null`-cursor expectation, so it's intentional. Please confirm that gateway invariant holds.

**3. `packages/react-core/src/v2/components/chat/CopilotChat.tsx` (~L492) — dead abort wiring for the actual target** *(minor / dead code)*
`wakeReconnectAbortController` is assigned to the agent only `if (agent instanceof HttpAgent)`, but this feature targets `IntelligenceAgent` (extends `AbstractAgent`). So `wakeReconnectAbortController.abort()` on cleanup is a no-op for the real path; cancellation actually relies on `detachActiveRun()`. Not harmful, but misleading — either drop the controller or wire cancellation uniformly.

**4. `packages/react-core/src/v2/components/chat/CopilotChat.tsx` (~L565) — best-effort wake-dedup races** *(low)*
Rapid remote notifications with distinct runIds can drop a runId from `pendingWakeRunIdRef` (last-write-wins) or from `activeWakeRunIdsRef` (unconditionally `.clear()`'d on thread switch, though the ref isn't effect-scoped), occasionally firing one redundant `/connect`. Replay is idempotent, so the impact is a wasted reconnect, not incorrect state.

### Bottom line
Blocker resolved; last round's two structural findings are fixed with tests. The main residual is #1 — a replay-truncation risk that hinges on the gateway's `replay_complete`/`stream_idle` ordering, which is exactly what the merge-order note and the still-pending two-device / old-gateway manual testing defer to Intelligence#458. I'd confirm #1 and #2 against the gateway contract before this leaves Draft, and clean up #3 opportunistically.

BenTaylorDev (commented):
Re-review at `12739f9b4` (branch force-rebased onto v1.62.1 since my last pass at `e0553b0`). One new functional commit — `fix(core): gate idle reconnect fallback by cursor` — resolves residuals #1 and #3. Ran `vitest run src/__tests__/intelligence-agent.test.ts` against the new code: **54/54 pass**.

### ✅ Addressed since last review
- **#1 — 100ms fallback truncating replay** → the unconditional `streamIdle$ + delay(100ms)` fallback is now gated by `canFallbackCompleteConnect(payload, reconnectCursor, latestObservedReplayCursor)`. It only completes connect early when the idle cursor is (a) absent, (b) equal to the reconnect cursor (nothing new since resume), or (c) equal to the latest event actually observed during replay (caught up). All three cursors funnel through `readDurableEventId`, so the comparison is apples-to-apples (durable ids only, `cpki_ingested` rejected) — no namespace mismatch. New test asserts an ahead-of-observed idle cursor does **not** complete after 150ms. 👍
- **#3 — dead abort wiring** → `wakeReconnectAbortController` removed entirely; cancellation relies on `detachActiveRun()`, which was already the real path.

### Residuals (none blocking)

**1. The gate trades truncation-risk for a narrower hang-risk** *(contract-dependent)*
The gate correctly prevents early completion when `stream_idle` arrives ahead of delivered events — but in that case the fallback branch's `take(1)` has already consumed the single `stream_idle`, so completion now depends entirely on `replay_complete` (the `combineLatest` path). If the gateway ever emits `stream_idle` with a cursor ahead of the events it delivered **and** never sends `replay_complete`, connect hangs again — the original blocker's failure mode in a narrower form. This is a strict correctness win over silent truncation and only triggers on a gateway-contract violation, so it's the right call for now — worth pinning `replay_complete` always precedes/accompanies a caught-up `stream_idle` in the Intelligence#458 contract.

**2. Rejecting `cpki_ingested` can stall cursor advance** *(carried over, by-design)*
Unchanged. If a thread's events/control cursors only ever carry `cpki_ingested`, the reconnect cursor never advances (and now `latestObservedReplayCursor` also stays `null`, so the gate falls back to the reconnect-cursor / null-idle branches). Correct **iff** the gateway always stamps durable events with `cpki_event_id`. Please confirm that invariant.

**3. Best-effort wake-dedup races** *(carried over, low)*
`pendingWakeRunIdRef` / `activeWakeRunIdsRef` logic is untouched; rapid distinct-runId notifications can still fire one redundant idempotent `/connect`. Wasted reconnect, not wrong state.

### Bottom line
Both remaining actionable items from last round are fixed with test coverage. What's left (#1, #2) are gateway-contract confirmations that the merge-order note and pending two-device / old-gateway manual testing already route to Intelligence#458. Good to leave Draft once that contract is confirmed and the manual matrix is run.

BenTaylorDev (approved):
Approving at `12739f9b4`. Both actionable items from my prior round are fixed with test coverage (ran `vitest run src/__tests__/intelligence-agent.test.ts` → 54/54 pass), and the client-side rollout is additive under version skew per the merge-order note.

Two residuals are **gateway-contract confirmations**, not code changes — surfacing them here so they're tracked against Intelligence#458 / the pending two-device + old-gateway manual matrix rather than blocking merge:

1. **`replay_complete` ordering vs. `stream_idle`.** The cursor gate on the fallback (`canFallbackCompleteConnect`) means that when `stream_idle` arrives with a cursor ahead of delivered events, completion depends entirely on `replay_complete` (the fallback's `take(1)` has already consumed the idle event). Please confirm the gateway always sends `replay_complete` alongside/before a caught-up `stream_idle` — otherwise connect hangs in that narrow case.

2. **`cpki_event_id` durability.** `readDurableEventId` rejects `cpki_ingested*`, so a thread whose events/control cursors only ever carry `cpki_ingested` never advances the reconnect cursor (and re-replays the full range each reconnect). Correct iff the gateway always stamps durable events with `cpki_event_id`.

Neither blocks this PR — they're the exact contract points the merge order (this PR first, then Intelligence#458) already defers to. Ship the client changes; confirm 1 & 2 before Intelligence#458 lands.

packages/core/src/core/run-handler.ts
IanVS: Can someone explain why this change was made?  It seems like a regression in product experience.  I've got some tests that were asserting on this behavior, and they're now failing in recent versions of copilotkit.  Now, if a HITL decision must be made, and the page is refreshed, the HITL can no longer proceed.

PR #5750, opened 2026-06-28
fix(showcase): re-assert SSOT replica count on promote + skip probe-ineligible services locally

jpr5:
Two related defects in `showcase/bin/railway`'s promote path. Both are now fixed with red-green Ruby specs that drive the real methods (not stubs of the code under test).

## Defect 1 (primary) — promote never re-asserts the SSOT replica count (the 6→1 trap)

**Root cause.** `PromoteCommand.pin_and_verify` (`showcase/bin/railway`) issued `serviceInstanceUpdate(input: { source: { image } })` followed by `serviceInstanceDeployV2` with **no `numReplicas`**. The authoritative count lives in `showcase/scripts/railway-envs.ts` as `workerProvisioning.<env>.effectiveReplicas` (=6 for `harness-workers`) and is exposed in the generated SSOT JSON, but no promote/deploy code path ever pushed it to Railway. Because `serviceInstanceUpdate` is a partial PATCH (omitted keys preserved), a promote didn't itself zero the override — but it never **re-asserted** it either, so once Railway's effective `multiRegionConfig.us-west2.numReplicas` drifted to 1 (manual edit / partial scale that only touched the top-level mirror / Railway-side reset), every subsequent promote left it at 1.

**Fix.** `pin_and_verify` now read-modify-writes `multiRegionConfig.us-west2.numReplicas` from the SSOT `effectiveReplicas` when the target service+env declares a worker count, mirroring the existing `healthcheckPath` self-heal. The `serviceInstanceUpdate` input/mutation is built dynamically (`RestoreCommand.build_update_image_mutation`) so any combination of `{image, healthcheckPath, replicas}` is a faithful partial PATCH.

**Footgun handled (called out in code).** `serviceInstanceUpdate` is a partial PATCH and `multiRegionConfig` is **input-only / non-queryable** — we cannot read-modify-write the live region object. So we set **only** the nested `multiRegionConfig."us-west2".numReplicas` key, add the region object **only** for services the SSOT declares a count for, and **OMIT `multiRegionConfig` entirely otherwise** — never null-clearing sibling region config and never forcing a count on a service that declares none. The top-level `numReplicas` mirror is deliberately not sent (touching the mirror alone is exactly the partial-scale footgun that caused the drift).

The promote call site sources `effectiveReplicas` from the generated SSOT record (`workerProvisioning.prod.effectiveReplicas`) and passes it as `replicas:`; nil for services with no declared count.

### Red → Green (Defect 1)

`showcase/bin/spec/test_promote_replicas_reassert.rb` drives the real `pin_and_verify` and asserts the constructed `serviceInstanceUpdate` input/vars.

RED (origin/main `bin/railway`):
```
PromoteReplicasReassertTest:
  ArgumentError: unknown keyword: :replicas
    bin/railway:1154:in `pin_and_verify'
3 runs, 0 assertions, 0 failures, 3 errors, 0 skips
```

GREEN (with fix):
```
3 runs, 20 assertions, 0 failures, 0 errors, 0 skips
```

The generated mutation (replicas + healthcheck case):
```
input: { source: { image: $image }, healthcheckPath: $healthcheckPath, multiRegionConfig: { "us-west2": { numReplicas: $numReplicas } } }
```
and OMITs `multiRegionConfig` entirely when no count is declared.

**Post-merge live check (not runnable pre-merge):** the next prod promote of `harness-workers` must preserve **6** replicas (confirm via Railway's effective `multiRegionConfig.us-west2.numReplicas`, since the top-level mirror is non-authoritative for this single-region service).

## Defect 2 (secondary) — local `bin/railway` refuses `probe:false` services

**Root cause.** `check_p3_staging_live_green` → `run_staging_probe` (`showcase/bin/railway`) shelled to `verify-deploy.ts` **without `--skip-ineligible`**, so `harness-workers` (`probe.staging=false`, domainless queue worker) went red and P3 emitted `REFUSE: P3: staging is not green`. CI already fixed this in merged commit `160ba5a4aa` (`showcase_promote.yml` passes `--skip-ineligible`); the **local** CLI was missed, which is why the workaround had been `--no-require-staging-green`.

**Fix.** Plumb `--skip-ineligible` into the local `run_staging_probe` invocation — a faithful 1-line port of the merged CI behavior. The flag does not relax green for eligible services, and an unknown/typo name is still a hard error inside `verify-deploy.ts`.

### Red → Green (Defect 2)

`showcase/bin/spec/test_promote_staging_probe_skip_ineligible.rb` drives the real `run_staging_probe`, intercepts the spawn, and asserts the constructed argv.

RED (origin/main `bin/railway`):
```
Expected ["npx", "--yes", "tsx", ".../verify-deploy.ts", "--env", "staging", "--services", "harness-workers", {:err=>[:child, :out]}] to include "--skip-ineligible".
1 runs, 4 assertions, 1 failures, 0 errors, 0 skips
```

GREEN (with fix):
```
1 runs, 12 assertions, 0 failures, 0 errors, 0 skips
```

## Files changed
- `showcase/bin/railway` — `build_update_image_mutation` helper + `replicas:` param on `pin_and_verify`; `--skip-ineligible` in `run_staging_probe`; replica sourcing at the promote call site. Removed the now-unused `UPDATE_IMAGE_AND_HEALTHCHECK_MUTATION` constant (superseded by the dynamic builder).
- `showcase/bin/spec/test_promote_replicas_reassert.rb` — new (Defect 1).
- `showcase/bin/spec/test_promote_staging_probe_skip_ineligible.rb` — new (Defect 2).
- `showcase/bin/spec/test_snapshot_ivar_lint.rb` — renumbered the line-pinned allowlist to track the shifted lines (no behavior change).

Full suite: `ruby showcase/bin/spec/all_tests.rb` → `179 runs, 709 assertions, 0 failures, 0 errors, 0 skips`.

### DIFF
--- a/showcase/RAILWAY.md
+++ b/showcase/RAILWAY.md
@@ -325,20 +325,27 @@ live staged config now reads `6` in both envs.
   (= Railway `serviceInstance.drainingSeconds`, env mirror
   `RAILWAY_DEPLOYMENT_DRAINING_SECONDS`). See **Deploy rollover** below.
 
-### Applying a replica count change (MANUAL)
+### Changing the replica count
 
-The `emit-railway-envs-json.ts` emitter and `bin/railway` tooling are
-**VERIFY-ONLY** with respect to the replica count — they do not write replica
-counts to Railway. To change the replica count:
+The `emit-railway-envs-json.ts` emitter is **VERIFY-ONLY** with respect to the
+replica count (it reads/reports it; it never writes it). `bin/railway promote`,
+however, **RE-ASSERTS** the SSOT-declared `effectiveReplicas` on every pin: when
+a service declares a count, the pin mutation sets
+`multiRegionConfig.us-west2.numReplicas` from the SSOT, so a service whose live
+override silently drifted (the 6->1 trap) self-heals to the tracked count on the
+next promote. (Services that declare no count are left untouched — promote never
+forces or null-clears their region config.) So to **change** the declared count:
 
 1. Change the `effectiveReplicas` value in `railway-envs.ts` (SSOT) — and the
    `numReplicas` mirror alongside it (keep them equal for this single-region
    service).
 2. Regenerate the snapshot: `npx tsx showcase/scripts/emit-railway-envs-json.ts`
 3. Commit both files (`railway-envs.ts` + `railway-envs.generated.json`).
-4. Apply the change to Railway manually via the Railway Dashboard (Service >
-   Settings > Replicas, which edits `multiRegionConfig.us-west2.numReplicas`) or
-   the Railway GraphQL API.
+4. The next `bin/railway promote` propagates the new count to Railway (it sets
+   `multiRegionConfig.us-west2.numReplicas` from the SSOT). To apply it out of
+   band before the next promote, edit it manually via the Railway Dashboard
+   (Service > Settings > Replicas, which edits
+   `multiRegionConfig.us-west2.numReplicas`) or the Railway GraphQL API.
 
 The CI drift gate (`showcase/scripts/__tests__/harness-workers-provisioning.test.ts`)
 will fail if `railway-envs.ts` and `railway-envs.generated.json` disagree on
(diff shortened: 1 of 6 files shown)


jpr5:
Closing as obsolete. Both defects already landed on main the same day via separate PRs: the SSOT replica re-assert by #5756 (using a `replica_config` region→count hash API) and the skip-ineligible behavior by #5752 (filtering `STAGING_PROBE_INELIGIBLE` services out before `run_staging_probe`, rather than an argv flag). This PR's parameter shapes and test assertions diverged from what shipped — rebasing would be a full rewrite with no net value over current main.

PR #5761, opened 2026-06-29
feat(bot,runtime): managed bots SDK — run the bot SDK from Intelligence-delivered events (OSS-360/361)

AlemTuzlak:
## Summary

Lets the `@copilotkit/bot` SDK run from **Intelligence-delivered events** without a second programming model, and adds the runtime `bots` declaration API. A managed event (delivered by Intelligence) runs the *same* customer handlers, tools, context, commands, Bot UI, and agents as local/custom adapters — the managed path is "just another `PlatformAdapter`," fed by injected transports.

This is the **OSS / SDK slice** of the Hosted Managed Bots work. The credentialed transports (Realtime Gateway, Connector Outbox) and the frozen shared contracts live elsewhere (see *Out of scope*); this PR ships the seams they plug into, fully runnable headless.

Relates to **OSS-360** (runtime bots API), **OSS-361** (run the SDK from Intelligence events), **OSS-363** (Slack render/codec reuse).

## What's in here

- **`intelligenceAdapter()` bridge** (`@internal`, not publicly documented) — implements `PlatformAdapter` over two injected transports: `DeliverySource` (inbound) + `EgressSink` (outbound). Ingress → `onTurn`/`onCommand`/`onInteraction`/`onThreadStarted`/`onReaction`; ack on success / nack on throw (at-least-once). Egress emits generic operations carrying `BotNode[]` IR with **deterministic ids** (`turnId:seq`, reset per turn) so a redelivered turn reproduces the same ids for the Connector Outbox to dedupe. Idempotency lives at egress, so the managed path skips ingress dedup (`skipIngressDedup`) — a redelivery re-runs rather than being dropped.
- **Runtime `bots` API** — `new CopilotRuntime({ intelligence, bots })`, accepted by TypeScript **only when `intelligence` is configured** (discriminated union). `createBot({ name })`; `startManagedBots()` validates names (required, identifier-style, unique — fail-loud), builds activation metadata, and wires each bot to its resolved transport.
- **`PlatformCodec` seam** + Slack egress codec (`slackCodec`) composing the existing pure `renderSlackMessage`, so IR→native rendering is shared (no Bolt/creds) instead of duplicated.
- **Backwards-compatible SDK foundations**: `bot.addAdapter()` + optional `adapters`, deferred backend resolution at `start()` with `stateStore`-provider precedence (+ multi-provider warning), `bot.transcripts` throws pre-start, optional `eventId`/`turnId`/`deliveryId` on ingress + handler context. Existing `createBot` callers and every `PlatformAdapter` implementer are unaffected.
- **In-memory transports + fixture tests** — the full dispatch path (envelope in → handler runs → egress op out) runs with zero Slack/Intelligence/network.

## Out of scope (external / separate tickets)

- **Realtime Gateway + Connector Outbox transports** — implemented in the closed-source repo against the `DeliverySource`/`EgressSink` interfaces shipped here.
- **Shared contracts freeze (OSS-377)** — consumed here via a minimal, isolated placeholder (`managed/contracts.ts`, marked `TODO(OSS-377)`); swaps in via one import change.
- **OSS-363 ingress normalization** — the egress codec is done; extracting the pure Slack event→neutral mapping out of the Bolt listener (so local + Intelligence ingress share it) is the remaining, higher-risk half and is left to that ticket (`TODO(OSS-363)`).

## Testing

TDD throughout (RED→GREEN per behavior). New: managed adapter dispatch/ack-nack/ids/run-renderer/exclusivity, all-kinds routing, name validation + metadata + lifecycle, runtime `bots` option, Slack codec. Full suites green: `bot` 147, `bot-slack` 256, `runtime` 1574. All builds typecheck (`bot`/`bot-slack`/`bot-discord`/`runtime`); oxlint/oxfmt clean.


packages/bot-intelligence/src/contracts.ts
```diff
@@ -0,0 +1,95 @@
...
+import type { BotNode, MessageRef } from "@copilotkit/bot-ui";
+
+/**
+ * Opaque return address minted by Intelligence and echoed back on egress. The
+ * SDK never interprets it — it only carries it from ingress to the egress sink.
+ */
+export type EgressRoute = unknown;
+
+/** Fields common to every managed ingress envelope. */
+export interface ManagedIngressBase {
```
AlemTuzlak, line 17: @tylerslaton important for your work

BenTaylorDev (commented):
## Review — managed bots SDK foundations

Reviewed at the current head (post main-merge). The adapter core (deterministic egress ids, at-least-once ack/nack) and the codec reuse are genuinely good, and the backwards-compat *intent* is careful. But "fully backwards-compatible" overstates it, there's one reinit hazard, a user-visible ingress double-fire, and a couple of cross-repo/contract couplings worth settling before this base lands. Findings below, most severe first; file:line included.

### Correctness
- **C1 — MED-HIGH: `start()` has no idempotency guard** (`packages/bot/src/create-bot.ts:804`). A second `start()` re-resolves the backend and unconditionally rebuilds Transcripts/Telemetry/ActionRegistry and re-calls `adapter.start()` on every adapter. With a `MemoryStore` all lock/dedup/transcript/action state is wiped; real adapters (Slack/Discord) get a second connect/port-bind. `main` built the backend once at construction, so moving init into `start()` introduced this. `started` is only consulted to block `addAdapter`. Fix: guard the top of `start()`.
- **S1 — MED: managed thread-reply @-mention double-responds** (`packages/bot-slack/src/ingress-normalize.ts:184`). Native `slack-listener.ts` skips a thread reply that @-mentions the bot (`app_mention` runs separately); the managed `thread_reply` branch has no equivalent guard and takes no `botUserId`. Slack delivers a threaded @-mention as two events → the managed path emits both an `app_mention` and a `thread_reply` turn → double response. The new `ingress-normalize.test.ts` case codifies the divergent behavior. Fix: port the duplicate-mention skip (accept `botUserId`, return `undefined` when the reply mentions the bot).
- **S2 — MED: `bots` silently no-ops at runtime without `intelligence`** (`packages/runtime/src/v2/runtime/core/runtime.ts:376`). The discriminated union rejects it at compile time, but `hasIntelligenceOptions` keys only on `intelligence`, so a JS / `as any` caller passing `{ agents, bots }` builds a plain runtime with `bots` dropped — no throw, no warn. Add a runtime guard mirroring the type.
- **S3 — MED: `startManagedBots` has no partial-start rollback** (`packages/bot-intelligence/src/runtime.ts:130`). Sequential `addAdapter` + `await start()`; if bot N fails, bots 0..N-1 are already live but no handle is returned → leaked listeners/connections. Wrap and `stop()` the started ones before rethrowing.

### Backwards-compatibility (the "transparent" claim overstates)
- **C2 — MED:** `bot.transcripts` now throws before `start()` (`create-bot.ts:716`); it was a concrete property on `main`. Existing `createBot(...).transcripts.…`-before-start callers break. Defensible design — but changelog it rather than describe as transparent.
- **C3 — MED:** telemetry `oss.bot.configured` moved from construction-time to `start()`-time (`create-bot.ts:832`); a constructed-but-never-started bot no longer emits it. Confirm this analytics-semantics shift is intended.

### Cross-repo / contract (Intelligence integration)
- **I1 — MED-HIGH: render-event `kind:"file"` is rejected on the realtime path.** The SDK contract (`contracts.ts`) and app-api's HTTP render-accept (zod) both accept `file` (`thread.postFile`), but the realtime gateway validator has no `file` clause → `invalid_render_event`. So outbound files work over the HTTP transport but silently fail over the realtime/Phoenix path (the "production render/delivery path"). Needs a `file` clause on the Intelligence gateway before realtime `postFile` works.
- **I2 — MED: `duplicate_skipped` acceptance is now dead.** `RenderAccepted.acceptance` still lists `duplicate_skipped`, but the Intelligence side now returns only `duplicate_accepted` or a 409 conflict on a payload mismatch for the same `(delivery, turn, slot, seq)`. A mismatched re-push becomes a thrown render error (turn nack) rather than a benign skip. Drop the dead value and handle the 409.
- **I3 — lease-token fencing:** neither transport sends a `leaseToken` on render (`HttpRenderEventSink` exposes only `scopeFor`; the Phoenix transport omits it too). When Intelligence tightens render-accept to require the lease token, both call sites need it — tracked on the Intelligence side, flagging that this is where it's rooted.

### Lower
- **S4 — LOW/MED:** `stripMentions` (`ingress-normalize.ts:14`) misses Slack's labeled form `<@U123|handle>` → leftover `|handle>`; DMs also keep mention tokens (unlike app_mention/thread_reply).
- **S5 — LOW:** bot-name uniqueness is case-sensitive (`"Support"` vs `"support"` both pass), and nothing guards against two bots resolving to the same transport (both then receive every delivery).
- **S6/S7/S8 — LOW:** name validation is deferred entirely to `startManagedBots` (no fail-fast at construction); `buildActivationMetadata` silently filters nameless bots; empty `bots: []` yields a silent no-op handle.
- **M1 — LOW:** the adapter's per-turn `seq` Map (`intelligence-adapter.ts:215`) is never cleared → unbounded growth over a long-running managed bot.
- **M2 — LOW:** `buildContentParts` fail-open — an inbound file whose fetch fails is logged and silently dropped from the model's context (user attaches an image, model never sees it).

### Test coverage
- **T1 — MED:** near-zero coverage for the managed-foundation surface this PR adds — `assertExclusive` (managed+direct rejection), the `skipIngressDedup` redelivery-not-dropped branch, and `botName` propagation to `adapter.start(ctx)` are untested; store-resolution tests assert the *warning* but not the *store actually chosen*; and there's no test for the threaded-mention dedup (S1) or labeled-mention stripping (S4).

### What's solid
`resolveBackend` precedence is correct across 0/1/N (explicit wins, N warns + uses `providers[0]`, no silent wrong-store); the new `PlatformAdapter`/capability/ingress fields are additive+optional (existing implementers compile unchanged); egress parity is real (`slackCodec.renderEgress` is the identical `renderSlackMessage`); fail-loud posture unchanged (no new empty catches / unawaited promises); no production `as any`; deterministic egress-id design is sound and cross-instance-safe.

### Suggested order
1. C1 (start-guard) and S1 (double-reply) before merge — state-corrupting / user-visible.
2. S2, S3, I1.
3. C2/C3 changelog; T1 + the S4/M-series; I2/I3 coordinate with the Intelligence side.

BenTaylorDev (approved):
## Approving — review findings addressed

Re-reviewed at the current head. All actionable findings from my earlier review are resolved correctly (verified each diff, not just presence), with tests, and CI is green (unit 20.x/24.x, runtime node/bun, e2e/showcase):

- **start() idempotency** — guard added + `started` reset in `stop()` so restart still re-inits (idempotent-second-start and real-restart both tested).
- **thread-reply @-mention double-response** — skipped via `botUserId` (plain + labeled forms), with a test.
- **`bots` without `intelligence`** — SSE runtime now throws (fail-loud) instead of silently dropping.
- **startManagedBots partial-start rollback** — started bots are stopped before rethrow.
- **labeled mentions `<@U…|handle>`** — regex handles them; DMs now strip mentions too.
- **case-insensitive bot-name uniqueness**, construct-time missing-name check, `buildActivationMetadata` throws instead of filtering, empty-`bots` warning.
- **adapter seq Map** cleared per turn; **inbound-file fetch failure** now surfaces a visible note instead of dropping silently.
- **`duplicate_skipped`** removed from the render-acceptance union and documented against the 409-conflict behavior.

### Not blocking this PR (tracking separately)
- **Realtime `kind:"file"` render frame** is rejected by the Intelligence gateway validator (no `file` clause) — a cross-repo fix on the Intelligence side, not this SDK PR. Filing separately so realtime `thread.postFile` isn't silently broken.
- **Lease-token on render** is deferred to the Intelligence-side follow-up (OSS-446).
- **Changelog** note for the two intentional behavior changes (`transcripts` throws pre-start; telemetry `configured` timing) would be nice.
- Minor: `start()` sets `started = true` before the init block, so a synchronous throw there wedges a retried `start()` until `stop()` — low-probability, cleaner to set it at the end / reset on catch.

Nice, thorough turnaround. LGTM.

PR #5768, opened 2026-06-29
feat(web-inspector): add unified thread debugger

marthakelly:
## Summary
- Adds the unified web-inspector thread debugger surface as a standalone `cpk-thread-inspector` custom element backed by a `ThreadDebuggerProvider` contract for thread metadata, messages, AG-UI events, and state.
- Keeps the existing full Web Inspector shell compatible by making the internal `cpk-thread-details` element extend the new shared inspector implementation.
- Normalizes persisted AG-UI event history into a first-class Timeline view with run/message/tool/state/raw rows, source-event links into the Raw AG-UI Events tab, metadata pills, and raw-event preservation for top-level fields.
- Improves timeline/state loading behavior for standalone and runtime-backed usage: state stays lazy-loaded, provider/runtime/header swaps refetch safely, stale async responses are ignored, and message-backed or raw-only threads no longer show an empty first tab.
- Adds a standalone web-inspector dev harness with AG-UI events, Messages only, and Raw event only scenarios for shared UI validation outside an app shell.
- Updates package/dev configuration for the harness (`dev:standalone`, Vite dev dependency/config allowlist) and bumps `@copilotkit/web-inspector` to `1.61.3`.

## Tests and Checks
- Expanded `packages/web-inspector` tests for the provider contract, normalized timelines, raw event rendering, lazy state loading, stale request handling, header-driven refetches, runtime fallbacks, and first-tab timeline population.
- PR checks currently passing include build, check-types, format, oxlint, package-quality, unit on Node 20/22/24, Python unit tests, shell script tests, Validate Showcase, bundle-size, config/binary checks, and integration starter/showcase jobs.
- Manual validation path documented in `packages/web-inspector/README.md`: `pnpm nx run @copilotkit/web-inspector:dev:standalone`, then validate AG-UI events, Messages only, Raw event only, source-event navigation, and State at `http://127.0.0.1:5177/`.

## Companion PR
- Intelligence managed debugger integration: https://github.com/CopilotKit/Intelligence/pull/461

Screenshot 2026-06-29 at 2 50 47 PM
Screenshot 2026-06-29 at 2 49 50 PM
e733" />
Screenshot 2026-06-29 at 2 49 58 PM
Screenshot 2026-06-29 at 2 50 07 PM


BenTaylorDev (dismissed):
## Review

Refactors `ɵCpkThreadDetails` into a shared `CpkThreadInspector` (`cpk-thread-inspector`) backed by a pluggable `ThreadDebuggerProvider`, replaces the Conversation/Agent-State/AG-UI-Events tabs with a normalized **Timeline** / **State** / **Raw AG-UI Events** layout, and adds a standalone Vite dev harness. The old element name stays as a thin compatible subclass, so the existing shell keeps working.

This is well-guarded — abort controllers, `threadId` re-checks, and the composite `loadKey` correctly handle provider/runtime/header swaps, and test coverage is thorough. I traced consumers (the only host is `WebInspectorElement` in the same file), checked the shared-cache-slot path for stale reads, and walked the 501/header-change/thread-switch races — **no correctness bugs surfaced.** The findings below are quality and fragility concerns; none block the feature.

**1. Timeline normalization re-runs on every render, partially defeating the panel-template cache** — `src/index.ts` (`renderTimeline`)
`renderTimeline()` reads `this.activeTimelineItems` (an unmemoized getter running the full O(n) `timelineItemsFromEvents(activeEvents)` pass) for its empty-check *before* consulting `cachedPanelTpl`. On a thread with hundreds of events, every re-render and tab-switch back to Timeline re-iterates all events even when the cached template is returned unchanged — exactly the cost the panel cache was added to avoid. (`fetchEvents` computes it a third time for a length-check that's redundant with `length === 0`, since every non-empty event list yields ≥1 item.)

**2. `renderTimeline` and `renderConversation` share the `"timeline"` cache slot** — `src/index.ts` (`renderConversation`)
When events are empty, `renderTimeline` falls back to `renderConversation()`, which caches under `"timeline"` with a length-3 key, while `renderTimeline`'s own path uses a length-1 key. This is safe *only by coincidence* — `cachedPanelTpl` compares key length first, so the differing lengths force a rebuild on each flip. A future change making the key arrays the same length would silently serve a stale panel. A distinct slot (e.g. `"timeline-fallback"`) removes the trap.

**3. Unrelated formatting churn bloats the diff** — `src/index.ts` (WebInspectorElement render)
A large fraction of the +1922/−201 is cosmetic re-wrapping of untouched markup (events-table `<col />`, announcement block, message-role badge, `CpkThreadList` error SVG) far from the refactor, which obscures the functional changes. Worth splitting out to keep the diff focused.

**4. Dev harness couples to an exact source line via string replace** — `dev/vite.config.ts`
The transform does `code.replace('import tailwindStyles from "./styles/generated.css";', …)`. If that import ever changes (quotes, whitespace, path), the replace silently no-ops, CSS isn't injected, and `unsafeCSS` breaks with no error. A guard that throws when the expected substring isn't found would fail loudly instead.

Items 1–2 are the ones I'd most want addressed; 3–4 are low-severity.

packages/web-inspector/src/index.ts
```diff
@@ -1775,7 +2391,11 @@ export class ɵCpkThreadDetails extends LitElement {
     // threads (those would render identically for every thread on the same
     // agent and mislead the reader).
     if (this._eventsNotAvailable) return [];
-    return this._fetchedEvents ?? this.agentEventsInput ?? [];
+    const events = this._fetchedEvents ?? this.agentEventsInput ?? [];
+    if (events.every((event) => event.sourceIndex != null)) return events;
+    return events.map((event, index) =>
```
BenTaylorDev: **Timeline & Raw-events caches are defeated on the live `agentEventsInput` path.**

On the fetch path every event carries `sourceIndex` (stamped by `mapApiEvents`), so `activeEvents` returns the stable `_fetchedEvents` reference and the panel/normalization caches hold. But on the live fallback path this `.map(...)` runs and returns a **brand-new array on every getter call**. The shell feeds `InspectorEvent[]` into `agentEventsInput`, and `InspectorEvent` has no `sourceIndex` (see line 194), so this branch is taken whenever `threadEndpoints.inspect === false` (`canFetchEvents()` false → events never fetched → fall through to `agentEventsInput`).

Per render this means:
- `renderTimeline` keys the panel cache on `[events]` by identity → always misses → `timelineItemsForEvents` sees a new array → re-normalizes → `cachedPanelTpl` rebuilds the whole timeline DOM.
- `renderEvents` (Raw tab) does the same — and this is a **regression**: the old `activeEvents` returned `agentEventsInput` unchanged, so the old `ag-ui-events` cache held.

`liveMessageVersion` bumps on every streaming chunk drive re-renders, so it thrashes exactly on the hot path with hundreds of events (the multi-second stall the cache was built to prevent).

**Fix:** stamp `sourceIndex` once when ingesting `agentEventsInput` (or memoize `activeEvents` by input-array reference) so the returned identity is stable across renders.

packages/web-inspector/src/index.ts
```diff
@@ -1473,26 +1773,25 @@ export class ɵCpkThreadDetails extends LitElement {
...
-        { headers: { ...this.headers }, signal: controller.signal },
-      );
-      if (controller.signal.aborted || this.threadId !== threadId) return;
-      if (!res.ok) throw new Error(`HTTP ${res.status}`);
-      const data = (await res.json()) as { messages: ApiThreadMessage[] };
+      const messages = this.provider?.getMessages
+        ? await this.provider.getMessages(threadId, {
+            signal: controller.signal,
+          })
+        : await this.fetchRuntimeMessages(threadId, controller.signal);
```
BenTaylorDev, line 1798: **`_timelineItemsCache` is not cleared here alongside `_panelTplCache`.**

`resetLoadedThreadData` resets `_panelTplCache` but leaves `_timelineItemsCache` (declared ~line 857) intact. Today this is masked — the fetch path always rebuilds `_fetchedEvents`, and the live path returns fresh arrays each call so the cache misses anyway — so it isn't observable as written.

But it becomes a live stale-rows hazard the moment the finding above is fixed by memoizing `activeEvents`: two threads sharing the same `agentId` could then share an events array reference, and `_timelineItemsCache.events === events` would serve thread A's normalized rows for thread B.

**Fix:** add `this._timelineItemsCache = null;` to this reset, in the same change as the `activeEvents` memoization.

packages/web-inspector/src/index.ts
marthakelly: Fixed in 5df64923cc: live fallback events without sourceIndex now reuse a memoized source-indexed array, so Timeline and Raw Events cache keys stay stable across renders. Added a regression test covering both panels.

packages/web-inspector/src/index.ts
marthakelly: Fixed in 5df64923cc: resetLoadedThreadData now clears _timelineItemsCache alongside the panel cache, and the existing thread-switch cache test now asserts the timeline-items cache is cleared.

BenTaylorDev (approved):

PR #5767, opened 2026-06-29
docs(cookbook): mark OpenBox recipe source as "to follow" until the showcase lands

contextablemark:
## What does this PR do?

The **OpenBox Governance** cookbook recipe (#5686) merged to `main` ahead of its companion showcase (#5685), so the recipe's **"Get the code"** section linked to:

`https://github.com/CopilotKit/CopilotKit/tree/main/examples/showcases/openbox-governed-copilotkit`

…which **404s** today because that code isn't on `main` yet.

This swaps the broken link for a plain-text **"Full source to follow"** note (no hyperlink, so nothing 404s) that still describes what the showcase will contain. The live **upstream reference-repo** link directly below it is kept.

```diff
- Full source: [`examples/showcases/openbox-governed-copilotkit`](https://github.com/.../tree/main/examples/showcases/openbox-governed-copilotkit) — `agent/` … and `frontend/` …
+ Full source to follow — the runnable showcase (the `agent/` LangGraph service with OpenBox middleware and the `frontend/` CopilotKit V2 chat with the wrapped runtime and approval route) will be published under `examples/showcases/openbox-governed-copilotkit`.
```

**Follow-up:** once the showcase merges to `main`, restore the direct `tree/main/...` link.

_Note:_ the recipe's run-instructions still `cd` into that path (expected — the whole "run it yourself" flow assumes the code), so those steps only work once the showcase lands. Left as-is since they read as setup instructions, not a clickable link.

## Related PRs and Issues

- Docs recipe (merged): #5686
- Companion showcase (pending): #5685

## Checklist

- [x] I have read the [Contribution Guide](https://github.com/copilotkit/copilotkit/blob/master/CONTRIBUTING.md)
- [x] If the PR changes or adds functionality, I have updated the relevant documentation
- [x] "Allow edits by maintainers" is checked

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Fr5HVeDzDyC4S6DjyhAFWZ

---
_Generated by [Claude Code](https://claude.ai/code/session_01Fr5HVeDzDyC4S6DjyhAFWZ)_

PR #5686, opened 2026-06-24
docs(cookbook): OpenBox Governance recipe

jerelvelarde:
## What this adds

A new **"OpenBox Governance"** recipe in the cookbook (`showcase/shell-docs`), documenting how to add **OpenBox runtime governance** — guardrails, OPA/Rego policy, redaction, human-in-the-loop approvals, and halt — to a **CopilotKit + LangGraph** agent, with decisions rendered as generative UI.

Follows the existing cookbook recipe pattern (Oracle / Arcade / Daytona):

- `src/content/docs/cookbook/openbox-governed-copilotkit.mdx` — the recipe (how it works, the stack, prerequisites, run + **provisioning** steps, the four-verdict governance matrix, the key pieces in code, going further).
- `meta.json` — registers the recipe in cookbook nav.
- `index.mdx` — overview `<Card>`.
- `src/lib/sidebar-icon.tsx` — `custom/openbox` sidebar icon.
- `public/logos/openbox.png` — brand mark (Git LFS).

The recipe documents the companion showcase **at full parity**: the provisioning step (`npm run openbox:admin:setup`), the Allow / Constrain / Approval / Block / Halt matrix using the demo's real suggestion prompts, and "key pieces in code" snippets pulled verbatim from the final showcase source (the LLM-driven governed engine, the OpenBox-middleware-first agent, and the runtime + approval routes).

## Red → green (TDD)

The cookbook-nav test in `src/lib/__tests__/docs-render.test.ts` is the red/green hook — it hard-asserts the recipe count, titles, slugs, and URLs (6 entries incl. `["OpenBox Governance", "cookbook/openbox-governed-copilotkit"]`).

## Verification

- `docs-render` test suite: **15/15** (nav asserts 6 entries).
- `next build`: ✅ (all cookbook routes, no MDX/link errors).
- The run-steps, four-verdict matrix, and code snippets were verified against the **final** demo source (companion PR #5685), so the docs and the runnable code stay in sync.

## Companion demo PR

The runnable showcase this recipe documents: **https://github.com/CopilotKit/CopilotKit/pull/5685**

> **Ships standalone.** This recipe no longer links the hosted live demo — that link now lives on the companion demo PR (#5685), so the cookbook recipe can merge independently of the demo deployment.

PR #5770, opened 2026-06-29
docs(cookbook): polish OpenBox recipe per launch feedback

jerelvelarde:
## What does this PR do?

Follow-up polish on the **OpenBox Governance** cookbook recipe (shipped in #5686), addressing review feedback from the launch. All changes are in the single recipe page; no functional/code changes.

| Feedback | Change |
| --- | --- |
| "Make this an actual diagram — hard to read as ASCII" | Replaced the ASCII flow under **How it works** with a real, color-coded **inline-SVG architecture diagram** (same self-contained pattern as the Oracle recipe — no external assets). |
| "Huge, scary warning — condense it" | Cut the wall-of-text **provisioning** warning down to a few lines while keeping the essentials (idempotent, which keys it needs, how to verify). |
| "Prompts are unreadable in the chart layout → accordions" | Converted the **Try it** governance-matrix table into one **accordion per prompt** (verdict-labeled), so the long prompts are on-demand instead of crammed into table cells. |
| "Highlight the important stuff — guide the user. Applies to all code." | Added line highlighting across the bash + TypeScript samples so the key lines (the runtime wrap, `selfGovernedToolNames`, middleware-first, the provisioning command, the approval schema) stand out. |
| "Move this to the top top and make it an accordion" | Moved the **coding-agent prompt** from the bottom to just under the intro, collapsed in an accordion. |

## Verification

Rendered locally against `shell-docs` (`next dev`) and confirmed in-browser: the SVG diagram renders, all five accordions expand correctly, the warning is condensed, and code highlighting shows on both bash and TS blocks (no leaked `[!code]` markers).

## Related

- Recipe (merged): #5686
- Source-link hotfix (merged): #5767
- Companion showcase (pending): #5685 — once it lands, restore the direct `tree/main/...` source link in **Get the code**.

### DIFF
--- a/showcase/shell-docs/src/content/docs/cookbook/openbox-governed-copilotkit.mdx
+++ b/showcase/shell-docs/src/content/docs/cookbook/openbox-governed-copilotkit.mdx
@@ -10,216 +10,8 @@ doc_type: how-to
 
 The example is a **governed business assistant**: it triages operations queues, drafts customer updates and exception reports, prepares vendor handoffs, issues service credits, and more — ten governed actions in all, each produced by a **governed LLM generation** step. Every request is checked by OpenBox first, so money movement is held for a human Approve/Reject, goal-drifting exports are blocked, and critical payment-control changes halt the session outright.
 
-## How it works
-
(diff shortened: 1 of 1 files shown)


contextablemark (approved):
Looks good

PR #5772, opened 2026-06-30
fix(runtime): don't forward hop-by-hop/platform headers to agents

### LINKED ISSUE #5712: Runtime forwards inbound `authorization`/`x-*` headers to agents and overrides server-set headers (breaks service-to-service auth to private Cloud Run)
## Summary

When a self-hosted `CopilotRuntime` is configured with a server-side agent (`agents: { myAgent }`, an `@ag-ui/client` `HttpAgent` with static `headers`), the runtime **forwards every inbound `authorization` and `x-*` request header onto the agent's outgoing call and lets them override the headers the server configured on the agent** — with no allowlist control or opt-out.

This silently breaks **service-to-service auth to a secured backend**: a server-set `Authorization: Bearer <service-token>` can be replaced (or shadowed) by whatever the browser/edge/platform put on the inbound request.

## Where

`copilotRuntimeNextJSAppRouterEndpoint` routes to the v2 single-route handler, which clones the agent and calls `configureAgentForRequest` before the run:

- `dist/v2/runtime/handlers/shared/agent-utils.mjs:45` — the merge puts forwarded headers **last**, so they win on collision:
  ```js
  agent.headers = {
      ...agent.headers,                      // what the server configured
      ...extractForwardableHeaders(request), // inbound request headers — override the above
  };
  ```
- `dist/v2/runtime/handlers/header-utils.mjs:7` — the allowlist forwards **`authorization` and every `x-*`**:
  ```js
  function shouldForwardHeader(headerName) {
    const lower = headerName.toLowerCase();
    return lower === "authorization" || lower.startsWith("x-");
  }
  ```

**Versions:** `@copilotkit/runtime@1.61.0`, `@copilotkit/react-core@1.61.0`, `@ag-ui/client@0.0.57`.

## Why this is dangerous in practice (the case that bit us)

We run the CopilotKit runtime inside a Next.js proxy on **Firebase App Hosting** (Google Cloud Run) and bridge to a **private** ADK agent on Cloud Run via an `HttpAgent`. The proxy mints a Google **OIDC ID token** and sets it as the agent's `Authorization` for IAM invoker auth.

Google's serverless platform injects headers onto the inbound request to the proxy — notably **`X-Serverless-Authorization`**, an alternative IAM-token header. Because it's an `x-*` header, the runtime forwards it to the agent. **Cloud Run's IAM frontend validates `X-Serverless-Authorization` in preference to `Authorization`**, so the agent service validated that forwarded (foreign, wrong-audience) token instead of our valid `Authorization` → `401 "The request was not authorized to invoke this service. The access token could not be verified."`

The bearer we set was provably correct (right `aud`, `email`, signature — the identical token authenticated a sibling endpoint we call with a direct `fetch()` that does *not* go through the runtime). The only difference was the runtime forwarding the platform header. Forwarding hop-by-hop / platform headers (`x-serverless-*`, `x-forwarded-*`, `x-cloud-trace-context`, etc.) across a trust boundary to the upstream agent is the core problem.

## Reproduction

1. `CopilotRuntime({ agents: { a: new HttpAgent({ url, headers: { Authorization: 'Bearer <service-token>' } }) } })` behind `copilotRuntimeNextJSAppRouterEndpoint`.
2. Make the inbound browser→runtime request carry an `authorization` (or any `x-*`) header — e.g. a platform/edge-injected one, or a browser that sets `Authorization`.
3. The outgoing request to the agent carries the **inbound** `authorization`/`x-*`, not (or in addition to) the server-set one. (When both an inbound `authorization` and a server `Authorization` collide, undici can also comma-join them into a "multiple JWTs" value.)

## Expected vs actual

- **Expected:** headers a server explicitly configures on a registered agent are authoritative; inbound client headers should not silently override them, and there should be a way to control/disable header forwarding (esp. for non-`x-copilotkit-*` platform headers).
- **Actual:** inbound `authorization` + all `x-*` are forwarded and take precedence, with no opt-out.

## Suggested fixes (any one would help)

1. **Precedence:** make server-configured `agent.headers` win over forwarded headers (`{ ...extractForwardableHeaders(request), ...agent.headers }`), or skip forwarding a header the agent already sets.
2. **Opt-out / allowlist control:** a runtime/agent option to disable forwarding or restrict it to a caller-specified header allowlist.
3. **Tighter default allowlist:** don't forward `authorization` or hop-by-hop/platform `x-*` headers (`x-serverless-*`, `x-forwarded-*`, `x-cloud-trace-context`, …) by default — these describe the client→runtime hop and shouldn't reach the upstream agent.
4. **Docs:** document the `@ag-ui/client` `HttpAgent` `fetch` option as the supported way to attach service-to-service auth that the runtime can't override.

## Workaround (for others hitting this)

Pass a custom `fetch` to the `HttpAgent` that builds the outgoing headers **from scratch** and forwards nothing from the inbound hop — it runs as the actual HTTP call, after `configureAgentForRequest`, and is preserved across the runtime's per-request `agent.clone()`:

```ts
new HttpAgent({
  url,
  fetch: async (target, init) => {
    const inbound = new Headers(init.headers);
    const headers = new Headers();
    headers.set('Content-Type', inbound.get('content-type') ?? 'application/json');
    headers.set('Accept', inbound.get('accept') ?? 'text/event-stream');
    headers.set('Authorization', `Bearer ${await mintServiceToken()}`); // your auth, authoritative
    return fetch(target, { ...init, headers });
  },
});
```

Likely related: #3170 ("Authorization header not passed to agents using A2A").

serhiizghama:
Closes #5712.

The v2 runtime forwards inbound `authorization` and every `x-*` header onto the agent's outgoing call. That sweeps up platform/proxy headers that only describe the client→runtime hop and shouldn't cross the trust boundary to the upstream agent.

The concrete breakage from the issue: on Cloud Run, the platform injects `X-Serverless-Authorization`, and Cloud Run's IAM frontend validates *that* in preference to `Authorization`. Since it's an `x-*` header it got forwarded, so the agent service validated the foreign platform token instead of the service-to-service bearer the server set on the `HttpAgent` → 401.

Fix is in the allowlist: `shouldForwardHeader` now drops the well-known hop-by-hop/platform headers (`x-forwarded-*`, `x-serverless-*`, `x-vercel-*`, `x-real-ip`, `x-cloud-trace-context`) while still forwarding `authorization` and ordinary `x-*` app headers. Kept it narrow on purpose — this doesn't touch the precedence/override behavior (client `x-*` headers can still override agent defaults, which #5712's existing test relies on), it just stops infra headers from leaking upstream.

Updated the existing `X-Forwarded-For` assertion (it was asserting the buggy behavior) and added tests for the dropped headers.

### DIFF
--- a/packages/runtime/src/v2/runtime/__tests__/header-utils.test.ts
+++ b/packages/runtime/src/v2/runtime/__tests__/header-utils.test.ts
@@ -16,7 +16,6 @@ describe("header-utils", () => {
       expect(shouldForwardHeader("x-custom")).toBe(true);
       expect(shouldForwardHeader("X-Custom")).toBe(true);
       expect(shouldForwardHeader("x-request-id")).toBe(true);
-      expect(shouldForwardHeader("X-Forwarded-For")).toBe(true);
     });
 
     it("blocks standard headers", () => {
@@ -27,6 +26,16 @@ describe("header-utils", () => {
       expect(shouldForwardHeader("accept")).toBe(false);
       expect(shouldForwardHeader("cookie")).toBe(false);
       expect(shouldForwardHeader("host")).toBe(false);
+    });
+
+    it("blocks hop-by-hop and platform headers from the inbound hop", () => {
+      expect(shouldForwardHeader("x-forwarded-for")).toBe(false);
+      expect(shouldForwardHeader("X-Forwarded-For")).toBe(false);
+      expect(shouldForwardHeader("x-forwarded-proto")).toBe(false);
+      expect(shouldForwardHeader("x-real-ip")).toBe(false);
+      expect(shouldForwardHeader("x-cloud-trace-context")).toBe(false);
+      expect(shouldForwardHeader("X-Serverless-Authorization")).toBe(false);
+      expect(shouldForwardHeader("x-vercel-id")).toBe(false);
     });
   });
 
@@ -49,6 +58,26 @@ describe("header-utils", () => {
         "x-custom": "custom-value",
         "x-request-id": "req-123",
         authorization: "Bearer token",
+      });
+    });
+
+    it("drops platform headers so they don't reach the upstream agent", () => {
+      const request = new Request("https://example.com", {
+        method: "POST",
+        headers: {
+          Authorization: "Bearer service-token",
+          "X-Serverless-Authorization": "Bearer platform-token",
+          "X-Forwarded-For": "203.0.113.1",
+          "X-Cloud-Trace-Context": "trace/span",
+          "X-Custom": "keep-me",
+        },
+      });
+
+      const result = extractForwardableHeaders(request);
+
+      expect(result).toEqual({
+        authorization: "Bearer service-token",
+        "x-custom": "keep-me",
       });
     });
 
--- a/packages/runtime/src/v2/runtime/handlers/header-utils.ts
+++ b/packages/runtime/src/v2/runtime/handlers/header-utils.ts
@@ -1,9 +1,27 @@
 /**
+ * Hop-by-hop and platform/proxy headers that describe the inbound
+ * client→runtime hop and must not leak to the upstream agent. They match the
+ * `x-*` allowlist but shouldn't cross the trust boundary — e.g. Cloud Run
+ * validates `x-serverless-authorization` in preference to `authorization`, so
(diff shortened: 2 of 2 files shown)


serhiizghama:
Closing — #5712 was already resolved by #5782 and #5783, so this is redundant now. Thanks!

PR #5799, opened 2026-07-02
feat(suggestions): stateless suggestion generation via /suggest (ENT-1018)

BenTaylorDev:
> **Update (SSE rework — addresses review):** `/suggest` now **streams AG-UI SSE** instead of buffering a JSON `{ messages }` response, and the run **forwards the consumer's messages + state**. Server-side it reuses `createSseEventResponse` (the runner's event pipeline minus `GLOBAL_STORE` persistence, `captureTelemetry:false`); client-side it drives a stock `HttpAgent` at `/agent/:id/suggest`, so chips fill in progressively via `onMessagesChanged` and the run never touches the Intelligence websocket delegate. The stateless and fallback paths now share one `runAgent` flow (net −68 LOC of production code). The `## How` / behavioral notes below are updated to match.

## What & why

Dynamic suggestions were implemented as a **full shadow-thread agent run** (`SuggestionEngine.generateSuggestions` clones the provider agent onto a throwaway `threadId` and calls `runAgent`). Against a thread-persisting backend (CopilotKit Intelligence) every reload materialized a **listed, auto-named thread** (plus gateway events + a run lock), flooding the thread drawer. Root cause is the design: a suggestion is a **stateless structured completion**, not a persisted conversation.

This makes it one. A suggestion now runs the provider agent **directly** behind a dedicated `POST /agent/:agentId/suggest` handler — no thread, lock, gateway, runner-store, or name-gen — for **all** CopilotKit-runtime backends. The debris is structurally impossible rather than hidden.

## How

- **`get-runtime-info`** advertises a `suggestions` capability on `GET /info` (`RuntimeInfo.suggestions`).
- **`handleSuggestAgent`** (new) runs the resolved provider agent directly and **streams its AG-UI events as SSE** via `createSseEventResponse` — the runner's event pipeline (`agent.runAgent({ onEvent })` + `finalizeRunEvents`) **minus** the `GLOBAL_STORE` persistence that backs SSE-mode thread listings. `captureTelemetry:false` and no debug bus keep suggestions out of run telemetry/the inspector. Header-forwarding only; no middleware.
- **Router**: `POST /agent/:agentId/suggest` wired through `fetch-router`/`fetch-handler` (multi- and single-route), with `assertNever` exhaustiveness guards.
- **core `SuggestionEngine`**: when the runtime advertises `suggestions` (and transport isn't single-route), it seeds a stock `HttpAgent` (pointed at `/agent/:id/suggest`, credentials via a fetch wrapper) with the consumer's deep-cloned messages + state and `runAgent`s it — a plain `HttpAgent` only speaks REST SSE, so it never routes through the Intelligence websocket delegate. Otherwise it falls back to the clone+`runAgent` path. Both paths share one flow; abort-aware; failures logged.

## Testing

Unit + integration across both packages (**504 core + 1587 runtime green**, `check-types` clean): the **no-thread-leak** proof (real `InMemoryAgentRunner`, asserts the suggest thread never appears in the live listing), progressive-streaming + state-forwarding assertions, cross-mode SSE handler behavior, error/abort robustness (Safari `DOMException` + undici non-`AbortError`), header/credential forwarding, multi-route `200` + `GET→405`, and single-route/`suggestions:false` fallback.

Validated end-to-end against a running runtime + real `CopilotKitCore` client over HTTP (local built packages):
- **In-memory managed threads:** a normal `/run` creates exactly 1 listed thread; **4 dynamic-suggestion reloads add 0**. Chips stream progressively (`onSuggestionsChanged` 0→1→2); the `/suggest` request carries the consumer's state.
- **Live hosted CopilotKit Intelligence** (`mode: "intelligence"`, real gateway): `/info` advertises the capability, suggestions stream via the stateless path, and the **hosted thread listing is byte-identical (50→50) after 6 reloads** — the drawer's actual data source is untouched.

## Compatibility

**Migrations:** none. No schema/DB change; self-hosted deployments need no coordinated update — the change is entirely in `@copilotkit/core` + `@copilotkit/runtime`. No release flag.

**Additive + capability-gated — version skew is safe both directions:**
- **New core + old runtime** (no `/suggest`, capability absent): `core.suggestions` is `undefined` → client uses today's clone+`runAgent` path.
- **Old core + new runtime**: the new optional `RuntimeInfo.suggestions` field is ignored; the `/suggest` endpoint sits unused.
- **New + new**: stateless `/suggest`.

`RuntimeInfo.suggestions` is a new **optional** field and the `/agent/:agentId/suggest` route (+ `RouteInfo`/`METHOD_NAMES` entries) are purely additive. No exported symbol was removed or renamed.

**Behavioral notes (matched new versions):**
- Dynamic suggestions **stream progressively** over SSE (chips fill in as the provider emits them), matching the pre-`/suggest` UX.
- Suggestion runs no longer flow through the run handler, so they create **no thread** (the fix) and won't appear in `oss.runtime.*` run telemetry.
- Single-route transport and non-CopilotKit AG-UI backends keep the clone+`runAgent` fallback.

**⚠️ Deploy note:** a reverse proxy that path-allowlists CopilotKit routes must add `POST /agent/:agentId/suggest`. If the capability is advertised but that path is blocked, suggestions silently no-op — the client does **not** fall back on a `/suggest` error (by design, to avoid re-introducing the thread flood).

## Review

Landed through a full CR loop (3 rounds / 21 agent-reviews + a bucket-(c) promotion audit) — bucket (a) converged to zero. Known pre-existing follow-ups tracked separately (out of scope): `resolveLicenseStatus` grace-period ordering and `AgentRegistry.getAgent` using `in` vs `hasOwnProperty` for reserved ids.


mme (changes requested):
1. No streaming — server buffers and returns one JSON `{ messages }`; chips appear all at once (`handle-suggest.ts:118-139`).
2. No state — client sends `state: {}` on the stateless path (`suggestion-engine.ts:224`), vs. the old path which deep-cloned the consumer's state.

BenTaylorDev:
Both addressed in 9f938e8 — reworked to reuse the existing AG-UI SSE path rather than a buffered response.

**1. Streaming.** `/suggest` now streams AG-UI SSE via `createSseEventResponse` (the runner's event pipeline minus the `GLOBAL_STORE` persistence, `captureTelemetry:false`). The client drives a stock `HttpAgent` at `/agent/:id/suggest`, so chips fill in progressively via `onMessagesChanged` — same wire format as `/run`. A plain `HttpAgent` only speaks REST SSE, so the run still never touches the runner or the Intelligence websocket delegate → no thread persistence.

**2. State.** The run now forwards the consumer's deep-cloned `messages` + `state` (was `state: {}`), matching the clone fallback. The stateless and fallback paths now share one `runAgent` flow (net −68 LOC of production code).

Validated end-to-end against a running runtime + real `CopilotKitCore` client over HTTP:
- **In-memory managed threads:** a normal `/run` → 1 listed thread; **4 dynamic-suggestion reloads → 0 added**. Chips stream (`onSuggestionsChanged` 0→1→2); `/suggest` carries the consumer state.
- **Live hosted Intelligence** (`mode: "intelligence"`): `/info` advertises the capability, suggestions stream via the stateless path, and the **hosted thread listing is byte-identical (50→50) after 6 reloads** — the drawer's actual data source is untouched.

Test counts updated to 504 core + 1587 runtime green.

mme (approved):

PR #5804, opened 2026-07-02
fix(web-inspector): collapse timeline event details

marthakelly:
## Summary
- Collapse structured timeline event details by default in `<cpk-thread-inspector>`.
- Add an accessible `Show details` / `Hide details` toggle with `aria-expanded` for timeline JSON payloads.
- Keep timeline row headers and source-event links visible while leaving plain message bodies unchanged.

## Test Plan
- `NX_TUI=false ./node_modules/.bin/nx run @copilotkit/web-inspector:test -- --run src/__tests__/web-inspector.spec.ts -t "collapses structured timeline event details"`
- `PNPM_CONFIG_DANGEROUSLY_ALLOW_ALL_BUILDS=true NX_TUI=false ./node_modules/.bin/nx run @copilotkit/web-inspector:test`
- `PNPM_CONFIG_DANGEROUSLY_ALLOW_ALL_BUILDS=true NX_TUI=false ./node_modules/.bin/nx run @copilotkit/web-inspector:check-types`
- `PNPM_CONFIG_DANGEROUSLY_ALLOW_ALL_BUILDS=true NX_TUI=false ./node_modules/.bin/nx run @copilotkit/web-inspector:build`
- `./node_modules/.bin/oxfmt --check packages/web-inspector/src/index.ts packages/web-inspector/src/__tests__/web-inspector.spec.ts`
- `./node_modules/.bin/oxlint packages/web-inspector/src/index.ts packages/web-inspector/src/__tests__/web-inspector.spec.ts` (warnings only, pre-existing in touched files)

## Notes
- Local pnpm 11 ignores the repo's legacy `package.json#pnpm.overrides` config, so local installs needed the existing overrides represented during install to avoid unrelated React 18/19 mismatches in downstream package tests. No lockfile or workspace config changes are included in this PR.

### DIFF
--- a/packages/web-inspector/src/__tests__/web-inspector.spec.ts
+++ b/packages/web-inspector/src/__tests__/web-inspector.spec.ts
@@ -1119,12 +1119,17 @@ describe("CpkThreadInspector provider contract", () => {
 
     expect(internals.activeTimelineItems).toHaveLength(1);
     expect(el.shadowRoot?.textContent ?? "").toContain("THREAD_STATE_WRITTEN");
-    expect(el.shadowRoot?.textContent ?? "").toContain("Show details");
+    expect(el.shadowRoot?.textContent ?? "").toContain("Details");
     expect(el.shadowRoot?.textContent ?? "").not.toContain("checkpointId");
     expect(el.shadowRoot?.textContent ?? "").toContain("Source event #1");
     expect(el.shadowRoot?.textContent ?? "").not.toContain(
       "No timeline events captured",
     );
+    expect(
+      el.shadowRoot
+        ?.querySelector<HTMLButtonElement>(".cpk-td__timeline-details-toggle")
+        ?.getAttribute("aria-label"),
+    ).toBe("Show event details");
 
     el.shadowRoot
       ?.querySelector<HTMLButtonElement>(".cpk-td__timeline-details-toggle")
@@ -1160,10 +1165,15 @@ describe("CpkThreadInspector provider contract", () => {
     await flushProviderWork(el);
 
     expect(el.shadowRoot?.textContent ?? "").toContain("Run started");
-    expect(el.shadowRoot?.textContent ?? "").toContain("Show details");
+    expect(el.shadowRoot?.textContent ?? "").toContain("Details");
     expect(el.shadowRoot?.textContent ?? "").not.toContain(
       "very chonky run-started payload",
     );
+    expect(
+      el.shadowRoot
+        ?.querySelector<HTMLButtonElement>(".cpk-td__timeline-details-toggle")
+        ?.getAttribute("aria-label"),
+    ).toBe("Show event details");
 
     el.shadowRoot
       ?.querySelector<HTMLButtonElement>(".cpk-td__timeline-details-toggle")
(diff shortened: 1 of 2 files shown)


MikeRyanDev (changes requested):
Thanks for tightening up the inspector output. I found one blocking regression in the new details toggle path:\n\n- packages/web-inspector/src/index.ts:2899: timeline rows that have both `item.body` and `item.details` render the new Show/Hide details button, but expanding never shows the structured details because the render branch exits on `item.body` first. This affects real `RUN_ERROR` / `ERROR` rows: clicking Show details changes the label to Hide details while still only showing the summary message, so error payload fields such as codes, stack/context, or other debug metadata remain inaccessible from the timeline. Please render the body and expanded details independently, or only show the toggle when it can reveal the JSON payload. A focused test with an error event containing both `message` and another payload field would catch this.

BenTaylorDev (commented):
### 🔴 High — bulk "Expand all / Collapse all" controls are untested
`renderActiveBulkControls`, `renderRawEventBulkControls`, `expand/collapseTimelineDetails`, `expand/collapseRawEventDetails`, and the `?disabled=${allExpanded/allCollapsed}` states are ~90 lines with no test coverage — the new tests only exercise the per-item toggle and collapse-by-default. Would like: expand-all reveals all details, collapse-all hides them, disabled states flip correctly, and the `<= 1` "no bulk controls" branch.

### 🟠 Medium — dead "Show details" toggle on error events
`RUN_ERROR` / `ERROR` items are built with **both** `body` and `details` (`index.ts:2211-2222`), but the render precedence is `item.body ? <body> : item.details ? <pre> : nothing` — when a body exists it always wins and the details `<pre>` never renders. The new toggle is gated only on `item.details`, so an error *with a message* (nearly all of them) gets a "Show details" button that flips its label/`aria-expanded` but reveals nothing — exactly the case a debugger user cares about. Suggest gating the toggle on `item.details && !item.body`, or restructuring so an item can show body *and* expandable details.

### 🟠 Medium — undocumented scope + UX: the Raw AG-UI Events tab now collapses too
Summary/Test Plan describe only the timeline, but the PR also collapses the **Raw AG-UI Events** payloads by default (`_expandedRawEvents` + the raw-events render path). That's a behavior change to a second tab whose whole purpose is showing raw payloads (now every one needs a click or "Expand all"). Worth calling out in the description and sanity-checking that collapsed-by-default is actually desired for the *raw* tab — timeline collapsing is clearly right; raw-events is more debatable.

### 🟡 Low / nits
- **`rawEventId` collision risk** — `raw-event-${event.sourceIndex ?? event.timestamp ?? event.type}`: when `sourceIndex` is absent, the fallback to `timestamp`/`type` collides easily (events share types), linking their toggles. Folding in the array index would make it collision-proof.
- **a11y** — buttons set `aria-expanded` but no `aria-controls`, and collapsed content is removed from the DOM (`nothing`). Adding an `id` on the `<pre>` + `aria-controls` completes the linkage.
- **Duplication** — the ~20-line chevron-SVG + label button is copy-pasted between the timeline and raw-event paths, and the raw-event button reuses the `cpk-td__timeline-details-toggle` class. A shared `renderDetailsToggle(expanded, onClick)` helper would DRY both and fix the class-name mismatch.
- **Perf (note only)** — because the `Set` is in the cache key, each toggle busts the whole-panel cache and re-renders + re-highlights every item. Fine at these sizes for a debug inspector; just noting.

MikeRyanDev (approved):
Fixed the blocking details-toggle regression and confirmed the updated commit is green in CI.

BenTaylorDev (approved):

PR #5805, opened 2026-07-02
fix(react-ui): reduce sidebar CSS specificity

### LINKED ISSUE #263: When overriding CSS properties, sometimes the overridden css property does not take precedence

godququ5-code:
## Summary
- lowers CopilotSidebar-scoped CSS selector specificity with `:where(.copilotKitSidebar)`
- keeps the existing sidebar styles and layout behavior unchanged
- adds a regression test to keep sidebar selectors easy for app CSS to override

## Root cause
Sidebar-specific selectors such as `.copilotKitSidebar .copilotKitWindow` were more specific than ordinary user overrides like `.copilotKitWindow`, so user CSS could lose precedence in Chrome.

Fixes #263.

## Validation
- `pnpm -C packages/react-ui test`
- `pnpm exec oxfmt --check packages/react-ui/src/css/sidebar-specificity.test.ts packages/react-ui/src/css/window.css packages/react-ui/src/css/header.css packages/react-ui/src/css/input.css`
- `pnpm exec oxlint packages/react-ui/src/css/sidebar-specificity.test.ts`

### DIFF
--- a/packages/react-ui/src/css/header.css
+++ b/packages/react-ui/src/css/header.css
@@ -14,7 +14,7 @@
   z-index: 2;
 }
 
-.copilotKitSidebar .copilotKitHeader {
+:where(.copilotKitSidebar) .copilotKitHeader {
   border-radius: 0;
 }
 
--- a/packages/react-ui/src/css/header.css
+++ b/packages/react-ui/src/css/header.css
@@ -34,6 +34,10 @@
     border-top-left-radius: 8px;
     border-top-right-radius: 8px;
   }
+
+  :where(.copilotKitSidebar) .copilotKitHeader {
+    border-radius: 0;
+  }
 }
 
 .copilotKitHeader > button {
--- a/packages/react-ui/src/css/input.css
+++ b/packages/react-ui/src/css/input.css
@@ -19,7 +19,7 @@
   border-bottom-right-radius: 0.75rem;
 }
 
-.copilotKitSidebar .copilotKitInputContainer {
+:where(.copilotKitSidebar) .copilotKitInputContainer {
   border-bottom-left-radius: 0;
   border-bottom-right-radius: 0;
 }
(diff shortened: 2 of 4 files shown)


BenTaylorDev (changes requested):
Thanks — the approach is right and the goal is worth doing: wrapping the qualifier in `:where(.copilotKitSidebar)` drops each rule from (0,2,0) to (0,1,0) so a consumer's single-class override can win, and for the non-media rules the sidebar variant is authored after its base, so the source-order tie still favors the sidebar (defaults preserved). Nicely, `window.css` stays correct because it has a dedicated `@media` sidebar variant that comes after the `@media` base.

One regression to fix before merge, in `header.css`. There's no media-scoped sidebar override there, so dropping the specificity creates a tie the later library rule wins at desktop widths:

- Sidebar variant (`header.css:17`): `:where(.copilotKitSidebar) .copilotKitHeader { border-radius: 0 }` — now (0,1,0).
- `@media (min-width: 640px)` base `.copilotKitHeader` (`:31`): `border-top-left-radius: 8px; border-top-right-radius: 8px` — also (0,1,0), authored later.

Previously the sidebar variant was (0,2,0) and won per-longhand, so the sidebar header stayed square at all widths. After this change it's a tie, and the later `@media` base wins → at ≥640px the sidebar header's top corners revert from square to rounded 8px, contrary to the variant's explicit `border-radius: 0`. (It's desktop-only — the media rule doesn't apply below 640px.)

Small fix — mirror what `window.css` already does and add a media-scoped sidebar override inside the `@media (min-width: 640px)` block:

```css
:where(.copilotKitSidebar) .copilotKitHeader {
  border-radius: 0;
}
```

(or have the non-media sidebar variant set the longhands `border-top-left-radius`/`border-top-right-radius: 0` so it isn't contested at that breakpoint). With that added, this looks good — the `:where()` idea is sound and complete across the other inflated selectors. A quick Chrome check at ≥640px to confirm the sidebar header stays square would be worth it too.

BenTaylorDev (approved):
The re-assert at `≥640px` is exactly right — the media-scoped `:where(.copilotKitSidebar) .copilotKitHeader { border-radius: 0 }` is authored after the base `.copilotKitHeader` radius rule, so it wins the same-specificity source-order tie and the sidebar header stays square at desktop widths. That closes the only regression; the `:where()` specificity reduction is sound and complete across the other selectors. Thanks for the quick turnaround!

PR #5807, opened 2026-07-02
fix(python-sdk): fold App Context into system prompt

### LINKED ISSUE #5610: 🐛 Bug: LangGraph CopilotKitMiddleware injects App Context as a separate, non-consecutive SystemMessage (breaks Anthropic / langchain-anthropic)
### ♻️ Reproduction Steps

1. Use the Python SDK LangGraph integration: build an agent with `langchain.agents.create_agent(...)`, passing a system prompt (so the model request carries a head system message), and attach CopilotKitMiddleware.
2. Use a `langchain-anthropic` chat model (ChatAnthropic / Claude).
3. Send any App Context via `copilotkit.context` (or runtime.context), e.g. `useCopilotReadable` on the frontend or RunAgentInput.context.
4. Have at least one prior human/assistant turn in the thread so the conversation is not "system-only".
5. Run the agent.

### ✅ Expected Behavior

App Context should be merged into the single existing system prompt (one `system` block on the wire), so providers that require system messages to be first/consecutive — Anthropic in particular — accept the request.

### ❌ Actual Behavior

`before_agent` in `sdk-python/copilotkit/copilotkit_lg_middleware.py` builds the `"App Context:"` `SystemMessage` and inserts it *positionally* after the first system message, then returns `{"messages": updated_messages}`.

LangGraph feeds that return value through the `add_messages` reducer, which is **append/merge-by-ID**, not positional. The freshly built `SystemMessage` has a brand-new UUID that matches no existing message, so the reducer **appends it to the end** of `state["messages"]` — the positional insert is silently discarded.

The agent's own system prompt sits at the head, the App Context system message lands at the tail, and any human/AI/tool messages sit between them. That is **two non-consecutive system messages**. `langchain-anthropic`'s `_format_messages` then raises:

```
ValueError: Received multiple non-consecutive system messages.
```

**Root cause in code** (first-injection path: `existing_context_index == -1` → fresh-UUID `SystemMessage`):
- `v1.59.3` / commit `b2f10f6753`: `before_agent` ~L593
- `main` as of 2026-06-22: `before_agent` ~L715, positional insert ~L820–830

**Suggested fix:** fold App Context into `request.system_message` inside `wrap_model_call` (append to the existing `system_message` string), instead of inserting a `SystemMessage` into `state["messages"]` from `before_agent`. That bypasses the `add_messages` reducer entirely and guarantees a single system block.

### Screenshots

_No response_

### 𝌚 CopilotKit Version

```shell
Python SDK `copilotkit` (LangGraph integration).
Reproduced against copilotkit_lg_middleware.py shipped with @copilotkit/runtime@1.59.3
(commit b2f10f6753) and confirmed same code path present on `main` as of 2026-06-22.
Model: langchain-anthropic (ChatAnthropic / Claude).
```

### 📄 Logs (Optional)

```shell
ValueError: Received multiple non-consecutive system messages.
  (raised in langchain_anthropic ... _format_messages)
```

godququ5-code:
## Summary
- folds CopilotKit App Context into `ModelRequest.system_message` during LangGraph model calls
- stops injecting App Context as a separate `SystemMessage` from `before_agent`, avoiding non-consecutive system messages after LangGraph reducers run
- updates middleware regression tests for state context, runtime context, forwarded-header filtering, and message-history no-op behavior

## Root cause
`before_agent` inserted a new App Context `SystemMessage` into `state["messages"]`, but the LangGraph message reducer can append new message IDs rather than preserving positional insertion. With an existing system prompt and prior conversation turns, this can produce multiple non-consecutive system messages, which `langchain-anthropic` rejects.

Fixes #5610.

## Validation
- `uv run pytest -q`

### DIFF
--- a/sdk-python/copilotkit/copilotkit_lg_middleware.py
+++ b/sdk-python/copilotkit/copilotkit_lg_middleware.py
@@ -338,6 +338,61 @@ class CopilotKitMiddleware(AgentMiddleware[StateSchema, Any]):
             system_message=SystemMessage(content=f"{base}\n\n{note}")
         )
 
+    def _build_app_context_note(
+        self,
+        state: dict[str, Any],
+        runtime_context: Any = None,
+    ) -> str | None:
+        copilotkit_state = state.get("copilotkit", {})
+        app_context = copilotkit_state.get("context") or runtime_context
+
+        if isinstance(app_context, dict):
+            app_context = {
+                k: v
+                for k, v in app_context.items()
+                if k != "copilotkit_forwarded_headers"
+            }
+
+        if not app_context:
+            return None
+        if isinstance(app_context, str) and app_context.strip() == "":
+            return None
+        if isinstance(app_context, dict) and len(app_context) == 0:
+            return None
+
+        if isinstance(app_context, str):
+            context_content = app_context
+        else:
+            if hasattr(app_context, "model_dump"):
+                app_context = app_context.model_dump()
+            elif isinstance(app_context, list):
+                app_context = [
+                    item.model_dump() if hasattr(item, "model_dump") else item
+                    for item in app_context
+                ]
+            context_content = json.dumps(app_context, indent=2)
+
+        return f"App Context:\n{context_content}"
+
+    def _apply_app_context_note(self, request: ModelRequest) -> ModelRequest:
+        note = self._build_app_context_note(
+            request.state or {},
+            getattr(request.runtime, "context", None),
+        )
+        if not note:
+            return request
+        existing = request.system_message
+        if existing is None:
+            return request.override(system_message=SystemMessage(content=note))
+        base = (
+            existing.content
+            if isinstance(existing.content, str)
+            else str(existing.content)
+        )
+        return request.override(
+            system_message=SystemMessage(content=f"{base}\n\n{note}")
+        )
+
     # ------------------------------------------------------------------
     # Auto-A2UI tool injection
     # ------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


BenTaylorDev (approved):
**Approving.** Adversarially reviewed against current `main`:

- Correct root cause & fix: `main`'s `before_agent` injected App Context as a separate `SystemMessage` into `state["messages"]`, which the `add_messages` reducer appends to the end → the non-consecutive system message that breaks `langchain-anthropic` (#5610). This folds App Context into `request.system_message` at model-call time instead, mirroring the existing `_apply_state_note`/`_build_state_note` pattern already on `main` (`:281/:325`), applied right after it (`:538-539`, `:742-743`) so system content stays consecutive.
- **Complete:** applied in both `wrap_model_call` and `awrap_model_call`; `before_agent` → `return None` with `abefore_agent` delegating to it, so both sync and async paths are neutralized.
- **No regression:** the only remaining `App Context:` reference in non-test code is the writer (`:375`) — nothing reads that prefix from message history anymore.
- **No drift** despite the branch age — the target file's base is byte-identical to current `main`. Mechanism is genuinely tested: `_run_wrap` invokes the real `wrap_model_call` and asserts on the transformed `system_message` (only `request.runtime` is mocked, which is the fallback input, not the folding logic).

One minor note for the record: the `request.runtime.context` fallback is only mock-tested; the common `state["copilotkit"]["context"]` path is real and covered. CI is green (32 checks pass, 5 skipped).

PR #5810, opened 2026-07-03
fix(runtime): route proxied MCP reads over SSE

maxkorp:
## Summary
- route proxied MCP Apps resource reads over the HTTP/SSE path instead of the Intelligence delegate in `ProxiedCopilotRuntimeAgent`
- only bypass the Intelligence run handler for proxy-shaped MCP Apps requests that match configured MCP Apps servers for the target agent
- reject malformed or unconfigured proxied MCP attempts with JSON errors instead of falling through to local agent runs
- add core and runtime regressions covering the Excalidraw-style `resources/read` flow and the unconfigured-proxy rejection path

## Testing
- `NX_TUI=false fnm exec --using 22 pnpm nx run @copilotkit/core:test -- src/__tests__/intelligence-agent.test.ts`
- `NX_TUI=false fnm exec --using 22 pnpm nx run @copilotkit/runtime:test -- src/v2/runtime/__tests__/handle-run.test.ts`
- `NX_TUI=false fnm exec --using 22 pnpm nx run @copilotkit/core:check-types`
- `NX_TUI=false fnm exec --using 22 pnpm nx run @copilotkit/runtime:check-types`
- `NX_TUI=false fnm exec --using 22 pnpm nx run @copilotkit/core:build`
- `NX_TUI=false fnm exec --using 22 pnpm nx run @copilotkit/runtime:build`
- `fnm exec --using 22 pnpm oxlint packages/core/src/agent.ts packages/core/src/__tests__/intelligence-agent.test.ts packages/runtime/src/v2/runtime/handlers/handle-run.ts packages/runtime/src/v2/runtime/__tests__/handle-run.test.ts`
- `fnm exec --using 22 pnpm oxfmt --check packages/core/src/agent.ts packages/core/src/__tests__/intelligence-agent.test.ts packages/runtime/src/v2/runtime/handlers/handle-run.ts packages/runtime/src/v2/runtime/__tests__/handle-run.test.ts`
- pre-commit hook: `test, publint, attw` for affected packages passed; Nx reported `@copilotkit/sqlite-runner:test` as flaky after it passed

### DIFF
--- a/packages/core/src/__tests__/intelligence-agent.test.ts
+++ b/packages/core/src/__tests__/intelligence-agent.test.ts
@@ -1847,6 +1847,74 @@ describe("IntelligenceAgent", () => {
 });
 
 describe("ProxiedCopilotRuntimeAgent (intelligence mode)", () => {
+  it("runs proxied MCP requests over HTTP instead of the intelligence delegate", async () => {
+    const sseBody = [
+      `data: ${JSON.stringify({
+        type: EventType.RUN_STARTED,
+        threadId: "thread-1",
+        runId: "run-1",
+      })}`,
+      "",
+      `data: ${JSON.stringify({
+        type: EventType.RUN_FINISHED,
+        threadId: "thread-1",
+        runId: "run-1",
+      })}`,
+      "",
+    ].join("\n");
+
+    mockFetch.mockResolvedValueOnce(
+      new Response(sseBody, {
+        status: 200,
+        headers: { "Content-Type": "text/event-stream" },
+      }),
+    );
+
+    const agent = new ProxiedCopilotRuntimeAgent({
+      runtimeUrl: "http://localhost:4000/api/copilotkit",
+      agentId: "default",
+      runtimeMode: RUNTIME_MODE_INTELLIGENCE,
+      intelligence: { wsUrl: "ws://localhost:4401/client" },
+    });
+
+    const events: BaseEvent[] = [];
+    await new Promise<void>((resolve, reject) => {
+      agent
+        .run({
+          ...defaultInput,
+          forwardedProps: {
+            __proxiedMCPRequest: {
+              method: "resources/read",
+              serverId: "example_mcp_app",
+              serverHash: "server-hash",
+              params: { uri: "ui://excalidraw/mcp-app.html" },
+            },
+          },
+        })
+        .subscribe({
+          next: (event) => events.push(event),
+          complete: resolve,
+          error: reject,
+        });
+    });
+
+    expect(mockFetch).toHaveBeenCalledTimes(1);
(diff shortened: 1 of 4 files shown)


MikeRyanDev (changes requested):
Blocking issue: `handleRunAgent` now treats the mere presence of `forwardedProps.__proxiedMCPRequest` as enough to skip `handleIntelligenceRun` and swap in `InMemoryAgentRunner` (`packages/runtime/src/v2/runtime/handlers/handle-run.ts:14-18`, `:86-99`). That field comes from the request body, and the new regression test even sets `mcpApps: undefined` (`packages/runtime/src/v2/runtime/__tests__/handle-run.test.ts:214-227`), so an Intelligence runtime with no MCP Apps middleware can be forced into a local SSE agent run just by sending this forwarded prop. That bypasses the normal Intelligence path: thread creation/locking, canonical run ids, realtime credentials, and persistence. Please gate this bypass on a validated MCP Apps proxy request that the runtime can actually service (for example, non-null/object payload plus configured/matching `runtime.mcpApps` server), and return an error for malformed/unserviceable proxy requests instead of falling through to a normal local agent run. The runtime test should also configure MCP Apps so it proves the real middleware proxy path, not just an arbitrary SSE fallback.

maxkorp:
Yes, after digging more the most precise invariant is "bypass Intelligence if and only if MCPAppsMiddleware would intercept this run before the downstream agent is invoked". Fix incoming

maxkorp:
Addressed in `3361e20c2`.

Runtime now only uses the local SSE runner for Intelligence-mode MCP Apps proxy requests when:
- `__proxiedMCPRequest` is a proxy-shaped object with a string `method`
- the request includes a string `serverId` or `serverHash`
- the target agent has an applicable configured `runtime.mcpApps` server matching that selector

Malformed or unconfigured proxy attempts now return JSON `400` responses instead of falling through to a local agent run. The runtime regression now configures a real MCP Apps server and verifies the middleware-backed `resources/read` path, plus a negative test for the previous `mcpApps: undefined` hole.

MikeRyanDev (changes requested):
`3361e20c` fixes my original bypass finding: an Intelligence request now uses the local MCP Apps path only when the proxy payload is structured and matches a server configured for that agent; unconfigured requests return `400`, and the test exercises the real middleware path.

The PR is still not merge-ready because it conflicts with current `main` in `packages/core/src/agent.ts`. Resolve that conflict and rerun the full checks before another review. The current ADK, Mastra, and PydanticAI failures are connection/UI timeouts and may be unrelated, but they are from this stale pre-rebase head.

The exact-head focused runtime suite passes 37 tests, and runtime/shared type checks pass.

maxkorp:
Closing as no longer relevant.

PR #5817, opened 2026-07-04
fix(shared): don't append cleanup events after a terminal event

### LINKED ISSUE #5812: 🐛 Bug: `finalizeRunEvents` appends `TEXT_MESSAGE_END` after `RUN_ERROR` — stop button crashes with "run has already errored"
### ♻️ Reproduction Steps

1. Run the standard Node-proxy setup: a `CopilotRuntime` with an `HttpAgent` pointing at a remote AG-UI endpoint, served via `copilotRuntimeNodeHttpEndpoint`. Frontend uses `CopilotChat`.
2. Start a chat and send a message that produces a streamed assistant response of a few seconds.
3. While the assistant is still streaming text (i.e. between `TEXT_MESSAGE_START` and `TEXT_MESSAGE_END`), press the **Stop** button.
4. Observe the console error (below). The assistant message is left in a broken state and the chat becomes unreliable.

Minimal server setup that reproduces:

```js
const { CopilotRuntime, EmptyAdapter, copilotRuntimeNodeHttpEndpoint } = require('@copilotkit/runtime');
const { HttpAgent } = require('@ag-ui/client');

  agents: { my_agent: new HttpAgent({ url: 'http://localhost:8000/api' }) },
});
const handler = copilotRuntimeNodeHttpEndpoint({ runtime, serviceAdapter: new EmptyAdapter(), endpoint: '/copilotkit' });
```

…where `http://localhost:8000/api` is any AG-UI-compliant agent that streams a `TEXT_MESSAGE_START` → `TEXT_MESSAGE_CONTENT`… response and does not close the message before the run is aborted (e.g. pydantic-ai's `AGUIAdapter` — see pydantic/pydantic-ai#3108).

### ✅ Expected Behavior

Pressing Stop should end the run cleanly: the open text message is closed and a single terminal event (`RUN_FINISHED`) is emitted. No console error. The chat remains usable for the next message.

Per the AG-UI spec invariant: **at most one terminal event per run, and all sub-events (TEXT_MESSAGE_END, TOOL_CALL_END) must close *before* the terminal.**

### ❌ Actual Behavior

The browser throws:
```
Agent execution failed: Cannot send event type 'TEXT_MESSAGE_END':
The run has already errored with 'RUN_ERROR'. No further events can be sent.
```

### Screenshots

_No response_

### 𝌚 CopilotKit Version

```shell
@copilotkit/react-core 1.62.2
@copilotkit/react-ui 1.62.2
@copilotkit/runtime 1.62.2
```

### 📄 Logs (Optional)

```shell
Error message in the browser:

Agent execution failed: D: Cannot send event type 'TEXT_MESSAGE_END': The run has already errored with 'RUN_ERROR'. No further events can be sent.


Example data in response

...

data: {"type":"TEXT_MESSAGE_CONTENT","timestamp":1783057434585,"messageId":"225f30a2-c662-4d0d-a05e-789cb51e17cc","delta":" frames"}

data: {"type":"RUN_ERROR","message":"This operation was aborted","code":"abort","rawEvent":{}}

data: {"type":"TEXT_MESSAGE_END","messageId":"225f30a2-c662-4d0d-a05e-789cb51e17cc"}
```

serhiizghama:
Closes #5812. Pressing Stop while an agent is still streaming text throws `Cannot send event type 'TEXT_MESSAGE_END': The run has already errored with 'RUN_ERROR'` in the browser and leaves the chat unusable.

`finalizeRunEvents` closed any open message/tool call unconditionally. When the agent already emitted a terminal event (an abort-time `RUN_ERROR` from e.g. pydantic-ai's `AGUIAdapter`, or a normal `RUN_FINISHED`) that terminal has already been dispatched to the client by the time finalize runs, so the appended `TEXT_MESSAGE_END`/`TOOL_CALL_END` land after it and the AG-UI client rejects them. Now finalize bails out as soon as a terminal event is present — the synthetic cleanup only exists to repair streams that ended without a terminal, so there's nothing to do once the run has terminated.

Added a regression for the stop-mid-message case and updated the existing terminal-present tests, which were asserting the old append-after-terminal behavior.

### DIFF
--- a/packages/runtime/src/v2/runtime/runner/__tests__/finalize-events.test.ts
+++ b/packages/runtime/src/v2/runtime/runner/__tests__/finalize-events.test.ts
@@ -82,7 +82,7 @@ describe("finalizeRunEvents", () => {
     expect(errorEvent?.message).toContain("terminal event");
   });
 
-  it("only appends structural fixes when a terminal event already exists", () => {
+  it("appends nothing once a terminal event already exists", () => {
     const events: BaseEvent[] = [
       createTextStart("msg-1"),
       createToolStart("tool-1"),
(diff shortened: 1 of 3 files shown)


BenTaylorDev:
Closing as superseded — merged #5885 already stops `finalizeRunEvents` from appending cleanup events after a terminal event (`packages/shared/src/finalize-events.ts` early-returns when a terminal event is present). Issue #5812 is closed as completed. Thanks @serhiizghama!

PR #5823, opened 2026-07-06
feat(web-components,react,vue,angular): CopilotThreadsDrawer UX redesign (ENT-1051)

BenTaylorDev:
Redesigns the shared `<copilotkit-threads-drawer>` element (`@copilotkit/web-components`) to the new Figma UX, keeping the React, Vue, and Angular wrappers in lockstep. Pure-VIEW change — no `@copilotkit/core`, runtime, or `useThreads` changes.

**Ticket:** [ENT-1051](https://linear.app/copilotkit/issue/ENT-1051) · **Figma:** [Thread Drawer](https://www.figma.com/design/feSsBJw1qCfLp0JNnOurrJ/CopilotKit-Intelligence?node-id=723-78) · **Spec:** [Notion](https://app.notion.com/p/3953aa381852819ab464dae3894e7f18)

## What changed
- **Header** → right-aligned icon row. On desktop it holds the **collapse** toggle (sidebar glyph); on mobile the **close** toggle. No title text, no "+ New" pill. Optional `slot="header"` preserved (empty by default; the toggle right-aligns after it).
- **New Conversation** row (plus-square + label) below the header — keeps `part="new-thread-button"` + the `new-thread` event.
- **Recent Conversations** heading + **funnel** filter icon → Active/All popover. Preserves `_filter` + `filter-change` and `part="filter-active"`/`filter-all"`.
- **Per-row kebab menu** (⋮) holding Archive/Unarchive + Delete — preserves those events + parts. An open kebab now shields the rest of the list from hover so it reads as a single surface (see Review fixes).
- **Delete confirm** is a native `<dialog>` opened with `showModal()` (browser top layer), centered over the drawer's visible box — it can't paint under other UI or drop below the fold. jsdom falls back to the `open` attribute.
- **Archived rows** render italic/muted inline in the "All" view.
- **Desktop collapse** → `collapsed` / `collapsible` props (default **expanded**) + a `collapse-change` event / `CollapseChangeDetail`. Collapsing sets `--cpk-drawer-reserved-width: 0` on the document root so the host grid reclaims the column with no hydration flicker.
- **Unified floating cluster** (Figma "closed" mockup) = `[sidebar-toggle] [+ New Conversation]`, shown in both the mobile-closed and desktop-collapsed states. Mobile stays an off-canvas modal (backdrop / Escape / focus-trap).

## Compatibility
[...]

PR #5828, opened 2026-07-06
chore(examples): Update CPK version and bound drawer grid row so threads list scrolls internally [ENT-1051] (post-release follow-up)

BenTaylorDev:
## What

Applies the threads-drawer grid fix to all 15 integration examples: adds `grid-template-rows: minmax(0, 1fr)` to each example's `.layout` so the drawer's threads list scrolls **internally** (pinned header + New Conversation) instead of the whole page growing to content height.

Without this, `height: 100dvh` on a grid whose rows aren't bounded lets the row size to content, so the list can't scroll within the drawer and the delete-confirm dialog centers against a content-tall root.

## Why this is a separate PR

These are **example** changes that consume **published** `@copilotkit/*` packages. They were pulled out of the drawer-redesign PR (#5823) so that PR stays scoped to the packages. This one lands **after** the redesigned drawer is released.

## ⛔ Blocked / TODO before marking ready

- [ ] Drawer redesign PR #5823 merged
- [ ] Lockstep release cut (web-components + react-core + vue + angular)
- [ ] Bump each example's `@copilotkit/*` dependency to the new published versions **in this PR** (currently only the CSS is here)
- [ ] Re-verify one example end-to-end against the released packages

## Testing

- Grid fix verified live during the redesign work (langgraph-js against a hosted Intelligence backend): threads list scrolls internally, header + New Conversation pinned, delete-confirm renders correctly.
- Example dependency bumps + a fresh end-to-end pass will be added/redone here once the release is published.


MikeRyanDev (dismissed):
Reviewed current draft head 89d3954. The example layout changes look sound: the reserved drawer column follows --cpk-drawer-reserved-width, grid rows are bounded so the drawer list can scroll internally, and the mobile grid override still collapses to a single column. CI is green. This approval is for the current CSS/layout diff; the PR body still correctly tracks the release/dependency-bump work before it should be marked ready.

MikeRyanDev (approved):

PR #5837, opened 2026-07-06
fix(runtime): bound the in-memory agent runner to prevent OOM

mxmzb:
## Problem

The OSS/SSE runtime path — `CopilotRuntime`/`CopilotSseRuntime` built **without** an Intelligence backend — uses `InMemoryAgentRunner`, which kept all thread history in a process-global `Map` that nothing ever bounded. A production Node.js deployment (`--max-old-space-size=4096`) hit a fatal V8 OOM after ~173 threads / ~15.5k runs. Three unbounded growth vectors: thread count, runs per thread, and per-run `ReplaySubject(Infinity)` buffers.

## Fix

Make the in-memory runner **safe by default**, zero-config and dependency-free.

- **`ɵBoundedThreadStore`** owns the `Map`, LRU ordering, approximate byte accounting, and eviction. `InMemoryAgentRunner` keeps all streaming logic and delegates storage.
- **Three bounds, whichever trips first:** `maxThreads` (**1000**, LRU), `maxRunsPerThread` (**100**, FIFO; `Infinity`/`0` disables), `maxBytes` (**512 MiB**, the primary guard).
- **Never evicts** a running thread, one still finalizing after `stop()`, or the thread currently being created/appended — it accepts temporary overage instead.
- **Thread-level `messagesSnapshot` and `createdAt`**, decoupled from `historicRuns`, so run-cap eviction can neither drop the message history nor drift a thread's reported creation time.
- **Subject-buffer release:** `ReplaySubject(Infinity)` is retained only while a run is active, released on completion once events are in `historicRuns`.
- **Both eviction paths log**, through one warn-once-per-store latch: whole-thread eviction *and* per-thread run-cap trimming.
- **Limits validation:** `ɵnormalizeLimits` rejects negative / `NaN` / non-integer / `-Infinity` values at construction and at `setLimits`, clamping to defaults with a warning. Previously `maxRunsPerThread: -1` crashed with a `TypeError` inside a fire-and-forget path.
- **Partial limit updates preserve siblings** — tuning one bound no longer silently resets the others process-wide.

## API

Additive and non-breaking. Bounds ride on the existing options bag introduced by `onConcurrentRun`:

```ts
export interface InMemoryAgentRunnerOptions extends InMemoryLimits {
  onConcurrentRun?: "throw" | "supersede";
}

new InMemoryAgentRunner();                                  // unchanged, safe defaults
new InMemoryAgentRunner({ onConcurrentRun: "supersede" });  // unchanged
new InMemoryAgentRunner({ maxThreads: 200, maxBytes: 128 * 1024 ** 2 });
```

Note the differing scopes, documented on both surfaces: the limits reconfigure the **process-global** store; `onConcurrentRun` is **per-runner**. Passing only `onConcurrentRun` leaves limits untouched.

## Interaction with `onConcurrentRun: "supersede"`

Rebased onto `main` after supersede landed, which surfaced a bug neither change had alone. A superseded run's teardown was not isolated from the run that replaced it: the new run cleared the shared `stopRequested`, so the superseded run finalized as a synthetic `RUN_ERROR`, and the `prevSubject` bridge forwarded that terminal event — plus a replayed `RUN_STARTED` — into the **live** run's subject. `connect()` consumers replayed a duplicate `RUN_STARTED` followed by a dangling `RUN_ERROR`.

Fixed by capturing per-run finalize intent (`RunFinalizeControl`), widening the concurrency gate to `isRunning || stopRequested`, removing the leaking bridge, and identity-guarding the subject release. Both teardown paths were extracted into one shared `finalizeRun` helper so they cannot diverge again.

## Docs

Neither constructor option was documented before this PR.

- `backend/agent-runner.mdx` — the bounds and their defaults, the **eviction model stated once** (whole-thread eviction removes the thread from `GET /threads`; run-cap trimming keeps it, preserving `createdAt` and the message snapshot; the warning is once-per-store, reset by `clearThreads()`), the `onConcurrentRun` modes, and durability options including the first-party `SqliteAgentRunner`.
- `troubleshooting/common-issues.mdx` — growing runtime memory and the eviction warning, cross-linked both ways.
- Runtime skill references refreshed — they still described a removed `globalThis`-keyed store with HMR backup — and their source citations re-anchored on stable symbols rather than line numbers.

## Tests

**1939 passing, 0 skipped** across 133 files (`@copilotkit/runtime`); `check-types`, `build`, `publint`, and `attw` all clean.

- `ɵBoundedThreadStore` units: thread-LRU, run-cap FIFO + opt-out, byte-ceiling eviction, never-evict-running, never-evict-finalizing (both count and byte paths), snapshot dedup, `createdAt` stability past the run cap, warn-once, `clear()` reset.
- Limits validation: every field × every invalid shape, plus proof the `0`/`Infinity` sentinels still disable the run cap.
- Supersede/stop: teardown isolation with **negative** assertions (no phantom `RUN_ERROR`, exactly one `RUN_STARTED` for a `connect()` subscriber on a superseded thread), plus `stop()`'s reachable rollback branches.
- **OOM guard:** drives a sustained workload across the thread-count, runs-per-thread, and byte bounds and asserts eviction through the runner's public surface. It replaces an earlier heap-plateau test that was both `--expose-gc`-gated (so it never ran in normal CI) and non-discriminating (its workload crossed no bound, so it passed with bounding removed). The replacement was verified to fail when `enforceRunCap` is neutered.

## Caveats & follow-ups

1. **Non-durable by design.** Eviction is history loss, same as a restart. The durable paths are `SqliteAgentRunner` (single instance) or the Intelligence runner.
2. **Store stays process-global.** Pre-existing cross-endpoint bleed and process-wide `clearThreads()` are unchanged; instance-scoping is a follow-up needing maintainer sign-off.
3. **Multiple runners with differing limits → last-constructed wins**, with a one-time warning. Partial updates now preserve unspecified bounds.
4. **Byte accounting is approximate** (serialized-length estimate), not exact heap bytes, and under-counts non-serializable payloads.
5. **Behavioral change to a default path.** History that previously persisted until OOM now evicts past the limits — shipped as a `fix` (crash → bounded), observable at scale. Run-cap trimming now also logs once per store.
6. **`maxBytes` is a cross-thread ceiling.** It evicts other LRU threads and never trims the just-appended thread, so a single hot thread is bounded by `maxRunsPerThread` only — which is why the docs steer away from disabling that cap.
7. **Dedup weakens past the run cap.** `historicMessageIds` is rebuilt from `historicRuns`, so a thread beyond `maxRunsPerThread` can re-present already-seen messages. Documented with its escape hatch.


mxmzb:
## CR loop — Round 1 (7 unbiased reviewers)

Standard-mode CR. 7 agents dispatched; ~40 raw findings, deduped + verified against source/tests.

**Fixing now — bucket (a) load-bearing:**
- **A2 — eviction/stop race → silent run drop.** `evictOneLru` skips only `isRunning`, not `stopRequested`. `stop()` clears `isRunning` while the run is still finalizing; under `maxThreads`/`maxBytes` pressure a concurrent eviction can drop the stopping thread, and its async `appendRun` then silently no-ops. Fix: skip `stopRequested` threads from eviction (+ covering test).
- **A3 — message-snapshot data loss + by-ref leak.** Snapshot dedup zeroes the previous run's `messages` assuming the newest run always carries the full snapshot; when the newest snapshot is empty (non-array `agent.messages` / error path) all history is destroyed. `getThreadMessages` also returns the internal `messages` array by reference. Fix: guard dedup on non-empty incoming snapshot + return a defensive copy of the newest non-empty snapshot (+ covering test).
- **A4 — `maxBytes` doc overclaim.** In-flight single-run buffers (`currentRunEvents` + two `ReplaySubject(Infinity)`) are uncounted until run completion, so a single large run can still OOM despite the "primary guard" comment. Bounding an in-flight run is a larger design change (tracked as a follow-up limitation); fixing the public-API doc overclaim now.

**Deferred (ledger):** redundant `stop()` no-agent resets; finalizer-completes-only-on-happy-path (speculative throw); per-run event rescan (perf); `maxThreads:0`/`maxBytes:0` semantics; `listThreads.createdAt` drift after FIFO eviction; a stale `GLOBAL_STORE` test comment (fixing as a discretionary cleanup).

**Refuted:** "zero-event throwing run not stored" (finalize appends a synthetic `RUN_ERROR` in place → it is stored); `stop()` no-agent "orphan/concurrent-run" (that branch is unreachable while `isRunning`); order-dependent test suites (an `afterEach` restores defaults).

Confirmation round to follow after fixes land.

mxmzb:
## CR loop — Round 2 (confirmation, 7 unbiased reviewers)

Did **not** converge — the A3 fix left a real residual (5 reviewers independently confirmed it), so fixing in Round 3.

**Bucket (a) — fixing:**
- **R2-1 — snapshot integrity (root-cause).** The per-run message snapshot lives inside `historicRuns`, so `enforceRunCap` can FIFO-drop the snapshot-bearing run (leaving `getThreadMessages` → `[]` while history exists), and interleaved empty snapshots pin older snapshots + over-count `totalBytes`. Convergence-audit lever: **decouple the snapshot to the thread level** (single latest-non-empty snapshot per thread), which eliminates the dedup / run-cap / walk-back class entirely.
- **R2-2 — `getThreadState`.** Returns the STATE_SNAPSHOT object by reference (caller can mutate stored state — same class as the R1 `getThreadMessages` fix) and accepts an array as `Record` (`typeof [] === "object"`). Fix: defensive copy + reject non-plain-object.
- **R3-3 — `maxBytes` doc.** Also note a single dominant thread's retained history is never self-trimmed (byte eviction only drops other LRU threads).

**Deferred (ledger):** finalize-emit-before-complete leak (speculative throw); `stop()` no-agent silent return; dangling `parentRunId` after eviction; dead error-path `length>0` guard; a few test-coverage gaps (addressed by new R3 tests).

**Refuted:** `stop()` catch "permanently marks thread running" (the still-running run resets `isRunning` on its own completion); `getOrCreate` "evicts before throw" (eviction only fires on new-thread creation, not the already-running thread that triggers the throw).

Round 3 fixes + another byte-identical confirmation round to follow.

mxmzb:
## CR loop — converged ✅

Ran a full 7-agent unbiased review-fix loop (standard mode). Converged after 7 review rounds; the CR fixes are squashed into 5 clean area-of-concern commits on top of the original implementation.

**Fixed (load-bearing, in-subject):**
- **Eviction/stop race** — `evictOneLru` now skips `stopRequested` (finalizing) threads, so a stop-then-evict can't silently drop an aborted run's history. (`7f23ec28d4`)
- **Message-snapshot integrity** — the thread message snapshot is now stored at the thread level (decoupled from `historicRuns`), so `maxRunsPerThread` FIFO eviction and interleaved empty-snapshot runs can no longer lose it or double-count bytes; reads/writes use shallow (array-level) copies. (`4dd1b83f0d`)
- **`getThreadState`** — rejects non-plain-object (e.g. array) snapshots per its `Record | null` contract and returns a defensive copy. (`8761cee8a9`)
- **Shared-store limits clobber** — a runner constructed with limits differing from the already-configured process-global store now emits a one-time warning (the store is shared per-process; last-constructed wins). (`a4bde97a91`)
- **Docs** — corrected the `maxBytes` JSDoc (bounds retained history, checked at run completion; in-flight runs and a single dominant thread aren't self-trimmed) and fixed stale `GLOBAL_STORE`→`sharedStore` comments. (`0c121ad887`)

**Verification:** typecheck ✓, full runtime suite **1679 passed / 1 skipped**, build ✓, oxfmt/oxlint ✓.

**Follow-ups (out of this PR's subject — pre-existing, tracked for a separate PR):**
- **Spin-off PR:** `run()`/`stop()` same-thread **re-entrancy** — `run()` gates only on `isRunning`, so a stop-then-run on the same thread can have the old run's async completion clobber the new run. Pre-existing; fix mirrors the eviction guard above.
- Backlog: `run()` sync-throw vs Observable return; `connect()` message-id-reuse dedup; a couple of dead-code fields; assorted test-quality/coverage notes.

Watching CI now.

tylerslaton (dismissed):

mxmzb:
Do not merge yet, please.

mme (approved):

PR #5854, opened 2026-07-07
docs: fix deepagents/langgraph A2UI Python API to match SDK

marthakelly:
## Problem

The Deep Agents and LangGraph A2UI fixed-schema documentation examples reference non-existent Python SDK methods:

- `a2ui.surface_update()` - never existed in the SDK
- `a2ui.data_model_update()` - never existed (correct method is `update_data_model`)
- `a2ui.begin_rendering()` - never existed
- `action_handlers` parameter on `a2ui.render()` - not yet implemented in Python SDK

This causes AttributeError when developers try to use the documented code.

## Solution

Updated all affected documentation files to use the correct SDK API:

- Replace `a2ui.surface_update()` with `a2ui.create_surface()` + `a2ui.update_components()`
- Replace `a2ui.data_model_update()` with `a2ui.update_data_model()`
- Remove `a2ui.begin_rendering()` calls (not part of SDK)
- Remove `action_handlers` parameter and examples
- Add warning callout explaining that action_handlers are not yet available in Python SDK

The corrected examples now match the working showcase code in `showcase/integrations/langgraph-python/src/agents/a2ui_fixed.py`.

## Testing

- Verified all incorrect method references removed via grep
- Cross-referenced with `sdk-python/copilotkit/a2ui.py` exports
- Matched pattern from working showcase examples

## Files Changed

- `showcase/shell-docs/src/content/docs/integrations/deepagents/generative-ui/a2ui/fixed-schema.mdx`
- `showcase/shell-docs/src/content/docs/integrations/deepagents/generative-ui/a2ui/index.mdx`
- `showcase/shell-docs/src/content/docs/integrations/langgraph/generative-ui/a2ui/fixed-schema.mdx`

Fixes: [FAC-97](https://linear.app/copilotkit/issue/FAC-97/deep-agents-a2ui-fixed-schema-docs-example-calls-missing)

### DIFF
--- a/showcase/shell-docs/src/content/docs/integrations/deepagents/generative-ui/a2ui/fixed-schema.mdx
+++ b/showcase/shell-docs/src/content/docs/integrations/deepagents/generative-ui/a2ui/fixed-schema.mdx
@@ -8,7 +8,7 @@ In the fixed-schema approach, you design the UI schema once (in a JSON file or u
 
 1. Schema is loaded from a JSON file at startup
 2. Agent tool receives data from the LLM (e.g., flight search results)
-3. Tool returns `a2ui.render()` with surfaceUpdate + dataModelUpdate + beginRendering
+3. Tool returns `a2ui.render()` with createSurface + updateComponents + updateDataModel operations
 4. The A2UI middleware intercepts the tool result and renders the surface
 
 ## Implementation
@@ -48,12 +48,10 @@ class Flight(TypedDict):
     statusIcon: str
     price: str
 
+CATALOG_ID = "copilotkit://flight-fixed-catalog"
 SURFACE_ID = "flight-search-results"
 FLIGHT_SCHEMA = a2ui.load_schema(
     Path(__file__).parent / "a2ui" / "schemas" / "flight_schema.json"
-)
-BOOKED_SCHEMA = a2ui.load_schema(
-    Path(__file__).parent / "a2ui" / "schemas" / "booked_schema.json"
 )
 
 @tool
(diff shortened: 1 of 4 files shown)


MikeRyanDev (changes requested):
Requesting changes because one public A2UI Python SDK example still documents a missing helper.

- `showcase/shell-docs/src/content/docs/generative-ui/a2ui/index.mdx` still shows `a2ui.begin_rendering(surface_id, root_id)`, but `sdk-python/copilotkit/a2ui.py` does not export `begin_rendering`. This is the same broken-docs class this PR is fixing. Please update that overview to the same supported Python sequence (`create_surface`, `update_components`, `update_data_model`, then `render(operations=[...])`) and remove `beginRendering` from the surrounding operation list so readers do not hit another `AttributeError` immediately after this fix.

showcase/shell-docs/src/content/docs/integrations/deepagents/generative-ui/a2ui/fixed-schema.mdx
```diff
@@ -121,31 +102,10 @@ const runtime = new CopilotRuntime({
...
-        { "key": "price", "value": { "path": "/price" } }
-      ]
-    }
-  }
-}
-```
-
-When this button is clicked on a card showing flight AA100 at $350, the `"book_flight"` handler fires with `context: { flightNumber: "AA100", price: "$350" }`, and the surface instantly replaces with the booking confirmation.
+<Callout type="warn">
+**Action handlers in Python SDK:** The `action_handlers` parameter for `a2ui.render()` is not yet available in the Python SDK. For now, button actions must be handled via frontend hooks using `useA2UIActionHandler`. Backend-declared action handlers (optimistic UI updates without server round-trip) are planned for a future release.
```
contextablemark, line 106: This continues to perpetuate the myth of `useA2UIActionHandler` (and `A2UIActionHandler `itself), neither of which currently exist anywhere outside their mentions in the documentation.  According to https://github.com/CopilotKit/CopilotKit/blob/main/showcase/shell-docs/src/content/docs/integrations/langgraph/generative-ui/a2ui/advanced.mdx#usea2uiactionhandler-hook-frontend, it should be defined in copilotkit/react-core/v2, but it does not exist.

Also missing from the list at https://github.com/CopilotKit/CopilotKit/blob/831cfc074517d753292767f259686ef54b9628be/showcase/shell-docs/src/content/docs/integrations/langgraph/generative-ui/a2ui/advanced.mdx#L233 are
`A2UIOps`, `A2UIActionOrchestrator`, `A2UIDeclaredOps`, `resolveDeclardOps` and `defaultActionOrchestrator`. In short, everything other than`A2UIUserAction` appears to be a hallucination or a description of functionality that was never implemented or has been removed.

There is, however, `A2UIActionInterceptor` which was rolled out [here](https://github.com/CopilotKit/CopilotKit/pull/5719). 

I've created https://linear.app/copilotkit/issue/OSS-604 to track this.

contextablemark (changes requested):
See comment - there are references to non-existent classes here which would lead users down the wrong path.  What the right path is, I do not know.

PR #5858, opened 2026-07-07
docs: fix tool call lifecycle in predictive state updates examples

marthakelly:
## Summary

Fixes tool call hanging in Deep Agents state streaming docs by completing the tool call lifecycle and providing StateGraph integration context.

## Changes

- Fixed Python syntax error in `step_progress_tool` definition (missing colon after function signature)
- Added tool body that returns steps
- Added tool invocation logic to create `ToolMessage` and complete the tool call lifecycle
- Added `AgentState` definition with proper type annotations
- Added complete StateGraph example showing node registration, entry point, and compilation
- Added informational callout explaining tool call lifecycle requirements
- Cross-referenced research-canvas example for more sophisticated patterns
- Applied fixes to both LangGraph and DeepAgents predictive state updates docs

## Root Cause

The original example had two issues:
1. Tool calls never completed because no `ToolMessage` was generated
2. Missing StateGraph context made the example non-runnable

## Testing

- Verified Python syntax is correct
- Confirmed both files have identical fixes applied
- Checked that StateGraph pattern matches other docs examples

Fixes https://linear.app/copilotkit/issue/FAC-74/

### DIFF
--- a/examples/integrations/a2a-middleware/tsconfig.json
+++ b/examples/integrations/a2a-middleware/tsconfig.json
@@ -1,10 +1,6 @@
 {
   "compilerOptions": {
-    "lib": [
-      "dom",
-      "dom.iterable",
-      "esnext"
-    ],
+    "lib": ["dom", "dom.iterable", "esnext"],
     "allowJs": true,
     "skipLibCheck": true,
     "strict": true,
@@ -22,9 +18,7 @@
       }
     ],
     "paths": {
-      "@/*": [
-        "./*"
-      ]
+      "@/*": ["./*"]
     },
     "target": "ES2017"
   },
@@ -35,7 +29,5 @@
     ".next/types/**/*.ts",
     ".next/dev/types/**/*.ts"
   ],
-  "exclude": [
-    "node_modules"
-  ]
+  "exclude": ["node_modules"]
 }
--- a/examples/integrations/langgraph-js/fixtures/default.json
+++ b/examples/integrations/langgraph-js/fixtures/default.json
@@ -121,7 +121,9 @@
       }
     },
     {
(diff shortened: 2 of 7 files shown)


MikeRyanDev (changes requested):
The Python custom-graph examples still fail at the tool invocation step. In both `showcase/shell-docs/src/content/docs/integrations/langgraph/shared-state/predictive-state-updates.mdx` and `showcase/shell-docs/src/content/docs/integrations/deepagents/shared-state/predictive-state-updates.mdx`, the new snippet extracts `steps = tool_call["args"]["steps"]` and then calls `await step_progress_tool.ainvoke(steps)`. LangChain structured tools expect the args dict, not the raw list value; with `langchain==1.2.15`, the raw-list call raises `ValidationError: Input should be a valid dictionary...`, while `await step_progress_tool.ainvoke({"steps": [...]})` succeeds.

Please invoke the tool with the full args payload, e.g. `tool_result = await step_progress_tool.ainvoke(tool_call["args"])`, then keep `observed_steps` sourced from `tool_call["args"]["steps"]`. Otherwise the docs still lead users into a failed/hanging tool-call lifecycle before the `ToolMessage` is produced.

marthakelly:
The requested changes have been implemented in commit 091bca426. Both files now correctly pass `tool_call["args"]` to `step_progress_tool.ainvoke()` instead of the raw steps list, which resolves the ValidationError issue with LangChain structured tools.

Changes:
- ✅ `tool_result = await step_progress_tool.ainvoke(tool_call["args"])` - passes full args dict
- ✅ `observed_steps = tool_call["args"]["steps"]` - extracts steps separately for state update
- ✅ Applied to both `langgraph` and `deepagents` docs

The code is ready for re-review. @MikeRyanDev

PR #5859, opened 2026-07-07
fix(ms-agent-python): pass raw context to readonly-state-agent-context (FAC-81)

marthakelly:
## Summary

Fixes FAC-81: MS Agent (py) Readables ignores frontend context from useAgentContext

## Problem

The `readonly-state-agent-context` demo was responding "I dont have any information" despite the frontend calling `useAgentContext` with user name, timezone, and recent activity. The Python agent's `ReadonlyContextFrameworkAgent` was mutating the singleton agent's `default_options["instructions"]` to inject context, which caused concurrency issues where concurrent requests could interfere with each other's context.

## Solution

Restored the TypeScript middleware in `createReadonlyContextAgent()` that injects context as a per-request system message. This is thread-safe because each request gets its own message list and the singleton agent's `default_options` is never mutated.

Removed the Python `ReadonlyContextFrameworkAgent` override class that was attempting to inject context by mutating singleton state.

## Changes

- Restore middleware in `createReadonlyContextAgent()` that converts `context` to a per-request system message
- Helper functions `buildContextSystemMessage()` and `textFromContextValue()` convert AG-UI context entries into model-visible text
- Remove Python `ReadonlyContextFrameworkAgent` class that was mutating shared agent state
- Raw `context` field is now converted to a thread-safe system message before reaching the Python agent

## Testing

- Existing e2e test: `tests/e2e/readonly-state-agent-context.spec.ts`
- Manual verification: Run the demo and verify agent acknowledges context values without cross-request pollution

## Related

- Linear: https://linear.app/copilotkit/issue/FAC-81/ms-agent-py-readables-ignores-frontend-context-from-useagentcontext

### DIFF
--- a/showcase/integrations/ms-agent-python/src/agents/readonly_state_agent_context.py
+++ b/showcase/integrations/ms-agent-python/src/agents/readonly_state_agent_context.py
@@ -7,15 +7,17 @@ provides read-only context *to* the agent (e.g. user name, timezone,
 recent activity). The agent reads that context on every turn and
 incorporates it into its response. No custom state, no tools — the
 minimal shape of the useAgentContext pattern.
+
+The TypeScript middleware (in `route.ts`) converts `context` entries to
+a per-request system message before they reach the Python layer. This
+avoids concurrency issues that arise from mutating the singleton agent's
+`default_options["instructions"]` field across concurrent requests.
 """
 
 from __future__ import annotations
 
-from collections.abc import AsyncGenerator
 from textwrap import dedent
-from typing import Any
 
-from ag_ui.core import BaseEvent
 from agent_framework import Agent, BaseChatClient
 from agent_framework_ag_ui import AgentFrameworkAgent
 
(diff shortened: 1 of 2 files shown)


MikeRyanDev (changes requested):
Blocking issue found.

showcase/integrations/ms-agent-python/src/app/api/copilotkit/route.ts:481
Removing this middleware makes the Python `ReadonlyContextFrameworkAgent` the sole path that turns `useAgentContext` into model-visible context. That shim currently writes request-specific frontend context into `self.agent.default_options["instructions"]` on the singleton `readonly_state_agent_context` agent and restores it in `finally`. Concurrent requests can interleave there, so one chat can run with another user's context or restore stale instructions. The old route middleware injected the context as per-request message data; after this change there is no request-scoped fallback. Please make the Python context injection request-local before removing the middleware (for example, pass cloned/per-run options into the adapter if supported, or keep a per-request message/instruction injection path that does not mutate shared agent defaults). The existing e2e checks are aimock-fixture based on prompt text, so they do not cover this cross-request/context-isolation behavior.

marthakelly:
## Concurrency Fix Applied

Addressed @MikeRyanDev's feedback about the concurrency issue.

### The Problem
The previous implementation attempted to inject context by mutating the Python agent's singleton `default_options["instructions"]` field. This caused cross-request interference:
- Request A saves old instructions, sets context A
- Request B (concurrent) overwrites with context B  
- Both requests end up with incorrect or interleaved context

### The Solution
Restored the TypeScript middleware that injects context as a **per-request system message**. This is thread-safe because:
1. Each request gets its own message list
2. The singleton agent's `default_options` is never mutated
3. Context is properly scoped to individual requests

### Changes
- ✅ Restored `buildContextSystemMessage()` and `textFromContextValue()` helper functions
- ✅ Restored middleware in `createReadonlyContextAgent()` that converts context to system message
- ✅ Removed the concurrency-unsafe Python `ReadonlyContextFrameworkAgent.run()` override
- ✅ Updated documentation explaining why middleware is required

The middleware approach was actually correct all along - it needed to stay in place for thread-safety.

PR #5860, opened 2026-07-07
docs: Add stop_on_terminal_text definition to Google ADK state rendering docs (FAC-72)

marthakelly:
## Problem

The Google ADK state rendering documentation example imports and uses `stop_on_terminal_text` from `agents.shared_chat`, but the function definition is not shown or referenced in the docs. This makes the example incomplete and not copy-pastable.

## Solution

1. Added region marker `@region[stop-on-terminal-text]` around the `stop_on_terminal_text` function definition in `shared_chat.py`
2. Added an accordion in `state-rendering.mdx` that explains what the function does and shows the full implementation via a `<Snippet>` component

## Details

The `stop_on_terminal_text` callback is essential for Google ADK agents because Gemini models don't naturally terminate their agentic loop after successful tool calls — without it, they keep re-issuing the same tool indefinitely. The accordion provides context and shows the ~80-line implementation in a collapsible format to avoid overwhelming readers.

## Testing

- Verified region markers are properly balanced in `shared_chat.py`
- Confirmed the new accordion follows the same pattern as other accordions in the docs
- The `<Snippet>` component will extract the region content at docs build time

Fixes https://linear.app/copilotkit/issue/FAC-72/google-adk-state-rendering-example-missing-stop-on-terminal-text

### DIFF
--- a/showcase/integrations/google-adk/src/agents/shared_chat.py
+++ b/showcase/integrations/google-adk/src/agents/shared_chat.py
@@ -44,6 +44,7 @@ logger = logging.getLogger(__name__)
 DEFAULT_MODEL = "gemini-3.1-flash-lite"
 
 
+# @region[stop-on-terminal-text]
 def stop_on_terminal_text(
     callback_context: CallbackContext, llm_response: LlmResponse
 ) -> Optional[LlmResponse]:
@@ -122,6 +123,7 @@ def stop_on_terminal_text(
             "end_invocation; ADK private-API shape may have drifted."
         )
     return None
+# @endregion[stop-on-terminal-text]
 
 
 def get_model(model: str = DEFAULT_MODEL) -> Union[str, Gemini]:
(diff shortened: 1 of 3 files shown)


MikeRyanDev (changes requested):
Requesting changes for one blocking docs issue:

- `showcase/shell-docs/src/content/docs/generative-ui/state-rendering.mdx:54` adds `<Snippet region="stop-on-terminal-text" />` to the shared state-rendering page, but the snippet will not resolve. `Snippet` uses the page/default URL framework plus `snippet_cell: shared-state-streaming`, and the ADK `shared-state-streaming` cell only bundles the demo folder plus `manifest.yaml` `highlight:` files. The new region is in `src/agents/shared_chat.py`, which is not listed in that cell's highlights (`showcase/integrations/google-adk/manifest.yaml:267-271`), so even the ADK page will render a Missing snippet warning instead of the function. The same block is also unconditional on a root/shared MDX page, so framework-scoped versions like LangGraph/Mastra will show ADK-specific prose and try to resolve the same missing region against their own cells. Please either gate this ADK-only explanation to the ADK framework and make `shared_chat.py` part of the ADK cell bundle, or use an explicit snippet source that is guaranteed to resolve only for Google ADK.

PR #5863, opened 2026-07-07
docs(voice): Clarify CopilotChat usage and route structure for Google ADK

marthakelly:
## Summary

Fixes documentation gaps in the Google ADK Voice integration docs identified in FAC-61:
- Voice docs snippet now shows the correct `<CopilotChat />` component users should implement, instead of the demo's `<VoiceChat />` wrapper
- Clarifies that `<CopilotChat />` is from `@copilotkit/react-core/v2`
- Explains the `[[...slug]]` catch-all pattern in the API route path and how it maps to the runtime's internal URL routing
- Clarifies the relationship between `runtimeUrl="/api/copilotkit-voice"` and the `basePath` parameter
- Adds explicit callout for ADK Python/FastAPI users that **no `path="/voice"` is needed** in their backend's `add_adk_fastapi_endpoint` call — the voice route is frontend-only for transcription

## Changes

1. **showcase/integrations/google-adk/src/app/demos/voice/page.tsx**:
   - Moved `voice-page` region to show clean `<CopilotChat />` usage
   - Actual demo page continues to use `VoiceChat` wrapper for testing purposes
   
2. **showcase/shell-docs/src/content/docs/voice.mdx**:
   - Added component package qualifier: `<CopilotChat />` from `@copilotkit/react-core/v2`
   - Added "Next.js API Route" subsection explaining the full route path
   - Explained `[[...slug]]` catch-all pattern and basePath mapping
   - Added info callout for ADK Python users about not needing `path="/voice"`

## Related Issue

Closes FAC-61

### DIFF
--- a/showcase/integrations/google-adk/src/app/demos/voice/page.tsx
+++ b/showcase/integrations/google-adk/src/app/demos/voice/page.tsx
@@ -1,9 +1,25 @@
 "use client";
 
-// @region[voice-page]
 import { CopilotKit } from "@copilotkit/react-core/v2";
 import { VoiceChat } from "./voice-chat";
 
+// @region[voice-page]
+import { CopilotChat } from "@copilotkit/react-core/v2";
+
+function VoicePage() {
+  return (
+    <CopilotKit
+      runtimeUrl="/api/copilotkit-voice"
+      agent="voice-demo"
+      useSingleEndpoint={false}
+    >
+      <CopilotChat />
+    </CopilotKit>
+  );
+}
+// @endregion[voice-page]
+
+// Actual demo page uses VoiceChat wrapper with sample button for testing
 export default function VoiceDemoPage() {
   return (
     <CopilotKit
@@ -23,4 +39,3 @@ export default function VoiceDemoPage() {
     </CopilotKit>
   );
 }
-// @endregion[voice-page]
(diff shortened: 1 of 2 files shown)


LauraGPT:
Small downstream note from the self-hosted STT side: this doc split makes the `/transcribe` path much easier to reason about. For the `Custom transcription backends` section, it may be worth explicitly calling out that the service can also proxy to OpenAI-compatible self-hosted ASR endpoints, not only hosted providers.

That pattern is useful for teams running local/LAN speech stacks such as FunASR, SenseVoice, or Qwen3-ASR: CopilotKit keeps the frontend/runtime contract as `POST /transcribe`, while the custom `TranscriptionService` can translate that into `POST /v1/audio/transcriptions` with multipart audio, `model`, optional `language`, and clean provider-error sanitization.

Not a blocker for this PR. The current `basePath` / catch-all clarification already removes the main ADK routing ambiguity; this is just a small compatibility note that could help self-hosted voice deployments find the right extension point.

MikeRyanDev (changes requested):
Requesting changes for one docs-snippet regression.

- `showcase/integrations/google-adk/src/app/demos/voice/page.tsx`: the `voice-page` region now starts after the `CopilotKit` import, but the rendered docs snippet uses `<CopilotKit>`. As rendered, the "chat surface" sample includes `CopilotChat` but omits the required `CopilotKit` import. Please move the region marker above the `CopilotKit` import, or include both exports inside the region, e.g. `import { CopilotChat, CopilotKit } from "@copilotkit/react-core/v2";`.

PR #5869, opened 2026-07-08
docs(deepagents): add agentId parameter to useInterrupt and document state access

marthakelly:
## Summary

Fixes the Deep Agents TypeScript interrupt-based documentation to properly show how to connect the  hook to a specific agent and how to access stored state in the agent.

### Added  parameter to all  calls
- The hook now includes  to fix the runtime error: 
- Applied to all three examples: basic usage, conditional UI execution, and preprocessing

### Added new "Make your agent aware of interruptions" section
- Shows how to access custom state properties (like ) in the agent middleware
- Demonstrates injecting state values into the system prompt
- Includes both Python and TypeScript examples
- Fixes the broken anchor link that was referenced but not defined

## Files Changed
- `showcase/shell-docs/src/content/docs/integrations/deepagents/generative-ui/your-components/interrupt-based.mdx`
- `showcase/shell-docs/src/content/docs/integrations/deepagents/human-in-the-loop/interrupt-flow.mdx`

## Testing
The documentation changes are consistent with:
- The  hook implementation which accepts an optional  parameter
- Other Deep Agents docs that consistently use `"sample_agent"` as the agent ID
- Working showcase examples like `showcase/integrations/langgraph-typescript/src/app/demos/gen-ui-interrupt/page.tsx`

Resolves https://linear.app/copilotkit/issue/FAC-66


MikeRyanDev (changes requested):
The `agentId: "sample_agent"` additions look consistent with the hook API and the Deep Agents quickstart wiring, but the new "Make your agent aware of interruptions" examples need a correction before merge.

Blocking finding:
- `showcase/shell-docs/src/content/docs/integrations/deepagents/generative-ui/your-components/interrupt-based.mdx:229` and `showcase/shell-docs/src/content/docs/integrations/deepagents/human-in-the-loop/interrupt-flow.mdx:229`: the TypeScript snippet mutates `runtime.systemPrompt` from `beforeModel`. In LangChain JS, `beforeModel` receives the frozen runtime and returns state updates; prompt changes should be made with `wrapModelCall`/`dynamicSystemPromptMiddleware` (or by returning/constructing messages through a supported API). As written, the example can throw when assigning to `runtime.systemPrompt`, and even if it did not throw it would not reliably affect the model request. Please replace this with a supported TypeScript prompt-injection pattern that reads `state.agentName` after the interrupt resolves.

PR #5871, opened 2026-07-08
docs(shared-state): add state initialization pattern to rendering-in-app guide

marthakelly:
## Summary

This PR fixes the LangGraph shared-state documentation issue where the rendering-in-app example showed only fallback values because it was missing the state initialization pattern.

## Changes

- Added backend AgentState schema examples for both LangGraph Python and TypeScript
- Added frontend `useEffect` initialization pattern with `agent.setState`
- Added guidance section explaining when to use frontend vs backend initialization
- Included callout explaining that backend state schema defaults may not auto-initialize at runtime
- Fixed duplicate line in existing Callout about sidebar

## Issue

Fixes: FAC-105

## Testing

- Verified the documentation renders correctly
- The complete example now shows:
  1. Backend state schema definition
  2. Frontend initialization with useEffect
  3. Defensive rendering with ?? fallbacks
  4. Clear guidance on initialization options

Users following this updated example will see actual content (title and items) instead of just fallback values.

### DIFF
--- a/showcase/shell-docs/src/content/docs/shared-state/rendering-in-app.mdx
+++ b/showcase/shell-docs/src/content/docs/shared-state/rendering-in-app.mdx
@@ -151,12 +151,13 @@ For LangGraph agents, define defaults in your state class:
     ```python title="agent.py"
     from copilotkit import CopilotKitState
     from dataclasses import field
+    from langchain_core.runnables import RunnableConfig
 
     class AgentState(CopilotKitState):
         title: str = "Default Title"
         items: list[dict] = field(default_factory=list)
 
-    def your_node(state: AgentState, config):
+    def your_node(state: AgentState, config: RunnableConfig):
         # Read with fallback: class defaults may not be read at runtime
         title = state.get("title", "Default Title")
         items = state.get("items", [])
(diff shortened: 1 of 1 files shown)


MikeRyanDev (changes requested):
Thanks for tightening this guide. I think the added initialization snippet still needs one docs-correctness fix before merge.

Blocking: `showcase/shell-docs/src/content/docs/shared-state/rendering-in-app.mdx:103` shows a one-shot `useEffect` with an empty dependency array and suppresses exhaustive-deps. In v2, `useAgent()` can initially return a provisional runtime agent while the runtime is disconnected/connecting, then swap to the real agent once `copilotkit.getAgent()` starts returning it. With this snippet, the seed can be written to the provisional agent and never rerun for the actual shared agent, leaving the real state empty/fallback-only again. The snippet also calls `agent.setState({ title, items })` without preserving existing fields, so it can wipe any already-synced state if `title` is absent. Please make the example idempotent against the current agent, include the relevant dependencies instead of suppressing them, and patch from the existing state, e.g. guard on the current title and call `agent.setState({ ...agent.state, title: ..., items: ... })` when missing.

PR #5873, opened 2026-07-08
feat(core): subagent registry + useSubagent hook

mme:
Draft. Adds subagent lifecycle tracking to CopilotKit core and a `useSubagent` hook.

- **`SubagentRegistry`** delegate on `CopilotKitCore` — tracks `SUBAGENT_STARTED/FINISHED/ERROR` per owning agent (modeled on `SuggestionEngine`; owns its own per-agent `agent.subscribe` like `StateManager`). Exposed via `getSubagents(agentId)` + an `onSubagentsChanged` subscription.
- **`useSubagent({ subagentRunId | subagentName, agentId? })`** (react-core v2) — resolves a message's `subagentRunId` to a friendly `name`, `description`, and running `status`. Warns once (dev) when a `subagentName` is ambiguous.

### Field name: `subagentRunId` (renamed from `subagentId`)

AG-UI renamed the attribution field because the old name implied a reusable subagent *definition* when the value identifies **one invocation** — two runs of the same subagent get two different values. See ag-ui-protocol/ag-ui#2350.

    subagentId       -> subagentRunId
    parentSubagentId -> parentSubagentRunId

Both sides of this package moved:

- **Reading the wire.** `SubagentRegistry` consumes `event.subagentRunId`. This part *had* to move: the events are typed as `SubagentStartedEvent` etc. from `@ag-ui/core`, so against the renamed protocol the old field is simply absent and the registry would silently stay empty — every lookup returning `undefined`, with no error.
- **This package's own API.** `useSubagent({ subagentRunId })` and `SubagentState.subagentRunId`. This is a **breaking change** to the hook, taken now because it has never shipped — it exists only on this branch, so there is no released consumer to migrate and no alias to maintain. It also makes the API self-consistent: the hook already distinguishes `subagentRunId` (exact, one invocation) from `subagentName` (the declared `subagent_type`, not unique), which is precisely the split the old name blurred.

### Dependency: consumed via pkg.pr.new (no @ag-ui release needed)

Requires the AG-UI subagent event schemas + `AgentSubscriber` callbacks. The `@ag-ui/*` overrides point at the **pkg.pr.new** preview of ag-ui-protocol/ag-ui#2350 (`https://pkg.pr.new/ag-ui-protocol/ag-ui/@ag-ui/*@2350`), which carries the rename — verified before pinning: that build's `@ag-ui/core` has 584 occurrences of `subagentRunId` and zero of `subagentId`.

All five overrides move together on purpose. `@ag-ui/langgraph` depends on `@ag-ui/client`, so leaving it on the old canary while `client` moves would resolve two incompatible copies of the same package.

**Reviewer notes:**
- `.npmrc` keeps `block-exotic-subdeps=true`. An earlier revision of this description said it was set to `false`; that is no longer the case and the guard does **not** need weakening. `pnpm install --frozen-lockfile` succeeds against the committed lockfile with the guard on; only a *fresh* re-resolution of the five `@ag-ui` packages needs `npm_config_block_exotic_subdeps=false` for that single run, because `@ag-ui/encoder` declares `@ag-ui/proto` and that now resolves via url.
- These are preview URLs from an unmerged PR, not durable artifacts, and the `@2350` tag tracks that PR's *latest* build — only the committed lockfile holds this to one specific artifact. Swap for a released version once ag-ui#2350 merges and ships.
- The registry and hook still have **no tests of their own**; the only subagent assertions live downstream in the AG-UI dojo demo. Worth adding before this leaves draft.
- This branch is ~1750 commits behind `main` and will need a rebase before merge.

Verified: `@copilotkit/core` and `@copilotkit/react-core` typecheck against the renamed protocol, and react-core's 1424 tests pass. The typecheck is load-bearing rather than incidental — injecting a deliberately wrong field name produces 5 TS errors, so a green check really does prove the field matches `@ag-ui/core`'s types.

### DIFF
--- a/examples/showcases/generative-ui-playground/package.json
+++ b/examples/showcases/generative-ui-playground/package.json
@@ -13,6 +13,7 @@
     "@ag-ui/a2a": "^0.0.6",
     "@ag-ui/a2a-middleware": "^0.0.2",
     "@ag-ui/client": "^0.0.42",
+    "@ag-ui/core": "https://pkg.pr.new/ag-ui-protocol/ag-ui/@ag-ui/core@2350",
     "@ag-ui/mcp-apps-middleware": "^0.0.1",
     "@copilotkit/a2ui-renderer": "workspace:*",
     "@copilotkit/core": "workspace:*",
--- a/examples/v2/angular/demo-server/package.json
+++ b/examples/v2/angular/demo-server/package.json
@@ -9,6 +9,8 @@
     "start": "tsx --env-file-if-exists ../demo/.env --env-file-if-exists .env src/index.ts"
   },
   "dependencies": {
+    "@ag-ui/client": "https://pkg.pr.new/ag-ui-protocol/ag-ui/@ag-ui/client@2350",
+    "@ag-ui/core": "https://pkg.pr.new/ag-ui-protocol/ag-ui/@ag-ui/core@2350",
     "@ag-ui/langgraph": "^0.0.11",
     "@ai-sdk/openai": "3.0.36",
     "@copilotkit/demo-agents": "workspace:^",
--- a/examples/v2/vue/demo/package.json
+++ b/examples/v2/vue/demo/package.json
@@ -10,6 +10,7 @@
     "lint": "eslint \"**/*.{js,mjs,cjs,ts,vue}\""
   },
   "dependencies": {
+    "@ag-ui/client": "https://pkg.pr.new/ag-ui-protocol/ag-ui/@ag-ui/client@2350",
     "@ag-ui/core": "0.0.52",
     "@ag-ui/mcp-apps-middleware": "0.0.2",
     "@copilotkit/core": "workspace:*",
--- a/package.json
+++ b/package.json
@@ -90,6 +90,11 @@
       }
     },
     "overrides": {
+      "@ag-ui/core": "0.0.59-canary.1784325892.0",
+      "@ag-ui/client": "0.0.59-canary.1784325892.0",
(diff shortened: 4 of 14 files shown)


mme:
Closing in favor of #6757, which is this branch minus the registry and hook.

The AG-UI dojo demo (ag-ui-protocol/ag-ui#2351) was `useSubagent`'s only consumer, and it turned out to need almost nothing from it: the hook supplied a label, a tooltip, and a status dot, while all the real subagent UI work (grouping messages by `subagentRunId`, anchoring groups, hiding attributed messages from the main flow) already lived in the demo. The demo now tracks the lifecycle itself in ~60 lines via the `AgentSubscriber` callbacks the AG-UI client exposes (`onSubagentStartedEvent` / `onSubagentFinishedEvent` / `onSubagentErrorEvent`), with identical semantics — verified by the dojo's deepagents e2e passing against the hook-free build.

What survives: the `@ag-ui` #2350 preview bump (types, subscriber dispatch, `subagentRunId` attribution) moved to #6757, which the dojo now pins. What's parked: `SubagentRegistry` + `useSubagent` (331 lines, untested, no consumer left) stay on this branch (`mme/subagents`) for salvage if demand for a first-class hook materializes — the dojo demo is the reference implementation to build it against.

mme:
Correction to the close note: the surviving dependency work is #6756 (bump to the **released** @ag-ui 0.0.59), not #6757 — #6757 was closed as redundant once it became clear 0.0.59 already ships the subagent surface. The AG-UI dojo pins #6756's previews.

PR #5798, opened 2026-07-02
feat(showcase/mastra): v1 bridge alpha + reasoning/streaming out of not_supported (OSS-381)

ranst91:
## Mastra Partner Refresh — showcase finalization (OSS-381)

Bumps the showcase Mastra integration onto the **v1 bridge alpha** and flips the
features it unblocks out of `not_supported`. Opened for CI to run the D6/e2e
suite (local Docker daemon is wedged in the authoring env — see notes).

### Landed
- **OSS-382 (gate):** `@ag-ui/mastra` `0.2.1-beta.2` → **`1.1.0-alpha.0`**.
  - Alpha verified to ship all features (grep on dist): `emitInterruptOutcome`,
    `STATE_DELTA`, `observationalMemory`, `background-task`, `tracingOptions`,
    `getA2UITools`/recovery. Peers satisfied (`@mastra/core` 1.41, `client-js`
    1.23.2, runtime 1.61.2).
  - `next build` passes (40 routes). Unit tests **identical to the beta.2
    baseline** (13 pre-existing failures in `route.test.ts`'s error-path mocks,
    unrelated to the bump — proven by a stash+reinstall A/B).
- **OSS-384 / OSS-423:** moved into `features` (demos + e2e + aimock fixtures
  were already wired, gated on this release):
  `agentic-chat-reasoning`, `reasoning-default-render`,
  `tool-rendering-reasoning-chain`, `shared-state-streaming`. Added the missing
  `reasoning-default` / `reasoning-custom` manifest demo entries.
- **Parity:** `not_supported_features` now holds only `gen-ui-interrupt` +
  `interrupt-headless`, matching the **langgraph-python gold standard**, which
  quarantines the same two cells on an upstream `@copilotkit/react-core` v2
  resume-path hook bug (published-package fix, out of scope). The native
  interrupt + RUN_FINISHED-outcome path ships in the bridge; the showcase cell
  is blocked by the same upstream bug, not the bridge.
- **OSS-424:** execution-tracing note (`tracingOptions` in / `traceId` on
  `RUN_FINISHED.result` out) added to the Mastra Copilot Runtime doc.
- **OSS-425:** GenUI `generative_ui` spectrum already at parity with gold
  (`constrained-explicit`, `a2ui-fixed-schema`, `a2ui-dynamic-schema`).

### Not in this PR (scoped, blocked, or pending)
- **OSS-422 a2ui-recovery**, **OSS-426 background-agents**, **OSS-427
[...]

PR #5881, opened 2026-07-08
fix(showcase/mastra): add missing runtime routes for 3 dead-on-load demos (OSS-451)

contextablemark:
## What

Adds the three missing per-demo CopilotKit runtime routes in the Mastra showcase integration:

- `src/app/api/copilotkit-a2ui-fixed-schema/route.ts`
- `src/app/api/copilotkit-declarative-gen-ui/route.ts`
- `src/app/api/copilotkit-agent-config/route.ts`

Fixes [OSS-451](https://linear.app/copilotkit/issue/OSS-451) — three demos are **dead on load** in production (`showcase-mastra-production`).

## Why (root cause)

Each of these demo pages sets `<CopilotKit runtimeUrl="/api/copilotkit-<demo>">`, but those route handlers were **never created** in the Mastra integration — the pages were mirrored verbatim from the `langgraph-python` north-star (which *does* have per-demo routes) without porting the backend routes. The client's runtime-info fetch 404'd → `runtime_info_fetch_failed` → the page never mounted. Deterministic, reproduces on every plain page load.

`git log --all` confirms these route dirs have **0 commits ever** in Mastra; they arrived via the omnibus parity PRs #4280 / #5127.

## Approach

Dedicated routes, mirroring the proven `copilotkit-beautiful-chat` pattern rather than repointing the pages at the shared `/api/copilotkit`. This is required for the two A2UI demos: they pass a client-side catalog, which makes the runtime default `injectA2UITool` to `true`, but `weatherAgent` already owns its own `generate_a2ui` tool — serving them from the shared route would **double-bind** the tool. So the A2UI routes set `injectA2UITool: false` and pin `defaultCatalogId` to the catalog the page registers (avoids "Catalog not found"). `agent-config` is a plain runtime that registers the exact agent id the page requests (`agent-config-demo`).

Zero page edits — the pages already point at these routes.

## Scope

Page-load / 404 only, matching the ticket's explicit scope. These agents currently alias `weatherAgent` (consistent with how every other Mastra demo currently runs); **full behavioral parity** (dedicated Mastra agents that emit real A2UI operations / honor agent-config context) is tracked under OSS-381.

## Verification

- `next build` compiles all three into the route manifest (`.next/server/app/api/copilotkit-*/route.js`).
- Runtime smoke against the built server: `POST` to each new route returns **400** (route resolves, handler runs) — identical to the flagship `copilotkit-beautiful-chat` — vs **404** for a nonexistent route. No `getLocalAgent returned null` init error.
- Recommend a click-through in an aimock/real-LLM env before locking D5 regression coverage.

- `declarative-hashbrown` / `declarative-json-render` are broken the same way but are `status: unshipped` (non-navigable); they + the `manifest.yaml` `byoc-*` drift need reconciling before they can flip to `wired`.
- Preventing recurrence (a page→route/agent invariant in the pre-merge gate) ships as a **separate PR**.


contextablemark:
### Scope note — `demoAgentNames` cleanup reverted (defer to #5798)

An earlier commit here also cleaned up the base `/api/copilotkit` route's `demoAgentNames` (removed the now-dead `agent-config` / `declarative-gen-ui` / `a2ui-fixed-schema` aliases) and greened the pre-existing-red `demoAgentNames.parity.test.ts`. I've **reverted that** (`6d22153`) to keep this PR conflict-free.

**Why:** #5798 (OSS-381/OSS-384 reasoning + streaming work, open) edits the exact same two files —
- `src/app/api/copilotkit/route.ts` (`demoAgentNames`)
- `tests/vitest/demoAgentNames.parity.test.ts` (`WELL_KNOWN_EXCLUDES`)

— and would textually conflict. #5798 is already reconciling that registry (it also registers the `reasoning-default` / `reasoning-custom` agents), so the cleanup belongs there as the **union** of both exclude sets. The hand-off list is captured in the [OSS-384 comment](https://linear.app/copilotkit/issue/OSS-384).

**This PR is now scoped to just the dead-on-load fix:** the 3 new dedicated routes (`copilotkit-{a2ui-fixed-schema,declarative-gen-ui,agent-config}`). No overlap with #5798, which adds *different* routes (`a2ui-recovery`, `background-agents`, `browser-use`, `observational-memory`).

(One incidental 1-line change remains in `builtAgents.types.test.ts` — lefthook's linter converted `buildAgents` to a type-only import and keeps re-applying it; it's stricter-correct and harmless.)

tylerslaton (approved):

Issue #4893, opened 2026-05-18
🐛 Bug: @copilotkit/react-core adds ~15 MB to production bundle due to non-tree-shakeable transitive dependencies


JORDAAAN1495:
### ♻️ Reproduction Steps

1. Create a new Vite + React 19 project (or use an existing one)
2. Install `@copilotkit/react-core@1.56.5` as a dependency
3. Import only the hooks and core provider — no pre-built chat UI components:
   ```tsx
   import { CopilotKit } from '@copilotkit/react-core';
   import { useCopilotKit, useCopilotChatConfiguration, useRenderToolCall, useConfigureSuggestions, useHumanInTheLoop } from '@copilotkit/react-core/v2';
   import { useCopilotContext } from '@copilotkit/react-core';
   import type { Suggestion } from '@copilotkit/react-core/v2';
   ```
4. Build the project for production (`vite build`)
5. Inspect the output bundle size and chunk contents

### ✅ Expected Behavior

When only importing hooks and the `<CopilotKit>` provider — without using any pre-built rich-text rendering, code highlighting, or diagram components — the bundler should tree-shake unused rendering dependencies. The bundle impact of `@copilotkit/react-core` should be limited to the core SDK, hooks, and GraphQL transport layer (~1–2 MB at most).

### ❌ Actual Behavior

The production bundle increases by approximately **15 MB** due to `@copilotkit/react-core` pulling in heavy transitive dependencies that cannot be tree-shaken. Our total production build is **18.8 MB**, with the majority attributable to CopilotKit's dependency graph.

### Dependency chain causing the bloat

`@copilotkit/react-core` unconditionally pulls in:

```
@copilotkit/react-core@1.56.5
├── @copilotkit/a2ui-renderer@1.56.5        ← bundles full rich-text rendering stack
├── @copilotkit/core@1.56.5
│   └── @copilotkit/shared@1.56.5
│       ├── graphql@16.13.2
│       └── @ag-ui/client → @ag-ui/encoder → @ag-ui/proto
├── @copilotkit/runtime-client-gql@1.56.5
│   ├── @urql/core@5.2.0
│   │   └── @0no-co/graphql.web@1.2.0
│   └── wonka@6.3.6
├── @copilotkit/shared@1.56.5
│   ├── graphql@16.13.2
│   └── @copilotkit/license-verifier@0.2.0
└── @copilotkit/web-inspector@1.56.5
    └── @copilotkit/core (duplicated subtree)
```

[...]

PR #5883, opened 2026-07-08
fix(react-core): expose lean hooks from /v2/headless to cut bundle bloat (#4893)

tylerslaton:
## What & why

`@copilotkit/react-core`'s `/v2` entry (and the root entry) re-exports from a
single **monolithic shared chunk**, so importing *any* symbol — even one hook —
pulls the built-in chat-message rendering stack (`streamdown` → shiki, `mermaid`,
`cytoscape`, `katex`) into the consumer's bundle: **~3 MB gzip / ~15 MB raw**,
with no way to tree-shake it. Consumers who build a fully custom chat UI and only
use hooks pay the full cost. (Issue #4893.)

A separate lean build entry — `@copilotkit/react-core/v2/headless` — already
ships those hooks in their own small chunk **without** that stack (it's how
`@copilotkit/react-native` mounts CopilotKit). This PR makes it usable for a
custom web UI.

Measured with esbuild (react/react-dom external):

| import | bundled JS |
|---|---|
| hooks from `@copilotkit/react-core/v2` | **~2.96 MB** gzip (≈ importing full `CopilotChat`) |
| same hooks from `@copilotkit/react-core/v2/headless` | **~0.03 MB** gzip |

> Note on the report: `@copilotkit/a2ui-renderer` doesn't bundle the rich-text
> stack (its deps are `@a2ui/web_core`, `lit`, `clsx`, `zod`). The weight is
> entirely `streamdown` (shiki/mermaid/cytoscape) + `katex`, pulled by the
> built-in `CopilotChat*` message components.

## Changes

- **`headless.ts`** — export `useCopilotKit` + `useRenderToolCall` (both DOM-free
  and rendering-stack-free); fix `UseAgentUpdate` (a runtime `enum`) being
  re-exported via `export type`, which stripped its runtime value under
  `isolatedModules` (so `useAgent`'s `updates` option was unusable from headless
  — a bug `tsc` can't catch). `useDefaultRenderTool` / `useRenderCustomMessages`
  / `useRenderActivityMessage` stay in `/v2` (web-only markup or `a2ui-renderer`).
- **Remove a `tailwind-merge` leak** — extract the tailwind-free ref helpers
  (`shallowEqual` / `useShallowStableRef`) into `lib/shallow-stable-ref.ts` so the
  headless graph no longer pulls `tailwind-merge` via
  `CopilotChatConfigurationProvider`.
- **Test** — a small vitest export-surface test guarding the hook surface and the
  `UseAgentUpdate` runtime value.

## Verification

`react-core` typecheck + vitest (1427) and `react-native` typecheck pass. Shipped
`dist/v2/headless.mjs` imports only `react`, `@ag-ui/client`, `@copilotkit/core`,
`@copilotkit/shared`, `@copilotkit/react-core/v2/context`, `zod` — no rendering
stack.

## Notes / follow-ups

- Headless hooks read a **different React context** than the prebuilt `/v2`
  `<CopilotKitProvider>`, so the two can't be mixed — mount a lean provider over
  `/v2/context` (as `@copilotkit/react-native` does). Making the `/v2` provider
  reuse the standalone `/v2/context` singleton would remove that sharp edge and
  is the natural follow-up.
- Existing bundle-size CI (`compressed-size-action`) already tracks `headless.mjs`,
  so no new size tooling is added here.

Addresses #4893.

### DIFF
--- a/packages/react-core/src/v2/__tests__/headless-exports.test.ts
+++ b/packages/react-core/src/v2/__tests__/headless-exports.test.ts
@@ -1,0 +1,54 @@
+import { describe, it, expect } from "vitest";
+import * as headless from "../headless";
+
+// Guards the two regressions issue #4893 fixed that neither the bundle guard nor
+// `tsc` catches: (1) a documented hook silently dropped from the entry, and
+// (2) UseAgentUpdate — a runtime enum — re-exported via `export type`, which
+// type-checks clean but strips the runtime binding under isolatedModules, leaving
+// useAgent's `updates` option undefined at runtime.
+//
+// Typing the name lists as `(keyof typeof headless)[]` also fails `tsc` if a
+// listed export is removed, so the surface is guarded at build time too.
+describe("@copilotkit/react-core/v2/headless runtime exports", () => {
+  it("exports the documented hooks as runtime functions", () => {
+    const hooks: (keyof typeof headless)[] = [
+      "useCopilotKit",
+      "useAgent",
+      "useFrontendTool",
+      "useComponent",
+      "useHumanInTheLoop",
+      "useInterrupt",
+      "useSuggestions",
+      "useConfigureSuggestions",
+      "useAgentContext",
+      "useThreads",
+      "useRenderTool",
+      "useRenderToolCall",
+      "useCapabilities",
+      "useCopilotChatConfiguration",
+    ];
+    for (const name of hooks) {
+      expect(
+        typeof headless[name],
+        `${name} should be a runtime function`,
+      ).toBe("function");
+    }
+  });
+
+  it("exports the provider, core class, and helpers as runtime values", () => {
+    const values: (keyof typeof headless)[] = [
+      "CopilotKitCoreReact",
+      "CopilotChatConfigurationProvider",
+      "CopilotChatDefaultLabels",
+      "defineToolCallRenderer",
+    ];
+    for (const name of values) {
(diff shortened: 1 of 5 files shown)


BenTaylorDev (approved):
Approving — verified end to end, not just read:

- **Bundle claim confirmed live.** Built the dist and measured `useAgent` imported from each entry (react family external, minified+gzipped): `/v2` = **2.81 MB gzip** (bundles streamdown/shiki/mermaid/cytoscape/katex) vs `/v2/headless` = **64 KB gzip** with zero heavy-stack refs — a ~44× reduction, matching the report.
- **Runtime correctness.** The new `headless-exports.test.ts` passes 3/3, including the guard that `UseAgentUpdate` resolves as a runtime enum — confirming the `export type` → `export` fix (the isolatedModules-stripping bug) actually works at runtime.
- **No leak, no regression.** Traced every headless-reachable module (incl. the re-exported `CopilotChatConfigurationProvider`) — none imports the render stack; the built `headless.mjs` chunk contains none of it. Moving `shallowEqual`/`useShallowStableRef` into the tailwind-free leaf module is backward-compatible via the re-export left in `slots.tsx` (existing `CopilotChat` + tests still resolve).
- CI green (bundle-size + copilotchat-import-size gates included); mergeable on current main.

Clean, surgical, well-documented, and it fixes a real runtime bug on top of the bundle win.

PR #5885, opened 2026-07-08
fix(runtime): stop finalizeRunEvents emitting events after a terminal (#5812)

### LINKED ISSUE #5812: 🐛 Bug: `finalizeRunEvents` appends `TEXT_MESSAGE_END` after `RUN_ERROR` — stop button crashes with "run has already errored"

tylerslaton:
## Summary

Pressing **Stop** while an assistant message is streaming (CopilotRuntime + `HttpAgent` proxy) crashed the chat with:

```
Cannot send event type 'TEXT_MESSAGE_END': The run has already errored with 'RUN_ERROR'. No further events can be sent.
```

Root cause: `finalizeRunEvents` appended a trailing `TEXT_MESSAGE_END` **after** the `RUN_ERROR` that the aborted agent had already emitted.

Fixes #5812.

## Root cause

When the upstream agent (e.g. pydantic-ai's `AGUIAdapter`) is aborted mid-stream it emits a live `RUN_ERROR` while a text message is still open — it does **not** close the message first. All runners (`in-memory`, `intelligence`, `sqlite`) stream `finalizeRunEvents`' output *after* everything the agent already emitted, so the appended closer landed past the terminal:

| | outgoing event order |
|---|---|
| **Before** | `… TEXT_MESSAGE_CONTENT → RUN_ERROR → TEXT_MESSAGE_END` ❌ verifier throws |
| **After** | `… TEXT_MESSAGE_CONTENT → RUN_ERROR` ✅ terminal closes the message client-side |

Per the AG-UI invariant: at most one terminal event per run, and no sub-events after it. I confirmed against the real `@ag-ui/client` `verifyEvents` (the verifier the browser runs) that a terminal arriving with a message still open is valid — the terminal implicitly closes it.

## Fix

`finalizeRunEvents` (in `@copilotkit/shared`) now returns early and appends **nothing** when the stream already contains a terminal event (`RUN_FINISHED` or `RUN_ERROR`). The abrupt-end path (no terminal → close open streams + synthesize a terminal, in the correct order) is unchanged. No API/signature change; the in-memory, intelligence, and sqlite runners all inherit the fix.

## Testing

RED→GREEN verified — each new/updated assertion was confirmed to fail against the pre-fix code:

- **`finalize-events.test.ts`** — terminal-present appends nothing (parametrized over `RUN_FINISHED` and `RUN_ERROR`) + a named #5812 case.
- **`in-memory-runner.test.ts`** — end-to-end mid-stream-stop regression: a fake `HttpAgent`-style agent is stopped between `TEXT_MESSAGE_START` and `TEXT_MESSAGE_END`; asserts no events follow `RUN_ERROR` **and** that the collected stream passes `verifyEvents` (before the fix this threw the exact browser error).
- **`intelligence-runner.test.ts`** — corrected a pre-existing assertion that had encoded the buggy post-terminal `TEXT_MESSAGE_END`.

Green: full `@copilotkit/runtime` suite, `@copilotkit/sqlite-runner`, `@copilotkit/shared`, `check-types`, `oxlint` (0 errors), and build.

## Reviewer notes

- The behavior change is a single early-return in `finalize-events.ts`; the `terminalEventMissing` guards simplify away because they're only reachable when no terminal exists.
- Diff is +204/−55 across 4 files, the bulk of it tests.

### DIFF
--- a/packages/runtime/src/v2/runtime/runner/__tests__/finalize-events.test.ts
+++ b/packages/runtime/src/v2/runtime/runner/__tests__/finalize-events.test.ts
@@ -1,10 +1,10 @@
 import { describe, expect, it } from "vitest";
-import {
+import type {
   BaseEvent,
-  EventType,
   ToolCallResultEvent,
   RunErrorEvent,
 } from "@ag-ui/client";
+import { EventType } from "@ag-ui/client";
 import { finalizeRunEvents } from "@copilotkit/shared";
 
 const createTextStart = (messageId: string): BaseEvent =>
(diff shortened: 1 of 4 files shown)


BenTaylorDev (approved):
The fix is correct and well-scoped, and the tests are genuinely strong. Verified the key points:

- The early return (`if (hasTerminalEvent) return appended;`) sits after the scan loop that only *builds* open-message/tool-call state, so `appended` is still `[]` — appending nothing is correct, and no needed work is skipped. Happy path and the abrupt-end (no-terminal → close + synthesize) path are unchanged.
- The load-bearing invariant (a terminal may arrive with a message still open; the terminal implicitly closes it) is proven by the issue's own repro: `verifyEvents` accepted the `RUN_ERROR` while the message was open and only threw on the *subsequent* `TEXT_MESSAGE_END`. Dropping that trailing END is exactly right.
- The `in-memory-runner` e2e test drives the real `@ag-ui/client` `verifyEvents` over a mid-stream stop and asserts nothing follows `RUN_ERROR` — a real regression lock, not a rubber-stamp.

**Non-blocking follow-up (not for this PR):** for an agent that self-emits `RUN_ERROR` code `abort` on stop (the pydantic-ai case here), the client still processes that terminal → `onRunErrorEvent` (`run-handler.ts:1005`) → `emitAgentError(AGENT_RUN_ERROR_EVENT, runtimeErrorCode:"abort")` → CopilotChat's internal error handler (`Chat.tsx:566-570`) → `triggerChatError` → the default error banner ("This operation was aborted"). There's no `code:"abort"` suppression at any hop (the only abort filter is for the local fetch-abort rejection at `run-handler.ts:318`, a different path). So after this fix the crash and broken chat state are gone and the stream is valid — a strict improvement — but a user-initiated Stop still renders as an error banner. Worth a separate issue: suppress `AGENT_RUN_ERROR_EVENT` for `code:"abort"` on user stop (in core's `onRunErrorEvent` or react-ui's `triggerChatError`). Doesn't block this PR.

Issue #228, opened 2024-02-19
CopilotTask does not continue execution after returning a response


ataibarkai:
**Describe the bug**
The CopilotTask successfully calls a function from the global context. But it does not pass the function's return value back into the execution context.

**To Reproduce**
- In the powerpoint presentation example, remove the words "do not carry research" from the task.
- See how a research task is successfully initiated when executing the CopilotTask.
- Once the research concludes, the task goes inert. The response is not fed back to the copilot engine, and a new slide is not appended.

**Expected behavior**
Same behavior as the chatbot.

jpr5:
CopilotTask and the underlying execution model have been significantly reworked since early 2024. If you're still hitting this with a current version, please open a new issue with repro steps against the latest SDK.

PR #5891, opened 2026-07-09
feat(telemetry): emit telemetry-registry fragments for runtime + docs surfaces

BenTaylorDev:
## What

Adds the CopilotKit side of the [telemetry event registry](https://github.com/CopilotKit/oss-path-to-production/blob/main/docs/telemetry-registry-publish-roadmap.md): tooling + CI that generate this repo's registry **fragments** and open path-limited PRs into `CopilotKit/oss-path-to-production`, where the reconciler folds them into `telemetry-events.json`.

Two surfaces, two mechanisms (per the surface-owns-its-extractor design):

| Surface | Events | Extraction | Trigger |
|---|---|---|---|
| **runtime** | 5 `oss.runtime.*` | **bespoke catalog** — reads the `AnalyticsEvents` type map (names + properties), scans `capture()` sites for `call_sites`; **fails loud if the v1/v2 catalogs diverge** | stable **monorepo** release (`on: release`, tag `vX.Y.Z`) |
| **docs** (`showcase/shell-docs`) | 11 | **callee mode** — inline `posthog.capture("name", {…})` literals; drops `$`-reserved events | push to `main` touching `showcase/shell-docs/src/**` (excluding `src/content`) |

## Key properties

- **Content-gated.** The emitter leaves the target fragment byte-for-byte untouched when the extracted event set is unchanged, so a PR opens **only when telemetry actually changes** — no per-release / per-commit churn.
- **Reconciled canonical in every PR.** Both workflows run the registry's `pnpm reconcile` and commit `telemetry-events.json` alongside the fragment, matching the registry's shipped emitters — a fragment-only PR fails its `telemetry-reconcile` staleness gate.
- **Least-privilege cross-repo token.** No explicit `owner` (defaults to the app installation's org) + bare `repositories: oss-path-to-production` + `contents`/`pull-requests` write only; mint gated on a job-level env var (GitHub rejects `secrets.*` in `if:`).
- **zizmor clean** at CI's `--min-severity low` (one `cache-poisoning` suppression, justified in `.github/zizmor.yml`: the workflow configures no cache and publishes a PR, not build artifacts).

## Files

- `scripts/telemetry/extract.ts` — pure extraction (callee scan + catalog reader), deterministic output.
- `scripts/telemetry/emit-fragment.ts` — CLI: `--surface runtime|docs --out <path>`, assembles + content-gates the fragment.
- `scripts/__tests__/telemetry-fragment.test.ts` — 13 unit tests (fixtures) + a loose real-catalog drift smoke test.
- `.github/workflows/telemetry-{runtime,docs}-fragment.yml` — the two CI jobs.

## Testing

Rebased onto `main` (`55aaad21a6`) and revalidated end-to-end on 2026-08-05 — the branch had fallen 1345 commits behind.

**Unit / static**

- `vitest run scripts/__tests__/telemetry-fragment.test.ts` → **13/13 passed**.
- `tsc --noEmit --strict --esModuleInterop` over both scripts → **clean** (`scripts/` has no tsconfig, so this is the ad-hoc invocation).
- `oxlint scripts/telemetry` → **0 warnings, 0 errors**; `oxfmt --check` → **all files correctly formatted**.
- `zizmor --min-severity low --config .github/zizmor.yml .github/workflows` (CI's exact invocation) → **No findings to report** (32 ignored, 233 suppressed).

**Runtime surface — mechanism proven against the live rebased tree**

```
$ tsx scripts/telemetry/emit-fragment.ts --surface runtime --out /tmp/CopilotKit.runtime.json
runtime: wrote 5 events → /tmp/CopilotKit.runtime.json (released_in runtime@1.66.2)
```

Diffed event-for-event against the registry's committed `CopilotKit.runtime.json`: **semantically identical** (same 5 events, same `call_sites`, same `properties_seen`) — the only difference is ordering, since the emitter sorts alphabetically and the hand-seeded fragment is in catalog-declaration order. Confirmed the reorder is a no-op at the canonical level (see below), so the first automated run opens one reordering PR with an empty `telemetry-events.json` diff and is quiet thereafter.

Also confirmed the catalog is still complete on current `main`: the only `oss.*` event literals anywhere under `packages/runtime/src` + `packages/shared/src` are the 5 catalog entries (43/22/12/9/9 occurrences), so no untyped event is being silently dropped. Both v1 and v2 catalogs remain byte-identical, so the divergence guard passes.

**Docs surface**

```
$ tsx scripts/telemetry/emit-fragment.ts --surface docs --out /tmp/CopilotKit.docs.json
docs: wrote 11 events → /tmp/CopilotKit.docs.json (released_in shell-docs@5855496103)
```

11 events (up from 7 when this PR was authored — the docs site grew): `cli_command_copied`, `docs_conversion_clicked`, `docs_conversion_copied`, `docs.framework_selected`, `docs.frontend_selected`, `docs.journey_continued`, `hero_command_copied`, `markdown_copied`, `open_in_llm_clicked`, `talk_to_us_clicked`, `try_for_free_clicked`. `$pageview` correctly dropped.

**End-to-end against the real registry**

Dropped both emitted fragments into a clean `oss-path-to-production@main` worktree and ran its own `pnpm reconcile`:

- Both fragments **validate against `fragment.schema.json`** (ajv, via the reconciler's loader).
- Reconcile succeeded; `telemetry-events.json` grew by 216 lines with 11 new `"surface": "docs"` observations.
- **Zero `oss.runtime.*` entries changed** — confirming the runtime fragment's reordering has no canonical effect.

## Fixed during revalidation

- **`add-paths` bug in the docs workflow (would have failed on first run).** It ran `pnpm reconcile` but listed only the fragment in `add-paths`, so its PR would have landed a fresh fragment beside a stale `telemetry-events.json` and tripped the registry's `telemetry-reconcile` staleness gate — the exact failure the runtime workflow was already fixed for. Verified against the registry's shipped emitters: every automated fragment PR there (`website.corp` #232/#220, Intelligence surfaces #228) carries `telemetry-events.json` alongside its fragment.
- **Stale action pins.** Refreshed to the SHAs `main` now uses everywhere: `actions/checkout` v7, `actions/setup-node` v7.0.0, `pnpm/action-setup` v6.0.10.
- **Over-broad docs trigger.** Narrowed from `showcase/shell-docs/**` to the code under `src/**`, excluding `src/content/**` — 1012 MDX + 140 JSON prose files with zero `.ts`/`.tsx`, none of which can hold a `posthog.capture` call site. Prose edits no longer fire a full monorepo install.
- **zizmor justification accuracy.** `setup-node` v7 adds a `package-manager-cache` input defaulting to `true`; per its `action.yml` it engages only when `package.json` declares **npm**, and this repo declares pnpm — so the workflow is still cacheless and the suppression still holds. Noted inline.

## Prerequisite — now satisfied

The registry App secrets (`TELEMETRY_REGISTRY_APP_ID`, `TELEMETRY_REGISTRY_APP_PRIVATE_KEY`) are configured on this repo (added 2026-07-09), and `app/copilotkit-telemetry-bot` is demonstrably installed on `oss-path-to-production` — it has been opening fragment PRs there from other surfaces (#232, #228, #220). No further setup needed.

## Not in this PR

- The registry-side seed of the **docs** surface. The docs fragment first appears via this workflow's initial run, which now also carries the reconciled canonical, so it lands green.
- The **web-inspector** surface, hand-seeded in the registry since this PR was authored, remains manual. Automating it is a follow-up.

### DIFF
--- a/.github/workflows/telemetry-docs-fragment.yml
+++ b/.github/workflows/telemetry-docs-fragment.yml
@@ -1,0 +1,95 @@
+# telemetry / docs fragment
+#
+# Regenerates the `CopilotKit.docs` telemetry-registry fragment (the
+# showcase/shell-docs site) and opens a path-limited PR against
+# CopilotKit/oss-path-to-production whenever the docs site's telemetry could
+# have changed. Callee-mode extraction (inline posthog.capture), content-gated:
+# if the extracted event set is unchanged the fragment is left untouched and no
+# PR is opened, so routine docs code changes don't churn the registry.
+name: telemetry / docs fragment
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - "showcase/shell-docs/**"
+      - "scripts/telemetry/**"
+      - ".github/workflows/telemetry-docs-fragment.yml"
+  workflow_dispatch: {}
+
+permissions:
+  contents: read
+
+concurrency:
+  group: telemetry-docs-fragment
+  cancel-in-progress: false
+
+jobs:
+  fragment:
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    permissions:
+      contents: read
+    # GitHub rejects secrets.* in `if:`; surface the gate as an env var.
+    env:
+      HAS_BOT_KEY: ${{ secrets.DEVOPS_BOT_PRIVATE_KEY != '' }}
+    steps:
+      - name: Checkout CopilotKit
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+        with:
+          persist-credentials: false
+
+      - name: Setup pnpm
+        uses: pnpm/action-setup@0e279bb959325dab635dd2c09392533439d90093 # v6.0.8
+
+      - name: Setup Node
+        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
+        with:
+          node-version: 22.x
+
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile
+
+      # Cross-repo token, least privilege: NO explicit owner + bare
+      # `repositories`; contents+pull-requests write only.
+      - name: Mint token for oss-path-to-production
+        if: env.HAS_BOT_KEY == 'true'
+        id: token
+        uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1 # v3.2.0
+        with:
+          app-id: 1108748
+          private-key: ${{ secrets.DEVOPS_BOT_PRIVATE_KEY }}
+          repositories: oss-path-to-production
+          permission-contents: write
+          permission-pull-requests: write
+
+      - name: Checkout oss-path-to-production
+        if: steps.token.outputs.token != ''
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+        with:
+          repository: CopilotKit/oss-path-to-production
+          token: ${{ steps.token.outputs.token }}
+          path: p2p
+          persist-credentials: false
+
+      - name: Generate docs fragment
+        if: steps.token.outputs.token != ''
+        run: pnpm tsx scripts/telemetry/emit-fragment.ts --surface docs --out p2p/telemetry-registry/fragments/CopilotKit.docs.json
+
+      - name: Open fragment PR
+        if: steps.token.outputs.token != ''
+        uses: peter-evans/create-pull-request@5f6978faf089d4d20b00c7766989d076bb2fc7f1 # v8
+        with:
+          token: ${{ steps.token.outputs.token }}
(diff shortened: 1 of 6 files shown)


BenTaylorDev:
Correction: switched the cross-repo token from `DEVOPS_BOT` (CopilotKit's release bot, wrong app) to the dedicated **telemetry-registry GitHub App** — `secrets.TELEMETRY_REGISTRY_APP_ID` + `secrets.TELEMETRY_REGISTRY_APP_PRIVATE_KEY`, the same App the Intelligence CLI release workflow uses (installed on oss-path-to-production only).

**Action required before this automation can open PRs:** add those two secrets to the **CopilotKit** repo. They currently exist only on `CopilotKit/Intelligence` (repo-scoped, not org-level). Until then the mint step self-skips (gated on `env.REGISTRY_APP_ID`), so nothing breaks — it just won't produce fragment PRs.

MikeRyanDev (approved):

Issue #310, opened 2024-04-22
Support for Sveltekit


anasmohammed361:
**Is your feature request related to a problem? Please describe.**
Port of the React Library in Svelte

**Describe the solution you'd like**
Writing a wrapper or a library for sveltekit

**Describe alternatives you've considered**
Exposing the apis with vanilla Js , So it can be framework agnostic

**Additional context**
Any Contributions are appreciated

yurashiwa:
Any update?

SPRIME01:
Any update?

tylerslaton:
Hey all, we will eventually do this. If anyone would like to do this work please feel free to submit a PR but the path we will go down is to implement a Vanilla JS layer that we then port React, Svelte, ..., <insert JS framework here> to use.

Right now the core team is really heads-down on polishing everything else and we'll come back to this this year.

orefalo:
same request here

sinaatalay:
This would be really great.

99bits:
Amazing project.  Please add Sveltekit 5

samvallad33:
SvelteKit support without a durable memory story still leaves agents amnesiac across SSR and client boundaries. Framework wiring alone does not seal session recall. Are you defining where agent memory persists in the SvelteKit lifecycle before claiming support?

orefalo:
> SvelteKit support without a durable memory story still leaves agents amnesiac across SSR and client boundaries. Framework wiring alone does not seal session recall. Are you defining where agent memory persists in the SvelteKit lifecycle before claiming support?

You raise a valid point. The calls to the memory agent are outside the scope of SvelteKit and should be abstracted in a way that allows any implementation to be integrated. I am somewhat surprised that this is only now being raised as an issue, as it seems rather self-evident. In fact to current react implementation should already have it.

PR #5905, opened 2026-07-09
feat(svelte): add initial Svelte SDK support

agrittiwari:
## What does this PR do?

Introduces the initial `@copilotkit/svelte` SDK contribution for CopilotKit, [https://github.com/CopilotKit/CopilotKit/issues/310](https://github.com/CopilotKit/CopilotKit/issues/310) including:

- a new Svelte package built around the V2/headless CopilotKit APIs
- Svelte chat UI primitives such as `CopilotChat`, `CopilotSidebar`, and `CopilotPopup`
- Svelte hook equivalents for the core agent interaction flows
- a Svelte demo app and workspace wiring
- follow-up runtime and hook fixes needed to make the package build cleanly
- the lockfile updates required for the new package and demo workspace

## Scope of the contribution

This PR is intentionally focused on bringing up a substantial first Svelte surface, not full React parity in one pass.

Implemented and testable in the current package:

- basic chat rendering and streaming
- frontend tools via `useFrontendTool`
- agent context via `useAgentContext`
- render tools via `useRenderTool`
- human-in-the-loop flows via `useHumanInTheLoop`
- suggestions, interrupts, attachments, capabilities, and threads
- sidebar and popup chat surfaces

> Added Example app for Sveltekit.
image

## Current parity gaps vs React

Compared with `@copilotkit/react-core` and `@copilotkit/react-ui`, the largest remaining gaps are:

- no V1 legacy hooks such as `useCopilotAction`, `useCopilotReadable`, or `useCopilotChat`
- no CoAgent or LangGraph state-rendering surface
- no generative UI / A2UI surface yet
- no memories or learning hooks
- no modal, threads drawer, audio recorder, or several React-specific UI utilities
- no exported CSS or theming system yet
- no dev-console style debugging surface

In practice, the biggest real-world gaps are CoAgent state rendering, generative UI, and theming/styling ergonomics.

## Why this shape

The Svelte package follows the modern V2 CopilotKit API model rather than trying to reproduce the entire legacy React surface up front. That keeps the initial contribution coherent and useful while leaving room for follow-up parity work in smaller, reviewable steps.

## Testing

Verified during this branch work:

- `packages/svelte` direct package build passes
- `packages/svelte` tests pass
- branch commit history was normalized to conventional commits

Repo-wide hook-driven verification was partially blocked locally because the workspace `pnpm` install state wanted an interactive reinstall of `node_modules`. The PR should therefore be treated as needing fresh CI confirmation on the upstream environment.

## Related PRs and Issues

- Upstream OSS PR: this branch targets `CopilotKit/CopilotKit:main`
- No linked issue in this branch

## Checklist

## Follow-up areas

Natural follow-ups after this initial contribution would be:

- CoAgent and LangGraph state rendering
- generative UI / A2UI support
- CSS and theming exports
- additional React-parity UI components and hooks


agrittiwari:
@tylerslaton 
Hey Tyler, I wanted to know if I should wait for this to get merged or push more changes to the same branch for the leftover. A direction would be really helpful.

NathanTarbert:
Hey @agrittiwari, this is great, a Svelte SDK has been on the wishlist for a while and we've had folks from the community request it. I'm really glad to see someone take a real swing at it. Thanks for putting in the work here.

Went through the diff and also spun up the demo locally to poke at a few things live (mocked the agent using AIMock so I could drive streaming/tool-call events deterministically). Found a handful of things worth fixing before merge - nothing that says "don't do this," just things that'll bite real users if they land as-is.

**Bugs I'd fix before merge:**

1. **`use-agent.svelte.ts:209`** — the state-change handlers reassign `agent` to the same object reference, and Svelte 5's `$state` bails out on identity-equal reassignment. So when a co-agent streams a `STATE_DELTA`, nothing re-renders — the UI just goes stale mid-run. React's version uses `forceUpdate()`, Vue's uses `triggerRef()` — this needs the same kind of forcing mechanism.

2. **Autoscroll doesn't work — confirmed this one live.** `CopilotChat.svelte:20` threads an `autoScroll` prop through, but it's never actually read anywhere, including in the real scroll container (`CopilotChatMessageView`). I mocked a long streaming reply and watched the message list stop 289px short of the bottom once it finished — the user has to scroll down manually every time. Happy to share the repro script if useful.

3. **`CopilotChatAssistantMessage.svelte:255`** — `isLatestAssistantMessage` checks against the literal last item in the whole message array, not the last *assistant* message. Once a tool call's result gets appended (`role: "tool"`), the toolbar (copy/regenerate) can appear on a message that's technically still streaming. Flagging this one as "very likely a bug from reading the code" rather than fully proven live — I tried to reproduce it end-to-end and my mock didn't route through the real tool-execution round-trip, so take it as high-confidence-but-not-triple-checked.

4. **`CopilotChatInput.svelte:433`** — `localValue` seeds once from `value` and never resyncs (there's a `svelte-ignore` suppressing the warning, which is usually a tell). Any parent updating `value` after mount won't be reflected in the textarea.

5. **`CopilotChat.svelte:22`** — `inputValue` is destructured as a prop but never actually passed down to `CopilotChatView`. Controlled input mode is a no-op right now.

6. **`CopilotChat.svelte:39`** — `agentId` gets resolved once at mount instead of reactively, so swapping agents on a live component (like an agent-switcher UI) silently keeps talking to the original one.

7. **`CopilotKitProvider.svelte:275`** — small race in `executingToolCallIds`: it snapshots the next state before a `setTimeout(0)` deferral, so a second tool call starting in that window can get wiped from the set by a stale snapshot.

8. **`use-agent.svelte.ts:31`** — `globalThreadCloneMap` never evicts entries, so an app that pages through a lot of threads over its lifetime will slowly leak agent clones.

9. **`use-agent-context.svelte.ts:70`** — context values passed in as plain objects get frozen at mount since the effect has no reactive dependency on their fields; later `$state` changes never make it to `core.addContext`. The same pattern shows up in `useCapabilities`/`useInterrupt`'s `agentId` forwarding.

**Test coverage** — right now, the only test file covers the core wrapper class; none of the 11 hooks or 16 chat components are covered (streaming, tool calls, HITL, and attachments are all untested). Totally get that this is a first pass establishing the API surface, but some baseline tests on the hooks would go a long way toward catching regressions as this evolves.

**One small thing**: `packages/svelte/scripts/fix-dts-paths.mjs` looks like it exists specifically to patch relative type paths that `vite-plugin-dts` leaves behind, but it's not actually wired into the `build` script. Worth double-checking whether `dist/index.d.ts` needs it — if it does, a standalone `npm install` of `@copilotkit/svelte` could hit broken type resolution.

Didn't find anything scary in the one shared runtime file this touches (`sse-response.ts`) — clean, backward-compatible cleanup, no blast radius concerns for React/Angular consumers.

Again, really appreciate you pushing this forward. Let me know if any of the above need more context or a repro; I'm happy to jump on a call too if that's easier than back-and-forth in comments.

agrittiwari:
Hi @NathanTarbert  Just following up on here. Would be super ahppy to go through the PR over a call.
I would like to extend this support once merged.

agrittiwari:
Hi @tylerslaton, @jpr5, Would like some direction here guys. Whenever you have time.
cc @NathanTarbert.

NathanTarbert:
Hey @agrittiwari, thanks for the patience on this one, and sorry it took so long to come back to you.

First thing I owe you an apology for: this is a fork PR, which means CI needs a maintainer to kick it off manually, and engineering has been shipping constantly, so that's not a great contribution experience. That's on us, not you. I've approved the run, so you'll get real CI on every push from here on.

Since it hadn't run, I pulled the branch down and did the whole thing locally instead. Install, build, tests, svelte-check, eslint, publint and attw all pass clean — 50 tests across 14 files. That last part is the thing I want to call out first, because you went from one test file to fourteen covering hooks, HITL, attachments, threads, interrupts and agent switching. That's what made this round actually reviewable, and it's a real chunk of work. Thank you.

I also walked every item from the last review. All of them are closed — the `agentRevision` counter is exactly the right Svelte equivalent of React's `forceUpdate()`, the assistant-message check filters on role properly now, `agentId` is reactive, the `executingToolCallIds` race is gone, the clone map has a cap, and I watched `fix-dts-paths.mjs` actually run during the build and rewrite the declaration paths. Nice work.

So this round is new ground. A few things I'd want fixed before this merges, then a pile of smaller stuff that can land whenever.

**The one with the widest blast radius:** `package.json:198` adds `"onlyBuiltDependencies": ["vite-plugin-dts"]`. In pnpm that field is a whitelist rather than an addition — once it's set, only the packages listed in it are allowed to run install scripts. That quietly turns off native builds for esbuild, sharp, @swc/core, workerd, lmdb, puppeteer and nx for everyone working in the repo, not just for the Svelte package. The `ignoredBuiltDependencies` block right above it looks like pnpm's own generated output from a local install. Could you drop both? And if `vite-plugin-dts` genuinely needs its install script to run, flag that and we'll sort it separately rather than have it ride along with the SDK.

**Two smaller ones in the same neighbourhood.** `package.json:26` has `demo:svelte` running `pnpm install --frozen-lockfile=false` before the dev server — none of the other demo scripts install, and that flag lets a `pnpm demo:svelte` quietly rewrite the lockfile. And `AGENTS.md` plus `.gitignore` both picked up changes that look like local tooling coming along for the ride. A `git checkout origin/main -- AGENTS.md .gitignore` should clear those and keep the diff Svelte-only.

**The chat input never clears after you send.** Confirmed this live. `CopilotChatInput.svelte:34` calls `onSubmit(trimmed)` and nothing resets `localValue` afterward, not in `send()` and not in `CopilotChat.handleSubmitMessage`. There's a wrinkle worth knowing about while you're in there: `CopilotChatView.svelte` renders two separate input instances, one in the welcome branch on line 58 and one in the message branch on line 77. The first send destroys the welcome one and mounts a fresh one, so it *looks* like it clears — then every send after that keeps the text sitting in the textarea. Worth collapsing those two into a single instance outside the `{#if}` regardless, which would also stop the placeholder drifting between "Type your message..." and "Type a message...".

**Human-in-the-loop can't handle two calls at once.** `register-human-in-the-loop.svelte.ts:32` holds a single `resolvePromise` in closure scope. If the agent opens a second call to the same HITL tool before the first is answered, the second `handler()` overwrites the reference and the first promise never settles, so that tool call hangs for the rest of the run. Keying pending resolvers by `toolCallId` should sort it. Flagging this one as read-from-the-code rather than reproduced live — I didn't have a clean way to drive two concurrent HITL calls through my mock.

**Component styles don't reach dist consumers.** The build emits `dist/svelte.css` with all the compiled `<style>` blocks in it, but `dist/index.mjs` never imports it and it isn't in the exports map either, so nothing can actually reach it. Anyone resolving through the `svelte` condition is fine, since they compile from `src` and get scoped styles inline — but the `import`/`require` path renders completely unstyled. Either fold it into `dist/styles.css` or add the import to the entry.

**A few runtime deps are getting bundled.** `vite.config.ts:30` externalizes the CopilotKit and AG-UI packages plus katex and zod, but `@humanspeak/svelte-markdown`, `marked` and `lucide-svelte` are all runtime dependencies and aren't in the list, so they end up inside the 426kB bundle. The Svelte components among them also get compiled against this package's Svelte version instead of the consumer's, which is the part that'd bite people. Driving the list off `Object.keys(pkg.dependencies)` would be the robust fix, and it's worth widening `"svelte"` to `/^svelte(\/.*)?$/` at the same time so subpaths like `svelte/transition` are covered.

Last blocking one is small: `packages/svelte/package.json` is on `1.62.2` while the rest of the monorepo packages are at `1.67.1`, so that wants aligning before this goes anywhere near the release tooling.

---

Then the smaller stuff, none of which I'd hold the merge for.

Autoscroll is wired up now, but it pins to the bottom unconditionally on every message change (`CopilotChatMessageView.svelte:24`), with no check for where the user actually is. Scroll up to re-read something mid-stream and you get pulled straight back down on the next token. The usual fix is a "user is near the bottom" threshold that suppresses the scroll when they've deliberately moved away.

Context reactivity is most of the way there. `connect-agent-context.svelte.ts:29` reads `context.value` inside the effect, so deep mutations on a `$state` object do track correctly. But the argument itself is a plain object captured at call time, so reassigning `description` or swapping `value` wholesale never propagates. `createCapabilities` and `createSuggestions` already accept a `() => string` for `agentId`, so extending that pattern to `connectAgentContext` and `registerFrontendTool` would make the whole surface consistent.

Controlled input is still one-way — `CopilotChat.svelte:72`'s `handleInputChange` is an empty stub and there's no `onInputChange` on `CopilotChatProps`, so a parent can set the value but never see edits come back. Related: `labels`, `attachments` and `onError` are declared on `CopilotChatProps` but never destructured in the component, so they typecheck and then silently do nothing. I'd either wire them or drop them until they're real, just because a prop that looks supported and isn't makes for a rough first impression.

There's also a bit of a split personality on styling. The message components use Tailwind `cpk:` classes with proper `dark:` variants, while `CopilotChatView`, `CopilotChatInput` and `CopilotChatMessageView` use plain `<style>` blocks with hardcoded light values like `#fff` and `#1f2937`. Dark mode half-works as a result.

Smaller still: `CopilotChat.svelte:63` and `:79` fire `copilotkit.runAgent()` without awaiting or catching, so a failure surfaces as an unhandled rejection instead of anything a user would see. `CopilotKitProvider.svelte:278` and `:281` subscribe with empty handlers and can go. `svelte-core.ts:98`'s `setRenderActivityMessages` doesn't notify subscribers while the two setters right beside it do. And the clone eviction is FIFO rather than LRU, since `getOrCreateThreadClone` doesn't refresh insertion order on a cache hit — low impact at a cap of 50, but easy to fix while you're in there.

One genuine question rather than a note: the demo aliases `@segment/analytics-node` out to a hand-written stub in `examples/v2/svelte/demo/src/lib/segment-stub.js`. I'm guessing `@copilotkit/runtime` wouldn't bundle under SvelteKit SSR without it? If that's what you hit, I'd much rather chase the root cause than ship the stub, partly because it also disables runtime telemetry in the demo. What was the error?

---

Two things you should explicitly not worry about. The `use*` to `create*`/`register*`/`connect*`/`load*` rename reads well on its own and fits Svelte's idiom, but it does diverge from `useAgent` in React, Vue and Angular, which affects how our cross-framework docs line up — and `useCopilotKit` stayed behind as the one `use*` holdout. Don't guess on that one, I'll get you a proper answer from the team, and it's a mechanical rename either way. Separately, `@copilotkit/svelte` isn't registered in `static_compat.yml` or the bundle-size workflow yet, which is ours to sort out once the package settles.

So, where this leaves us: the rework closed every functional bug from the last round and brought real test coverage with it. What's left is three quick reverts, five actual bugs, and polish that can land incrementally.

Once you've pushed the blocking set, I'll get one of our engineers to do a proper review pass — someone deeper in the runtime internals than I am, who can sign off on the packaging and the core integration side of it. And you'll have CI running on your pushes now, so you won't be flying blind like you have been.

This is a solid Svelte SDK and it's the first real one anyone's built for us. Really glad you stuck with it.

If any of the above would be faster to talk through than to write back and forth, I'm still happy to jump on a call — just say the word and I'll set something up.

PR #5907, opened 2026-07-09
feat(channels-intelligence): managed-over-Phoenix launcher + slack managed entrypoint (OSS-406 Phase 1)

BenTaylorDev:
## Summary

**OSS-406 Phase 1** — the missing composition that runs a managed bot over the **Phoenix realtime path**, plus a real consumer of it.

The realtime foundation is live on main (gateway `hosted_bots:project:<id>` channel, Redis fan-out, app-api durable authority + lease fencing), and the SDK primitives — `startManagedBots`, `connectPhoenixHostedBotChannel`, `PhoenixRealtimeTransport` — exist and are unit-tested. But **nothing composed them**, so the managed adapter fell back to its HTTP default and Phoenix was never exercised end-to-end. This adds the launcher **and wires an example to it** so it isn't an unused export.

## Launcher (`channels-intelligence/phoenix-launcher.ts`)

- **`startManagedBotsOnChannel(bots, { channel, scope, runtimeInstanceId, log? })`** — wraps an already-connected channel in a `PhoenixRealtimeTransport` (delivery source + render sink) and starts the bots via `startManagedBots`. Split out so the *behavior* is unit-testable against a fake channel.
- **`startManagedBotsOverPhoenix(bots, config)`** — thin glue: `connectPhoenixHostedBotChannel` → delegate → `disconnect()` on `stop()`.
- **`phoenixEgress`** — fail-loud `EgressSink`. `intelligenceAdapter` is exclusive and, with a render sink wired, routes every `post`/`update`/run-render through it — the generic `EgressSink` must never be hit.

## Consumer (`examples/slack/app/managed.ts`)

A **real caller** of the launcher: the *same* Slack bot as `examples/slack/app/index.ts` — identical agent, tools, context, commands, and turn handlers — run in **managed mode over Phoenix** instead of the native `slack()` adapter. `index.ts` (native/self-hosted) is left untouched; `managed.ts` holds no Slack creds and no public endpoint, just a runtime key + a Phoenix connection. Demonstrates the "same bot, swap the transport" thesis concretely:

```
native:   createBot({ adapters: [ slack({ botToken, appToken }) ] })   // index.ts
managed:  startManagedBotsOverPhoenix([ createBot({ … }) ], { … })     // managed.ts
```

The managed `onMention` handler passes the current message as `prompt` to `runAgent` — see the validation note below for why this is required on the managed path (and not on native).

## Tests

Drive a **real `createBot`** through the **full managed path** over a fake channel:
- a Phoenix-delivered turn → handler runs → `render_event` frame → `complete_requested` (completion **intent**, never a self-ack);
- a throwing handler → `fail` intent (no completion, no ack).

Build + 97 package tests green; `examples/slack` `managed.ts` type-clean.

## Validation — live E2E over the real Phoenix path

This didn't just pass unit tests against a fake channel; the whole loop was driven end-to-end on a real local stack with a **real OpenAI backend**, and I verified the message actually traveled the websocket path (not the HTTP fallback).

**Stack:** Intelligence `main` via docker-compose (postgres, postgres-ops, redis, keycloak, minio, tei, realtime-gateway on `:4401`) + app-api on `:7050` (graphile migrations applied). Managed-bots entitlement was granted via the managed-service path (a stub ops entitlement endpoint + `FF_MANAGED_BOTS=true` on both app-api and the gateway; gateway join otherwise rejects with `disabled_by_feature_flag`).

**Provisioning:** created a managed Slack bot + attached a Slack adapter (fixture workspace/creds) against app-api, then triggered a real `app-mention` event through a fake-Slack provider into app-api's signed ingress.

**What was proven:**
- app-api ingress → Redis publish → gateway leases the delivery and pushes `delivery.available` over Phoenix → `managed.ts` (via the launcher) receives it, runs the **real OpenAI** agent, and streams `render_event` frames back → durable render acceptances written (`run_started`, `text_delta`, `text_end`, `finalize`) → `complete_requested` intent → app-api commits the ack. Delivery ended `succeeded`.
- **It genuinely used the websocket path.** With gateway debug logging bumped, frames showed `HANDLED hosted_bot.render_event.v1 INCOMING ON hosted_bots:project:<id> (SdkChannel)`. All app-api delivery-side calls originated from the gateway's HTTP client (`hackney`), with **zero launcher-originated HTTP delivery calls**, and the launcher contains no HTTP delivery code at all. The render frames went bot → gateway over the Phoenix socket.

**Bug found and fixed during validation (the reason `managed.ts` passes `prompt`):**
The first real run failed with an OpenAI `400 input.messages=0` — the agent received **zero messages**. Root cause: `runAgent` (packages/channels `thread.ts`) only injects a user message when `extra.prompt` is set; otherwise it relies on the adapter's reconstructed history. The **native** path gets away with omitting `prompt` because the native adapter's `getHistory` rebuilds the live thread *including* the triggering message. The **managed** path does not: app-api's `GET /api/bots/history` (`reconstructManagedThreadHistory`) rebuilds only *prior committed* deliveries and structurally excludes the in-flight turn (the current delivery is handed to the SDK as the delivery envelope, not via history). So a managed handler that omits `prompt` drops the newest user message — turn 1 → 0 messages → 400; turn 2+ → the agent answers a *stale* prompt. FakeAgent unit tests never caught it because they don't exercise the history round-trip.

Fixed here by having the managed `onMention` pass `message.contentParts ?? message.text` as `prompt`. This is a **systemic** footgun for every managed entrypoint; it's tracked with a durable-fix recommendation (adapter auto-injects the current message so managed handlers match native) in **OSS-459**.

## Scope

- **This PR:** the launcher composition + its first consumer + unit/contract coverage, **plus the live-stack E2E above** which closes OSS-406's "realtime path is unproven" gap and surfaced/fixed the managed `prompt` bug.
- **Deferred (OSS-459):** Teams managed entrypoint, multi-tenant multiplexing, BYO, promote to a deployable Intelligence managed-bot runtime app, HTTP-path deprecation, shared bot-def extraction, and the durable managed-`prompt` fix.

Refs OSS-406.

### DIFF
--- a/examples/slack/app/managed.ts
+++ b/examples/slack/app/managed.ts
@@ -1,0 +1,128 @@
+/**
+ * Managed (Intelligence-hosted) entrypoint for the same Slack bot as
+ * `app/index.ts`.
+ *
+ * `index.ts` is the SELF-HOSTED variant: it holds the Slack bot/app tokens and
+ * talks to Slack directly via the native `slack()` adapter. This file is the
+ * MANAGED variant: it holds no Slack credentials and no public endpoint — it
+ * connects to the Intelligence realtime-gateway over Phoenix, receives leased
+ * deliveries, and streams render frames back. Intelligence owns the Slack edge
+ * (signed ingress → app-api, egress via the Connector Outbox).
+ *
+ * The bot itself — the agent, tools, context, commands, and turn handlers — is
+ * IDENTICAL to the native bot; only the transport changes. `intelligenceAdapter`
+ * is exclusive, so the managed bot is created WITHOUT a native adapter and
+ * {@link startManagedBotsOverPhoenix} attaches the managed transport.
+ *
+ *   native:   createBot({ adapters: [slack({ botToken, appToken }) ] })  // index.ts
+ *   managed:  startManagedBotsOverPhoenix([ createBot({ … }) ], { … })   // this file
+ *
+ * Run: `pnpm --filter slack-example managed` with the INTELLIGENCE_* env set
+ * (see `.env.example`).
+ */
+import "dotenv/config";
+import { randomUUID } from "node:crypto";
+import { createBot } from "@copilotkit/channels";
+import {
+  defaultSlackTools,
+  defaultSlackContext,
+  SanitizingHttpAgent,
+} from "@copilotkit/channels-slack";
+import { startManagedBotsOverPhoenix } from "@copilotkit/channels-intelligence";
+import { appTools } from "./tools/index.js";
+import { appContext } from "./context/app-context.js";
+import { appCommands } from "./commands/index.js";
+import { senderContext } from "./sender-context.js";
+import { fileIssueSubmit, FILE_ISSUE_CALLBACK } from "./modals/file-issue.js";
+import { closeBrowser } from "./render/browser.js";
+
+const required = (name: string): string => {
+  const v = process.env[name];
+  if (!v) {
+    console.error(`Missing required env var: ${name}`);
+    process.exit(1);
+  }
+  return v;
+};
+
+async function main() {
+  const agentUrl = required("AGENT_URL");
+  const agentHeaders = process.env.AGENT_AUTH_HEADER
+    ? { Authorization: process.env.AGENT_AUTH_HEADER }
+    : undefined;
+
+  const projectId = Number(required("INTELLIGENCE_PROJECT_ID"));
+  if (!Number.isInteger(projectId) || projectId < 0) {
+    console.error(
+      `Invalid INTELLIGENCE_PROJECT_ID: "${process.env.INTELLIGENCE_PROJECT_ID}"`,
+    );
+    process.exit(1);
+  }
+  const botName = required("INTELLIGENCE_BOT_NAME");
+
+  // Same bot as the native example, minus the adapter: the managed transport is
+  // attached by startManagedBotsOverPhoenix. Slack is the only managed provider
+  // here, so it always ships the Slack tools/context (the native example adds
+  // these conditionally per active adapter).
+  const bot = createBot({
+    name: botName,
+    agent: (threadId) => {
+      const a = new SanitizingHttpAgent({ url: agentUrl, headers: agentHeaders });
+      a.threadId = threadId;
+      return a;
+    },
+    tools: [...appTools, ...defaultSlackTools],
+    context: [...appContext, ...defaultSlackContext],
+    commands: appCommands,
+  });
+
+  // Turn + feature handlers — identical to the native example (app/index.ts).
+  bot.onMention(async ({ thread, message }) => {
+    try {
+      await thread.runAgent({
+        context: senderContext(message.user, thread.platform),
+      });
+    } catch (err) {
+      console.error("[bot] agent run failed", err);
+      await thread
+        .post("Sorry — I hit an error handling that. Please try again.")
+        .catch(() => {});
+    }
+  });
+  bot.onModalSubmit(FILE_ISSUE_CALLBACK, fileIssueSubmit);
+  bot.onThreadStarted(async ({ thread, user }) => {
+    if (!user?.name) return;
+    await thread.setSuggestedPrompts([
+      { title: `Triage ${user.name}'s issues`, message: "Triage my open issues" },
+      { title: "What shipped this week?", message: "Summarize what shipped this week" },
+    ]);
+  });
+
+  const handle = await startManagedBotsOverPhoenix([bot], {
+    wsUrl: required("INTELLIGENCE_GATEWAY_WS_URL"),
+    apiKey: required("INTELLIGENCE_API_KEY"),
+    scope: {
+      organizationId: required("INTELLIGENCE_ORG_ID"),
+      projectId,
+      botId: required("INTELLIGENCE_BOT_ID"),
+      botName,
+    },
+    runtimeInstanceId:
+      process.env.INTELLIGENCE_RUNTIME_INSTANCE_ID ??
+      `rti_${randomUUID().replace(/-/g, "")}`,
+    adapter: "slack",
+    log: (msg, meta) => console.log(`[managed] ${msg}`, meta ?? ""),
+  });
+  console.log(`[bot] started managed (Phoenix) as "${botName}" on project ${projectId}`);
+
+  const shutdown = async (signal: string) => {
+    console.log(`\n[bot] received ${signal}, stopping…`);
+    await handle.stop();
+    await closeBrowser().catch(() => {});
+    process.exit(0);
+  };
+  process.on("SIGINT", () => void shutdown("SIGINT"));
+  process.on("SIGTERM", () => void shutdown("SIGTERM"));
+}
+
+void main();
--- a/examples/slack/app/managed.ts
+++ b/examples/slack/app/managed.ts
@@ -67,7 +67,10 @@ async function main() {
   const bot = createBot({
     name: botName,
     agent: (threadId) => {
-      const a = new SanitizingHttpAgent({ url: agentUrl, headers: agentHeaders });
+      const a = new SanitizingHttpAgent({
+        url: agentUrl,
+        headers: agentHeaders,
+      });
       a.threadId = threadId;
       return a;
     },
@@ -93,8 +96,14 @@ async function main() {
   bot.onThreadStarted(async ({ thread, user }) => {
     if (!user?.name) return;
     await thread.setSuggestedPrompts([
-      { title: `Triage ${user.name}'s issues`, message: "Triage my open issues" },
-      { title: "What shipped this week?", message: "Summarize what shipped this week" },
+      {
+        title: `Triage ${user.name}'s issues`,
+        message: "Triage my open issues",
+      },
+      {
+        title: "What shipped this week?",
+        message: "Summarize what shipped this week",
+      },
     ]);
   });
 
@@ -113,7 +122,9 @@ async function main() {
     adapter: "slack",
     log: (msg, meta) => console.log(`[managed] ${msg}`, meta ?? ""),
   });
-  console.log(`[bot] started managed (Phoenix) as "${botName}" on project ${projectId}`);
+  console.log(
+    `[bot] started managed (Phoenix) as "${botName}" on project ${projectId}`,
+  );
 
   const shutdown = async (signal: string) => {
     console.log(`\n[bot] received ${signal}, stopping…`);
--- a/examples/slack/app/managed.ts
+++ b/examples/slack/app/managed.ts
@@ -82,7 +82,14 @@ async function main() {
   // Turn + feature handlers — identical to the native example (app/index.ts).
   bot.onMention(async ({ thread, message }) => {
     try {
+      // Managed history (app-api /api/bots/history) does NOT include the
+      // in-flight turn (unlike native adapters whose getHistory rebuilds the
+      // live thread), so pass the current message explicitly as `prompt` —
+      // otherwise runAgent runs with zero messages. Prefer multimodal parts.
       await thread.runAgent({
+        prompt: message.contentParts?.length
+          ? message.contentParts
+          : message.text,
         context: senderContext(message.user, thread.platform),
       });
     } catch (err) {
--- a/examples/slack/app/managed.ts
+++ b/examples/slack/app/managed.ts
@@ -139,17 +139,19 @@ async function main() {
 
   const shutdown = async (signal: string) => {
     console.log(`\n[bot] received ${signal}, stopping…`);
+    let exitCode = 0;
     try {
       await handle.stop();
     } catch (err) {
       console.error("[bot] error stopping managed runtime", err);
+      exitCode = 1;
     }
     // Browser teardown is best-effort, but still surface a failure rather than
     // swallow it silently.
     await closeBrowser().catch((err: unknown) =>
       console.error("[bot] browser cleanup failed (continuing shutdown)", err),
     );
-    process.exit(0);
+    process.exit(exitCode);
   };
   // A failed shutdown must not vanish — log it and exit nonzero.
   const runShutdown = (signal: string): void => {
--- a/examples/slack/app/managed.ts
+++ b/examples/slack/app/managed.ts
@@ -127,6 +127,10 @@ async function main() {
       process.env.INTELLIGENCE_RUNTIME_INSTANCE_ID ??
       `rti_${randomUUID().replace(/-/g, "")}`,
     adapter: "slack",
+    // DEBUG-ONLY logging. `meta` (and the raw `err` in the onMention catch
+    // above) can contain message content/payloads — the design says telemetry
+    // must not include raw message text. In production, drop this `log` or trim
+    // `meta` to safe fields (ids, counts) before emitting.
     log: (msg, meta) => console.log(`[managed] ${msg}`, meta ?? ""),
   });
   console.log(
--- a/examples/slack/app/managed.ts
+++ b/examples/slack/app/managed.ts
@@ -139,12 +139,39 @@ async function main() {
 
   const shutdown = async (signal: string) => {
     console.log(`\n[bot] received ${signal}, stopping…`);
-    await handle.stop();
-    await closeBrowser().catch(() => {});
+    try {
+      await handle.stop();
+    } catch (err) {
+      console.error("[bot] error stopping managed runtime", err);
+    }
+    // Browser teardown is best-effort, but still surface a failure rather than
+    // swallow it silently.
+    await closeBrowser().catch((err: unknown) =>
+      console.error("[bot] browser cleanup failed (continuing shutdown)", err),
+    );
     process.exit(0);
   };
-  process.on("SIGINT", () => void shutdown("SIGINT"));
-  process.on("SIGTERM", () => void shutdown("SIGTERM"));
+  // A failed shutdown must not vanish — log it and exit nonzero.
+  const runShutdown = (signal: string): void => {
+    shutdown(signal).catch((err: unknown) => {
+      console.error(`[bot] fatal during ${signal} shutdown`, err);
+      process.exit(1);
+    });
+  };
+  process.on("SIGINT", () => runShutdown("SIGINT"));
+  process.on("SIGTERM", () => runShutdown("SIGTERM"));
 }
 
-void main();
+// Fail loud, not silent: surface any stray async error instead of letting it
+// kill the process with no log (mirrors the native entrypoint).
+process.on("unhandledRejection", (reason) => {
+  console.error("[bot] unhandledRejection:", reason);
+});
+process.on("uncaughtException", (err) => {
+  console.error("[bot] uncaughtException:", err);
+});
+
+main().catch((err: unknown) => {
+  console.error("[bot] fatal: failed to start managed runtime", err);
+  process.exit(1);
+});
--- a/packages/channels-intelligence/src/phoenix-launcher.ts
+++ b/packages/channels-intelligence/src/phoenix-launcher.ts
@@ -1,0 +1,132 @@
+import type { Bot } from "@copilotkit/channels";
+import { startManagedBots } from "./runtime.js";
+import type { ManagedBotsHandle } from "./runtime.js";
+import { connectPhoenixHostedBotChannel } from "./phoenix-channel.js";
+import { PhoenixRealtimeTransport } from "./phoenix-transport.js";
+import type {
+  HostedBotChannel,
+  HostedBotRealtimeScope,
+} from "./phoenix-transport.js";
+import type { EgressSink } from "./transports.js";
+
+/**
+ * Phoenix-path egress is vestigial: with a render sink wired (the transport
+ * itself), the adapter routes every `post`/`update` and the run-render stream
+ * through the render sink and never through the generic {@link EgressSink}. Fail
+ * loud if that invariant is ever broken, rather than silently dropping an op.
+ */
+const phoenixEgress: EgressSink = {
+  emit: async () => {
+    throw new Error(
+      "startManagedBotsOverPhoenix: EgressSink.emit was called, but the Phoenix " +
+        "path routes all egress through the render sink — this indicates a " +
+        "wiring bug (the render sink was not set on the adapter).",
+    );
+  },
+};
+
+/** Options for {@link startManagedBotsOnChannel} — an already-connected channel. */
+export interface ManagedBotsOnChannelOptions {
+  /** The joined realtime-gateway bot-IO channel (`hosted_bots:project:<id>`). */
+  channel: HostedBotChannel;
+  /** Authoritative org/project/bot scope echoed on every SDK→gateway envelope. */
+  scope: HostedBotRealtimeScope;
+  /** Stable runtime instance id (`rti_…`), echoed on every envelope. */
+  runtimeInstanceId: string;
+  /** Diagnostic sink for dropped deliveries / transport events. */
+  log?: (message: string, meta?: unknown) => void;
+}
+
+/**
+ * Compose the managed runtime over an already-connected Phoenix channel: wrap
+ * the channel in a {@link PhoenixRealtimeTransport} (delivery source + render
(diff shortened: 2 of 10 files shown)


BenTaylorDev:
## How the managed vs self-hosted path is selected

Worth spelling out since this PR introduces the managed composition alongside the existing native one.

**Nothing at runtime picks it — it's the entrypoint/composition you launch.** There's no env flag or auto-detection; you run the process wired for the mode:

|  | Self-hosted (native) | Managed |
|---|---|---|
| Entrypoint | `app/index.ts` (`pnpm slack`) | `app/managed.ts` (`pnpm managed`) |
| Bot construction | `createBot({ adapters: [ slack({ botToken, appToken }) ] })` | `createBot({ … })` **with no native adapter** |
| Transport attached by | the native `slack()` adapter itself (talks to Slack directly) | `startManagedBotsOverPhoenix([bot], …)` attaches `intelligenceAdapter` |
| Credentials | holds Slack tokens + public endpoint | holds none — just a runtime key + a Phoenix connection |

The enforcing mechanism is that **`intelligenceAdapter` is exclusive**: a managed bot is created *without* any native adapter, and `startManagedBots` / `startManagedBotsOverPhoenix` is what attaches the managed adapter. So the fork is decided at composition time by which factory wraps the bot.

There's a **second** selection *inside* the managed path — HTTP vs Phoenix transport:

- `intelligenceAdapter()` **defaults to the HTTP claim-loop** (`HttpDeliverySource` / `HttpEgressSink`, configured from env) when no transports are injected.
- The launcher **injects a `PhoenixRealtimeTransport`** instead — which is the whole point of this PR. Before it existed, even "managed" silently fell back to HTTP and Phoenix was never exercised end-to-end.

So: **native vs managed** = which entrypoint/adapter you compose; **Phoenix vs HTTP within managed** = whether the launcher injects the Phoenix transport or leaves `intelligenceAdapter` on its HTTP default.

packages/channels-intelligence/src/phoenix-launcher.ts
```diff
@@ -0,0 +1,132 @@
...
+ * managed bot over the realtime path — the transport primitives are unit-tested;
+ * this wires them for a live run. The returned handle's `stop()` stops the bots
+ * and then disconnects the socket.
+ */
+export async function startManagedBotsOverPhoenix(
+  bots: Bot[],
+  config: ManagedPhoenixConfig,
+): Promise<ManagedBotsHandle> {
+  const adapter = config.adapter ?? "slack";
+  const channel = await connectPhoenixHostedBotChannel({
```
tylerslaton, line 172: Small ordering thing: we open the websocket connection here first, and the bot names only get checked afterwards (inside `startManagedBotsOnChannel` → `startManagedBots` → `assertValidBotNames`). So if a bot has a missing or duplicate name, we've already connected and sent a join with `botName: undefined` before the check runs — and when that check throws, this socket never gets closed, so it leaks.

The design also asks us to fail fast on missing/duplicate names. Could we run `assertValidBotNames(bots)` before connecting? Then a bad name fails instantly and we never open a connection we'd have to clean up.

packages/channels-intelligence/src/phoenix-launcher.ts
```diff
@@ -0,0 +1,132 @@
...
+    wsUrl: config.wsUrl,
+    apiKey: config.apiKey,
+    projectId: config.scope.projectId,
+    join: {
+      runtimeInstanceId: config.runtimeInstanceId,
+      declaredBots: bots.map((bot) => ({ botName: bot.name!, adapter })),
```
tylerslaton: This join tells Intelligence the runtime id + each bot's name and adapter, but not much else. The design's "Runtime Activation" section wants a little more so the runtime shows up properly in Intelligence's health view — the environment (dev/staging/prod), versions, and each bot's commands/capabilities.

The nice part: that data is already computed by `buildActivationMetadata` inside `startManagedBots` (it comes back on `handle.metadata`), and this join type already has `runtimeMetadata` + per-bot `renderCapabilities` slots ready for it. Right now it's computed *after* we connect and basically goes unused — worth building it before the join and passing it through here.

packages/channels-intelligence/src/phoenix-launcher.ts
```diff
@@ -0,0 +1,132 @@
...
+  const handle = await startManagedBotsOnChannel(bots, {
+    channel,
+    scope: config.scope,
+    runtimeInstanceId: config.runtimeInstanceId,
+    ...(config.log ? { log: config.log } : {}),
+  });
+  return {
+    ...handle,
+    stop: async () => {
+      await handle.stop();
```
tylerslaton: Tiny robustness nit: if `handle.stop()` throws, the `channel.disconnect()` below won't run and the socket stays open. Wrapping it as `try { await handle.stop() } finally { channel.disconnect() }` would make sure we always close the connection.

examples/slack/app/managed.ts
```diff
@@ -0,0 +1,146 @@
...
+      organizationId: required("INTELLIGENCE_ORG_ID"),
+      projectId,
+      botId: required("INTELLIGENCE_BOT_ID"),
+      botName,
+    },
+    runtimeInstanceId:
+      process.env.INTELLIGENCE_RUNTIME_INSTANCE_ID ??
+      `rti_${randomUUID().replace(/-/g, "")}`,
+    adapter: "slack",
+    log: (msg, meta) => console.log(`[managed] ${msg}`, meta ?? ""),
```
tylerslaton, line 134: Low priority, and it's just example code — but the design says logs/telemetry shouldn't include raw message text or payloads. This `log` prints whatever `meta` is, and the `catch` further up logs the raw `err`, both of which could contain message content. Since people copy this example as a starting point, maybe a short note that these logs are debug-only (or trim `meta` to safe fields).

tylerslaton (commented):
Took a pass against the [Hosted Bots in Intelligence](https://linear.app/copilotkit/project/hosted-bots-in-intelligence-10c2236404ca/overview) design doc. Overall this is a clean, well-scoped Phase 1 — the "same bot, swap the transport" idea reads clearly, and I especially like that the tests pin the delivery contract exactly the way the design describes (completion *intent*, never a self-ack). Nice work.

A few things worth a look, all left inline:

- **The one that matters:** we open the socket connection before checking bot names, so a missing/duplicate name sends a broken join and leaks the socket — and the design specifically asks us to fail fast there.
- The activation join is thinner than the design's "Runtime Activation" section (no environment / versions / commands / capabilities), even though the helper that builds all of that already exists.
- A small `stop()` cleanup nit, and a low-priority logging note in the example.

One bigger-picture question (not blocking): the design's target API is `new CopilotRuntime({ intelligence, bots: [bot] })`, but this ships a standalone `startManagedBotsOverPhoenix(...)` entrypoint. Totally reasonable for Phase 1 — I just want to make sure the eventual move to the `CopilotRuntime({ bots })` shape is tracked somewhere, since OSS-459 looks like it covers the other follow-ups but not that one. (The CI timeout bump looks unrelated but harmless.)

Thanks Ben! 🙌

packages/channels-intelligence/src/phoenix-launcher.ts
BenTaylorDev: Good catch — fixed in d33f95d8. `assertValidBotNames(bots)` now runs **before** `connectPhoenixHostedBotChannel`, so a missing/duplicate/invalid name throws instantly and no socket is ever opened. Added a test that a duplicate name rejects while the injected WebSocket ctor is never constructed.

packages/channels-intelligence/src/phoenix-launcher.ts
BenTaylorDev: Done in d33f95d8. The launcher now builds `buildActivationMetadata` **before** connecting and threads it into the join: per-bot `commands` plus `runtimeMetadata` (`runtimeEnv`, `nodeVersion`, and package versions when supplied). It also forwards the same `env` overrides into `startManagedBots` so `handle.metadata` agrees with what we declared on join. `renderCapabilities` is left as a reserved slot — bots don't expose capabilities yet (the `buildActivationMetadata` TODO(OSS-377)); package versions populate once a caller passes them via the new `env` override.

packages/channels-intelligence/src/phoenix-launcher.ts
BenTaylorDev: Fixed in d33f95d8 — now `try { await handle.stop() } finally { channel.disconnect() }`, so the socket always closes even if stopping the bots throws.

examples/slack/app/managed.ts
BenTaylorDev: Addressed in d33f95d8 — added a DEBUG-ONLY note on the example's `log` (and flagged the raw `err` in the onMention catch): `meta`/`err` can carry message content, so drop or trim to safe fields (ids/counts) in production.

BenTaylorDev:
Thanks for the thorough pass, Tyler! All four inline comments addressed in `d33f95d8`:
- **fail-fast names** — `assertValidBotNames` moved ahead of the socket connect (+ a test proving no connection opens on a bad name);
- **activation join** — `buildActivationMetadata` built pre-connect and passed into the join (`runtimeMetadata` + per-bot `commands`), with `handle.metadata` kept in sync;
- **stop() cleanup** — `try/finally` around `channel.disconnect()`;
- **example logging** — debug-only caveat added.

On the bigger-picture question — the design's target `new CopilotRuntime({ intelligence, bots: [bot] })` shape vs this PR's standalone `startManagedBotsOverPhoenix(...)`: agreed it should be tracked. I've added it to **OSS-459** (Phase Next) as an explicit API-convergence follow-up. Phase 1 intentionally ships the standalone composition; converging on `CopilotRuntime({ bots })` is a Phase-2 surface item.

packages/channels-intelligence/src/phoenix-launcher.ts
```diff
@@ -0,0 +1,193 @@
...
+ * part with behavior — is unit-testable against a fake channel, leaving the
+ * socket connect as thin glue. `intelligenceAdapter` is exclusive, so the
+ * Phoenix transport is each bot's ONLY adapter; egress is served by the render
+ * sink, not the generic {@link EgressSink} (see {@link phoenixEgress}).
+ */
+export async function startManagedBotsOnChannel(
+  bots: Bot[],
+  opts: ManagedBotsOnChannelOptions,
+): Promise<ManagedBotsHandle> {
+  const transport = new PhoenixRealtimeTransport({
```
tylerslaton, line 86: Blocking: this creates one stateful `PhoenixRealtimeTransport` and returns that same object for every bot. Each attached `IntelligenceAdapter` calls `source.start(...)`; `PhoenixRealtimeTransport.start()` overwrites its single `onDelivery` callback while `channel.on(...)` appends another listener. With two valid bots, every listener therefore dispatches through the last bot's callback (potentially multiple times), while earlier bots receive nothing.

Could we either add real bot-name routing/multiplexing here, or fail fast on `bots.length !== 1` for Phase 1? Whichever contract we choose needs a two-valid-bot regression test; the duplicate-name test does not exercise this path.

packages/channels-intelligence/src/phoenix-launcher.ts
```diff
@@ -0,0 +1,193 @@
...
+        commands: Object.fromEntries(
+          activation.declaredBots.map((b) => [b.name, b.commands]),
+        ),
+      },
+      observedAt: new Date().toISOString(),
+    },
+    ...(config.timeoutMs !== undefined ? { timeoutMs: config.timeoutMs } : {}),
+    ...(config.webSocket !== undefined ? { webSocket: config.webSocket } : {}),
+  });
+  const handle = await startManagedBotsOnChannel(bots, {
```
tylerslaton: Blocking: name validation now happens before connect, and `stop()` cleans up after a handle exists, but the gap between them can still leak the socket. If `startManagedBotsOnChannel()` throws after the channel joined (for example, because a bot was already started or has a conflicting adapter), this function rejects without disconnecting. Please wrap startup in a `try`/`catch` that disconnects before rethrowing. The lower-level connector should also disconnect its socket when the join rejects or times out, since callers never receive a handle in those cases.

packages/channels-intelligence/src/phoenix-launcher.ts
```diff
@@ -0,0 +1,193 @@
...
+  assertValidBotNames(bots);
+
+  // Build activation metadata up front so the join carries the Runtime
+  // Activation data Intelligence's health view expects (runtime env, node
+  // version, per-bot commands) rather than just name+adapter. The same
+  // `envOverrides` is forwarded to startManagedBots so `handle.metadata` agrees
+  // with what we declared on join.
+  const envOverrides: Partial<ActivationEnv> = {
+    runtimeInstanceId: config.runtimeInstanceId,
+    ...(config.env ?? {}),
```
tylerslaton, line 164: `config.env` can include `runtimeInstanceId` because `ActivationEnv` exposes it, and this spread wins over the required `config.runtimeInstanceId`. The join and realtime envelopes still use `config.runtimeInstanceId`, while `handle.metadata` can use the overridden value, breaking the stated kept-in-sync invariant. Can we make the required config ID authoritative (spread `env` first), or omit `runtimeInstanceId` from this public `env` type, and cover the mismatch in an activation test?

examples/slack/app/managed.ts
```diff
@@ -0,0 +1,150 @@
...
+ *
+ * The bot itself — the agent, tools, context, commands, and turn handlers — is
+ * IDENTICAL to the native bot; only the transport changes. `intelligenceAdapter`
+ * is exclusive, so the managed bot is created WITHOUT a native adapter and
+ * {@link startManagedBotsOverPhoenix} attaches the managed transport.
+ *
+ *   native:   createBot({ adapters: [slack({ botToken, appToken }) ] })  // index.ts
+ *   managed:  startManagedBotsOverPhoenix([ createBot({ … }) ], { … })   // this file
+ *
+ * Run: `pnpm --filter slack-example managed` with the INTELLIGENCE_* env set
```
tylerslaton, line 20: This points users at `.env.example`, but that file contains none of the managed `INTELLIGENCE_*` settings consumed below: gateway URL, API key, org/project/bot IDs, bot name, or the optional runtime instance ID. Could we add a managed section there so this new consumer is runnable from its documented instructions?

tylerslaton (changes requested):
Second pass after d33f95d8. The original four comments are addressed in their intended cases. I found two remaining launch-correctness issues that should block merge: shared-source multi-bot routing, and socket cleanup when startup fails after the channel connects. I also left two smaller API/docs comments around runtime ID consistency and the missing managed environment setup.

The scoped Nx typechecks, tests, builds, format/lint, publint, and ATTW checks are green, so this round is about behavior rather than CI.

packages/channels-intelligence/src/phoenix-launcher.ts
BenTaylorDev: Fixed in 852bd04a — went with fail-fast for Phase 1 (the option you offered). Added `assertSingleBotForPhase1` (throws on `bots.length !== 1`) in **both** `startManagedBotsOnChannel` (the shared-transport site) and `startManagedBotsOverPhoenix` (before connect, so no socket opens for a multi-bot call). Real bot-name routing/multiplexing over a shared channel is tracked as Phase 2 in OSS-459. Added two-valid-bot regression tests on both entrypoints — a distinct path from the duplicate-name test.

packages/channels-intelligence/src/phoenix-launcher.ts
BenTaylorDev: Fixed in 852bd04a, both layers you flagged: (1) `startManagedBotsOverPhoenix` now wraps the post-join `startManagedBotsOnChannel` in `try/catch` → `channel.disconnect()` → rethrow, so a throw *after* the channel joined no longer leaks the socket; (2) `connectPhoenixHostedBotChannel` now calls `socket.disconnect()` on both the join `error` and `timeout` before rejecting, since the caller never receives a channel it could close in those cases.

packages/channels-intelligence/src/phoenix-launcher.ts
BenTaylorDev: Fixed in 852bd04a — good catch, that was a regression from the r1 change. Two parts: the public `ManagedPhoenixConfig.env` type now `Omit`s `runtimeInstanceId` so a caller can't set it via `env` at all; and the merge spreads `config.env` **first**, then `runtimeInstanceId` last, keeping the required id authoritative for the join *and* `handle.metadata`. Added an activation test asserting `env` overrides flow into `handle.metadata` (the join↔metadata sync plumbing).

examples/slack/app/managed.ts
BenTaylorDev: Fixed in 852bd04a — added a **Managed (Intelligence-hosted) mode** section to `examples/slack/.env.example` with every var `managed.ts` reads: `INTELLIGENCE_GATEWAY_WS_URL`, `_API_KEY`, `_ORG_ID`, `_PROJECT_ID`, `_BOT_ID`, `_BOT_NAME`, and optional `_RUNTIME_INSTANCE_ID` (noting `AGENT_URL`/`AGENT_AUTH_HEADER` above are reused as the agent brain).

BenTaylorDev:
Round 2 addressed in `852bd04a` — both blockers + both smaller items:
- **Multi-bot routing (blocking):** fail-fast `bots.length !== 1` guard in both entrypoints (Phase-1 single-bot; multiplexing → OSS-459 Phase 2), with two-valid-bot regression tests.
- **Socket cleanup on startup failure (blocking):** `try/catch → channel.disconnect()` around post-join startup, **and** the connector now `socket.disconnect()`s on join error/timeout.
- **runtimeInstanceId consistency:** `env` type now excludes it + required id spread last (authoritative); activation test added.
- **Managed `.env.example` section** added.

101 package tests pass (+3), `tsc` clean. Re-requesting review — thanks again Tyler 🙏

packages/channels-intelligence/src/phoenix-launcher.ts
```diff
@@ -0,0 +1,227 @@
...
+  /** The joined realtime-gateway bot-IO channel (`hosted_bots:project:<id>`). */
+  channel: HostedBotChannel;
+  /** Authoritative org/project/bot scope echoed on every SDK→gateway envelope. */
+  scope: HostedBotRealtimeScope;
+  /** Stable runtime instance id (`rti_…`), echoed on every envelope. */
+  runtimeInstanceId: string;
+  /** Activation env overrides forwarded to the runtime (so `handle.metadata`
+   * matches what the caller declared on join); omitted fields are gathered from
+   * the process. */
+  env?: Partial<ActivationEnv>;
```
tylerslaton: The common `startManagedBotsOverPhoenix` path now keeps the runtime ID authoritative, but this exported sibling still permits the same mismatch: `env` accepts `runtimeInstanceId`, the transport uses `opts.runtimeInstanceId`, and `opts.env` is forwarded unchanged into `handle.metadata`. A direct caller can therefore emit envelopes as `rti_A` while metadata reports `rti_B` (or omits the ID entirely).

Could we apply the same contract here: exclude `runtimeInstanceId` from this `env` type, merge `opts.runtimeInstanceId` into the activation env inside `startManagedBotsOnChannel`, and assert `handle.metadata.runtimeInstanceId === opts.runtimeInstanceId` in the regression test?

packages/channels-intelligence/src/phoenix-launcher.ts
```diff
@@ -0,0 +1,227 @@
...
+  let handle: ManagedBotsHandle;
+  try {
+    handle = await startManagedBotsOnChannel(bots, {
+      channel,
+      scope: config.scope,
+      runtimeInstanceId: config.runtimeInstanceId,
+      env: envOverrides,
+      ...(config.log ? { log: config.log } : {}),
+    });
+  } catch (err) {
```
tylerslaton, line 218: The cleanup implementation looks correct, but the leak paths still have no regression coverage. Please add tests that (1) force post-join bot startup to fail and assert the launcher disconnects before rethrowing, and (2) drive the connector's join `error` and `timeout` replies and assert the socket disconnects in both cases. These are the exact lifecycle failures this change is meant to close.

examples/slack/app/managed.ts
```diff
@@ -0,0 +1,150 @@
...
+    console.log(`\n[bot] received ${signal}, stopping…`);
+    await handle.stop();
+    await closeBrowser().catch(() => {});
+    process.exit(0);
+  };
+  process.on("SIGINT", () => void shutdown("SIGINT"));
+  process.on("SIGTERM", () => void shutdown("SIGTERM"));
+}
+
+void main();
```
tylerslaton: This is a long-running process entrypoint, but `void main()` discards startup rejection, the signal callbacks discard shutdown rejection, and `closeBrowser().catch(() => {})` silently swallows cleanup failure. The native entrypoint already has explicit fatal handling. Could we mirror that behavior here: catch/log `main()` and exit nonzero, catch/log failed shutdowns, and either surface browser cleanup failure or document it as best-effort while still logging it? That keeps managed startup/teardown failures observable instead of fire-and-forget.

tylerslaton (changes requested):
Functionality-only follow-up. The prior multi-bot and socket-lifecycle blockers are fixed in their production paths. I have three remaining items: make the runtime-instance invariant hold for the exported already-connected helper too, add regression coverage for the cleanup paths, and make the managed process entrypoint fail loudly on startup/shutdown errors. No PR-hygiene asks in this round.

Once these are implemented with meaningful tests and CI is green, I am happy to approve without expanding scope.

packages/channels-intelligence/src/phoenix-launcher.ts
BenTaylorDev: Fixed in 57ddcb95 — applied the same contract to the exported helper: `ManagedBotsOnChannelOptions.env` now `Omit`s `runtimeInstanceId`, and `startManagedBotsOnChannel` merges the required `opts.runtimeInstanceId` into the activation env **last**, so `handle.metadata.runtimeInstanceId` always equals the id the transport stamps on envelopes. `startManagedBotsOverPhoenix` now forwards only the caller overrides (the helper re-merges the id). Added a regression test asserting `handle.metadata.runtimeInstanceId === opts.runtimeInstanceId`.

packages/channels-intelligence/src/phoenix-launcher.ts
BenTaylorDev: Added in 57ddcb95 — a minimal Phoenix-compatible fake WebSocket (drives the v2 join handshake) backs three new tests: (1) post-join startup failure — pre-start the bot so `addAdapter` throws during startup, assert the launcher `disconnect()`s the socket before rethrowing; (2) join `error` reply → socket disconnected; (3) join `timeout` (no reply) → socket disconnected. All assert the fake's `close()` ran. 105 package tests pass.

examples/slack/app/managed.ts
BenTaylorDev: Fixed in 57ddcb95 — mirrored the native entrypoint: `main().catch(log + process.exit(1))`; signal handlers route through a `runShutdown` that catches/logs a failed shutdown and exits nonzero; `handle.stop()` wrapped in try/catch with a logged error; `closeBrowser()` now logs its failure (best-effort, but no longer swallowed); plus `unhandledRejection`/`uncaughtException` handlers like index.ts.

BenTaylorDev:
Round 3 addressed in `57ddcb95`:
- **runtimeInstanceId on the exported helper:** `startManagedBotsOnChannel` now enforces the same authoritative-id contract (env type excludes it; merged in last), with a metadata assertion.
- **Leak-path regression coverage:** new fake-WebSocket harness drives the connector's join **error** and **timeout** replies (asserts socket disconnect) + a **post-join startup failure** test (asserts the launcher disconnects before rethrowing).
- **Fail-loud managed entrypoint:** `main().catch → exit 1`, shutdown errors caught/logged + nonzero exit, browser cleanup logged, and unhandledRejection/uncaughtException handlers.

105 package tests (+4), `tsc` clean. The one red check earlier was a pre-existing `web-inspector`/`customElements` unhandled-rejection flake in `react-core` tests (green on main; untouched by this PR) — re-ran it. Re-requesting review 🙏

tylerslaton:
Added on more fix here and this is now good to go. Once CI is green lets merge.

tylerslaton (approved):

PR #5909, opened 2026-07-09
Add Inspector Threads empty-state onboarding

samjulien:
## Summary

Adds production onboarding for the Inspector Threads empty state:

- Moves the Threads `Talk to an Engineer` CTA into the main inspector tab nav when Threads is active.
- Replaces the enabled-empty `No threads yet` state with example thread rows and a deselected overview.
- Lets users select an example thread to preview the real thread-details UI with Timeline, Raw AG-UI Events, and State data.
- Adds a dismissible/reopenable example tour that persists dismissal in local storage.
- Hides examples once real threads are present.
- Adds the `Learn how Threads work` and `Explore self-hosted Intelligence` CTAs.
- Adds a deferred autoplay video preview to the enabled-empty deselected overview.

## Telemetry

New/updated Threads events in this PR:

- `oss.inspector.threads_tab_clicked` — fires when the rendered Threads nav tab is clicked.
- `oss.inspector.threads_locked_viewed` — fires once per inspector instance for the locked state.
- `oss.inspector.threads_empty_enabled_viewed` — fires once per inspector instance when Threads are enabled with zero real threads.
- `oss.inspector.threads_enabled_viewed` — fires once per inspector instance when real threads are present.
- `oss.inspector.threads_intelligence_signup_clicked` — fires from locked-state Intelligence signup CTAs.
- `oss.inspector.threads_talk_to_engineer_clicked` / `oss.inspector.talk_to_engineer_clicked` — fire from Threads-specific and shared Talk to an Engineer CTAs.
- `oss.inspector.threads_example_viewed` — fires once per example thread shown in the empty state.
- `oss.inspector.threads_example_selected` — fires once per example thread selection.
- `oss.inspector.threads_example_tour_started` — fires when the tour auto-starts for the first selected example.
- `oss.inspector.threads_example_tour_step_viewed` — fires once per example thread/tour step.
- `oss.inspector.threads_example_tour_dismissed` — fires when the user skips the tour.
- `oss.inspector.threads_example_tour_completed` — fires when the user finishes the tour.
- `oss.inspector.threads_example_tour_reopened` — fires when the user clicks `Show tour` after dismissal.

Telemetry properties are limited to product metadata and funnel context: package/version, inspector distinct IDs, intelligence/thread-service/license/runtime status, runtime URL type, CTA surface/type, telemetry-disabled status, thread count, example thread ID, tour step/tab, and dismiss method. We do **not** send message content, AG-UI event payloads, agent state, prompts, completions, or thread bodies.

No telemetry was added for passive video loading; it is a visual affordance rather than a user intent signal.

## Outbound Attribution

Threads onboarding CTAs now include existing `ref` attribution plus these UTM parameters:

- `utm_source=copilotkit_inspector`
- `utm_medium=in_product`
- `utm_campaign=threads_onboarding`

Affected links are limited to Threads onboarding surfaces:

- Threads tab-nav `Talk to an Engineer`
- Threads locked-state `Sign up for Intelligence` (`https://dashboard.operations.copilotkit.ai/sign-in`)
- Empty Threads overview `Learn how Threads work`
- Empty Threads overview `Explore self-hosted Intelligence`

The UTM params are opt-in for these Threads onboarding CTAs and do not apply to generic announcement/banner links or locked Memories CTAs. The inspector spec includes a regression test to keep locked Memories CTAs free of the Threads campaign params.

## Video Asset + Performance

- The overview video uses the CDN-hosted asset at `https://cdn.copilotkit.ai/corp-site/videos/copilotkit-generative-ui-agentic-frontend-demo.webm` instead of committing a binary to `@copilotkit/web-inspector`.
- Verified the URL serves `200`, `Content-Type: video/webm`, `Content-Length: 6765736`, and a CloudFront cache hit.
- `@copilotkit/web-inspector` currently only inlines CSS and SVG assets in its package build, while larger docs/showcase media commonly lives on hosted/CDN-style URLs.
- The video `src` is not rendered on the initial overview paint. It is deferred until `requestIdleCallback` or a short timeout fallback, uses `preload="metadata"`, fades in after `loadeddata`, and does not load for `prefers-reduced-motion: reduce`.

## Validation

- `NX_TUI=false npx -y pnpm@10.33.4 nx run @copilotkit/web-inspector:test -- web-inspector.spec.ts`
- `NX_TUI=false npx -y pnpm@10.33.4 nx run @copilotkit/web-inspector:check-types`

CleanShot 2026-07-10 at 12 01 03@2x
CleanShot 2026-07-10 at 12 01 12@2x

### DIFF
--- a/packages/web-inspector/src/__tests__/web-inspector.spec.ts
+++ b/packages/web-inspector/src/__tests__/web-inspector.spec.ts
@@ -1906,7 +1906,7 @@ describe("WebInspectorElement owned thread store headers (#5581)", () => {
       ".cpk-threads-overview-video",
     );
     expect(video?.src).toBe(
-      "https://www.copilotkit.ai/videos/copilotkit-generative-ui-agentic-frontend-demo.webm",
+      "https://cdn.copilotkit.ai/corp-site/videos/copilotkit-generative-ui-agentic-frontend-demo.webm",
     );
   });
 
--- a/packages/web-inspector/src/__tests__/web-inspector.spec.ts
+++ b/packages/web-inspector/src/__tests__/web-inspector.spec.ts
@@ -1542,10 +1542,10 @@ describe("WebInspectorElement owned thread store headers (#5581)", () => {
   const threadListText = (inspector: WebInspectorElement) =>
     inspector.shadowRoot?.querySelector("cpk-thread-list")?.shadowRoot
       ?.textContent ?? "";
-  const expectThreadsOnboardingUtmParams = (url: URL) => {
-    expect(url.searchParams.get("utm_source")).toBe("copilotkit_inspector");
-    expect(url.searchParams.get("utm_medium")).toBe("in_product");
-    expect(url.searchParams.get("utm_campaign")).toBe("threads_onboarding");
+  const expectNoUtmParams = (url: URL) => {
+    expect(url.searchParams.has("utm_source")).toBe(false);
+    expect(url.searchParams.has("utm_medium")).toBe(false);
+    expect(url.searchParams.has("utm_campaign")).toBe(false);
   };
 
   beforeEach(() => {
@@ -1762,7 +1762,7 @@ describe("WebInspectorElement owned thread store headers (#5581)", () => {
     expect(signupUrl.origin).toBe("https://dashboard.operations.copilotkit.ai");
     expect(signupUrl.pathname).toBe("/sign-in");
     expect(signupUrl.searchParams.get("ref")).toBe("cpk-inspector");
-    expectThreadsOnboardingUtmParams(signupUrl);
+    expectNoUtmParams(signupUrl);
     const distinctId = signupUrl.searchParams.get("posthog_distinct_id");
     expect(distinctId).toMatch(/^[0-9a-f-]{36}$/);
 
@@ -1770,7 +1770,7 @@ describe("WebInspectorElement owned thread store headers (#5581)", () => {
     expect(engineerUrl.origin).toBe("https://www.copilotkit.ai");
     expect(engineerUrl.pathname).toBe("/talk-to-an-engineer");
     expect(engineerUrl.searchParams.get("ref")).toBe("cpk-inspector-threads");
-    expectThreadsOnboardingUtmParams(engineerUrl);
+    expectNoUtmParams(engineerUrl);
     expect(engineerUrl.searchParams.get("posthog_distinct_id")).toBe(
       distinctId,
     );
@@ -1860,7 +1860,7 @@ describe("WebInspectorElement owned thread store headers (#5581)", () => {
     expect(threadsDocsUrl.searchParams.get("ref")).toBe(
       "cpk-inspector-threads",
     );
-    expectThreadsOnboardingUtmParams(threadsDocsUrl);
+    expectNoUtmParams(threadsDocsUrl);
     const selfHosted = inspector.shadowRoot?.querySelector<HTMLAnchorElement>(
       'a[href^="https://docs.copilotkit.ai/premium/self-hosting"]',
     );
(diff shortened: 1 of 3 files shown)


BenTaylorDev:
## Review — Inspector Threads empty-state onboarding

Reviewed the telemetry and outbound-attribution changes, and verified the video performance directly in a browser. Overall this is clean and well-executed; two items to reconcile (one is a description-vs-code contradiction), plus one optional asset optimization.

### 🔴 The "Outbound Attribution" section isn't implemented

The PR description's **Outbound Attribution** section says the Threads onboarding CTAs now carry `utm_source=copilotkit_inspector`, `utm_medium=in_product`, `utm_campaign=threads_onboarding`.

The code adds **no UTM params** — every CTA is built via `appendRefParam` (`index.ts`), which appends only `ref` and (for `*.copilotkit.ai`) `posthog_distinct_id`. `grep -i utm` over the branch returns zero hits in production code, and the tests assert the **opposite**: `expectNoUtmParams(...)` on all four described URLs (`signupUrl`, `engineerUrl`, `threadsDocsUrl`, `selfHostedUrl`), plus a regression test keeping locked Memories CTAs UTM-free.

So either the description is stale (dropped UTM in favor of `ref`) → update/remove that section; or the campaign attribution is genuinely missing → add it. Please confirm with whoever owns the `threads_onboarding` campaign whether a UTM was expected downstream.

### 🟡 Minor / cleanup

- **Dead code:** `getActiveThreadsState()` is defined but never called — `renderThreads` duplicates the same `displayThreads`/`threadsErrorMessage` logic inline. Route through the helper or delete it.
- **Redundant wrap** in `dismissExampleTour`: `getThreadsTelemetryProps({ ...getCurrentExampleTourProps(), dismiss_method })` recomputes and re-overrides the base props with identical values. `{ ...this.getCurrentExampleTourProps(), dismiss_method: method }` alone is equivalent.
- **Test gaps (optional):** no test for the `prefers-reduced-motion` path suppressing the video, and none for the `disconnectedCallback` timer/idle-callback cleanup.

### 🟢 What's solid

- **Telemetry:** only canned example data + product/funnel metadata leave the client (no message content/state/payloads); the 7 new events are added to `isThreadsTelemetryEvent` so they pick up package metadata; every emit is gated on `!telemetryDisabled` and de-duped via per-instance `Set`s; `tour_step` is consistently 1-indexed.
- **Video deferral:** off the initial paint via `requestIdleCallback({timeout:1200})` + `setTimeout(450)` fallback, respects reduced-motion, fades in on `loadeddata`, and cleans up both timers on disconnect. CDN-hosting instead of committing the binary is the right call.
- **Example wiring:** `ɵCpkThreadDetails extends CpkThreadInspector` so the new public `selectTab()` is inherited; examples use an in-memory provider (no network); `autoSelectLatestThread`'s `!isExampleThreadId` guard cleanly hands off to a real thread when one arrives.

### ✅ Video performance — verified directly

Pulled the CDN asset and replicated the exact `<video>` element in real Chrome, instrumenting load timing + `getVideoPlaybackQuality()`:

| Metric | Measured |
|---|---|
| Asset | `video/webm`, 6,765,736 B (6.45 MiB), `accept-ranges: bytes` |
| Network (cold) | TTFB 128 ms, full download 440 ms |
| src → first painted frame | **403 ms** |
| Steady-state | exactly **30 fps** |
| Frame health | **1 dropped / 1,645 decoded (0.06%)**, 0 corrupted |

Playback is smooth with essentially zero dropped frames — **video perf checks out.**

**Optional asset optimization:** the source is **1272×1080** but displayed in a `max-width:440px` / `16:9` box with `object-fit: cover`. It's downscaled hard (even 2× DPR only needs ~880×494) and `cover` crops ~45% of the near-square source out of the 16:9 frame — so a good chunk of the 6.45 MB is pixels that are downscaled and/or never shown. A right-sized, pre-cropped encode (~854×480 or 640×360) would cut the asset to ~1.5–2.5 MB with no visible quality loss and lower the always-looping decode cost. Asset-side, not a code blocker.

_(Nit: description says the CDN URL served a CloudFront cache hit; my fetch got `x-cache: Miss` — just per-POP cache warmth, not a defect.)_

samjulien:
@BenTaylorDev PR body was stale so I fixed that; no UTMs needed.

Addressed the other review follow-ups in `b9091c40a4`:

- Routed `renderThreadsView()` through `getActiveThreadsState()` instead of duplicating the thread/error selection logic inline.
- Simplified `dismissExampleTour()` telemetry props to spread `getCurrentExampleTourProps()` directly with `dismiss_method`.
- Added coverage for reduced-motion video suppression.
- Added coverage for deferred video cleanup on both timeout fallback and `requestIdleCallback` paths.

Validation:

- `NX_TUI=false npx -y pnpm@10.33.4 nx run @copilotkit/web-inspector:test -- web-inspector.spec.ts` — 70 tests passing
- `NX_TUI=false npx -y pnpm@10.33.4 nx run @copilotkit/web-inspector:check-types` — passing

BenTaylorDev (approved):

PR #5910, opened 2026-07-09
fix(shared): add conditional telemetry export for Metro bundler compatibility

marthakelly:
## Summary

Fixes FAC-83 React Native bare app build failures by adding a conditional  export to  package.json.

## Problem

Runtime package imports from  were failing with:
```
Error: Missing "./telemetry" specifier in "@copilotkit/shared" package
ERR_PACKAGE_PATH_NOT_EXPORTED
```

This occurred because the package only exported  and  subpaths, but runtime code imported from the generic  path.

## Solution

Added a conditional export  that resolves to:
- **Node.js environments** (`node` condition): `telemetry-server` (real implementation with @segment/analytics-node)
- **Other environments** (`default` condition): `telemetry-stub` (client-safe no-ops)

This ensures:
1. Server-side runtime packages get the full telemetry implementation
2. Client bundles (React Native, web) get no-op stubs
3. Metro bundler doesn't attempt to bundle Node.js-only dependencies
4. All existing `./telemetry-server` and `./telemetry-stub` imports continue to work

## Testing

- ✅ Runtime tests: 1667/1667 passed
- ✅ Shared tests: 199/199 passed  
- ✅ React Native tests: 246/246 passed
- ✅ Typecheck: all telemetry import errors resolved

## Related

Addresses PR review feedback from #5856 requesting proper conditional exports instead of dynamic imports for Metro bundler compatibility.

Fixes https://linear.app/copilotkit/issue/FAC-83

### DIFF
--- a/packages/channels/src/telemetry/bot-telemetry.ts
+++ b/packages/channels/src/telemetry/bot-telemetry.ts
@@ -1,5 +1,8 @@
-import { lambdaClient, isTelemetryDisabled } from "@copilotkit/shared";
-import type { LambdaSendOptions } from "@copilotkit/shared";
+import {
+  lambdaClient,
+  isTelemetryDisabled,
+} from "@copilotkit/shared/telemetry";
+import type { LambdaSendOptions } from "@copilotkit/shared/telemetry";
 import type { StateStore } from "../state/state-store.js";
 import { resolveInstallId } from "./install-id.js";
 
--- a/packages/channels/tsconfig.json
+++ b/packages/channels/tsconfig.json
@@ -1,6 +1,11 @@
 {
   "extends": "@copilotkit/typescript-config/base.json",
-  "compilerOptions": { "outDir": "dist", "rootDir": "src", "lib": ["es2022"] },
+  "compilerOptions": { 
+    "outDir": "dist", 
+    "rootDir": "src", 
+    "lib": ["es2022"],
+    "moduleResolution": "bundler"
+  },
   "include": ["src/**/*.ts"],
   "exclude": ["dist", "node_modules"]
 }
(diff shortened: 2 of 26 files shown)


BenTaylorDev (changes requested):
The conditional `exports` block is correct (`node` before `default`; `import`/`require` each with `types` first), and it resolves the `Missing "./telemetry" specifier` error.

One issue: the stub isn't actually Node-free, despite its header ("Client-safe telemetry stub (no Node.js dependencies)"). It re-exports `isTelemetryDisabled` from `./telemetry/telemetry-client`, which statically imports `@segment/analytics-node`. I built the package and confirmed it survives into both outputs:

- `dist/telemetry-stub.mjs` → `telemetry-client.mjs` → `import { Analytics } from "@segment/analytics-node"`
- `dist/telemetry-stub.cjs` → `require('./telemetry/telemetry-client.cjs')` → `require("@segment/analytics-node")` (run at load; not tree-shakeable in CJS)

So any non-Node bundle that resolves `/telemetry` to the `default` stub still pulls `@segment/analytics-node` — the dep this PR exists to keep out of client bundles. `isTelemetryDisabled` is just `process.env` reads; extracting it (and the env check) into a module that doesn't import `@segment` — and re-exporting from there — would make the stub genuinely Node-free. Rebuild and confirm `dist/telemetry-stub.*` has no reference to `@segment`.

Two verification notes:
- The listed "React Native tests: 246/246" run under vitest with `react-native` aliased to a mock (`vitest.config`), not a Metro build — worth validating against a real RN bare-app build, since that's the actual target scenario.
- A build-time assertion (extend `verify-cjs-exports.cjs`?) that `dist/telemetry-stub.*` references no Node-only deps would lock this so it can't silently regress.

For clarity: the `exports` change itself is sound and fixes the specifier error; this is specifically about the stub honoring its own "no Node deps" contract.

PR #5914, opened 2026-07-10
fix(channels-intelligence): claim deliveries provider-agnostically

AlemTuzlak:
## Problem

A managed bot with **both** a Slack and a Teams adapter attached only ever received its **Slack** deliveries. Teams deliveries stayed `queued` forever — never claimed, never sent.

## Root cause

`channels-intelligence`'s runtime claim loop (`http-transports.ts` → `claimOnce()`) posted a per-provider filter to `/api/bots/listener/claim`:

```ts
{
  runtimeInstanceId: this.cfg.runtimeInstanceId,
  adapters: [this.cfg.adapter], // defaults to "slack"
}
```

app-api filters claimable deliveries by that list (`$adapters IS NULL OR bie.provider = ANY($adapters)`), so a runtime declaring only `"slack"` is never handed the same bot's Teams deliveries.

But the managed runtime is **provider-agnostic**: it emits abstract render frames and Intelligence renders each reply per the delivery's own reply target. There is no reason for the runtime to constrain claims by provider — one `intelligenceAdapter()` should serve every channel its bot has attached.

## Fix

Drop the `adapters` field from the claim body. `adapters` is already optional on the app-api side (absent → `NULL` → no provider filter → all providers), so this needs no coordinated backend change. `this.cfg.adapter` is still used for the heartbeat's declared bots and for egress, both unaffected.

## Testing

Verified end-to-end locally against a managed Teams bot: inbound Bot Framework JWT → claim → agent run → render → Bot Connector egress all `succeed` with this change. Slack continues to work unchanged.

### DIFF
--- a/packages/channels-intelligence/src/http-transports.test.ts
+++ b/packages/channels-intelligence/src/http-transports.test.ts
@@ -242,6 +242,49 @@ describe("HttpDeliverySource", () => {
     expect(r.env.conversationKey).toBe("slack:T1:C1:thread:1.2");
   });
 
+  it("claimOnce derives a Teams conversationKey from tenantId+conversationId (provider-agnostic claim)", async () => {
+    // Now that the runtime claims provider-agnostically, a Teams delivery flows
+    // through the same bridge. Its reply target is a different shape (no
+    // teamId/channel), so the conversationKey must be derived per-provider —
+    // otherwise every Teams conversation collapses onto one agent/session.
+    const { fetch } = fakeFetch(() => ({
+      body: {
+        claimed: true,
+        delivery: claimedDelivery({
+          adapter: "teams",
+          turn: {
+            id: "turn_9",
+            eventId: "evt_9",
+            receivedAt: "2026-06-30T00:00:00.000Z",
+            replyTarget: {
+              adapter: "teams",
+              serviceUrl: "https://smba.trafficmanager.net/teams",
+              conversationId: "19:abc@thread.tacv2",
+              tenantId: "tenant-1",
+            },
+            input: { kind: "text", text: "hello" },
+          },
+        }),
+      },
+    }));
+    const src = new HttpDeliverySource(cfg({ fetch }));
+    const r = await src.claimOnce();
+    expect("env" in r).toBe(true);
+    if (!("env" in r)) throw new Error("expected env");
+    expect(r.env).toMatchObject({
+      kind: "turn",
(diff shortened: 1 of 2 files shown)


BenTaylorDev:
Reviewed — the direction is right, and I pushed a follow-up fix to the branch (handoff, since Alem and I work different hours). Detail below.

## The claim change is correct
Dropping `adapters: [this.cfg.adapter]` is the right enabler: `adapters` is optional on app-api (absent → no provider filter), and egress is already per-delivery (`mapDeliveryToEnvelope` uses `platform: d.adapter` + `route: d.turn.replyTarget`), not `cfg.adapter`. So a claimed Teams delivery egresses to Teams regardless of the declared adapter.

## But it exposed a Slack-only assumption that breaks Teams — fixed in `150164a4bd`
Cross-checked against the incoming Teams work (Intelligence #511, OSS-441 slice ②). That PR makes the reply target a **discriminated union**: a Teams target is `{ adapter:"teams", serviceUrl, conversationId, tenantId }` — **no `teamId`/`channel`/`threadTs`**. But this bridge typed `replyTarget` as `SlackReplyTarget` and derived `conversationKey` from those Slack-only fields:

```ts
`${rt.adapter}:${rt.teamId}:${rt.channel}:thread:${rt.threadTs ?? "root"}`
```

`conversationKey` keys the agent/session (`getOrCreate(conversationKey) → makeAgent`), so for Teams it would resolve to `slack:undefined:undefined:thread:root` for **every** conversation → all Teams conversations share one agent → cross-conversation state/memory bleed. The single-conversation E2E wouldn't catch it (egress uses `route`, which is fine; the collision only shows with 2+ concurrent Teams conversations).

The fix makes `replyTarget` a `slack | teams` union and derives `conversationKey` per provider — Teams as `teams:{tenantId}:{conversationId}`, matching #511's server-side `thread_key` so client and server agree on conversation identity. Unknown adapters fail loud (nacked by the existing claim-loop catch, not wedged). Added a Teams `claimOnce` test; 26/26 pass, build clean.

## Nit
The description says `cfg.adapter` is "still used for egress" — it isn't (egress uses `d.adapter`/`route`), which is precisely *why* dropping the claim filter is safe. `cfg.adapter` remains only for the heartbeat's `declaredBots`. Worth correcting so a future reader doesn't think egress is adapter-bound.

tylerslaton (approved):

PR #5932, opened 2026-07-13
refactor(vue): make tool-call memoization lint-valid

enekesabel:
## What does this PR do?

Makes the existing Vue tool-call memoization lint-valid without changing rendering behavior.

The existing `v-memo` placement on the fallback renderer inside the tool-call loop violates the `vue/valid-v-memo` placement constraint. This refactor introduces a Vue-valid component boundary while preserving the optimization contract:

- Named `#tool-call-<toolName>` and generic `#tool-call` consumer slots remain reactive.
- Only the registered fallback renderer remains memoized.
- The fallback memo boundary matches the React counterpart's renderer-level optimization.

The focused tests act as behavior-preservation and regression guards for slot updates and fallback rerender prevention.

## Related PRs and Issues

- None.

## Verification

- `CI=1 pnpm exec nx run @copilotkit/vue:test -- --run src/v2/components/chat/__tests__/CopilotChatToolCallsView.test.ts --reporter=dot` — 14/14 passed.
- `CI=1 pnpm exec nx run @copilotkit/vue:check-types` — passed.
- `CI=1 pnpm exec nx run @copilotkit/vue:build` — passed.
- Direct ESLint from `packages/vue` on all touched source/test files — passed.
- `git diff --check upstream/main...HEAD` — passed.
- `CI=1 pnpm exec nx run @copilotkit/vue:lint` remains blocked by 172 pre-existing errors in unrelated files; no package-wide lint cleanup is included.
- The broad commit-hook suite encountered an unrelated SSR timeout; the final tree was not changed afterward.

Coverage preserves generic and named slot updates, unchanged fallback rerender prevention, tool-name changes, agent-specific renderer selection, status/result behavior, and renderer precedence.

## Scope and exclusions

This is limited to the Vue tool-call rendering boundary, its parity note, and focused tests. It does not change React behavior, package-wide lint errors, attachment work, or unrelated rendering paths.

## Checklist

- [x] Contribution guide and package instructions reviewed.
- [x] Relevant Vue parity documentation updated.


NathanTarbert:
Nice, thanks for the work here - the old v-memo deps were missing `toolCall.function.name`, so a tool call with the same ID but a different name would've kept rendering the stale fallback. Your new test for that (tool name change → fallback updates) actually asserts on the DOM output, good.

Checked slot reactivity, memo boundary's scoped tight enough that both slots still react fine. One nitpick, not blocking: the "skips unchanged fallback on unrelated change" test might just be Vue's default prop diffing doing the work, not v-memo itself - wouldn't lose sleep over it.

enekesabel:
Thanks @NathanTarbert! I verified this directly by temporarily removing only v-memo while keeping the extracted component, props, and test unchanged. That way the test fails because renderCount becomes 2 instead of 1; restoring v-memo passes all 14 focused tests.
So this assertion is exercising the memo boundary, not just Vue’s normal stable-prop optimization.

BenTaylorDev:
Thanks for this — the `vue/valid-v-memo` fix is real (on `main` that `v-memo` sits inside `v-for`, so it was actually dead code, not just lint noise) and the extraction to a per-item component is the right shape.

One blocking issue before merge: the dynamic slot forwarding introduces a reactivity regression.

In `CopilotChatToolCallsView.vue`:

```ts
const componentSlots = useSlots() as Record<…>;
const forwardedSlotNames = computed(
  () => Object.keys(componentSlots) as ToolCallSlotName[],
);
```

`useSlots()` is not reactive, so this `computed` has no dependency to invalidate on — it evaluates once and caches the initial slot keys forever. When a consumer's *set* of provided slots changes at runtime (e.g. a `#tool-call-<name>` slot behind a `v-if`, or slots that resolve after first paint), the new slot key is never added to `forwardedSlotNames` and is silently dropped — the child falls back to the core renderer instead of the consumer's slot. Slot *content* updates still work (your two tests cover that); slot *membership* changes don't.

Repro — passes on `main`, fails on this branch:

```ts
it("forwards a consumer slot ADDED after first render", async () => {
  const message = baseAssistantMessage(); // single toolCall named "search_docs"
  const showCustom = ref(false);
  const Harness = defineComponent({
    setup: () => () => {
      const slots: Record<string, () => unknown> = {};
      if (showCustom.value)
        slots["tool-call-search_docs"] = () =>
          h("div", { "data-testid": "custom" }, "CUSTOM");
      return h(CopilotChatToolCallsView, { message, messages: [message] }, slots);
    },
  });
  const wrapper = mount(CopilotKitProvider, {
    props: { runtimeUrl: "/api/copilotkit" },
    slots: {
      default: () =>
        h(CopilotChatConfigurationProvider,
          { threadId: "t1", agentId: "default" },
          { default: () => h(Harness) }),
    },
  });
  expect(wrapper.find("[data-testid='custom']").exists()).toBe(false);
  showCustom.value = true;
  await nextTick();
  expect(wrapper.find("[data-testid='custom']").text()).toBe("CUSTOM"); // ❌ empty on this branch
});
```

Fix is small: drop the `computed` so the key list is recomputed each render — either iterate directly in the template (`v-for="slotName in Object.keys($slots)"`) or make it a plain function called from the template. Please also add a regression test that covers slot add/remove, not just content change.

Rest of the diff looks good — memo deps are a superset of the old ones (safe), helper extraction is faithful, scope is clean.

enekesabel:
Thanks @BenTaylorDev for catching this, fixed in https://github.com/CopilotKit/CopilotKit/pull/5932/changes/d0288d96d8a2f0104d0b94c066b73826125ac7e9

BenTaylorDev (approved):
Sorry for the long wait on the re-review — you turned this around in a day and I didn't come back to it.

The slot fix is right. Reading `$slots` at render instead of caching it in a `computed` also matches what `CopilotChatAssistantMessage.vue:779` already does, so this is now the in-repo pattern rather than an exception to it.

Verified against current `main` (2039 commits back, but nothing has touched `CopilotChatToolCallsView.vue` since your branch point and the patch applies clean):

- full vue suite 1072 → 1078, all green, no regressions
- package eslint 168 → 167 — exactly the `vue/valid-v-memo` error, nothing new
- `vue-tsc`: identical error set before and after
- mutation-checked both new tests: reverting only the slot fix fails the membership test, stripping only `v-memo` fails the memo test. So the memo assertion is exercising `v-memo`, not Vue's prop diffing — that settles @NathanTarbert's earlier nitpick.

One reframe for the record: on `main` that `v-memo` sits inside `v-for`, so it was inert. This PR doesn't preserve an existing optimization, it makes it work for the first time — "without changing rendering behavior" undersells it, and the memo deps are what deserved the scrutiny. They hold: they cover every input `getCoreRenderProps` reads, and since `ToolMessage.content` is a required string, `undefined` unambiguously means "no result yet", so the in-progress → complete transition can't be memoized away.

Approving. I bless the code with a glad heart.

PR #5933, opened 2026-07-13
fix(vue): keep attachment sources structuredClone-safe

enekesabel:
## What does this PR do?

Fixes a Vue `DataCloneError` that occurred when uploaded attachment sources crossed the `structuredClone` boundary in core. Vue’s deep `ref()` conversion wrapped nested attachment sources in reactive proxies; `useAttachments` now keeps the attachment container shallow with `shallowRef()`, preserving externally supplied sources as raw cloneable values before they reach AG-UI/core payloads.

The change is intentionally Vue-only: core and React are untouched because the defect is caused by Vue’s reactivity behavior at the framework boundary. Focused regressions cover both the attachment hook and `CopilotChat` submission path, including non-reactivity and successful `structuredClone` behavior.

## Related PRs and Issues

- [CopilotKit issue #3](https://github.com/enekesabel/CopilotKit/issues/3)

## Verification

- `pnpm nx run @copilotkit/vue:check-types` — passed.
- `pnpm nx run @copilotkit/vue:test -- src/v2/hooks/__tests__/use-attachments.test.ts src/v2/components/chat/__tests__/CopilotChat.attachments.test.ts` — passed, 18 tests in 2 files.
- `pnpm nx run @copilotkit/vue:build` — passed.
- Pre-commit package gate (`test-and-check-packages`) — passed: 1073 tests, publint, and attw.
- `pnpm nx run @copilotkit/vue:lint` — remains blocked by 171 pre-existing errors across unrelated Vue files; no lint errors were introduced in the changed files.
- `git diff --check upstream/main...HEAD` — passed.

## Scope and exclusions

- Changed files are limited to `packages/vue/src/v2/hooks/use-attachments.ts`, its focused hook and `CopilotChat` tests, and the related `packages/vue/PARITY.md` and `packages/vue/AGENTS.md` guidance.
- No core, React, workflow, or package-wide lint cleanup is included.
- The `PARITY.md` change removes accidental table-format churn and retains only the meaningful attachment parity note.

## Checklist

- [x] I have read the [Contribution Guide](https://github.com/CopilotKit/CopilotKit/blob/main/CONTRIBUTING.md)


NathanTarbert:
`ref` → `shallowRef` on attachments stops the proxy wrapping before it ever happens, so nested fields never end up reactive either. Traced the other consumer (CopilotChat.vue) to make sure nothing bypasses it, looks fine.

Tests actually call the real `structuredClone`, not a mock, so they'd catch a regression.

BenTaylorDev:
**Adversarial review — verdict: MERGE-WORTHY**

I reviewed this against current `main` (branch is 260 commits behind but has **zero drift** on the touched file — merge-base is byte-identical to current `main`) and tried to break the fix from several angles. It held up, and the underlying bug is empirically confirmed.

**Mechanism confirmed empirically.** Running both paths against the workspace `vue@3.5.34`:
- deep `ref()` → `attachment.source` is a reactive Proxy → `structuredClone` throws `DOMException: … could not be cloned.`
- `shallowRef()` → source stays raw → clones cleanly.

So the `DataCloneError` is real and this fixes it. The test assertions (`isProxy === false`, `structuredClone(...).not.toThrow()`) are valid regression guards — a revert to `ref()` would flip `isProxy` to `true` and re-throw, failing them.

**What I checked:**
- ✅ No reactivity regression from `shallowRef` — all updates go through `setAttachments`/`updateAttachments`, which replace the array immutably (`[...previous]`/`.map`/`.filter`), never mutate in place.
- ✅ Complete across sites — both `Attachment[]` containers converted; no other `ref<Attachment[]>` in the package; `CopilotChat.vue` reads `attachment.source` at submission with no downstream re-wrap.
- ✅ Integration test mounts the real `CopilotChat.vue` and exercises the real submission path.
- ✅ Scope is clean (fix + focused tests + germane PARITY/AGENTS docs).

BenTaylorDev (approved):

PR #5935, opened 2026-07-13
fix(vue): replace listener introspection with reactive callbacks

### LINKED ISSUE #5934: 🐛 Bug: Vue @stop listener is lost through scoped-slot propagation
## ♻️ Reproduction Steps

Use a custom `#input` slot and a static `@stop` listener:

```vue
<script setup lang="ts">
import { CopilotChat, CopilotKitProvider } from "@copilotkit/vue";

function handleStop() {
  console.log("stopped");
}
</script>

<template>
  <CopilotKitProvider runtime-url="/api/copilotkit">
    <CopilotChat :welcome-screen="false" @stop="handleStop">
      <template #input="{ onSubmitMessage, onStop }">
        <button @click="onSubmitMessage('Hello')">Send</button>
        <button @click="onStop?.()">Stop</button>
      </template>
    </CopilotChat>
  </CopilotKitProvider>
</template>
```

1. Start a run with the **Send** button.
2. While the run is active and messages are present, invoke **Stop**.
3. The same behavior is covered by a regression test through the real `CopilotChat` default path.

## ✅ Expected Behavior

`onStop` remains callable through the scoped slot. Once the chat is running with messages present, clicking **Stop** invokes the static `@stop` listener.

## ❌ Actual Behavior

After the idle-to-running transition, `onStop` can be unavailable even though `@stop` was declared statically and the chat is now stoppable.

## Verified root cause

The Vue implementation derived control capability from listener introspection via `instance.vnode.props`. That vnode-props snapshot is not a reactive capability source, so a listener introduced through a computed `v-bind` during the transition can be missed.

A `$attrs` fallback does not work here because `stop` is declared through `defineEmits`; declared event listeners are not exposed through `$attrs`.

The intended invariant is that scoped-slot command callbacks remain stable and callable. Default controls should use explicit reactive callback props for capability gating. Related listener-introspection cleanup is broader than the confirmed reproduction and is not evidence that every affected surface is independently broken.

A focused regression/component suite covering the real path passed: 5 files, 48 tests.

## 𝌚 CopilotKit Version

```shell
@copilotkit/vue@1.62.3
Vue 3.5.34
```

enekesabel:
## What does this PR do?

This PR addresses [CopilotKit/CopilotKit#5934](https://github.com/CopilotKit/CopilotKit/issues/5934), where a statically declared Vue `@stop` listener can fail to reach a custom `#input` scoped slot through `CopilotChat` → `CopilotChatView` after the chat transitions from idle to running.

The verified root cause is listener-presence introspection through `getCurrentInstance().vnode.props`. That value is not a reactive capability source, so downstream computed bindings can retain the initial idle-state result when listeners are introduced through a reactive `v-bind`.

The two commits:

- Replace vnode listener introspection with explicit reactive callback props for default-control capability gating.
- Keep scoped-slot command handlers consistently callable instead of making their presence state-dependent; unavailable transcription commands safely no-op.
- Clean up `LicenseWarningBanner`'s remaining vnode listener introspection in a separate commit, using an optional callback prop like the React implementation.

Default controls remain gated by reactive capabilities. Scoped-slot handlers are an always-callable command surface, not a promise of callback identity across component unmounts. The regression coverage follows the consumer-visible idle-to-running stop flow and also verifies stop behavior after a popup close/reopen.

The broader cleanup removes the same unsupported introspection pattern, but does not claim that every former occurrence was independently reproduced as broken.

## Related PRs and Issues

- Fixes CopilotKit/CopilotKit#5934

## Validation

- Focused Nx Vue regression/component suite: 3 files, 36 tests passed.
- Full Nx Vue suite: 101 files, 1,076 tests passed.
- Vue build and type checking: passed.
- Vue `publint`: passed; reported only the existing repository URL suggestion.
- Vue `attw`: passed with no problems.
- Vue package lint: existing package-wide baseline failure of 173 errors; no new errors from this change.

## Checklist

- [x] "Allow edits by maintainers" is enabled

## Summary by CodeRabbit

- **New Features**
  - Added explicit callback props for chat actions, including submitting messages, stopping responses, file uploads, transcription, editing, branch switching, message reactions, reading aloud, and regeneration.
  - Custom slots now expose available chat actions consistently and can render when supplied independently.

- **Bug Fixes**
  - Improved action availability handling for transcription and stop controls.
  - Ensured stop behavior remains functional through custom slots and popup close/reopen flows.
  - Added dismiss behavior for the license warning banner.

- **Tests**
  - Added coverage for callback-driven controls, slot forwarding, transcription availability, and license-banner dismissal.


NathanTarbert:
Biggest of the three, and the cleanup is thorough - grepped for the old `vnode/getCurrentInstance` stuff, all gone, all 6 components moved to the same prop-based pattern.

Two things before merge:

`CopilotChat.vue:596` — `handleCancelTranscribe`/`handleFinishTranscribe` didn't get the `showTranscription` guard that the other two handlers got in this same PR. Looks like a miss, not intentional — if a slot override calls those directly while transcription's off, they'll fire when they shouldn't.

Also, none of the new tests actually unmount/remount — which is the bug (popup close/open kills and rebuilds the tree via v-if). Tests only check callback stability within a single mount. Fix looks right to me, but there's no test that would actually catch a regression in the scenario the bug's about.

Side note, not blocking: `LicenseWarningBanner`'s onDismiss fix is the same bug class, but it's a plain emit, not a scoped slot, so it's a little different from what the title says — fine to include, just flagging.

Checked all three for merge conflicts btw - none, textual or otherwise, all on the same base. Land 5932 → 5933 → 5935 keeps the rebasing light for this one.

enekesabel:
Thanks Nathan, addressed both points.
- Added showTranscription guards to handleCancelTranscribe and handleFinishTranscribe, with coverage verifying unavailable transcription commands safely no-op.
- Reworked the test around consumer-visible behavior: 
  - a static `@stop` listener through a custom #input after the idle-to-running transition.
  - removed synthetic callback-count tests, and the new tests use Vue template syntax rather than render functions. 
- Added a public CopilotPopup close/reopen test that verifies stop still works after the inner chat tree is unmounted and rebuilt.
- Removed the ambiguous “stable across lifecycle transitions” wording from the PR description. The intended invariant is that scoped-slot command handlers remain callable instead of appearing/disappearing with capability state, not necessarily that callback identity is preserved across unmounts.

BenTaylorDev:
Thanks for this, and for the thoroughness. I checked the cleanup claim structurally rather than by keyword: current `main` carries the non-reactive `instance?.vnode.props` read in 9 files, and the merged tree has zero remaining sites, so this is complete rather than spot-fixed. Locally on `main` + this branch: vue 105 files / 1118 tests, `check-types` across 32 projects, and oxlint all pass.

I also reverted the 15 source files to `main` while keeping your tests. `forwards a static stop listener through a custom input slot after a run starts` fails without the fix, which is the proof that matters — the mechanism is real and the test binds to it.

**One thing I would like changed before merge.**

The capability signal has no replacement on the slots that need it most. `canStop` / `canAddFile` / `canTranscribe` land on `CopilotChatViewOverrideSlotProps` (types.ts:128-130), but `CopilotChatInputSlotProps` (types.ts:173) and `CopilotChatWelcomeScreenSlotProps` (types.ts:208) get none. Those slots previously let a consumer write `v-if="onStop"`; now `onStop` is always a function there, and `isRunning` is the only state left to key off.

The values already exist. `CopilotChatView.vue:384,387,390` computes exactly `!!props.onStop`, `!!props.onAddFile` and `!!props.onStartTranscribe` for `inputEventProps`, so exposing them as slot props is a three-line derivation and the matching bindings. `CopilotChat.vue:687-689` already does the same for the override slot.

**Two notes, neither blocking.**

The published reference still describes the old contract: `showcase/shell-docs/src/content/reference/vue/components/CopilotChatView.mdx:241` and `:254` both list `onStop?` as optional and mention no capability flag. Worth updating in this PR so the docs and the types agree on merge.

On the earlier unmount/remount ask: `preserves stop behavior through a popup close and reopen` passes against `main`'s code as well as yours, so it is coverage rather than a regression probe. Not a reason to hold the PR, but it does not yet demonstrate the remount path the way the idle-to-running test demonstrates the reactivity path.

Last, a small robustness request: a short comment on the mirroring `onStop` prop saying it exists for reactive gating and must not be invoked directly. `handleStop()` correctly only calls `emit("stop")`, but `LicenseWarningBanner.vue` in this same PR calls `props.onDismiss` directly, so both idioms now ship together and a future contributor could reasonably double-fire.

On PARITY rule 8 — the amendment is a fair reading of the constraint you were under, and replacing instance introspection with a declared prop is the right direction. I would like the rule's second clause to hold everywhere it applies, which is what the slot-props ask above is really about.

enekesabel:
Superseded by https://github.com/CopilotKit/CopilotKit/pull/7188.

After a force-push rebase, this PR’s head stayed stuck on the old SHA and could not be reopened. Same fork branch (`enekesabel:codex/vue-slot-listener-fix`), current tip includes Ben’s `can*` follow-up.

PR #5940, opened 2026-07-13
fix(react-core): scope queued MCP ui/message follow-ups to enqueue-time thread (#5819)

### LINKED ISSUE #5819: 🐛 Bug: MCP app ui/message follow-up work can lose its enqueue-time thread context
### ♻️ Reproduction Steps

On latest origin/main:

```text
c0d8eaa44 feat(memory): user-scoped long-term memory (core store, useMemories/injectMemories, inspector tab, runtime endpoints) (#5667)
```

I added a focused regression probe to:

```text
packages/react-core/src/v2/components/chat/__tests__/MCPAppsUiMessage.e2e.test.tsx
```

The test shape is:

1. Render an MCP app activity message with a mock agent whose threadId starts as "thread-A".
2. Send a JSON-RPC ui/message request from the iframe.
3. Force the MCP follow-up work through the queue by making the agent appear busy/running.
4. Before the queued work executes, mutate the same agent's threadId to simulate the host/chat moving to a different thread.
5. Assert that the delayed runAgent call still executes under the thread that was active when the MCP ui/message was enqueued.

The important implementation detail is in MCPAppsActivityRenderer.tsx:

```ts
const threadId = agent.threadId || "default";
this.processQueue(threadId, agent);
```

The queue is keyed by the thread at enqueue time, but the delayed work still executes through a mutable agent object later.

Focused command:

```bash
pnpm --filter @copilotkit/react-core exec vitest run \
  src/v2/components/chat/__tests__/MCPAppsUiMessage.e2e.test.tsx \
  -t "keeps queued ui/message follow-up work scoped"
```

### ✅ Expected Behavior

Queued MCP app work should preserve the thread context from the moment the work was created.

In plain terms:

```text
MCP ui/message born in thread-A
-> waits for current run to finish
-> user/app switches to another thread
-> queued follow-up still runs as thread-A
```

This seems especially important for MCP app ui/message follow-up behavior, because a message sent from an embedded app should not accidentally execute against whichever thread the shared agent object happens to point at later.

At minimum, if this is not the intended contract, it would be helpful to document the recommended host-app pattern for avoiding cross-thread execution when MCP app work is queued.

### ❌ Actual Behavior

The focused probe fails cleanly. The delayed runAgent call does not preserve "thread-A".

Observed failure:

```text
expected 'thread_1783183784209_akt2yg4eg' to be 'thread-A'
```

So the queued work appears to execute using a later/current agent thread value rather than the thread that was active when the MCP ui/message was enqueued.

This suggests a possible thread-context leak in the MCP app activity queue:

```text
message born in thread-A
-> queued while agent is busy
-> agent/thread context changes
-> queued follow-up executes under a different thread
```

### Screenshots

_No response_

### 𝌚 CopilotKit Version

```shell
Local source base used for the probe:


origin/main c0d8eaa44
```

### 📄 Logs (Optional)

```shell
Focused Vitest output:


FAIL src/v2/components/chat/__tests__/MCPAppsUiMessage.e2e.test.tsx
  MCP Apps ui/message followUp behavior
    x keeps queued ui/message follow-up work scoped to the thread active at enqueue time

AssertionError:
expected 'thread_1783183784209_akt2yg4eg' to be 'thread-A'

Test Files  1 failed (1)
Tests       1 failed | 5 skipped (6)


## Notes

This may be related to MCPAppsRequestQueue storing queues by threadId, while queued callbacks still close over a mutable AbstractAgent.

A possible fix shape may be to capture the enqueue-time thread ID and ensure delayed runAgent / proxied MCP requests execute with that captured identity, rather than reading agent.threadId at execution time.
```

BenTaylorDev:
Fixes #5819.

## Problem

An MCP app `ui/message` follow-up is queued (via the module-level MCP request queue) until the agent is idle. The queued closure ran `copilotkit.runAgent({ agent })` against the **shared registry agent** — reading `agent.threadId`/`agent.messages` at *execution* time, not enqueue time.

CopilotKit keeps a single shared agent per id and mutates its `threadId`/`messages` on a thread switch (`CopilotChat.tsx` sets `agent.threadId = resolvedThreadId` and `setMessages(...)`). So if the host switches threads while a follow-up is queued (agent busy), the follow-up — and its streamed events — execute against and render into the **now-foreground thread**. That's the leak in #5819.

### Why this is a regression

This class of bug was previously prevented by the per-thread clone routing in the activity renderers (#3630); commit `762370a4e5` **deliberately** reverted that ("remove per-thread agent cloning, restore single registry agent per id"), re-opening CPK-7155/#2957. Re-introducing implicit clones would undo that decision.

## Fix — skip stale follow-ups

Capture the thread the follow-up is enqueued for, and at execution compare it against the shared agent's current thread (`ɵrunMcpFollowUp`):

- **Same thread** (common case) → run on the shared agent, unchanged (still streams into the visible chat).
- **Thread changed** → the shared agent has moved on and can no longer run the follow-up in its originating thread's context, so **drop it** (with a warning) rather than leak it into the current thread. The MCP app already received its `ui/message` ack at enqueue time; only the optional agent turn on an abandoned thread is skipped.

### Why not re-home the run to the original thread?

An earlier revision re-homed the stale run onto an isolated `registerProxiedAgent` proxy pinned to the origin thread. Review (thanks @tylerslaton) surfaced two P1s that are inherent to *running* a stale follow-up, both confirmed:

1. **Tools/context leak.** `buildFrontendTools`/`getContextForAgent` include *unscoped* entries (the `useFrontendTool`/`useAgentContext` default) for any agent id, read live — so the background thread-A run received thread-B's current context and could execute global frontend handlers against B.
2. **Lost continuity.** With multiple queued follow-ups, a fresh per-item proxy seeded from an enqueue-time snapshot meant run 2 never saw run 1's result (the runtime replaces the clone with `input.messages`/`input.state`, so persistence didn't recover it).

Both dissolve if the stale follow-up simply doesn't run. Skip-stale is also far smaller and needs no proxy lifecycle, so it doesn't fight the post-revert "single registry agent" architecture.

## Testing

- **TDD.** The e2e regression test reproduced #5819 on the pre-fix code (`run` invoked once with the switched-to `threadId`) and passes after.
- **Unit** (`mcp-followup.test.ts`) — `ɵrunMcpFollowUp` runs on the shared agent when the thread matches (incl. the `undefined`→`"default"` case) and drops (no run, warns) when it changed.
- **e2e** (`MCPAppsUiMessage.e2e.test.tsx`) — "does not run a queued ui/message follow-up against a thread the host switched to (#5819)".
- 30/30 across the MCP + follow-up suites; existing ui/message tests unchanged. `tsc` 0 errors, `oxlint` 0 errors (my files), `oxfmt` clean.


packages/react-core/src/v2/components/MCPAppsActivityRenderer.tsx
```diff
@@ -3,8 +3,91 @@
...
+      const proxyId = `${agent.agentId ?? runtimeAgentId}::mcp-followup::${capturedThreadId}::${crypto.randomUUID()}`;
+      const { agent: scopedAgent, unregister } = host.registerProxiedAgent({
+        agentId: proxyId,
+        runtimeAgentId,
+      });
+      try {
+        scopedAgent.threadId = capturedThreadId;
+        scopedAgent.setMessages(capturedMessages);
+        scopedAgent.setState(capturedState);
+        return await host.runAgent({ agent: scopedAgent });
```
tylerslaton: [P1] Suppress live global tools/context for the background run

`core.runAgent` still injects `buildFrontendTools(scopedAgent.agentId)` and `getContextForAgent(scopedAgent.agentId)`. Both include unscoped entries, and `useFrontendTool`/`useAgentContext` register unscoped by default. After A→B, this thread-A run therefore receives B's current readable context and may execute global frontend handlers against B. The distinct proxy id only filters entries explicitly scoped to the original agent. Please suppress these for the re-homed run (as the PR description says), or capture thread-A values at enqueue time, and cover the outbound `tools`/`context` payload.

packages/react-core/src/v2/components/MCPAppsActivityRenderer.tsx
```diff
@@ -3,8 +3,91 @@
...
+    const runtimeAgentId = agent.runtimeAgentId ?? agent.agentId;
+    if (runtimeAgentId) {
+      const proxyId = `${agent.agentId ?? runtimeAgentId}::mcp-followup::${capturedThreadId}::${crypto.randomUUID()}`;
+      const { agent: scopedAgent, unregister } = host.registerProxiedAgent({
+        agentId: proxyId,
+        runtimeAgentId,
+      });
+      try {
+        scopedAgent.threadId = capturedThreadId;
+        scopedAgent.setMessages(capturedMessages);
```
tylerslaton: [P1] Carry earlier queued results into later follow-ups

Each queued `ui/message` captures before any queued item executes, but this branch creates and unregisters a fresh proxy per item. With two requests queued on A, run 1's generated messages/state live only on proxy 1; proxy 2 is then reset to its stale snapshot and runs without run 1's response/state. The runtime replaces the request clone with `input.messages`/`input.state`, so persisted history does not fix the model context. Please reuse a per-thread proxy across the queue drain or otherwise feed each result into the next item, with a two-item regression test.

tylerslaton (changes requested):
The single-run thread routing and cleanup are well covered, and the full React Core test/type-check targets pass locally. I found two blocking isolation/order issues in the re-home path.

packages/react-core/src/v2/components/MCPAppsActivityRenderer.tsx
BenTaylorDev: Confirmed — `buildFrontendTools` filters `(!tool.agentId || tool.agentId === agentId)` and `getContextForAgent` filters `!entry.agentIds || …`, so unscoped tools/context (the `useFrontendTool`/`useAgentContext` default) get injected for the proxy id too, live from the current (thread-B) state. The re-home was leakier than the PR claimed.

Rather than suppress-and-run, I removed the re-home entirely and switched to **skip-stale**: when the shared agent's `threadId` no longer matches the thread the follow-up was enqueued for, the follow-up is dropped (with a warning) instead of executed. No background run → no tools/context to leak. This also resolves the sibling P1 below by construction.

packages/react-core/src/v2/components/MCPAppsActivityRenderer.tsx
BenTaylorDev: Confirmed — fresh per-item proxy + enqueue-time snapshot meant run 2 ran on a proxy reset to a pre-run-1 snapshot, and (as you note) the runtime replaces the clone with `input.messages`/`input.state`, so persistence didn't recover the model context.

Resolved by dropping the re-home path: stale follow-ups are now skipped rather than run, so there are no accumulated-across-queue results to carry. The queued-continuity problem only existed for the background re-homed runs, which no longer happen. The e2e regression test asserts a queued follow-up does not run against a switched-to thread.

tylerslaton (approved):
Re-reviewed the latest rebased head. Replacing the proxy re-home with skip-stale removes both prior blockers by construction: no background run can receive the active thread's global tools/context, and there is no per-item proxy state to lose across queued follow-ups. The enqueue-time thread guard is covered at the helper and renderer levels. Full local React Core Nx test/type-check validation passed (114 test files, 1,428 tests).

PR #5942, opened 2026-07-13
fix(core,react-core): defer runtime /info out of the core constructor (#5801)

### LINKED ISSUE #5801: Runtime /info fired from the core constructor during render → duplicate runtime connections (dozens of /info) under concurrent rendering — @copilotkitnext 1.54
### Summary

`CopilotKitProvider` constructs the runtime core **as a side-effect during render**, and that constructor synchronously starts a network request (`/info`). Because React can start-and-discard renders (concurrent rendering in production builds, Suspense, StrictMode), each discarded mount attempt re-runs the construction and fires **another** `/info` request. Combined with `updateRuntimeConnection()` having no in-flight de-duplication, a single page load can fire **dozens** of runtime `/info` requests instead of one.

We observed **70–80 `POST .../info` requests on one page load** in a production build, against a self-hosted runtime with realistic network latency.

### Package / version

- `@copilotkitnext/react` **1.54.0**
- `@copilotkitnext/core` **1.54.0**
- React 19 concurrent renderer, Next.js 16 (App Router) **production build**

### Root cause

**1. The core is constructed during render** — `providers/CopilotKitProvider`:

```js
const copilotkitRef = useRef(null);
if (copilotkitRef.current === null)
  copilotkitRef.current = new CopilotKitCoreReact({ runtimeUrl, ... }); // side effect in render
```

For a **mounting** fiber there is no persisted `useRef` yet, so if React discards the in-progress render (interrupted by a higher-priority update, a suspended child, or StrictMode’s double-invoke) and re-runs it, `copilotkitRef.current` is `null` again → **a brand-new core is constructed on each discarded attempt**.

**2. The constructor kicks off a network request** — `CopilotKitCoreReact` ctor calls `agentRegistry.setRuntimeUrl(runtimeUrl)`, and:

```js
setRuntimeUrl(runtimeUrl) {
  const normalized = runtimeUrl ? runtimeUrl.replace(/\/$/, "") : void 0;
  if (this._runtimeUrl === normalized) return;
  this._runtimeUrl = normalized;
  this.updateRuntimeConnection();   // → fetch(.../info)
}
```

So every discarded-and-recreated core immediately issues its own `/info` request.

**3. `updateRuntimeConnection()` has no in-flight guard** — it unconditionally sets status to `Connecting` and fetches, even if a previous call is still in flight:

```js
async updateRuntimeConnection() {
  if (typeof window === "undefined") return;
  if (!this.runtimeUrl) { /* disconnect */ return; }
  this._runtimeConnectionStatus = Connecting;   // no "already connecting/connected?" bail-out
  const runtimeInfoResponse = await this.fetchRuntimeInfo();
  ...
}
```

The number of duplicate `/info` calls scales with (a) how long the provider’s mount render stays interruptible and (b) `/info` latency — a slow runtime lets many pile up before the first resolves.

### Steps to reproduce

1. Wrap a **large** app subtree as `children` of `<CopilotKit runtimeUrl=... />` (so the provider’s mount render is long / interruptible).
2. Build for **production** (concurrent renderer) — a dev build usually commits the first render, hiding this.
3. Make `/info` slow (real backend RTT, or a cold serverless route).
4. Trigger boot-time state updates around mount (store hydration, data fetching) so the mount render is interrupted.
5. Hard-load the page and watch the network tab.

**Expected:** exactly one runtime connection / one `/info` per provider mount.
**Actual:** N cores constructed and N `/info` requests (we saw 70–80 in ~800ms), all with `method: "info"`.

### Suggested fix

- Don’t perform network I/O from the core constructor. Separate **construction** (pure) from **connection** — start `updateRuntimeConnection()` from an effect (`useEffect`) rather than synchronously in the ctor / during render.
- Add an in-flight/idempotency guard to `updateRuntimeConnection()` (bail if already `Connecting`, or de-dupe by `runtimeUrl`), so concurrent/duplicate invocations collapse to one request.
- Consider constructing the core lazily in a way that tolerates discarded renders (and dispose orphaned instances).

### Workaround (app side)

Rendering `<CopilotKit>` with **no children on the first commit** and mounting children one tick later (inside a `<Suspense>` boundary) shrinks the mount render so it commits before it can be interrupted — this collapsed our count from ~78 `/info` to **1**. It’s a mitigation, not a fix for the underlying render-time side effect.

BenTaylorDev:
Fixes #5801.

## Problem

`CopilotKitProvider` constructs the core **during render** (the `useRef` init pattern), and the `CopilotKitCore` constructor synchronously starts the `/info` network request. React can start-and-discard renders (concurrent rendering / Suspense / StrictMode), and for a *mounting* fiber `copilotkitRef.current` is `null` again on each discarded attempt — so a brand-new core is constructed, and fires its own `/info`, per discarded render. The reporter observed **70–80 `/info` requests on a single page load** (production build, self-hosted runtime with real latency).

Root causes (all confirmed on current `main`):
1. Core constructed during render — `CopilotKitProvider.tsx`.
2. Constructor kicks off network I/O — ctor → `setRuntimeUrl` → `updateRuntimeConnection()` → `fetch(/info)`.
3. `updateRuntimeConnection()` had no in-flight guard.

The dominant multiplier is **cross-instance** (N discarded renders → N cores → N fetches), so an in-flight guard alone can't fix it — the fetch must not fire during construction.

## Fix

Separate **construction** (pure) from **connection** (network I/O), as the issue recommends:

- **core** — new `deferInitialConnection` option: the constructor records the runtime config (so `copilotkit.runtimeUrl` stays available *synchronously* to hooks like `use-threads` / the CopilotChat context — no first-render regression) but does **not** start the `/info` fetch. A new `connect()` method starts the single connection; it's idempotent (no-op unless status is `Disconnected`, which `updateRuntimeConnection` flips to `Connecting` synchronously), so a StrictMode double-invoked mount effect collapses to one request. `updateRuntimeConnection` also gains an in-flight guard keyed by `url + requested transport` so concurrent same-target calls de-dupe (a change to a different target supersedes).
- **react-core** — the provider constructs with `deferInitialConnection: true` and calls `copilotkit.connect()` from its existing commit-phase mount effect. Renders discarded before commit never reach the effect, so they never fetch. The subscribe effect runs before `connect()`, so the status-change event is caught (no reliance on the catch-up read for the common path).

**Backward compatible:** without `deferInitialConnection`, the constructor still connects — Vue/Angular/vanilla consumers are unaffected.

## Testing

- **TDD.** New core tests fail on the pre-fix code (constructor fetches) and pass after.
- **core** (`core-defer-runtime-connection.test.ts`, 7 tests) — the authoritative proof of the mechanism: deferred construct fires **zero** `/info`; `runtimeUrl` still exposed synchronously; `connect()` → exactly one `/info`; `connect()` idempotent across repeated calls; **N orphaned cores that never `connect()` → 0 fetches, only the committed one fetches** (directly models the discarded-render bug); in-flight de-dupe; backward-compat.
- **react-core** (`CopilotKitProvider.deferWiring.test.tsx`) — asserts the provider constructs with `deferInitialConnection: true` and drives `connect()` from an effect; **this is the provider-level test that distinguishes the fix from the bug** (it fails if the deferral wiring is dropped — verified by reverting).
- **react-core** (`CopilotKitProvider.deferConnection.test.tsx`, 2 tests) — normal-mount regression guards: exactly one `/info` on mount, and one under `React.StrictMode`.

  *Scope note (found via adversarial review):* the raw dozens-of-`/info` symptom only manifests with React *discarding* in-progress renders and constructing multiple orphaned cores — Testing Library can't reproduce that (a single committed mount yields one `/info` either way, and the ctor's fetch is several microtasks deep so ordering can't distinguish it). Hence the multi-instance proof lives in the core test above, and the provider is guarded by the wiring assertion.

- Full suites green: **core 576/576**, **react-core provider 141/141** (updated one fake-core mock to implement `connect()`). `tsc` 0 errors (core + react-core); `oxlint` 0 errors; `oxfmt` clean.

### DIFF
--- a/packages/core/src/__tests__/core-defer-runtime-connection.test.ts
+++ b/packages/core/src/__tests__/core-defer-runtime-connection.test.ts
@@ -1,0 +1,147 @@
+/**
+ * Deferred runtime connection (#5801).
+ *
+ * `CopilotKitProvider` constructs the core during React's render phase, and
+ * React can start-and-discard renders (concurrent rendering, Suspense,
+ * StrictMode). When the constructor fires the `/info` request synchronously,
+ * every discarded-and-recreated core issues its own request — a single page
+ * load was observed firing 70-80 `/info` requests.
+ *
+ * The fix separates construction (pure) from connection (network I/O):
+ * `deferInitialConnection` lets the constructor set the runtime config WITHOUT
+ * fetching, and `connect()` — driven from a commit-phase effect — starts the
+ * single connection for the surviving instance. `connect()` is idempotent so
+ * StrictMode's double-invoked effect collapses to one request.
+ */
+import { describe, it, expect, beforeEach, afterEach, vi } from "vitest";
+import { CopilotKitCore } from "../core";
+
+const RUNTIME_URL = "https://runtime.example/rest";
+const infoResponse = { version: "1.0.0", agents: {} };
+
+describe("CopilotKitCore — deferred runtime connection (#5801)", () => {
+  const originalWindow = (globalThis as { window?: unknown }).window;
+  const originalFetch = global.fetch;
+  let fetchMock: ReturnType<typeof vi.fn>;
+
+  beforeEach(() => {
+    (globalThis as { window?: unknown }).window = {};
+    fetchMock = vi.fn().mockResolvedValue(
+      new Response(JSON.stringify(infoResponse), {
+        status: 200,
+        headers: { "content-type": "application/json" },
+      }),
+    );
+    global.fetch = fetchMock as unknown as typeof fetch;
+  });
+
+  afterEach(() => {
+    vi.restoreAllMocks();
+    global.fetch = originalFetch;
+    if (originalWindow === undefined) {
+      delete (globalThis as { window?: unknown }).window;
+    } else {
+      (globalThis as { window?: unknown }).window = originalWindow;
+    }
+  });
+
+  const infoCalls = () =>
+    fetchMock.mock.calls.filter(([u]) => String(u).includes("/info"));
+
+  it("does not fetch /info from the constructor when deferInitialConnection is set", async () => {
+    const core = new CopilotKitCore({
+      runtimeUrl: RUNTIME_URL,
+      runtimeTransport: "rest",
+      deferInitialConnection: true,
+    });
+    expect(core.runtimeUrl).toBe(RUNTIME_URL);
+    // Wait a macrotask — a constructor-initiated fetch fires after an internal
+    // await, so a microtask wait would miss it and pass falsely.
+    await new Promise((r) => setTimeout(r, 20));
+    expect(infoCalls().length).toBe(0);
+  });
+
+  it("still exposes runtimeUrl synchronously when deferred", () => {
+    const core = new CopilotKitCore({
+      runtimeUrl: RUNTIME_URL,
+      runtimeTransport: "rest",
+      deferInitialConnection: true,
+    });
+    expect(core.runtimeUrl).toBe(RUNTIME_URL);
+  });
+
+  it("connect() triggers exactly one /info fetch after a deferred construct", async () => {
+    const core = new CopilotKitCore({
+      runtimeUrl: RUNTIME_URL,
+      runtimeTransport: "rest",
+      deferInitialConnection: true,
+    });
+    core.connect();
+    await vi.waitFor(() => expect(infoCalls().length).toBe(1));
+    expect(infoCalls()[0]?.[0]).toBe(`${RUNTIME_URL}/info`);
+  });
+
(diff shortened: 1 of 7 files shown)


tylerslaton (approved):

tylerslaton (commented):
Overall, this cleanly moves the initial `/info` request out of the React render phase, and the orphan-core plus provider-wiring coverage checks the actual regression.

Tests (non-blocking): `packages/core/src/__tests__/core-defer-runtime-connection.test.ts:133` does not exercise the new in-flight guard. The core is constructed with `runtimeTransport: "rest"`, so `setRuntimeTransport("rest")` is a no-op, and repeated `connect()` calls are stopped by `connectRuntime()` before `updateRuntimeConnection()` is reached. This test would still pass if `_connectionInFlight` were removed; please either remove the unnecessary guard or add coverage that actually reaches it.

PR #5954, opened 2026-07-14
fix(showcase): consolidate upstream provider keys to Railway shared vars (credential hygiene)

jpr5:
## Problem

The **same real upstream OpenAI/Anthropic/Google key** is duplicated in plaintext across ~34 showcase services' Railway env, while **aimock** — the only service that ever reaches real providers — holds none.

Empirically verified on live staging (values never printed; compared by sha256 prefix):

| Key | Distinct values | # services holding it |
|---|---|---|
| `OPENAI_API_KEY` | 2 (one value on 34, one outlier) | 35 |
| `ANTHROPIC_API_KEY` | 1 | 34 |
| `GOOGLE_API_KEY` | 1 | 28 |

aimock staging carries **zero** provider keys. Even the `shell`/`docs`/`dashboard` frontends carry the real keys. Net: one scraped demo env = the real key; rotation = ~34 edits.

## STEP 1 — Feasibility finding: aimock own-key = **NO**

Can aimock authenticate upstream with its OWN key instead of forwarding the client's? **No.** In `@copilotkit/aimock@1.26.1`:

- The proxy passthrough (`dist/recorder.js`) copies every non-stripped client header to upstream. `STRIP_HEADERS` strips hop-by-hop headers but **not** `authorization` / `x-api-key` — the client key is forwarded verbatim.
- No `process.env.OPENAI_API_KEY` / `ANTHROPIC_API_KEY` / `GEMINI_API_KEY` is read anywhere on the proxy path.
- The CLI `--provider-openai/anthropic/gemini` flags set upstream **URLs only**; there is no key flag.

So aimock cannot own upstream auth. Per the task's fallback instruction, this PR implements the **lighter design**: define each provider key **once** as a Railway **environment-shared variable** and reference it (`${{shared.OPENAI_API_KEY}}`) from the services that need it — single source, one-edit rotation, a scraped service env yields only a reference token.

## What this PR changes (repo side — modeling + guardrail)

The real key *values* live only in Railway, so the durable repo-side change is to **model and enforce** the single-source arrangement:

- **`railway-envs.ts`** — additive `sharedRefs?: string[]` on each service (the provider-key env-var NAMES it sources from a shared variable), a `PROVIDER_KEY_SHARED_VARS` allow-list, a `sharedRefsFor` accessor, and an `assertSharedRefsValid` module-load invariant. Populated from live staging reality (OPENAI everywhere; ANTHROPIC/GOOGLE where present; `strands-typescript` = OPENAI only).
- **`emit-railway-envs-json.ts`** — emit `sharedRefs` additively (frozen legacy shape preserved; byte-stability golden still passes).
- **`railway-envs.generated.json`** — regenerated deterministically via the generator.
- **Tests** — red-green coverage: every aimock-routed service must declare a shared `OPENAI_API_KEY` ref; aimock declares none; malformed/duplicate refs fail loud.

Values live **only in Railway, never in this repo.**

## Red / Green

Exercised at the SSOT/wiring level (a live provider passthrough was not run in-session; see note below).

- **RED** — with the `sharedRefs` data stripped (the pre-fix state where each service still models its own per-service key), the test *"every aimock-routed service declares OPENAI_API_KEY as a shared ref"* fails, listing all 30+ agent/starter services as offenders.
- **GREEN** — with `sharedRefs` populated, all 8 new tests pass; the full `railway-envs` + `emit` suites pass (127/127), including the byte-stability golden and `emit --check`.

**Not live-validated in-session:** an end-to-end passthrough (demo backend → aimock → real provider using the shared key) requires creating the Railway shared variable and switching a service to `${{shared.*}}`, which is a prod/staging mutation intentionally left as the manual ops step below. The wiring/assertion level is proven here.

## Manual follow-up (ops — not done in this PR)

- [ ] Create environment-shared variables `OPENAI_API_KEY`, `ANTHROPIC_API_KEY`, `GOOGLE_API_KEY` in each Railway env (staging + prod), sourced from 1Password.
- [ ] Switch every service listed in `sharedRefs` to reference `${{shared.<NAME>}}` and delete its per-service copy.
- [ ] Live-validate one demo backend passthrough (fixture-miss → real completion) with only the shared reference present.
- [ ] **MANUAL: rotate the exposed key.** The real `OPENAI_API_KEY` surfaced in plaintext across ~34 services in a debugging session — rotate it in the OpenAI dashboard, store the new value in 1Password, and reference it via the shared variable (never commit it).

## Coordination

Independent of and mergeable in **any order** with `fix/aimock-private-egress`. This PR touches only the `*_API_KEY` modeling (`sharedRefs`) — it does **not** modify base-URL host resolution. The one shared artifact, `railway-envs.generated.json`, is deterministically regenerable via the generator on the merged SSOT (never hand-merge).

### DIFF
--- a/showcase/aimock/RAILWAY.md
+++ b/showcase/aimock/RAILWAY.md
@@ -208,6 +208,47 @@ None are required for the default configuration. Notes:
   network.
 - `PORT` is injected by Railway but not read by the current startCommand
   (port is hardcoded to `4010`). Harmless.
+- **No provider API keys.** aimock holds NO `OPENAI_API_KEY` /
+  `ANTHROPIC_API_KEY` / `GOOGLE_API_KEY`. On the proxy-fallthrough path it does
+  NOT authenticate with its own key — it **forwards the client-supplied
+  `Authorization` / `x-api-key` header verbatim** to the real provider (see
+  §6.1). The real key therefore lives on the CLIENT services, not here.
+
+### 6.1 Upstream auth model + provider-key credential hygiene
+
+aimock cannot own upstream provider auth. In `@copilotkit/aimock`, the
+proxy-fallthrough recorder copies every non-hop-by-hop client header straight
+to the upstream provider; the header strip-list does NOT include
+`authorization` / `x-api-key`, and there is no CLI flag or env var to inject an
+aimock-owned provider key (`--provider-openai/anthropic/gemini` set upstream
+**URLs** only). So a fixture-miss request reaches the real provider using
+**whatever key the calling service sent**.
+
+Because every showcase agent/starter (and the `shell`/`docs`/`dashboard`
+frontends) routes provider traffic through aimock, each one must carry a
+provider key for those fallthrough requests. To avoid duplicating the SAME real
+secret across ~34 services (one scraped env = the real key; rotation = ~34
+edits), each provider key is defined **once** as a Railway
+**environment-level SHARED variable** and referenced from the services that
+need it:
+
+- Create `OPENAI_API_KEY`, `ANTHROPIC_API_KEY`, `GOOGLE_API_KEY` as
+  environment-shared variables in each env (staging + prod), sourced from
+  1Password — never committed to this repo.
+- Each consuming service references `${{shared.<NAME>}}` instead of holding its
+  own literal copy. The SSOT (`showcase/scripts/railway-envs.ts`) models this
+  per-service via the additive `sharedRefs` field (the provider-key NAMES a
+  service resolves from the shared variable), and the Ruby promote preflight
+  asserts each listed key resolves to the shared variable rather than a
+  distinct per-service literal.
+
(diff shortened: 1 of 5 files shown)


jpr5:
Closing this in favor of a cleaner fix. We dug into aimock's source and found it currently forwards the caller's `Authorization`/`x-api-key` header verbatim on passthrough (`recorder.ts:337` `buildForwardHeaders`) — it has no own-key surface, which is *why* the real key has to sit on every demo backend.

The shared-variable approach here still resolves the real key into all ~20 backend runtimes, so it only buys one-edit rotation, not a smaller blast radius. Turns out aimock can own the key with a small (~1-file) change: aimock holds the one real key, backends carry only the `sk-aimock-*` dummy, and aimock swaps its key in on a fixture-miss passthrough (a real caller key still overrides). That's the real consolidation.

Superseded by the forthcoming aimock own-key PR + a showcase config flip. Spec incoming.

PR #5957, opened 2026-07-14
fix(showcase): resolve dashboard cell-model fold under Turbopack (next dev)

jpr5:
## The webpack-only gap

PR #5955 fixed the dashboard's `next build` (webpack) resolution of the shared cell-model fold's internal `.js` specifiers by adding `webpack.resolve.extensionAlias` to `next.config.ts`. That fix is **webpack-only**.

This package's `dev` script is `next dev --turbopack --port 3002` — so **local `next dev` runs under Turbopack**, which ignores the webpack callback entirely. The fold files (`harness/src/shared/cell-model/{cell-model,live-status,staleness,format-ts}.ts`) carry explicit `.js` extensions on their internal relative imports (correct for the harness's pure-Node-ESM runtime); `export *` in the dashboard's `src/lib/*` shims pulls those internal `.js` edges into the dashboard graph. Under Turbopack they failed to resolve.

## Empirical RED (Turbopack, current main tip)

`next build --turbopack` (deterministic repro of the dev resolver), with #5955's webpack `extensionAlias` present but ignored:

```
Module not found: Can't resolve './live-status.js'
Module not found: Can't resolve './staleness.js'
Module not found: Can't resolve './format-ts.js'
  at .../harness/src/shared/cell-model/cell-model.ts:17:1
  at .../harness/src/shared/cell-model/live-status.ts:17:1
> Build error occurred
```

## Why the obvious Turbopack knobs don't work

Turbopack has **no `extensionAlias` equivalent** — mapping a relative explicit `.js` specifier to its `.ts` source is an open, unimplemented feature request: [vercel/next.js#82945](https://github.com/vercel/next.js/issues/82945). Verified empirically that all of these leave the identical 3 RED lines:

- `turbopack.resolveAlias { "./x.js": "./x.ts" }` (and absolute-path / extensionless variants) — resolveAlias does not match relative specifiers
- `turbopack.resolveExtensions` with the full default list — only affects **extensionless** imports, not a `.js`→`.ts` remap

## The fix

The documented workaround from #82945: a `turbopack.rules` transform loader (`turbopack-strip-js-ext-loader.cjs`) that strips the trailing `.js` off **relative** import/re-export/dynamic-import specifiers, so Turbopack's normal `.ts`/`.tsx`/`.js` extension resolution takes over. The rule glob scopes the loader to the fold sources **only** (`**/harness/src/shared/cell-model/*.ts`); no other dashboard module is rewritten. The webpack `extensionAlias` from #5955 is retained unchanged. The fold's on-disk `.js` specifiers are untouched (still correct for the harness Node-ESM runtime).

## GREEN

- `next build --turbopack` → `✓ Compiled successfully in 2.5s` (fold errors gone)
- `next dev --turbopack` → route `/` returns HTTP 200, `✓ Compiled / in 2.3s`, zero `Module not found` (the real reviewer-flagged surface: `dashboard-page.tsx` → `src/lib` shims → harness fold compiles cleanly)

## No regressions

- **Webpack path** (`next build`) still `✓ Compiled successfully` — #5955's fix is not disturbed.
- **Harness fold untouched** — `git diff` on the fold is empty; the `.js` internal imports remain.
- The only error remaining on **both** bundlers is a pre-existing, orthogonal `Cannot find module 'croner'` in test/CI type-check (present on the #5955 baseline with these changes stashed — 2 test files fail identically before and after). Not caused or fixed by this PR.

### DIFF
--- a/showcase/shell-dashboard/next.config.ts
+++ b/showcase/shell-dashboard/next.config.ts
@@ -22,18 +22,34 @@ import type { NextConfig } from "next";
  *   2. The ops base URL stays out of the client bundle (no `NEXT_PUBLIC_*`
  *      exposure).
  *
- * `webpack.resolve.extensionAlias`: the dashboard re-exports the shared
- * cell-model fold from the harness (`src/lib/{cell-model,live-status,
- * staleness,format-ts}.ts` → `../../../harness/src/shared/cell-model/*`).
- * Those fold files are authored for the harness's pure-Node-ESM runtime, so
- * their INTERNAL relative imports carry explicit `.js` extensions (e.g.
- * `import { formatTs } from "./format-ts.js"`). `export *` does NOT rewrite
- * those internal edges, so `next build`'s webpack sees a literal `./x.js`
- * specifier that only exists on disk as `./x.ts` and fails to resolve. The
- * extensionAlias tells webpack to try the TypeScript sources when a `.js`
- * specifier is requested — the standard bundler complement to TS's
- * NodeNext `.js`-import convention. This covers the `next build` (webpack)
- * path that CI uses.
+ * === Resolving the shared cell-model fold's internal `.js` specifiers ===
+ *
+ * The dashboard re-exports the shared cell-model fold from the harness
+ * (`src/lib/{cell-model,live-status,staleness,format-ts}.ts` →
+ * `../../../harness/src/shared/cell-model/*`). Those fold files are authored
+ * for the harness's pure-Node-ESM runtime, so their INTERNAL relative imports
+ * carry explicit `.js` extensions (e.g. `import { formatTs } from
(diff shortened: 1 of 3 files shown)


jpr5:
Superseded by #5959 — closing this regex/mask Turbopack loader. The 3-round CR showed hand-rolled JS-lexing source-rewriting is inherently fragile (raw regex → statement-anchored → mask-then-match, each leaked another JS construct: strings/comments → dynamic import() → regex literals). #5959 instead runs the dashboard dev server on webpack (which already resolves the shared cell-model fold via #5955's resolve.extensionAlias) — no source-rewriter, `pnpm dev` just works. The proper long-term fix (extract the shared cell-model into a real workspace package so the dashboard stops reaching into harness/src raw source) is being specced separately.

PR #5925, opened 2026-07-11
docs(channels): clarify managed connection paths

tylerslaton:
## Summary

- clarifies the available paths for operating Channels
- adds ownership and setup guidance for managed connections
- refreshes related terminology and entry points

## Why

The current docs do not clearly explain where provider operations end and application-owned agent behavior begins.

## How

- makes `/channels` the decision page
- adds conceptual and guided managed paths
- aligns adjacent Slack, Intelligence, and Channels reference material
- validates the shell-docs build and leaves a local preview running

A small generated-handoff follow-up is tracked in [#5926](https://github.com/CopilotKit/CopilotKit/issues/5926). This PR will remain draft until that alignment lands.

PR #5963, opened 2026-07-14
feat(runtime): handler owns managed channels activation & lifecycle (OSS-473)

BenTaylorDev:
## Summary

Makes the normal `createCopilotRuntimeHandler` own Managed Channels activation, readiness, reconnect, and shutdown. A developer declares a Channel next to their agents in an existing `CopilotRuntime`, adds `intelligence`, and mounts the normal handler — **no separate launcher, no gateway URLs, no org/project/channel/runtime-instance IDs**.

```ts
const support = createChannel({ name: "support", agent: () => supportAgent });
const runtime = new CopilotRuntime({ agents, intelligence, identifyUser, channels: [support] });
export const handler = createCopilotRuntimeHandler({ runtime, basePath: "/api/copilotkit" });

await handler.channels.ready({ timeoutMs: 10_000 });
handler.channels.status();
await handler.channels.stop();
```

Implements the CopilotKit-SDK half of the Managed Channels SoT **workstream A**. Closes OSS-473.

## What changed

- **`channels:` runtime option** on the Intelligence runtime (public alias replacing the internal `bots` field); the SSE runtime rejects it.
- **`ChannelManager`** (`packages/runtime/.../core/channel-manager.ts`) owns the managed lifecycle: lazy, idempotent `activate()`; `ready({ timeoutMs })`; `status()` (`connecting | online | setup_required | reconnecting | stopped | unmanaged | error`); idempotent, resilient `stop()`. Activation config (`wsUrl`, `apiKey`, `projectId`, `channelName`, `provider`, `runtimeInstanceId`) is **derived from the `intelligence` config + the declared channel** — `projectId` is parsed from the `cpk-{projectId}_` API-key prefix, and the managed `provider` is declared **per-Channel** via `createChannel({ provider })` (type `ManagedChannelProvider`, defaults to `slack`; `teams` is gated/coordinated — the gateway accepts only `slack` at join today, so `teams` is SDK-ready but not GA until Intelligence OSS-450/#511 lands) — so no infrastructure IDs are supplied by the developer.
- **`createCopilotRuntimeHandler` returns a callable object** — for an Intelligence runtime with declared channels the `channels` control is **non-optional** (`((req) => Promise<Response>) & { channels: ChannelsControl }`; optional for SSE / channel-less runtimes). Activation is **lazy and serverless-safe**: handler creation opens **no** connection; the persistent gateway socket opens on the first `await handler.channels.ready()`, so a Fetch host that cannot own a listener (Cloudflare Workers, Next.js App Router) never opens one. Call `ready()` once at startup on a long-running host. Idempotent per-runtime via a `WeakMap`; additive and non-breaking; propagated through the node/express/hono endpoint wrappers.
- **Clean-break rename** `createBot → createChannel` / `Bot → Channel` (and `CreateChannelOptions`, `ChannelHandler/Component/Tool/Command`, `defineChannelTool/Command`) across `@copilotkit/channels`, `-intelligence`, the adapter packages, `@copilotkitnext/teams`, and the examples — no public `Bot` aliases, per the SoT.
- **`examples/slack`** managed path rewritten to the no-launcher DX (`createChannel` + `channels:` + `createCopilotNodeListener` + `handler.channels.stop()` on SIGTERM/SIGINT); the six `INTELLIGENCE_*` launcher env vars removed.
- The `channels-intelligence` realtime launcher is kept working and is now driven internally by the handler.

## Reconnection & connection health (note for reviewers)

The actual reconnect/rejoin is **delegated to the Phoenix connection layer**, not a runtime-managed backoff loop. Verified against the live gateway contract (Intelligence #511): Phoenix's `Socket` auto-reconnects and auto-rejoins, which re-runs the gateway `join/3` → `record_heartbeat` (re-registers the listener), and `terminate/2` releases the dead socket's lease. A manager-level re-activation loop would be both **redundant** and **incorrect** (the `Channel` is single-start — re-running `addAdapter`/`start` throws).

What the handler owns is **observation of that connection**. The session exposes `onStateChange` and the manager reflects it in `status()`: a dropped socket **or a Phoenix channel-level close/error** moves the channel to `reconnecting`; a successful (re)join moves it back to `online` — so `online` means the managed path can currently send. A **bounded give-up window** (`reconnectGiveUpMs`, default 60s) surfaces a prolonged outage as `error`, but it is **recoverable**: if Phoenix rejoins afterwards the channel returns to `online` — the state always tracks the live transport, never latching. Join rejections are **classified per-Channel** from the gateway's declaration states: genuinely unconfigured/waiting states (`adapter_setup_required`, `*_waiting_for_runtime`, `no_channels_yet`, …) surface as `setup_required` (and `ready()` resolves); a `runtime_conflict` or a `*_failed`/hard state surfaces as an **error** and is never silently downgraded to setup-required. `ready()` keeps one-shot semantics — it settles on the initial activation outcome; later health transitions move only `status()`.

## Testing

Unit + integration: handler-owned activation with config derived purely from `intelligence` (asserts **no** org/channelId on the wire); **first request does not trigger activation**; `ready`/`status`/`stop`; a socket-drop path; duplicate-name fail-loud; setup-required surfacing; the default engine's opts mapping + module-not-found path. `build` + `check-types` + `test` green across `@copilotkit/channels`, `-intelligence`, and `runtime` (1726 tests).

## Residual risk / verification

The dynamic `import()` of `@copilotkit/channels-intelligence`, the real Phoenix socket join, and the example's node listener are proven against fakes but not a live gateway. A **packaged-SDK → Realtime Gateway → provider e2e smoke** is recommended before production reliance (SoT workstream D; pairs with landing Intelligence #511).

## Follow-ups (out of this PR's activation-lifecycle subject)

Filed separately: managed transport parity/reliability — cross-turn history (OSS-436), command/interaction/reaction ingress over the realtime path (OSS-416/419/434), `thread.delete()` egress, empty-turn redelivery, egress-failure fail-loud, durable StateStore on the managed path, and the reconnect-observability / `stop()`-timeout hardenings. Docs update tracked in #5925. Id-less gateway topic in OSS-480.

## Adapter transport model (clarified per review)

This PR is **exclusive per Channel**, not per platform, and there is no simultaneous coexistence. **Any** developer-supplied **direct** adapter on a Channel declared in `channels:` makes the **whole** Channel `unmanaged` and skips it for managed activation — regardless of platform — rather than dual-activating, because attaching the managed gateway transport beside a direct adapter would open two live connections to the provider and deliver every event twice (`assertExclusive` enforces the adapter exclusivity). The skipped Channel is not silently "healthy": it is reported with an explicit **`unmanaged`** status (a runtime whose only Channel is direct reads `overall: "unmanaged"`, never `online`), and the handler neither starts nor stops it — the developer owns its `channel.start()`. Having the handler own the lifecycle of a direct Channel too (either-or per Channel) is **OSS-486**; true managed+direct coexistence on one Channel is **OSS-484**.


packages/runtime/src/v2/runtime/core/channel-manager.ts
```diff
@@ -0,0 +1,624 @@
...
+    // that is NOT the Intelligence managed adapter (a developer-supplied
+    // slack/discord/... adapter, which lacks `__intelligenceChannel`) is a
+    // DIRECT channel: it is started by the developer via `channel.start()`, not
+    // managed-activated here. Managed-activating it would attach a second
+    // (Intelligence) adapter and trip the SDK's `assertExclusive` guard, moving
+    // the Channel to `error`. Per the SoT rule, never infer managed intent from
+    // a local direct adapter — direct adapters remain additive. A managed-eligible
+    // Channel has an empty `adapters` at declaration time.
+    for (const channel of this.channels) {
+      const isDirect = channel.adapters.some((a) => !a.__intelligenceChannel);
```
tylerslaton, line 367: [P1] Preserve managed activation when a direct adapter is present

This makes a developer-supplied adapter disable managed delivery for the entire Channel. As a result, the same Channel cannot use both a direct transport and Intelligence, and it is not even declared to Intelligence. The lower-level `assertExclusive` guard also rejects attaching the Intelligence adapter beside an existing one, so this is not just a status-filtering issue. That contradicts the PR's advertised additive transport behavior; the activation path needs to attach/start the managed adapter alongside direct adapters rather than skip the Channel.

packages/runtime/src/v2/runtime/core/channel-manager.ts
```diff
@@ -0,0 +1,624 @@
...
+    scope: { projectId: config.projectId, channelName: config.channelName },
+    runtimeInstanceId: config.runtimeInstanceId,
+    adapter: config.adapter,
+  });
+}
+
+/** Whether `err` signals a missing managed provider rather than a hard failure. */
+function isSetupRequired(err: unknown): boolean {
+  return (
+    err instanceof ChannelSetupRequiredError ||
```
tylerslaton, line 228: [P1] Make `setup_required` reachable from the production engine

The default path never preserves either signal checked here. `connectRealtimeGateway()` wraps every join rejection in a fresh generic `Error` after formatting the server reason, and `@copilotkit/channels-intelligence` never throws `ChannelSetupRequiredError` or an error with `code === "SETUP_REQUIRED"`. Therefore an unconfigured managed provider becomes `error` and makes `ready()` reject instead of degrading this Channel to `setup_required`. Please translate the gateway's stable setup-required response before discarding its code (and cover the real default-engine path, not only an injected fake).

packages/runtime/src/v2/runtime/core/channel-manager.ts
```diff
@@ -0,0 +1,624 @@
...
+   * and makes NO state change and NO re-activation: reconnection is delegated to
+   * the Phoenix connection layer (see {@link ChannelManager}), which auto-rejoins
+   * the dropped socket under the persistent adapter. Leaving `status` at `online`
+   * is the honest state — the transport self-heals invisibly to the manager.
+   *
+   * @param name - The Channel name (map key).
+   * @param entry - The Channel's activation entry.
+   */
+  private registerOnClose(name: string, entry: ChannelEntry): void {
+    entry.handle?.onClose?.(() => {
```
tylerslaton: [P1] Do not report a permanently disconnected Channel as online

This callback only logs, so after a socket drop `status()` remains `online` forever—even if Phoenix cannot rejoin because the gateway is down, credentials were revoked, or setup changed. `ready()` also continues to resolve. Since this is the public lifecycle/health surface, `online` needs to mean the basic managed path can currently send. Expose reconnect/rejoin outcomes from the session and transition through `reconnecting` to `online` or a bounded `error` state.

packages/runtime/src/v2/runtime/core/fetch-handler.ts
```diff
@@ -112,11 +118,96 @@ export interface CopilotRuntimeHandlerOptions {
...
+ */
+export type CopilotRuntimeFetchHandler = ((
   request: Request,
-) => Promise<Response>;
+) => Promise<Response>) & {
+  /**
+   * Present only when the handler activated managed Channels for an
+   * Intelligence runtime; the lifecycle control surface for those Channels.
+   */
+  channels?: ChannelsControl;
```
tylerslaton, line 179: [P2] Make the advertised `handler.channels` snippet type-check

The factory always returns an optional `channels`, so strict TypeScript rejects the PR's documented `await handler.channels.ready(...)` with “possibly undefined.” The new tests confirm this accidentally by requiring `handler.channels!`, and the example uses `?.`. Please narrow/overload the return type for an Intelligence runtime with declared Channels, or otherwise provide a non-optional control surface for this supported setup, and add a compile-time consumer test using the documented call verbatim.

packages/channels/telemetry-events.json
```diff
@@ -1,12 +1,12 @@
 {
   "global_properties": {
     "anonymous_id": "Persisted per-install UUID (durable store -> project-local cache file -> per-process).",
-    "bot_session_id": "Random UUID minted per createBot() call.",
+    "bot_session_id": "Random UUID minted per createChannel() call.",
```
tylerslaton: [P2] Finish the Channels-only rename in telemetry and diagnostics

This schema still publishes `bot_session_id` and every event remains `oss.bot.*`; the implementation also exposes `[CopilotKit Bot]` / `[bot]` user-facing diagnostics and `BotTelemetry` names. Since this PR explicitly claims a clean-break Channels rename including telemetry, landing this state creates a new permanent Bot vocabulary contract underneath `createChannel`. Rename these surfaces (or explicitly separate the compatibility migration) and update the end-to-end telemetry coverage accordingly.

tylerslaton (changes requested):
The handler-owned lifecycle direction is good, and CI is green, but the current implementation still breaks several contracts advertised by this PR. The inline findings cover three runtime blockers (managed/direct coexistence, setup-required classification, and truthful connection health), plus the public TypeScript surface and the incomplete Channels-only rename.

packages/runtime/src/v2/runtime/core/channel-manager.ts
BenTaylorDev: Agreed that skip ≠ additive — but "attach the managed adapter alongside direct adapters" isn't safe as written, so splitting this into its own ticket (OSS-484) rather than landing it here.

The hazard: `assertExclusive`'s rule is *"Channel and direct delivery are alternative modes **per platform** — Intelligence holds the platform creds, or the runtime does, never both,"* and `createChannel().start()` starts every attached adapter. The `intelligenceAdapter` is a full transport (DeliverySource + EgressSink + RenderEventSink over the gateway), so attaching it beside a direct Slack adapter opens two live connections to the same Slack workspace → every mention delivered twice → the agent runs and replies twice. Compounding, the `intelligenceAdapter`'s `platform` is generically `"intelligence"` (the managed provider is chosen server-side), so a same-platform collision can't be detected by comparing adapter platforms.

So true coexistence is a real feature with a model choice, not a filter tweak: (i) multi-platform transports on one Channel (direct Slack + managed Teams) — needs the managed adapter to declare its target platform + per-platform dedup; or (ii) Intelligence provides Threads/StateStore while the direct adapter owns the transport — matches the SoT direct-delivery prose but needs a new non-transport managed mode. Tracked in **OSS-484** with this analysis.

For this PR I've kept exclusive-per-platform: a Channel carrying a direct adapter is skipped for managed activation with an info-log (no crash, no double delivery), and I'm correcting the PR's "additive" wording to state exclusive-per-platform explicitly with the both-transports caveat pointing at OSS-484.

packages/runtime/src/v2/runtime/core/channel-manager.ts
BenTaylorDev: Fixed in `3f4385f527`. `connectRealtimeGateway` no longer flattens the join rejection — it extracts the server reason and, for a setup-required reason, rejects with `RealtimeGatewaySetupRequiredError` (`code === "SETUP_REQUIRED"`, raw reason preserved); the manager's existing `isSetupRequired` maps it to `setup_required` (not `error`), so `ready()` resolves. Translated reasons: `channel_declaration_unavailable` (the canonical gateway signal per `sdk_channel.ex`) + defensively `adapter_setup_required`/`not_configured` — flagged in a code comment for you/the gateway owner to confirm the exact set. Covered on the REAL path: a manager test drives `defaultActivateChannel` → real launcher → real `connectRealtimeGateway` (only the WebSocket faked) and asserts `setup_required` + `ready()` resolving.

packages/runtime/src/v2/runtime/core/channel-manager.ts
BenTaylorDev: Fixed in `3f4385f527`. `ConnectedRealtimeGatewaySession` now exposes `onStateChange` (`online | reconnecting | gave_up`): an unexpected socket drop → `reconnecting`, a successful (re)join → `online` (driven off the join-push `ok` recHook, which re-fires on Phoenix auto-rejoin), and a bounded give-up window (`reconnectGiveUpMs`, 60s default — covers gateway-down and creds-revoked) → `gave_up`. The manager wires these to `status()`: `reconnecting`→`reconnecting`, `online`→`online`, `gave_up`→`error`; `computeOverall` ranks `error > reconnecting > setup_required > connecting > online`. So `online` now means the basic managed path can currently send. `ready()` keeps one-shot semantics (settles on the initial outcome; later health transitions move only `status()`) — documented. Tests: drop→reconnecting, rejoin→online, deterministic give-up→error.

packages/runtime/src/v2/runtime/core/fetch-handler.ts
BenTaylorDev: Fixed in `3da3e772dc`. The documented snippet now type-checks verbatim (no `!`/`?.`).

Approach: a class constructor can't vary its return type across overloads (it's pinned to the instance type), so branding the class is out. Instead `CopilotRuntime` is a `const` typed as a `CopilotRuntimeConstructor` interface with two `new` overloads — `intelligence` + a **non-empty `channels` tuple** (`readonly [Channel, ...Channel[]]`) returns a phantom-branded `RuntimeWithDeclaredChannels`; everything else returns plain `CopilotRuntime`. `createCopilotRuntimeHandler` then overloads on that brand (with `activateChannels` constrained to `true | undefined`) → non-optional `channels`. The name `CopilotRuntime` is preserved as a type for the ~15 existing annotation/cast sites; runtime behavior is byte-for-byte identical (internal shim class).

Honesty of the narrowing: SSE runtimes, intelligence-without-channels, `channels: []`, a non-literal `Channel[]` variable, and `activateChannels: false` all fall through to the optional overload — nothing gets a dishonest non-optional `.channels`.

Compile test: `handler-channels-types.test.ts`, checked by `tsc --noEmit`. Confirmed failing on the parent commit (`TS2344` on the required-`.channels` assertion) and passing after. Also dropped the `!` at 5 sites in `handler-channels.test.ts` so the suite exercises the documented shape. Full-package `tsc --noEmit` clean; 34 tests green.

One note for the record: this package's pre-commit hook doesn't run `tsc` (vitest transpiles without typechecking), so CI `check-types` is the guard that catches this class — the committed state passes it independently.

packages/channels-intelligence/src/realtime-gateway.ts
```diff
@@ -33,11 +33,107 @@ export interface ConnectRealtimeGatewayOptions {
...
+ * Gateway join-error reasons that mean "the declared channel/provider is not
+ * configured server-side" rather than a hard failure. The canonical signal is
+ * `channel_declaration_unavailable` (Intelligence gateway `sdk_channel.ex`
+ * `validate_heartbeat_response` → `{:error, %{"reason" =>
+ * "channel_declaration_unavailable", ...}}`). `adapter_setup_required` and
+ * `not_configured` are modeled defensively (app-api uses those shapes); the
+ * exact string set should be coordinated with the gateway owner.
+ */
+const SETUP_REQUIRED_JOIN_REASONS: ReadonlySet<string> = new Set([
+  "channel_declaration_unavailable",
```
tylerslaton: [P1] Preserve the per-Channel rejection state before calling this setup-required

`channel_declaration_unavailable` is not specific to missing setup. Intelligence's gateway emits that top-level reason whenever any heartbeat declaration is non-live, while the attached `channels[]` payload distinguishes `runtime_not_declared` from `runtime_conflict`. This mapping therefore turns a duplicate live listener into `setup_required`, so `handler.channels.ready()` resolves instead of surfacing the conflict. Please inspect/preserve the declaration states and map only the genuinely unconfigured states to `SETUP_REQUIRED`; conflicts and other hard failures need to remain errors.

packages/channels-intelligence/src/realtime-gateway.ts
```diff
@@ -83,26 +180,159 @@ export async function connectRealtimeGateway(
...
+  };
+  const enterReconnecting = (): void => {
+    if (closingIntentionally || connectionState === "gave_up") return;
+    // Arm the give-up window on the FIRST drop of an outage episode; a later
+    // drop while still reconnecting keeps the original deadline so a flapping
+    // connection that never stabilizes still gives up after one bounded window.
+    if (giveUpTimer === undefined) {
+      giveUpTimer = setTimeout(() => {
+        giveUpTimer = undefined;
+        emitState("gave_up");
```
tylerslaton, line 313: [P1] Make `gave_up` agree with the transport's real behavior

This timer only emits `gave_up`; it never disconnects the Phoenix socket or stops its automatic retries. At the same time, `emitState` permanently ignores every transition after `gave_up`. If the gateway returns after 60s, Phoenix can successfully rejoin and resume processing deliveries while `handler.channels.status()` remains `error` forever. Either actually terminate the transport when this is terminal, or allow a later successful rejoin to restore `online` (and name/model the state accordingly).

packages/channels-intelligence/src/realtime-gateway.ts
```diff
@@ -83,26 +180,159 @@ export async function connectRealtimeGateway(
...
+  // the socket merely reopening — the channel may still be rejoining.
+  socket.onClose(() => {
+    notifyClose();
+    enterReconnecting();
+  });
+  socket.onError(() => {
+    notifyClose();
+    enterReconnecting();
+  });
+  channel.onClose(() => notifyClose());
```
tylerslaton: [P1] Mark channel-level failures as non-sendable too

Only socket-level close/error calls `enterReconnecting()`. Phoenix can emit a Channel close/error and begin a channel rejoin while the underlying socket remains open; these handlers then fire the breadcrumb but leave the public state `online`, even though channel pushes cannot currently send. Route channel close/error through the same health transition (deduping notifications as needed) so `online` keeps the contract established by this PR.

packages/runtime/src/v2/runtime/core/channel-manager.ts
```diff
@@ -0,0 +1,679 @@
...
+    const channels: Record<string, ChannelStatus> = {};
+    for (const [name, entry] of this.entries) {
+      channels[name] = entry.status;
+    }
+    return { overall: this.computeOverall(Object.values(channels)), channels };
+  }
+
+  /** Fold per-Channel statuses into a single overall status (see {@link status}). */
+  private computeOverall(values: ChannelStatus[]): ChannelStatus {
+    if (values.length === 0) {
```
tylerslaton, line 610: [P1] Do not report skipped declared Channels as healthy

Because direct-adapter Channels never get an entry, a runtime whose only declared Channel has a direct adapter returns `{ overall: "online", channels: {} }`, and `ready()` resolves—even though this handler neither starts nor stops that Channel. This is also whole-Channel exclusivity, not the advertised per-platform rule: direct Slack + managed Teams is skipped wholesale. The duplicate-delivery concern in the existing thread is valid, but if coexistence stays deferred this path must at least fail loud or expose an explicit unsupported/non-managed state; silently healthy is not a truthful lifecycle contract.

packages/runtime/src/v2/runtime/core/fetch-handler.ts
```diff
@@ -300,6 +430,24 @@ export function createCopilotRuntimeHandler(
...
+
+  // Managed Channel activation happens here — at handler-creation time, not in
+  // the Runtime constructor and not inside the per-request closure above (the
+  // first HTTP request must not trigger activation). Only for an Intelligence
+  // runtime that declares Channels and hasn't opted out via activateChannels.
+  if (
+    isIntelligenceRuntime(runtime) &&
+    runtime.channels &&
+    runtime.channels.length > 0 &&
+    options.activateChannels !== false
```
tylerslaton, line 476: [P1] Do not auto-open a process-scoped socket on every Fetch host

This factory is explicitly documented for Cloudflare Workers and Next.js App Router, but this branch opens a long-lived WebSocket at handler creation. Module-scope network I/O is not a valid lifecycle on Workers, and request-driven/serverless isolates can freeze or recycle between messages; separate cold starts also mint separate listener IDs and can conflict. The `WeakMap` only deduplicates one in-memory runtime object. Managed Channels need an explicit long-running-host lifecycle contract here (or a request-safe transport); the generic Fetch path cannot silently default to a persistent listener on hosts that cannot own one.

packages/runtime/src/v2/runtime/core/channel-activation-config.ts
```diff
@@ -0,0 +1,144 @@
...
+  // Return the TRIMMED value so a padded adapter like `"  teams  "` resolves
+  // to `"teams"` rather than being forwarded with its whitespace intact.
+  const trimmedAdapter = adapter?.trim();
+
+  return {
+    wsUrl,
+    apiKey,
+    projectId,
+    channelName,
+    adapter: trimmedAdapter ? trimmedAdapter : "slack",
```
tylerslaton: [P2] Do not describe the managed provider as derived while hard-coding Slack

No public handler/runtime option supplies `adapter`, so the production manager always reaches this fallback and declares every managed Channel as Slack. That means a Teams-backed Channel configured in Intelligence cannot activate, and multiple Channels in one runtime cannot target different managed providers. If next week's gate is intentionally Slack-only, scope the public contract/docs honestly and track the server-resolved/per-Channel provider negotiation; otherwise the provider needs to come from the Channel's Intelligence configuration rather than this global default.

packages/channels/README.md
```diff
@@ -22,13 +22,13 @@ pnpm add @copilotkit/channels-slack
 ## Quickstart
 
 ```ts
-import { createBot } from "@copilotkit/channels";
+import { createChannel } from "@copilotkit/channels";
 import { slack } from "@copilotkit/channels-slack"; // a concrete PlatformAdapter
 
-const bot = createBot({
+const bot = createChannel({
```
tylerslaton: [P2] Finish the public Channels-only vocabulary in the quickstart

The API name changed, but the primary example still teaches `const bot` / `bot.onMention()` / `bot.start()`, and the same application-level noun remains throughout the example docs. Provider terms such as Slack `botToken` are real platform vocabulary; the CopilotKit abstraction is not. Since this PR promises a clean break and the product decision is that Channels is the only public noun, use `channel` consistently in user-facing code and prose.

tylerslaton (changes requested):
The follow-up commits materially improved this: the real setup signal now reaches the manager, connection state is observable, the documented base-handler snippet type-checks, and the telemetry contract is Channels-only. I still see several correctness gaps before this is a safe handler-owned lifecycle: a declaration conflict is currently downgraded to setup-required, the terminal reconnect state can disagree with a live socket, channel-only failures remain falsely online, skipped direct Channels report healthy without being started, and the generic Fetch handler auto-opens a process-scoped socket on request-driven hosts. I also left two P2 contract notes for provider selection and the remaining public Bot terminology.

packages/channels-intelligence/src/realtime-gateway.ts
BenTaylorDev: Fixed in `572f0e6f48`. Replaced the top-level-reason set with a per-Channel classifier: on `channel_declaration_unavailable` it inspects `resp.channels[].state` (verified against `sdk_channel.ex` `validate_heartbeat_response` + `CHANNEL_STATE_KINDS`). Only genuinely unconfigured/waiting states map to `SETUP_REQUIRED` (`no_channels_yet`, `adapter_setup_required`, `slack/channel_setup_complete_waiting_for_runtime`, `disabled_by_entitlement`, `disabled_by_feature_flag`); `runtime_conflict`, `platform_setup_failed`, `delivery_failed`, `egress_failed`, `runtime_offline` are hard errors (new `RealtimeGatewayChannelStateError`, no `code` → `ready()` rejects). Fails loud on the worst state when a batch is mixed; offending states preserved on the error. `runtime_not_declared` I classified as a hard error (during our own join the runtime *is* declaring itself), with a `TODO(gateway-owner)` to confirm it is never a benign join-vs-heartbeat race — please confirm.

packages/channels-intelligence/src/realtime-gateway.ts
BenTaylorDev: Fixed in `572f0e6f48`. Dropped the "ignore all transitions after gave_up" latch from both `emitState` and `enterReconnecting`, so `gave_up` is no longer terminal: it means "still down past the 60s window" (manager still maps → `error`, truthfully non-sendable), but a later successful rejoin restores `online` and a fresh drop re-arms the window. The socket is not force-disconnected (recoverable path — a transient gateway outage self-heals without a process restart). Trade-off to flag: against a persistently-dead gateway the state can oscillate `error ↔ reconnecting` at the backoff cadence; each instantaneous reading is truthful to the transport. If you would rather `gave_up` be sticky-until-online, that is a one-line change (keep the `enterReconnecting` guard, drop only the `emitState` one) — say the word.

packages/channels-intelligence/src/realtime-gateway.ts
BenTaylorDev: Fixed in `572f0e6f48`. `channel.onClose`/`channel.onError` now go through `enterReconnecting()` (previously only `notifyClose()`), so a Phoenix channel-level drop with the socket still open moves public state to `reconnecting`; a channel rejoin → `online` via the join-push `ok` recHook. Dedupe is natural: `emitState` same-state + the single-arm give-up timer means a socket drop that co-fires socket+channel hooks is still one transition. Test injects a `phx_error` for the joined topic without closing the socket and asserts `reconnecting` → rejoin → `online`.

packages/runtime/src/v2/runtime/core/channel-manager.ts
BenTaylorDev: Fixed in `a3cc95330a`. A skipped direct-adapter Channel now gets an explicit entry with a new `unmanaged` status instead of no entry, so it is always keyed in `status().channels` and can never vanish into a false `online`. `computeOverall` folds `unmanaged` out and ranks the remaining managed statuses; a runtime whose only Channel is direct now reads `overall: "unmanaged"` (never `online`), and a managed Channel beside an unmanaged one keeps its real health while the unmanaged one stays visible per-Channel. `stop()` leaves `unmanaged` untouched (never claimed to own it); `ready()` still resolves but its doc now states resolution implies no health. Actually having the handler start/own direct Channels (either-or per Channel, replacing the skip) is **OSS-486**; same-Channel coexistence is **OSS-484**.

packages/runtime/src/v2/runtime/core/fetch-handler.ts
BenTaylorDev: Fixed in `1465404c18`. Activation is now lazy everywhere: `getOrCreateChannelManager` only constructs + caches the manager — the `manager.activate()` call at handler creation is gone. The persistent socket opens on the first `await handler.channels.ready()` and never before, on any host (including the node/express/hono wrappers, which make no `activate`/`ready` call of their own). So a Fetch host that never calls `ready()` (Workers, Next.js App Router request path) never opens a listener. `handler.channels` still exists at creation (the non-optional typing holds); `status()` before `ready()` reports `connecting` (never `online`); a duplicate/missing-name misconfig now surfaces as a rejected `ready()` (fail-loud preserved), and a cached un-activated manager re-rejects. Documented the long-running-host lifecycle contract in the `createCopilotRuntimeHandler` TSDoc (call `ready()` once at startup; do not call per-request on serverless).

packages/runtime/src/v2/runtime/core/channel-activation-config.ts
BenTaylorDev: Fixed in `7633973d50`. Added a per-Channel `provider: "slack" | "teams"` (exported `ChannelProvider`) to `createChannel`/`Channel`; `deriveChannelActivationConfig` reads `channel.provider` per-Channel instead of the manager-wide field (removed — production never set it). Unset defaults to `slack` as an explicit, documented default, and the comment now states the real contract (the SDK declares the per-Channel provider on join; the gateway resolves the connection) rather than "derived." Two Channels can now target different providers. Cross-repo dependency to flag: the gateway (`sdk_channel.ex`) validates `adapter("slack")` / else `invalid_adapter` today, so declaring `provider: "teams"` is SDK-ready but blocked end-to-end until Alems gateway work (OSS-450 / Intelligence #511) accepts `teams`. One API-shape question: I kept `provider` a closed union `"slack" | "teams"` per the finding — if more managed providers are near, we could widen to `string`. Preference?

packages/channels/README.md
BenTaylorDev: Fixed in `357b7fba47`. The quickstart and prose now use `channel` (`const channel`, `channel.onMention()`, `channel.onMessage()`, `await channel.start()`, and the "channel-level defaults"/"the channel"/"the channel's declared commands" prose). Kept `slack({ botToken, appToken })` — those are real Slack platform params, not the CopilotKit abstraction. The remaining internal-only `bot` identifiers (the local `bot` variable and the `adapter.start({ botName })` SPI) are tracked as a mechanical follow-up in **OSS-485**, out of scope for the public-surface rename.

tylerslaton (dismissed):
Three final changes before merge:

1. After the reconnect window emits `gave_up`, keep the public Channel status at `error` while Phoenix continues retrying. Only a successful rejoin may restore `online`; failed retries during the same outage must not emit `reconnecting` again. Please add a regression test proving: long outage becomes and stays `error`; a later successful rejoin returns to `online`; and a new drop after recovery can enter `reconnecting` again.

2. This implementation is exclusive per **Channel**, not per platform: any direct adapter causes the whole Channel to be reported `unmanaged` and skipped for managed activation. Please correct the comments, diagnostics, and PR description accordingly, and remove any claim that direct + managed coexistence or additive behavior already works on the same Channel. OSS-484 and OSS-486 remain deferred; do not implement them here.

3. Keep the managed-provider type as a closed union and name it `ManagedChannelProvider`. The union should reflect real or coordinated gateway support, and Teams should be described as gated/coordinated rather than generally available until its corresponding path is ready.

Once these are fixed, please merge this PR behind the existing feature gate. Production rollout is a separate decision and still requires a packaged SDK → real gateway → Slack smoke test; that path has not been validated by this PR.

BenTaylorDev:
All three landed:

1. **`gave_up` is now sticky** (`64a3d9e2b4`). A `gaveUp` latch suppresses further `reconnecting` emissions during the same outage — the status stays `error` while Phoenix keeps retrying; only a successful rejoin clears the latch and restores `online`, which re-arms it so a genuinely new drop can enter `reconnecting` again. Regression test proves all three legs: long outage → `gave_up` and stays (a failed retry emits no new `reconnecting`); later rejoin → `online`; fresh drop after recovery → `reconnecting`. (Confirmed the "no second reconnecting" assertion fails against the prior recoverable code.)

2. **Corrected to exclusive per Channel** (`1d2671f6c2`). The `activate()` skip comment, the `unmanaged` status doc, and the skip info-log now state it's per **Channel** — any direct adapter makes the whole Channel `unmanaged` regardless of platform — and the "additive"/"coexistence works" claims are removed (OSS-484 and OSS-486 referenced as deferred, not implemented here). PR description updated to match.

3. **`ChannelProvider` → `ManagedChannelProvider`** (`9e94a8ac53`), kept a closed `"slack" | "teams"` union. TSDoc now describes `slack` as GA and `teams` as gated/coordinated (gateway accepts only `slack` at join today → SDK-ready, not GA until Intelligence OSS-450/#511). `grep` confirms zero remaining `ChannelProvider` references.

Green across the three packages (channels-intelligence 141, channels 155, runtime 1742); CI running on the pushed HEAD. Agreed the packaged-SDK → real gateway → Slack smoke is a separate pre-rollout gate, not validated here. Ready to merge behind the existing feature gate on your go.

AlemTuzlak (approved):

PR #5969, opened 2026-07-14
fix(channels-intelligence): Teams thread history (getHistory + getMessages)

AlemTuzlak:
## Problem
On the managed Teams path the agent always reported "no earlier messages" / "I don't see the image in this thread." `HttpDeliverySource.getHistory` was **Slack-only**: it keyed off `threadTs` and returned `[]` for any route without one. A Teams route is `{ adapter: 'teams', tenantId, conversationId }` (no `threadTs`), so it short-circuited *before making any request* — starving **both** history mechanisms on Teams:
- `agent.messages` seeding (`conversationStore.getOrCreate`), and
- the `read_thread` tool (via `thread.getMessages()`).

## Fix
- **`getHistory` is now adapter-aware** (mirrors `conversationKeyFromReplyTarget`'s per-adapter switch): Slack keys off `teamId`/`channel`/`threadTs`; Teams sends `adapter=teams` + `tenantId` + `conversationId`, matching app-api's `teams:{tenantId}:{conversationId}` thread_key. app-api's `/api/channels/history` route already accepts this shape. **Slack query and order are unchanged.**
- **Add `getMessages` to the adapter** so `thread.getMessages()` (the `read_thread` tool) reads reconstructed history via the transport and maps it to `ThreadMessage[]`. Without it `Thread.getMessages()` returns `[]` and thread-reading tools (summaries, "what was in the image") see nothing even when history exists.

## Tests
- New: Teams-shaped `getHistory` query, and the missing-`tenantId`/`conversationId` short-circuit (no request).
- All 115 `channels-intelligence` tests pass; `check-types` clean.

## Notes
Verified end-to-end against a live managed Teams bot as a dist patch before porting to source (history seeding + `read_thread` both start returning the thread's messages). The app-api counterpart (Teams ingress: reactions, inline-media/Graph file ingest, slash commands) is a separate PR in the Intelligence repo.

### DIFF
--- a/packages/channels-intelligence/src/http-transports.test.ts
+++ b/packages/channels-intelligence/src/http-transports.test.ts
@@ -592,6 +592,37 @@ describe("HttpDeliverySource.getHistory", () => {
     expect(calls).toHaveLength(1);
   });
 
+  it("sends the Teams-shaped query (adapter/tenantId/conversationId) for a teams route", async () => {
+    const { calls } = stubGlobalFetch((url) => {
+      expect(url).toBe(
+        "http://x/api/channels/history?adapter=teams&tenantId=tenant-1&conversationId=19%3Ac%40thread.tacv2%3Bmessageid%3D1&limit=5",
+      );
+      return { json: { messages: [{ id: "m1", role: "user", text: "hi" }] } };
+    });
+    const src = new HttpDeliverySource(cfg({}));
+    const history = await src.getHistory(
+      {
+        adapter: "teams",
+        tenantId: "tenant-1",
+        conversationId: "19:c@thread.tacv2;messageid=1",
+      } as unknown as Parameters<typeof src.getHistory>[0],
+      5,
+    );
+    expect(history).toEqual([{ id: "m1", role: "user", content: "hi" }]);
+    expect(calls).toHaveLength(1);
+  });
+
+  it("returns [] for a teams route missing tenantId/conversationId (no request)", async () => {
+    const { calls } = stubGlobalFetch(() => ({ json: { messages: [] } }));
+    const src = new HttpDeliverySource(cfg({}));
+    const history = await src.getHistory(
+      { adapter: "teams" } as unknown as Parameters<typeof src.getHistory>[0],
+      5,
+    );
+    expect(history).toEqual([]);
+    expect(calls).toHaveLength(0);
+  });
+
   it("hydrates a historical file ref into content parts (text inline + image data part)", async () => {
     const png = new Uint8Array([1, 2, 3, 4]);
     stubGlobalFetch((url) => {
--- a/packages/channels-intelligence/src/http-transports.ts
+++ b/packages/channels-intelligence/src/http-transports.ts
@@ -627,24 +627,46 @@ export class HttpDeliverySource implements DeliverySource {
     replyTarget: EgressRoute,
     limit: number,
   ): Promise<AgentMessage[]> {
-    // Slack-shaped for the V1 slice; `EgressRoute` is opaque, so a second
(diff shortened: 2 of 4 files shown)


BenTaylorDev:
Nice, tightly-scoped fix — the `getHistory` adapter switch keeps the Slack query/order byte-for-byte identical and the Teams short-circuit is well tested. One requested change plus a few minor notes:

**Requested change — add a test for the new `IntelligenceAdapter.getMessages`.**
The `getHistory` transport path is well covered, but `getMessages` itself is untested. Its mapping has real branching worth pinning: `content` as `string`, as `AgentContentPart[]` (joined), and the `role → isBot`/`user` derivation. Suggest a test feeding a mixed string/array history and asserting the resulting `ThreadMessage[]`.

**Minor (optional):**
- Silent `catch {}` in `getMessages` degrades to `[]` with no log, unlike `conversationStore.getOrCreate` (`"intelligence getHistory failed; starting fresh"`) and `HttpDeliverySource.getHistory` (loud 4xx). Since `HttpDeliverySource` already swallows its own failures, this outer catch only fires on an unexpected throw — exactly when a `this.opts.config?.log?.` line would help. Worth matching the seeding path.
- `isBot: m.role !== "user"` and `who = m.role === "user" ? "user" : "bot"` encode the same boolean twice — could fold into one `const isBot`.
- `getMessages` flattens content to text only (`ThreadMessage` has no content parts), so `read_thread` is text-only for images — the "what was in the image?" capability rides the seeding path, not this one. Just noting for future readers; no change needed.

AlemTuzlak:
Thanks — all addressed in `3adea157`:

**Requested change — `getMessages` test:** added. Feeds a mixed history and asserts the `ThreadMessage[]` mapping across all the branches you flagged: `content` as a string, `content` as `AgentContentPart[]` (text parts joined; a non-text image part contributes `""`), and the `role → isBot`/`user` derivation. Also added the no-`getHistory`-transport case → `[]`.

**Minor:**
- Added a `this.opts.config?.log?.("intelligence getMessages failed; returning no history", err)` on the outer catch, matching `conversationStore.getOrCreate`'s seeding path. As you noted, `HttpDeliverySource` already swallows its own fetch failures, so this only fires on an unexpected throw — exactly when the line helps.
- Folded the duplicated role boolean: `isBot` computed once, drives both `isBot` and the `user` id/name.

**Text-only flattening:** confirmed intentional — `ThreadMessage` carries no content parts, so `read_thread` is text-only, and the "what was in the image?" capability rides the seeding path (`getHistory` → `buildContentParts` → real image content parts on `agent.messages`), not `getMessages`. Left as-is per your note.

check-types clean, all 117 package tests green.

BenTaylorDev (approved):
Reviewed the diff — clean, well-scoped fix. Making `getHistory` adapter-aware (Slack path byte-for-byte unchanged, Teams as a new guarded branch) and adding `getMessages` so `Thread.getMessages()` stops returning `[]` both address the "no earlier messages" / "no image" symptom at the right layer. `getMessages` mirrors the `conversationStore` seeding read for consistency, correct `URLSearchParams` encoding of the Teams `conversationId`, and solid test coverage (Teams query shape, missing-field short-circuit, string + content-part mapping, no-`getHistory` fallback).

No correctness issues found. The minor polish items (content-part join double-space, `read_thread` JSDoc wording, duplicate per-turn `getHistory` fetch) are non-blocking and tracked in OSS-488, so nothing here should hold the merge. The app-api `/api/channels/history` Teams `thread_key` counterpart lands in the Intelligence repo — flagging only so the two land compatibly (description says it was verified against a live managed Teams bot).

PR #5978, opened 2026-07-15
fix(vue): restore v1 action compatibility

enekesabel:
## What does this PR do?

Restores Vue v1 action and readable compatibility with the current React v1 contract while preserving the existing Vue v2 adapter boundaries.

## Why

The Vue v1 adapters had drifted from React in action classification, live configuration reads, renderer/result normalization, public types, and readable semantics. Valid React-compatible configurations could therefore route differently, retain stale values after dependency-triggered updates, or fail to compile in Vue.

## What changed

- Mirrors React's property-presence routing precedence for catch-all, HITL, frontend, render-only, remote, disabled, handler, and invalid configurations.
- Keeps handler, renderer, metadata, schema, follow-up, and dependency reads live at the adapter boundary; HITL continues through `useHumanInTheLoopV2` so its existing lifecycle and abort behavior remain intact.
- Aligns JSON/VNode/component result normalization and parameterless render schemas with the React v1 behavior.
- Aligns the public v1 types and root exports, including strict discriminated callback renderer unions, Vue component prop declarations, and separate legacy renderer overloads for both `useCopilotAction` and `useFrontendTool`.
- Rejects v1-only `agentId` at inline object boundaries without forwarding it at runtime, while retaining React-compatible structural extra-property behavior. Compatibility-only fields remain accepted but runtime-non-consumed.
- Aligns readable duplicate/update/removal and converter behavior, including React's one-argument runtime invocation of the two-parameter public converter type.
- Adds branch-derived runtime coverage and a Vitest/`vue-tsc` compile-time suite exposed as the Nx `test:types` target.

The only intentional framework boundary is Vue observability: refs, reactive fields, and dependency sources are observed; mutation inside an unreactive plain object is not observed unless a dependency source triggers registration. Function renderers retain React's discriminated union, while Vue component renderers use equivalent flattened prop declarations because Vue declares component props independently.

No React, shared/core, or Vue v2 production code is changed.

## Validation

- `pnpm nx run @copilotkit/vue:test:types --skip-nx-cache` — 11 type tests, no type errors
- Focused v1 runtime suites — 32 tests passed
- `pnpm nx run @copilotkit/vue:check-types --skip-nx-cache`
- `pnpm nx run @copilotkit/vue:build --skip-nx-cache`
- `pnpm nx run @copilotkit/vue:publint --skip-nx-cache`
- `pnpm nx run @copilotkit/vue:attw --skip-nx-cache`
- Changed-file ESLint and `git diff --check`


BenTaylorDev:
Thanks for this, and apologies it sat. The gap it closes is real — I read main's hooks rather than grepping for it: `FrontendAction` in `packages/vue/src/hooks/use-copilot-action.ts` still has no `disabled` or `pairedAction`, and `UseCopilotReadableOptions` in `use-copilot-readable.ts` has no `parentId` or `categories`, while React's v1 has all four. Worth saying explicitly since the timing invites the opposite conclusion: v1 was deprecated on 2026-08-20 but is still supported, and a v1 parity gap is still a live bug, so this work is wanted. Vue's v1 hooks also still live at `packages/vue/src/hooks/`, so nothing here needs re-targeting.

**The rebase is the ask, and it is not mechanical.**

Two files conflict, both for the same reason. Main gained WebMCP support in the v1 hooks on 2026-09-02 in `aeea432cb1`, with the follow-up stability fix `808113b923`. Main's `use-copilot-action.ts` and `use-frontend-tool.ts` each carry seven `webmcp` references; this branch's versions carry none, and the branch rewrites the classifier that webmcp routes through.

So resolving this means re-applying a shipped feature on top of your new routing precedence, not replaying a diff. Could you rebase onto current main and, in the process:

- keep the `webmcp?: boolean | WebMCPToolConfig` option on the v1 action surface and its path through the classifier,
- preserve the stale-rejection and reactive-change stability behavior from `808113b923`,
- and add a case to the new v1 compatibility suite asserting that a `webmcp` action still registers through the v1 path after the rewrite?

That last one matters most. The risk in this rebase is a silent deletion rather than a test failure, and your suite is the natural place to pin it.

**One smaller thing.** The Vue type tests are wired in as a dedicated step in `.github/workflows/test_unit.yml:176`, gated on the affected list containing `@copilotkit/vue`. Please wire `test:types` into the package's own targets instead, so no workflow edit is needed — a package-specific step in the shared workflow will drift the next time that file is restructured. It also has a side effect worth knowing about: by that workflow's own logic, any PR touching `test_unit.yml` forces a full-repo build and test, so this branch's CI runs everything on every push.

Happy to review as soon as it is rebased.

PR #4069, opened 2026-04-18
docs(langgraph): add missing RunnableConfig imports to Python snippets

Koushik-Salammagari:
## Summary

Several Python code examples across LangGraph integration docs used `RunnableConfig` as a type annotation without importing it. Copy-pasting these snippets would raise `NameError: name 'RunnableConfig' is not defined`.

Adds `from langchain_core.runnables import RunnableConfig` to 14 files across:
- `human-in-the-loop/interrupt-flow.mdx`
- `advanced/exit-agent.mdx`
- `advanced/disabling-state-streaming.mdx`
- `agent-app-context.mdx`
- `auth.mdx`
- `configurable.mdx`
- `coagent-troubleshooting/common-coagent-issues.mdx`
- `coagent-troubleshooting/migrate-from-v0.2-to-v0.3.mdx`
- `generative-ui/tool-rendering.mdx`
- `generative-ui/your-components/interrupt-based.mdx`
- `shared-state/in-app-agent-read.mdx`
- `shared-state/predictive-state-updates.mdx`
- `shared-state/state-inputs-outputs.mdx`
- `shared-state/workflow-execution.mdx`

Follows the same pattern as #3755 which fixed the same issue in `langgraph/generative-ui/state-rendering.mdx`.

## Test plan

- [x] Verified all Python blocks using `RunnableConfig` now include the import
- [x] No logic changes — import additions only

PR #3656, opened 2026-04-08
fix(docs): replace dots with hyphens in Anthropic model identifiers

Abubakar-01:
## What
Updated all Anthropic model specifiers in the docs to use hyphens instead
of dots in version numbers.

## Why
Model identifiers with dots (e.g. `anthropic:claude-sonnet-4.5`) are invalid
and will fail at runtime. The correct format uses hyphens
(e.g. `anthropic:claude-sonnet-4-5`), consistent with how Anthropic
formats its model IDs.

## Fix
Replaced dots with hyphens in the version portion of the following specifiers:

| Before | After |
|--------|-------|
| `anthropic:claude-sonnet-4.5` | `anthropic:claude-sonnet-4-5` |
| `anthropic:claude-3.7-sonnet` | `anthropic:claude-3-7-sonnet` |
| `anthropic:claude-opus-4.1`   | `anthropic:claude-opus-4-1`   |
| `anthropic:claude-3.5-haiku`  | `anthropic:claude-3-5-haiku`  |

Models with whole number versions (`claude-sonnet-4`, `claude-opus-4`)
were already correct and left unchanged.

## Verified
All 6 models confirmed as live and valid against Anthropic's official
models overview at platform.claude.com/docs/en/about-claude/models/overview.
The hyphenated format matches Anthropic's official API IDs
(e.g. `claude-sonnet-4-5-20250929`, `claude-opus-4-1-20250805`).

PR #3509, opened 2026-03-24
docs: replace Turborepo instructions with Nx / pnpm

electricalen:
## What does this PR do?

Updates contributor docs (code contributions + package linking) to match the post–#3207 setup: root pnpm scripts and pnpm exec nx instead of global turbo / turbo run …. Covers the root pages, shared snippets, and integration copies of those guides.

## Related PRs and Issues

Related PRs
* #3207

Fixes
* #3508

## Checklist

- [ X] I have read the [Contribution Guide](https://github.com/copilotkit/copilotkit/blob/master/CONTRIBUTING.md)
- [ X] If the PR changes or adds functionality, I have updated the relevant documentation

PR #5982, opened 2026-07-15
docs: refresh stale contributing commands, langgraph RunnableConfig imports, and Anthropic model IDs

BenTaylorDev:
## What & why

Consolidates three community docs fixes that were each opened against the retired `docs/content/docs/` tree (now just a symlink to `showcase/shell-docs/`), so none of them could merge as-is even though the underlying doc bugs are still live. This re-applies them against the current shell-docs source and, where the original proposals had drifted, uses commands verified against the current repo.

### 1. Contributing / package-linking guides — supersedes #3509
Turborepo has been **fully removed** from the repo (no `turbo` dependency, no `turbo.json`), but every contributing guide still instructs `turbo run …`. Across the root guide, all per-integration copies, and the shared snippets:
- Drop the "Turborepo v2.x installed globally" prerequisite; bump pnpm to **v10.x** to match the root `packageManager` (`pnpm@10.33.4`).
- Reframe the monorepo as a pnpm workspace orchestrated by **Nx**.
- Replace commands with equivalents verified against the current root `package.json` / Nx targets:
  - `turbo run build|dev|format|lint` → `pnpm run build|dev|format|lint`
  - `turbo run link:global` / `unlink:global` → `pnpm exec nx run-many -t link:global` / `unlink:global`
  - per-package dev (`turbo run dev --filter=…`) → `pnpm exec nx watch --projects=packages/<name> -- pnpm run build`
  - also fixed a pre-existing stale reference: `pnpm run example-dev` → `pnpm run dev:examples` (the real script)

### 2. Anthropic model IDs — supersedes #3656
`built-in-agent/model-selection` listed dotted IDs that aren't valid; hyphenate `claude-3-7-sonnet`, `claude-opus-4-1`, `claude-3-5-haiku`.

### 3. LangGraph `RunnableConfig` imports — supersedes #4069
Twelve langgraph reference pages annotate `config: RunnableConfig` in Python snippets without importing it. Add `from langchain_core.runnables import RunnableConfig` to each such block. (Tutorial pages were intentionally left out to avoid disrupting their step-by-step narrative.)

## Credit
Thanks to the original authors whose fixes this incorporates: @electricalen (#3509), @Abubakar-01 (#3656), and @Koushik-Salammagari (#4069). Those PRs can be closed in favor of this one once merged.

## Testing
- `grep` confirms **0** remaining `turbo run` / `Turborepo` / `turbo@2` references and **0** remaining `example-dev` references across the docs content.
- All replacement scripts/targets verified present in the root `package.json` (`build`, `dev`, `format`, `lint`, `dev:examples`) and as per-package Nx targets (`link:global`, `unlink:global`).
- Confirmed **0** langgraph Python blocks remain that use `RunnableConfig` without importing it (excluding the intentionally-skipped tutorial pages).
- Spot-checked insertion placement/indentation in both deeply-nested (`interrupt-flow.mdx`) and flat (`auth.mdx`) code blocks.

### DIFF
--- a/showcase/shell-docs/src/content/docs/(other)/contributing/code-contributions/index.mdx
+++ b/showcase/shell-docs/src/content/docs/(other)/contributing/code-contributions/index.mdx
@@ -9,8 +9,7 @@ This guide will help you get started as smoothly as possible.
 ## Step 1: Install Prerequisites
 
 - [Node.js](https://nodejs.org/en/) 20.x or later
-- [pnpm](https://pnpm.io/) v9.x installed globally (`npm i -g pnpm@^9`)
-- [Turborepo v2.x](https://turborepo.org/) installed globally (`npm i -g turbo@2`)
+- [pnpm](https://pnpm.io/) v10.x installed globally (`npm i -g pnpm@^10`)
 
 ## Step 2: Repository Setup
 
@@ -30,8 +29,7 @@ cd CopilotKit
   <Step>
   ### Install Dependencies
   <Callout type="info">
-    The CopilotKit repository is a monorepo based on [Turborepo](https://turborepo.org/).
-    We use [pnpm](https://pnpm.io/) as our package manager.
+    The CopilotKit repository is a monorepo using a [pnpm](https://pnpm.io/) workspace. Tasks across packages are orchestrated with [Nx](https://nx.dev/).
   </Callout>
 
 Install the dependencies using pnpm:
@@ -45,7 +43,7 @@ pnpm install
   ### Build Packages
   To make sure everything works, let's build all packages once:
   ```bash
-  turbo run build
+  pnpm run build
   ```
   </Step>
 </Steps>
@@ -56,10 +54,10 @@ Now that everything is set up and works as expected, you can get started develop
 
 ```bash
 # Start all packages in development mode
-turbo run dev
+pnpm run dev
 
 # Start a specific package in development mode
-turbo run dev --filter="@copilotkit/package-name"
+pnpm exec nx watch --projects=packages/package-name -- pnpm run build
 ```
 
 Now you can start making changes to the code.
@@ -82,12 +80,12 @@ In a separate terminal, run the following command to start the example:
 ```bash
 cd examples/next-openai
 export OPENAI_API_KEY=<your-openai-api-key>
-pnpm run example-dev
+pnpm run dev:examples
 ```
 
 <Callout type="info">
-  We use the `pnpm run example-dev` command to run examples, which is different
-  from the `turbo run dev` command we use to work on the individual packages.
+  We use the `pnpm run dev:examples` command to run examples, which is different
+  from the `pnpm run dev` command we use to work on the individual packages.
 </Callout>
 
 Now navigate to http://localhost:3000/presentation and you should see the example running. Any changes you make to the CopilotKit packages will immediately be reflected here.
@@ -97,13 +95,13 @@ Now navigate to http://localhost:3000/presentation and you should see the exampl
 Before committing your changes, ensure your files are formatted properly by running the following at the root of the monorepo:
 
 ```bash
-turbo run format
+pnpm run format
 ```
 
 Additionally, ensure you have no linting errors:
 
 ```bash
-turbo run lint
+pnpm run lint
 ```
 
 ## Step 6: Submit a Pull Request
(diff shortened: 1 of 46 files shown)


showcase/shell-docs/src/content/docs/(other)/contributing/code-contributions/index.mdx
```diff
@@ -82,12 +80,12 @@ In a separate terminal, run the following command to start the example:
 ```bash
 cd examples/next-openai
 export OPENAI_API_KEY=<your-openai-api-key>
-pnpm run example-dev
+pnpm run dev:examples
```
samjulien: This command cannot start the documented app. The preceding path `examples/next-openai` does not exist; the actual legacy package is `examples/v1/next-openai`, and it exposes `example-dev`, not `dev:examples`. The workspace-level `dev:examples` script only builds/watches examples. Please correct this in the root page and the authored copies.

showcase/shell-docs/src/content/docs/integrations/langgraph/auth.mdx
```diff
@@ -74,6 +74,8 @@ async def authenticate(authorization: str | None):
 ### Access User in Agent
 
 ```python
+from langchain_core.runnables import RunnableConfig
```
samjulien, line 77: All LangGraph variants use `docs_mode: generated`, so the route resolver prefers root `src/content/docs/auth.mdx` and bypasses this framework copy when the root page exists. The live generated auth page therefore still has two `RunnableConfig` examples without imports. Please move/apply this fix to the root source.

samjulien (changes requested):
Thanks for consolidating these fixes. I traced the shell-docs resolution paths across both `docs_mode: generated` and `docs_mode: authored` and found a few correctness gaps that need addressing before merge:

1. The replacement example command does not start the documented app. The guide changes into `examples/next-openai` (which does not exist), then runs `dev:examples` (which is not defined by the actual `examples/v1/next-openai` package). The workspace-level `dev:examples` script only builds/watches examples; it does not start the Presentation server. Please use the real package path and dev script, or point the guide at the maintained presentation showcase. This needs to remain consistent across the root page and authored integration copies.

2. The LangGraph auth imports were added to the framework-owned file, but all LangGraph variants are showcase-driven. Generated routes prefer the root `src/content/docs/auth.mdx`, so the live auth page still contains two Python blocks that use `RunnableConfig` without importing it. Please apply the fix to the root source that those routes render.

3. The non-tutorial `integrations/langgraph/shared-state/predictive-state-updates.mdx` page still has a Python fence whose `chat_node(..., config: RunnableConfig)` lacks the import. A later fence imports the type, but each code fence needs to be self-contained; this also means the PR description's zero-missing-import claim is not yet accurate.

4. If the intent is to refresh contributor prerequisites comprehensively, the root and per-integration `docs-contributions.mdx` pages still require pnpm 9, leaving generated and authored Contributing sections with conflicting pnpm requirements.

The model-ID corrections and the overall root/authored duplication strategy otherwise look good, and CI is green.

BenTaylorDev:
Pushed f126acfb43 addressing all four.

**1. Example command** — Confirmed `examples/next-openai` doesn't exist and `dev:examples` is the workspace build/watch script (no server). Repointed to `examples/v1/next-openai` and its per-package `example-dev` (`next dev`), which serves `/presentation`. Applied to the shared snippet + all per-integration `index.mdx` copies.

**2. LangGraph auth** — Verified all three langgraph variants are `docs_mode: generated`, so `/auth` resolves root `docs/auth.mdx`. Added the import to both Python blocks in the root source. (Left the framework-copy edit in place; it's inert under generated resolution but harmless.)

**3. Self-contained fences** — Fixed `predictive-state-updates.mdx`, then swept every non-tutorial langgraph Python fence; found and fixed one more in `snippets/integrations/langgraph/frontend-tools.mdx`. Non-tutorial langgraph blocks missing the import is now 0.

**4. pnpm** — Bumped the `docs-contributions` guides (root + per-integration) from pnpm 9 to pnpm 10.

Out of scope, flagging for visibility: the same missing-`RunnableConfig` pattern exists in untouched, non-langgraph files — `ag-ui/concepts/state.mdx` (authored upstream), `deepagents/shared-state/*`, and root `agent-config.mdx`. Happy to fold those in here or split them into a follow-up, your call.

samjulien (approved):

PR #5983, opened 2026-07-15
feat(channels-intelligence): realtime transport parity — kinds, identity, history/files, delete

BenTaylorDev:
## What & why

The Realtime Gateway transport had drifted from the HTTP polling transport, so managed channels running over the gateway behaved differently from direct. This brings the realtime path to full parity.

Before, the realtime transport:
- built a **text-only** ingress envelope — commands/reactions/interactions were coerced into empty turns;
- keyed conversations **per-turn** (`conversationKey = turn.id`), so threaded follow-ups didn't share agent/session state;
- **dropped the provider actor** — `env.user` was never populated (on *both* transports, in fact);
- implemented **none** of `fetchFile`/`getHistory`/`uploadFile`, so the realtime path silently ran with no history and no file support;
- had **no `delete` render kind**, so `thread.delete` couldn't render over the gateway.

## Changes

- **`claim-mapping.ts` (new, shared):** extract the claim→ingress mapping — `ClaimedDelivery`, `conversationKeyFromReplyTarget`, `mapDeliveryToEnvelope` — into one module both transports import, so they can't drift again (drift *was* the bug). Adds the provider `actor` to the claim turn and maps it to `env.user` — fixing identity on **both** paths.
- **realtime transport:** builds the envelope via the shared mapper (real kind discrimination, thread-stable `conversationKey`, actor→user) instead of the inline text-only build. Fail-closed: an unmodeled reply-target/kind is dropped+logged, not crashed.
- **`intelligence-file-history.ts` (new, shared):** extract `fetchFile`/`getHistory`/`uploadFile` into `IntelligenceFileHistoryClient`. These are **HTTP-only** — the gateway relays the render-event stream but never file bytes or history. The realtime transport gains them when configured with `appApiBaseUrl` + `apiKey` (threaded through the launcher); without those, the methods stay `undefined` and the adapter degrades exactly as before.
- **`delete` render kind:** add `{ kind: "delete"; ref }` to `ChannelRenderEvent` (mirrors the frozen Intelligence contract) and route `thread.delete` through a render frame when a render sink is wired, like post/update/file.

## Tests

- `claim-mapping.test.ts`: actor→user, kind discrimination (command/reaction/interaction/text), thread-stable conversationKey (slack/teams), unmodeled-adapter throw.
- `render-events.test.ts`: realtime non-text kind + identity + thread-stable key; file/history capability toggling on config; a delete-render-frame test.
- Full suite (160) + `check-types` + `build` green.

## Companion

This is the CopilotKit SDK half of the managed-transport parity work; the Intelligence-side half (gateway `render_event/1` validator for file/delete, DB CHECK, Connector Outbox `chat.delete`, actor on the claim) landed separately. Together they close the direct-vs-managed parity matrix.

### DIFF
--- a/packages/channels-intelligence/src/intelligence-adapter.ts
+++ b/packages/channels-intelligence/src/intelligence-adapter.ts
@@ -369,7 +369,17 @@ export class IntelligenceAdapter implements PlatformAdapter {
       turnId: env.turnId,
       deliveryId: env.deliveryId,
     };
-    const user = env.user ? { id: env.user.id } : undefined;
+    // Forward the full provider identity the claim mapper resolved (OSS-476),
+    // not just the id — otherwise `displayName` is plumbed through the mapper
+    // only to be dropped one layer before it reaches handlers.
+    const user = env.user
+      ? {
+          id: env.user.id,
+          ...(env.user.displayName !== undefined
+            ? { displayName: env.user.displayName }
+            : {}),
+        }
+      : undefined;
 
     switch (env.kind) {
       case "turn": {
@@ -632,6 +642,19 @@ export class IntelligenceAdapter implements PlatformAdapter {
     let acc = "";
     for await (const c of chunks) acc += c;
     const target = _target as ChannelReplyTarget;
+    if (acc.length === 0) {
+      // An empty stream (no chunks, or only empty strings) has nothing to post.
+      // Posting textNode("") violates the render contract's min-1-text
+      // constraint (the frame is rejected → the whole turn nacks), and the HTTP
+      // egress fallback guards the same case. Skip the post and return a
+      // synthetic ref keyed to the turn — there is nothing to update/delete.
+      return {
+        id: `${target.turnId}:empty`,
+        __route: target.route,
+        __turnId: target.turnId,
+        __deliveryId: target.deliveryId,
+      };
+    }
     if (this.renderSink) {
       return this.postRenderFrame(target, {
         kind: "post",
--- a/packages/channels-intelligence/src/intelligence-adapter.ts
+++ b/packages/channels-intelligence/src/intelligence-adapter.ts
@@ -373,14 +373,16 @@ export class IntelligenceAdapter implements PlatformAdapter {
       turnId: env.turnId,
       deliveryId: env.deliveryId,
     };
-    // Forward the full provider identity the claim mapper resolved (OSS-476),
-    // not just the id — otherwise `displayName` is plumbed through the mapper
-    // only to be dropped one layer before it reaches handlers.
+    // Forward the provider identity the claim mapper resolved (OSS-476), not
+    // just the id. The wire field is `displayName`; the public `PlatformUser`
+    // exposes it as `name` (the direct Slack adapter populates `name` too), so
+    // map onto `name` — otherwise `message.user.name` is undefined on managed
+    // turns and callers can't read the profile through the typed API.
     const user = env.user
       ? {
           id: env.user.id,
           ...(env.user.displayName !== undefined
-            ? { displayName: env.user.displayName }
+            ? { name: env.user.displayName }
             : {}),
         }
       : undefined;
--- a/packages/channels-intelligence/src/intelligence-adapter.ts
+++ b/packages/channels-intelligence/src/intelligence-adapter.ts
@@ -642,6 +642,16 @@ export class IntelligenceAdapter implements PlatformAdapter {
   }
 
   async delete(ref: MessageRef): Promise<void> {
+    // Realtime path: stream a `delete` render frame so the Connector Outbox
+    // resolves the ref and calls chat.delete (OSS-420) — mirroring post/update.
+    // Fallback (no render sink — in-memory tests): the egress op path.
+    if (this.renderSink) {
+      await this.postRenderFrame(targetFromRef(ref), {
+        kind: "delete",
+        ref: ref.id,
+      });
+      return;
+    }
     await this.emit(targetFromRef(ref), { kind: "delete", ref: ref.id });
   }
 
--- a/packages/channels-intelligence/src/intelligence-adapter.ts
+++ b/packages/channels-intelligence/src/intelligence-adapter.ts
@@ -235,6 +235,10 @@ export class IntelligenceAdapter implements PlatformAdapter {
                     ? part
                     : ((part as { text?: string } | null)?.text ?? ""),
                 )
+                // Drop empty parts (non-text content contributes "") BEFORE
+                // joining, so a `read_thread` transcript isn't corrupted with
+                // doubled/leading/trailing spaces around image/file parts.
+                .filter(Boolean)
                 .join(" ")
             : "";
       const isBot = m.role !== "user";
--- a/packages/channels-intelligence/src/intelligence-adapter.ts
+++ b/packages/channels-intelligence/src/intelligence-adapter.ts
@@ -434,7 +434,8 @@ export class IntelligenceAdapter implements PlatformAdapter {
         //
         // CONTRACT (app-api side): the interaction delivery MUST carry a turnId
         // distinct from the turn that originally posted the card. Egress op ids
-        // are `${turnId}:${seq}` (see mintOp/postRenderFrame) and seq resets to
+        // are `${turnId}:${seq}` (see mintOp; the render path keys on the
+        // 3-part `${turnId}:${slot}:${seq}`) and seq resets to
         // 0 per delivery, so a reused turnId makes the update's op id collide
         // with the original post's — the Connector Outbox dedupes on op id and
         // SILENTLY DROPS the update, so the card never flips. This layer can't
--- a/packages/channels-intelligence/src/realtime-gateway-launcher.ts
+++ b/packages/channels-intelligence/src/realtime-gateway-launcher.ts
@@ -72,6 +72,11 @@ export interface StartChannelsWithGatewaySessionOptions {
   scope: ChannelRealtimeScope;
   /** Stable runtime instance id (`rti_…`), echoed on every envelope. */
   runtimeInstanceId: string;
+  /** Intelligence app-api HTTP base URL — enables file/history parity on the
+   * realtime path (OSS-476), which are HTTP-only. With {@link apiKey}. */
+  appApiBaseUrl?: string;
+  /** Project runtime API key (`cpk-…`) for the app-api file/history calls. */
+  apiKey?: string;
   /** Activation env overrides forwarded to the runtime (so `handle.metadata`
    * matches what the caller declared on join); omitted fields are gathered from
    * the process. `runtimeInstanceId` is excluded — the required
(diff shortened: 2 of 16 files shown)


BenTaylorDev:
## Pre-merge adversarial CR (7-agent, 2 review + 2 confirmation rounds)

Rebased this branch onto `main` after OSS-473/#5972 landed (one constructor conflict, resolved as a union of `deliveryTimeoutMs` + the file/history client), then ran a full adversarial review of the realtime-transport work. **Converged: the delivery lifecycle is validated correct — ack/nack XOR (delete-before-wire), poison-payload dead-lettering, per-turn timeout, `mapDeliveryToEnvelope` parity + exhaustiveness, and `readCapped` byte-cap all confirmed sound by every agent, with no regressions.**

### Fixed in this PR (from the CR)
- **Poison-payload re-lease loop** — an unmappable delivery (unmodeled reply-target adapter / unknown `input.kind`) with a valid lease was logged + dropped, re-leasing the identical payload forever. Now fails **non-retryable** (dead-letter) via `PoisonDeliveryError` → `nack(…, retryable: false)`, mirroring the HTTP path. `nack()` gains a `retryable` param.
- **Double-terminal-signal race** — realtime `ack`/`nack` deleted state *after* the wire push, so a per-turn-timeout `nack` could race a late dispatch `ack` and emit both `fail` + `complete_requested`. Now **delete-before-push** (XOR), matching HTTP.
- **Concurrent dispatch** → **serial** processing (parity with the HTTP runLoop), so an at-least-once redelivery of an in-flight `turnId` can't reset the shared per-turn `seq` counter or run two turns of one conversation at once.
- **Graceful `stop()`** — halts new intake (a `stopped` guard; the session has no `off`) and **drains** the in-flight delivery before clearing state, so a turn settling at stop time still sends its terminal signal instead of silently redelivering.
- **`actor.displayName`** was carried through the mapper then dropped in `dispatchTo` — now forwarded to handlers (the identity half of OSS-476).
- **`fetchFile`** now streams + aborts past `MAX_INBOUND_FILE_BYTES` (the declared `content-length` alone didn't bound a chunked/understated body).
- **`getHistory`** guards `limit <= 0` (was `slice(-0)` → returned everything) and caps to the most recent `limit` **before** hydrating file bytes.
- `stream()` empty-frame guard; HTTP `withTimeout`/`defaultSleep` `unref`; realtime `nack` reason truncation; `getMessages` drops empty parts before `join(" ")` (was corrupting `read_thread` transcripts); corrected `thread_started`/op-id/`ClaimResponse` comments; removed dead imports; genuine test coverage for the `no-getHistory` branch.

New/updated tests: poison non-retryable, single-terminal XOR, serial dispatch, `stop()` drain + post-stop guard, `displayName` forwarding, `getHistory` limit≤0 + cap. **channels-intelligence 170 tests + check-types green; CI green.**

### Deferred (flagged, not guessed in-PR)
- **OSS-491** (delivery-terminal-signal contract — needs app-api coordination): the empty-turn completion signal (a reaction/command that posts nothing has no valid `complete_requested` under the frozen contract → redelivers to `max_attempts`); the per-turn-timeout `retryable` nack that can overlap a still-running `turnId`; `run_error` swallowed on the HTTP-fallback render path; the realtime `push()` no-state fallback vs HTTP's fail-loud (documented as intentional — parity question for you).
- **OSS-497** (pre-existing HTTP-transport robustness cluster): heartbeat starvation during a long turn, `stop()` shutdown latency, `HttpEgressSink` empty-`update`→new-post, static `adapter` on egress/heartbeat, strict-`number` `projectId` coercion, injectable `fetch` not honored by the file/history client, HTTP `push` defaulting a malformed acceptance to success.
- **Durable StateStore** on the managed realtime path (`buildDefaultStore` returns `undefined` when transports are injected) — the item OSS-482 excluded; addressed server-side by OSS-473/OSS-474 (canonical Threads + durable KV). Flagging to confirm SDK-side rehydration is covered.

The reconnect `gave_up → error` escalation and the delivery-contract items above are the only launch-relevant gaps left; the rest are tracked robustness follow-ups.

packages/channels-intelligence/src/realtime-gateway-launcher.ts
```diff
@@ -72,6 +72,11 @@ export interface StartChannelsWithGatewaySessionOptions {
   scope: ChannelRealtimeScope;
   /** Stable runtime instance id (`rti_…`), echoed on every envelope. */
   runtimeInstanceId: string;
+  /** Intelligence app-api HTTP base URL — enables file/history parity on the
+   * realtime path (OSS-476), which are HTTP-only. With {@link apiKey}. */
+  appApiBaseUrl?: string;
```
tylerslaton, line 77: [P1] Wire the app-api URL through the Runtime activation path

The normal managed path never sets this option: `defaultActivateChannel()` only passes `wsUrl`, `apiKey`, `scope`, `runtimeInstanceId`, `adapter`, and `log`, and its structural module type does not expose `appApiBaseUrl`. Because the transport installs `fetchFile`, `getHistory`, and `uploadFile` only when this is set, Channels started by the normal `CopilotRuntime` handler still run without history or any file support. Please carry `intelligence.ɵgetApiUrl()` through `ChannelActivationConfig`/`defaultActivateChannel()` and cover that integration, instead of only supporting callers that invoke this low-level launcher manually.

packages/channels-intelligence/src/realtime-gateway-transport.ts
```diff
@@ -394,14 +549,31 @@ export class RealtimeGatewayTransport
...
     const acceptedThrough = Array.from(state.accepted.entries()).map(
       ([slot, seq]) => ({ turnId: state.turnId, slot, seq }),
     );
+    this.deliveries.delete(deliveryId);
     if (acceptedThrough.length === 0) {
       // Nothing was accepted (e.g. an empty turn). Without an accepted frame
       // there is nothing to complete; let the lease lapse / redeliver instead
       // of sending an invalid (empty acceptedThrough) intent.
-      this.deliveries.delete(deliveryId);
+      // NOTE (OSS-491): a legitimately empty turn (a reaction/command handler
```
tylerslaton, line 566: [P1] Complete successful deliveries that render no frames

A successful command, reaction, or interaction handler may perform work without posting UI, but this branch deletes local state and sends no terminal intent. The lease then expires and the same handler runs again until `max_attempts`, which can duplicate tools or other side effects and ultimately dead-letter successful work. Now that these non-text kinds are routed through realtime, this is a normal reachable path. The managed path needs a valid completed-empty terminal signal (or equivalent finalization) before this can provide transport parity.

packages/channels-intelligence/src/intelligence-adapter.ts
```diff
@@ -369,7 +373,17 @@ export class IntelligenceAdapter implements PlatformAdapter {
...
     };
-    const user = env.user ? { id: env.user.id } : undefined;
+    // Forward the full provider identity the claim mapper resolved (OSS-476),
+    // not just the id — otherwise `displayName` is plumbed through the mapper
+    // only to be dropped one layer before it reaches handlers.
+    const user = env.user
+      ? {
+          id: env.user.id,
+          ...(env.user.displayName !== undefined
+            ? { displayName: env.user.displayName }
```
tylerslaton: [P2] Map the wire display name to `PlatformUser.name`

`IngressSink` exposes this value as `PlatformUser`, whose public field is `name`, not `displayName`; the direct Slack adapter also populates `name`. Passing `displayName` through as an extra property leaves `message.user.name` undefined on managed turns, so callers cannot observe the profile through the typed API. Please map `env.user.displayName` to `name` here and make the test assert `IncomingMessage.user.name` rather than a custom `displayName` shape.

tylerslaton (changes requested):
The transport work is close, but the normal Runtime activation path still does not enable history/files, successful no-output deliveries can be executed repeatedly, and the managed user profile does not match the public Channel type. These affect the behaviors this PR is intended to make equivalent, so I’m requesting changes.

packages/channels-intelligence/src/realtime-gateway-launcher.ts
BenTaylorDev: Fixed in `a8407f84fd`. Threaded the app-api URL through the normal activation path: `ChannelActivationConfig` gains `apiUrl`, set from `intelligence.ɵgetApiUrl()` in `deriveChannelActivationConfig`, and `defaultActivateChannel` now passes it as `startChannelsOverRealtimeGateway({ appApiBaseUrl })`. The launcher + transport already accepted `appApiBaseUrl`/`apiKey`; the only gap was the runtime activation path not forwarding it, so Channels started by the `CopilotRuntime` handler now get `fetchFile`/`getHistory`/`uploadFile`, not just manual low-level callers.

Assumptions I verified against the code before wiring it:
- `ɵgetApiUrl()` returns the app-api **HTTP** base (`#apiUrl`) — distinct from `ɵgetRunnerWsUrl()`, and the same value already used for the agent `/mcp` endpoint — so it's the correct `baseUrl` for the `/api/channels/files|history` routes.
- `apiKey` = `ɵgetRunnerAuthToken()` = the `cpk-` project key, which this PR's transport already uses as the app-api bearer (the HTTP path does the same); I'm just feeding the runtime's existing key into that contract, not introducing a new auth assumption.
- Forwarding is intact at every hop: outer `startChannelsOverRealtimeGateway` (line 77) → `startChannelsWithGatewaySession` (lines 107–113) → the transport constructor's file/history wiring.

Coverage: the config-deriver test asserts `apiUrl === ɵgetApiUrl()`, and the `defaultActivateChannel` test asserts `appApiBaseUrl` reaches the launcher opts. (That opts test executes in CI — the runtime vitest suite hits a local optional-peer-dep import-resolution flake, so I verified the type surface via the runtime build instead.)

packages/channels-intelligence/src/intelligence-adapter.ts
BenTaylorDev: Fixed in `a8407f84fd`. `dispatchTo` now maps `env.user.displayName` → `PlatformUser.name`, so `message.user.name` is populated on managed turns (parity with the direct Slack adapter, which sets `name`).

Verified the field mapping is correct on the real managed path, not just the in-memory test:
- The claim mapper emits `env.user = { id, displayName }` (`claim-mapping.ts:157-160`), and the wire shape is `{ id; displayName? }` (`contracts.ts:44`) — so `env.user.displayName` is the right source field.
- `PlatformUser` is `{ id; name?; handle?; email? }` — `name` is the public field, confirmed.

The wire envelope still carries `displayName`; only the public projection changed. Test now asserts `message.user.name`.

packages/channels-intelligence/src/realtime-gateway-transport.ts
BenTaylorDev: Confirmed reachable, and verified the mechanism: `createRunRenderer` is never called from `create-channel.ts` — the discrete `onCommand`/`onReaction`/`onInteraction`/`onModalSubmit` paths just `await handler(ctx)` with no `finish()`/`finalize`. So a handler that does work but posts nothing produces zero accepted frames, `ack()` has no valid `acceptedThrough`, and the delivery redelivers to `max_attempts`. (Agent runs avoid this because the adapter's run renderer `finish()` always pushes a `finalize` frame → accepted → non-empty `acceptedThrough`.)

The fix is a terminal-signal contract decision I don't want to guess from the SDK side:
1. **SDK** pushes a terminal `finalize` frame for a no-output handler turn (mirroring the agent-run `finish()`), so the existing `complete_requested` path fires — viable entirely in the SDK **iff** the gateway/Connector Outbox accepts a lone `finalize` (no preceding content) as a valid completion.
2. **app-api** accepts a completed-empty terminal signal (`complete_requested` with empty `acceptedThrough`, or a distinct "completed, no output" event).

Can you confirm option 1 is accepted by the frozen contract, or point me at the intended signal for option 2? I'll wire whichever once it's settled. Tracked in OSS-491.

tylerslaton (approved):

Issue #5931, opened 2026-07-13
Best practice: persist & restore MCP Apps history when using self-hosted Agno sessions (not Enterprise Threads)


likebean:
## Context

We're building a self-hosted stack:

- **Agent backend:** Agno AgentOS (AG-UI via `/dynamic_agui/...`)
- **Frontend:** CopilotKit + `@ag-ui/mcp-apps-middleware` / runtime `mcpApps`
- **Session store:** Agno sessions (`GET /sessions/{id}/runs`), **not** CopilotKit Enterprise Intelligence Threads

Live MCP Apps work fine: middleware injects UI tools → Agno pauses with `external_execution` → middleware runs `tools/call` → emits `TOOL_CALL_RESULT` + `ACTIVITY_SNAPSHOT` (`activityType: "mcp-apps"`) → `MCPAppsActivityRenderer` shows the iframe.

## Problem

After reload / switching sessions we hydrate chat via Agno runs → `setMessages(...)`.

What Agno persists for UI tools looks like:

- run status: **paused** (`RunPaused`)
- `tools[].result`: **`null`** (middleware never resumes Agno with trailing `role: "tool"` messages)
- no `ACTIVITY_SNAPSHOT` / `role: "activity"` in the runs payload

So history shows empty tool results and **no MCP App iframe**.

We understand official persistence for MCP Apps is **thread event replay** (Enterprise Threads / persistent `AgentRunner` storing the AG-UI event stream including `ACTIVITY_SNAPSHOT`). We want guidance for the **Agno-session-as-source-of-truth** case.

## What we think the right approach is

1. After middleware executes the MCP UI tool, **resume the Agno run** by sending trailing tool messages (`TOOL_CALL_RESULT` content) so Agno stores a non-null `tools[].result` in the session.
2. Prefer storing the **full MCP result object** (not only `extractTextContent` string), plus enough metadata to rebuild UI (`resourceUri`, stable `serverId` / `serverHash`, `toolInput`).
3. On hydrate (`runs` → messages), **synthesize** an AG-UI activity message:

```ts
{
  role: "activity",
  activityType: "mcp-apps", // MCPAppsActivityType
  content: {
    result,       // full MCP tool result
    resourceUri,
    serverHash,
    serverId,
    toolInput,
  }
}
```

validated against `MCPAppsActivityContentSchema`, then `setMessages`, relying on the built-in `MCPAppsActivityRenderer`.

[...]

Issue #5434, opened 2026-06-13
How should I create a thread on first submit using my own create-thread API?


phongphongg:
Current behavior:

- CopilotKit usually generates a thread ID on the frontend (SDK-minted UUID) and the runtime/platform creates the backend thread when the first message is submitted.

My requirement:

- I own a separate backend API that creates threads (POST /threads → returns { threadId }).
- I want to allow users to type a message with no thread yet. When they press send for the first time, my frontend should:
  1) call my create-thread API to get a real threadId,
  2) use that returned threadId to submit the same message into CopilotKit so the runtime persists under my ID,
  3) reuse that threadId for subsequent messages.

**Question:**

Where is the right integration point to intercept the first submit so I can call my API and then continue the submission with the returned threadId? Specifically:
   - Is there a public “pre-submit” hook or submit interceptor in CopilotChat? 
   - If not, do you recommend using a lower-level API (e.g. calling runtime endpoints directly or using the agent/run APIs) and wiring the response back into the UI?

BenTaylorDev:
The new **Thread & History Lifecycle** guide (merged in #5988) covers this in the first-message thread-creation section — mint the threadId up front and drive creation from there. Note `onSubmitMessage` is headless-only (the built-in `CopilotChat` overrides it), which is the usual snag when wiring your own create-thread API. Closing as answered; reopen if the guide doesn't cover your flow.

Issue #4790, opened 2026-05-13
What JSON Schema Does CopilotChat Require for Restoring Conversation History?


Aimee-min-code:
How should chat history be loaded into CopilotChat?

When initializing the page, what JSON response format should the backend return so that previous messages can be rendered correctly in CopilotChat?

Are there any required fields, internal message types, or special CopilotKit configurations for restoring conversation history (e.g. threadId, event replay, or specific hooks/APIs)?

Aimee-min-code:
In CopilotKit v2, CopilotChat no longer supports the initialMessages prop. What is the recommended approach for initializing chat history or hydrating previous messages in the new architecture?

BenTaylorDev:
This is now covered by the new **Thread & History Lifecycle** guide in the CopilotKit docs (merged in #5988). See the history-hydration section: you restore a conversation by hydrating messages via an explicit-threadId `connectAgent()` replay (with a server-side store) or `agent.setMessages()` — there's no v2 `initialMessages` prop. Closing as answered; please reopen if the guide leaves a gap for your case.

Issue #4778, opened 2026-05-12
Clarification on threadId generation and lifecycle handling


Aimee-min-code:
I would like to understand the generation mechanism of threadId in CopilotKit.

Currently, when sending a CopilotKit request, the request body includes a threadId, which represents a single conversation session.

My questions are:

How is the threadId generated, and at what point in the system lifecycle is it created?
Since threadId represents a conversation context, what is the expected behavior when it is lost due to actions such as:
Page refresh
User logout
Session expiration or similar scenarios

In these cases, should the backend be responsible for generating a new threadId and returning it to the frontend, or is the threadId expected to be persisted and managed entirely on the client side?

I would appreciate clarification on the intended design and best practices for handling threadId lifecycle management in CopilotKit.

BenTaylorDev:
Answered by the new **Thread & History Lifecycle** guide (merged in #5988), which documents threadId handling end-to-end: it's minted client-side (uuid v4, with a remount caveat), and you switch/create threads via `setActiveThreadId(id, { explicit })` / `startNewThread()`. Closing as answered — reopen if anything's unclear after reading it.

PR #5988, opened 2026-07-15
docs: add Thread & History Lifecycle guide

BenTaylorDev:
## What & why

Adds a new framework-agnostic guide — **Thread & History Lifecycle** (`docs/threads-lifecycle.mdx`) — that walks the full client-side lifecycle of a conversation thread as one narrative. It fills the gap between the [Headless Threads](/threads) how-to and the [Threads & Persistence Architecture](/premium/threads-explained) explanation, and directly answers a recurring cluster of support questions.

## Sections
- **How the threadId is created** — UUID v4, client-minted at mount, the resolution precedence, and the remount-stability caveat (auto-minted ids re-mint on remount / StrictMode; pass an explicit `threadId` for continuity).
- **How history is restored** — explicit-`threadId` `connect()` replay vs. manual `agent.setMessages(...)`; clarifies there is **no v2 `initialMessages`** and no v2 `useCopilotChat` (read via `useAgent().agent.messages`).
- **Switching / starting threads** — `setActiveThreadId(id, { explicit })` and `startNewThread()`, plus the prop-controlled no-op guard.
- **Creating a thread with your own API on first message** — mint-up-front + `setActiveThreadId`/`threadId` prop as the robust path; the headless `CopilotChatInput.onSubmitMessage` seam for submit-time interception (noting the built-in `<CopilotChat>` overrides it).
- **CopilotKit threads vs. your framework's checkpointer** — two layers correlated only by `threadId`; a LangGraph checkpointer creates checkpoint tables, not a CopilotKit "threads" table.
- **MCP Apps activity & history** — activity messages are frontend/middleware constructs (no server store); re-synthesize on hydration.
- **v1 vs v2** disambiguation (incl. the two different `useThreads` hooks).

## Addresses
Recurring thread-lifecycle questions: #4790, #4778, #5434, #2242, #5931. (I'll close those pointing here once this lands.)

## Testing / accuracy
All referenced APIs verified present on `main`: `useThreads` (v2), `useCopilotChatConfiguration` (`setActiveThreadId`/`startNewThread`), `useAgent().agent.setMessages/addMessage`, `CopilotChatInput.onSubmitMessage`, `<CopilotChat threadId>`, and the `mcp-apps` activity type. All five cross-doc links resolve. Added to the "Threads" group in `docs/meta.json`.

Note: written against the current v2 APIs — three details were corrected against source during authoring (`startNewThread` not `createThread`; `onSubmitMessage` is headless-only; no v2 `useCopilotChat`/`initialMessages`).


samjulien (changes requested):
Requesting changes before merge: (1) use agentId rather than agent in both CopilotChat snippets; agent is not a v2 prop. (2) Add threads-lifecycle to every authored-framework Threads navigation group: Built-in Agent, Mastra, CrewAI, LlamaIndex, Agno, AG2, Pydantic AI, Microsoft Agent Framework, and Deep Agents. The root meta only covers generated docs. (3) Qualify the lifecycle summary statement that the platform persists runs server-side; replay depends on configuring a server-side store. The remaining v2 lifecycle and API details look good.

BenTaylorDev:
Thanks Sam — all three addressed in `fa319dffe5`:

**1. `agentId` not `agent`** — fixed in both `<CopilotChat>` snippets. (Confirmed `agentId?: string` is the v2 prop in `CopilotChat.tsx`; `agent` was wrong.)

**2. Per-framework Threads nav** — added `threads-lifecycle` to every authored-framework group: Built-in Agent, Mastra, CrewAI, LlamaIndex, Agno, AG2, Pydantic AI, Microsoft Agent Framework, and Deep Agents. Rather than duplicate the page nine times, I followed the existing single-source convention (`threads` / `headless-threads` / `threads-import` are all thin wrappers over `@/snippets/shared/threads/…`): moved the body into a new shared snippet and made the root + per-framework pages thin wrappers importing it.

**3. Persistence caveat** — qualified the "Run" glance bullet: persistence/replay depends on a configured server-side store (Enterprise Intelligence or a persisting `AgentRunner`); a runtime with no persistence layer keeps nothing server-side. (This lines up with the existing replay callout further down.)

Ready for another look 🙏

samjulien (changes requested):
One follow-up remains: the new Threads page changes the expected navigation order, but the docs-render test still expects the old list. Please add Thread and History Lifecycle after Overview in the expected array in showcase/shell-docs/src/lib/__tests__/docs-render.test.ts so the local shell-docs test suite passes. The public image test failures in this checkout are unrelated Git LFS placeholders.

samjulien (dismissed):
Approved. The navigation expectation now covers the new lifecycle page, and the earlier API, persistence, and cross-framework navigation findings are resolved.

samjulien (changes requested):
Requesting two documentation edits:

1. Reorder the Threads navigation consistently across every framework to: Overview, Threads Drawer, Headless Threads, Import Thread History, Threads & Persistence Architecture, then Thread & History Lifecycle. Update the root and authored-framework `meta.json` files, plus the navigation-order test.

2. Keep Thread & History Lifecycle and Threads & Persistence Architecture as two documents, but tighten their boundary instead of merging them. The Lifecycle page should own client-facing React behavior: minting IDs, hydration, switching, and manual history. The Architecture page should own Enterprise Intelligence Platform behavior: persistence, replay, realtime synchronization, locks, and failure modes. Reduce the duplicated “start / resume / switch” walkthrough in Architecture and link to Lifecycle for client behavior; Lifecycle should retain only brief persistence context and link to Architecture for the deeper model.

Please also do a writing pass to reduce the heavy use of em dashes. It is not stylistically consistent with the rest of the docs; prefer shorter sentences, commas, or parentheses where they read naturally, reserving em dashes for the rare case where they add real clarity.

samjulien (approved):

PR #5993, opened 2026-07-16
fix(shell-docs): preserve trusted MDX expressions

zcxGGmu:
## Summary
- allow trusted shell-docs MDX renders to preserve JSX expression props while keeping dangerous-expression filtering enabled
- share the trusted MDX options across docs, framework, reference, AG-UI, setup, and partial render paths
- add a regression test covering IframeSwitcher template-literal props

## Verification
- npm run test -- src/lib/__tests__/trusted-mdx-options.test.tsx
- npm run test -- src/lib/__tests__/docs-render.test.ts
- npm run typecheck

### DIFF
--- a/showcase/shell-docs/src/app/[framework]/[[...slug]]/page.tsx
+++ b/showcase/shell-docs/src/app/[framework]/[[...slug]]/page.tsx
@@ -51,6 +51,7 @@ import {
   parseFrontendRoutePath,
 } from "@/lib/frontend-options";
 import { transformerMeta } from "@/lib/rehype-code-meta";
+import { createTrustedMdxRemoteOptions } from "@/lib/trusted-mdx-options";
 import {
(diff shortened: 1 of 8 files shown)


BenTaylorDev:
Thanks for this — the underlying diagnosis is correct: `next-mdx-remote@6` defaults `blockJS: true`, which strips JSX attribute expressions like `IframeSwitcher`'s template-literal `exampleUrl`/`codeUrl` props.

Closing rather than merging because the fix flips the security default (`blockJS: false`) across every MDX render path at once, widening the JS-execution surface for all rendered MDX and leaning on the best-effort `blockDangerousJS` guard. We'd prefer to address this more deliberately (scope it to the affected render paths / audit the content set / reconsider the dependency behavior) rather than globally. Not a reflection on the analysis, which was sound.

PR #5994, opened 2026-07-16
fix(research-canvas): terminate Python graph

zcxGGmu:
## Summary
- add an explicit chat_node -> END edge to the research-canvas Python LangGraph workflow
- add a stdlib unittest topology regression that checks the termination edge is declared

## Verification
- python3 -m unittest discover examples/v1/research-canvas/agents/python/tests
- python3 -m py_compile examples/v1/research-canvas/agents/python/src/agent.py examples/v1/research-canvas/agents/python/tests/test_graph_topology.py

### DIFF
--- a/examples/v1/research-canvas/agents/python/src/agent.py
+++ b/examples/v1/research-canvas/agents/python/src/agent.py
@@ -5,7 +5,7 @@ It defines the workflow graph and the entry point for the agent.
 
 import os
 
-from langgraph.graph import StateGraph
+from langgraph.graph import END, StateGraph
 
 from src.lib.chat import chat_node
 from src.lib.delete import delete_node, perform_delete_node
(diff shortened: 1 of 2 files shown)


BenTaylorDev:
Closing — this doesn't fix a real defect. `chat_node` already terminates the graph dynamically: it's typed `Command[Literal[…, "__end__"]]` and defaults `goto = "__end__"` when the model responds with no tool call. I verified against `langgraph` 1.x (this example pins `>=1.0.5`) that the graph's behavior is identical with and without the added `add_edge("chat_node", END)` — it's a no-op that also runs counter to the `Command`-based routing this graph uses.

The added test only AST-greps `agent.py` for the literal `add_edge("chat_node", END)` call; it never compiles or runs the graph, so it asserts the diff exists rather than that termination works. If there's an actual hanging run, a reproducing case on `main` would help — happy to reopen with one.

PR #5666, opened 2026-06-24
fix: route frontend tool calls (WriteReport, WriteResearchQuestion) to chat_node in research-canvas TS agent

deepshekhardas:
Fixes #5607

The \
oute()\ function in the research-canvas TypeScript agent only recognized \Search\ and \DeleteResources\ tool calls. Frontend CopilotKit actions (\WriteReport\, \WriteResearchQuestion\, etc.) would fall through to \
eturn END\, silently dropping them.

Now routes any unrecognized tool call back to \chat_node\ so the AG-UI channel can pick up frontend actions.

PR #5995, opened 2026-07-16
fix(research-canvas): route TypeScript tool calls

zcxGGmu:
## Summary
- route TypeScript research-canvas tool calls based on tool_calls instead of AIMessageChunk constructor names
- send non-specialized tool calls back to chat_node instead of ending the graph
- replace the placeholder test script with a Node static regression test for route behavior

## Verification
- npm test

### DIFF
--- a/examples/v1/research-canvas/agents/typescript/package.json
+++ b/examples/v1/research-canvas/agents/typescript/package.json
@@ -7,7 +7,7 @@
   "author": "",
   "main": "index.js",
   "scripts": {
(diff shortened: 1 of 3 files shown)


BenTaylorDev:
Closing as a duplicate — this is the same research-canvas TypeScript tool-routing fix (#5607) already covered by the existing open PRs #5627 and #5666 (identical `else if (toolName) return "chat_node"` change in `examples/v1/research-canvas/agents/typescript/src/agent.ts`). Consolidating on the existing PRs.

PR #6003, opened 2026-07-16
fix(runtime): support appending state deltas to new arrays

### LINKED ISSUE #5998: 🐛 Bug:  OPERATION_PATH_CANNOT_ADD when LLM emits AGUISendStateDelta with /- append on uninitialized array
### ♻️ Reproduction Steps

1. Set up a CopilotKit runtime with any agent that exposes the `AGUISendStateDelta` tool (default behavior)
2. Configure the agent's initial state as `{}` (no `STATE_SNAPSHOT` prior to `STATE_DELTA`)
3. Ask the LLM a question that triggers it to maintain a list in agent state (e.g. "keep a todo list")
4. The LLM calls `AGUISendStateDelta` with something like:
   `{op: "add", path: "/todos/-", value: {completed: false, text: "Buy milk", id: ...}}`
5. The runtime crashes with `OPERATION_PATH_CANNOT_ADD`

### ✅ Expected Behavior

The runtime should gracefully handle the LLM appending to an array that hasn't been explicitly initialized. The state delta should be applied, creating the array if needed, or the error should be caught and the delta handled gracefully.

### ❌ Actual Behavior

fast-json-patch in strict mode throws OPERATION_PATH_CANNOT_ADD because the /- path is only valid when the parent array already exists in the document. The error propagates through compactEvents() in @ag-ui/client (called from the in-memory runner), crashing the run.

### Screenshots

_No response_

### 𝌚 CopilotKit Version

```shell
N/A (present across versions using @ag-ui/client@0.0.57)
```

### 📄 Logs (Optional)

```shell
OPERATION_PATH_CANNOT_ADD: Cannot perform an `add` operation at the desired path
name: OPERATION_PATH_CANNOT_ADD
index: 0
operation: { op: "add", path: "/todos/-", value: { ... } }
tree: {}
    at applyOperation (fast-json-patch)
    at applyPatch (fast-json-patch)
    at $ (compactEvents, @ag-ui/client)
    at runAgent (in-memory runner)
```

zcxGGmu:
Fixes #5998.

## Summary
- Normalize in-memory runner state deltas before compaction when an `add` operation appends to a missing array path.
- Preserve existing non-array values and only insert initializer operations for missing parents.
- Add a regression test for `/todos/-` state delta appends from an empty state.

## Verification
- `pnpm nx run @copilotkit/runtime:test --excludeTaskDependencies --skip-nx-cache -- src/v2/runtime/runner/__tests__/in-memory-runner.test.ts`
- `pnpm nx run @copilotkit/runtime:check-types --excludeTaskDependencies --skip-nx-cache`

### DIFF
--- a/packages/runtime/src/v2/runtime/runner/__tests__/in-memory-runner.test.ts
+++ b/packages/runtime/src/v2/runtime/runner/__tests__/in-memory-runner.test.ts
@@ -15,7 +15,7 @@ import type {
 } from "@ag-ui/client";
 import { AbstractAgent, EventType, verifyEvents } from "@ag-ui/client";
 import { EMPTY, Observable, firstValueFrom } from "rxjs";
-import { toArray } from "rxjs/operators";
+import { timeout, toArray } from "rxjs/operators";
 
 const stripTerminalEvents = (events: BaseEvent[]) =>
   events.filter(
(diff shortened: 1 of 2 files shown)


BenTaylorDev:
Closing to avoid parallel edits on the runtime in-memory state-delta path, which is under active human work (#5847). #5998 stays open for normal review. (Unsolicited machine-generated PR.)

PR #6008, opened 2026-07-16
Fix stale activity content after deltas

### LINKED ISSUE #6001: CopilotChat keeps stale A2UI activity content after ACTIVITY_DELTA from a userAction run
## Description

When an A2UI user action starts a new Runtime run and the backend updates an existing activity message with a valid `ACTIVITY_DELTA`, the full `CopilotChat` component continues rendering the previous activity content.

The Runtime receives the action, updates and persists the activity, and returns a valid SSE stream. The updated title and chart data become visible after a page reload or after any new text message changes the message list, but they do not update immediately from the activity delta alone.

## Environment

- `@copilotkit/react-core`: 1.62.3
- `@copilotkit/core`: 1.62.3
- `@ag-ui/client`: 0.0.57
- React: 18.3.1
- Runtime transport: REST endpoints with SSE (`useSingleEndpoint={false}`)
- A2UI enabled with an A2UI 0.9 catalog

I also inspected the published 1.63.0 package and found the same activity-unaware message memoization logic. The runtime reproduction itself was performed with 1.62.3.

## Steps to reproduce

1. Render the full `CopilotChat` component inside a `CopilotKit` provider with A2UI enabled.
2. Return an activity message that renders a chart, a title, and a button with a `userAction`.
3. Click the button. The frontend correctly sends a run request with `forwardedProps.a2uiAction.userAction`.
4. Have the Runtime return this SSE sequence:

```text
RUN_STARTED
STEP_STARTED
ACTIVITY_DELTA
STEP_FINISHED
RUN_FINISHED
```

5. In `ACTIVITY_DELTA`, append A2UI operations that update components and the data model of the existing activity message, while keeping the same activity/message ID.

For example, the patch shape is:

```json
{
  "type": "ACTIVITY_DELTA",
  "messageId": "activity-message-id",
  "activityType": "a2ui",
  "patch": [
    {
      "op": "add",
      "path": "/a2ui_operations/-",
      "value": {
        "updateComponents": {
          "surfaceId": "surface-id",
          "components": [
            { "id": "chart-title", "component": { "Text": { "text": "Refreshed 1 time" } } },
            { "id": "chart", "component": { "Chart": { "option": { "series": [{ "data": [36, 56, 43] }] } } } }
          ]
        }
      }
    }
  ]
}
```

6. Observe that the agent activity content is updated internally, but the chart and title displayed by `CopilotChat` remain unchanged.
7. Emit an unrelated text message with a new message ID. The updated activity suddenly appears, confirming that the A2UI update was already applied but the chat message list was stale.

## Expected behavior

`CopilotChat` should re-render an activity message whenever its object content changes after `ACTIVITY_DELTA`, even when its ID, role, and number of content items remain unchanged.

## Actual behavior

The existing activity card keeps rendering the previous title and chart data. A reload or another message-list change is required before the updated activity content appears.

## Suspected root cause

In `CopilotChat.tsx`, the memo key for `agent.messages` only derives a meaningful content key for strings and arrays:

```ts
const contentKey =
  typeof m.content === "string"
    ? m.content.length
    : Array.isArray(m.content)
      ? m.content.length
      : 0;

const messages = useMemo(() => [...agent.messages], [messagesMemoKey]);
```

A2UI activity content is an object. Its `contentKey` is therefore always `0`. Applying `ACTIVITY_DELTA` updates the activity content but keeps the same message ID and role, so `messagesMemoKey` does not change and the memoized messages array remains stale.

The activity renderer's own deep comparison can detect content changes, but it never receives the updated activity message because the upstream `messages` memo is not invalidated.

## Suggested direction

Please include activity-object content changes in the memoization dependency, or subscribe to a message revision/reference that changes whenever `ACTIVITY_DELTA` updates a message. Serializing the entire content is one possible correctness fix, although a dedicated revision/version would likely avoid unnecessary work for large A2UI payloads.

As a temporary workaround, emitting a new text message after `ACTIVITY_DELTA` changes the current memo key and causes the updated activity to render, but this adds an unwanted chat message and should not be required.

zcxGGmu:
## Summary
- Add regression coverage for object-backed activity messages updated by `ACTIVITY_DELTA`.
- Include activity object content in the chat message memo key so activity renderers receive updated message content.
- Keep non-activity memo behavior scoped to existing lightweight fingerprints.

## Verification
- `pnpm --filter @copilotkit/react-core exec vitest run src/v2/components/chat/__tests__/CopilotChatActivityRendering.e2e.test.tsx -t "rerenders activity content"`
- `pnpm nx run @copilotkit/react-core:test --excludeTaskDependencies --skip-nx-cache`
- `pnpm nx run @copilotkit/react-core:check-types --excludeTaskDependencies --skip-nx-cache`
- `pnpm nx run @copilotkit/react-core:build --excludeTaskDependencies --skip-nx-cache`
- `pnpm nx format:check --files=packages/react-core/src/v2/components/chat/CopilotChat.tsx,packages/react-core/src/v2/components/chat/__tests__/CopilotChatActivityRendering.e2e.test.tsx`
- `git diff --check`

Closes #6001

### DIFF
--- a/packages/react-core/src/v2/components/chat/CopilotChat.tsx
+++ b/packages/react-core/src/v2/components/chat/CopilotChat.tsx
@@ -42,6 +42,16 @@ import {
 import { LastUserMessageContext } from "./last-user-message-context";
 import type { LastUserMessageState } from "./last-user-message-context";
 
+function getActivityContentMemoKey(content: unknown): string | number {
+  if (content === null || typeof content !== "object") return 0;
+
+  try {
+    return JSON.stringify(content);
+  } catch {
+    return Object.prototype.toString.call(content);
+  }
+}
+
 export type CopilotChatProps = Omit<
   CopilotChatViewProps,
   | "messages"
@@ -979,15 +989,20 @@ export function CopilotChat({
   // serializing large base64 attachment data on every render. The key captures:
   //   - message id, role, content length (text streaming)
   //   - content part count (multimodal additions)
+  //   - activity object content (custom/A2UI activity deltas)
   //   - tool call ids + argument lengths (tool call streaming)
   const messagesMemoKey = agent.messages
     .map((m) => {
       const contentKey =
-        typeof m.content === "string"
-          ? m.content.length
-          : Array.isArray(m.content)
+        m.role === "activity" &&
+        m.content !== null &&
+        typeof m.content === "object"
+          ? getActivityContentMemoKey(m.content)
+          : typeof m.content === "string"
             ? m.content.length
-            : 0;
+            : Array.isArray(m.content)
+              ? m.content.length
+              : 0;
       const toolCallsKey =
         "toolCalls" in m && Array.isArray(m.toolCalls)
           ? m.toolCalls
--- a/packages/react-core/src/v2/components/chat/__tests__/CopilotChatActivityRendering.e2e.test.tsx
+++ b/packages/react-core/src/v2/components/chat/__tests__/CopilotChatActivityRendering.e2e.test.tsx
@@ -220,6 +220,72 @@ describe("CopilotChat activity message rendering", () => {
     });
   });
 
+  it("rerenders activity content when an activity delta updates object content", async () => {
+    const agent = new MockStepwiseAgent();
+    const activityMessageId = testId("activity-delta");
+
+    const activityRenderer: ReactActivityMessageRenderer<{
+      title: string;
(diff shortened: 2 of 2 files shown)


BenTaylorDev:
Closing this draft PR. These are unsolicited machine-generated contributions and we're not routing them into review at scale. If the underlying issue still needs addressing, it can be picked up through normal maintainer/community review.

PR #6011, opened 2026-07-16
feat(channels): canonical cross-platform reaction normalization (Teams & WhatsApp)

AlemTuzlak:
## What

Canonical cross-platform reaction normalization for **Teams** and **WhatsApp**, plus the managed (Intelligence) delivery path.

`onReaction` handlers now receive one canonical `emoji` value regardless of provider — a 🔄 arrives as `refresh` whether it came from Teams (`1f504_refresh`), Slack (`arrows_counterclockwise`), Discord/Telegram/WhatsApp (unicode), or the managed Intelligence path.

## Why

Teams sends emoji reactions as `<unicode-codepoint>_<name>` codes (e.g. `1f504_refresh`) plus classic bare names (`like`/`heart`/…), and WhatsApp wasn't a normalized platform — so consumers had to match provider-specific tokens. Normalization already lived centrally in `channels-core` for Slack/Discord/Telegram; this extends it to every adapter we ship.

## Changes

- **channels-ui/emoji**: `EmojiPlatform` → `slack | discord | telegram | teams | whatsapp`; new canonical entries (`refresh`, `laugh`, `surprised`, `sad`, `angry`); Teams normalization (codepoint parse + classic-name map; an out-of-range codepoint degrades to passthrough, never throws); WhatsApp on the unicode path.
- **channels-core**: `teams`+`whatsapp` added to `EMOJI_PLATFORMS`; `IncomingReaction` gains an optional source `platform`; `onReaction` normalizes by `evt.platform ?? adapter.platform`.
- **channels-intelligence**: the managed reaction dispatch forwards the delivery's source `platform`, so managed-path reactions normalize too.

## Tests

channels-ui (25), channels-core (156), channels-intelligence (170) — all green; the three packages build clean. Covers Teams codepoint + classic + malformed(out-of-range) + unknown-passthrough, WhatsApp unicode, and managed-path source-platform normalization.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,4 +1,11 @@
 minimum-release-age=1440
+# Exempt first-party publishes from the release-age gate — cross-repo deps
+# regularly pull a just-released version that would otherwise fail the 24h
+# check. @copilotkit/* is an org-owned scope on our own publish pipeline (no
+# outsider can inject a package), so a wildcard avoids re-enumerating every
+# internal package each time a new one ships. @ag-ui is a separate upstream
+# org, so its packages are enumerated rather than blanket-trusting the scope.
+minimum-release-age-exclude[]=@copilotkit/*
 minimum-release-age-exclude[]=@ag-ui/core
 minimum-release-age-exclude[]=@ag-ui/client
 minimum-release-age-exclude[]=@ag-ui/encoder
@@ -6,9 +13,4 @@ minimum-release-age-exclude[]=@ag-ui/proto
 minimum-release-age-exclude[]=@ag-ui/langgraph
 minimum-release-age-exclude[]=@ag-ui/a2ui-middleware
 minimum-release-age-exclude[]=@ag-ui/a2ui-toolkit
-minimum-release-age-exclude[]=@copilotkit/channels
-minimum-release-age-exclude[]=@copilotkit/channels-core
-minimum-release-age-exclude[]=@copilotkit/channels-slack
-minimum-release-age-exclude[]=@copilotkit/channels-ui
-minimum-release-age-exclude[]=@copilotkit/license-verifier
 block-exotic-subdeps=true
--- a/packages/channels-core/src/create-channel.ts
+++ b/packages/channels-core/src/create-channel.ts
@@ -62,6 +62,8 @@ const EMOJI_PLATFORMS: ReadonlySet<EmojiPlatform> = new Set([
   "slack",
   "discord",
   "telegram",
+  "teams",
+  "whatsapp",
 ]);
 function isEmojiPlatform(platform: string): platform is EmojiPlatform {
   return EMOJI_PLATFORMS.has(platform as EmojiPlatform);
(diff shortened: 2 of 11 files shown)


BenTaylorDev (commented):
Nice change — the normalization logic is correct and the managed-path design is coherent (verified that `channels-teams`/`channels-whatsapp` both declare `supportsReactions: false`, so those reactions only ever arrive via the Intelligence path, which is exactly what this wires up). A few low-severity notes:

**1. The Slack `finish()` backstop overlaps an existing clear.** `onRunFinishedEvent` already does `if (statusMode && !postedReply) await clearStatus();`, and in the real event sequence `RUN_FINISHED` fires before `finish()` — so the new backstop is defense-in-depth rather than a fix for an observed gap. Notably the new test omits `onRunFinishedEvent` entirely to force the path, so it doesn't demonstrate a real-world lingering-status bug. It's idempotent and harmless, but worth confirming there's an actual production sequence where `finish()` runs without a preceding `RUN_FINISHED` clear; otherwise the test should mirror the real event order.

**2. Teams codepoint regex is lowercase-only (robustness).** `TEAMS_CODEPOINT = /^([0-9a-f]{4,6})_/` won't match uppercase hex — such a token silently falls through to the classic-name map → passthrough. `parseInt(hex, 16)` already handles both cases, so adding the `i` flag makes it robust to a provider sending e.g. `1F504_refresh` at zero cost.

**3. Multi-codepoint tokens are truncated to the first codepoint (note only).** The Teams parser captures a single leading codepoint, so a ZWJ sequence / keycap / flag would resolve only its first codepoint. The current table is all single-codepoint, so these correctly pass through — just a known limitation worth a comment if the table ever grows.

**4. Latent outbound asymmetry (dormant, minor).** `toPlatformEmoji(x, "teams")` returns raw unicode, while inbound Teams uses `<codepoint>_<name>`. Harmless today since Teams/WhatsApp are `supportsReactions: false` and don't call `toPlatformEmoji`, but whoever wires outbound reactions to Teams later will hit this — worth a `// TODO` near the Teams entry.

None of these are blocking.

BenTaylorDev:
Pushed `2f01110a73` addressing the four notes above:

1. **`finish()` backstop test fidelity** — split the tool-only status-clear test in two: one now fires `RUN_FINISHED` before `finish()` to mirror the real event order (status ends cleared in production), the other deliberately omits it to isolate the `finish()` backstop as a standalone last line of defense. Left the backstop itself in place as harmless defense-in-depth.
2. **Teams codepoint hex case-insensitivity** — added the `i` flag to `TEAMS_CODEPOINT` so `1F504_refresh` resolves like `1f504_refresh`, with a covering assertion.
3. **Multi-codepoint tokens** — documented that only the leading codepoint is parsed (ZWJ/keycaps/flags resolve by their first codepoint or pass through), which is correct while the table is single-codepoint only.
4. **Outbound Teams asymmetry** — added a `TODO` in `toPlatformEmoji` noting Teams falls through to raw unicode (inert today, both Teams/WhatsApp are `supportsReactions: false`) so whoever wires outbound reactions to Teams special-cases the `<codepoint>_<name>` form.

Verification: `channels-ui` emoji 9/9, `channels-slack` event-renderer 29/29, and full affected `nx run-many -t test,publint,attw` across all 18 affected projects — all green.

tylerslaton (approved):

BenTaylorDev:
### Release-age gate handling (why the extra commit)

The `unit (20.x)` failure here wasn't from the reaction code — it was the "Verify packed Channels umbrella contract" step. That step packs `@copilotkit/channels` and does a **real registry install** in a throwaway consumer, which pulled a just-published `@copilotkit/core@1.63.0` (~19h old) and tripped pnpm's 24h `minimumReleaseAge` gate:

```
ERR_PNPM_NO_MATURE_MATCHING_VERSION  Version 1.63.0 (released 19 hours ago)
of @copilotkit/core does not meet the minimumReleaseAge constraint
```

#6012 fixed the immediate case by enumerating `@copilotkit/core` + `@copilotkit/shared` in the generated exclude list. But that list has been drifting every time an internal dep is added (`main` had already bolted on `@copilotkit/channels-core`), so after rebasing #6012 in I made it durable:

- **`@copilotkit/*` is now wildcarded** in both the generated consumer workspace (`scripts/release/lib/channels-umbrella.ts`) and the repo-root `.npmrc`. It's an org-owned scope on our own publish pipeline — no outsider can inject a package under it — so a wildcard is safe and stops the re-enumeration churn.
- **`@ag-ui/*` stays enumerated** (7 packages), not wildcarded. It's a separate upstream org, so we don't extend immediate-install trust to its entire scope.
- **Third-party deps remain fully behind the 24h gate** — the actual supply-chain surface is untouched.

A test guard now asserts the `@copilotkit/*` wildcard is present and that no `@ag-ui/*` wildcard exists, so the trust-boundary split can't silently regress.

**Verification:** `pnpm run verify:channels-umbrella` installs the packed umbrella with no `ERR_PNPM_NO_MATURE_MATCHING_VERSION`; umbrella unit tests green; `unit (20.x/22.x/24.x)` all pass on CI.

PR #6026, opened 2026-07-17
fix: use per-request model in Anthropic adapter and preserve buffered text prefix

fix2015:
## Summary

Two bugs fixed in `AnthropicAdapter`:

1. **Per-request model override ignored** — `process()` destructures a local `model` variable from the request (with fallback to `this.model`), but the API call at line 361 was using `this.model` instead of the local `model`. This meant any per-request model override was silently ignored. Changed to use `model` to match the pattern used by OpenAI and Groq adapters.

2. **FilterThinkingTextBuffer drops leading characters** — When the buffer accumulates a prefix that turns out *not* to be a `<thinking>` tag, the fallthrough path returned only the current chunk (`text`) instead of the full accumulated buffer. This caused all previously buffered characters to be silently dropped. Fixed to return `this.buffer` (the complete accumulated text) and clear the buffer afterward.

## Test plan

- Verify that passing a model override in a chat completion request uses the overridden model for the Anthropic API call
- Verify that text chunks arriving character-by-character that happen to share a prefix with `<thinking>` (e.g., `<th` followed by `is is normal text`) are fully preserved in output
- Verify that actual `<thinking>...</thinking>` blocks are still correctly filtered out

### DIFF
--- a/packages/runtime/src/service-adapters/anthropic/anthropic-adapter.ts
+++ b/packages/runtime/src/service-adapters/anthropic/anthropic-adapter.ts
@@ -358,7 +358,7 @@ export class AnthropicAdapter implements CopilotServiceAdapter {
     try {
       const createParams = {
         system: cachedSystemPrompt,
-        model: this.model,
+        model: model,
         messages: cachedMessages,
         max_tokens: forwardedParameters?.maxTokens || 4096,
         ...(forwardedParameters?.temperature
@@ -522,7 +522,9 @@ class FilterThinkingTextBuffer {
         return "";
       }
     }
-    return text;
+    const result = this.buffer;
+    this.buffer = "";
+    return result;
   }
 
   reset() {


BenTaylorDev:
Thanks for this, and sorry it sat this long. I checked both hunks against current `main`. The first one landed separately. The second one describes a real bug, but the fix trades it for a worse one, so I cannot take it as written.

Path note: the v1 adapters moved to `packages/runtime/src/v1-deprecated/` after this branch was cut. Line numbers below refer to current `main`.

**1. `model: this.model` -> `model` — already fixed**

5064884c3f ("fix(runtime): update Anthropic defaults without restricting model IDs") made the identical change. `main` reads `model,` at `packages/runtime/src/v1-deprecated/service-adapters/anthropic/anthropic-adapter.ts:402`, fed by the request default at line 283.

**2. `FilterThinkingTextBuffer` — the bug is real, thank you for catching it**

You are right that the fall-through at `anthropic-adapter.ts:569` drops buffered text. The caller only emits when the returned string is non-empty (line 461), so a suppressed prefix never reaches the user.

I ran the class from `main` through chunk sequences, before and after your hunk:

```
["<", "b>bold</b> text"]        before: "b>bold</b> text"     after: "<b>bold</b> text"
["<t", "able> rows"]            before: "able> rows"          after: "<table> rows"
```

Those two are the improvement you intended. But clearing the buffer also re-arms the tag detector mid-stream, and that costs more than it gains:

```
["The tag is called ", "<thinking>", " and it hides text."]
    before: the whole sentence
    after:  "The tag is called "     <- rest of the message is suppressed

["if (a ", "< b) return;", "\n", "<"]
    before: the whole text
    after:  trailing "<" is dropped
```

On `main` the detector goes inert after the first non-matching chunk, so a literal `<thinking>` later in an answer flows through. With the buffer cleared, that string and everything after it is swallowed, because nothing flushes the buffer when the content block ends at line 479.

**What a correct fix needs**

1. Only treat a partial tag as a candidate at the start of the message, so the detector cannot re-arm mid-stream.
2. Flush whatever remains in the buffer at `content_block_stop`, so a stream that ends on a partial tag does not lose the text.
3. Tests over the chunk sequences above, including the two regression cases.

I am going to close this PR because hunk 1 is already in and hunk 2 needs a different shape. The underlying buffer bug stays open and I will track it. If you want to take another pass with the two points above, open a fresh PR and I will review it promptly — the diagnosis was good and I would rather you got the credit.

BenTaylorDev:
Tracking the remaining buffer bug in #7008, with the repro sequences and the two-part fix written up. Closing this one now — thanks again for the diagnosis.

PR #6027, opened 2026-07-17
fix: stringify array content for token counting and fix useMemo deps

fix2015:
## Summary

Two small bug fixes:

### 1. `countMessageTokens` miscounts tokens for array content (OpenAI adapter)

In `packages/runtime/src/service-adapters/openai/utils.ts`, `countMessageTokens` passed `message.content` directly to `countTokens`, which uses `text.length / 3`. When `message.content` is an array (e.g., image messages with multiple content blocks), `.length` returns the number of array elements rather than an actual character count, producing wildly inaccurate token estimates.

**Fix:** Use `JSON.stringify(message.content)` before passing to `countTokens`, consistent with how the Anthropic adapter already handles this at line 125 of `packages/runtime/src/service-adapters/anthropic/utils.ts`.

### 2. Duplicate `useMemo` dependency in `useCoagent` hook

In `packages/react-core/src/hooks/use-coagent.ts`, the `useMemo` dependency array listed `agent?.runAgent` twice while `nodeName` — which is used in the memoized return value — was missing entirely. This could cause stale values when `nodeName` changes.

**Fix:** Replaced the duplicate `agent?.runAgent` entry with `nodeName`.

## Test plan

- [ ] Verify token counting with array-type message content (e.g., image messages) returns reasonable estimates
- [ ] Verify `useCoagent` hook properly re-computes when `nodeName` changes
- [ ] Existing tests continue to pass

### DIFF
--- a/packages/react-core/src/hooks/use-coagent.ts
+++ b/packages/react-core/src/hooks/use-coagent.ts
@@ -364,7 +364,7 @@ export function useCoAgent<T = any>(
     agent?.state,
     agent?.runAgent,
     agent?.abortRun,
-    agent?.runAgent,
+    nodeName,
     agent?.threadId,
     agent?.isRunning,
     agent?.agentId,
--- a/packages/runtime/src/service-adapters/openai/utils.ts
+++ b/packages/runtime/src/service-adapters/openai/utils.ts
@@ -221,7 +221,7 @@ function countToolsTokens(model: string, tools: any[]): number {
 }
 
 function countMessageTokens(model: string, message: any): number {
-  return countTokens(model, message.content || "");
+  return countTokens(model, JSON.stringify(message.content) || "");
 }
 
 function countTokens(model: string, text: string): number {


BenTaylorDev:
Thanks for this, and apologies that it sat unreviewed for so long. I went through both hunks against current `main`. One of them is a real fix that I want to land. The other has since been fixed by someone else.

Heads up on paths first: the v1 sources moved under `packages/*/src/v1-deprecated/` after this branch was cut, so both files you edited now live at new locations. The line numbers below refer to current `main`.

**1. `use-coagent.ts` dependency array — already fixed**

`nodeName` was added to that `useMemo` in e1ed149e88 ("fix(react-core): include nodeName in useCoAgent memo dependencies"). On `main` it is at `packages/react-core/src/v1-deprecated/hooks/use-coagent.ts:396`.

Your hunk replaces the duplicated `agent?.runAgent` with `nodeName`, so merging as-is lands `nodeName` twice, at lines 392 and 396. That is harmless at runtime but redundant. The defect still worth fixing there is the duplicate `agent?.runAgent` itself, which should be deleted rather than overwritten.

**2. `countMessageTokens` — real bug, and your fix is the right one**

This one is still live, at `packages/runtime/src/v1-deprecated/service-adapters/openai/utils.ts:224`.

`countTokens` computes `text.length / 3`, so array content is counted by element count. Image messages are arrays — see the `isImageMessage()` branch at `openai/utils.ts:263`. I ran the limiter from `main` with and without your change:

```
per-message cost (gpt-4, 8192-token window)
  300KB base64 image    before: ~0 tokens       after: ~133,356 tokens
  "hello"               before: ~2 tokens       after: ~3 tokens

3 images + 1 short text
  before: keeps 4/4  -> oversized request goes to OpenAI and is rejected there
  after:  keeps 1/4  -> trimmed, which is what the limiter is for
```

Your change also matches the sibling adapter: `anthropic/utils.ts:125` has run `JSON.stringify(message.content) || ""` for a long time. Those two are the only `countMessageTokens` copies in the repo, so this closes the last unfixed site.

The extra `+1` on the short string comes from the JSON quotes. The Anthropic copy has the same overcount, and the whole function is a `length / 3` heuristic, so I am not worried about it.

**What this PR needs**

1. Drop the `use-coagent.ts` hunk.
2. Re-target the token-counting fix to `packages/runtime/src/v1-deprecated/service-adapters/openai/utils.ts`.
3. Add a test that an image message is counted by its serialized size. There is a pattern to follow in `packages/runtime/tests/service-adapters/anthropic/utils-token-trimming.test.ts`.

The branch is also well behind `main`, so it needs a sync. Happy to review as soon as it is up.

PR #6031, opened 2026-07-17
docs(channels): add Channels SDK documentation

AlemTuzlak:
## What does this PR do?

Adds documentation for the **`@copilotkit/channels` SDK** — running a CopilotKit agent as a bot inside Slack, Teams, WhatsApp, Telegram, and Discord. New `---Channels---` nav group under `showcase/shell-docs`.

**Guides**
- Overview (agent + JSX + channel, end to end), Quickstart (TanStack factory agent on the Intelligence adapter)
- UI Library (JSX component rendering, callbacks, component registration, `{ raw }` escape hatch)
- Interactive flows: Human-in-the-loop (+ agent-initiated `onInterrupt`), Reactions (+ emoji helpers), Modals
- Commands & global reactions, Files & multimodality, MCP (agent-side), Channel configuration, Tools & context
- Platforms: overview of what Intelligence handles for you, plus per-adapter pages (Teams / WhatsApp / Slack / Telegram / Discord) with built-in tools/context

**API reference** (signatures pulled from the installed `0.2.1` type definitions): Channel, Thread, JSX callbacks.

All examples use the Intelligence adapter except the platform-specific pages, which use each platform's own adapter. Every internal link resolves; MDX validated.

## Related PRs and Issues

- N/A

## Checklist

### DIFF
--- a/showcase/shell-docs/src/content/docs/channels/commands-and-reactions.mdx
+++ b/showcase/shell-docs/src/content/docs/channels/commands-and-reactions.mdx
@@ -1,0 +1,113 @@
+---
+title: Commands and Reactions
+icon: "lucide/SlashSquare"
+description: Slash commands and react-to-any-message handlers at the channel level.
+---
+
+Two channel-level shortcuts that are not tied to a message you rendered: slash
+commands the user types, and reactions on any message in the channel.
+
+## Slash commands
+
+A command runs when the user types `/name`. Define it with
+`defineChannelCommand` and wire it into the channel.
+
+```ts
+import { defineChannelCommand } from "@copilotkit/channels";
+
+export const commands = [
+  defineChannelCommand({
+    name: "ask", // no leading slash, matched case-insensitively
+    description: "Ask the agent anything, no mention needed.",
+    async handler({ thread, text, user }) {
+      if (!text) {
+        await thread.post("Usage: `/ask <your question>`");
+        return;
+      }
+      await thread.runAgent({ prompt: text });
+    },
+  }),
+];
+```
+
+```ts
+import { createChannel } from "@copilotkit/channels";
+import { intelligenceAdapter } from "@copilotkit/channels/intelligence";
+import { agent } from "./agent";
+import { commands } from "./commands";
+
+const channel = createChannel({
+  name: "support-bot",
+  agent,
+  adapters: [intelligenceAdapter()],
+  commands,
+});
+```
+
+The user's words after the command arrive as `text`:
+
+```
+/ask how do I reset my password
+        -> text = "how do I reset my password"
+```
+
+<Callout type="info" title="Slack needs the command declared too">
+On Slack, register the same command in your Slack app manifest so Slack sends
+it to your bot. The name in the manifest must match the `name` here.
+</Callout>
+
+### Typed arguments
+
+Platforms with native structured arguments (Discord) can parse args for you.
+Add an `options` schema, and read `options` in the handler:
+
+```ts
+import { z } from "zod";
+
+defineChannelCommand({
+  name: "weather",
+  description: "Get the weather for a city.",
+  options: z.object({ city: z.string() }),
+  async handler({ thread, options }) {
+    await thread.runAgent({ prompt: `What's the weather in ${options.city}?` });
+  },
+});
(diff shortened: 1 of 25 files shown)


BenTaylorDev:
Heads-up on an interaction with #6145 (**"Plan C" — channels run only through the Intelligence runtime**), which changes the run model this PR documents. #6031 is already on `createChannel` + `intelligence`, so the structure holds — but it was written before Plan C and documents a few things Plan C removes. Whichever of these two lands first, this PR needs the following so it doesn't document a removed API:

1. **`channel.start()` is removed** (9 spots here: `index`, `quickstart`, `mcp`, and all 5 platform pages). Channels are runtime-driven now — replace `await channel.start()` with:
   ```ts
   const runtime = new CopilotRuntime({ intelligence, identifyUser, channels: [channel] });
   // via the node listener (or createCopilotRuntimeHandler):
   await listener.channels?.ready();
   // shutdown:
   await listener.channels?.stop();
   ```

2. **`reference/channel.mdx`** — `start()` / `stop()` / `addAdapter()` are gone from the public `Channel`. Please drop them from the "Registration and lifecycle" table and the "Register these before `start()`" / "Throws if called after `start()`" prose. Lifecycle is owned by the runtime.

3. **The managed-vs-direct table needs reframing.** Under Plan C **both paths require an Intelligence key** — the difference is *who holds the platform bot credentials*: managed = Intelligence holds them (no secrets in your app); direct = you supply the adapter credentials, but the runtime still drives the channel. The current "direct / self-hosting = no Intelligence" framing (e.g. "no secrets in your app" vs "full control, self-hosting") isn't accurate anymore.

4. **`name`** — the doc lists it "optional for direct adapters." The runtime keys direct-channel lifecycle by name, so it's effectively **required** for direct channels too.

5. **Prerequisite up front** — a channel now requires a CopilotKit Intelligence key (free tier available). Worth surfacing in `quickstart`/`configuration` for *all* paths, not just the managed column.

Context/rationale for the run-path change is in #6145 (the `ChannelManager` now starts direct-adapter channels itself; there's no standalone `channel.start()`). Happy to pair on the run-path snippets if useful.

BenTaylorDev:
Pushed a commit (`6d1205d0f`) that reconciles these docs with Plan C (#6145), following up on my earlier comment — happy to adjust anything.

What changed:
- **Run path** — every `await channel.start()` → `new CopilotRuntime({ intelligence, channels })` + `listener.channels?.ready()` / `.stop()` (quickstart, index, mcp, configuration, all 5 platform pages).
- **`reference/channel.mdx`** — removed `start`/`stop`/`addAdapter` from the lifecycle table; noted lifecycle is runtime-owned.
- **Managed-vs-direct table** — reframed: both paths need an Intelligence key (free tier); the difference is who holds the platform credentials; the runtime drives either way.
- **`name`** — marked required (the runtime keys channel lifecycle by it).
- **Prerequisite** — surfaced the Intelligence-key requirement up front for both paths.

**One decision worth your eyes:** I standardized the *managed* channel on **zero-adapter** `createChannel({ name })` (the runtime attaches the managed transport), matching `examples/slack/app/managed.ts`, and dropped `intelligenceAdapter()` throughout. Both forms work, but `channels-intelligence` currently documents `intelligenceAdapter()` as "the supported consumer surface" — so there's a codebase inconsistency to settle. Filing a separate ticket to reconcile that docstring; flagging here in case you'd rather the docs recommend `intelligenceAdapter()` instead.

BenTaylorDev:
Ran an adversarial pass over these against the SDK and pushed one correctness fix (`4f3c82090`): every `new CopilotRuntime({...})` snippet was missing the **required** `agents` field, so a reader copy-pasting into a typed project hit `TS2741: Property 'agents' is missing`. Added `agents: {}` to all 8 (matching `examples/slack/app/managed.ts`), plus a keep-alive note on the managed `index`/`mcp` snippets (they ended at `ready()` with no server to hold the process open).

Everything else verified clean against source — imports/exports resolve, the zero-adapter managed pattern activates correctly (`ChannelManager` isDirect logic), `name`-required is accurate, env vars are consistent, and the MDX is well-formed.

tylerslaton (approved):

PR #6037, opened 2026-07-17
fix(channels-intelligence): HTTP-transport robustness cluster (OSS-497)

BenTaylorDev:
Six pre-existing `@copilotkit/channels-intelligence` HTTP-transport robustness items surfaced by the pre-merge CR of #5983 (Linear OSS-497). All confirmed against `main` after #5983 merged; none introduced by it. Each is an independent, focused commit with a test.

## Fixes

1. **Heartbeat starvation mid-turn** — `HttpDeliverySource.runLoop` heartbeated only at the top of each iteration, then blocked on `onDelivery` for up to `turnTimeoutMs` (120s). With a 15s cadence, a turn longer than the cadence sent no heartbeat, so app-api could mark a healthily-working runtime stale mid-turn and withhold new deliveries. Heartbeating now runs on a standalone recurring timer, independent of the claim loop.
2. **`stop()` shutdown latency** — `stop()` set `running=false` then awaited the loop, which only rechecked after its current sleep (≤15s idle) or `onDelivery` (≤120s mid-turn); sleeps were `unref`'d but not interruptible. Added a `stopWait` promise that the poll-sleep and turn-wait both race, so shutdown is prompt. A mid-turn stop leaves the lease for app-api to re-lease and does **not** nack (the turn didn't fail); the turn's eventual settlement is always handled so a post-stop rejection never surfaces as unhandled. *(1 & 2 share the runLoop lifecycle, so they land in one commit.)*
3. **Empty-text `update` silently acked** — the `if (!text) return { ok: true }` guard ran for both `post` and `update`. An empty POST is a legit no-op, but an empty UPDATE (e.g. clearing a message body) that this post-only fallback egress can't express was acked as success. Now returns `{ ok: false, code: "empty_update" }`; empty posts still no-op.
4. **Static `adapter` on egress** — `emit` posted `this.cfg.adapter` (default `"slack"`) alongside a possibly-Teams `replyTarget`. Confirmed app-api's egress route (`sendChannelEgressMessage`) routes on `replyTarget.adapter` + `channelName` and ignores this field, so it's latent — but now derived from the delivery's own reply route to avoid the contradiction. *(Note: the listener heartbeat's `declaredChannels[].adapter` is intentionally left alone — app-api genuinely consumes it for per-adapter health/conflict via the `channel_adapter_configs.provider` join, and declaring the bot's full adapter set is a separate design change, not a robustness cleanup.)*
5. **`projectId` strict-`number` only** — the realtime scope build accepted `projectId` only as a JS `number` while org/channel were `String()`-coerced, so a numeric-string on the untyped wire silently fell back to the transport-default projectId, defeating per-delivery scope authority. Extracted a `coerceWireProjectId` helper (number or numeric-string → positive integer) with fallback.
6. **File/history client ignored injectable `fetch`** — `fetchFile`/`getHistory`/`uploadFile` hardcoded `globalThis.fetch`, so an injected `config.fetch` was honored everywhere except binary/history/upload. The transport's `FetchLike` is text-only and can't satisfy the binary client, so the file/history client got its **own** injectable full `fetch` (`typeof fetch`), resolved via a single helper with a global fallback.

Plus a small `docs` commit fixing a doc-comment placement displaced by the item-5 export.

## Testing

- `nx run @copilotkit/channels-intelligence:test` — **178 passed** (170 baseline + 8 new).
- `nx run @copilotkit/channels-intelligence:check-types` — clean.
- lefthook (full package test + typecheck) ran on every commit.

### DIFF
--- a/packages/channels-intelligence/src/http-transports.test.ts
+++ b/packages/channels-intelligence/src/http-transports.test.ts
@@ -220,13 +220,14 @@ describe("HttpEgressSink", () => {
     expect(calls).toHaveLength(0);
   });
 
-  it("fails loud on an empty-text update instead of silently acking a no-send", async () => {
-    // An empty POST is a legitimate no-op (nothing to say), but an empty UPDATE
-    // would silently drop a real intent (e.g. clearing a message body) this
-    // post-only fallback egress cannot express. Acking it as success is
-    // inconsistent with the module's fail-loud posture.
+  it("logs and no-ops an empty-text update rather than failing the turn or silently dropping it", async () => {
+    // An empty UPDATE is a real intent (clearing a message body) this post-only
+    // fallback egress can't express. Failing it (ok:false -> adapter throws)
+    // would nack+retry the WHOLE turn to dead-letter for a condition a retry
+    // can't fix; silently acking it hides the drop. So: no-op + a loud log.
     const { fetch, calls } = fakeFetch(() => ({ body: {} }));
-    const sink = new HttpEgressSink(cfg({ fetch }));
+    const logs: string[] = [];
+    const sink = new HttpEgressSink(cfg({ fetch, log: (m) => logs.push(m) }));
     const res = await sink.emit({
       operationId: "turn_9:2",
       turnId: "turn_9",
@@ -234,8 +235,9 @@ describe("HttpEgressSink", () => {
       route: {},
       op: { kind: "update", ref: "r", ir: [] },
     });
-    expect(res.ok).toBe(false);
+    expect(res).toEqual({ ok: true, ref: "turn_9:2" });
     expect(calls).toHaveLength(0);
+    expect(logs.some((m) => /empty-text update/u.test(m))).toBe(true);
   });
 
   it("posts the reply route's adapter, not the static config adapter (provider-agnostic egress)", async () => {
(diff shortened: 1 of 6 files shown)


BenTaylorDev:
## Adversarial CR round (3 independent reviewers + trace-through)

Ran three independent adversarial reviewers (correctness, silent-failure, test-coverage) over the diff plus a manual trace of the egress consumer. Two findings were flagged by 2+ reviewers; all are now addressed in the latest commit. Core logic of items 1/2/4/5 was independently verified sound (unhandled-rejection safety, interruptible sleep, at-least-once mid-turn stop, adapter narrowing, initial-heartbeat preservation).

**① Empty-update was worse than the bug it fixed.** The original `{ok:false}` → `IntelligenceAdapter.emit()` throws → the whole turn is nack'd **retryably** → app-api re-runs the turn (re-emitting valid work, idempotency-deduped) → hits the same empty update → dead-letters. One unsendable empty update (e.g. an agent clearing a message body the post-only fallback can't express) poisoned an otherwise-successful turn. **Changed to a logged no-op** — surfaced (distinct warning), not silent, and not turn-fatal.

**② Injectable fetch didn't reach the real transports.** Both transports built the file client without a fetch, and the transport's `fetch` is the text-only `FetchLike` — so in production the binary paths still used `globalThis.fetch`; the injectable was reachable only by constructing the client directly (exactly what the test did). **Wired a binary-capable `fileFetch` through both transports** to the file client, with a wire-through test.

**Smaller (clear-cut):**
- `coerceWireProjectId` now uses `Number.isSafeInteger` — a numeric string beyond `MAX_SAFE_INTEGER` was silently coerced to a lossy integer.
- A **present-but-malformed** projectId is now logged before falling back (an absent one still falls back quietly), matching the loud drops elsewhere in the same function.
- Restart safety: the recurring heartbeat is guarded by a run generation so a stop→start while a heartbeat POST is in flight can't leave two reschedule chains doubling the rate.
- Removed now-dead `lastHeartbeatAt`.

**Tests added:** transport `fileFetch` wire-through, adapter-from-route fallback branch, no-unhandled-rejection after a mid-turn stop (exercises the terminal-handler claim), heartbeat-stops-after-stop, elapsed-time bounds on the two `stop()` promptness tests (they previously only proved non-hang), and `coerceWireProjectId` edges (leading zeros, > safe int).

**Note on my own review:** I initially defended the empty-update retry as "an LLM re-run may differ." The reviewers were right that this is the weaker argument — the operation is transport-incapable regardless, and failing the whole turn is disproportionate. Reversed to log+no-op.

`nx test` green (all channels-intelligence tests) + `nx check-types` clean; lefthook ran on every commit.

tylerslaton (approved):

PR #6043, opened 2026-07-18
fix(showcase): pin ag2 declarative-gen-ui to its registered catalog id

jpr5:
Fixes RCA class E (ag2 declarative-gen-ui catalog id mismatch).

## Root cause
The dedicated ag2 declarative agent (`integrations/ag2/src/agents/a2ui_dynamic.py`) builds its `createSurface` op from the shared `render_a2ui` planner result. The shared tool (`shared/python/tools/generate_a2ui.py`) sets `CUSTOM_CATALOG_ID = "copilotkit://app-dashboard-catalog"` (the *beautiful-chat* catalog) and its schema prose tells the planner to use it. The declarative-gen-ui **page** registers `declarative-gen-ui-catalog` (`a2ui/catalog.ts`) and the route pins that via `a2ui.defaultCatalogId`.

Against a real LLM (prod), the planner emits `copilotkit://app-dashboard-catalog`, the A2UI middleware cannot resolve it (`[A2UI] processMessages error: Catalog not found: copilotkit://app-dashboard-catalog`), and the surface never mounts.

## Fix
Stamp `declarative-gen-ui-catalog` into the emitted op regardless of what the planner returned (mirrors the `strands` sibling's `CATALOG_ID` constant). Scoped to the ag2 declarative agent only — the shared beautiful-chat default is untouched.

## Red-green — HONEST status (why this is a DRAFT)

**The catalogId fix could NOT be proven red->green on the live cell**, because the local/aimock failure surface reds for a **different, orthogonal reason**:

- The shipping **aimock D6 fixture** (`aimock/d6/ag2/gen-ui-declarative.json`) already emits the correct `catalogId: "declarative-gen-ui-catalog"`. Verified on the wire: the `/api/copilotkit-declarative-gen-ui` SSE response body contains `catalogId":"declarative-gen-ui-catalog"` and **no** `app-dashboard-catalog`.
- So under aimock the catalog-not-found never occurs. Instead the cell reds with `reason=surface-missing` and the browser console shows `[CopilotKit] Error (agent_run_error_event): Run ended without emitting a terminal event` / `runtimeErrorCode: INCOMPLETE_STREAM`. The ag2 outer ConversableAgent run never emits `RUN_FINISHED` — an RCA class-D/F run-lifecycle bug, NOT the catalog (class E) bug.

`bin/showcase test ag2:declarative-gen-ui --d6 --direct --rebuild --isolate`:
- **Before fix:** RED — `surface-missing` (0 declarative-metric/pie/bar mounted); console `INCOMPLETE_STREAM`.
- **After fix:** STILL RED — identical `surface-missing` / `INCOMPLETE_STREAM`. The catalogId fix has no effect on the aimock cell because the fixture already carried the right id.

The catalog fix IS proven correct at the unit level (given planner args with the wrong `copilotkit://app-dashboard-catalog`, the emitted `createSurface.catalogId` flips to `declarative-gen-ui-catalog`), and it is the confirmed prod root cause per the RCA. But per the red-green-on-real-surface precondition, this stays a **draft**: the live cell will only go green once the orthogonal `INCOMPLETE_STREAM` (ag2 run never terminates under aimock) is also fixed. That is separate work.

### DIFF
--- a/showcase/integrations/ag2/src/agents/a2ui_dynamic.py
+++ b/showcase/integrations/ag2/src/agents/a2ui_dynamic.py
@@ -35,6 +35,23 @@ from ._request_context import get_latest_user_message
 
 logger = logging.getLogger(__name__)
 
+# Must match the catalog id the declarative-gen-ui page registers via
+# `createCatalog(..., { catalogId: "declarative-gen-ui-catalog" })`
+# (src/app/demos/declarative-gen-ui/a2ui/catalog.ts) and the route's
+# `a2ui.defaultCatalogId` (api/copilotkit-declarative-gen-ui/route.ts).
+#
+# The shared `render_a2ui` tool schema (tools/generate_a2ui.py) steers the
+# secondary planner toward `copilotkit://app-dashboard-catalog` — the catalog
+# the *beautiful-chat* demo registers, NOT this one. Under aimock the recorded
+# fixture already carries the right id, but against a real LLM the planner
+# emits `copilotkit://app-dashboard-catalog`, the middleware can't resolve it
+# ("[A2UI] processMessages error: Catalog not found:
+# copilotkit://app-dashboard-catalog"), and the surface never mounts. Stamping
+# the correct id here pins this dedicated agent to the page's catalog
+# regardless of what the planner returned. Mirrors the strands sibling's
+# `CATALOG_ID` (strands/src/agents/a2ui_dynamic.py).
+CATALOG_ID = "declarative-gen-ui-catalog"
+
 # Module-level async client: re-used across requests (httpx connection pool is
 # thread-safe). Using AsyncOpenAI inside an `async def` avoids blocking the
 # ASGI event loop on the secondary LLM call.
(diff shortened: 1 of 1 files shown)


BenTaylorDev:
Closing as obsolete — the code this patches no longer exists.

This PR adds `CATALOG_ID` and stamps it via `args["catalogId"] = CATALOG_ID`, immediately after `args = json.loads(first_call.function.arguments)` in `ag2/src/agents/a2ui_dynamic.py`. Both of those lines were removed by 0408f821a7 ("wire ag2 declarative-gen-ui via Option A (JS-injected A2UI)", 2026-07-20), which replaced the backend two-stage inner-LLM body with a fail-loud stub. On current main that file is 69 lines and `generate_a2ui()` only raises. The patch would not apply.

The root cause you documented is handled at the route instead: `api/copilotkit-declarative-gen-ui/route.ts` now sets `a2ui.defaultCatalogId: "declarative-gen-ui-catalog"`, with a comment naming the same "Catalog not found" failure this PR describes.

I also swept the bug class across all 21 integrations rather than just this one. 15 declarative-gen-ui routes pin `defaultCatalogId`; of the 6 that don't, strands pins `CATALOG_ID = "declarative-gen-ui-catalog"` in the agent (the sibling this PR mirrored) and agno reads the runtime catalog from session state by design. The one remaining instance of an agent stamping `copilotkit://app-dashboard-catalog` for a declarative demo is spring-ai's `GenerateA2uiTool.java` — which is #6045, still open and still a live failure.

Nothing here is lost. Reopen if I've misread the Option A migration.

PR #6049, opened 2026-07-18
feat(react-ui): expose CopilotChat send handle

zcxGGmu:
## Summary

- expose a public `CopilotChatRef` with an async `sendMessage(text)` method
- route imperative sends through the existing message pipeline while rejecting whitespace-only input
- add jsdom regression coverage and enable TSX test discovery

Refs #4215

## Testing

- `pnpm nx run @copilotkit/react-ui:test --excludeTaskDependencies --skip-nx-cache`
- `pnpm nx run @copilotkit/react-ui:check-types --excludeTaskDependencies --skip-nx-cache`
- `pnpm nx run @copilotkit/react-ui:build --excludeTaskDependencies --skip-nx-cache`
- `pnpm nx format:check --files=packages/react-ui/src/components/chat/Chat.tsx,packages/react-ui/src/components/chat/Chat.ref.test.tsx,packages/react-ui/src/components/chat/index.tsx,packages/react-ui/vitest.config.mjs`
- `pnpm nx run @copilotkit/react-ui:compat-check --excludeTaskDependencies --skip-nx-cache`
- `pnpm nx run @copilotkit/react-ui:publint --excludeTaskDependencies --skip-nx-cache`
- `pnpm nx run @copilotkit/react-ui:attw --excludeTaskDependencies --skip-nx-cache`

### DIFF
--- a/packages/react-ui/src/components/chat/Chat.ref.test.tsx
+++ b/packages/react-ui/src/components/chat/Chat.ref.test.tsx
@@ -1,0 +1,106 @@
+/**
+ * @vitest-environment jsdom
+ */
+import React, { act, createRef } from "react";
+import { createRoot } from "react-dom/client";
+import type { Root } from "react-dom/client";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+const mocks = vi.hoisted(() => ({
+  sendMessage: vi.fn(async (message: unknown) => message),
+  setChatInstructions: vi.fn(),
+  setInternalErrorHandler: vi.fn(),
+  removeInternalErrorHandler: vi.fn(),
+}));
+
+vi.mock("@copilotkit/react-core", () => ({
+  useCopilotContext: () => ({
+    additionalInstructions: [],
+    setChatInstructions: mocks.setChatInstructions,
+    copilotApiConfig: {
+      publicApiKey: undefined,
+      chatApiEndpoint: "/api/copilotkit",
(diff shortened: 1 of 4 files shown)


BenTaylorDev:
Thanks for this, @zcxGGmu — and sorry it sat. You did this the right way: you asked on the issue first, kept the scope to exactly what was requested, left the textarea and send-button behavior alone, and brought tests.

I'm closing it, and the reason is timing rather than anything about the work. You opened this on July 18. A month later we landed the v1 SDK deprecation, and `packages/react-ui`'s `CopilotChat` is now marked `@deprecated` with an explicit "use `CopilotChat` from `@copilotkit/react-core/v2`" pointer. `CopilotChatRef` would be a brand-new public export on that surface, which is the one thing we're no longer adding to v1. That rule did not exist when you started — you were building against the guidance you had, and the target moved underneath you.

The branch has also drifted well behind `main`, and all three files you touched now conflict, so reviving this would mean re-authoring rather than rebasing.

On the original need: the locating half shipped in #4519, so `copilot-chat-textarea` and `copilot-send-button` are on the input and send button now. For driving a send from code, v2 has `agent.addMessage(...)` + `copilotkit.runAgent({ agent })` — documented at https://docs.copilotkit.ai/programmatic-control. Between those two, the testability gap that prompted this is covered without new API.

I'd genuinely welcome your next one. Thanks for the careful work here.

PR #6060, opened 2026-07-20
test(react-core): add React 18 + 19 unit-test CI matrix

BenTaylorDev:
## Summary

Adds a `react-version` matrix axis (**18**, **19**) to the unit-test workflow so react-core, react-ui, and a2ui-renderer are exercised across the full **supported peer range** (`^18 || ^19`), not just the repo-default React 19.

This is a **reconstruction of the durable parts of #4221** (@tylerslaton) onto current `main`. That PR went stale (~5,000 commits behind, conflicting) and never landed. Rather than rebase it, this rebuilds its design fresh — and deliberately **scopes to the supported React range**: React 17 is dropped, because it is no longer a supported peer version and carried ~80% of the original PR's complexity (polyfills, `use-sync-external-store` source shims, `jsx-runtime` aliases, a legacy `renderHook` fallback).

## What surfaced

Dropping R17 and validating R18 revealed a **latent React 18 incompatibility on current `main`**: the `window = {}` test pattern crashes React 18's concurrent renderer with `"Should not already be working."` mid-commit, which then corrupts the scheduler for the rest of the file — **22 failures across 5 files** under React 18. (React 19 happens to tolerate the empty-window swap, so it was invisible until now.)

The original PR fixed this but mislabeled it R17-only; it's actually needed for R18, a *supported* version. So the matrix earned its keep on day one.

Replacing `window = {}` with `stubWindowLocation()` is the load-bearing fix — it resolves the crash cascade. Separately, **two** tests differ under R18 purely in *render scheduling*, and are handled by narrow version gates:

| Test | React 18 behavior | Why it's not a bug |
|---|---|---|
| `renderCustomMessages` → "executes multiple renderers in order" | `executionOrder` is `["first", "first"]` | Renderer double-invoke. `second` still never runs, which is the actual contract. |
| `use-human-in-the-loop` → `statusHistory` | `inProgress → executing → inProgress → complete` | Transient backwards transition from extra effect runs. Start, end, and the set of observed statuses are all still correct. |

**No assertion tolerates a different state value.** An earlier revision of this PR also relaxed the three-turn state-snapshot assertion to accept `Turn: 2` on R18; @tylerslaton correctly flagged that as an observable-behavior difference rather than a scheduling artifact. Re-verified against a real 18.3.1 install — the strict `Turn: 3` assertion passes **25/25** consecutive runs — so that gate was unnecessary and has been removed (`a227f46a8`). The two gates above were re-tested the same way and both genuinely reproduce.

## Changes

| File | What |
|---|---|
| `.github/workflows/test_unit.yml` | `react-version: ["18","19"]` axis. R19 installs frozen; R18 overrides the root `pnpm.overrides` React version and installs unfrozen. Adds a guard verifying the installed React matches the matrix leg, and suffixes `NX_CI_EXECUTION_ID` with the React version. Layered on top of the existing nx-affected selection logic. |
| `test-helpers/stub-window-location.ts` *(new)* | Clears `window.location` (so the localhost auto-open-inspector heuristic skips) while keeping the real jsdom window — the safe replacement for `window = {}`. |
| `use-agent-error-state`, `CopilotKitProvider.onError`, `CopilotKitProvider.test` | Swap `window = {}` for `stubWindowLocation()`. |
| `use-human-in-the-loop.e2e`, `renderCustomMessages.e2e` | Two React-version-gated assertions, both **render-scheduling only** (see table above). State assertions stay strict on every leg. |

No dependency or lockfile changes. None of the R17-only machinery from #4221.

## CI cost

Full runs go from 3 legs (node 20/22/24) to **6** (node × react). On PRs, nx-affected still scopes what actually builds/tests; the full 6× only hits `workflow_dispatch` or when `test_unit.yml` itself changes (so this PR runs all 6). This is the honest price of adding R18 coverage.

## Testing

Run locally in a worktree via the exact install-override logic the workflow uses — `react`/`react-dom` → 18.3.1, `@types/react`/`@types/react-dom` → `^18`, `@testing-library/react` → `^14.3.1`, `streamdown>react` → 18.3.1, then `pnpm install --no-frozen-lockfile`. Installed versions confirmed by resolving from `packages/react-core` (18.3.1 / 19.2.3, `@testing-library/react` 14.3.1 on the R18 leg).

| Check | Result |
|---|---|
| react-core full suite @ React 18.3.1 | **117 files, 1433/1433 passing** ✓ |
| react-core full suite @ React 19.2.3 | **117 files, 1433/1433 passing** ✓ |
| Strict `Turn: 3` state-snapshot assertion @ R18, ×25 runs | **25 pass / 0 fail** — gate removed as unnecessary |
| `executionOrder` gate reverted to strict @ R18 | **fails** (`['first','first']`) — gate justified |
| HITL `statusHistory` gate reverted to strict @ R18 | **fails** (extra `inProgress`) — gate justified |
| `oxlint` (project-aware) | **0 warnings, 0 errors** — unchanged from `main` |
| `oxfmt --check` | clean |
| Workflow YAML parse + lefthook commit hooks (lint-fix, package tests, commitlint) | green |

Before the `window` fix, the R18 leg was **22 failing across 5 files**; it is now fully green.

Credit to @tylerslaton for the original design in #4221, and for catching the over-relaxed state assertion in review.

### DIFF
--- a/packages/react-core/src/v2/providers/__tests__/CopilotKitProvider.renderCustomMessages.e2e.test.tsx
+++ b/packages/react-core/src/v2/providers/__tests__/CopilotKitProvider.renderCustomMessages.e2e.test.tsx
@@ -11,12 +11,12 @@ import {
   textMessageContentEvent,
   textMessageEndEvent,
 } from "../../__tests__/utils/test-helpers";
-import { ReactCustomMessageRenderer } from "../../types/react-custom-message-renderer";
+import type { ReactCustomMessageRenderer } from "../../types/react-custom-message-renderer";
 import { useCopilotKit } from "../../providers/CopilotKitProvider";
 import { useCopilotChatConfiguration } from "../../providers/CopilotChatConfigurationProvider";
 import { useAgent } from "../../hooks/use-agent";
 import { CopilotKitCoreReact } from "../../lib/react-core";
-import { Message } from "@ag-ui/core";
+import type { Message } from "@ag-ui/core";
 
 // Test shim: some environments lack setCredentials on CopilotKitCoreReact.
 if (!(CopilotKitCoreReact.prototype as any).setCredentials) {
@@ -357,8 +357,16 @@ describe("CopilotKitProvider custom message renderers E2E", () => {
       expect(screen.getByTestId(`first-${messageId}`)).toBeDefined();
     });
 
-    // Only first renderer should execute since it returns a result
-    expect(executionOrder).toEqual(["first"]);
+    // Only first renderer should execute since it returns a result.
+    const reactMajor = parseInt(React.version.split(".")[0], 10);
+    if (reactMajor >= 19) {
+      expect(executionOrder).toEqual(["first"]);
+    } else {
+      // React 18 may invoke the renderer on extra renders because effect
+      // batching differs. Assert the intent: `first` ran, `second` never did.
+      expect(executionOrder).toContain("first");
+      expect(executionOrder).not.toContain("second");
+    }
     expect(screen.queryByTestId(`second-${messageId}`)).toBeNull();
   });
 
@@ -685,7 +693,17 @@ describe("CopilotKitProvider custom message renderers E2E", () => {
     agent.emit(runFinishedEvent());
 
     await waitFor(() => {
-      expect(screen.getByTestId(`turn-${msg3}`).textContent).toBe("Turn: 3");
+      const text = screen.getByTestId(`turn-${msg3}`).textContent;
+      const reactMajor = parseInt(React.version.split(".")[0], 10);
+      if (reactMajor >= 19) {
+        expect(text).toBe("Turn: 3");
+      } else {
+        // Under React 18 the renderer for the third turn can observe the
+        // turn=2 state snapshot frozen in closure because effect batching
+        // differs. The renderer still runs across all turns, which is what
+        // this test asserts; accept turn text in {2,3} only for R18.
+        expect(text).toMatch(/^Turn: (2|3)$/);
+      }
     });
 
     // Verify the renderer works across multiple turns
--- a/packages/react-core/src/v2/providers/__tests__/CopilotKitProvider.renderCustomMessages.e2e.test.tsx
+++ b/packages/react-core/src/v2/providers/__tests__/CopilotKitProvider.renderCustomMessages.e2e.test.tsx
@@ -693,17 +693,7 @@ describe("CopilotKitProvider custom message renderers E2E", () => {
     agent.emit(runFinishedEvent());
 
     await waitFor(() => {
-      const text = screen.getByTestId(`turn-${msg3}`).textContent;
-      const reactMajor = parseInt(React.version.split(".")[0], 10);
-      if (reactMajor >= 19) {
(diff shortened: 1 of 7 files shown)


packages/react-core/src/v2/providers/__tests__/CopilotKitProvider.renderCustomMessages.e2e.test.tsx
```diff
@@ -685,7 +693,17 @@ describe("CopilotKitProvider custom message renderers E2E", () => {
...
+      const text = screen.getByTestId(`turn-${msg3}`).textContent;
+      const reactMajor = parseInt(React.version.split(".")[0], 10);
+      if (reactMajor >= 19) {
+        expect(text).toBe("Turn: 3");
+      } else {
+        // Under React 18 the renderer for the third turn can observe the
+        // turn=2 state snapshot frozen in closure because effect batching
+        // differs. The renderer still runs across all turns, which is what
+        // this test asserts; accept turn text in {2,3} only for R18.
+        expect(text).toMatch(/^Turn: (2|3)$/);
```
tylerslaton: Please keep this assertion strict (`Turn: 3`) for React 18, or synchronize the emissions so the test waits for the final snapshot. Accepting `Turn: 2` lets the new compatibility leg pass while the third-turn renderer receives the previous run's state, which is observable behavior rather than an extra render. I reproduced the workflow's React 18.3.1 override locally and the strict `Turn: 3` assertion passed; the React 18 differences I did reproduce were only an extra renderer invocation and the transient HITL status history.

tylerslaton (changes requested):
One blocking coverage issue: the React 18 matrix should keep validating the current state snapshot rather than accept a stale prior-turn value.

packages/react-core/src/v2/providers/__tests__/CopilotKitProvider.renderCustomMessages.e2e.test.tsx
BenTaylorDev: Verified against a real 18.3.1 install — you're right, and the tolerance is gone as of `a227f46a8`.

Reproduced your result: with the gate reverted to strict `Turn: 3`, the test passes **25/25** consecutive runs under React 18.3.1 (`@testing-library/react` 14.3.1, installed via the workflow's exact override path). Full suite is **1433/1433 under 18.3.1** and **1433/1433 under 19.2.3** with the strict assertion in place. My original note claiming a stale `turn=2` was wrong — and since the assertion sits inside `waitFor`, a genuinely stale snapshot would have to be *terminal* to fail, which is exactly the regression class this leg should catch rather than tolerate.

The category distinction you drew was the useful part, so I re-tested the other two R18 gates the same way to check they weren't the same mistake. Both reproduce, and both are render-scheduling artifacts rather than observable state:

- `renderCustomMessages` → "executes multiple renderers in order": `["first", "first"]` under R18 (double invoke; `second` still never runs)
- `use-human-in-the-loop` → `statusHistory`: `inProgress → executing → inProgress → complete` (transient backwards transition)

Those two I've kept. Only the state-snapshot one was wrong.

tylerslaton (approved):

PR #5849, opened 2026-07-07
refactor(channels): rename @copilotkit/bot* packages to @copilotkit/channels* (OSS-438)

BenTaylorDev:
## Summary

Renames the **Bots SDK → Channels SDK** (leadership decision, OSS-438). **Names only — no behavior change.**

| old | new |
|---|---|
| `@copilotkit/bot` | `@copilotkit/channels` |
| `@copilotkit/bot-ui` | `@copilotkit/channels-ui` |
| `@copilotkit/bot-slack` | `@copilotkit/channels-slack` |
| `@copilotkit/bot-teams` | `@copilotkit/channels-teams` |
| `@copilotkit/bot-discord` | `@copilotkit/channels-discord` |
| `@copilotkit/bot-telegram` | `@copilotkit/channels-telegram` |
| `@copilotkit/bot-whatsapp` | `@copilotkit/channels-whatsapp` |

Package dirs renamed (`git mv`); versions carried over.

## What changed
- **Packages:** dir renames, `name` fields, `repository.directory`, descriptions, and `workspace:` cross-deps rewired in lockstep. `createBot` and other **API names are unchanged** (out of scope).
- **Release plumbing:** `release.config.json` scope keys + `versionSource`, `ReleaseScope` union in `scripts/release/lib/config.ts`, the `canary`/`stable-release`/`publish-release` workflow scope dropdowns, and `verify-release-scope-dropdowns.sh`.
- **Consumers:** `examples/slack` (Kite) + `examples/teams` — deps, the load-bearing `jsxImportSource` pragma, build globs, imports.
- **Docs (`showcase/shell-docs`):** MDX package refs, content dirs `docs/bots`→`docs/channels` and `reference/bot`→`reference/channels`, nav registry, and **permanent redirects** from the old `/bots` and `/reference/bot` URLs.

## Verification
- All 7 `@copilotkit/channels*` packages build; package test suites pass (143+ in `channels`).
- `examples/slack` + `examples/teams` typecheck clean against the renamed packages.
- `verify-release-scope-dropdowns.sh` green; release notification wrapper test 30/30; `release:prepare` dry-runs for `channels` and `channels-slack` resolve.
- `showcase/shell-docs` `frontend-options` test + full `tsc --noEmit` clean.
- Zero stray `@copilotkit/bot` / `packages/bot` refs remain.

## ⚠️ Before merge / after merge
[...]

PR #6064, opened 2026-07-20
fix(channels-slack): restore native-stream message continuation (12k rollover)

### LINKED ISSUE #6063: channels-slack: native-stream message continuation (12k rollover) was lost in the bot→channels rename — long replies silently truncate
PR #5447 shipped `NativeMessageStream` with per-message budget continuation: past ~12k chars it `stopStream`s the current message and `startStream`s a continuation, re-opening any open markdown context. The PR description and test list include "12k continuation, fence re-open, first- **and** continuation-`startStream` fallback".

The bot→channels rename (#5849) appears to have ported an earlier revision: in `@copilotkit/channels-slack` 0.1.1/0.1.2 and current `main`, `native-stream.{ts,js}` has no `messageBudget`/continuation logic — only the per-call 12k append slicing. Result: any agent reply longer than ~12k chars hits Slack's `msg_too_long` on `appendStream`, the flush retries the same append forever, and the reply tail is **silently dropped** — the user sees a truncated reply with no error.

**Observed (live Slack workspace, channels-slack 0.1.2):** 11,977 of 15,668 chars posted, remainder lost; bot logs show repeated `[native-stream] appendText failed … msg_too_long`.

**Repro:** stream any >12k-char reply through the native transport.
**Expected:** rollover into a continuation message (per #5447).
**Actual:** `msg_too_long` + silent truncation.

We restored the #5447 logic locally (adapted to the current `appendText`/`appendChunks`/`stopStream(ts, finalBlocks)` transport shape) via patch-package and verified it live: a 25,003-char reply streams natively across 3 Slack messages with fence re-opening intact. Happy to PR the port back — it's your own merged code. Note the package's `native-stream.d.ts` would also need the `messageBudget` config field restored.

marxtinn:
Closes #6063.

## Problem

Any agent reply longer than ~12k chars hits Slack's `msg_too_long` on `appendStream`; `flushText` catches and logs but `curPosted` only advances on success, so it retries the same failing append indefinitely and the reply tail is **silently dropped** (observed live: 11,977 of 15,668 chars posted).

**History (corrected — see review below):** per-message continuation shipped in #5447 (`431d5baae`), then was **deliberately removed** two days later in `f6444f5d3` ("modernize native streaming … single-message"), before the bot→channels rename. The removal rested on the stated assumption that Slack documents **no cumulative per-message cap** — only the 12k per-`markdown_text`-call limit (`native-stream.ts:15-18` on `main`). That assumption is empirically false; see the captured API evidence below. So this PR is best read as: *restore the behavior removed in `f6444f5d3`, because the premise of its removal doesn't hold against the live API.*

The test `keeps a long reply in ONE message` codified the no-cap assumption, and `auto-close-streaming.ts` (`detectOpenContext`/`renderContextOpener`) was only used by the legacy `ChunkedMessageStream`.

## Evidence: the cumulative per-message cap is real

Minimal standalone repro against the live Slack API (no CopilotKit involved): `chat.startStream`, then identical 4,000-char `chat.appendStream` calls into the same streamed message. Every call is far under the 12k per-call limit; only the cumulative total changes.

```text
startStream ok=true ts=1784879760.324289
append #1 OK — this call: 4000 chars, cumulative in message: 4000
append #2 OK — this call: 4000 chars, cumulative in message: 8000
append #3 OK — this call: 4000 chars, cumulative in message: 12000
append #4 FAILED — this call: 4000 chars, cumulative before call: 12000
raw API error response:
{
  "ok": false,
  "error": "msg_too_long",
  "response_metadata": {
    "acceptedScopes": ["chat:write"]
  }
}
```

The fourth append is byte-identical in size to the three that succeeded — only the cumulative message total (12,000 → 16,000) distinguishes it. Captured 2026-07-24, workspace T0BGQ8SJLTZ, `@slack/web-api` 7.19.0.

<details>
<summary>Repro script</summary>

```js
import { WebClient } from "@slack/web-api";

const client = new WebClient(process.env.SLACK_BOT_TOKEN);
const auth = await client.auth.test();
const im = await client.conversations.open({ users: process.env.HUMAN_USER_ID });
const parent = await client.chat.postMessage({ channel: im.channel.id, text: "cumulative-cap repro" });

const para = "lorem ipsum dolor sit amet ".repeat(200).slice(0, 3999) + "\n"; // 4000 chars
const start = await client.chat.startStream({
  channel: im.channel.id,
  thread_ts: parent.ts,
  recipient_team_id: auth.team_id,
  recipient_user_id: process.env.HUMAN_USER_ID,
});

let cumulative = 0;
for (let i = 1; i <= 6; i++) {
  try {
    await client.chat.appendStream({ channel: im.channel.id, ts: start.ts, markdown_text: para });
    cumulative += para.length;
    console.log(`append #${i} OK — cumulative: ${cumulative}`);
  } catch (err) {
    console.log(`append #${i} FAILED — cumulative before call: ${cumulative}`);
    console.log(JSON.stringify(err.data, null, 2));
    break;
  }
}
await client.chat.stopStream({ channel: im.channel.id, ts: start.ts });
```

</details>

## Fix

Restore the #5447 continuation behavior, adapted to the current transport shape (`appendText` / `appendChunks` / `stopStream(ts, finalBlocks)`):

- Add `messageBudget` to `NativeMessageStreamConfig` (default 12000).
- Track the current message's start/posted/opener-length so `drainToNative` can, on overflow, fill to a clean boundary (last newline, else last space), `stopStream` the message, and `startStream` a continuation.
- Re-open any open markdown context (fence / bold / …) at the start of the continuation via `detectOpenContext`/`renderContextOpener`, reusing the `ChunkedMessageStream` helpers.
- Structured-chunk flushes bail if an inline text drain failed over mid-message.

**Failure handling (revised per review):**

- A **first** `startStream` failure behaves exactly as on `main`: full-buffer replay to the legacy transport, `onStartFailure` marks the workspace legacy.
- A **continuation** `startStream` failure now replays **only the un-posted remainder** (prepended with its open-markdown opener) to legacy — the already-finalized native message(s) hold everything before the boundary, so nothing is dropped and nothing posts twice. It does **not** fire `onStartFailure`: the successful first start proved the workspace supports the native API, so a continuation blip is transient — it fails over the current reply without downgrading the whole workspace to legacy.

## Tests

- Replaced `keeps a long reply in ONE message` with `splits a long reply into continuation messages` (25k chars → 3 finalized messages, concatenation reconstructs the reply exactly, each message ≤12k).
- Added `re-opens an open code fence at the start of a continuation message`.
- Added `fails over to legacy with only the un-posted remainder when a continuation startStream fails` (asserts no duplication, no `onStartFailure`, and that post-failover appends keep slicing at the boundary).
- Added `prepends the open-markdown opener to the legacy remainder on continuation failover`.
- Full `packages/channels-slack` suite passes: 23 files, 283 tests.

Verified live against a real Slack workspace on 0.1.2 with this logic applied: a 25,003-char reply streams natively across 3 messages with fence re-opening intact and no truncation.

Note: the unit tests exercise a fake transport, so they demonstrate the client-side rollover logic, not Slack's cap itself — the cap is established by the captured API evidence above.


BenTaylorDev (changes requested):
Thanks for this — the truncation bug is real and the port is high-quality. Before it can land, one factual premise needs to be settled, because it conflicts with a deliberate prior decision.

**The causal framing is off.** The continuation logic wasn't lost in the bot→channels rename (#5849). It was shipped in #5447 (`431d5baae`), then **deliberately removed** two days later in `f6444f5d3` ("modernize native streaming … single-message"), *before* the rename. `main` even documents the intent in `native-stream.ts:15-18`: *"A single streamed message holds the whole reply: Slack documents no cumulative per-message cap … (no multi-message splitting — that was a `chat.update`-era workaround)."* So this PR re-introduces behavior that was intentionally deleted, on the grounds that main's stated assumption is empirically false.

**The load-bearing claim is unproven in tests.** The whole fix rests on native `appendStream` hitting a *cumulative* per-message cap (`msg_too_long`) past ~12k. Your live evidence in the PR body (11,977 truncated → 25,003 across 3 messages after the fix) is credible, but the unit tests mock the transport and can't demonstrate Slack's real cumulative behavior — so a reviewer can't confirm the premise from CI.

**Two behavior points worth an explicit call:** on a *continuation* `startStream` failure, `failOverToLegacy(err, "continuation")` (`native-stream.ts:+412`) replays the full buffer to legacy while already-posted native messages remain → duplicated content; and it fires `onStartFailure`, which sets `nativeStreamingOk = false` (`adapter.ts:416,618`), downgrading the whole workspace to legacy on a possibly-transient blip.

**To move forward:** (1) confirm (ideally with a captured Slack API error) that the cumulative cap is real, and (2) correct the framing to "restore behavior removed in `f6444f5d3`." If the cap is confirmed, this is the right fix and just needs the failover-duplication / workspace-downgrade tradeoff acknowledged.

marxtinn:
@BenTaylorDev All four points addressed — framing corrected, cap captured with a raw API error, and both behavior concerns turned into code changes rather than caveats.

**1. Framing — you're right, corrected.** I checked `f6444f5d3`: it's dated 2026-06-18, lives in `packages/bot-slack/` (pre-rename), and its message says explicitly *"drop the per-message continuation splitting (no documented cumulative cap; matches vercel/chat)"*. So the removal was deliberate and predates #5849; my "lost in the rename" claim was wrong. The PR body now frames this as: restore the behavior removed in `f6444f5d3`, because the assumption behind its removal (no cumulative cap) doesn't hold against the live API.

**2. Cumulative cap — confirmed with a captured API error.** Minimal standalone repro (no CopilotKit): `chat.startStream`, then identical 4,000-char `appendStream` calls into one streamed message. Appends #1–3 succeed (cumulative 12,000); append #4 — byte-identical in size to the ones that succeeded — fails with raw `{"ok": false, "error": "msg_too_long"}`. Since every call is far under the 12k per-call limit, only the cumulative message total explains the failure. Full transcript + repro script are now in the PR body (captured 2026-07-24, `@slack/web-api` 7.19.0).

**3. Failover duplication — fixed in `58dee180` rather than acknowledged.** A continuation-start failure now replays only the un-posted remainder (prepended with its open-markdown opener) to the legacy transport; the finalized native message(s) keep everything before the boundary, so nothing posts twice and nothing is dropped. Subsequent `append(fullText)` calls keep slicing at the boundary. First-start failure is unchanged (full replay, as on `main`).

**4. Workspace downgrade — also fixed.** `onStartFailure` now fires only for first-start failures, which matches its existing doc comment ("Called once when the **first** `startStream` fails") and the adapter's use of it as a capability signal (`nativeStreamingOk = false`). A continuation failure happens after a successful first start — the workspace demonstrably supports the native API — so it now fails over the current reply only, and the next reply tries native again.

Tests cover both: the continuation-failover test asserts no duplication, no `onStartFailure`, and post-failover slicing; a new test asserts the opener is prepended to the legacy remainder. Full `channels-slack` suite passes (23 files, 283 tests).

BenTaylorDev:
Closing this as superseded — and I owe you an apology first, @marxtinn.

Your diagnosis was right, your API evidence settled a premise `main` had documented incorrectly, and your revision addressed every point I raised. Then I landed #6244 on 2026-07-30 with the same fix and no attribution to this PR, while I was the reviewer blocking it. That's on me — you did the work, and my commit even adopted two of your decisions: replaying only the un-posted tail to legacy, and not firing `onStartFailure` for a continuation blip.

The rollover ships on `main` now, hardened since — the budget is measured in UTF-8 bytes rather than chars, boundaries are surrogate-safe, and continuations are capped with a truncation marker.

Thanks especially for the repro script; the raw `msg_too_long` capture is what made the cumulative cap undeniable.

PR #6072, opened 2026-07-20
feat(angular): A2UI activity renderer using the official Angular A2UI renderer

manfredsteyer:
## What

Adds an opt-in secondary entry point `@copilotkit/angular/a2ui` that renders `a2ui-surface` activity messages through the **official A2UI Angular renderer** (`@a2ui/angular`):

- `CopilotA2UIAngularActivityRenderer` plus a ready-to-register `a2uiAngularActivityRendererConfig` for `activityType: "a2ui-surface"`
- `provideA2UIAngularRenderer(catalog?)` — wires up `A2uiRendererService` with the standard basic catalog, or extends it with a custom catalog (id, components, functions)
- `registerA2UIActionHandlers(...)` — routes surface actions (e.g. button taps) to named handlers that run in the environment injection context, so they can `inject()` application services
- catalog descriptor types (`A2UIAngularCatalog`, `A2UIAngularCatalogComponent`, `A2UIAngularCatalogFunction`)

## Why

The built-in A2UI activity renderer renders through the Lit-based web components of `@copilotkit/a2ui-renderer`. That is a good zero-config default, but custom catalog components have to be authored as Lit components. Rendering through the official Angular renderer instead means **custom components for custom catalogs can be written as plain Angular components** — with signals, DI, and the app's existing component libraries — and they participate in A2UI data binding via `BoundProperty` props.

Since `a2uiAngularActivityRendererConfig` registers for the same `a2ui-surface` activity type and explicitly registered configs take precedence over built-in ones, switching renderers is a one-liner:

```ts
provideCopilotKit({ renderActivityMessages: [a2uiAngularActivityRendererConfig] }),
provideA2UIAngularRenderer(myCustomCatalog),
```

## Design notes

- Ships as a **secondary entry point**, so the main entry stays free of `@a2ui` imports (verified against the built FESM). `@a2ui/angular` and `@a2ui/web_core` are **optional peer dependencies**: only apps that import `@copilotkit/angular/a2ui` need to install them. This matters because `@a2ui/angular` requires Angular ≥ 21.2 while the main entry keeps supporting Angular 19–21.
- `zod` moves from devDependencies to dependencies. It types the catalog descriptors and validates activity content; it was already an implicit runtime requirement via `zod-to-json-schema`.

## Testing

- 13 new specs: surface rendering via `ngComponentOutlet` (including the second-surface-append regression), rendering a custom Angular component emitted by the agent, catalog/function registration incl. argument validation, and action-handler routing, injection context, and unsubscribe-on-destroy.
- `vitest run` (29 files, 191 tests), `tsc --noEmit`, `ng-packagr` build (both entry points), and `publint` are green.

### DIFF
--- a/packages/angular/a2ui/ng-package.json
+++ b/packages/angular/a2ui/ng-package.json
@@ -1,0 +1,6 @@
+{
+  "$schema": "../../../node_modules/ng-packagr/ng-package.schema.json",
+  "lib": {
+    "entryFile": "../src/a2ui/index.ts"
+  }
+}
--- a/packages/angular/package.json
+++ b/packages/angular/package.json
@@ -19,6 +19,12 @@
         "default": "./dist/fesm2022/copilotkit-angular.mjs"
       }
     },
+    "./a2ui": {
+      "import": {
+        "types": "./dist/types/copilotkit-angular-a2ui.d.ts",
+        "default": "./dist/fesm2022/copilotkit-angular-a2ui.mjs"
+      }
+    },
     "./styles.css": "./dist/styles.css"
   },
   "publishConfig": {
@@ -51,9 +57,12 @@
     "rxjs": "^7.8.1",
     "tailwind-merge": "^2.6.0",
     "tslib": "^2.6.0",
+    "zod": "^3.25.76",
     "zod-to-json-schema": "^3.24.5"
   },
   "devDependencies": {
+    "@a2ui/angular": "^0.9.1",
+    "@a2ui/web_core": "^0.9.2",
     "@analogjs/vite-plugin-angular": "^1.20.2",
     "@analogjs/vitest-angular": "^1.20.2",
     "@angular/cdk": "^21.2.13",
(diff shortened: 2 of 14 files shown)


wolfmanfx:
@manfredsteyer Thx for the PR! We will take a look and see how it fits into the roadmap. Right now we will focus on other features first on the angular part. I will ping you on that.

manfredsteyer:
Thanks @wolfmanfx. Where can I find the roadmap?

lukasmoschitz:
Closing this one in favour of #6073, @manfredsteyer — not rejecting the idea, consolidating it.

#6073 builds directly on this branch's commit (0ca5083d6f is an ancestor of it) and its file set is a strict superset, so everything here lives on there. Keeping both open just splits the same discussion.

That discussion is still open, and it's an architecture call rather than a code review: `main` already has an A2UI Angular renderer at `packages/angular/src/lib/components/a2ui/` going through `@copilotkit/a2ui-renderer/web-components`, and your approach goes through `@a2ui/angular` instead. Whether we maintain both paths is the question to settle — and it belongs on #6073.

Also: you asked @wolfmanfx where to find the roadmap back in July and never got an answer. Sorry about that. Follow #6073 for this thread.

rainerhahnekamp:
@lukasmoschitz All good from my side.

Just to clarify the roadmap question: we discussed this internally with @manfredsteyer on August 12 but can share it here as well. It was:
- Fix issues that surfaced while bringing the Angular integration to feature parity.
- Review existing features to make sure they fit well with Angular conventions and patterns.
- Ensure that we maintain feature parity with React over the long term.

The “Angular way” part is particularly important to us. On the one hand, we want CopilotKit to feel as consistent as possible across frameworks. On the other hand, Angular developers shouldn’t feel like they are using a React library wrapped for Angular.

Because these architectural questions often need broader consideration (i.e. takes longer), we are currently prioritizing the more implementation-specific issues, where improvements can be made independently of those larger design decisions. Important to mention #6714 here as well.

One of the bigger architectural topics is resource integration. There is no direct equivalent in React, while resources are a first-class concept in Angular, so this deserves some dedicated thought.

We’ll very likely add a ROADMAP.md as well, both to make these priorities more transparent and to open the roadmap up to input from the community.

manfredsteyer:
@lukasmoschitz: Thanks for the update.

The main reason I implemented the Angular-native renderer for A2UI was to enable custom catalogs built directly on top of Angular components. I think this is an important use case.

@rainerhahnekamp: A roadmap.md would be great to give contributors a clearer idea of the current priorities and where their contributions would be most helpful.

PR #6073, opened 2026-07-20
feat(angular): typed custom-catalog helpers + agent metadata derivation for A2UI + Angular-native A2UI Renderer

manfredsteyer:
> **Stacked on #6072** — please review only the last commit here until #6072 is merged; the first commit is that PR.

## What

Extends the `@copilotkit/angular/a2ui` entry point (introduced in #6072) with the authoring side of custom catalogs:

**Typed catalog definition**
- `binding(schema)` — wraps a prop schema in a union with a path-binding schema, so the agent can pass a literal (`"Paris"`) or a data-model binding (`{ path: "/flight/to" }`) for every prop
- `ContextFromSchema<Schema>` — derives the shape of a custom component's `props` input (each prop as a `BoundProperty`) from its Zod schema
- `createCustomComponent` / `createCustomFunction` / `createCustomCatalog` — identity helpers with literal type inference that keep name, description, schema, and implementation together

**Metadata / schema derivation for the agent**
- `catalogToContextEntry` / `catalogIdToContextEntry` — serialize a catalog into an AG-UI context entry (`A2UI_CATALOG_CONTEXT_DESCRIPTION`), with prop schemas emitted as inline JSON Schema (`$refStrategy: "none"`)
- `provideA2UIAngularRenderer(catalog)` now registers that context entry automatically and removes it when the providing injector is destroyed. `sendCatalogDescription: false` restricts the entry to the catalog id — for production setups where the server resolves the trusted catalog descriptor from its own registry instead of trusting client-supplied metadata.

## Why

Custom catalogs only work when two sides stay in sync: the renderer must know the Angular component behind each catalog name, and the agent must know which components exist and which props they accept. Without helpers, both halves are stringly typed and drift silently. With this PR the Zod schema is the single source of truth: it types the component's `props` input at compile time, validates at runtime, and is what the agent sees as JSON Schema:

```ts
const ticketWidget = createCustomComponent({
  name: "TicketWidget",
  description: "Boarding-pass style widget for a booked flight.",
  component: TicketWidget,
  schema: z.object({
    from: binding(z.string()),
    to: binding(z.string()),
    delay: binding(z.number()).optional(),
  }).strict(),
});

provideA2UIAngularRenderer(createCustomCatalog({
  id: "https://example.com/catalogs/flights",
  components: [ticketWidget],
}));
```

## Testing

- 14 new specs: `binding` parsing (literal / path / strict rejection), identity helpers incl. execute-arg inference, context-entry serialization (inline schemas, no `$ref`, empty-catalog case), automatic context registration, id-only mode, no-catalog case, and removal on injector destroy.
- `vitest run` (31 files, 205 tests), `tsc --noEmit`, and the two-entry-point `ng-packagr` build are green.

### DIFF
--- a/packages/angular/a2ui/ng-package.json
+++ b/packages/angular/a2ui/ng-package.json
@@ -1,0 +1,6 @@
+{
+  "$schema": "../../../node_modules/ng-packagr/ng-package.schema.json",
+  "lib": {
+    "entryFile": "../src/a2ui/index.ts"
+  }
+}
--- a/packages/angular/package.json
+++ b/packages/angular/package.json
@@ -19,6 +19,12 @@
         "default": "./dist/fesm2022/copilotkit-angular.mjs"
       }
     },
+    "./a2ui": {
+      "import": {
+        "types": "./dist/types/copilotkit-angular-a2ui.d.ts",
+        "default": "./dist/fesm2022/copilotkit-angular-a2ui.mjs"
+      }
+    },
     "./styles.css": "./dist/styles.css"
   },
   "publishConfig": {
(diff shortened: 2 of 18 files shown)


manfredsteyer:
Short update:
* As mentioned in #6072, the Angular-native A2UI renderer allows us to directly implement custom components with Angular
* If we decide to move on with this PR (the ideas here), I should also use registerComponent which was recently added with CopilotKit for Angular 0.5.0.

PR #6080, opened 2026-07-20
Feat/height measurement cleanup

wolfmanfx:
Fixes two layout bugs in the Angular chat view:

1. **Floating-input height stuck at `0`.** Measurement ran once in `ngAfterViewInit`,
   but on the welcome screen the input overlay isn't in the DOM yet — retries
   expired and the height never recovered. After the first message, the list
   rendered under the floating input and the scroll-to-bottom button overlapped it.
2. **Scroll wrapper sized to the viewport** (`h-[calc(100vh-9rem)]`), which
   overshoots when the chat is embedded in a smaller panel.

Changes:
- New **`CopilotChatViewInputMeasure`** directive on the input slot — its lifecycle
  mirrors the overlay, so it re-measures on the welcome → chat transition.
  `afterNextRender` + `ResizeObserver`, SSR-safe, height exposed as a signal.
- Scroll wrapper → `h-full max-h-full` so it tracks its container.

Regression tests cover both fixes.

## Checklist

- [ X ] I have read the [Contribution Guide](https://github.com/copilotkit/copilotkit/blob/master/CONTRIBUTING.md)
- [ X ] If the PR changes or adds functionality, I have updated the relevant documentation
- [ X ] "Allow edits by maintainers" is checked (lets us help iterate on your PR directly — faster turnaround for everyone)

### DIFF
--- a/examples/v2/runtime/node/src/index.ts
+++ b/examples/v2/runtime/node/src/index.ts
@@ -8,9 +8,7 @@ import { createCopilotNodeHandler } from "@copilotkit/runtime/v2/node";
 
 const runtime = new CopilotRuntime({
   agents: {
-    default: new BuiltInAgent({
-      model: process.env.MODEL ?? "openai/gpt-5-mini",
-    }),
+    default: new BuiltInAgent({ model: "openai/gpt-5-mini" }),
   },
 });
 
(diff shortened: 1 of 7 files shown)


examples/v2/runtime/node/src/index.ts
```diff
@@ -8,7 +8,9 @@ import { createCopilotNodeHandler } from "@copilotkit/runtime/v2/node";
 
 const runtime = new CopilotRuntime({
   agents: {
-    default: new BuiltInAgent({ model: "openai/gpt-5-mini" }),
+    default: new BuiltInAgent({
```
rainerhahnekamp: Shouldn't we do this in a separate PR?

rainerhahnekamp (changes requested):

rainerhahnekamp (approved):

PR #6097, opened 2026-07-21
feat(examples): add Angular + ADK integration example (adk-angular) [OSS-561]

BenTaylorDev:
## What & why

Adds the first **Angular** scaffoldable integration example: `examples/integrations/adk-angular/` — an Angular 21 SPA wired to a Google ADK agent over AG-UI, a peer to the React `examples/integrations/adk`. Angular support is selective (ADK first), so this surfaces as a new **"Angular + ADK"** entry in `copilotkit init` (CLI wiring is a companion PR in the Intelligence repo).

## Architecture — 3 processes

Angular is a SPA, so (unlike the Next.js React example, which hosts the runtime in an API route) the runtime lives in a standalone Node server — the documented `@copilotkit/angular` pattern.

| Process | Port | Source |
| --- | --- | --- |
| Angular app (`ng serve`) | 4200 | `src/` |
| Copilot Runtime (`tsx server.ts`) | 8200 | `server.ts` (`createCopilotNodeListener`) |
| Python ADK agent (`uv`) | 8000 | `agent/` — copied from `examples/integrations/adk` |

One `npm run dev` runs all three via `concurrently`.

## Feature parity with the React ADK example

- Shared agent state (proverbs) via `injectAgentStore` + `agent.setState`
- Generative UI (weather card) via `renderToolCalls`
- Frontend tool (`setThemeColor`) via `registerFrontendTool`
- Static suggestions (the same 5)
- Threads drawer + env-gated managed Intelligence (`CopilotKitIntelligence` when `COPILOTKIT_LICENSE_TOKEN` is set, else `InMemoryAgentRunner` — matching the React `route.ts` gating)
- Dev web inspector (`cpk-web-inspector`), matching the richer React examples

### Two behaviors that required real fixes (found via live testing, not static review)

1. **Frontend tools now reach the ADK agent.** `setThemeColor` silently failed because the ADK agent's tools were static and `ag_ui_adk` only forwards client tools when the agent includes an `AGUIToolset()` placeholder. Added it (this affected the React `adk` example identically — fixed there in a companion PR).
2. **The weather generative-UI card now renders.** A `renderToolCalls` entry scoped with `agentId` filtered the server-side tool call out; removed it.

## Chat UI — floating, React-`CopilotSidebar`-style

Angular has no `CopilotSidebar`, so the chat is a hand-rolled collapsible panel that mirrors React's behavior:

[...]

PR #6110, opened 2026-07-22
fix(examples): adk-angular a11y (closed chat inert) + dev inspector out of prod bundle

BenTaylorDev:
Two issues found in review after `examples/integrations/adk-angular` merged (#6097).

## 1. Closed chat stayed keyboard-accessible off-screen
The collapsible chat panel is only translated off-screen (`transform: translateX(100%)`), so its `copilot-chat` inputs/buttons remained in the tab order and the accessibility tree while "closed."

**Fix:** mark the `<aside>` `inert` while closed (`[attr.inert]="chatOpen() ? null : ''"`); removed when open.

## 2. Dev inspector shipped in the production bundle
The `cpk-web-inspector` dev aid (and its `@copilotkit/web-inspector` dependency, ~660 kB) was statically imported, so it landed in the **4.59 MB** production initial bundle.

**Fix:** gate it behind `@defer (when isDev)` (`isDev = isDevMode()`), splitting it into a lazy chunk that a production build (`isDev` = false) never loads.

## Testing (live)

- **Prod build** (`ng build`): initial bundle **4.59 MB → 3.93 MB**; `@copilotkit/web-inspector` is now a separate lazy chunk (660 kB) that prod never loads.
- **Dev serve**: the inspector still mounts (`@defer` triggers when `isDev`).
- **a11y**: with the chat closed, focusing a control inside the panel is blocked (`document.activeElement` falls back to `body`); open panel unaffected; the toggle FAB still reopens it.
- `oxlint` 0/0.

Follow-up to OSS-561.

### DIFF
--- a/examples/integrations/adk-angular/src/app/app.ts
+++ b/examples/integrations/adk-angular/src/app/app.ts
@@ -1,4 +1,4 @@
-import { Component, signal } from "@angular/core";
+import { Component, isDevMode, signal } from "@angular/core";
 import {
   CopilotChat,
   CopilotThreadsDrawer,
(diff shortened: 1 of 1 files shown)


MikeRyanDev (approved):
Reviewed the Angular template change and verified the production build. The inspector is emitted as a lazy 660.56 kB chunk and is not requested on a production page load. In Chrome, the closed chat has inert and blocks focus; opening it removes inert and restores focus. No blocking issues found.

PR #6118, opened 2026-07-23
fix(runtime): bump uuid off deprecated v10.0.0

adrienplg:
## Summary

`packages/runtime` declares its own `"uuid": "^10.0.0"` dependency, but nothing in the package's source actually imports it directly — id generation in `@copilotkit/runtime` goes through `randomUUID()` re-exported from `@copilotkit/shared`, which already depends on `uuid@^11.1.0`. The unused v10 pin just adds an npm deprecation warning for every consumer installing `@copilotkit/runtime`:

```
npm warn deprecated uuid@10.0.0: uuid@10 and below is no longer supported. For ESM codebases, update to uuid@latest. For CommonJS codebases, use uuid@11 (but be aware this version will likely be deprecated in 2028).
```

This bumps it to `^11.1.0` to match `@copilotkit/shared` and clears the warning.

## Test plan

- [x] `pnpm --filter "@copilotkit/runtime^..." run build` — all workspace dependencies build cleanly
- [x] `pnpm --filter @copilotkit/runtime run check-types` — no type errors
- [x] `pnpm --filter @copilotkit/runtime run test` — 126 test files / 1746 tests passing
- [x] Confirmed no file in `packages/runtime/src` imports `uuid` directly (grepped for both `from "uuid"` / `from 'uuid'` and `require("uuid")` — zero matches)
- [x] Confirmed `pnpm-lock.yaml` now resolves `uuid@11.1.0` for this dependency, which is not on npm's deprecated-versions list


BenTaylorDev (changes requested):
Thanks — the intent here is right. `uuid@10` is deprecated, and bumping runtime to `^11.1.0` actually aligns it with `@copilotkit/shared`, which is already on `^11.1.0`, so this reduces version drift across the monorepo. runtime doesn't import `uuid` in its source and v11 ships its own types, so there's no API risk.

One thing to fix before this can merge: the `pnpm-lock.yaml` diff isn't scoped to `uuid`. It also downgrades `@babel/core` (`7.29.0` → `7.28.5`) and drops `babel-plugin-macros` resolutions across a number of entries — that looks like the lockfile was regenerated on a branch that's now behind `main`, so it's pulling in unrelated (and backwards) dependency movement.

Could you rebase onto the latest `main` and regenerate the lockfile so the only change is the `uuid` bump? Once the lockfile diff is uuid-only this is good to go.

adrienplg:
The lockfile now only touches `uuid`

BenTaylorDev (approved):
Confirmed — the lockfile diff is now uuid-only. The whole PR is `packages/runtime/package.json` +1/−1 and two lines in `pnpm-lock.yaml`; the `@babel/core` downgrade and `babel-plugin-macros` churn are gone. Thanks for the rebase.

Verified locally on this branch:

- `pnpm install --frozen-lockfile` succeeds, so the lockfile is consistent with the manifests.
- `@copilotkit/runtime` and `@copilotkit/shared` both resolve `uuid@11.1.0` — the version drift between them is closed and no `uuid@10` remains in the install.
- `@copilotkit/runtime` and its workspace dependencies build clean, declarations included.
- Runtime test suite: 135 files / 1983 tests passing.
- The built `dist` contains no reference to `uuid` at all, which confirms the dependency was unused — so the v11 major is a no-op for consumers beyond dropping the deprecation warning.

Full CI is also green on this head.

Approving.

PR #6120, opened 2026-07-23
feat(runtime): add agentId to AgentRunnerConnectRequest

### LINKED ISSUE #5911: 🚀 Feature Request: Add agent parameter in AgentRunner connect() function
related to https://github.com/CopilotKit/CopilotKit/pull/3173  and https://github.com/CopilotKit/CopilotKit/issues/3553

### Problem or Motivation

Hello, AgentRunner  class should include the agent in connect function params for hydrating messages when cache is empty 
```ts
declare abstract class AgentRunner {
  readonly ɵsupportsLocalThreadEndpoints?: boolean;
  abstract run(request: AgentRunnerRunRequest): Observable<BaseEvent>;
  abstract connect(request: AgentRunnerConnectRequest): Observable<BaseEvent>;
  abstract isRunning(request: AgentRunnerIsRunningRequest): Promise<boolean>;
  abstract stop(request: AgentRunnerStopRequest): Promise<boolean | undefined>;
}
```
thanks 

### 🎤 Tell us your idea

add agentId optional in AgentRunnerConnectRequest

- [x] Yes, I’d love to work on it!

lesbass:
Closes #5911

Adds an optional `agentId` field to `AgentRunnerConnectRequest` so custom agent runners can hydrate messages when the cache is empty. Also passes the already-available `agentId` in `handleSseConnect` through to `runner.connect()`.

## Changes

- `packages/runtime/src/v2/runtime/runner/agent-runner.ts`: Added `agentId?: string` to `AgentRunnerConnectRequest` interface
- `packages/runtime/src/v2/runtime/handlers/sse/connect.ts`: Pass `agentId` to `runner.connect()`

## Verification

- Runtime package builds successfully (`pnpm --filter @copilotkit/runtime build`)
- The change is purely additive — the field is optional and does not affect existing runners

### DIFF
--- a/packages/runtime/src/v2/runtime/handlers/sse/__tests__/sse-connect-agent-id.test.ts
+++ b/packages/runtime/src/v2/runtime/handlers/sse/__tests__/sse-connect-agent-id.test.ts
@@ -79,6 +79,46 @@ describe("handleSseConnect → debug envelope agentId", () => {
 });
 
 /**
+ * Regression guard: `handleSseConnect` must forward the route-resolved
+ * `agentId` into the `runner.connect` request so custom runners can
+ * hydrate messages from it when the cache is empty.
+ */
+describe("handleSseConnect → agentId in runner.connect call", () => {
+  it("forwards agentId into the runner.connect request", async () => {
+    const event: BaseEvent = {
+      type: EventType.RUN_STARTED,
+      threadId: "t-1",
+      runId: "r-1",
+    } as BaseEvent;
+
+    const connectSpy = vi.fn(() => of(event));
+    const fakeRuntime = {
+      debugEventBus: new DebugEventBus(),
+      forwardHeadersPolicy: defaultPolicy,
+      runner: { connect: connectSpy },
+    } as any;
+
+    const response = handleSseConnect({
+      runtime: fakeRuntime,
+      request: new Request("http://localhost/agent/weather-agent/connect", {
+        method: "POST",
+      }),
+      agentId: "weather-agent",
+      threadId: "t-1",
+    });
+
+    const reader = response.body!.getReader();
+    while (true) {
+      const { done } = await reader.read();
+      if (done) break;
+    }
+
+    expect(connectSpy).toHaveBeenCalledTimes(1);
+    expect(connectSpy.mock.calls[0][0].agentId).toBe("weather-agent");
+  });
+});
+
+/**
  * Regression guard for issue #5712 on the /connect path. `handleSseConnect`
  * used to forward raw inbound `authorization`/`x-*` headers straight to
  * `runner.connect`, ignoring server-configured `agent.headers` entirely — so
--- a/packages/runtime/src/v2/runtime/handlers/sse/__tests__/sse-connect-agent-id.test.ts
+++ b/packages/runtime/src/v2/runtime/handlers/sse/__tests__/sse-connect-agent-id.test.ts
@@ -15,6 +15,7 @@ vi.mock("../../../telemetry", () => ({
 }));
 
 import { handleSseConnect } from "../connect";
+import type { AgentRunnerConnectRequest } from "../../../runner/agent-runner";
 import { DebugEventBus } from "../../../core/debug-event-bus";
 import { resolveForwardHeadersPolicy } from "../../header-utils";
 import type { ResolvedForwardHeadersPolicy } from "../../header-utils";
@@ -91,7 +92,7 @@ describe("handleSseConnect → agentId in runner.connect call", () => {
       runId: "r-1",
     } as BaseEvent;
 
-    const connectSpy = vi.fn(() => of(event));
+    const connectSpy = vi.fn((_req: AgentRunnerConnectRequest) => of(event));
     const fakeRuntime = {
       debugEventBus: new DebugEventBus(),
       forwardHeadersPolicy: defaultPolicy,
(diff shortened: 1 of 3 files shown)


BenTaylorDev (changes requested):
Small, clean addition and the shape is right (optional, matching the existing forward-looking `headers` field on `AgentRunnerConnectRequest`). One ask before merge: **please add a test that asserts `agentId` is forwarded into the `runner.connect` request.** It's new API surface on the connect request with no coverage — the parallel `headers` plumbing on this same interface is tested (`handle-connect.test.ts`, and the header-precedence cases in `sse-connect-agent-id.test.ts`), so this is the one gap. `sse-connect-agent-id.test.ts` is the natural home — it already owns the connect→agentId behavior and spies on `runner.connect`. Same rationale as the agentId-in-envelope guard already in that file: the /run path wouldn't catch a /connect regression. Everything else looks good to merge.

lesbass:
Added a test in `sse-connect-agent-id.test.ts` that asserts `agentId` is forwarded into the `runner.connect` request, following the same spy pattern as the existing header precedence tests.

The test registers a spy on `runner.connect`, calls `handleSseConnect` with `agentId: "weather-agent"`, and asserts `connectSpy.mock.calls[0][0].agentId` equals `"weather-agent"`.

All 6 tests in the file pass (the 5 existing plus the new one).

BenTaylorDev (changes requested):
Thanks — this is exactly the test I was asking for, and it exercises the real `handleSseConnect`→`runner.connect` path. One thing before it can land: it doesn't type-check. The new `connectSpy` is declared `vi.fn(() => of(event))` (untyped args), so `connectSpy.mock.calls[0][0]` is an empty-tuple index and `tsc --noEmit` fails on the new assertion (`sse-connect-agent-id.test.ts:117` — TS2493/TS2532). The existing header-precedence tests in this same file avoid this by typing the spy param — do the same here, e.g.:

```ts
const connectSpy = vi.fn((_req: AgentRunnerConnectRequest) => of(event));
```

(or inline `{ threadId: string; agentId?: string }`), and `mock.calls[0][0].agentId` will type-check. With that it's good to go — the runtime `check-types` gate includes test files, so it needs to pass there.

lesbass (commented):
Fixed the type-check issue: added `AgentRunnerConnectRequest` import and typed the `connectSpy` parameter as `(_req: AgentRunnerConnectRequest)`, matching the pattern used by the header-precedence tests in the same file.

Tests pass (6/6) and `tsc --noEmit` is clean for the file.

lghinet:
should you add ` agentId?: string to AgentRunnerConnectRequest`  or ` agent?: AbstractAgent` given the fact that where AgentRunnerConnectRequest is used agent instance is already available

lesbass:
Great question. We chose `agentId?: string` deliberately: `AgentRunnerConnectRequest` is a serializable, wire-level DTO (it travels across process/network boundaries as part of the connect handshake), so embedding the live `AbstractAgent` instance would couple the request shape to an in-memory object graph and break that contract. The `agentId` lets the receiving side resolve the agent through its own registry while keeping the request self-contained — the same forward-looking pattern as the existing `headers` field on this interface.

That said, if the team prefers passing the resolved `agent` at the call site instead, we're happy to adjust — the current change keeps it optional precisely so the shape stays flexible. Thanks for the review!

lesbass:
Hi @BenTaylorDev, just a gentle follow-up. We've addressed both requests:

1. Added the test that asserts  is forwarded into  (see ).
2. Fixed the type-check by typing the spy parameter as , matching the existing pattern.

Both tests pass and  is clean. Could you please re-review when you have a chance? Thanks!

BenTaylorDev (approved):
Re-reviewed against current `main` — both earlier asks are addressed and this is good to merge.

**What I verified locally** (this is a fork PR, so only CodeRabbit reported on it; the unit workflow never ran). Merged the branch onto current `main` in a scratch worktree — it applies cleanly, and the merged tree differs from `main` by exactly these three files:

- `packages/runtime` tests: 200/200 pass, including the new sixth case in `sse-connect-agent-id.test.ts`.
- The new test is a real guard, not a vacuous one — deleting `agentId,` from `connect.ts:49` makes it fail (`received: undefined`).
- `tsc --noEmit` over `packages/runtime` (which includes test files) reports zero errors in the three changed files; the `TS2493`/`TS2532` I flagged at `sse-connect-agent-id.test.ts:117` is gone. Typing the spy param was the right fix.
- oxlint clean.
- `handlers/sse/connect.ts:47` is the only production `runner.connect` call site, so the plumbing is complete — the Intelligence branch already passes `agentId` through to `ɵconnectThread` (`handlers/intelligence/connect.ts:50`). No shipped runner spreads the request object, so the extra field can't leak anywhere: in-memory and intelligence both destructure `threadId` only, and the telemetry runner delegates `...args`.

`AgentRunnerConnectRequest` is public surface (it ships in `dist/v2/index.d.cts` via the `@copilotkit/runtime/v2` entry), and since it's only ever a parameter type, the optional field is non-breaking for external runner implementations.

One correction for the record, since it may come up again: `AgentRunnerConnectRequest` isn't a serializable wire-level DTO — its sibling `AgentRunnerRunRequest` embeds `agent: AbstractAgent` directly (`agent-runner.ts:11`), and the resolved per-request clone is already in scope at this call site. So that argument for `agentId?: string` over `agent?: AbstractAgent` doesn't hold. The choice is still the right one here: it's what #5911 asked for, it's minimal, and `agent?:` remains addable later without a break.

Thanks for the two clean turnarounds on the review rounds.

PR #6121, opened 2026-07-23
fix(react-core): forward credentials to provisional agents

### LINKED ISSUE #6116: 🐛 Bug: Provisional agent from `useAgent` connects without `credentials`, causing 403 on cross-origin runtimes
### ♻️ Reproduction Steps

1. Serve a CopilotKit app from origin A (e.g. `https://app.example.com`).
2. Set `runtimeUrl` on `<CopilotKit>` to an absolute or relative URL that resolves to origin B (cross-origin, e.g. `https://api.example.com/api/copilotkit`).
3. Set `credentials="include"` on `<CopilotKit>`.
4. Configure the runtime (or an auth gateway in front of it) to require cookies for authenticated `agent/connect` requests.
5. Load the page and open the browser network tab.

### ✅ Expected Behavior

Every request `<CopilotKit>` fires to `runtimeUrl` — including the first `agent/connect` triggered by `CopilotChat`'s connect effect — should carry cookies, because the app declared `credentials="include"`. The response for both `POST /runtime` (body `{method:"info"}`) and the follow-up `POST /runtime` (body `{method:"agent/connect", ...}`) should be 200.

### ❌ Actual Behavior

Three requests fire on page load:

- `POST /runtime` (body `{method:"info"}`) → **200** — uses `core.credentials` correctly, cookies present.
- `POST /runtime` (body `{method:"agent/connect", ...}`) → **403** — no cookies attached; rejected by the auth gateway before it reaches the runtime.
- `POST /runtime` (body `{method:"agent/connect", ...}`) → **200** — retry with the real agent, cookies present.

The first `agent/connect` request has no `Cookie` header; the retry does. CopilotKit's own error subscriber system fires `AGENT_CONNECT_FAILED` for the failed one, which any consumer surfacing coded errors from `onError` will treat as a user-visible error and (in our case) put the input into an unrecoverable state.

### Screenshots

_No response_

### 𝌚 CopilotKit Version

```shell
- `@copilotkit/react-core`: **1.62.2** (bug also verified on `main` at time of report; latest published is 1.63.1)
- `@copilotkit/core`: **1.62.2**
- Browser: any (reproduced in Chrome and Safari)
```

### 📄 Logs (Optional)

```shell
In `packages/react-core/src/v2/hooks/use-agent.tsx`, the two provisional-agent constructors don't pass `credentials`:


// lines ~114-118 and ~140-144 on main
const provisional = new ProxiedCopilotRuntimeAgent({
  runtimeUrl: copilotkit.runtimeUrl,
  agentId: resolvedAgentId,
  transport: copilotkit.runtimeTransport,
  runtimeMode: "pending",
  // credentials: MISSING
});
copilotkit.applyHeadersToAgent(provisional);
// no equivalent applyCredentialsToAgent(provisional)


The "real" agent constructor in `packages/core` (built in `packages/core/dist/index.mjs:1157`) *does* pass `credentials: this.core.credentials`, and `CopilotKitCore` even has an `applyCredentialsToAgent(agent)` helper — but it's only called from `updateRuntimeConnection`, never from the `useAgent` provisional path.

The immediate `useEffect` in `CopilotChat` (`dist/copilotkit-lHheGQWt.cjs:8043` in 1.62.2) unconditionally fires `copilotkit.connectAgent({ agent })` against whichever agent `useAgent` returns — so on the first render, that's the credential-less provisional. `fetch` then falls back to its default `same-origin` credentials mode, cookies are omitted, and the cross-origin request is rejected.
```

shixi-li:
## What does this PR do?

Provisional agents returned by `useAgent` now receive the provider's `credentials` setting when they are created. Cached provisional agents also refresh that setting when the provider configuration changes, while keeping the same agent instance.

The regression test covers both the initial value and a later `include` to `omit` update on the cached provisional agent.

### Verification

- `pnpm nx run @copilotkit/react-core:test`
- `pnpm nx run @copilotkit/react-core:check-types`
- `pnpm nx run @copilotkit/react-core:build`
- `pnpm check-format`

## Related PRs and Issues

- Closes https://github.com/CopilotKit/CopilotKit/issues/6116

## Checklist

### DIFF
--- a/packages/react-core/src/v2/hooks/__tests__/use-agent-stability.test.tsx
+++ b/packages/react-core/src/v2/hooks/__tests__/use-agent-stability.test.tsx
@@ -5,7 +5,10 @@ import type { AbstractAgent } from "@ag-ui/client";
 import { useCopilotKit } from "../../context";
 import { MockStepwiseAgent } from "../../__tests__/utils/test-helpers";
 import { useAgent } from "../use-agent";
-import { CopilotKitCoreRuntimeConnectionStatus } from "@copilotkit/core";
+import {
+  CopilotKitCoreRuntimeConnectionStatus,
+  ProxiedCopilotRuntimeAgent,
+} from "@copilotkit/core";
 
 // Mock the CopilotKit context to control copilotkit state directly
 vi.mock("../../context", () => ({
@@ -27,6 +30,7 @@ describe("useAgent stability during runtime connection", () => {
     runtimeUrl: string | undefined;
     runtimeConnectionStatus: CopilotKitCoreRuntimeConnectionStatus;
     runtimeTransport: string;
+    credentials: RequestCredentials | undefined;
     headers: Record<string, string>;
     agents: Record<string, AbstractAgent>;
     applyHeadersToAgent: (agent: AbstractAgent) => void;
@@ -43,6 +47,7 @@ describe("useAgent stability during runtime connection", () => {
       runtimeConnectionStatus:
         CopilotKitCoreRuntimeConnectionStatus.Disconnected,
       runtimeTransport: "rest",
+      credentials: undefined,
       headers: {},
       agents: {},
       // Additive stand-in for core's merge (core headers ON TOP of the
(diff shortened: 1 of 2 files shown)


BenTaylorDev (changes requested):
Mechanism looks correct and the provisional path now matches the real-agent path (`agent-registry.ts:317` + `applyCredentialsToAgent`). One change before merge:

**Apply credentials in the header-freshness effect, not inside `useMemo`.** This hook deliberately keeps HttpAgent headers current in the `useEffect` at `use-agent.tsx:237`, with the comment explaining why: *"Keep HttpAgent headers fresh without mutating inside useMemo, which is unsafe in concurrent mode (React may invoke useMemo multiple times and discard intermediate results, but mutations always land)."* Setting `cached.credentials` inside the `useMemo` (lines 110 / 140) reintroduces exactly the pattern that comment warns against. It happens to be idempotent today, but it's the same class of concurrent-mode footgun the effect was added to avoid.

Please move the credentials sync alongside the headers sync in that effect, e.g.:

```ts
useEffect(() => {
  if (agent instanceof HttpAgent) {
    copilotkit.applyHeadersToAgent(agent);
    if (isCredentialedAgent(agent)) {
      agent.credentials = copilotkit.credentials;
    }
  }
  // eslint-disable-next-line react-hooks/exhaustive-deps
}, [agent, JSON.stringify(copilotkit.headers), copilotkit.credentials]);
```

Then drop the two `cached.credentials = copilotkit.credentials` mutations from the `useMemo` (the constructor `credentials:` for the fresh-provisional case stays). That keeps all live request-setting sync in one concurrent-safe place. Everything else — the constructor plumbing and the regression test — looks good.

shixi-li:
Addressed the review feedback in `50cf16caf`:

- moved cached provisional credential updates out of `useMemo` and into the existing commit-phase request-settings effect
- limited live credential updates to `ProxiedCopilotRuntimeAgent`, matching the core registry guard
- added `copilotkit.credentials` to the effect dependencies while retaining constructor-time initialization for new provisional agents

Validation:
- `use-agent-stability.test.tsx`: 5/5 passed
- `@copilotkit/react-core:check-types` (including 33 dependencies): passed
- binary, formatting, lint, and commit-message checks: passed
- the affected-package run reached 117 files / 1,437 tests passed before an unrelated A2UI test hit its 5s timeout under concurrency; that A2UI file then passed in isolation (8/8, 1.68s)

@BenTaylorDev, would you mind taking another look?

BenTaylorDev (approved):
**Re-approving** — the credentials sync is moved out of the `useMemo` into the post-commit effect (`use-agent.tsx:239-240`, gated on `instanceof ProxiedCopilotRuntimeAgent`, with `copilotkit.credentials` in the deps), and the constructor `credentials:` is kept for the fresh-provisional path — the concurrent-mode concern is resolved. Verified locally on the updated branch against current `main`: `use-agent-stability` 5/5 (the new test exercises the effect path), lint 0 errors, no type errors in the changed file.

PR #6141, opened 2026-07-23
fix(react-core): honor threadId prop on useAgent

davidmckayv:
## Problem

`UseAgentProps` (`packages/react-core/src/v2/hooks/use-agent.tsx`) exposed only `{ agentId?, updates?, throttleMs? }` — there was **no `threadId`**. Yet the shipped V2 React Native demo calls:

```ts
const { agent } = useAgent({ agentId: "default", threadId });
```

(`examples/v2/react-native/demo/src/ChatScreen.tsx`)

Because the prop didn't exist on the type, it was **silently ignored**. The demo's `threadId` state — and its "New Chat" reset (`setThreadId(generateThreadId())`) — never reached the agent. The demo has no `<CopilotChatConfigurationProvider>` in its tree (see `App.tsx`), which was the *only* source the hook read threadId from, so the agent ran under its own auto-minted UUID the whole time. A misleading API: the demo advertised a supported prop that did nothing.

## Fix

Accept an optional `threadId` on `UseAgentProps` and honor it. Resolution precedence:

1. An explicit `threadId` prop wins — lets headless callers (e.g. React Native) scope the run to a thread with no chat-configuration provider in the tree.
2. Otherwise fall back to the surrounding chat configuration's threadId, gated on `hasExplicitThreadId` exactly as before.

The threadId still lands on `agent.threadId`, which `ProxiedCopilotRuntimeAgent` uses to address `/agent/run`, `/agent/connect`, and `/agent/stop`.

## Backward compatibility

Fully backward compatible: `threadId` is optional, and when omitted the behavior is byte-for-byte unchanged (config-sourced, `hasExplicitThreadId`-gated). No demo change is needed — the RN demo's existing `useAgent({ agentId, threadId })` call now works as it always appeared to.

## Tests

Extends the existing threadId-propagation contract test (`packages/react-core/src/__tests__/threadid-propagation.contract.test.tsx`) with the prop path:

- honors an explicit `threadId` prop with **no** `CopilotChatConfigurationProvider` (the RN demo scenario)
- prefers the `threadId` prop over the chat configuration's threadId
- re-syncs `agent.threadId` when the prop changes

The pre-existing config-sourced invariants remain covered and unchanged.


packages/react-core/src/v2/hooks/use-agent.tsx
```diff
@@ -54,7 +67,12 @@ export interface UseAgentProps {
   throttleMs?: number;
 }
 
-export function useAgent({ agentId, updates, throttleMs }: UseAgentProps = {}) {
+export function useAgent({
+  agentId,
+  threadId,
+  updates,
+  throttleMs,
+}: UseAgentProps = {}) {
```
mme, line 148: This introduces an unexpected behavior when dealing with two agents with different `threadIds`. For example:

```
const { agent: agent1 } = useAgent({ agentId: "myAgent",threadId: "threadId-1" });
const { agent: agent2 } = useAgent({ agentId: "myAgent",threadId: "threadId-2" });
```

Now each time the `useAgent` hook executes, the agent called `"myAgent"` would first be set to `"threadId-1"` then to `"threadId-2"` because the agent registered under a specific `agentId` is a singleton.

The solution CopilotKit offers is to route agents registered on the frontend to a different ID in runtime.

```
const { agent, unregister } = copilotkit.registerProxiedAgent({
  agentId: "chat-1",       // distinct local registry id
  runtimeAgentId: "default", // the existing runtime agent it routes to
});
```

A possible API using this mechanism under the hood could look like this:

```
const { agent: agent1 }  = useAgent({ agentId: "myAgent-1", threadId: "threadId-1", runtimeAgentId: "myAgent" });
const { agent: agent2 }  = useAgent({ agentId: "myAgent-2", threadId: "threadId-2", runtimeAgentId: "myAgent" });
```

where `runtimeAgentId` would be required when providing `threadId`

mme (changes requested):

packages/react-core/src/v2/hooks/use-agent.tsx
davidmckayv: Adopted the `registerProxiedAgent` approach. `useAgent` now takes `runtimeAgentId`, and `runtimeAgentId` is required whenever `threadId` is set — passing `threadId` alone throws with a message pointing at the fix. When `runtimeAgentId` is present the hook registers a private proxied agent (`agentId` → `runtimeAgentId`) and writes the threadId onto that instance, so two hooks routing to the same runtime agent under distinct local `agentId`s no longer clobber each other:

```ts
useAgent({ agentId: "chat-1", runtimeAgentId: "assistant", threadId: "thread-1" });
useAgent({ agentId: "chat-2", runtimeAgentId: "assistant", threadId: "thread-2" });
```

Register/unregister run as a single balanced effect (StrictMode-safe: cleanup unregisters before the remount re-registers), and the registered proxy is exposed via state so the hook swaps off the provisional stand-in deterministically rather than relying on the provider's agents-changed subscription.

Contract tests cover: the two-hooks-same-runtimeAgentId no-clobber case, the throw when `threadId` is passed without `runtimeAgentId`, and register-on-mount / unregister-on-unmount. The RN demo is updated to `{ agentId: "rn-chat", runtimeAgentId: "default", threadId }`.

One call I made that's worth a look: I went with a hard throw (not a dev-only warn) for `threadId`-without-`runtimeAgentId`, since the prop is unreleased and silent clobbering is the failure we're preventing. Happy to soften to a warning if you'd rather.

mme (approved):

mme (approved):

tylerslaton (approved):

PR #6144, opened 2026-07-23
docs(react-native): expand into a production guide

davidmckayv:
## What

Expands the React Native docs page from a "get to first chat" quickstart into a full production guide, matching the depth of the Slack/Teams frontend pages. The existing quickstart Steps are unchanged; the new material is added after them.

### Production sections added
- **Headless imports** — `@copilotkit/react-native/headless` for a custom-UI app, avoiding the chat/attachment native peer deps the default barrel re-exports.
- **Metro configuration for release bundles** — `unstable_enablePackageExports` + browser-first `unstable_conditionNames` so the transitive `jose` dependency resolves to its browser build (fails only at release-bundle time otherwise).
- **Polyfills** — documents the real auto-installed barrel + granular subpaths, including the `crypto` polyfill (backed by `Math.random`) and the recommendation to install a secure RNG for production.
- **Provider options** — the full `CopilotKitProvider` prop surface: `runtimeUrl`, `headers` (auth), `credentials`, `properties`, `onError`, `debug`, `defaultThrottleMs`, `useSingleEndpoint`.
- **Runtime and model wiring** — the string model form, `OPENAI_BASE_URL` (and Anthropic/Google equivalents) for OpenAI-compatible endpoints, `createCopilotNodeListener` zero-dep serving, and `BuiltInAgent` options.
- **Connecting from a device or bench** — LAN IP, Android emulator `10.0.2.2`, `adb reverse` for USB, and Android release cleartext HTTP via `expo-build-properties`.
- **Frontend tools and generative UI** — `useFrontendTool` with zod/Standard Schema, the send-a-turn recipe, `useAgent` read-only subscribe + `threadId`, discriminated-union params, and variable-length `ScrollView` rendering.
- **Run lifecycle and UI** — `isRunning` outliving the visible compose, and gating overlays/reveal motion accordingly.
- **Voice and speech-to-text** — DIY STT with a graceful-fallback warning for de-Googled/automotive hardware.
- Refreshed **What's included** and **Known limitations**, plus a **Next steps** section.

### Dependency on runtime / react-native fixes

Three of the four fixes this guide relies on are now **merged to `main`**, so it documents them as current behavior:

- `@copilotkit/react-native/headless` subpath (headless provider + hooks without chat/attachment deps) — #6142 ✅ merged
- `resolveModel` honoring `OPENAI_BASE_URL` / `ANTHROPIC_BASE_URL` / `GOOGLE_GENERATIVE_AI_BASE_URL` for the string model form — #6140 ✅ merged
- union / `anyOf` / `oneOf` tool-param schemas surviving JSON-schema→zod conversion — #6143 ✅ merged

The fourth — `useAgent` accepting an optional `threadId` (#6141) — is **still open**. The guide therefore does **not** document it as working; it is listed under **Known limitations** instead, and will be promoted once #6141 lands.

Every API, export, env var, and behavior in the page was independently re-verified against the package source (three parallel source audits over `packages/{react-native,react-core,runtime}`). That pass turned up one overstatement, fixed in 6eefe2b: only `/headless` and the root barrel side-effect-import the polyfill barrel — `@copilotkit/react-native/components` does not, so the "all three surfaces auto-install" wording was corrected.


tylerslaton (dismissed):

showcase/shell-docs/src/content/docs/frontends/react-native.mdx
```diff
@@ -284,15 +284,252 @@ hideTOC: true
...
 ## What's included
 
-| Export | Description |
+| Import | Description |
 |--------|-------------|
-| `CopilotKitProvider` | Lightweight provider that points React Native at Copilot Runtime |
-| `useAgent`, `useFrontendTool`, etc. | Re-exported hooks from `@copilotkit/react-core` |
-| `@copilotkit/react-native/polyfills` | ReadableStream, TextEncoder, crypto, DOMException, location |
-| `@copilotkit/runtime` | Server-side runtime used to host your `default` agent |
+| `@copilotkit/react-native` | The full barrel: `CopilotKitProvider`, the re-exported hooks, and the pre-built chat components (`CopilotChat`, `CopilotModal`, `CopilotSidebar`, `CopilotPopup`) plus `useAttachments` |
```
samjulien: The package surface is not accurate here. The default barrel’s `CopilotChat` and `CopilotModal` are the headless wrappers from `src/CopilotChat.tsx` and `src/CopilotModal.tsx`; the fully rendered chat and bottom-sheet modal are exported from `@copilotkit/react-native/components`. The root barrel also does not pull `@gorhom/bottom-sheet` through those headless exports. Please document the root, `/headless`, and `/components` surfaces separately so readers import the UI they expect.

showcase/shell-docs/src/content/docs/frontends/react-native.mdx
```diff
@@ -284,15 +284,252 @@ hideTOC: true
...
+| `@copilotkit/react-native/headless` | Just `CopilotKitProvider` and the hooks — no chat/attachment components, so Metro never resolves their native peer dependencies (see [Headless imports](#headless-imports)) |
+| `@copilotkit/react-native/polyfills` | Barrel that installs every polyfill (streams, encoding, crypto, dom, location) plus the streaming-fetch shim |
+| `@copilotkit/react-native/polyfills/{streams,encoding,crypto,dom,location}` | Granular subpaths for installing a single polyfill |
+| `@copilotkit/runtime` | Server-side runtime (`CopilotRuntime` / `BuiltInAgent` / `createCopilotNodeListener`) used to host your agent |
+
+The provider re-exports the platform-agnostic hooks from `@copilotkit/react-core` — `useAgent`, `useFrontendTool`, `useHumanInTheLoop`, `useCopilotKit`, `useSuggestions`, `useInterrupt`, `useRenderTool`, and more. They behave identically to the web SDK.
```
samjulien: The provider does not re-export hooks; the package does. “They behave identically to the web SDK” is also too broad: React Native substitutes its own `useRenderTool` path and deliberately excludes web-only rendering hooks. Please enumerate the supported shared hooks or link the React Native API reference instead of promising full behavioral identity.

showcase/shell-docs/src/content/docs/frontends/react-native.mdx
```diff
@@ -284,15 +284,252 @@ hideTOC: true
...
+  CopilotKitProvider,
+  useAgent,
+  useFrontendTool,
+} from "@copilotkit/react-native/headless"; // [!code highlight]
+```
+
+The `headless` entry exports the provider and every platform-agnostic hook — everything in this guide except the pre-built chat components. Polyfills are still auto-installed. The default barrel re-exports everything from `headless` plus the chat UI, so existing `@copilotkit/react-native` imports keep working unchanged.
+
+<Callout type="info" title="Use headless for custom UIs">
+  If you build your own chat interface (as the quickstart does), prefer `@copilotkit/react-native/headless`. You only need the default barrel when you actually render `CopilotChat` / `CopilotSidebar` / `CopilotPopup` or use `useAttachments`.
```
samjulien: The quickstart above imports the provider and hooks from the default barrel for a fully custom UI, but this callout says that exact reader should use `/headless` to avoid native peer dependencies. Please update the quickstart imports (currently lines 122 and 160) to `/headless` and reconcile the opening “headless” description with the package-surface section.

showcase/shell-docs/src/content/docs/frontends/react-native.mdx
```diff
@@ -284,15 +284,252 @@ hideTOC: true
...
+</Callout>
+
+## Metro configuration for release bundles
+
+CopilotKit's dependency graph reaches `jose` (a JWT/license dependency, pulled in transitively through `@copilotkit/shared`). `jose` ships separate Node and browser builds behind a package `exports` map, and Metro's default resolution selects the Node build, which imports `node:`-prefixed modules that can't be bundled for Hermes. This surfaces **only at release-bundle time** (`Unable to resolve module node:...`), never at typecheck. Enable package exports and put `browser` ahead of the Node conditions so `jose` (and any similar dual-build dependency) resolves to its browser build:
```
samjulien: This explanation of Metro condition resolution is inaccurate and the global configuration is risky. `unstable_conditionNames` asserts conditions; array order does not establish package-export priority, and globally asserting `browser` can change resolution for unrelated dependencies. The page also claims React Native 0.70+, although package exports arrived in 0.72 and are enabled by default in Metro 0.82 / React Native 0.79. Please document the tested version bounds and scope the browser override to `jose` with `resolveRequest` while preserving Metro’s defaults.

showcase/shell-docs/src/content/docs/frontends/react-native.mdx
```diff
@@ -284,15 +284,252 @@ hideTOC: true
...
+</Callout>
+
+## Polyfills
+
+Hermes lacks several Web APIs CopilotKit relies on: `ReadableStream` / `WritableStream` / `TransformStream`, `TextEncoder` / `TextDecoder`, `DOMException` and `Headers`, `crypto.getRandomValues`, and `window.location`. It also lacks streaming `fetch` bodies (`response.body.getReader()`), which AG-UI's SSE transport needs.
+
+**These are installed automatically the moment `@copilotkit/react-native` (or its `headless` entry) is imported** — the package runs its polyfill barrel and installs an XHR-based streaming-fetch shim on load. The manual `import "@copilotkit/react-native/polyfills"` shown in the quickstart is optional belt-and-suspenders: it's useful when you want the polyfills to run before other imports at your entry point, or when you're installing a secure crypto source first (below). Each polyfill is a no-op if the global it defines already exists, so importing more than once is safe.
```
samjulien: This contradicts the quickstart, which says the manual polyfill barrel import is required before any other code. The production secure-RNG path also needs `react-native-get-random-values` to load before any CopilotKit import, while the quickstart currently installs CopilotKit’s insecure fallback first. Please make the quickstart canonical and avoid appending a later correction.

showcase/shell-docs/src/content/docs/frontends/react-native.mdx
```diff
@@ -284,15 +284,252 @@ hideTOC: true
...
+</Callout>
+
+## Provider options
+
+`CopilotKitProvider` is a context-only provider (it renders no view of its own), so it composes cleanly as the outermost wrapper around your app. `runtimeUrl` is the only required prop, but several others are available:
+
+| Prop | Type | Description |
+|------|------|-------------|
+| `runtimeUrl` | `string` | URL of the Copilot Runtime endpoint (**required**) |
+| `headers` | `Record<string, string> \| (() => Record<string, string>)` | Custom headers sent with every request. Use this for auth — e.g. an `Authorization` bearer token. Pass a function to compute headers per request |
```
samjulien: The callback is not evaluated per request. `CopilotKitProvider` invokes `headersProp()` during React render, memoizes the resulting object, and passes that object to the core. Please say it is evaluated when the provider renders—or change the implementation before promising per-request freshness. This matters for refreshed auth tokens.

showcase/shell-docs/src/content/docs/frontends/react-native.mdx
```diff
@@ -284,15 +284,252 @@ hideTOC: true
...
+</CopilotKitProvider>
+```
+
+<Callout type="info" title="Cloud props aren't wired on React Native yet">
+  The web provider's `publicApiKey` / `licenseToken` cloud props are not supported by the React Native provider. Point `runtimeUrl` at your own Copilot Runtime.
+</Callout>
+
+## Runtime and model wiring
+
+The runtime from the quickstart hosts a `BuiltInAgent` and serves it with `createCopilotNodeListener`, which plugs straight into Node's built-in `node:http` `createServer` — no Hono or Express dependency required. `createCopilotNodeListener({ runtime, basePath, cors })` returns a request listener; `cors: true` is what lets the app reach it cross-origin from a device.
```
samjulien: Native React Native networking has no browser CORS model, so `cors: true` is not what lets a physical device reach this server. It can matter for Expo Web/browser clients; native reachability depends on the server bind address, routing/firewall, and platform transport security. Please correct this causal explanation.

showcase/shell-docs/src/content/docs/frontends/react-native.mdx
```diff
@@ -284,15 +284,252 @@ hideTOC: true
...
+
+## Connecting from a device or bench
+
+`runtimeUrl` must be an address the **device** can reach — `localhost` on a physical device or emulator resolves to the device itself, not your dev machine.
+
+| Setup | `runtimeUrl` host |
+|-------|-------------------|
+| iOS simulator | `localhost` reaches the host machine |
+| Android emulator | `10.0.2.2` is the emulator's alias for the host's `localhost` |
+| USB-attached device | run `adb reverse tcp:8200 tcp:8200`, then use `localhost` on the device |
```
samjulien: `adb reverse` applies only to Android USB devices, so this generic row is misleading for iOS. The LAN `http://` recommendation also needs iOS App Transport Security guidance; the section currently covers only Android cleartext. Please add an iOS physical-device/ATS path, label this row Android, and include the `expo-build-properties` installation step before its plugin configuration.

showcase/shell-docs/src/content/docs/frontends/react-native.mdx
```diff
@@ -284,15 +284,252 @@ hideTOC: true
...
+      return "rendered";
+    },
+  });
+}
+```
+
+Because the tool is registered on the device, the model's call reaches your `handler` on the client — the ideal place to drive a state reducer that your React Native views render declaratively (the agent sends *state*, the client owns rendering). Discriminated-union / `oneOf` / `anyOf` param schemas are preserved through the runtime's schema conversion, so `z.discriminatedUnion(...)` params work. It's still good practice to normalize defensively in the handler — coerce a partially populated tool call into valid state rather than letting a malformed one render broken UI.
+
+<Callout type="info" title="Rendering tool UI">
+  A frontend tool can also carry a `render` function that returns React Native components, drawn as the tool runs. For arbitrary, agent-authored content whose length you can't predict, render it in a `ScrollView` with a `maxHeight` set to the available band — it sizes to content when short and scrolls when long, so it never hard-clips.
```
samjulien: For inline UI rendered by React Native’s chat component, the documented hook should be the React Native `useRenderTool`. `useFrontendTool({ render })` registers with react-core’s renderer collection, while the prebuilt React Native `CopilotChat` reads the separate `RenderToolContext` registry populated by `useRenderTool`. Please separate handler-driven state for a custom UI from inline chat rendering; the latter is not drawn automatically by this `useFrontendTool` path.

showcase/shell-docs/src/content/docs/frontends/react-native.mdx
```diff
@@ -284,15 +284,252 @@ hideTOC: true
...
+</Callout>
+
+### Sending a turn and reading run state
+
+Running a user turn is the two-step the quickstart shows: `agent.addMessage(...)` then `copilotkit.runAgent({ agent })` (the runner lives on the core, not the agent). To cancel an in-flight run, call `copilotkit.stopAgent({ agent })`. To scope a run to a specific conversation without a chat-configuration provider in the tree, pass `threadId` to `useAgent`:
+
+```tsx
+const { agent } = useAgent({ agentId: "default", threadId }); // [!code highlight]
```
samjulien: `threadId` is not supported by `useAgent` on `main` yet: #6141 is still open. As published now, this snippet does not typecheck and the prop is ignored. Please block this docs PR on #6141 or remove the example until that API lands.

showcase/shell-docs/src/content/docs/frontends/react-native.mdx
```diff
@@ -284,15 +284,252 @@ hideTOC: true
...
+<Callout type="warn" title="Don't assume a platform speech recognizer exists">
+  Recognizers that wrap the OS speech service (iOS `SFSpeechRecognizer` / Android `SpeechRecognizer`) can be **absent** on de-Googled or automotive hardware, where the call throws rather than degrading. If you target that kind of device, guard the unavailable case and fall back to text input, or embed an on-device recognizer (e.g. Vosk) as a guaranteed floor. Always give voice a deterministic **Send** control rather than relying on silence detection to end a turn.
+</Callout>
 
 ## Known limitations
 
-- **No pre-built UI components** - you build the chat interface with React Native components
-- **Markdown rendering** - assistant messages with markdown render as plain text; bring your own renderer (e.g. `react-native-markdown-display`)
-- **Voice** - `@copilotkit/voice` has not been adapted for React Native yet
+- **No pre-built UI is required** — the package is headless by design; you build the chat interface with React Native components (the pre-built `CopilotChat` / `CopilotSidebar` / `CopilotPopup` are opt-in via the default barrel).
```
samjulien: “No pre-built UI is required” is product positioning, not a limitation, and it repeats the incorrect export-surface description above. Please reserve this section for actual unsupported or constrained behavior and point readers to the exact `/components` surface when discussing the rendered chat UI.

samjulien (changes requested):
Thanks for expanding this material—there is a lot of useful production context here. I’m requesting changes because several copy-paste paths conflict with the current React Native export/runtime behavior and could leave readers with missing UI, stale auth headers, or platform-specific connection failures.

The inline comments cover the source-backed accuracy issues. At the page level, please also revisit the information architecture: this grows the page to 535 lines while retaining `hideTOC: true` and the description “Get started with CopilotKit in React Native.” Please either enable the TOC and update the frontmatter or split the quickstart from the production guide. The production material also needs to be integrated into the existing steps rather than appended as later corrections—most notably the `/headless` imports and polyfill setup.

Current CI is green, but these are semantic and reader-UX issues that the existing docs checks do not catch.

davidmckayv:
Thanks — this was a genuinely useful review. I re-verified every point against the package source on `main` (`@copilotkit/react-native` 1.63.2) rather than taking them on faith, and **10 of 11 are sustained**. All are now fixed. The verification also turned up several errors you didn't flag, which I've corrected in the same pass.

Rebased onto current `main` first, so this is against `7cbc2b5`.

### Sustained and fixed

- **Package surface.** Confirmed: root's `CopilotChat`/`CopilotModal` are the headless wrappers (`src/CopilotChat.tsx:124-130` returns only a context provider), the rendered chat and bottom-sheet are `/components`, and root does **not** pull `@gorhom/bottom-sheet` — the only import is `src/components/CopilotModal.tsx:20`. Now documents all three surfaces separately with the native peers each one forces Metro to resolve (`/headless`: none; root: `expo-document-picker` + `expo-file-system`; `/components`: `@gorhom/bottom-sheet` + `react-native-streamdown`). Added a warning that `CopilotChat` means two different things depending on the import — that silent blank screen is a nasty trap.
- **Hooks.** Corrected to the *package*, enumerated the shared hooks, and dropped "behave identically to the web SDK". Also documented that RN's `useRenderTool` is its own hook (requires `parameters`, accepts `handler`, returns `ReactElement | null`, no `"*"` wildcard) and that the four web-only rendering hooks are deliberately excluded.
- **Quickstart vs `/headless`.** Real contradiction — fixed by moving the quickstart to `/headless` (lines 122/160 and the tools example), which also keeps readers out of the bundle failure the page warns about.
- **Metro.** You're right on all four sub-points. Replaced the global `unstable_conditionNames` with `resolveRequest` scoped to `jose`, and documented *why*: it's an unordered set of asserted conditions, priority comes from each package's own `exports` key order, so reordering that array does nothing. The old config also applied `browser` globally **and** dropped Metro's default `react-native` condition. Version bounds corrected to opt-in in RN 0.72 / Metro 0.76.1, default-on from Metro 0.82 / RN 0.79.
- **Polyfills.** Made the quickstart canonical instead of correcting it later, and dropped "optional belt-and-suspenders". Strengthened the crypto point: it's a *hard* ordering requirement, not a precaution — both CopilotKit's polyfill and `react-native-get-random-values` install only when `crypto.getRandomValues` is undefined, so any CopilotKit import evaluated first permanently locks in the `Math.random` fallback and silently no-ops the secure library.
- **`headers`.** Confirmed: `CopilotKitProvider` calls the function in its render body, memoizes on `JSON.stringify`, and pushes to the core, which snapshots onto each agent. Table entry corrected, plus a callout on rotating tokens — drive from state, or call `copilotkit.setHeaders()` on refresh. Worth noting a token in a ref or async cache with no accompanying render sends the stale value indefinitely.
- **`cors`.** Corrected. CORS is browser-enforced; RN's native stack sends no `Origin` and never consults `Access-Control-Allow-Origin`, so it's irrelevant to native reachability and only matters for Expo Web.
- **Device connectivity.** Labelled the `adb reverse` row Android-only, added an iOS device row, and added the missing `npx expo install expo-build-properties`. On the iOS path — your instinct was right but the fix needed adjusting: `expo-build-properties` has **no iOS ATS option** (its `PluginConfigTypeIos` doesn't include one), and `NSAllowsLocalNetworking` doesn't cover raw IPs on iOS 17+ (and its presence makes `NSAllowsArbitraryLoads` ignored). So it's `ios.infoPlist` with an `NSExceptionDomains` entry for the dev IP, plus `NSLocalNetworkUsageDescription` for the separate local-network permission gate. Documented both, and why `NSAllowsLocalNetworking` alone isn't enough.
- **Tool rendering.** Confirmed by tracing both registries — and there's an existing test asserting the fallback (`CopilotChat.test.tsx:285-303` expects `Called: myTool`), so a `render` on `useFrontendTool` genuinely never draws in the RN chat. Rewrote the callout to separate handler-driven custom UI from inline chat rendering via `useRenderTool`.
- **`threadId`.** Sustained — `UseAgentProps` has only `agentId`/`updates`/`throttleMs`, it isn't destructured, and the snippet fails with `TS2353`. Rather than block this PR on #6141, I removed the example and recorded it as a known limitation linking to #6141. This page no longer depends on that PR.

### One partial

- **Known limitations.** Agreed the "No pre-built UI is required" line was positioning, not a limitation — removed. For the record, the markdown and voice limitations weren't deleted; both survive reworded (they're at 527-528 in the version you reviewed). Separately, the *old* markdown limitation was already stale before this PR: the rendered chat renders markdown via `CopilotMarkdown`/`react-native-streamdown` (`AssistantMessage.tsx:41`), and it recommended `react-native-markdown-display`, which isn't what the package uses. Corrected rather than dropped.

### One push-back: the TOC / page split

I'd like to keep `hideTOC: true` and the single page. It isn't a deviation from house convention — it *is* the convention:

- **8 of 8** `.mdx` files in `frontends/` set `hideTOC: true`. None omit it.
- Comparable-length pages elsewhere keep it too: `integrations/langgraph/quickstart.mdx` (551 lines), `integrations/microsoft-agent-framework/quickstart.mdx` (535), `agent-spec/quickstart.mdx` (509), `integrations/aws-strands/quickstart.mdx` (476) — all `hideTOC: true`.

So a long page with `hideTOC` is the established pattern for quickstart-style docs here, and splitting this one would make it the odd one out among the frontends. Happy to revisit if we change the pattern across the section — I'd rather not do it for one page. I did take the frontmatter point: the `description` now reflects the page's actual scope instead of promising only "get started".

### Separate issues found while verifying (not fixed here)

Three package-level bugs, which I'll file separately rather than smuggle into a docs PR:

1. **`UseAgentUpdate` is exported type-only** from `react-native` (`src/headless.ts:61`), but react-core documents it as a runtime enum passed as a *value* to `useAgent`'s `updates` option ("must be a value export"). RN consumers can't use `UseAgentUpdate.*` at runtime.
2. **The root barrel's own comments are wrong.** `src/index.ts:20-27` and `src/headless.ts:8-15` state that the root chat components import `@gorhom/bottom-sheet`. They don't. This is where the incorrect surface description in my draft came from, so it's worth fixing before it misleads anyone else.
3. **`src/index.ts:56`** justifies excluding `useRenderToolCall` as "depends on DOM elements via DefaultToolCallRenderer", while `react-core/src/v2/headless.ts:74-80` explicitly says it "introduces no host DOM of its own". One of the two is wrong.

Also worth noting for a future cleanup: `jose` only reaches the RN graph because `@copilotkit/shared` pulls `@segment/analytics-node` for telemetry. If that were kept out of the client-reachable graph — as was already done for `@copilotkit/license-verifier`, which shared imports types-only for exactly this reason — the Metro workaround in this guide would become unnecessary.

showcase/shell-docs/src/content/docs/frontends/react-native.mdx
```diff
@@ -119,7 +132,7 @@ hideTOC: true
         Point the provider at the runtime endpoint. Android emulators reach the host machine at `10.0.2.2`; iOS simulators can use `localhost`.
 
         ```tsx title="App.tsx"
-        import { CopilotKitProvider } from "@copilotkit/react-native"; // [!code highlight]
+        import { CopilotKitProvider } from "@copilotkit/react-native/headless"; // [!code highlight]
```
samjulien, line 135: [P1] The unversioned install step currently resolves to stable `@copilotkit/react-native@1.63.2`, but that release does not export `@copilotkit/react-native/headless`; the subpath exists only in the current canary. The documented runtime base-URL and union-schema fixes also landed after `1.63.2`. As written, a reader following the quickstart gets an unresolved import here. Please gate the docs deployment on a stable release containing these APIs, or explicitly install/document a version that contains them.

showcase/shell-docs/src/content/docs/frontends/react-native.mdx
```diff
@@ -282,17 +296,351 @@ hideTOC: true
...
+
+  - `unstable_conditionNames` is an **unordered set of asserted conditions**, not a priority list. The winning target is chosen by the key order in *each package's own* `exports` map, so reordering this array changes nothing.
+  - It applies **globally**, so every dual-build dependency in your graph switches to its browser build — not just `jose`.
+  - Metro's React Native default is `["require", "react-native"]`. Replacing it wholesale drops the `react-native` condition, which changes resolution for any package that ships one.
+
+  Scoping with `resolveRequest` keeps Metro's defaults intact everywhere else.
+</Callout>
+
+<Callout type="info" title="Do you need `unstable_enablePackageExports`?">
+  Package `exports` resolution is **enabled by default from Metro 0.82 / React Native 0.79**, so no flag is needed there. It arrived opt-in in React Native 0.72 (Metro 0.76.1) — on React Native 0.72–0.78 set `resolver.unstable_enablePackageExports = true`, which is also what lets Metro resolve this package's `/headless`, `/components`, and `/polyfills` subpaths. React Native below 0.72 has no package-exports support.
```
samjulien, line 386: [P1] This correctly says React Native below 0.72 has no package-exports support, but the prerequisite still advertises React Native 0.70+ and every frontend example uses `/headless`. Please make 0.72 the guide minimum—ideally recommending RN 0.79 / Metro 0.82.3+—or provide a separate 0.70–0.71 path using the root export and its required peers.

showcase/shell-docs/src/content/docs/frontends/react-native.mdx
```diff
@@ -282,17 +296,351 @@ hideTOC: true
...
+
+```tsx title="App.tsx"
+import {
+  CopilotKitProvider,
+  useAgent,
+  useFrontendTool,
+} from "@copilotkit/react-native/headless"; // [!code highlight]
+```
+
+All three surfaces auto-install the polyfills, and the root barrel re-exports everything from `/headless`, so existing `@copilotkit/react-native` imports keep working unchanged.
```
samjulien: [P2] `/components` does not import the polyfill barrel: the root and `/headless` entries do, but `src/components/index.ts` only re-exports components. Please say that the two provider surfaces auto-install polyfills. An app pairing `/components` with `CopilotKitProvider` gets them through the provider import; importing `/components` alone does not.

samjulien (changes requested):
Thanks for the thorough revision—most of the original accuracy issues are now addressed. I found three remaining blockers around release availability, the supported React Native floor, and which entry points install polyfills; details inline.

Two non-blocking UX improvements would also make the guide easier to follow:
- The frontend-tool section says to add Zod but does not provide an installation command.
- The prerequisites promise both bare React Native and Expo, but app creation/run commands cover only bare CLI while the physical-device configuration covers only Expo. Consider adding framework-specific tabs or explicitly scoping each path.

I’m treating the no-TOC concern as section-wide docs UX debt rather than blocking this PR, given the established convention you documented.

samjulien (approved):

PR #6146, opened 2026-07-24
feat(channels): render arbitrary JSX as images via Takumi

AlemTuzlak:
`thread.post` now turns your own React JSX into a PNG and posts it. Takumi rasters the tree in-process (no browser). Native channel-UI components still render as Slack Block Kit / Teams Adaptive Cards.

This PR is merged onto latest `main`. The old `@copilotkit/channels/charts` library is gone. Keep Takumi, Render, and Carousel.

## What does this PR do?

- `thread.post` auto-routes arbitrary React JSX to a PNG (Takumi) and uploads it through `postFile`.
- Channel-UI components stay native. `update` / `postEphemeral` / `awaitChoice` reject image JSX with a clear error.
- `createChannel({ render })` takes fonts, stylesheets, size, and `allowImageUrl` (default denies private/loopback hosts).
- `Render` and `Carousel` mix a Takumi snapshot with native buttons on the same message.
- The Slack example shows this with `MrrCard`, branded showcase cards, and a product carousel.

## Related PRs and Issues

- None.

## Checklist

## Testing

**Commands run**

1. `nx run @copilotkit/channels-core:test` — 331 passed
2. `nx run @copilotkit/channels:test` — 2 passed
3. `pnpm --filter slack-example test` — 78 passed

I did not run `pnpm test:pr` (full monorepo). Lefthook did not run on these commits (`core.hooksPath` points at another clone).

**Manual test**

1. In `examples/slack`, set `INTELLIGENCE_API_KEY` and `OPENAI_API_KEY`.
2. Run `pnpm --filter slack-example runtime` and `pnpm --filter slack-example dev`.
3. In Slack, ask the bot for the MRR card (or run `/prs`). Confirm a PNG lands in the thread.
4. Send a native Linear issue card. Confirm it is Block Kit, not an image.
5. Run `/carousel`. Confirm the snapshot and the Buy button share one message.

**How this PR makes testing easy**

- Unit tests cover routing, Takumi nested-component raster, URL policy, and Slack example posts.
- `examples/slack` is a runnable demo of image posts and mixed native UI.

## Risk / rollback

Image posts cannot be edited in place. Takumi fetches remote URLs only when `allowImageUrl` allows them (default denies private/loopback). Rollback: revert the PR.

## Public API change

**Before**

```ts
await thread.post(nativeChannelJsx);
```

**After**

```ts
await thread.post(nativeChannelJsx);
await thread.post(mrrCardElement, { filename: "mrr.png" });
```

`react` is an optional peer of `@copilotkit/channels`. Install it only if you post images. `@copilotkit/channels/charts` is not an export any more.

## Summary by CodeRabbit

* **New Features**
  * Post React components as branded images with customizable fonts, styles, and dimensions.
  * Combine native channel UI with rendered snapshots, cards, and carousels.
  * Added Slack showcase commands for PR radar, weekly pulse, cycle standup, product carousels, and MRR cards.
  * Added staged image support across Slack, Teams, Discord, Telegram, and WhatsApp.
* **Documentation**
  * Added guides for JSX image posting and renderable carousels.
* **Bug Fixes**
  * Improved threaded replies, image delivery, upload handling, and port validation.
  * Replaced chart and diagram rendering tools with MRR and carousel experiences.


packages/channels-core/src/charts/bar-chart.ts
tylerslaton, line 1: When we spoke offline we decided that a part of this was that we didn't want to have pre-canned components. Is that still the case? As written it looks like we're shipping a component library. Was that the intent?

examples/slack/assets/fonts/LICENSE.txt
tylerslaton, line 1: Why are we brining the font in?

examples/slack/app/managed.ts
```diff
@@ -90,6 +89,13 @@ async function main() {
     tools: [...appTools, ...defaultSlackTools],
     context: [...appContext, ...defaultSlackContext],
     commands: appCommands,
+    // Takumi JSX image rendering config. No `fonts`/`googleFonts` (the latter
+    // can trigger a network fetch and hang tests/CI) — Takumi's built-in Latin
+    // covers this demo. The `charts` components render in color from their own
+    // fixed hex palette (no stylesheet needed), and `MrrCard` uses inline styles.
+    render: {
```
tylerslaton, line 96: How does the MRR component actually land into the channel? I don't see where that actually gets added right now.

tylerslaton (commented):
Had some questions I'd like cleared up before approving. Below this review you'll find one from the agent that I found had some valid concerns.

tylerslaton (changes requested):
> 🤖 **Agent-generated review (Codex):** This review was produced by an AI coding agent. The findings should be validated by a human maintainer before merging.

The overall direction is promising: lazy image rendering, preservation of the native channel path, and the chart coverage are all well designed. However, I found several issues that I believe should be addressed before approval:

1. **Potential SSRF:** Takumi can fetch remote image and CSS URLs, but `RenderConfig` exposes no URL policy. User-influenced URLs could potentially access internal services. Please deny remote access by default or expose a safe allowlist/private-network policy.

2. **Regular React JSX does not type-check:** `ReactElementLike` requires the internal `$$typeof` property, which React’s public `ReactElement` type does not expose. The tests currently hide this with `as never`. Please add a cast-free TypeScript usage test.

3. **Image posts return invalid `MessageRef`s:** `postImage()` treats `fileId` as a message ID. Slack returns no ID here, while Telegram and WhatsApp return media-storage IDs. This can break later delete, reaction, or update operations.

4. **Components can be evaluated multiple times:** JSX detection invokes unbranded components speculatively, followed by another invocation during rendering. `awaitChoice()` can invoke them three times. This risks duplicate API calls and other side effects.

5. **`FlowDiagram` does not render the declared edges:** Edges currently influence node layering, but individual connections and edge labels are discarded. Different graphs can therefore produce the same visual output.

The current CI is green, but these are public API, security, and adapter-integration gaps that the existing tests do not cover. I recommend addressing them and adding focused regression tests before merging.

packages/channels-core/src/charts/bar-chart.ts
AlemTuzlak: Intentional, and we want to keep them shipped — but I take the "component library" concern, so here's the reasoning.

They aren't a general UI kit. They exist because the image path has a hard constraint: Takumi is a static SSR rasterizer with no browser, so Recharts/Chart.js/Mermaid — every charting library a user would reach for — simply cannot run. "Post a chart to the thread" is the single most common thing people want out of `post(<JSX/>)`, and without these there is no way to do it at all; you'd have to hand-compute bar geometry in your own JSX. That's not a component library, it's the only chart primitive available on this render path.

Scope is deliberately narrow and won't grow: geometry in TS, no runtime deps, no interactivity, no theming system (fixed hex palette, plus `colors`/`className`/`style` per chart). They're behind a separate `@copilotkit/channels/charts` entrypoint, so nothing loads them unless you import them, and the chart/diagram *tools* did move out to `examples/slack` in 602c430 — what's left is just the presentational pieces.

If you'd still rather not ship them, the alternative is deleting the directory and moving these into `examples/slack` as copy-paste example code; say the word and I'll do that in a follow-up rather than block this PR. Happy to hop on a call if it's easier.

examples/slack/assets/fonts/LICENSE.txt
AlemTuzlak: Because Takumi has **no system fonts** — it ships one built-in Latin fallback (Geist) and nothing else. Any glyph outside that renders as tofu. So `render.fonts` is a real part of the API this PR adds, and the example has to demonstrate it with an actual font file.

Plus Jakarta Sans is the CopilotKit brand typeface, so the branded cards in `app/showcase` look like our brand instead of generic Geist. It's OFL-1.1 (hence the `LICENSE.txt` next to it), 2 static weights (Medium 500 / Bold 700), ~350KB total, and it's committed rather than fetched because the alternative — `googleFonts()` — does a network fetch at startup, which hangs CI (the code comments in `app/managed.ts` call this out).

Worth noting the managed variant (`app/managed.ts`) deliberately configures **no** fonts, so the "just use the built-in Latin" path is covered too. If you'd rather the repo not carry binaries at all, I can drop them and let the example render in Geist — it only costs the brand typeface on the cards.

examples/slack/app/managed.ts
AlemTuzlak: It lands via a tool, not via anything wired at this line. The chain:

1. `tools: [...appTools, ...defaultSlackTools]` (a few lines above) — `appTools` includes `renderMrrTool` (`app/tools/index.ts`).
2. `renderMrrTool` is `render_mrr` (`app/tools/render-mrr.tsx`), and its handler is just:

   ```tsx
   await thread.post(<MrrCard value={value} delta={delta} />, { filename: "mrr.png", title: "MRR" });
   ```
3. `<MrrCard/>` is arbitrary app JSX, so it isn't a branded channels-ui component. `thread.post` peeks at what it returns, sees a React element, and routes it to the Takumi image path, which rasterizes it and uploads through `postFile`. That auto-routing is the whole feature of this PR — no wrapper or registration at the call site.
4. The `render: { width: 720 }` at this line is only the **canvas config** for that rasterization (width; no `fonts`, so Takumi's built-in Latin is used). It doesn't register the component.

So: the agent decides to call `render_mrr` (the system prompt in `runtime.ts` maps "MRR / revenue summary" → `render_mrr`), the handler posts the JSX, and the JSX becomes a PNG in the thread. Nothing about `MrrCard` is declared here — the tool list is the only wiring.

AlemTuzlak:
Addressed the Codex review in 8c49a42. All five findings were real; each has a regression test. Answers to your three inline questions are in the threads above (short version: charts stay, with reasoning; the font is there because Takumi has no system fonts; the MRR card lands via `render_mrr` → `thread.post(<MrrCard/>)`, not via anything at that line).

**1. SSRF — fixed.** Takumi does fetch remote resources referenced by the JSX (``, CSS `url()`, emoji sheets), and it exposes an `allowUrl` hook we weren't passing. `RenderConfig.allowImageUrl` now gates every outbound fetch, defaulting to a policy that denies non-HTTP(S) schemes plus literally private/loopback/link-local/internal hosts — `169.254.169.254` and `metadata.google.internal` included. It's channel-wide only, deliberately: a per-post override is exactly the knob a prompt-injected tool argument would reach for. The docs state plainly that a synchronous hook can't resolve DNS, so rebinding isn't covered and untrusted URLs want an explicit allowlist.

**2. Regular React JSX now type-checks, cast-free.** You were right that `as never` was hiding it: `ReactElementLike` required React's internal `$$typeof` and typed `props` as `Record<string, unknown>`, which a props *interface* isn't assignable to. It's now structural and loose, mirroring Takumi's own `ReactElementLike`. New `render/post-typing.test.tsx` posts a host tag, a component with a typed props interface, a `createElement` element, and a `ReactElement<CardProps>` — all with zero casts, and tsc failing on that file is the test. The remaining `as never` casts in `thread.image.test.ts` are on `update`/`awaitChoice`/`postEphemeral`, which intentionally *don't* accept arbitrary JSX in their signatures; those tests assert the runtime rejection for untyped callers.

**3. Invalid `MessageRef`s — fixed.** `postFile` was conflating two different ids. `PostFileResult` now separates `messageId` (a real message id, the only thing safe as a `MessageRef`) from `fileId` (a media/storage handle). Per adapter: Slack reads the ts out of the uploaded file's `shares` map; Telegram reports `message_id` alongside `document.file_id`; WhatsApp reports the wamid alongside the media id; Discord/Teams post the attachment *as* a message, so their id moved to `messageId`. When no message id is available (managed transport → async outbox), `post()` returns `{id: ""}` and warns once per platform rather than handing back something delete/react would reject.

**4. Multiple component evaluation — fixed.** The peek verdict is memoized per element in a `WeakMap`, so classification invokes a component at most once no matter how many times the same element is classified (`awaitChoice`'s 3 → 1). The native path still calls it once more to bind, which is inherent to binding; purity is documented as the requirement it already was.

**5. `FlowDiagram` edges — fixed.** Each adjacent-layer gap now draws one arrow *per edge* crossing it, captioned with that edge's `label`, plus `from -> to` when the gap carries several edges (position alone can't say which boxes an arrow joins). Skip-level, backward, and self edges — which a layered layout has nowhere to route — are listed as text beneath the flow instead of being dropped. Tests assert a fan-out and a chain over the same nodes don't render identically, that labels survive, and that every edge of a cyclic graph appears.

Two incidental bugs the new tests caught while writing them: the URL policy missed bracketed IPv6 literals (`URL.hostname` keeps the brackets), and edge captions used `→`, which is tofu under Takumi's built-in Geist Latin — both now ASCII/bracket-stripped.

Full suite green (`channels-core` 225 tests, all adapters, `slack-example`), plus check-types and lint.

AlemTuzlak:
Pushed two fixes found while testing this end to end against a real Slack workspace.

**`fix(channels)`** — `<Render>` sizing and altText

- `<Render>` accepted a per-node `width` but no `height`, so every snapshot was drawn onto the render config's default (480px). `ProductCard` uses `h-full`, which never resolves in Takumi without a sized parent, so the card collapsed to its content height at the *top* of the canvas and the rest was dead space — which a carousel hero image crops, making the card look corrupted. Added the symmetric `height` prop and used it for the catalog cards.
- `thread.post` forwards `opts?.altText`, which is `undefined` when a caller omits it (as `renderCatalogCarousel` did). The key then drops out of the JSON and the gateway rejects the packet as `invalid_provider_effect`, since it requires 1..2000 chars. Now falls back to the filename.

**`fix(react-core)`** — unrelated, but it blocked committing at all

`assert-headless-purity.mjs` compared esbuild metafile keys (always POSIX-separated) against `path.relative` output (backslashes on Windows). Every scan aborted with *"the scan measured nothing"*, failing 8 tests and blocking the pre-commit hook on Windows for any commit in the repo. One-line separator normalisation; that suite now passes 45/0.

## Deploy ordering

This client change depends on a gateway fix: CopilotKit/Intelligence#1134.

Slack grants file access **only** via `channel_id` at `files.completeUploadExternal` time, with no way to share an upload afterwards. Staging block images unshared therefore makes them permanently uploader-only, and they render as placeholders for everyone but the bot. Intelligence#1134 defers completion so all staged files are completed in one call with `channel_id`, then rewrites that share message into the real message.

**The gateway must ship first.** A pre-ENT-1294 gateway rejects the `share` key outright with `invalid_provider_effect`, so this client would break against it.

Issue #603, opened 2024-09-27
⭐ 02 - Recipe Generator (Hacktoberfest Demo)


NathanTarbert:
### 🤩 Thank you for participating in the CopilotKit Hacktoberfest event. 

Follow [This Link to Get Started!](https://github.com/CopilotKit/CopilotKit/tree/main/community/content)🎉

#### NOTE: CopilotKit will allow multi-assign issues

  Please show your support and give a ⭐ to the CopilotKit repository.

### Description
Create a demo that suggests recipes based on available ingredients, using CopilotKit for natural language processing

#### Using CopilotKit
- **Required**: Your project must utilize CopilotKit as a core component.
- **Installation**: Include CopilotKit by following the installation guide at [CopilotKit Quickstart](https://docs.copilotkit.ai/quickstart).
- **Documentation**: Familiarize yourself with CopilotKit's features through the [Official Documentation](https://docs.copilotkit.ai).

#### Styling with Shadcn-UI
- **Required**: All UI components in your project should be styled using Shadcn-UI.
- **Installation**: Install Shadcn-UI [documentation](https://ui.shadcn.com/docs/installation)

![Swag_2](https://github.com/user-attachments/assets/d120bf97-95af-40a6-8618-6fb6a9198696)

sanketshinde3001:
Interested..

NathanTarbert:
@sanketshinde3001, please follow the link at the top to get started!

sanketshinde3001:
Yesssss..

sanketshinde3001:
@NathanTarbert Can i use  only NextJS for this ? And where to add this recipe generator in repo.

NathanTarbert:
@sanketshinde3001, yes you can use NextJS but you also have to include CopilotKit.

Follow [this link](https://github.com/CopilotKit/CopilotKit/tree/main/community/content) to get started

AJBrownson:
@NathanTarbert 
I'd like to work on this

devshah207:
Can I also contribute please??

sanketshinde3001:
@AJBrownson @devshah207 Thanks , but i started to work on it.

AJBrownson:
@sanketshinde3001 of course
But it is written up there:
 "NOTE: CopilotKit will allow multi-assign issues"

NathanTarbert:
That is correct @sanketshinde3001, we are allowing multi-assign

sanketshinde3001:
[...]

PR #6154, opened 2026-07-24
fix(channels): finalize direct render turns

tylerslaton:
## Problem

Realtime Channel deliveries could complete without a terminal `finalize` frame. Direct operations left the Connector Outbox lane open, output-free handlers had no accepted completion pointer, and a discrete operation after `runAgent()` could leave the agent's earlier finalize in the middle of the lane.

## Why

Only the agent renderer owned a finish lifecycle. Discrete operations bypassed it, and realtime delivery state started as already terminal, so handler completion did not guarantee that the lane ended in `finalize`.

## Fix

- Track sequence and terminal state together for each realtime turn.
- Start every realtime delivery as needing finalization, then append `finalize` after the handler only when its last accepted frame is non-terminal.
- Preserve deterministic sequence numbers on redelivery and avoid duplicating an agent run's existing finalizer.
- Treat the gateway's explicit listener-fenced signal as an immediate terminal error instead of claiming a clean Phoenix channel close will reconnect.
- Document that runtime instance ids are unique per live Channel activation, stable only across that activation's transport reconnects, and re-minted on restart.
- Cover direct post, output-free, and `finalize → post → finalize` lifecycles while retaining the never-self-ack assertion.
- Verify with formatting, lint, typechecks, builds, all 185 Channels Intelligence tests, all 1,755 runtime tests, and the repository pre-commit package checks.

### DIFF
--- a/packages/channels-intelligence/src/http-transports.ts
+++ b/packages/channels-intelligence/src/http-transports.ts
@@ -47,7 +47,7 @@ export interface IntelligenceTransportConfig {
   apiKey: string;
   /** Project-unique channel name; defaults from `createChannel({ name })`. */
   channelName: string;
-  /** Stable runtime instance id (`rti_…`); generated when omitted. */
+  /** Transport-instance id (`rti_…`); generated when omitted. */
   runtimeInstanceId: string;
   /** Adapter kind. First slice is `"slack"`. */
   adapter: string;
--- a/packages/channels-intelligence/src/intelligence-adapter.ts
+++ b/packages/channels-intelligence/src/intelligence-adapter.ts
@@ -269,6 +269,9 @@ export class IntelligenceAdapter implements PlatformAdapter {
   /** Per-turn egress sequence; reset at the start of each turn's processing so
    * a redelivered turn reproduces the same operation id sequence. */
   private readonly seq = new Map<string, number>();
+  /** Whether the last durably accepted realtime frame for a live turn still
+   * needs a terminal `finalize` before its delivery may complete. */
+  private readonly renderNeedsFinalize = new Map<string, boolean>();
 
   constructor(private readonly opts: IntelligenceAdapterOptions = {}) {
     this.source = opts.source;
@@ -356,8 +359,12 @@ export class IntelligenceAdapter implements PlatformAdapter {
     // true for at-least-once, lease-based delivery; overlapping redeliveries of
     // the same turnId would perturb the counter.
     this.seq.set(env.turnId, 0);
+    if (this.renderSink) {
+      this.renderNeedsFinalize.set(env.turnId, false);
+    }
     try {
       await this.dispatchTo(env);
+      await this.finalizeRenderedTurn(env);
       await this.requireSource().ack(env.deliveryId);
     } catch (err) {
       await this.requireSource().nack(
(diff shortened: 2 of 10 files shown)


BenTaylorDev (commented):
Read the full diff plus the surrounding code, and verified the load-bearing claims against `thread.ts`, the gateway transport, and app-api's `acceptChannelRenderEvent` in Intelligence.

Nothing here is a blocker. Reviewed alongside CopilotKit/Intelligence#603 — see the landing-order note at the bottom.

## Verified correct

I chased the three ways this could be wrong and all three hold:

- **No double-finalize on agent runs.** `thread.ts:461` always awaits `renderer.finish?.()`, which enqueues `finalize` then `drain()`s the serial push chain. So `recordAcceptedRenderEvent` has already flipped the turn terminal before `dispatchTo` returns, and `finalizeRenderedTurn` no-ops. The interrupt path is covered too: `markInterrupted()` enqueues `interrupt` + `finalize` and drains; `finish()` then early-returns on `aborted`.
- **No un-drained chain race.** The finalize push can't jump the queue, because `finish()`/`markInterrupted()` are the only exits from `runAgent` and both drain first.
- **The finalize makes it into the completion pointer.** `RealtimeGatewaySource.ack` builds `acceptedThrough` from the live `state.accepted` high-water map (`realtime-gateway-transport.ts:600`), which the finalize's receipt has already updated. And finalizing *before* ack is mandatory rather than incidental: `acceptChannelRenderEvent` gates on `bid.status = 'leased'`, so a frame pushed after the completion commits would be rejected.

The egress-fallback path is correctly untouched — `renderNeedsFinalize` is only seeded when `this.renderSink` exists.

## Findings

**1. The case the docstring advertises has no test — and it's the one that breaks downstream.**

`recordAcceptedRenderEvent`'s comment explicitly supports "a later post/update/file/delete after an agent-run finalize flips the turn back to non-terminal", i.e. a lane shaped `…finalize, post, finalize`. Nothing tests it, on either side of the split.

It matters because on the Intelligence side `driveRenderFrame` maps *both* finalizes to `renderer.finish?.()`, so the provider run renderer gets finished twice. `connector-outbox-render.ts:1250` asserts finish() is idempotent in a comment; no test pins it.

Worth adding an adapter unit test (`runAgent()` then `thread.post()` → exactly one trailing finalize, none duplicated after the run's own) and a matching outbox test driving `post, finalize, post, finalize`.

**2. OSS-491 is one character away, and this is the PR that owns it.**

`dispatch` seeds `renderNeedsFinalize` to `false`, so a handler that renders nothing still sends zero frames, and `ack()` bails with an empty `acceptedThrough` — the transport's own NOTE says that turn then redelivers until `max_attempts`.

Seeding `true` instead would give every render-sink turn a terminal frame and therefore a valid completion pointer, closing that pile-up SDK-side. The cost is that the Connector Outbox would claim + `startTurn` + finalize + `markSent` for output-free turns.

I'd want an explicit yes/no on this in the PR either way, since "every rendered delivery has a terminal frame" is exactly what the new test comment claims.

**3. Two parallel per-turn maps.** `seq` and `renderNeedsFinalize` are seeded and deleted in lockstep, keyed identically, and `renderNeedsFinalize.has()` is load-bearing as the "am I tracking this turn" gate. Collapsing to one `Map<string, {seq, needsFinalize}>` removes the chance of the two drifting. Optional.

**4. Test-only nit.** `renderKind` is ~30 lines of hand-rolled narrowing for a shape the tests already know statically. Also, the rename dropped "never self-ack" from the test title while the test still asserts it — the new title undersells its coverage.

## Landing order

This and CopilotKit/Intelligence#603 are two halves of one fix, and both are independently landable. #603 should go first: it's the compat repair for every SDK already deployed, and it becomes a no-op once this ships. This PR alone fixes nothing for existing deployments.

PR #6159, opened 2026-07-24
fix(showcase/ms-agent-python): keep the user's prompt on the multimodal PDF turn

jpr5:
`d6:ms-agent-python/multimodal` has been red in staging and prod since 2026-05-30. Turn 1 (image) passes; turn 2 (PDF) fails. This fixes it — **without touching the fixture**, because the fixture was never the problem.

## The verbatim turn-2 error

Backend (`showcase-ms-agent-python`), and reproduced locally:

```
[/multimodal] Streaming failed
openai.InternalServerError: Error code: 503 - {'error': {'message': 'Strict mode: no fixture matched',
  'type': 'invalid_request_error', 'param': None, 'code': 'no_fixture_match'}}
The above exception was the direct cause of the following exception:
agent_framework.exceptions.ChatClientException: ("<class
  'agent_framework_openai._chat_completion_client.OpenAIChatCompletionClient'> service failed to
  complete the prompt: Error code: 503 - {'error': {'message': 'Strict mode: no fixture matched', …
```

Surfaced in the browser as `An internal error has occurred while streaming events.`, with the probe reporting `failure_turn: 2`, `turns_completed: 1`.

## Request-shape diagnosis

This reads like a fixture gap and is not one. I pulled the **actual outbound request** off the local aimock's `GET /__aimock/journal` during a failing run. Turn 2, verbatim (bodies elided):

```
[0] role=system  "You are a helpful assistant. The user may attach images or documents…"
[1] role=user    "can you tell me what is in this demo image I just attached"
[2] role=user    [image_url <data:image/png;base64,iVBORw0K…>]
[3] role=user    [image_url <data:image/png;base64,iVBORw0K…>]
[4] role=assistant "The attached image is the CopilotKit logo — a clean, geometric mark…"
[5] role=user    "can you tell me what is in this demo pdf I just attached"
[6] role=user    "[Attached document]\nCopilotKit Quickstart\nAdd AI copilots to your React…"
[7] role=user    "[Attached document]\nCopilotKit Quickstart\nAdd AI copilots to your React…"
```

One logical user turn arrived as **three separate user messages**, and the *last* one carries only the flattened document — the question is nowhere in it. That is why aimock's strict mode refused it: `userMessage` is a substring match against the last user turn, and the last user turn was a PDF dump.

**Root cause:** `agent_framework_openai` emits **one OpenAI message per `Content`**. `_chat_completion_client._prepare_message_for_openai` builds a fresh `args` dict on every iteration of its content loop, so a user `Message` carrying `[prompt_text, flattened_doc_text]` serialises to two consecutive user messages — prompt-only, then document-only. `_PdfFlattenChatMiddleware` was appending the flattened `[Attached document]` text as a *second* text `Content` beside the prompt, which is exactly the shape that gets split.

Two corroborating details that make the mechanism airtight:

- **Why turn 1 (image) passes.** aimock already skips *text-less* trailing user messages (`getLastUserText` in `router.ts`, whose comment documents this exact MS Agent Framework behavior). The image turn's split-off trailing message has no text at all, so aimock falls back to the prompt message and matches. The PDF turn's trailing message *does* have text — the document — so there is nothing to skip past.
- **Why `langgraph-python` is green** doing the identical `[Attached document]` flattening: LangChain keeps multiple text parts *inside one message* rather than splitting them into separate messages.

This is a product bug, not a mock artefact. Against a real LLM it would not 503 — the model would just answer the wrong thing, because the question is buried behind a document dump instead of being the current turn.

## The fix

`showcase/integrations/ms-agent-python/src/agents/multimodal_agent.py`

1. **Merge** the flattened document *into* the message's existing prompt text content instead of appending it as a second content. The turn stays a single text content and serialises to a single user message: `"<prompt>\n[Attached document]\n<body>"`.
2. The merge **copies** the prompt `Content` rather than mutating it. This is load-bearing: the middleware restores the original `contents` list after `call_next`, and that restore only undoes the *list* swap — an in-place mutation would leak the raw PDF body into the AG-UI `MESSAGES_SNAPSHOT` and render a wall of PDF text in the user's chat bubble. There is a test for this.
3. **Attachment-only turns** (a PDF with no question) still work: with no text content to merge into, the flattened document stands alone as the message body.
4. **Dedupe identical flattened blocks.** The page's `LegacyConverterShim` appends a legacy `binary` mirror alongside every modern attachment part, so the same PDF reached the middleware twice and its body was being sent to the model twice (visible as the duplicated `[6]`/`[7]` above). Now emitted once.

Post-fix outbound turn 2, same journal endpoint:

```
[5] role=user "can you tell me what is in this demo pdf I just attached\n[Attached document]\nCopilotKit Quickstart\nAdd AI copilots to your React application with CopilotKit…"
matched fixture userMessage: "can you tell me what is in this demo pdf I just attached"
```

One user message, prompt intact, document intact, emitted once.

## The fixture is untouched

```
$ git diff --stat origin/main -- showcase/aimock/
(empty)
```

The existing `userMessage` match key was always correct; the corrected request shape is what satisfies it. Relaxing or re-recording the fixture to match the broken request was an explicit non-goal — it would have made the cell actively certify a model that never sees the user's question.

## Same-pattern audit

- `_PdfFlattenChatMiddleware` is the **only** `ChatMiddleware` in `ms-agent-python`, and the only place in the integration that constructs `Content` or reassigns `message.contents` (`grep` for `ChatMiddleware` / `Content.from_text` / `.contents =` across `src/` returns hits in this one file only). No second instance of the pattern to fix.
- `ms-agent-python` is the only MS-Agent-Framework Python integration doing PDF flattening — `ms-agent-dotnet` has a multimodal e2e spec but no Python agent. The other `[Attached document]` implementations (`langgraph-python`, `langgraph-fastapi`, `agno`, `claude-sdk-python`, `langroid`, `pydantic-ai`, `langgraph-typescript`, `built-in-agent`) run on frameworks that do not split a message's contents into separate wire messages, so they are not exposed to this. The upstream one-message-per-`Content` behavior is pinned by a dedicated test, so if it ever changes we find out by that test failing rather than by a silent regression.
- The file is a regular per-integration file, not a `shared/` symlink (`git ls-files -s` → `100644`). No shared code touched; `validate-shared-symlinks.ts` confirms no new erosion.

## Red / green / control

All three on the real probe surface, from a clean worktree at `origin/main` `38613623f4`.

### RED — before the change

```
$ bin/showcase test ms-agent-python:multimodal --d6 --direct --verbose --cycle --isolate

[conversation-runner] turn 1/2 — assistant settled { bubbleIndex: 0, textLength: 100, hasAssertions: true }
[conversation-runner] turn 1/2 — assertions passed
[conversation-runner] turn 2/2 — sending message { inputLength: 29, timeoutMs: 60000 }
[conversation-runner] turn 2/2 — FAILED {
  errorCategory: 'assertion-failed',
  turnsCompleted: 1,
  elapsedMs: 1577,
  bodyTextLength: 421,
  hasTextarea: true,
  hasErrorBoundary: false
}
[warn] CVDIAG component=harness-d6 boundary=fixture-match … status=miss … error=chat errored: copilot-error-banner visible — An internal error has occurred while streaming events.
[info] probe.e2e-full.service-complete {"slug":"ms-agent-python","passed":0,"failed":1,"skipped":0,"incapable":0,"total":1,"state":"red","durationMs":9384}
  ✗ d6:ms-agent-python red (9.5s)
    multimodal: chat errored: copilot-error-banner visible — An internal error has occurred while streaming events.

  0 passed, 1 failed (9.5s)
⚠ Tests failed for ms-agent-python:multimodal (exit 1)
```

Evidence the outbound request lacked the prompt — aimock journal from that run, 8 entries, `200,503,503,503,200,503,503,503` (2 attempts × 3 retries on turn 2):

```
[5] role=user STRING "can you tell me what is in this demo pdf I just attached"
[6] role=user STRING "[Attached document]\nCopilotKit Quickstart\nAdd AI copilots to…"
[7] role=user STRING "[Attached document]\nCopilotKit Quickstart\nAdd AI copilots to…"
status: 503
```

### GREEN — after the change, fixture unchanged

```
$ bin/showcase test ms-agent-python:multimodal --d6 --direct --verbose --rebuild --keep --isolate

[conversation-runner] turn 1/2 — assistant settled { bubbleIndex: 0, textLength: 100, hasAssertions: true }
[conversation-runner] turn 1/2 — assertions passed
[conversation-runner] turn 2/2 — assistant settled { bubbleIndex: 1, textLength: 233, hasAssertions: true }
[conversation-runner] turn 2/2 — assertions passed
[conversation-runner] conversation completed successfully { turnsCompleted: 2, totalDurationMs: 8279 }
[info] probe.e2e-full.feature-complete {"slug":"ms-agent-python","featureType":"multimodal","pass":true,"durationMs":8788}
[info] probe.e2e-full.service-complete {"slug":"ms-agent-python","passed":1,"failed":0,"skipped":0,"incapable":0,"total":1,"state":"green","durationMs":10187}
  ✓ d6:ms-agent-python green (10.5s)

  1 passed (10.5s)
✓ Tests passed for ms-agent-python:multimodal
```

Both turns pass. aimock journal for that run: **2 entries, statuses `200,200`** (down from 8 entries with six 503s — no retries needed). **The fixture was not modified**; `git diff origin/main -- showcase/aimock/` is empty and the diff is two files, both under `showcase/integrations/ms-agent-python/`.

### CONTROL — an already-green integration, same command, same stack

```
$ bin/showcase test langgraph-python:multimodal --d6 --direct --isolate

[conversation-runner] turn 2/2 — assistant settled { bubbleIndex: 1, textLength: 233, hasAssertions: true }
[conversation-runner] turn 2/2 — assertions passed
[conversation-runner] conversation completed successfully { turnsCompleted: 2, totalDurationMs: 8395 }
  ✓ d6:langgraph-python green (9.1s)

  1 passed (9.1s)
✓ Tests passed for langgraph-python:multimodal
```

Local harness, shared probe, shared frontend and fixtures are all sound — the red was specific to this integration.

## Covering test

`showcase/integrations/ms-agent-python/tests/python/test_multimodal_pdf_prompt.py` — 7 tests. Not fakes: each one drives the real `_PdfFlattenChatMiddleware` and then the real `OpenAIChatCompletionClient._prepare_message_for_openai`, and asserts against the actual OpenAI wire payload. The PDF is the bundled `public/demo-files/sample.pdf` through real `pypdf`, and the prompt asserted on is **read out of the real aimock fixture** rather than hardcoded, so the test fails if either side drifts.

Test-level red→green (stash the source change, keep the tests):

```
# pre-fix
FAILED test_multimodal_pdf_prompt.py::test_pdf_turn_last_user_message_contains_the_prompt
FAILED test_multimodal_pdf_prompt.py::test_pdf_turn_serialises_to_a_single_user_message
FAILED test_multimodal_pdf_prompt.py::test_duplicate_pdf_parts_are_flattened_once
3 failed, 4 passed in 2.37s
```

with the primary failure reading:

```
AssertionError: expected the PDF turn to serialise to 1 user message, got 2:
  ['can you tell me what is in this demo pdf I just attached',
   '[Attached document]\nCopilotKit Quickstart\nAdd AI copilots to']
```

```
# post-fix — full integration suite (6 pre-existing CVDIAG + 7 new), CI's exact invocation
$ PYTHONPATH=".:src" python -m pytest tests/python/ -q
13 passed in 2.40s
```

Coverage: prompt survives to the final user turn; the turn stays one user message; the upstream one-message-per-`Content` split is pinned; original `contents` restored and the prompt `Content` not mutated; duplicate mirror parts flattened once; attachment-only turn still flattens; image turn left byte-identical.

## Pre-push

`validate-parity.ts` 20/20 pass · `validate-shared-symlinks.ts` no new erosion · `aimock-fixtures.test.ts` 842 pass · full `tests/python/` suite 13 pass · lefthook `lint-fix` + `commitlint` clean · Python lines ≤88 cols matching the file's existing style · no lockfile churn, two files in the diff.

## Scope

One cell, one middleware, one integration. The other five red `multimodal` cells from the same sweep have five different root causes and are not addressed here.

https://claude.ai/code/session_01PYdjeveT8Xof9TyHWMLoJr


jpr5:
### Blocked on a CI-only failure introduced by main, not by this PR

Updating this branch onto `main` (`d28384a2eb`) turned CI red. Two of the new tests fail on both 3.10 and 3.12:

```
FAILED tests/python/test_multimodal_pdf_prompt.py::test_pdf_turn_last_user_message_contains_the_prompt
  - AssertionError: real pypdf text extraction produced nothing
FAILED tests/python/test_multimodal_pdf_prompt.py::test_attachment_only_turn_still_flattens_the_document
  - AssertionError: assert 'CopilotKit' in '[Attached document]\n(unable to extract text)'
```

**This is not a defect in the fix.** The same tests passed on the pre-update head `49e7b2174c`.

Root cause: commit `51cc63d974` ("chore(showcase): store demo assets uniformly as Git LFS pointers", #6163) converted
`showcase/integrations/ms-agent-python/public/demo-files/sample.pdf` from a real 2486-byte PDF into a 129-byte Git LFS
pointer. Verified:

| ref | blob size | first bytes |
|---|---|---|
| merge-base `38613623f4` | 2486 | `%PDF-1.4` |
| old head `49e7b2174c` | 2486 | `%PDF-1.4` |
| `main` `d28384a2eb`     | 129  | `version https://` |

The `Python unit tests` job in `showcase_validate.yml` checks out **without LFS** (its checkout has only
`persist-credentials: false`), so pytest reads the pointer text and `pypdf` extracts nothing.

Image builds are unaffected — `showcase_build.yml` has carried an unconditional `lfs: true` since `7bde1eef3a`
(2026-06-07), for exactly this reason. So the deployed demo still ships a real PDF; only this CI job is blind to LFS.

**Suggested fix (not applied here — it touches a shared workflow):** add a scoped LFS fetch to the `Python unit tests`
job rather than a blanket `lfs: true`. The repo has 223 LFS objects including 13–26 MB gifs, and that job has
`timeout-minutes: 10`, so pulling everything is a timeout risk. The demo assets are 20 files at KB scale:

```yaml
- name: Fetch LFS demo assets
  run: git lfs pull --include="showcase/integrations/*/public/demo-files/*"
```

Please avoid "skip the test if the PDF looks like a pointer" — that would silently stop exercising the real
extraction path in CI, which is the thing these tests exist to cover.

The other three Wave 3 PRs (#6162, #6161, #6158) were green and have been merged. This one is held until CI is green.

PR #6170, opened 2026-07-25
test(showcase): mutation manifest + runner for the toothless-guard class

jpr5:
## Why

Across two 12-reviewer rounds on #6156, the single most effective technique was **mutation testing** — deliberately breaking the implementation to check whether the tests notice. It caught what inspection missed entirely: dropping the whole short page left all 3 autocancel tests green; lowering `MAX_INITIAL_FETCH_PAGES` from 20 to 10, and even to 5, left the bounds suite 6/6 green; `Math.max`→`min` and `every`→`any` left the golden master byte-identical.

**Three different agents wrote those mutations by hand, independently, and threw them away each time.** This commits them.

## What

`showcase/scripts/mutation-gate/mutants.json` — a declarative manifest, one entry per mutation: target file, exact find/replace, test scope, the tests that MUST fail, and the defect class it guards. Entries known to slip past their scope declare `expect: "survive"` with a mandatory `knownGap` (reason + what closes it), so the manifest is a **ledger of guard strength**, not a list of mutations someone already knew would fail.

`mutation-gate/mutation-gate.ts` — applies each mutation in isolation, runs the declared scope, asserts the declared tests fail, reverts, reports. Fails loudly when a mutation does **not** produce its expected failure, and equally when a declared gap has been closed (a stale ledger is the same disease one level up).

`__tests__/mutation-gate-manifest.test.ts` — the cheap half, riding the `showcase/scripts` vitest step that **already runs in `showcase_validate.yml`**: asserts the manifest is valid and every anchor still occurs exactly once at HEAD. That is the failure mode that would otherwise rot silently — a refactor that moves an anchor breaks no test, it just means the guard is no longer measured.

## Seeded set — 21 mutations, every outcome measured, not assumed

Harvested from `round-1-fix-{A5,FETCH,INV7,G1}`, `round-1-slot-6-a`, `round-1-integration`, `round-2-slot-{3-b,5-b,6-a,7-a}` and `convergence-audit` (lever 3).

Full run at `938be2accf`: **11 killed, 10 declared-toothless, 0 fatal, 100s wall, tree verified clean.**

| mutant | expect | verdict |
|---|---|---|
| bulk-merge/drop-short-page | kill | KILLED (3) |
| bulk-merge/no-boundary | kill | KILLED (1) |
| bulk-merge/reverse-page-order | kill | KILLED (2) |
| bulk-fetch/requestkey-autocancel | kill | KILLED (3) |
| page-cap/widened-to-100000 | kill | KILLED (2) |
| **page-cap/halved-to-10** | survive | **TOOTHLESS** |
| **page-cap/cut-to-5** | survive | **TOOTHLESS** |
| supplemental/failure-becomes-fatal | kill | KILLED (1) |
| supplemental/drop-signal-projection | kill | KILLED (3) |
| **fanout/serialized-dispatch** | survive | **TOOTHLESS** |
| polarity/pre-pr-gray-gate | kill | KILLED (11) |
| polarity/drop-redsignalknown-conjunct | survive | TOOTHLESS (by design) |
| polarity/redsignalknown-family-scope | kill | KILLED (4) |
| **goldenmaster/redsignalknown-family-scope** | survive | **TOOTHLESS** |
| **goldenmaster/failcount-min-not-max** | survive | **TOOTHLESS** |
| aggregation/failcount-min-not-max | kill | KILLED (2) |
| **aggregation/softclass-any-not-every** | survive | **TOOTHLESS** |
| **goldenmaster/starter-window-doubled** | survive | **TOOTHLESS** |
| **goldenmaster/commerror-decode-disabled** | survive | **TOOTHLESS** |
| inv7/contributingkeys-total-drift | kill | KILLED (5) |
| **inv7/contributingkeys-partial-drift** | survive | **TOOTHLESS** |

Two findings that sharpen the reports they came from:

- **`aggregation/softclass-any-not-every` is worse than slot 7a claimed.** 7a scoped the `every`→`any` gap to the 56 golden-master entries; measured across the *whole* cell-model directory (198 tests) it is still green. The gap is total, not golden-master-only.
- **Slot 6's `Promise.all` → sequential-`await` mutation is a no-op**, for the same reason it stayed green: the page promises are created eagerly at `wave.push(pb…getList(…))`, so awaiting them in sequence does not serialize dispatch. `INITIAL_FANOUT_BATCH = 1` does — and 24/24 still pass, so the "dispatched concurrently" claim really is unpinned.

## The ratchet — both sides of a fix, each measured on the fix branch

A manifest entry cannot target code that does not exist at its own HEAD, so a gap closed on an unmerged branch had nowhere to live. `ratchet` is a **second declared state** measured at a named commit; the runner checks the entry's state first and the ratchet second and reports which one the tree is in (`RATCHETED`). An outcome matching **neither** is still fatal, and both states name their failing tests exactly — so this tolerates someone else's merge without buying laxity.

I measured these by running the gate against the fix branches, not by taking their word:

- **Q1 `fix/6156-truncated-boundary` @ `67c2c8c5c4`** — `page-cap/halved-to-10` and `page-cap/cut-to-5` both go `survive → kill`. The bounds suite grows 6 → 7 tests and **3 of the 7 fail at cap 10, the same 3 at cap 5**, where all 6 passed before. (Note: 3 *tests*, not 10 — vitest aborts an `it` at its first failing `expect`, so a count of 10 is assertions, not tests.)
- **Q4 `fix/6156-fixture-coverage-polarity` @ `3c1177838f`** — `goldenmaster/redsignalknown-family-scope`, `goldenmaster/failcount-min-not-max` and `aggregation/softclass-any-not-every` each go `survive → kill`, each via **exactly one** new heterogeneous fixture (`hetero-d4-green-stripped-infra-red`, `hetero-d4-red-failcount-max`, `hetero-starter-mixed-soft-hard`). Golden master 57 → 62.
- Gate run against both branches: **0 fatal**.

`pendingSources` records what cannot be seeded yet:
- **Q2 `fix/6156-pb-evaluator-fidelity`** — still in flight, so nothing can be anchored against it. Seeds listed (single-quote tokenizing, the parse throw escaping the `node:http` handler, `applyPbSort` no-op, case-sensitive `like()`, `_` pattern anchoring, string-compared `fail_count`).
- **Q3 `fix/6156-inv7-gap-carveout` @ `49bf40943f`** — measured; it moves **none** of the 21 existing verdicts, and its carve-out does not exist at this manifest's `measuredAt`, so an entry for it would fail the anchor pre-flight. Two entries queued for when it lands, including one that pins `opts.ladderGapAllowance` as load-bearing rather than dead configuration.

## Safety — this tool edits tracked source, so all four are proven, not asserted

1. **Refuses a dirty tree** (tracked modifications; `-uno` so gitignored generated artifacts don't block). No `--force`. → exit 2, verified.
2. **Never `git stash`** (the stash stack is shared repo-wide across worktrees and holds other sessions' entries) and never commits.
3. **Verifies the repo before the first write.** Pointed at a repo whose origin is `jpr5/env`: *"REFUSING TO WRITE … origin resolves to `jpr5/env`, but the manifest expects `CopilotKit/CopilotKit`"* → exit 2, target file byte-untouched.
4. **Reverts on interrupt.** Killed mid-run with `SIGINT` while `useLiveStatus.ts` was mutated (`1 file changed, 2 deletions`): reverted synchronously, exit 130, `git status` empty, journal removed. `SIGKILL` cannot be handled, so the journal survives it — after a `-9` the mutation is stranded and the **next invocation self-heals**: *"a PREVIOUS RUN DIED with `bulk-fetch/requestkey-autocancel` applied (pid 42704) … restored; tree is clean"* → exit 2.

Backups and the journal live inside the worktree's own git dir, so they are invisible to `git status` and per-worktree.

## CI viability — and the dependency, stated plainly

**Fast enough for CI: 100s for all 21 mutants** (harness cell-model runs are ~0.9s each; the four jsdom dashboard runs dominate). Runtime is `N × suite time`, so it scales with the manifest.

**But it gates nothing until #6168 lands.** No CI job runs the showcase harness or dashboard unit suites at all — 3,646 harness tests and 1,331 dashboard tests run nowhere, excluded twice over (`paths-ignore: showcase/**` *and* `--projects='packages/**'`), and the harness is not typechecked either because `check-types` runs `nx run-many -t check-types` while the harness script is named `typecheck`. **#6168 is the prerequisite.** Until then the only part of this PR that executes in CI is the manifest-integrity test, which rides the `showcase/scripts` vitest step — worth having on its own, since it is what stops the manifest rotting, but it runs no mutations.

Recommendation once #6168 is in: a **periodic / label-triggered** job rather than per-PR. It is fast, but it is a whole-tree source-rewriting tool and per-PR is the wrong blast radius for it. The per-PR guard is the anchor test.

## Base

Targets `fix/cold-load-gray-masks-red` (#6156), not `main` — **verified, not preference**: `--check --repo <main>` reports that `cell-model.contribution.ts` and `cell-model.coherence.test.ts` are not tracked on `main` at all. The manifest measures guards that #6156 introduces, so it cannot resolve before #6156 lands.

## Verification

- prettier `--check`: clean. oxlint: 0 warnings, 0 errors on all 3 files. `tsc -p showcase/scripts/tsconfig.json`: 0 errors from `mutation-gate` (7 pre-existing errors in `equivalence-gate.ts` / `generate-search-index.ts` / `verify-prod-resweep.ts` / `angular-package.ts`, untouched). `mutation-gate/**` was added to the scripts tsconfig `include` so the new code is actually covered.
- Manifest-integrity test: 5/5 pass.
- Scoped every run away from `tests/runtime-env-switch.spike.test.ts` (confirmed: hangs ~25 min at 0.0% CPU, never spawns its `next build`); the dashboard suite entry documents this and the scope stays under `src/`.
- `showcase/scripts` full suite: 13 pre-existing failures in `__tests__/generate-registry-pattern.test.ts`. **Not mine** — a base worktree at `938be2accf` with matched symlinks/artifacts gives the identical `13 failed | 1 passed`.

## Unverified

- Never run in CI (see above) — timings are local, on a machine with other agents active.
- One dashboard run died at 161s without writing its JSON report under heavy load. That is why the runner **retries a non-measurement once** (never a real `SURVIVED`) and why the dashboard timeout is 10 min. Both FATALs in the first full run were this, and both cleared on re-run.
- `pendingSources` seeds for Q2/Q3 are written from their reports, not measured — by construction, since neither can be anchored yet.

https://claude.ai/code/session_01PYdjeveT8Xof9TyHWMLoJr

### DIFF
--- a/showcase/scripts/__tests__/mutation-gate-manifest.test.ts
+++ b/showcase/scripts/__tests__/mutation-gate-manifest.test.ts
@@ -1,0 +1,92 @@
+/**
+ * The CHEAP half of the mutation gate, run by `showcase_validate.yml`'s
+ * existing `showcase/scripts` vitest step.
+ *
+ * The gate itself (applying mutations and re-running suites) needs the
+ * `shell-dashboard` and `harness` packages installed and their pretest
+ * artifacts generated, and no CI job runs those unit suites yet — see
+ * `mutation-gate/mutation-gate.ts` and the report for that dependency. What
+ * DOES gate today is manifest integrity:
+ *
+ *   1. the manifest is structurally valid, and
+ *   2. every mutation's anchor string still occurs EXACTLY ONCE in its target
+ *      file at HEAD.
+ *
+ * (2) is the failure mode that would otherwise rot silently. A refactor that
+ * moves `const lastIdx = …` does not break any test — it just means the
+ * mutation for that guard no longer applies, and the day someone runs the gate
+ * they get a green report from a manifest that tests nothing. Catching it here
+ * costs milliseconds and no test execution.
+ */
+
+import { execFileSync } from "node:child_process";
+import path from "node:path";
+import { describe, expect, it } from "vitest";
+
+import {
+  describeAnchorProblem,
+  loadManifest,
+  resolveAnchors,
+  validateManifest,
+} from "../mutation-gate/manifest";
+
+const REPO_ROOT = execFileSync("git", ["rev-parse", "--show-toplevel"], {
+  cwd: import.meta.dirname,
+  encoding: "utf8",
+}).trim();
+
+describe("mutation-gate manifest", () => {
+  const manifest = loadManifest();
+
+  it("is structurally valid", () => {
+    expect(validateManifest(manifest)).toEqual([]);
+  });
+
+  it("declares at least one known-toothless guard, so the ledger is honest", () => {
+    // A manifest in which every mutation is killed is either a genuinely
+    // airtight suite or — far more likely — a manifest that only records the
+    // mutations someone already knew would fail. The seeded set deliberately
+    // carries the survivors too; if this ever drops to zero, confirm it is
+    // because the gaps were CLOSED (each `survive` entry promoted to `kill`)
+    // and not because they were deleted.
+    const survivors = manifest.mutants.filter((m) => m.expect === "survive");
+    expect(survivors.length).toBeGreaterThan(0);
+  });
+
+  it("every anchor resolves exactly once against HEAD", () => {
+    const problems = resolveAnchors(manifest, REPO_ROOT);
+    expect(
+      problems.map(describeAnchorProblem).join("\n"),
+      problems.length === 0
+        ? ""
+        : `mutation-gate/mutants.json has rotted against the current source. ` +
+            `Each mutation below no longer applies, so the guard it checks is ` +
+            `unmeasured. Re-anchor or delete the entry:`,
+    ).toBe("");
+  });
+
+  it("every target file is inside showcase/ and tracked by git", () => {
+    const files = [
+      ...new Set(manifest.mutants.flatMap((m) => m.edits.map((e) => e.file))),
+    ];
+    for (const f of files) {
+      expect(f.startsWith("showcase/"), `${f} is outside showcase/`).toBe(true);
+      expect(() =>
+        execFileSync("git", ["ls-files", "--error-unmatch", f], {
+          cwd: REPO_ROOT,
+          stdio: "ignore",
+        }),
+      ).not.toThrow();
+    }
+  });
+
+  it("every declared suite package exists and names a vitest binary path", () => {
+    for (const [name, s] of Object.entries(manifest.suites)) {
+      expect(
+        path.isAbsolute(s.package),
+        `suite ${name}: \`package\` must be repo-relative`,
+      ).toBe(false);
+      expect(s.vitestBin, `suite ${name}`).toMatch(/vitest$/);
+    }
+  });
+});
--- a/showcase/scripts/mutation-gate/manifest.ts
+++ b/showcase/scripts/mutation-gate/manifest.ts
@@ -90,8 +90,37 @@ export interface Mutant {
    */
   mustFail?: string[];
   knownGap?: MutantKnownGap;
+  ratchet?: MutantRatchet;
   /** Optional per-mutant override of the suite timeout, in milliseconds. */
   timeoutMs?: number;
+}
+
+/**
+ * A SECOND declared state for the same mutation, measured at a different commit.
+ *
+ * A manifest entry cannot target code that does not exist at its own HEAD (the
+ * anchor pre-flight refuses), so an entry seeded from a base branch records the
+ * base's outcome in `expect` — while the fix that closes the gap is already
+ * measured on its own branch. `ratchet` carries that second measurement, so the
+ * manifest documents the RATCHET rather than only the current state, and the
+ * gate stays truthful on both sides of the merge.
+ *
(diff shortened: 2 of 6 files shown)


jpr5:
CI green on this draft: 14 pass, 0 pending, 0 fail.

Worth noting one thing that resolves an ambiguity in the body: locally, the full `showcase/scripts` suite shows 13 failures in `__tests__/generate-registry-pattern.test.ts`. `Validate Showcase` — the job that runs that suite — **passes here**, and a base worktree at `938be2accf` reproduces the identical `13 failed | 1 passed`. So those are an artefact of a symlinked worktree without a real `pnpm install` (the tests spawn `npx tsx` subprocesses), not a pre-existing CI failure and not from this branch.

Staying draft pending #6168 and the #6156 fix units.

PR #6173, opened 2026-07-25
fix(react-core): stop CopilotPopup remounting chat on resize

### LINKED ISSUE #6172: 🐛 Bug: CopilotPopup rescrolls chat from top to bottom on every resize
### ♻️ Reproduction Steps

1. Render a `<CopilotPopup>` (v2, `@copilotkit/react-core/v2`) with a conversation long enough to scroll.
2. Drive its `width`/`height` props from a drag-to-resize handle — a controlled value that changes when the user resizes the popup (e.g. committing the new dimensions to React state on `mouseup`):
   ```tsx
   const { width, height, handleProps } = useResize(...)
   return <CopilotPopup width={width} height={height} /* ... */ />
   ```
3. Scroll the message list to any position (top, middle, or bottom).
4. Resize the popup via the handle.

### ✅ Expected Behavior

Resizing the popup should keep the current scroll position. The message list should not jump.

### ❌ Actual Behavior

On every resize the entire chat subtree unmounts and remounts. The scroll container's `scrollTop` resets to `0`, then the default `initial="smooth"` auto-scroll animates the list from the top back down to the bottom — a visible "rescroll from top to bottom" on each resize, regardless of where the user was scrolled.

**Root cause:** `CopilotPopup` builds its `chatView` override component inside a `useMemo` keyed on `width`/`height` (`packages/react-core/src/v2/components/chat/CopilotPopup.tsx`). When `width`/`height` change, `useMemo` returns a **new component function**, which is passed to `CopilotChat` as the `chatView` slot and rendered via `React.createElement(newType, …)`. Because React sees a new element **type** at that position, it unmounts and remounts the whole subtree instead of re-rendering it, which resets scroll state.

This only surfaces when `width`/`height` are *controlled and change over time* (i.e. a resizable popup). Static-size popups never hit it because the memoized component identity stays stable.

A fix is prepared and attached as a PR (stable module-scope override identity + shell props passed through React context, plus a regression test).

### Screenshots

_N/A — behavioral. The list visibly animates top→bottom on each resize._

### 𝌚 CopilotKit Version

```shell
@copilotkit/react-core@1.62.x   # also reproduces on 1.63.x main; the useMemo predates both
```

### 📄 Logs (Optional)

```shell
No console errors — it is a silent remount. Confirmable by adding
`useEffect(() => console.log("mount"), [])` to any component inside the
popup's chat subtree: it logs once per resize.
```

Vishal0203:
## What does this PR do?

Fixes a bug where resizing a `CopilotPopup` (v2) rescrolls the message list from top to bottom on every resize, regardless of the user's scroll position.

**Root cause.** `CopilotPopup` built its `chatView` override component inside a `useMemo` keyed on `width`/`height`. When those props change — which happens continuously for consumers driving them from a drag-to-resize handle that commits new dimensions on `mouseup` — `useMemo` returns a **new component function**. That function is passed to `CopilotChat` as the `chatView` slot and rendered via `React.createElement(newType, …)`. Because React sees a new element **type** at that position, it unmounts and remounts the entire chat subtree rather than re-rendering it. The remount resets the scroll container's `scrollTop` to `0`, after which the default `initial="smooth"` auto-scroll animates the list from the top back down to the bottom — the visible "rescroll" on every resize.

**Fix.** Give the override a stable module-scope identity so its type never changes, and pass the popup shell props (`header`, `toggleButton`, `width`, `height`, `clickOutsideToClose`, `defaultOpen`) through React context instead of baking them into the override via `useMemo` deps. Resizing is now a plain style update on `CopilotPopupView` — no remount — so scroll position is preserved. `Object.assign` still copies the `CopilotChatView` static slot members onto the override so callers reaching them through the namespace continue to work.

**Test.** Adds a regression test (`CopilotPopup.resizeRemount.test.tsx`) that injects a mount-counting component into the chat subtree and asserts its mount count stays at `1` across several `width`/`height` changes. It fails on the pre-fix code (one extra mount per resize — observed `4` for 3 resizes) and passes with the fix.

Scope is limited to `packages/react-core`:

```
packages/react-core/src/v2/components/chat/CopilotPopup.tsx                               | 124 +++++++++------
packages/react-core/src/v2/components/chat/__tests__/CopilotPopup.resizeRemount.test.tsx  |  98 +++++++++++
2 files changed, 185 insertions(+), 37 deletions(-)
```

Verification: full `@copilotkit/react-core` unit suite green (1443 tests), `check-types` green.

## Related PRs and Issues

- Closes #6172

## Checklist

- [ ] If the PR changes or adds functionality, I have updated the relevant documentation — _N/A: internal behavior fix, no public API or documented behavior change_

### DIFF
--- a/packages/react-core/src/v2/components/chat/CopilotPopup.tsx
+++ b/packages/react-core/src/v2/components/chat/CopilotPopup.tsx
@@ -1,10 +1,74 @@
-import React, { useEffect, useMemo } from "react";
+import React, { useContext, useEffect, useMemo } from "react";
 import { useLicenseContext } from "../../providers/CopilotKitProvider";
 import { InlineFeatureWarning } from "../license-warning-banner";
 
-import { CopilotChat, CopilotChatProps } from "./CopilotChat";
-import CopilotChatView, { CopilotChatViewProps } from "./CopilotChatView";
-import CopilotPopupView, { CopilotPopupViewProps } from "./CopilotPopupView";
+import type { CopilotChatProps } from "./CopilotChat";
+import { CopilotChat } from "./CopilotChat";
+import type { CopilotChatViewProps } from "./CopilotChatView";
+import CopilotChatView from "./CopilotChatView";
+import type { CopilotPopupViewProps } from "./CopilotPopupView";
+import CopilotPopupView from "./CopilotPopupView";
+
+/**
+ * Carries the popup shell props (header, toggle, width, height, …) to the
+ * chatView override without baking them into the override's component
+ * identity. Width/height change on every resize; if they were `useMemo`
+ * dependencies of the override component, each change would mint a new
+ * component function, and rendering a new element *type* makes React unmount
+ * and remount the entire chat subtree — resetting scrollTop to 0 and firing
+ * `initial="smooth"`, which visibly scrolls the message list from top to
+ * bottom on every resize. Routing them through context keeps the override
+ * identity stable (no remount) while still re-rendering CopilotPopupView with
+ * the new dimensions on change.
+ */
+type PopupShellProps = {
+  header?: CopilotPopupViewProps["header"];
+  toggleButton?: CopilotPopupViewProps["toggleButton"];
+  width?: CopilotPopupViewProps["width"];
+  height?: CopilotPopupViewProps["height"];
+  clickOutsideToClose?: CopilotPopupViewProps["clickOutsideToClose"];
+  defaultOpen?: boolean;
+};
+
+const PopupShellPropsContext = React.createContext<PopupShellProps>({});
+
+// Stable override component. Its identity never changes, so the chat subtree
+// is never remounted on resize. Popup props flow in via context, so changing
+// width/height re-renders CopilotPopupView (a style update) instead.
+const PopupViewOverride: React.FC<CopilotChatViewProps> = (viewProps) => {
+  const {
+    header: viewHeader,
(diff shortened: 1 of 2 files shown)


BenTaylorDev (approved):
**Approving** (scoped to #6172, Popup). Verified against current `main`: the remount root cause is real — `CopilotPopup.tsx:39-64` builds `PopupViewOverride` inside a `useMemo`, so any dep change (a resize commit, or an inline `header`/`toggleButton`) yields a new component *type* at the `chatView` slot, and `renderSlot`'s comparator (`slots.tsx:95`) remounts the whole chat subtree → `scrollTop` reset → the top→bottom re-animation. The fix hoists the override to module scope (stable `$slot` identity) and flows resize props via `PopupShellPropsContext` — and since `React.memo` doesn't block context propagation to a `useContext` consumer, content still updates while the remount stops (no over-memoization/staleness). Open/close/portal untouched.

Verified locally on the updated branch: the new `CopilotPopup.resizeRemount.test.tsx` (real Popup→Chat→slot path, mount-counter via the `input` slot) passes, the full `@copilotkit/react-core` suite is green (the only failure is an unrelated pre-existing stale-dist artifact in `use-interrupt`), `tsc` clean on the changed file, lint 0.

Note: `CopilotSidebar.tsx:39-63` has the identical `useMemo`-override pattern and the same latent remount bug — filing a follow-up to apply the same fix there; out of scope for this Popup-focused PR.

PR #6174, opened 2026-07-26
fix(showcase/ci): make the :latest tag monotonic so an older build cannot overwrite a newer one

jpr5:
## The race, with evidence

Three PRs merged within 34 seconds on 2026-07-26. `Showcase: Build & Push` has **no concurrency group** (deliberately), so all three ran simultaneously and raced to push the same `:latest` tags.

| run | commit | PR | start → end |
|---|---|---|---|
| `30190815370` | `7b28934387` | #6162 | 06:18:05 → 06:29:44 |
| `30190823203` | `59f275eedc` | #6161 | 06:18:21 → 06:29:44 |
| `30190831480` | `db75a04837` | #6158 | 06:18:39 → **06:29:13** ← newest, finished FIRST |

The newest commit finished first, so the older builds overwrote its `:latest`. Per-service job completion times — older beating newer on every shared slot:

| service | newer (`db75a04837`) | older (`59f275eedc`) | older won by |
|---|---|---|---|
| `shell-dashboard` | 06:25:33 | 06:25:34 | +1s |
| `showcase-harness` | 06:27:51 | 06:27:55 | +4s |
| `shell` | 06:27:17 | 06:27:28 | +11s |
| `shell-dojo` | 06:25:10 | 06:25:26 | +16s |

**All three runs reported `success`.** Staging served pre-#6158 code while CI, the redeploy gate and deploy verification all looked clean. Same failure class as #6171: a success that doesn't mean what it says.

## What I verified in YAML vs took on trust

Verified by reading the files / querying the API:

- **Tagging** — `showcase_build.yml` pushed `:latest` **and** `:${{ github.sha }}` in one `depot/build-push-action` step, in both the `build` and `build-starters` matrices. **A per-commit sha tag already existed**; confirmed in GHCR (`showcase-shell-dashboard` has digests tagged `db75a04837…`, `59f275eedc…`, `d28384a2eb…`).
- **Nothing serialized the pushes.** No concurrency group; confirmed the header comment states this is intentional.
- **Deploy consumes `:latest`** — `verify-railway-image-refs.ts` is the SSOT assertion: staging is `ghcr.io/copilotkit/<repo>:latest` (mutable), **prod is `ghcr.io/copilotkit/<repo>@sha256:<digest>` (already immutably pinned)**. So this race is **staging-only**; prod was never exposed.
- **`Showcase: Verify Deploy` structurally cannot catch it.** It is a health probe; it asserts no digest or commit provenance anywhere. Its #6171 per-commit concurrency key is about *which run verifies*, not *what image is running*. A stale-but-healthy service passes.
- **The racing runs build DISJOINT service sets** (see below) — I pulled the actual job lists.

Taken on trust: nothing material. The issue description matched the API on every point I checked.

## Why NOT a concurrency group

`detect-changes` builds a **per-push, path-filtered** matrix, so concurrent runs build overlapping but **non-identical** service sets:

- `7b28934387` → ag2, agno, built-in-agent, claude-sdk-python, claude-sdk-typescript, crewai-crews, langgraph-fastapi, langgraph-python, langroid, llamaindex, mastra, pydantic-ai, shell-docs, spring-ai, strands (**15**)
- `db75a04837` → crewai-crews, llamaindex, shell, shell-dashboard, shell-docs, shell-dojo, showcase-harness (**7**)

`cancel-in-progress: true` would have cancelled the `7b28934387` run and the ~10 services **only it builds would never have shipped** — trading a stale-image bug for a never-shipped bug. Per [GitHub's docs](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#concurrency), `cancel-in-progress: false` is no better: *"Any previously pending job or workflow in the concurrency group will be canceled"* — with three rapid merges the **middle** commit's build is dropped outright. GitHub also does not guarantee FIFO ordering of queued runs.

Concurrent runs here are **not redundant**, so they must not be cancelled.

**Re the #6171 interaction:** this design does **not** use `cancel-in-progress`, so no build is ever superseded-and-cancelled and the cancelled-slot notifier is never tripped by this change. That interaction stays theoretical — deliberately.

## The design

Make the one shared mutable resource monotonic instead of serializing the fleet.

1. The build step pushes **only** the immutable `:<sha>` tag, plus an `org.opencontainers.image.revision` label.
2. A guard resolves the commit behind the current `:latest`, asks GitHub `compare/<theirs>...<ours>`, and advances `:latest` (registry-side retag, no pull) **unless ours is `behind`** — i.e. `:latest` already holds a descendant and moving it would roll staging back.

**Fails open by design.** No `:latest` yet, unlabelled legacy image, unreachable API, diverged history → advance. A stuck `:latest` is the very failure being fixed, so it declines only on *positive proof* of regression.

**Placement:** in `redeploy-staging` / `redeploy-staging-starters`, immediately before the Railway pull that consumes `:latest` — not per build slot. Those jobs already have Node (build slots do **not**, so per-slot would mean an unpinned `npx tsx` fetch on ~50 parallel runners), and deciding right before the pull makes the window as narrow as possible. The image list is the **same matrix ∩ build-success intersection** that decides what gets redeployed, so a failed build can never move a tag.

Also adds `showcase_build.yml` to `showcase_validate.yml`'s trigger paths — the new test asserts against that file's live text, and without the path a PR re-adding `:latest` would never run the test that catches it.

## Tradeoffs / what stays open

- **Residual sub-second TOCTOU.** GHCR has no compare-and-swap on tags, so two runs reading `:latest` simultaneously could still both advance. This narrows the window from the whole build (~10 min) to inspect→retag. Fully closing it means retiring the mutable staging tag and pinning staging to digests the way prod already is — a change to the Railway image-ref SSOT contract, not a workflow change. **Recommended follow-up.**
- **The guard is inert for one build per image.** Today's `:latest` images carry no labels (verified: `showcase-shell-dashboard:latest` has no `Labels` at all), so the first post-merge build fails open and advances unconditionally — same as today. Protection starts from the second build of each image.
- Failure to retag exits non-zero, redding the redeploy job and stopping the deploy. That is intended: redeploying against a tag that did not move is exactly the silent false green being fixed.

## Proof

**Red/green on the live YAML.** `advance-latest-tag.test.ts` parses the real `showcase_build.yml` (extending #6171's `redeploy-guard.test.ts` pattern). Reverting the workflow to its pre-fix state: **14 failed / 23 passed**. With the fix: **37 passed**. Full `showcase/scripts` suite: **2382 passed, 73 files**.

**The load-bearing predicate, verified live against the real incident commits:**

```
compare/db75a04837...59f275eedc  => behind     (older run arriving late → DECLINE)
compare/59f275eedc...db75a04837  => ahead      (newer run → advance)
compare/db75a04837...db75a04837  => identical
```

**Label reading, verified against a real multi-platform registry image** — `docker buildx imagetools inspect ghcr.io/astral-sh/uv:latest` piped through `extractRevisionLabel()` returns `3010295ae7ff572de459987ad70db315a62ecd61`, matching `jq` exactly. The platform-keyed shape is handled.

**Shell/jq transforms** exercised directly, including the empty-CSV edge case (empty → empty, step skipped by its `if:`).

**Lint:** `actionlint` finding counts byte-identical to the pre-change baseline (no new findings; the 11 pre-existing are unrelated). `zizmor --min-severity low` with the repo config: **no findings**.

**Typecheck:** both new files are in `showcase/scripts/tsconfig.json`'s include set and produce **zero** errors. Worth stating plainly: `nx run-many -t check-types` **does not reach `showcase/scripts`** — the project isn't in the nx graph and has no `check-types` target (there are 9 pre-existing type errors in sibling files, which is how I confirmed it). So the typecheck above is mine, not CI's. The *tests* are gated: `showcase_validate.yml` runs bare `pnpm exec vitest run` in `showcase/scripts`, which auto-discovers the new file.

### What I could NOT prove

**I did not construct a real concurrent race on scratch branches.** Doing it faithfully needs two builds pushing the same GHCR repo with controlled finish ordering, which means merging to `main` — the only branch the build workflow triggers on. No run IDs for a live race demonstration; I am not implying one.

Unproven until this runs on main: that `docker buildx imagetools create` retags cleanly under the runner's GHCR credentials, and that `npx tsx` behaves in the redeploy jobs (it is already the established invocation there — `redeploy-env.ts` — so this is low risk, not zero).

## Normal single-merge builds are unaffected

No concurrency group is added, so nothing queues or cancels. A lone merge finds `:latest` at its own parent → `ahead` → advances, exactly as before. Cost is one `imagetools inspect` + one `gh api` + one registry-side retag per built service, in a job that already exists — no extra job, no extra checkout, no change to build parallelism.

---

Branched from `db75a04837`; #6156/#6159 landed after, so this will need main merged in before it goes green.

Probable conflict with the concurrent `git lfs pull` work in `showcase_validate.yml` — my edit there is only the top-level `on: paths:` list, so it should merge cleanly, but flagging it.


jpr5:
## Parking this as a draft — record of two review rounds

This PR is being parked, not abandoned. It went through two full 7-reviewer rounds; both found Critical defects, and rather than run a third fix round we're leaving the findings on the PR so whoever picks it up starts from them. What follows is the synthesis of all 14 reports.

### What this set out to fix, and why it's still real

Concurrent merges to `main` could move `ghcr.io/copilotkit/<image>:latest` **backward**: two runs build in parallel, the older one's push lands last, and `:latest` ends up pointing at the older commit's image. Staging then redeploys onto code that was already superseded. This was observed live, and the ugly part is that all three runs involved reported `success` — there was no red anywhere, because nothing in the system compared the incoming commit against the one `:latest` already held.

The fix: stop the build matrix from pushing `:latest` at all (push only the immutable `:<sha>` plus an `org.opencontainers.image.revision` label), and move `:latest` advancement into a guard step (`showcase/scripts/advance-latest-tag.ts`) that reads the revision label off the current `:latest`, asks the GitHub compare API whether our sha is *behind* it, and declines the retag if so.

Mitigating fact that made parking reasonable: **prod is digest-pinned** (`verify-railway-image-refs.ts` enforces `@sha256:` refs), so a backward `:latest` is a staging-only problem. Manually serializing merges has been sufficient in practice.

### What's sound and should be kept

**The decision core.** A round-1 reviewer assessed `decideLatestTagAction` as genuinely well designed: pure, total over its input space, with the safety property stated as an iff ("`decline-regression` if and only if we have POSITIVE proof of regression") and then pinned by a test that filters the full compare-status set and asserts the declining set is exactly `["behind"]` — that pins the *complement* of the rule, so a new status added to the enum cannot silently start blocking deploys. The compare direction (`compare/<current>...<ours>`, base = what `:latest` holds) is correct per GitHub's semantics and separately pinned, because inverting it silently inverts the entire guard.

**The workflow-wiring assertions.** Round 2 mutation-tested these and they held: **9 of 9 wiring mutations were caught** — inverting the compare direction, also declining on `diverged`, skipping the digest probe, re-adding `:latest` to the build push step, renaming the parser's label constant, removing `packages: write` from `redeploy-staging`, having the `changed` step ignore the build-success set, dropping the `skip_build` exclusion from the `images` list, and moving the guard step after the Railway redeploy. Asserting against the live YAML rather than a mock is the right instinct and it demonstrably works. The related move of executing the real `changed` shell against the real matrix (instead of grepping the jq) is the correct pattern and should survive any rebuild.

**The digest-preservation investigation — do not lose this.** This is the most valuable artifact in the PR. The question was whether `docker buildx imagetools create -t X:latest X:<sha>` preserves the manifest digest, which matters because prod provenance ("which commit is live") is established by matching a running digest against GHCR's per-commit tags. Findings: buildx's `imagetools create` returns the *original bytes* when the source is an index (verified against `util/imagetools/create.go`, the single-source branch); our images are indexes only because Depot attaches a provenance attestation; and so `--prefer-index=false` was added to the retag so the property stops being contingent on that attestation continuing to exist. If someone later disables provenance on the Depot build, the flag is what keeps `:latest` and `:<sha>` sharing a digest. That chain of reasoning took real work and is not recoverable from the diff alone.

Also worth keeping: the incident reconstruction in the file header (run IDs, per-service timestamps, deltas) and the section explaining *why* a `concurrency` group is the wrong fix — it pre-empts the first objection any reviewer raises.

---

### The blockers, grouped by root cause

**1. Timeout composition — the per-call bound doesn't add up to a job bound, and the overrun is unalertable.** Flagged independently by three round-2 reviewers (and by three round-1 reviewers in its earlier form, when there was no timeout at all).

`SUBPROCESS_TIMEOUT_MS` defaults to 60s **per subprocess**, and `advanceLatestTags` walks the fleet strictly serially with no aggregate deadline. The workflow's own sizing comment says ~28 slots × ~3 calls = ~84 serial calls. 84 × 60s = **~84 minutes** against `timeout-minutes: 20`. It doesn't take a full outage — ~20 hung calls exhausts the budget, and a merely degraded registry averaging 10s/call gets to ~14 minutes before `setup-node`, the `npx` fetch, and the Railway redeploy loop that follows.

The reason this is Critical rather than a perf nit: a job killed by `timeout-minutes` concludes `cancelled`, and this workflow has *measured* (probe run 30166429073, documented inline) that a cancelled ancestor is invisible to both `failure()` and `cancelled()`. Reviewers traced every notifier: `notify` doesn't fire (`failure()` is false for a cancelled ancestor, `any_success` is `'true'`, `any_cancelled` describes build slots), `notify-cancelled-builds` doesn't fire (keyed on build-slot artifacts), `notify-all-builds-failed` doesn't fire. Net state: `:latest` advanced for the first N images and stale for the rest, no Railway redeploy, no `redeploy-summary` artifact (so `showcase_deploy.yml` reads "nothing redeployed" and skips its gate), and **not one Slack message**. That is the same silent-partial-failure class as the original incident, relocated one job downstream — and this PR is what creates it, since before the diff these jobs only POSTed to Railway.

Suggested shape: a step-level `timeout-minutes` (an overrun *fails*, which `failure()` can see) plus an in-script wall-clock deadline under it, so the script reports the overrun in its own voice instead of dying mid-loop.

**2. `defaultIo` — the entire registry/API I/O layer — is exercised by no test.** Two reviewers independently mutation-tested it; both landed on the same conclusion from different mutation sets.

The suite imports 11 symbols; `defaultIo` is not one of them, and nothing else reaches those ~120 lines. Mutations that leave the suite **109/109 green**:

- **Retag pushes `:stale` instead of `:latest`.** `:latest` never moves again, and because the build step no longer pushes it, staging freezes permanently on whatever image it last held. *That mutation is the original P0 this PR exists to fix, shipping green.*
- Remove `--prefer-index=false` — kills the digest-identity invariant described above, and with it prod's provenance-by-digest.
- Change the inspect template `--format '{{json .Image}}'` → `'{{json .Manifest}}'` — the parser then never finds a revision label, `readLatestRevision` returns `null` forever, `null` ADVANCES, and the guard is permanently blind. The test file's own header claims the fixtures exist specifically to rule this out; they pin the *parser* against real payloads but nothing pins the *command that produces them*.
- Swap the digest template `{{.Manifest.Digest}}` → `{{.Config.Digest}}` — the same-commit-rebuild detection silently compares the wrong thing.
- Inspect the wrong ref in `readLatestRevision`.
- Degrade a failed-read `::warning` to a `::notice` — defeats the module's stated "ambiguity advances, but never silently" doctrine.
- Change `SUBPROCESS_TIMEOUT_MS` default from `60_000` to `1`.

Fix direction: export the argv builders as pure functions and assert the exact arrays, and/or give `defaultIo` an injectable exec seam. Bind the inspect template to the fixtures explicitly so fixture and command cannot drift.

**3. Digest identity is load-bearing but never verified, and the tool it depends on is unpinned.** The retag discards its output and returns `void`. Nothing reads back `:latest` and `:<sha>` to confirm they name the same manifest — even though `readDigest` is already implemented and it's one round-trip on an operation that already made three. Meanwhile neither redeploy job uses `docker/setup-buildx-action`, so the buildx version is whatever `ubuntu-latest` ships that week. The moment that behaviour changes, `verify-railway-image-refs.ts`'s premise quietly stops holding and `showcase_reconcile_staging.yml` starts remediating drift that isn't real — with zero signal. This is the most expensive failure class in the change: not a missed deploy, but a *wrong answer to "which commit is live in prod"*, surfacing months later during an incident.

Related: the code documents the `--prefer-index` floor as buildx **v0.16**; the flag actually landed in **v0.17.0**. The floor is also unenforced, and on an older buildx `--prefer-index=false` fails with `unknown flag`, which the classifier buckets as `unknown`, every retag fails, and (per #5 below) the whole fleet's redeploy is blocked by a one-word CLI error.

**4. Command injection via the revision label.** `extractRevisionLabel` accepts **any** non-empty trimmed string as `currentRevision`. That value comes off the `org.opencontainers.image.revision` label of whatever image currently holds `:latest` — the same trust class the file explicitly treats as attacker-adjacent for subprocess stderr. It then reaches two sinks:

- Interpolated raw into the API path: `gh api repos/${repo}/compare/${base}...${head}`. `execFileSync` rules out shell injection but not *path* injection — a label containing `/`, `?`, `#`, or `../` reshapes the request.
- Written to stdout **unescaped** on the advance path. Only the decline branch goes through `annotation()`; `io.log(decision.reason)` does not, and every non-decline reason embeds `${currentRevision}` verbatim. `defaultIo.log` is `process.stdout.write(msg + "\n")`, and a workflow command only has to *start* a line — so a newline in the label forges a workflow command. That's precisely the forgery vector the file goes to some length to close for stderr, left open on the label path.

Fix is cheap and preserves fail-open: require `/^[0-9a-f]{7,40}$/i` in the extractor, so a junk label yields `null`, which already means "unknown → advance."

**5. Fail-closed blast radius.** `advanceLatestTags` deliberately attempts every image and collects failures — correct. But `main` turns any non-empty `failures` into `process.exit(1)`, which reds the step and skips `Redeploy changed services in staging` for the *whole fleet*. In a 20-image run where 19 retags succeed and the 20th 500s, 19 images now have `:latest` pointing at the new build and Railway is never told to pull any of them: **`:latest` is ahead of running staging**, and the divergence is only caught later and indirectly. The step comment's justification ("redeploying against a `:latest` that did not move would look like a fresh healthy deploy") is sound for the image that failed and wrong for the 19 that didn't. Compounding it, the failure summary prints `Succeeded for 19` when the practical outcome is "0 deployed", and folds `declined` — a deliberate stand-down — into the success count.

**6. Architectural: moving `:latest` advancement into `redeploy-staging` blinds the self-heal.** Before this change the build matrix pushed `:latest` at build time, which made `:latest` an *independent* record of "what the newest successful build produced." That independence is exactly what `showcase_reconcile_staging.yml` relies on: every 15 minutes it compares each staging service's actually-deployed digest against the GHCR `:latest` digest and re-runs the redeploy when the deployed digest lags.

After this change, `:latest` only advances inside the same job that performs the redeploy. If `redeploy-staging` dies *after* the builds succeeded, `:latest` and staging lag the freshly-pushed `:<sha>` **together**, so `deployedDigest === latestDigest`, and reconcile classifies every service as "positively confirmed current" and exits 0. The 15-minute watchdog that used to catch this class no longer can — and this diff *adds* new ways for the job to die before the redeploy (the new GHCR login, the new guard's deliberate non-zero exit, and the 20-minute timeout whose overrun is the invisible `cancelled` from #1). The run does go red on most of those paths, so it isn't a silent false green, but the automated remediation path is gone and the residual state is "GHCR holds a newer `:<sha>` that no watchdog will ever notice." Options raised: teach `reconcile-staging.ts` to compare against the newest built `:<sha>`, or run the guard in its own job depending on `aggregate-build-results` and independent of the Railway redeploy.

**7. Error-classification bugs.** The classifier is the mechanism by which a degraded guard is supposed to be able to *say* what happened, so wrong answers here are real defects, not cosmetics:

- Bare `429` in the throttling regex, while every other HTTP code is word-bounded (`\b401\b`, `\b403\b`, `\b404\b`, `\b50[0234]\b`). Registry stderr routinely echoes `sha256:` digests, and hex digests contain `429` often enough to matter. Because `throttled` is tested *before* `absent`, a genuine `404` on an image whose digest happens to contain `429` is reported as rate-limiting — a false incident signal claiming the exact concurrent-build pressure this guard exists to handle.
- Any `SIGTERM` classified as `timeout`. GitHub signals the whole process tree on cancellation, and the OOM killer reaches children the same way, so those get the confident and wrong message "the command exceeded its 60000ms budget" — in precisely the scenario #1 describes.
- The `unauthorized` alternation includes the bare substring `login`, and runs *ahead* of the `absent` check. Docker and registries append "you may need to run `docker login`" to not-found errors, so a genuine first build on a new image repo reports an auth failure and sends an operator to inspect `packages: write` and the login step when both are fine. Repeated false warnings are how a `::warning`-based guard gets tuned out, which defeats the file's own thesis.
- Additionally: `describeProbeFailure`'s text is written for the `:latest`-inspect call site but reused verbatim at three others, so a `gh api .../compare` 404 (force-push, deleted ref) renders as "the tag does not exist yet" — there is no tag in a compare — and a `gh api` 401 tells the operator to widen GHCR package permissions when the actual knob is `GH_TOKEN` / `contents: read`. Also: `ENOENT` (docker or gh not on PATH) classifies as `unknown`, and a malformed `ADVANCE_LATEST_TAG_TIMEOUT_MS` (`60s`, `60_000`, `abc`) silently falls back to 60000 — the one place in the file where a degraded configuration does not announce itself.

**8. Doc drift introduced across concurrent fixes.** Both of these describe the code as it was two fix rounds ago and will actively mislead the next reader:

- The script's "where this runs" section says these jobs invoke it through an **unpinned `npx tsx`** "exactly as the sibling `redeploy-env.ts` step does", and that pinning is "tracked separately." Both live call sites run `npx --yes tsx@4.21.0` — pinned, with an eight-line rationale block on the step. The `redeploy-env.ts` half is still true, which makes the wrong half easier to believe.
- Two places state the image list is "the SAME matrix ∩ build-success intersection that decides what gets redeployed." The caller applies an additional `map(select(.skip_build != true))` to `images` that `services` does not get, and the workflow documents that exclusion as load-bearing (`webhooks` is released by another repo, concludes `success`, and has no `:<sha>` from this run, so handing it to the guard fails with `manifest unknown` and blocks the whole fleet). The actual contract is `images = (matrix ∩ build-success) − skip_build`. A maintainer reading only the script would reasonably "simplify" the jq and reintroduce that outage.

Smaller items in the same drift class: a test comment claims a re-serialising formatter pass "would reorder keys" (it doesn't — the commit that added the comment says so itself; the real justification is byte-exact provenance); a fixture is described as having `config.Labels` **null** when the key is **absent**, and the assertion's `?? null` makes the two indistinguishable so nothing catches it; and a test row labelled "identical trees under different shas" describes something GitHub's compare API cannot return (`identical` means the same commit).

### Also open, smaller

`main()` and the entire documented exit-code contract have zero coverage (exit 2 on a valueless flag, exit 1 on partial retag failure — the value that stops a stale redeploy — exit 0 on empty `--images`, and the `--images` > `--image` > `$IMAGES` precedence chain). The workflow-wide `:latest` ban regexes are single-line-only, so a backslash-continued `docker buildx imagetools create \` evades all three, as does `docker buildx build --push -t …:latest` and `crane`/`regctl`/`skopeo`. The `absent` classification is the only failure path that emits no annotation at all, and it's reachable by misclassification. Both compute steps' fail-loud `exit 1` branches — the "we thought we shipped but didn't" guards — are never executed by any test, and the starter case hard-codes the one `BUILD_STARTERS_RESULT` value that gates both of them off.

---

### Recommendation for whoever resumes this

946 lines of script plus 1,182 of test to express "don't move a tag backward" is what generated the defect list above. Each round fixed real problems and, in fixing them, added surface — a classifier, a summary layer, an annotation-escaping layer, an env-tunable timeout, a CLI — and most of the round-2 Criticals live in that added surface rather than in the idea.

A minimal rebuild is likely to land where this didn't: **verify the digest after the retag, fail loud, drop the rest.** No classifier, no probe-summary vocabulary, no env-tunable budget, no CLI surface beyond what the workflow actually passes. The regression check itself is worth keeping — it's the part that's correct and well pinned — but it can be a much smaller thing.

Two hard constraints on any resumption:

1. **Test `defaultIo`.** That gap is what let a mutation reintroducing the original P0 (retag writes `:stale`) pass 109/109. A rebuild with the same seam and the same untested I/O layer would be no safer than this one.
2. **Bound the fleet, not just the call.** Any design that walks N images serially with a per-call timeout and no aggregate deadline inherits blocker #1, including the `cancelled`-is-unalertable part.

### Two findings that are independent of this PR, handled separately

- **The starter lane still launders `cancelled` → `skipped`.** The showcase lane's per-slot writer keeps `cancelled` distinct, with an emphatic comment explaining that collapsing it is exactly what made a partially-cancelled fleet build silent. The starter lane's writer still does the laundering, so a starter slot killed by its 20-minute budget records `skipped`, `build-starters` rolls up to `cancelled`, both starter fail-loud checks are gated off (they require `BUILD_STARTERS_RESULT = success`), the slug drops out of the CSV silently, and there is no starter analogue of `notify-cancelled-builds`. #6171 only fixed the showcase lane. Getting its own PR.
- **`showcase_build.yml:247`** (the "Build service matrix" step) has no `set -euo pipefail`, unlike its structurally identical starter sibling and unlike the steps this PR added. If the jq pipeline fails, `MATRIX` becomes the empty string, slips past the `[ "$MATRIX" = "[]" ]` typo guard, and the step writes `matrix=` **and** `has_changes=true` and exits 0 — so `build` dies on `fromJSON('')` several jobs downstream of the real cause. Separate fix.

showcase/scripts/advance-latest-tag.ts
```diff
@@ -0,0 +1,946 @@
...
+        action: "already-current",
+        reason: `\`:latest\` already points at ${sha} and carries the same manifest digest as \`:${sha}\` — nothing to do.`,
+      };
+    }
+    return {
+      action: "advance",
+      reason:
+        latestIsSameImage === false
+          ? `\`:latest\` is labelled ${sha} but holds a DIFFERENT manifest digest than \`:${sha}\` — the same commit was rebuilt and \`:latest\` still points at the older image. Advancing so staging and digest-pinned prod agree.`
+          : `\`:latest\` is labelled ${sha} but its digest could not be compared with \`:${sha}\` — advancing rather than assuming the two match.`,
```
AlemTuzlak, line 285: this can't ever be true as there's an early exit on line 274

AlemTuzlak (approved):
this seems right to me and the tests look right

PR #6179, opened 2026-07-27
fix(react-ui): preserve cursor during textarea resize

### LINKED ISSUE #6167: 🐛 Bug: CopilotChatInput - cursor jumps to the end of the textbox upon edit in expanded mode
### ♻️ Reproduction Steps

### ✅ Expected Behavior

### ❌ Actual Behavior

### Screenshots

Image
Image
Image

### 𝌚 CopilotKit Version

```shell
Copilotkit v1.63.1
Same issues on v1.5
```

### 📄 Logs (Optional)

```shell

```

brbousnguar:
## What Problem This Solves

Editing text in the middle of CopilotChat's expanded input can move the cursor to the end after each character.

## Why This Change Was Made

The auto-resizing textarea recalculates its height whenever its value changes. This change preserves the focused textarea's selection across that resize and restores it after the final height is applied.

## User Impact

Users can edit text mid-sentence without the cursor jumping to the end.

## Evidence

- `@copilotkit/react-ui` tests: 57 passed
- React UI TypeScript checks passed
- Pre-commit package checks, build, publint, and attw passed

Fixes CopilotKit/CopilotKit#6167

### DIFF
--- a/packages/react-core/src/v2/components/chat/CopilotChatInput.tsx
+++ b/packages/react-core/src/v2/components/chat/CopilotChatInput.tsx
@@ -707,12 +707,22 @@ export function CopilotChatInput({
       textarea.style.maxHeight = `${maxHeight}px`;
     }
 
+    // Save cursor position before height reset — Chrome reflows on `height = "auto"`
+    // which resets selectionStart/End to the end of the text (issue #6167).
+    const { selectionStart, selectionEnd, selectionDirection } = textarea;
     textarea.style.height = "auto";
     const scrollHeight = textarea.scrollHeight;
     if (maxHeight) {
       textarea.style.height = `${Math.min(scrollHeight, maxHeight)}px`;
     } else {
       textarea.style.height = `${scrollHeight}px`;
+    }
+    if (document.activeElement === textarea) {
+      textarea.setSelectionRange(
+        selectionStart,
+        selectionEnd,
+        selectionDirection as "forward" | "backward" | "none",
+      );
     }
 
     return scrollHeight;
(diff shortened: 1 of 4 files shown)


LiteracyFanatic:
I'm also affected by this bug. The `src/v2/components/chat/CopilotChatInput.tsx` needs the same fix.

brbousnguar:
Thanks for flagging this! I've extended the same cursor/selection preservation fix to `src/v2/components/chat/CopilotChatInput.tsx` in commit f252c7972 and added focused regression coverage.

BenTaylorDev:
Thanks for the patch and for including tests. I'm closing it, and I want to be specific about why so the reasoning is checkable.

The stated mechanism doesn't hold. Both halves save and restore the selection around `textarea.style.height = "auto"`, on the premise that the reflow resets `selectionStart`/`selectionEnd`. Setting a layout property doesn't move the caret in any browser — selection is only reset when the API `value` is written (the spec sets the text entry cursor to the end of the field when the value changes). In `react-ui/src/components/chat/Textarea.tsx` there is no `value` write at all, only `style.height`, so there's nothing there to correct.

The v2 half sits in `adjustTextareaHeight`, which runs on every keystroke, so the added `setSelectionRange` isn't inert — it re-applies a selection captured before the height write and would override any programmatic selection change made in between.

The actual cause of #6167 was the unconditional `ensureMeasurements()` in the narrow-viewport branch of `evaluateLayout` — that one does write `.value` — and #5432 removed it, shipping in v1.64.0.

One thing worth carrying into future work: `CopilotChatInput.cursor.test.ts` and `Textarea.cursor.test.ts` will pass with the change applied, but it's worth checking they *fail* with it removed. In jsdom a selection reset comes from the value setter, not from a height write, so a test built around the height write can pass either way.

PR #6190, opened 2026-07-27
feat: share Ops Clerk session in Docs header

marthakelly:
## Summary
- Add a shell-docs public Clerk provider backed by runtime config.
- Replace Docs desktop/mobile auth-entry controls with Clerk-aware controls that keep the public CTA until a signed-in Ops session is resolved.
- Preserve exact Docs return URLs in dashboard auth-entry links.

## PRD Coverage
- Docs remains public and ungated when Clerk is missing, loading, or unavailable.
- Signed-in shared Ops sessions render Clerk's standard `UserButton` in the existing auth-control slots.
- `NEXT_PUBLIC_CLERK_PUBLISHABLE_KEY` is included in server/client runtime config.
- No Clerk satellite or `satelliteAutoSync` behavior added.

## Validation
- `npm run test -- src/lib/runtime-config.test.ts src/lib/runtime-config.client.test.ts src/components/__tests__/brand-nav.test.tsx` (16 tests)
- `npm run test` (46 files, 239 tests; required hydrating existing Git LFS image assets locally)
- `npm run typecheck`
- `npm run lint` (passes with pre-existing warnings)
- `npm run build` (passes with pre-existing content/frontmatter warnings)


onsclom (changes requested):
Reviewed at `551d7b0e`. Threading the key through `runtimeConfig` is the right approach. One blocking issue, one gap.

**1. Ops discards `redirect_url`, so the return-to-origin criterion cannot pass.** Two independent reasons:

* `apps/ops-frontend/entrypoint/src/main.tsx:423` (`DefaultRedirect`) forwards `location.search` only when a `tier` param is present, otherwise `<Navigate to="/ready" replace />`.
* `apps/ops-frontend/org-flow/src/org-flow-routes.tsx:393` renders `<SignIn forceRedirectUrl="/org" ...>`, and Clerk documents `forceRedirectUrl` as taking precedence over search params.

UTMs are correctly preserved here, so this is only about the return param. ENT-1157 needs an Ops-side change or a descope. Related: `useDocsAuthEntryHref` fills in `redirect_url` from a `useEffect` with an empty dependency array, and `BrandNav` sits in the persistent root layout, so the value goes stale after client-side navigation and is absent pre-hydration. Worth fixing in the same pass once the Ops side is settled.

**2. The Clerk-unavailable fallback is never exercised.** `brand-nav.test.tsx` declares `clerkState.shouldThrow` and the mock throws on it, but no test sets it to `true`. Error boundaries also do not engage under `renderToStaticMarkup`, so `DocsAuthFallbackBoundary` cannot be covered that way. The PRD names this case, and it is the safety path for public docs when Clerk is down.

marthakelly:
Addressed the review feedback in `7eef0fd777`:

- `useDocsAuthEntryHref` now recomputes from the current client location after route/search changes instead of staying fixed from the persistent root layout mount.
- Added a direct regression test for the Clerk-unavailable fallback boundary rather than relying on `renderToStaticMarkup` to exercise an error boundary.
- Added coverage to guard the route/search dependency behavior.

The Ops-side redirect handling called out in review is now covered by https://github.com/CopilotKit/Intelligence/pull/623.

onsclom (dismissed):
Approving at `7eef0fd7`. Both items are addressed. The href recomputes on route and search changes, and the fallback boundary has real coverage now.

Two non-blocking follow-ups:

1. `public-clerk-provider.tsx:24` still has the pattern you just fixed. `afterSignOutUrl` is computed in a `useEffect` with empty deps in a provider above the router, so signing out after a client-side navigation returns the user to the first page they loaded. Same fix, one file over, and the Website provider has it too.
2. The `[pathname, searchParams]` test asserts source text rather than behavior, so a regression that keeps those strings would stay green. Fine to leave given the file's existing style, just noting the staleness case is not behaviorally covered.

marthakelly:
Addressed the non-blocking provider feedback in commit 532763500. The shell-docs PublicClerkProvider now refreshes Clerk's afterSignOutUrl when Next route state changes, and the new test renders the provider, changes mocked pathname/search state, rerenders, and checks the Clerk prop behavior instead of checking source text.

onsclom:
Pushed a small lockfile cleanup. The merge conflict had updated unrelated packages, so I reset the lockfile to current main and regenerated it with npm. It now adds only Clerk and its dependencies. The other changes look great! Reapproved.

onsclom (dismissed):

onsclom (approved):

PR #6193, opened 2026-07-28
docs: organize Channels guides by provider and framework

tylerslaton:
## Summary

- Make Slack and Microsoft Teams the production-ready Channels choices in the top provider picker, with framework-aware routes under `/slack/...` and `/teams/...`.
- Put ten task-oriented guides inside every provider/framework journey and remove the standalone Channels overview from navigation.
- Restore the global Channels SDK reference at `/reference/channels`, including 34 current core, UI, state, transcript, and direct-adapter entries.
- Add the complete CopilotKit Intelligence setup walkthrough, product screenshot, and a stable architecture-diagram slot that Mike's final artwork can replace in place.
- Qualify Discord and WhatsApp correctly: their direct adapters already ship, while managed Intelligence support is coming soon.

## Why

Developers should choose their chat provider and agent framework first, then stay in that context while they build and operate the integration. The previous structure mixed provider guides, an extra overview layer, and stale provider-specific reference pages, making it hard to find the supported path or understand which behavior was managed versus developer-operated.

This update keeps the guide journey provider-specific while returning API material to the normal global Reference surface. It also documents operational boundaries that matter in production instead of adding pages for their own sake.

## How

- Reuse provider-aware MDX across Slack, Teams, and all 19 public agent-framework integrations; the built-in agent keeps the shorter route without a framework segment.
- Organize the sidebar into Getting started, Build, Production, and API reference with guides for Intelligence, tools, rich and interactive messages, commands and reactions, files, state, persistence, transcripts, and operations.
- Pin the verified `@copilotkit/channels@0.4.0` and `@copilotkit/runtime@1.64.1` pair and align the copy with current SDK source plus live Intelligence behavior.
- Document managed capabilities and provider-specific realities, including active/standby runtimes, optional hosted endpoint defaults, output-free turn finalization, Slack manifest scopes, Teams attachment shapes and consent, delivery acceptance semantics, and durable state requirements.
- Preserve useful direct-adapter discoverability for Slack, Teams, Discord, Telegram, and WhatsApp without restoring obsolete symbol pages.
- Add one-hop redirects for retired routes and cover navigation, framework selection, raw-doc URLs, search, reference discovery, and sitemap output.

Validation:

- Full docs suite: 51 files / 348 tests
- Typecheck
- Lint with no errors (existing baseline warnings only)
- Production build: 222 static pages
- Live HTTP checks: Slack Intelligence, Teams + Mastra files, Slack rich messages, and direct-adapter reference all return 200
- Independent read-only correctness passes against the current Channels SDK, Runtime, Intelligence, and every public framework setup

Linear: https://linear.app/copilotkit/issue/OSS-615/channels-sdk-documentation-audit


samjulien (changes requested):
I reviewed the latest PR head locally and walked through the Channels journey. Since this is the Channels 0.3 launch documentation, I think the following should be addressed before publication.

### Blocking issues

1. **The documented dependency pair does not type-check.**

   Both quickstarts install:

   ```bash
   @copilotkit/channels@0.3.0
   @copilotkit/runtime@1.63.3-canary.rc-1
   ```

   I reproduced the documented setup in a clean project. `CopilotRuntime({ channels: [channel] })` fails TypeScript compilation because Channels uses `@copilotkit/channels-core@0.3.0`, while the pinned Runtime embeds `@copilotkit/channels-core@0.2.2-canary.rc-1`. The resulting `Channel` and `Thread` types are incompatible.

   We need to pin or publish a Runtime whose Channels Core and Intelligence dependencies align with `0.3.0`.

2. **The top-level Channels API reference no longer works as a reference surface.**

   Selecting Channels from `/reference` now redirects:

   - `/channels/reference/channel` → `/slack/reference/channel`
   - `/reference/channels` → `/slack/reference/channel`

   This PR deletes the 35 generated Channels reference pages but leaves Channels registered in the reference SDK selector and page-tree implementation. Most old symbol-specific URLs—including `createBot`, `defineBotTool`, `ActionStore`, `renderBlockKit`, and the Discord reference—now collapse into the generic Slack `Channel` guide.

   Removing obsolete pre-0.3 direct-adapter documentation makes sense, but the remaining UI still presents Channels as a normal SDK reference. We should either:

   - provide a real 0.3 API-reference surface, or
   - remove/reframe Channels in `/reference` and link explicitly to the new guides without silently redirecting into Slack.

### Navigation and copy

- **Channels sidebar:** Keep the `Channels` section heading, but render the page as a direct **Overview** navigation item. It should not be a one-page folder titled “Channels” with a disclosure affordance that has nothing to expand.

- **Production-ready labels:** Remove “Production ready” from the Slack and Teams card titles and provider-selector badges. These are the available providers, so the status label is redundant. “Build for Slack” and “Build for Microsoft Teams” are sufficient.

- **Coming-soon copy:** Use only the corrected, product-approved provider list—currently Discord and WhatsApp in the launch handoff—and remove the implementation-oriented explanation about roadmap items and selectable documentation routes. Suggested copy:

  > Discord and WhatsApp support is coming soon.

- **Configuration heading:** Rename “The three things to keep aligned” to something task-oriented. My preference is:

  > Match your Channels configuration

The local route/redirect tests pass because they currently assert the redirects, production-ready labels, and four-provider coming-soon list, so those tests will need to be updated alongside these changes.

samjulien (changes requested):
Thanks for addressing the earlier feedback. The navigation and overall organization are much clearer. I still see a few content and UX issues to address before merging:

1. On the Channels overview, please remove the paragraph beginning:

   > The retired symbol pages documented the pre-0.3 direct-adapter API.

   This is self-referential documentation history and reads like an AI-generated artifact. Users do not need to know about retired documentation on the overview page.

2. Add a screenshot near the top of the Channels overview showing what the feature actually delivers—for example, an agent responding inside Slack or Microsoft Teams. The current text explains the architecture but does not give users a concrete picture of the experience.

3. Add a screenshot near the top of “Configure the Channel in Intelligence” showing the Channels area and its configuration UI. Without that visual context, it is difficult to understand what users are about to configure or what the completed setup should look like.

4. The Intelligence guide presents a runtime snippet and a “Copy snippet” action, then immediately says:

   > Do not copy or run the wizard’s runtime snippet.

   That signals a product/docs mismatch. We should either make the generated snippet usable or direct users to the canonical supported quickstart instead of presenting a snippet they must ignore.

5. The provider constraint callout is too implementation-oriented. Terms such as “current managed SDK activation,” “runtime ownership,” and “gateway session” obscure the actual instruction. Please state it directly: create a separate Intelligence Channel for Slack and Teams, each with a unique Code.

6. Please remove wording such as “the pinned Runtime,” “pinned provider runner,” and “pinned Slack quickstart’s provider runner.” This exposes release mechanics and will age poorly. State the actual configuration requirement directly—for example, that both Intelligence base URLs must be configured.

The remaining structural changes look good, particularly the direct Overview navigation item, the updated Channels treatment on the Reference landing page, and the simplified coming-soon copy.

samjulien (approved):

PR #6203, opened 2026-07-28
feat(telemetry): inspector opened event and banner surface split (OSS-566/568)

BenTaylorDev:
Two related Inspector-telemetry tickets: **OSS-566** and **OSS-568**.

## OSS-566 — explicit "Inspector opened" event

There was no event recording that the panel was opened. Opens could only be inferred from in-panel activity (~1,655/90d, a floor) or from `banner_clicked` cta=`body` (~511), which misses the common floating-button path entirely.

Adds `oss.inspector.opened` with:

| property | values |
|---|---|
| `open_source` | `floating_button` \| `announcement_preview` |
| `has_unseen_announcement` | whether an announcement was on screen at open time |
| `license_status` / `runtime_mode` / `runtime_url_type` | same segmentation the threads events already carry |
| `package_name` / `package_version` / `inspector_distinct_id` | version segmentation |

**Restoring a persisted-open panel deliberately does not count.** Restore assigns `isOpen` directly instead of routing through `openInspector()`, so page reloads — and every `next dev` hot reload — stay out of the number.

## OSS-568 — banner surface + first-class dismissal

1. **`surface` on `banner_viewed`** — `collapsed_preview` (bubble on the collapsed widget) vs `expanded_card` (card inside the opened panel), stamped at fire time. Dedup is now per `(banner, surface)` instead of per banner, so opening the panel records the card impression as its own signal.
2. **`oss.inspector.banner_dismissed`** — emitted **in addition to** `banner_clicked { cta: "dismiss" }`, not replacing it, so dashboards reading the `cta` value keep working. Carries `surface` too, separating "swatted the bubble away" from "dismissed the card after opening".

Both new events clear the sink's `oss.inspector.` prefix gate, so **no telemetry-sink deploy is needed**.

## Testing

- **`packages/web-inspector` full suite — 112 passed (4 files)**, run locally in the worktree:
  ```
   ✓ dev/css-raw-import.spec.ts (1 test) 1ms
   ✓ src/__tests__/telemetry-egress-guard.spec.ts (3 tests) 2ms
   ✓ src/lib/__tests__/telemetry.test.ts (28 tests) 8ms
   ✓ src/__tests__/web-inspector.spec.ts (80 tests) 890ms
   Test Files  4 passed (4)
        Tests  112 passed (112)
  ```
- **New coverage**: payload shape for `opened` / `banner_dismissed`, incl. an allow-list assertion that no content/PII key can be added accidentally; collapsed→expanded surface sequence on open; per-surface dedup; open attribution for both sources; no event for an already-open panel; no event for a restored-open panel; nothing emitted when the runtime reports `telemetryDisabled`; an open still recorded while the runtime is disconnected.
- **`tsc --noEmit`** on `@copilotkit/web-inspector`: clean (after building `core` + `shared` dist in the worktree).
- **`tsdown` build**: succeeds; the test-only egress-guard helper is **not** present in `dist/`.
- **`oxfmt --check`**: clean. **`oxlint`**: 9 warnings, all pre-existing.
- `@copilotkit/runtime` (1,760) and `@copilotkit/shared` (199) also green — both are back on main's own test files in this PR.

## A test-only egress guard rides along

`vitest.setup.ts` installs a fetch guard that swallows requests to the telemetry sink. This is **not** CI plumbing — it is a prerequisite for the new events. These tests run in jsdom, where a real `fetch` exists, and inspector telemetry is fire-and-forget, so any test that drives a banner / threads / open path without stubbing fetch POSTs a real `oss.inspector.*` event to the live sink, from developer machines as well as CI. The announcement-dismissal tests were already doing this; the new `opened` / `banner_dismissed` tests hit the same send path. No environment variable can prevent it, because the inspector's opt-out arrives in the runtime's `/info` response and these tests never boot a runtime.

## Not in scope

Suppressing telemetry from CI jobs that boot real apps (**OSS-565**) was explored on this branch and removed. It needs a mechanism that does not depend on the `/info` handshake — the env → `/info` → core chain is asynchronous, so an early interaction beats it. That ticket stays open and unaddressed here.

Closes OSS-566, OSS-568.


MikeRyanDev (changes requested):
Requesting changes for three telemetry correctness/privacy blockers:

1. **Pre-handshake dismissal bypasses the runtime opt-out.** `AgentRegistry` initializes `telemetryDisabled` to `false` until `/info` completes, but `trackBannerDismissedOnce()` (and the companion `banner_clicked` call) posts immediately. A user or automated run that dismisses the announcement before the runtime connects will therefore send telemetry even if `/info` subsequently reports `telemetryDisabled: true`. Please queue these interactions behind the same connected-handshake gate used for `banner_viewed`/`opened`, and cover disconnected → connected-disabled in a test.

2. **Queued opens preserve pre-handshake placeholder segmentation.** `trackOpened()` snapshots `licenseStatus` and `runtimeMode` when the runtime may still be disconnected; at that point license status is undefined and mode defaults to `sse`. `flushPendingOpened()` later sends those stale values unchanged, so an early open against an Intelligence runtime is permanently attributed to SSE and loses license segmentation. Queue the intent-only fields, then populate runtime/license/url fields when the handshake has completed (or refresh them during flush), with a disconnected → Intelligence test.

3. **The new test env overrides can re-enable telemetry for later test files.** `get-runtime-info.test.ts` deletes the job-wide opt-out in `afterEach`, and the three `vi.hoisted` blocks delete it without restoring it. Vitest workers can execute multiple isolated files in one process; module/mock isolation does not reset direct `process.env` mutations between files. That leaves `COPILOTKIT_TELEMETRY_DISABLED` / `DO_NOT_TRACK` unset for whichever files run afterward, defeating OSS-565 and potentially allowing real egress. Capture the original values and restore them after the suite (the existing `telemetry-disclosure.test.ts` pattern is a good reference); the hoisted cases can return the originals from `vi.hoisted` and restore them in `afterAll`.

All current checks are green, and I also ran the full Nx test targets locally for `@copilotkit/web-inspector` (112), `@copilotkit/runtime` (1,760), and `@copilotkit/shared` (199). These issues are gaps in the covered transition paths rather than existing test failures.

BenTaylorDev:
All three fixed in e7e8922.

**1. Pre-handshake banner interactions.** Confirmed — `_telemetryDisabled` initializes `false` (`agent-registry.ts:85`), and both `trackBannerDismissedOnce` and `trackBannerClickedOnce` posted straight through. Note this was pre-existing for `banner_clicked`, so the gate closes an older hole too. Both now queue in `pendingBannerInteractions` behind the same connected gate as `banner_viewed`, with the same bound. Dedup stays at interaction time, so a double-click pre-handshake queues one event, not two. Covered by disconnected → connected-disabled (drops), disconnected → connected-enabled (sends, with `banner_clicked { cta: "dismiss" }` still riding alongside), and a double-click dedup test.

**2. Queued open segmentation.** Confirmed — `_runtimeMode` defaults to SSE and `_licenseStatus` is undefined pre-handshake. `pendingOpened` now holds intent only (`open_source`, `has_unseen_announcement`); `license_status` / `runtime_mode` / `runtime_url_type` are resolved in `flushPendingOpened` once `/info` has landed. Covered by a disconnected → Intelligence test asserting the flushed event carries `runtime_mode: intelligence` and `license_status: valid`.

**3. Test env restore — fixed, but the leak doesn't reproduce.** Restore is in: a shared `telemetry-opt-out-env` helper, called from `afterAll` in all four files; the hoisted cases snapshot inline and return it from `vi.hoisted`, since hoisted callbacks run before the imports they'd otherwise need.

On the mechanism, though: I measured it rather than assuming. Under the repo's default `isolate: true`, vitest forks a fresh process per test file, so a `process.env` delete in one file doesn't reach the next — two probe files reported pids 8223 and 8225, and the second still saw the var set. A `--no-isolate` control shared one pid (8419) and did leak, so the probe was sensitive. So `COPILOTKIT_TELEMETRY_DISABLED` isn't currently left unset for later files and there's no live egress path. I'd treat the restore as defense against ever flipping `isolate` off (which has been floated as a perf lever) rather than as a fix for something reachable today — happy to be shown a config where it does cross.

Verified: web-inspector 116, runtime 1,760, shared 199, all green with the opt-out vars set; `tsc --noEmit` clean on both packages; oxlint unchanged. 3 of the 4 new tests fail against the pre-fix code.

MikeRyanDev (changes requested):
Thanks — e7e8922 correctly addresses all three findings from my first review: banner interactions now wait for the handshake, queued opens are enriched at flush time, and the runtime tests restore the opt-out environment.

There is still one blocking gap in the broader OSS-565 opt-out guarantee: existing Inspector telemetry paths can post before the `/info` handshake establishes whether telemetry is disabled. `AgentRegistry._telemetryDisabled` starts as `false`, and `handleMenuSelect` directly calls `trackThreadsTabClicked` / `trackMemoriesTabClicked` while disconnected (`packages/web-inspector/src/index.ts:9970` and `:9981`). Other existing Threads/view-state/CTA telemetry paths use the same boolean-only guard.

I reproduced this on e7e8922 with a disconnected Inspector: selecting Threads immediately emitted `oss.inspector.threads_tab_clicked` with `telemetry_disabled: false`; completing the handshake afterward with `telemetryDisabled: true` was too late. The clean Nx suites still pass (web-inspector 116, runtime 1760, shared 199), so current coverage does not catch this race.

Please gate all Inspector telemetry on a known-connected/allowed state, ideally at a centralized egress boundary rather than adding per-event queues, and add disconnected -> connected-disabled coverage for at least the existing Threads tab/view-state path. Otherwise CI/runtime opt-out is not yet a complete safety property.

MikeRyanDev (changes requested):
The workflow-path failure from f717ad1 is fixed by 56c6f56: inlining the hosts entry before checkout removes the path and skipped-job problems. The clean Nx package suites also pass on this code (web-inspector 116, runtime 1,760, shared 199).

One privacy blocker from my prior review remains unresolved. f717ad1 removes the centralized handshake gate and explicitly changes the test contract to send while disconnected (`web-inspector.spec.ts:1718-1727`). `navigator.webdriver` is a good synchronous CI suppression signal, but it only protects automated browsers. In a normal browser, `COPILOTKIT_TELEMETRY_DISABLED` / `DO_NOT_TRACK` is still unknown until `/info` returns; any Inspector event emitted before then leaves with the default allowed state. This is the same disconnected -> later-disabled leak reproduced in the previous review, now reintroduced intentionally for every event path.

Please preserve the automation suppression and CI egress backstop, but restore a tri-state/known-allowed gate for runtime-derived opt-out semantics at the centralized Inspector egress boundary. Add coverage proving a normal browser emits nothing during disconnected -> connected-disabled. A runtime that never connects losing telemetry is the safe failure mode for an opt-out guarantee; failing open means an explicit deployment opt-out cannot promise zero browser egress.

BenTaylorDev:
## Scope cut: OSS-565 removed, this PR is now OSS-566 + OSS-568 only

Pushed in 783fc89. Every mechanism your two reviews were about is gone from the branch, so rather than iterate on it further I've excised the CI-suppression effort and reduced this to the two event additions.

**Removed:** the workflow-level `COPILOTKIT_TELEMETRY_DISABLED` / `DO_NOT_TRACK` env across 10 workflows, the runner egress-block steps, `examples/e2e/playwright.config.ts`, the runtime/shared test-env manipulation and the `telemetry-opt-out-env` helper, and the `navigator.webdriver` check. Also gone: the per-event handshake queues from e7e8922 and the centralized egress gate from 9599108 — no inspector telemetry waits on `/info` any more.

Every reverted path is byte-identical to the **merge-base**, which I verified rather than assumed: `git checkout origin/main -- <path>` would otherwise have quietly imported main's newer content for any of those files that had moved on in the 29 commits since. None had.

**Final diff is 7 files, all under `packages/web-inspector`.**

### Please read this part before re-reviewing

The scope cut does **not** resolve your findings — it removes the machinery they were about, and leaves the underlying race exactly as it is on `main` today. Concretely: `banner_dismissed` fires immediately, guarded only by the `core.telemetryDisabled` boolean, which reads `false` until `/info` answers. That is the same behavior `banner_clicked` has had on main all along, and the new event inherits it. I'm flagging it rather than implying it's fixed.

That was a deliberate call, on this reasoning: gating on the handshake means an Inspector whose runtime never connects — or whose `/info` fails, which sets status to `Error` rather than `Connected` — goes silent for the entire session. That penalizes users with broken setups, who are disproportionately likely to have the Inspector open. Closing the race properly needs a signal that doesn't depend on the handshake, which is a bigger piece of work than these two events warrant.

### A pre-existing defect I found while in there, worth its own ticket

`renderThreadsLockedView()` calls `trackThreadsViewStateOnce()` **from inside a render function**. So a panel restored open on the Threads tab emits `threads_locked_viewed` on first paint, reliably beating `/info` — which is why the race isn't hypothetical. Independent of telemetry, emitting side effects from `render()` is wrong: Lit can render more than once and only the dedup `Set` is hiding it. Untouched here, on `main` today.

### One piece kept, and it isn't CI plumbing

`vitest.setup.ts` still installs a test-only fetch guard that swallows requests to the telemetry sink. These tests run in jsdom where a real `fetch` exists, and inspector telemetry is fire-and-forget — so any test driving a banner / threads / open path without stubbing fetch POSTs a real `oss.inspector.*` event to the live sink, from developer machines as well as CI. The announcement-dismissal tests were already doing exactly that, and the new `opened` / `banner_dismissed` tests hit the same send path, so shipping the events without it would make our own dataset dirtier. No env var can cover it — the inspector's opt-out arrives in `/info` and these tests never boot a runtime. Its comments no longer reference OSS-565; the rationale stands on its own. Happy to pull it if you'd rather it went too.

### Verification

`@copilotkit/web-inspector` 112 passed; `@copilotkit/runtime` 1,760 and `@copilotkit/shared` 199 — the latter two now running main's own test files again. `tsc --noEmit` clean, `tsdown` build clean with no test-only helper in `dist/`, `oxfmt` clean, `oxlint` 9 warnings all pre-existing.

The one red check, `llama-index`, is an infrastructure flake unrelated to this branch: `@protobuf-ts/protoc failed to download protoc v35.1 — Error: socket hang up`, in the sibling **ag-ui** checkout's `@ag-ui/proto:generate` step. Dojo is green on the last five runs of `main`.

### OSS-565

Moved to **Scoped** with a writeup of what was tried, what we learned, and why we bailed — including the measurement that `navigator.webdriver` is `true` under Playwright in both headless and headed mode, the `/info`-failure behavior above, and the CI gotcha that a repo-relative script step exits 127 in the dojo job (it checks out into a subdirectory and gates its steps). The jsdom guard is noted there as already done so nobody redoes it.

MikeRyanDev (changes requested):
Thanks for the scope cut. I re-reviewed the final seven-file OSS-566/568 diff rather than carrying the removed OSS-565 implementation forward. The affected test, check-types, and build targets pass, and the GitHub checks are green. Two scoped blockers remain:

1. The surface refactor removes the existing pre-handshake queue for `banner_viewed`. On the base branch, `fetchAnnouncement()` stored `pendingBannerViewed`, and the flush waited for `connected` and dropped the event when `telemetryDisabled` was true. `maybeTrackBannerViewed()` now posts directly at `packages/web-inspector/src/index.ts:10850`. This is a new opt-out regression caused by this PR, not the acknowledged preexisting `banner_clicked` race. Please preserve the existing handshake behavior while queueing and deduping impressions per surface, with disconnected-to-disabled coverage.

2. `oss.inspector.opened` snapshots runtime segmentation while disconnected at `packages/web-inspector/src/index.ts:7871-7873`. `AgentRegistry` initializes `runtimeMode` to SSE before `/info` (`packages/core/src/core/agent-registry.ts:77`), so an early open against an Intelligence runtime is permanently recorded as SSE and without license status. That defeats a stated OSS-566 dimension. Please either queue the open intent and enrich it after `/info`, or omit runtime-derived fields unless connected, and add disconnected-to-Intelligence coverage.

The test-only egress guard and the remaining event/surface behavior look good.

BenTaylorDev:
Both correct. Fixed in 108f42d.

**1. `banner_viewed` queue restored.** You're right that this one is on this PR rather than inherited — the deferral existed on `main`, and removing it made the impression post on the `false` placeholder. Restored, and widened from a single slot to a bounded array: opening the panel reveals the second surface, which can happen before the first impression has flushed, so a single slot would drop one of them. Per-surface dedup is unchanged, and the flush still discards rather than sends when `/info` comes back with `telemetryDisabled: true`.

**2. `opened` — I took your second option rather than queueing.** `license_status` and `runtime_mode` are now omitted unless the handshake has landed, instead of being recorded from `licenseStatus: undefined` and the `sse` default. Reasoning for preferring omit over queue: a disconnected runtime shouldn't suppress a real user's open, and for segmentation an absent dimension is honest whereas a wrong one actively corrupts the split OSS-566 exists to provide. `runtime_url_type` stays unconditionally — it's derived from configuration, not from `/info`, so it's accurate immediately. If you'd rather have the open held and enriched so early opens keep full segmentation, say so and I'll switch it; both satisfy the finding, and the trade is "unsegmented but counted" vs "counted late but complete".

**One thing worth flagging, since it would have undermined the fix.** My first version of the pre-handshake `opened` test passed against the buggy code. `HeaderMockCore` had no `runtimeMode` property at all, so the assertion read `undefined` whether the bug was present or not — it would have shipped green over the exact defect it was written for. The real core initializes `runtimeMode` to `sse` before `/info` (`agent-registry.ts:77`), so the mock now mirrors that pre-handshake state (`runtimeMode: "sse"`, `licenseStatus` absent). Only after that correction did the test start failing pre-fix. Worth knowing if you review other telemetry assertions built on that mock — the same blind spot would apply to them.

**Coverage added** (4 cases): impression held then dropped when the runtime reports disabled; held then released when it allows; an open before the handshake omitting both fields while keeping `runtime_url_type`; and an open against a connected Intelligence runtime carrying `runtime_mode: intelligence` + `license_status: valid`. Three of the four fail against the previous commit — the fourth guards the connected path.

**Verified locally:** web-inspector 115 passed, `tsc --noEmit` clean, `tsdown` build clean, `oxfmt` clean, `oxlint` 9 warnings all pre-existing. CI is still running on 108f42d; 11 green and no failures so far.

Unrelated to this PR, and mentioned only because I tripped over it: `renderThreadsLockedView()` calls `trackThreadsViewStateOnce()` from inside a render function, so a panel restored open on the Threads tab emits on first paint. That's on `main` today and I've left it alone here, but it's a defect independent of telemetry policy — Lit can render more than once and only the dedup `Set` is masking it. Happy to file it separately.

MikeRyanDev (approved):
`108f42d2` resolves both scoped blockers from my prior review. Banner impressions remain queued per surface until the runtime handshake and are dropped when telemetry is disabled. Opens before the handshake no longer claim runtime-derived segmentation; connected Intelligence opens include it.

Exact-head validation passes: 115 web-inspector tests, web-inspector/core/shared type checks, and all current GitHub checks.

Issue #632, opened 2024-09-30
⭐ 20 - Email Organizer (Hacktoberfest Demo)


NathanTarbert:
### 🤩 Thank you for participating in the CopilotKit Hacktoberfest event. 

Follow [This Link to Get Started!](https://github.com/CopilotKit/CopilotKit/tree/main/community/content)🎉

#### NOTE: CopilotKit will allow multi-assign issues

  Please show your support and give a ⭐ to the CopilotKit repository.

### Description of app functionality:
Create an email management tool that helps users categorize and prioritize emails based on urgency and content

#### Using CopilotKit
- **Required**: Your project must utilize CopilotKit as a core component.
- **Installation**: Include CopilotKit by following the installation guide at [CopilotKit Quickstart](https://docs.copilotkit.ai/quickstart).
- **Documentation**: Familiarize yourself with CopilotKit's features through the [Official Documentation](https://docs.copilotkit.ai).

#### Styling with Shadcn-UI
- **Required**: All UI components in your project should be styled using Shadcn-UI.
- **Installation**: Install Shadcn-UI [documentation](https://ui.shadcn.com/docs/installation)

![Swag_2](https://github.com/user-attachments/assets/d120bf97-95af-40a6-8618-6fb6a9198696)

Code-Hacker26:
Hey @NathanTarbert, I’m interested in working on the Email Organizer demo, where users can categorize and prioritize emails based on urgency and content. Can you assign it to me ?

Rishikesh63:
@alexandredev3  and @Code-Hacker26  can we work together on this project to complete it early

Code-Hacker26:
> @alexandredev3 and @Code-Hacker26 can we work together on this project to complete it early

@Rishikesh63 sure I am in.

alexandredev3:
> @alexandredev3  and @Code-Hacker26  can we work together on this project to complete it early

Sure!

Rishikesh63:
@NathanTarbert is it looks like as 
![Screenshot 2024-10-15 012756](https://github.com/user-attachments/assets/02894b5a-72be-4614-9f5f-b0e41b8dbc86)

Rishikesh63:
or it is looks like 
![Screenshot 2024-10-15 013952](https://github.com/user-attachments/assets/47c6c8e1-3641-4e0b-b70e-211436db3f3f)

Rishikesh63:
[...]

PR #6204, opened 2026-07-28
fix(channels): hide Slack tool status by default

tylerslaton:
## Problem

Slack threads render every tool invocation as visible progress, leaving noisy `Used …` rows ahead of the final answer.

## Why

The direct Slack renderer treated tool status as enabled when `showToolStatus` was omitted, while the managed Intelligence renderer always emitted tool lifecycle frames. The managed renderer's opt-in was not reachable through the supported `createChannel()` and `CopilotRuntime` API.

## Fix

- Hide tool-call progress by default for direct and managed Slack routes.
- Expose `createChannel({ showToolStatus: true })` as the managed Channels opt-in and carry it through runtime activation to the Intelligence launcher.
- Preserve `slack({ showToolStatus: true })` as the direct Slack opt-in.
- Keep existing managed behavior for non-Slack routes.
- Continue capturing and executing tool calls while their status frames are hidden.
- Add regression coverage for the public managed API, legacy and native-streaming Slack, and Realtime Gateway paths.

Paired with https://github.com/CopilotKit/Intelligence/pull/632.


BenTaylorDev (commented):
## Overview

Flips tool-call progress to hidden-by-default on both Slack routes and adds a managed opt-in path:

- **Direct Slack** (`packages/channels-slack`): `createRunRenderer`'s `showToolStatus` default flips `true → false`; `slack({ showToolStatus: true })` stays the opt-in.
- **Managed** (`packages/channels-intelligence`): `IntelligenceAdapter` skips minting `tool_start`/`tool_end` render frames, defaulting off *only* for Slack routes (`route.adapter === "slack"`), preserving visible tool status for Teams/other providers.
- **New public opt-in**: `createChannel({ showToolStatus: true })` → `Channel.showToolStatus` → `ChannelActivationConfig` → `defaultActivateChannel` → launcher → adapter. Previously unreachable through the supported API.

Tool calls are still captured and executed in both renderers — the early returns sit after `captureToolCall`/`ensureRunStarted`. Correct ordering.

## Verification

Fresh worktree off `4806321de`, `pnpm install --frozen-lockfile`, then raw vitest/tsc:

| Suite | Result |
|---|---|
| `channels-slack` (23 files) | **282 passed** |
| `channels-intelligence` (10 files) | **207 passed** |
| `runtime` channel-manager + channels-integration | **42 passed** |
| `tsc --noEmit` channels-slack / channels-intelligence, channels-core build | **clean** |

## Findings

**1. Suppressing frames also drops them from the canonical transcript (main concern).** On the managed path, render frames have two consumers in Intelligence's `apps/app-api/src/channels/`: `connector-outbox-render.ts` (Slack display) **and** `render-frame-projection.ts`, which projects `tool_start`/`tool_end` into canonical `run_events` as `TOOL_CALL_START`/`END` — its own doc comment calls this "conversation truth." Not emitting the frames means managed Slack threads lose tool calls from the durable AG-UI transcript (dashboard thread view, replay), not just from the Slack bubble. The PR describes this as display-only, which holds for the direct adapter but not for managed. The clean split would be to keep minting frames and gate display Intelligence-side — but Intelligence hardcodes `showToolStatus: true` in its native turn renderer, so that needs a cross-repo change. Worth stating the trade-off explicitly and confirming transcript loss is acceptable.

**2. The managed opt-in will silently break on the legacy Slack path after Intelligence bumps the pin.** Intelligence's `createLegacyTurnRenderer` constructs the SDK renderer with **no** `showToolStatus`. Today it pins `@copilotkit/channels-slack@0.2.1` (default `true`), so opted-in frames render. Once the pin includes this PR, that fallback path (`chat.update` replay) will swallow tool frames even when the developer set `createChannel({ showToolStatus: true })`. Needs a paired Intelligence change or a tracked follow-up.

**3. Long single-iteration turns can lose the "is thinking…" indicator.** Per the code's own comments (`event-renderer.ts:343`, `:420`), Slack's status has a 2-minute timeout that every `setStatus` resets, and the per-tool-call `setStatus("is using \`tool\`…")` was one of the refreshers. With tool status off, refresh only happens on `RUN_STARTED` (once per tool-loop iteration), so a >2-min iteration goes silent. Cheap fix: when `!showToolStatus`, re-set the generic thinking text on tool events instead of returning early — refreshes the timeout without leaking tool names.

**4. Exported launchers ignore `channel.showToolStatus`.** `startChannels`, `startChannelsWithGatewaySession`, and `startChannelsOverRealtimeGateway` are all public exports and each takes only the global `opts.showToolStatus`; the per-channel flag arrives solely via the runtime's activation-config path. Anyone using a launcher directly gets `createChannel({ showToolStatus: true })` silently dropped. One-liner in `runtime.ts` (~L190): `showToolStatus: opts.showToolStatus ?? channel.showToolStatus` — the adapter is already built inside the per-channel loop. Not a leak risk on the runtime path, which activates one channel per call.

**5. `adapterFromRoute` sniffs a deliberately opaque type.** `EgressRoute` is declared `unknown` in `contracts.ts`, and the new helper reaches into it for `.adapter`. It does hold in reality — the route is Intelligence's `channelReplyTargetSchema`, a `z.discriminatedUnion('adapter')` — but the envelope already carries `platform` (`claim-mapping.ts`) and the launcher already receives `adapter` for `declaredChannels`. Keying off one of those, or threading platform onto `ChannelReplyTarget`, wouldn't depend on an opaque blob's shape. Related edge: a hand-built `dispatchTo` route with no `adapter` field defaults to **visible** on a Slack channel.

### Minor

- `packages/channels-telegram/src/event-renderer.ts:52` still defaults `?? true`. If the noisy-`Used …`-rows rationale is platform-agnostic, Telegram is now inconsistent — intentional Slack-only scope, or a follow-up?
- `examples/slack/app/index.ts:103` now passes `showToolStatus: false` redundantly, with a comment framing it as an opt-out. Worth updating or dropping.
- `createChannel({ showToolStatus })` is silently ignored for direct-adapter channels (documented, no runtime warning). Consistent with how `provider` behaves, so just a nit.
- `render-events.test.ts` shares a module-level `subscriberAgent`/`subscriberContext` across tests; a factory would be safer against cross-test coupling.
- Release notes for `channels-slack` should call out the default flip — it's a user-visible behavior change for existing `slack()` users.

## What's good

- Tests cover the right seams: managed default vs. opt-in through the **public** `createChannel` API, activation-config pass-through including the "omitted stays omitted" case, non-Slack managed default preserved, and both Slack legacy and native-streaming paths.
- `render-events.test.ts` drops the `as unknown as Sub` casts for real typed AG-UI events — a genuine quality improvement, and `@ag-ui/core` is correctly a devDependency pinned to `0.0.57` matching its siblings.
- Conditional-spread style for optional fields matches surrounding code throughout, and the docs (`configuration.mdx`, `channel.mdx`, slack reference, README) were all updated with the new default.

Nothing here blocks on correctness within this repo — the code does what it says and the suite is green. Findings 1 and 2 are the ones I'd want answered before merge, since both depend on the Intelligence side.

MikeRyanDev (approved):
No blocking findings at `cc453683`. The public opt-in reaches both managed launch paths, hidden Slack tool calls still refresh the generic status, and tool lifecycle frames remain available for execution and canonical history. Focused tests and type checks pass.

Ship this with Intelligence #632 so the strict gateway accepts the new `run_started` field.

PR #6206, opened 2026-07-28
fix(channels-core): default runAgent prompt from inbound message

tylerslaton:
## Problem

Calling `thread.runAgent()` without an explicit prompt can omit the inbound Channel message when reconstructed history excludes the in-flight turn. Applying the fallback unconditionally would duplicate that turn in native adapters whose conversation stores already seed it.

## Why

`Thread.runAgent()` had no adapter-declared signal for whether the conversation store already supplied the inbound message, and a text-only fallback would discard multimodal content parts.

## Fix

Add `ConversationStore.seedsInboundTurn`, declare the native-adapter and Intelligence behaviors, and default an omitted prompt only when the store has not already seeded the turn. Prefer non-empty inbound `contentParts` over `message.text`, preserve explicit prompts, and add composition plus adapter regression coverage for both paths.


BenTaylorDev (changes requested):
The gap this fixes is real, but the fix lands at a seam shared by two adapter families with opposite requirements — and I reproduced a duplicated user turn on the other family. Details below.

## The premise is only half true

> Calling `thread.runAgent()` without an explicit prompt omits the inbound Channel message.

That holds for the **Intelligence adapter** only. Its history comes from `source.getHistory()`, whose contract explicitly says (`channels-intelligence/src/transports.ts:62-64`) *"oldest→newest, **excluding the turn currently being dispatched**"*. So bare `runAgent()` on that path genuinely runs with zero messages — real bug.

For every **native adapter**, the inbound message is *already* in the agent's history before the handler runs:

| Adapter | Mechanism |
|---|---|
| Slack | `conversations.replies`/`history` rebuild includes the triggering message (`channels-slack/src/conversation-store.ts:141-161`) |
| Discord | *"The inbound (triggering) message is part of the reconstructed history, so the app passes NO prompt"* (`channels-discord/src/conversation-store.ts:29-30`) |
| Teams | `store.recordUser(...)` **before** `sink.onTurn` (`adapter.ts:234`), seeded into `agent.messages` (`conversation-store.ts:103-105`) |
| WhatsApp | `history.append(...)` **before** `sink.onTurn` (`webhook-listener.ts:102-113`) |
| Telegram | listener enqueues; `getOrCreate` drains via `agent.addMessage` (`conversation-store.ts:85-95`) |

We already document this exact split in `examples/slack/app/managed.ts:89-93`:

> *"Channel history (app-api /api/channels/history) does NOT include the in-flight turn (**unlike native adapters whose getHistory rebuilds the live thread**), so pass the current message explicitly as `prompt` — otherwise runAgent runs with zero messages. Prefer multimodal parts."*

## Blocking

### 1. Duplicates the inbound user message on all five native adapters

I applied the patch locally and ran it against a conversation store that seeds the inbound message the way Telegram/Slack/Teams do. On `main` it passes; on this branch the agent sees the turn twice:

```
USER MESSAGES SEEN BY AGENT: [
  {"id":"inbound","role":"user","content":"Say my name"},
  {"id":"23b444ff-…","role":"user","content":"Say my name"}   ← injected by the new default
]
```

Bare-`runAgent()` call sites already in the repo that this hits:

- `examples/teams/app/index.tsx:232` — and the comment directly above at :213-216 says the message is *"recorded into the conversation transcript by the adapter … so `runAgent()` picks them up from the seeded history with no extra wiring"*
- `packages/channels-telegram/README.md:63` — `bot.onMention(({ thread }) => thread.runAgent())`
- `packages/channels-whatsapp/README.md:55`
- `examples/slack/README.md:113` — `runAgent({ context: … })`, no prompt

Telegram is the worst case: the drained queue entry may be multimodal, so the duplicate is a lossy plain-text copy of it.

**Suggestion** — make this adapter-declared instead of a blanket default. The deciding fact belongs to the `ConversationStore` (or to `SurfaceCapabilities`, which already carries optional `supports*` flags at `platform-adapter.ts:20-40`):

```ts
// ConversationStore
/** True when getOrCreate seeds the in-flight inbound turn into agent.messages. */
seedsInboundTurn?: boolean;
```

Default the prompt only when that is falsy. Native stores set it `true`; the Intelligence adapter leaves it unset and gets the fix. That also turns the invariant into something enforced in one place, rather than implicit in five separate store comments.

### 2. The default silently drops multimodal attachments

`message.text` discards `message.contentParts`. Same local run, inbound text + image:

```
MULTIMODAL CASE: [{"role":"user","content":"look"}]   ← image part gone
```

Not a regression from a working state (`main` injects nothing here at all), but the new default degrades an image/CSV upload to text-only — and the new JSDoc frames that as the caller's job (*"Pass … `AgentContentPart[]` to preserve multimodal input"*) when the default can just be right. Both `examples/slack/app/managed.ts:95-97` and the existing `contentParts` test right below the new one already use the correct form:

```ts
prompt: input?.prompt ?? (message.contentParts?.length ? message.contentParts : message.text)
```

## Non-blocking

- **JSDoc regression** (`thread.ts:292-299`): the rewrite drops the accurate rationale (slash-command args are never posted to the channel; adapter-reconstructed history) and replaces it with a claim that is wrong for native adapters. The scope sentence *is* correct, though — I verified interaction/command/reaction/threadStarted threads are built without `extras`, so `deps.message` is undefined and those paths are unaffected.
- **The test doesn't reach the defect.** It uses `FakeAdapter`, whose `conversationStore` seeds nothing (`fake-adapter.ts:171-175`), so it models only the Intelligence shape. I ran all six channels suites on this branch — **979 tests, all green**. Nothing guards the store↔`thread.run` composition, which is precisely where this breaks. Worth a `channels-core` test with a seeding store, or per-adapter tests asserting exactly one user message reaches the agent.
- **Test hygiene**: the `agent.addMessage` capture monkeypatch is copy-pasted from the test immediately below — worth a shared helper. The new test also uses `getSink().onTurn()` where its neighbor uses `emitTurn()` + `tick()`; awaiting the sink directly is the better pattern, just inconsistent now.

## Edge cases that are fine

- `message.text === ""` (bare file upload) → `""` is falsy, so `run()`'s `if (extra?.prompt)` guard skips injection. No empty user message.
- Explicit `prompt: ""` → `??` preserves `""`; behavior unchanged from `main`.

Happy to hand over a patch implementing the `seedsInboundTurn` gate, the contentParts-preserving default, and the missing composition test if that saves you a round-trip.

PR #6208, opened 2026-07-28
fix(channels-intelligence): fence delivery terminal attempts

tylerslaton:
## Problem

Channel delivery state was keyed only by delivery ID, so terminal requests could outlive or cross lease attempts. A lost completion response also left the SDK without a safe, byte-equivalent retry, while failed completion handling risked turning a successful run into a nack and rerun.

[OSS-631](https://linear.app/copilotkit/issue/OSS-631/investigate-runtime-crash-while-using-channels-sdk)

## Why

At-least-once delivery can lease the same delivery more than once. Late render, ack, or fail work from an older attempt must never borrow the current lease token or race an opposite terminal intent. Once a turn has succeeded, completion failure also cannot become a nack because that would rerun the agent and duplicate side effects.

## Fix

- Thread exact delivery-attempt identity through claim mapping, adapter render state, HTTP delivery, and realtime gateway state.
- Fence stale or expired attempts, cap handler deadlines inside the lease safety margin, and centralize first-terminal-wins coordination.
- Freeze the first terminal payload so a retryable completion can retry once byte-for-byte; preserve OSS-491 behavior by logging any final completion failure without nacking a successful turn.
- Add regression coverage for stale attempts, terminal races, retry payload stability, lease expiry, and HTTP/realtime parity.

Validated with scoped formatting; repository lint; package typecheck, build, tests, publint, and ATTW; and the full uncached package test and build graphs.

### DIFF
--- a/packages/channels-intelligence/src/claim-mapping.test.ts
+++ b/packages/channels-intelligence/src/claim-mapping.test.ts
@@ -14,6 +14,8 @@ const baseDelivery = (
   channel: { id: "channel_1", name: "support" },
   adapter: "slack",
   leaseToken: "lease_1",
+  attempt: 2,
+  leaseExpiresAt: "2099-07-01T00:02:30.000Z",
   turn: {
     id: "turn_1",
     eventId: "evt_1",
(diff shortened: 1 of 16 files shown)


BenTaylorDev:
Closing this — the tree it patches no longer exists.

`5ddcb14233` (`refactor(channels): use gateway live sessions`) landed the day after the last commit here and replaced the delivery transport layer wholesale. Twelve of the sixteen files in this PR were deleted by it, including everything the fencing logic hangs off: `claim-mapping.ts`, `contracts.ts`, `http-transports.ts`, `transports.ts`, `in-memory-transports.ts`, `intelligence-adapter.ts`, `realtime-gateway-transport.ts`. The new packet model carries `ownerGeneration` on every packet and fences server-side, so the attempt-token approach here wouldn't port cleanly.

OSS-631 was canceled on Jul 30 and the companion Intelligence PR (CopilotKit/Intelligence#621) was closed unmerged.

PR #6218, opened 2026-07-29
Add Teams-native Adaptive Cards for Channels

tylerslaton:
## Summary

Add Teams-native Adaptive Cards to Channels, including a typed JSX catalog, direct codec, durable submit envelope, and platform-safe IR delivery.

## Why

Teams needs to preserve native card schema and form submissions without silently falling back to portable UI or persisting an action before platform validation succeeds.

## How

- introduce reserved platform IR validation before ActionStore writes, with managed target-platform propagation
- add `@copilotkit/channels-teams/ui`, direct/umbrella codec exports, and Adaptive Card serialization
- validate raw/native card payloads, input IDs, submit callbacks, schema/version limits, and unsupported actions
- carry submitted form values through the reserved `__copilotkit` envelope while retaining legacy inbound parsing
- cover native serialization, interaction decoding, public exports, preflight behavior, and package type/export checks

Not included in this review slice: Connector Outbox lifecycle work, live Teams tenant smoke coverage, and docs/examples.

### DIFF
--- a/packages/channels-core/src/action-registry.test.ts
+++ b/packages/channels-core/src/action-registry.test.ts
@@ -1,6 +1,7 @@
 import { describe, it, expect, beforeEach } from "vitest";
 import { ActionRegistry, ActionExpiredError } from "./action-registry.js";
 import { InMemoryActionStore } from "./action-store.js";
+import type { ActionSnapshot, ActionStore } from "./action-store.js";
 import { MemoryStore } from "./state/memory-store.js";
 import { kvActionStore } from "./state/kv-action-store.js";
 import type { ChannelNode, InteractionContext } from "@copilotkit/channels-ui";
@@ -165,5 +166,32 @@ describe("ActionRegistry", () => {
         ActionExpiredError,
       );
     });
+  });
+
+  it("runs preflight before writing any action snapshot", async () => {
+    const writes: string[] = [];
+    const store: ActionStore = {
+      async put(id: string, _snap: ActionSnapshot): Promise<void> {
+        writes.push(id);
+      },
+      async get(): Promise<ActionSnapshot | undefined> {
+        return undefined;
+      },
+      async delete(): Promise<void> {},
+    };
+    const reg = new ActionRegistry({ store });
+    const tree: ChannelNode = {
+      type: "button",
+      props: { onClick: () => undefined },
+    };
+
+    await expect(
+      reg.bindRenderable(tree, "conv", {
+        preflight: () => {
+          throw new Error("unsupported UI");
+        },
+      }),
+    ).rejects.toThrow("unsupported UI");
+    expect(writes).toEqual([]);
   });
 });
--- a/packages/channels-core/src/action-registry.ts
+++ b/packages/channels-core/src/action-registry.ts
@@ -19,6 +19,14 @@ export class ActionExpiredError extends Error {
 
 const EVENT_PROPS = ["onClick", "onSelect", "onSubmit"] as const;
 
+export interface BindRenderableOptions {
+  /**
+   * Runs after JSX expansion and before any action record is written. This is
+   * the boundary where an adapter rejects a UI tree it cannot render.
+   */
+  preflight?: (root: ChannelNode[]) => void | Promise<void>;
+}
+
 function isComponentElement(
   ui: unknown,
 ): ui is { type: ComponentFn; props: Record<string, unknown> } {
@@ -111,9 +119,11 @@ export class ActionRegistry {
     componentName: string,
     props: Record<string, unknown>,
     conversationKey: string,
+    options?: BindRenderableOptions,
   ): Promise<ChannelNode[]> {
     const fn = this.components.get(componentName);
     const root = renderToIR((fn ? fn(props) : props) as Renderable);
+    await options?.preflight?.(root);
     await this.walk(root, [], componentName, props, conversationKey);
     return root;
   }
@@ -128,6 +138,7 @@ export class ActionRegistry {
   async bindRenderable(
     ui: Renderable,
     conversationKey: string,
+    options?: BindRenderableOptions,
   ): Promise<{
     root: ChannelNode[];
     onReaction?: MessageReactionHandler;
@@ -146,9 +157,10 @@ export class ActionRegistry {
       component = fn.name || "anonymous";
       props = (ui.props ?? {}) as Record<string, unknown>;
       this.registerComponent(component, fn);
-      root = await this.bindTree(component, props, conversationKey);
+      root = await this.bindTree(component, props, conversationKey, options);
     } else {
       root = renderToIR(ui);
+      await options?.preflight?.(root);
       await this.walk(root, [], "", undefined, conversationKey);
     }
     const onReaction = takeMessageReaction(root);
--- a/packages/channels-core/src/create-channel.test.ts
+++ b/packages/channels-core/src/create-channel.test.ts
@@ -5,7 +5,13 @@ import { defineChannelCommand } from "./commands.js";
 import { FakeAdapter } from "./testing/fake-adapter.js";
 import { FakeAgent } from "./testing/fake-agent.js";
 import { MemoryStore } from "./state/memory-store.js";
-import { Section, Actions, Button } from "@copilotkit/channels-ui";
+import {
+  Section,
+  Actions,
+  Button,
+  platformNode,
+  PlatformUiMismatchError,
+} from "@copilotkit/channels-ui";
 import type { ChannelNode } from "@copilotkit/channels-ui";
 import type { PlatformAdapter } from "./platform-adapter.js";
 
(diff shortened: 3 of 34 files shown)


BenTaylorDev (commented):
## Hard review

**What it does:** adds a `$platform` reserved IR node to `channels-ui` with a shared `validatePlatformUi` boundary; wires that boundary into `Thread` as an `ActionRegistry.bindRenderable` **preflight** so an unrenderable tree is rejected *before* any ActionStore write; adds a typed Teams JSX catalog (`@copilotkit/channels-teams/ui`) that serializes straight to Adaptive Card JSON; moves the button envelope from flat `{ckActionId, value}` to a reserved `data.__copilotkit` namespace so Teams' named input values can ride at the top level; and extracts a credential-free `teamsCodec` for the managed Connector Outbox path.

The architecture is right. The preflight seam is the correct place for this, the reserved-namespace envelope is the right fix for the input-value collision, and the core mechanism holds end-to-end (verification below). Findings are mostly at the edges.

### What I verified (not just read)

Built a worktree at `edc9319406`, aliased `@copilotkit/*` to source, and ran the real suites plus targeted probes:

| Suite | Result |
|---|---|
| channels-ui | 29/29 pass |
| channels-teams | 93/93 pass |
| channels-core | 159/159 pass |
| channels-slack / discord / telegram | 279 / 191 / 150 pass |
| channels-intelligence | 198/198 pass |

Probe — native card through `ActionRegistry` → preflight → codec, **including a cold-cache dispatch** (hot cache cleared, handler re-derived from the store):

```
actionId: ck:3a4a82988ff58fea
EGRESS: {"card":{...,"body":[{"text":"Deploy?","type":"TextBlock"},
  {"id":"environment","choices":[...],"type":"Input.ChoiceSet"}],
  "actions":[{"title":"Deploy","type":"Action.Submit",
    "data":{"__copilotkit":{"version":1,"actionId":"ck:3a4a82988ff58fea","value":{"intent":"deploy"}}}}]}}
handler invoked times: 1
handler ctx values: {"environment":"prod"}
```

So the durable-submit claim is real. Also confirmed `ActionRegistry.walk` and `pluck` traverse `$platform` children generically, `renderToIR` normalizes single JSX children to arrays before validation, and every `Thread` post path (`post` / `update` / `postEphemeral` / `awaitChoice`) routes through `bindForPost`, so preflight coverage is complete.

### Blocking

**1. `rawAdaptiveCard()` is dead through the adapter — it posts an empty message.** `isPlainText` doesn't classify a `raw` node as rich, so the new `only?.type === "raw"` branch in `renderAdaptiveCard` (`adaptive-card.ts:47`) is unreachable from `teamsCodec.renderEgress` / `TeamsAdapter.render`. Probed:

```
IR: [{"type":"raw","props":{"value":{"type":"AdaptiveCard",...,"body":[{"type":"TextBlock","text":"hello from raw"}]}}}]
isPlainText: true
egress: {"text":""}
```

The card is silently dropped and Teams gets `{text: ""}`. Fix is one line — add `"raw"` to the `RICH` set in `isPlainText` (`adaptive-card.ts:440`) — but it needs a codec-level test; the current `native-adaptive-card.test.ts` only asserts `rawAdaptiveCard(card).raw === card`, i.e. validation, never delivery. This is also the same silent-drop class the PR elsewhere argues against.

**2. Three in-repo docs still specify the envelope this PR replaces.** `packages/channels-teams/docs/button-action-envelope.md` (the normative spec for exactly this wire format — lines 24, 47, 56, 58, 62), `ARCHITECTURE.md:102`, `README.md:132`, plus the `parseCardAction` JSDoc at `interaction.ts:44`. "Docs not in this slice" covers the docs site, not the package's own contract doc for the thing being changed.

**3. Teams becomes the only strict renderer, and it isn't called out.** `default:` now throws `UnsupportedUiNodeError` (`adaptive-card.ts:168`), while Slack (`block-kit.ts:300`), Telegram (`telegram.ts:418`) and Discord (`components-v2.ts:292`) all remain total renderers that skip. A shared tree that degrades on the other three now hard-fails the Teams post, and since handler throws propagate out of `onTurn`, that's a webhook 5xx and a platform redelivery. I think strict is the better default — but it should either land across the renderers or be flagged as a breaking behavior change in the PR body. Same for the `firstRowAsHeader` → `firstRowAsHeaders` fix (`adaptive-card.ts:315`): correct per the Adaptive Cards `Table` schema, but it changes rendering for column-less tables (the misspelled key meant the real property defaulted to `true`; it's now explicitly `false`), and it's currently invisible in the description.

### Should fix

**4. The happy path has no test.** `native-adaptive-card.test.ts` hand-stubs `(submit.props).onSubmit = { id: "ck:deploy" }`, so registry-binding → preflight → serialization is never exercised together — precisely the integration most likely to regress. It does work (probe above); please encode that probe as the test.

**5. Required child slots are omitted, which is the "opaque send error" class the validator claims to catch.** `withSlot` only emits the slot when `values.length > 0`:

```
{"body":[{"type":"Container"},{"type":"Table"}]}   // no items, no rows
```

`Container.items`, `Table.rows`, `ColumnSet.columns`, `FactSet.facts`, `ImageSet.images`, `RichTextBlock.inlines` are schema-required. Either emit `[]` or reject an empty container in `renderNode`.

**6. The reserved-namespace invariant isn't enforced on the raw path.** `assertInputId` blocks `id` values starting with `__copilotkit` for native inputs, but `assertAdaptiveCardPayload` applies no such rule, so a raw card can define `Input.Text id="__copilotkit"` and quietly break decoding (`parseCardAction` returns `undefined` → click silently ignored). Worth hoisting the id scan into the shared assert.

**7. `InteractionEvent.platform?` is optional while every sibling ingress type requires it** — `IncomingTurn.platform`, `IncomingCommand.platform`, `IncomingThreadStart.platform` are all `platform: string`. The optionality means any adapter that forgets to set it silently gets `adapter.platform` and, on the managed path, the wrong `thread.platform`. Making it required forces each adapter to be explicit (Teams is the only one updated here).

**8. `thread.platform` and `thread.capabilities` now disagree on the managed path.** `Thread.platform` becomes the source platform (`"teams"`), but `supportsBlockingChoice`, the capability flags, and the `postEphemeral` error string (`` `${this.platform} does not support ephemeral messages` ``) still come from the Intelligence adapter. Devs are now invited to branch on `thread.platform` against capabilities that describe a different adapter. Related and lower severity: the non-realtime egress fallback flattens IR via `irToText`, which yields `""` for a `$platform` tree (TextBlock text lives in `attributes`, not a `text` node) — harmless in practice since `renderSink` is wired by default in production, but it's a second empty-message path.

### Nits

- `@microsoft/teams.cards` is imported **type-only** in `ui.ts` but shipped as a hard runtime `dependency`. The `.d.ts` output does reference those types, so it can't just be a devDep — but say so in a comment, and note that the exact pin (the only one in that package) is deliberate because `TEAMS_SCHEMA_LOCK` names the version.
- `TEAMS_SCHEMA_LOCK.sha256` is dead metadata — nothing verifies it. Either add a test that pins it or drop the field.
- Unrelated lockfile churn: `vue-component-type-helpers` 3.3.7 → 3.3.8. Please revert.
- `renderAdaptiveCard` ends with `card as unknown as AdaptiveCardPayload`, and the portable path is the one branch that skips `assertAdaptiveCardPayload` — so portable cards get no cycle/size/`Action.Execute` check. Asymmetric with raw and native.
- `isRecord`'s `Object.getPrototypeOf(value) === Object.prototype` rejects class instances and null-prototype objects — including anything built with the `@microsoft/teams.cards` builder classes this PR just put on the dependency list.
- An unknown reserved envelope version (`metadata.version !== 1`) falls through to the legacy branch: the click is silently ignored *and* `__copilotkit` leaks into `values`. Prefer an explicit error.
- `createElement` destructures `data` off every element, so a `data` prop on a non-submit element is silently discarded.
- `makeOpenModal` (`create-channel.ts:497`) calls `renderToIR` with no registry and no preflight, so modal views bypass `validatePlatformUi` entirely. Pre-existing for the registry, but new for the validation boundary.
- Duplicated logic: the mixed-node check exists in both `renderAdaptiveCard:51` and `validatePlatformUi`, and `formatPath` in `native-adaptive-card.ts` duplicates the exported `formatPlatformPath`.
- Portable `renderInput` / `renderSelect` fall back to literal ids `"input"` / `"select"`, so two unbound inputs collide — while the native path forbids duplicates outright.
- `AdaptiveCard` is now both a deprecated type (from `/render`) and a component function (from `/ui`). Defensible, just worth a doc line.

### CI

`unit (20.x)` red is **unrelated**: `@copilotkit/web-inspector` › "does not load the empty overview video when reduced motion is preferred" (`web-inspector.spec.ts:2032`). Passes on 22.x and 24.x; nothing in this diff touches web-inspector. Branch is behind `main` but has zero file overlap with what landed since, so no conflict risk.

---

Nothing here is architectural rework — #1 is a one-line fix plus a test, #2 is doc updates, #3 is a description/scope decision.

tylerslaton:
Closing: this capability landed on main via #6331 (*feat(channels): add native channel JSX*), which supersedes this PR with generated native catalogs for both Slack and Teams — including the `Teams` Adaptive Card JSX namespace (`<Teams.AdaptiveCard>`, `Teams.Action.*`, `Teams.Chart.*`, …) this PR introduced. Shipped in `@copilotkit/channels@0.7.0`; see `packages/channels/native-catalogs.md` for the generated catalog.

Issue #636, opened 2024-09-30
⭐ 24 - Travel Budget Planner (Hacktoberfest Demo)


NathanTarbert:
### 🤩 Thank you for participating in the CopilotKit Hacktoberfest event. 

Follow [This Link to Get Started!](https://github.com/CopilotKit/CopilotKit/tree/main/community/content)🎉

#### NOTE: CopilotKit will allow multi-assign issues

  Please show your support and give a ⭐ to the CopilotKit repository.

### Description of app functionality:
Design a tool that helps users plan a budget for their trips, including airfare, lodging, and activities

#### Using CopilotKit
- **Required**: Your project must utilize CopilotKit as a core component.
- **Installation**: Include CopilotKit by following the installation guide at [CopilotKit Quickstart](https://docs.copilotkit.ai/quickstart).
- **Documentation**: Familiarize yourself with CopilotKit's features through the [Official Documentation](https://docs.copilotkit.ai).

#### Styling with Shadcn-UI
- **Required**: All UI components in your project should be styled using Shadcn-UI.
- **Installation**: Install Shadcn-UI [documentation](https://ui.shadcn.com/docs/installation)

![Swag_2](https://github.com/user-attachments/assets/d120bf97-95af-40a6-8618-6fb6a9198696)

vijaykr338:
can you assign this to me? I am a beginner

NathanTarbert:
@vijaykr338, please follow the link at the top to get started.

RohittCodes:
@NathanTarbert, I would like to work on this.

harshitajha4680:
I would like to work on this, please assign!

devshah207:
Can you assign me this issue please??

Abhrajitdas02:
Can you assign me this issue please??

UniquePratham:
Can you please assign this issue to me

UniquePratham:
@NathanTarbert  is there any way to get Open AI API key for free, because I dont have one and its seems like Copilotkit won't work without it , I want to build this project but the API key is holding me back, please help me!

NathanTarbert:
Hi @UniquePratham, if you set the model to 3.5 Turbo and use the free API key, it should work

Zedoman:
@NathanTarbert can you assign this to me?

Zedoman:
@quest-bot embark

Issue #638, opened 2024-09-30
⭐ 26 - Personal Finance Advice Bot (Hacktoberfest Demo)


NathanTarbert:
### 🤩 Thank you for participating in the CopilotKit Hacktoberfest event. 

Follow [This Link to Get Started!](https://github.com/CopilotKit/CopilotKit/tree/main/community/content)🎉

#### NOTE: CopilotKit will allow multi-assign issues

  Please show your support and give a ⭐ to the CopilotKit repository.

### Description of app functionality:
Build a financial advisor app that answers basic money management questions and suggests tips

#### Using CopilotKit
- **Required**: Your project must utilize CopilotKit as a core component.
- **Installation**: Include CopilotKit by following the installation guide at [CopilotKit Quickstart](https://docs.copilotkit.ai/quickstart).
- **Documentation**: Familiarize yourself with CopilotKit's features through the [Official Documentation](https://docs.copilotkit.ai).

#### Styling with Shadcn-UI
- **Required**: All UI components in your project should be styled using Shadcn-UI.
- **Installation**: Install Shadcn-UI [documentation](https://ui.shadcn.com/docs/installation)

![Swag_2](https://github.com/user-attachments/assets/d120bf97-95af-40a6-8618-6fb6a9198696)

Niharika0104:
Can you assign this issue to me?

steffy-lo:
Hi @NathanTarbert Can I work on this issue?

afrinashar:
I would like to work on this.

mutant138:
Can you please assign this to me also?
I am beginner I would like to work on this

smv1256:
Can you assign me this?

dha-stix:
Hi @NathanTarbert 
Can you assign this to me?

NathanTarbert:
Sure thing @dha-stix ;)

TRIPLE-ADE:
I would like to work on this, can you assign it

smv1256:
@quest-bot embark

PR #6220, opened 2026-07-29
feat(channels): unified Channel identity for Intelligence (OSS-643 Phase C)

BenTaylorDev:
Phase C of OSS-643. Wires the canonical Channel identity end to end, so a managed Slack/Teams turn runs as the person who sent it.

**Stacked on #6219 (Phase B)** — base is `ben1/oss-643-runtime-resolved-user`, so this diff shows only Phase C. Also depends on Intelligence [#636](https://github.com/CopilotKit/Intelligence/pull/636) (Phase A) being deployed, which is what sends `appUserId` and `conversationScope` on the claim.

## What this gives us

Before: a Channel turn ran with no Intelligence user, so memory and Enterprise Learning had nothing to attribute to. After: the sender's canonical id (`slack:T1:U9`) — the *same* id app-api wrote to `threads.end_user_id` — reaches the MCP memory headers.

| Commit | What |
| --- | --- |
| `7a7cbdc2e` | Identity from claim → envelope → `PlatformUser` → conversation store |
| `0a20a082d` | Runtime hook + privacy gate, optional `identifyUser`, isolation test, docs |

## Two ids, on purpose

`message.user.id` stays the **raw provider id** — mentions and every provider API call need it. `message.user.appUserId` is the canonical Intelligence identity, carried verbatim from app-api and **never re-derived here**: a second derivation drifts from canonical threads, and memory would attribute to a different user than the thread.

## Why per-turn middleware is safe here

`MCPMiddleware` bakes headers at construction, so per-turn identity needs a per-turn instance. That looked like it would force agent cloning — except `intelligence-adapter.ts` already mints a **fresh agent on every** `getOrCreate` and re-seeds history from the platform. There is no cross-turn agent reuse on the managed path, so nothing leaks between senders and the session model is untouched.

Identity travels as an explicit argument the whole way (`ChannelTurnContext` on `getOrCreate`), never adapter state — two senders' turns can be in flight at once, and ambient state would let one read the other's user.

## The privacy default

`channelMemoryPolicy` defaults to **`direct-only`**: a shared Slack channel or Teams group chat gets **no** user-private memory tools.

The reply in a group channel is visible to everyone in it, and the *model* decides when to call a memory tool. So attaching one there can disclose a sender's private data to their colleagues — triggered by their own message, with no moment where they chose to share it. There is a write-side version too: other participants' statements get written into whoever addressed the bot.

`shared` is an explicit operator opt-in and must never become the default. **This makes the headline capability invisible in group channels, which is where most Channel traffic lives** — that is a real product cost, accepted deliberately. OSS-649 (conversation-scoped memory) removes the tradeoff rather than gating it, and the option type is written to extend to a third `conversation` value.

Attachment also **requires** the canonical `appUserId`. The raw provider id is refused: Slack ids collide across workspaces and would not match `threads.end_user_id`.

## Channel-only runtimes may omit `identifyUser`

A managed turn carries its own verified sender, so demanding the callback forced a placeholder that attributed every turn to a fake user. Omitting it **without** channels still throws at construction; a mixed runtime that omits it fails HTTP Intelligence requests with an actionable 500 rather than running unauthenticated.

## Public API surface

Three additive, non-breaking changes. All three are shaped as the **general** per-turn identity path, not managed-only branches, because direct adapters are converging on managed delivery — migrating one is then: derive in `deriveAppUserId`, classify at ingress, forward the turn context.

- `PlatformUser.appUserId?`
- `ConversationStore.getOrCreate(..., turn?: ChannelTurnContext)` — optional, so today's slack/discord/telegram/whatsapp stores compile untouched
- `IntelligenceAdapterOptions.prepareAgentForTurn?`

## Testing

Run in `.worktrees/oss643`:

```
packages/runtime               Tests  1796 passed (1796)   (130 files)
packages/channels-core         Tests   161 passed (161)
packages/channels-intelligence Tests   214 passed (214)
packages/channels-ui           Tests    25 passed (25)

npx tsc --noEmit  (all four packages, clean)
```

New coverage per acceptance criterion:

- **AC 1/2** — `claim-mapping.test.ts`: canonical id carried without re-derivation, `user.id` stays raw, `appUserId` omitted when app-api could not scope it
- **AC 3** — `channel-turn-identity.test.ts` (9 cases): uses `appUserId` not the raw id, refuses a header-forging id, accepts a 200-char Teams id, falls back to the id when the sender has no display name
- **AC 4** — `channel-http-identity-isolation.test.ts`: **10 interleaved HTTP/Channel pairs**, asserting 20 distinct identities and that the Channel path never calls `identifyUser`. Interleaved, not sequential, so an ordering-dependent leak fails reliably rather than flaking green
- **AC 5** — privacy gate both directions (`direct` attaches, `shared` does not under the default, `shared` policy does), two concurrent senders get distinct agents, an absent scope defaults to `shared`, preparation failure still runs the turn

An exhaustive `toEqual` in `channel-activation-config.test.ts` now also pins the **fail-closed default**: the fake intelligence exposes no `ɵgetChannelMemoryPolicy`, so `direct-only` is asserted for a client that predates the option.

## Not verified — please weigh before un-drafting

**No end-to-end run against a live Intelligence.** Everything here is unit/integration level. The two places where "compiles and passes" is weakest evidence are the claim-time enrichment (Phase A) and this PR's privacy gate.

## Docs

`showcase/shell-docs/src/content/docs/channels/configuration.mdx` gains three sections: which id to use for what, the memory policy with its tradeoff stated plainly, and when `identifyUser` is optional.

## Out of scope

This unifies **identity, not execution**. Channel turns get user-scoped memory and learning, but still do not run *inside* the canonical thread the way an HTTP `/run` does — no fenced outer-run, no durable HITL resume, no selection pinning. That is OSS-599 §3, and per its boundary discipline it is built **in Intelligence, never as SDK-side helpers**. Do not "fix" this by routing Channel turns through `handleIntelligenceRun`.


BenTaylorDev:
> [!WARNING]
> **Stale base — do not un-draft and merge without a rebase and re-verification.** Flagged 2026-07-30.

This branch was authored **2026-07-29 ~03:54**, before Intelligence#638 (`feat(channels): replace queued delivery with live sessions`) landed. That PR moved managed provider egress into the realtime gateway and rewrote the SDK delivery path underneath this work — companion SDK PR #6236, merged.

**The specific hazard here is that this PR merges *clean*.** GitHub currently reports `MERGEABLE / CLEAN` for 23 files, +1008/-14. Its sibling Intelligence#636 (OSS-643 Phase A) touches four files that no longer exist on `main` — `canonical-thread-projection.ts`, its adversarial spec, `connector-outbox.integration.spec.ts`, `turn-metering.integration.spec.ts` — so that one will conflict loudly and announce itself. This one will not. A clean merge of stale work produces a silent wrong result rather than a red build, which is why it is worth a note rather than a rebase-when-convenient.

## Before this comes out of draft

1. **Rebase on current `main`** and re-run the suite; the pre-cutover results on this branch cannot be trusted.
2. **Check the overlap with what already shipped.** `main` now carries `app_user_id` on `cpki.channel_inbound_deliveries`, with the realtime-state constraint requiring it non-null from `prepared` onward, and both `canonical-thread-store.ts` and `prepared-delivery-store.ts` read it. #638 therefore already established app-api as the authority for a per-delivery app-user id — a large part of what the Phase A/C stack set out to build, reached by a different mechanism. Decide what remains rather than assuming the whole stack still applies.
3. **Confirm the phase ordering still holds.** #6219 (Phase B, `prepareAgentForTurn`) is also unmerged and is cited as a dependency by OSS-657.

## Still genuinely open regardless

`channelMemoryPolicy` has **zero** occurrences on `main`, so the memory-policy half of OSS-643 is unshipped — and OSS-649 (conversation-scoped memory) correctly still depends on it.

Context and the full file-level breakdown: OSS-643.

Issue #637, opened 2024-09-30
⭐ 25 - Code Snippet Generator (Hacktoberfest Demo)


NathanTarbert:
Follow [This Link to Get Started!](https://github.com/CopilotKit/CopilotKit/tree/main/community/content)🎉

🤩 Thank you for participating in the CopilotKit Hacktoberfest event. 

Please show your support and give a ⭐ to the CopilotKit repository.

### Description of app functionality:
Create a tool that suggests code snippets based on user prompts for various programming languages

#### Using CopilotKit
- **Required**: Your project must utilize CopilotKit as a core component.
- **Installation**: Include CopilotKit by following the installation guide at [CopilotKit Quickstart](https://docs.copilotkit.ai/quickstart).
- **Documentation**: Familiarize yourself with CopilotKit's features through the [Official Documentation](https://docs.copilotkit.ai).

#### Styling with Shadcn-UI
- **Required**: All UI components in your project should be styled using Shadcn-UI.
- **Installation**: Install Shadcn-UI [documentation](https://ui.shadcn.com/docs/installation)

RohittCodes:
Hey @NathanTarbert! I would like to work on this. Can you assign it to me?

PR #6221, opened 2026-07-29
feat(channels-intelligence): batch managed render output

MikeRyanDev:
## What changed

- replace the managed single-frame render sink with immutable ordered batches
- flush the first text at once and compact later text for up to 250 ms
- bound batches by frame, byte, and pending-lane limits
- add stable SHA-256 digests, batch IDs, durable high-water receipts, and the `render_batch_v1` runtime capability
- update HTTP, Realtime Gateway, fallback, and in-memory transports to the batch-only contract

## Why

Managed Channel latency currently grows with source delta count because each delta takes one durable round trip. Batching makes durable work track response time and semantic boundaries instead of token count.

Refs OSS-654. This PR follows the one-cutover contract in the companion service plan, so it does not provide the older-server sequential fallback described in that ticket.

Companion service PR: https://github.com/CopilotKit/Intelligence/pull/637

## Validation

- `pnpm test` in `packages/channels-intelligence` — 223 passed
- `pnpm run check-types` in `packages/channels-intelligence`
- pre-commit Nx test, publint, and attw suite — passed
- 20-delta and 200-delta runs each produced two durable batches and equal modeled receipt time

The service PR records the cross-repo E2E, load, fault, Kubernetes, and image checks.


BenTaylorDev:
Adversarial review of this PR paired with its companion CopilotKit/Intelligence#637. Findings that span the two repos are filed on whichever side owns the fix, with cross-links; the server-side half of this review is on #637.

One recalibration up front: managed Intelligence-based channels aren't released yet, so I'm **not** treating the batch-only capability gate as a customer-facing break. With an internal-only population and both sides publishing together, the hard cutover is a legitimate call and the companion ADR's "one release set" premise holds.

## Blocker — `MessageRef.id` no longer matches the key the Outbox records posts under

This is the one I'd hold the merge on. It spans both PRs; the SDK half is here.

`postRenderFrame` now reads:

```ts
id: receipt.egressOperationId || `${target.turnId}:main:${seq}`
```
`src/intelligence-adapter.ts:686`

Combined with #637 making `egressOperationId` required and always populated, that fallback is unreachable — so `post()` returns `eop_…`, which is **constant for the whole turn**, since one egress operation covers a (turn, slot).

The Connector Outbox resolves update/delete refs by `${turnId}:${slot}:${seq}` and stores them in a flat jsonb map keyed on exactly that string (Intelligence `apps/app-api/src/workers/connector-outbox.ts:1107`). I went looking for a secondary key on the operation id specifically to rule this finding out — `recordRef`/`loadRef` and `findEgressRefByProviderTs` all key on the frame ref alone. So a same-turn `update`:

- **Slack** — `loadRef('eop_…')` misses; `eop_…` can't match `/^\d+\.\d+$/`; `ts` is undefined → the edit is logged and **silently skipped**.
- **Teams** — `ts = 'eop_…'` → activity PUT against a nonexistent activity id.

`delete` resolves the same way, so it breaks identically. `<Message onReaction>` is likely affected too — the reverse-lookup docstring notes handlers are persisted under the post-time ref.

On `main` this was correct by construction: the acceptance INSERT never set `egress_operation_id`, so the receipt omitted it for `post` frames and `??` yielded `turnId:main:seq`.

Worth calling out why CI is green on both sides: `InMemoryRenderEventSink` fabricates the same `eop_${batch.turnId}`, so the SDK suite locks in the broken value, and #637's `renderBatchStream` fixtures contain zero `update`/`delete` events and zero `ref:` values. Nothing in either suite carries a real ref across the boundary.

Suggested guard: post → same-turn update where the update's ref comes from the post's returned `MessageRef`, asserting it equals `turnId:main:seq`.

## Retry loop amplifies at exactly the wrong moment

`flushBatch` retries `pushBatch` up to 3 times with **no backoff**, and classifies permanence by substring match on `CHANNEL_RENDER_BATCH_CONFLICT`, `CHANNEL_RENDER_SEQUENCE_GAP`, and `-> 409`.

#637 adds a gateway admission limiter that rejects with `"reason" => "app_api_busy"` (`app_api_capacity_limiter.ex:85-87`). That isn't in the permanent set, so it's treated as transient: every lane triples its request rate the instant App API hits capacity. That's a congestion-collapse shape, and it's a few lines to fix — classify `app_api_busy` as retry-with-backoff and add even a small delay between attempts.

Matching on `"-> 409"` is also brittle: it's a magic substring of one transport's error formatting, and the realtime path wraps errors differently. A typed/coded error would survive both transports.

## Minor

- **`run_started` skips the pending-text flush.** In `enqueue`, every non-text kind calls `flushPendingText()` before `addFrame` except `run_started`, which returns early. If any text is buffered when one arrives it gets a lower `seq` than text that preceded it. Unreachable in a normal single-run turn, but the asymmetry has no comment explaining it.
- **`.changeset/fast-bats-wave.md` is inert.** There's no changesets CLI dependency in the root `package.json` and no CI reference — the tooling was removed a while back and the two files on `main` are leftovers. It won't be consumed, so it isn't carrying the release note it looks like it's carrying.
- **Pre-release makes a cleanup free:** since nothing published depends on the old shape, consider dropping `RenderFrame.deliveryId` / `turnId` / `slot` outright rather than shipping them `@deprecated` on day one. Keeping them is what invites the identity confusion behind the blocker above.

## Deploy ordering

Neither PR states it, and it's the reverse of the usual instinct: **Intelligence first, then publish this package, then upgrade internal runtimes.** Server-first means old runtimes are cleanly rejected until they upgrade; SDK-first publishes a package that talks to nothing, since this PR ships no old-server fallback by design.

## Credit

First-token latency is genuinely preserved — `firstTextFlushed` flushes immediately and only later text compacts. Batch immutability across retries is done properly: stable `batchId`, content digest, frozen frames. I also went looking for a `batchId` collision across turns and couldn't construct one — distinct turns get distinct egress operations, and within one operation `startSeq` differs. That part holds up.

BenTaylorDev:
Proposed fix for the blocker I raised above (the `MessageRef.id` / `postRef` mismatch). Built and verified on a worktree off `25e02778`; the Intelligence-side counterpart is proposed on CopilotKit/Intelligence#637.

The fix is smaller than the finding: since the Connector Outbox keys every post's provider message id on `turnId:slot:seq`, the SDK should always mint that, and the `egressOperationId` branch should go away entirely rather than be re-ordered. On `main` this worked only by accident — the receipt happened to omit `egressOperationId` for `post` frames.

```diff
--- a/packages/channels-intelligence/src/intelligence-adapter.ts
+++ b/packages/channels-intelligence/src/intelligence-adapter.ts
@@ -676,14 +676,22 @@ export class IntelligenceAdapter implements PlatformAdapter {
     * so the Connector Outbox renders it as Block Kit — preserving rich JSX/IR
     * instead of flattening to text. Returns a {@link MessageRef} keyed to the
     * frame so a later update/delete can re-address it.
+   *
+   * The id MUST be the frame key `turnId:slot:seq`, never the receipt's
+   * `egressOperationId`: the Connector Outbox records each post's provider
+   * message id under exactly that key (`recordRef(\`${turnId}:${slot}:${seq}\`)`)
+   * and resolves a later update/delete by looking it up. One egress operation
+   * covers the whole (turn, slot), so keying on `egressOperationId` would give
+   * every post in a turn the same ref and resolve to nothing — Slack silently
+   * skips the edit, Teams PUTs a nonexistent activity id.
     */
    private async postRenderFrame(
      target: ChannelReplyTarget,
      event: ChannelRenderEvent,
    ): Promise<MessageRef> {
-    const { receipt, seq } = await this.pushRenderFrame(target, event);
+    const { seq } = await this.pushRenderFrame(target, event);
      return {
-      id: receipt.egressOperationId || `${target.turnId}:main:${seq}`,
+      id: `${target.turnId}:main:${seq}`,
        __route: target.route,
        __turnId: target.turnId,
        __deliveryId: target.deliveryId,
```

## The guard

The existing delete test is why this got through — it asserts the delete frame carries `ref.id`, which is a tautology that passes for any value including a colliding one:

```ts
expect((deleteFrame?.event as { kind: "delete"; ref: string }).ref).toBe(ref.id);
```

So the new test asserts the concrete format and, more importantly, that two posts in one turn don't collapse onto the same ref:

```ts
it("keys each post's MessageRef to its own frame so the Outbox can resolve a later update", async () => {
  const renderSink = new InMemoryRenderEventSink();
  const adapter = intelligenceAdapter({
    source: new InMemoryDeliverySource(),
    egress: new InMemoryEgressSink(),
    renderSink,
  });
  const card = [
    { type: "section", props: { children: "card" } },
  ] as unknown as Parameters<typeof adapter.post>[1];

  const first = await adapter.post(target, card);
  const second = await adapter.post(target, card);
  await adapter.update(second, card);

  // Distinct posts must not collapse onto one ref.
  expect(first.id).not.toBe(second.id);
  expect(first.id).toMatch(/^turn_[A-Za-z0-9_-]+:main:\d+$/);
  expect(second.id).toMatch(/^turn_[A-Za-z0-9_-]+:main:\d+$/);

  // The update must re-address the second post's frame key.
  const updateFrame = renderSink.frames.find((f) => f.event.kind === "update");
  expect((updateFrame?.event as { kind: "update"; ref: string }).ref).toBe(
    second.id,
  );
});
```

## Verification

Red-green, both directions:

- **Against this PR's current code** the new test fails exactly as predicted — `AssertionError: expected 'eop_turn_t1' not to be 'eop_turn_t1'`. Both posts really do collapse onto one ref.
- **With the fix**: `vitest run` → **13 files, 224 passed** (this PR reports 223; the +1 is the new test). No existing test needed changing, which is worth noting — the fix restores what `main` did by accident, so nothing was depending on the new behavior.
- `tsc -p tsconfig.check.json --noEmit` is unchanged: the only two errors are the pre-existing `seedsInboundTurn`/`ConversationStore` pair, identical on the pristine branch and absent from CI.

Happy to push this to `agent/channels-realtime-batches` if you'd rather not hand-apply — say the word. Not pushing to your branch unasked.

One thing this fix does **not** address, and I don't think it should: `RenderBatchAccepted.egressOperationId` is still required and still returned on every batch. That's fine for the receipt (the worker genuinely needs it), it just isn't a message identity. If you want to make the distinction impossible to get wrong again, the stronger version is a nominal type — `type RenderFrameKey = string & { readonly __brand: "RenderFrameKey" }` on `MessageRef.id` — so an `eop_…` can't be assigned to it at all. Probably a follow-up rather than this PR.

.changeset/fast-bats-wave.md
tylerslaton, line 1: lets delete these, we don't have changesets anymore.

packages/channels-intelligence/src/intelligence-adapter.ts
```diff
@@ -939,40 +957,179 @@ export class IntelligenceAdapter implements PlatformAdapter {
...
+        if (
+          pendingText &&
+          (pendingText.messageId !== event.messageId ||
+            Buffer.byteLength(pendingText.delta + event.delta, "utf8") >
+              RENDER_TEXT_MAX_BYTES)
+        ) {
+          flushPendingText();
+          flushBatch();
+        }
+        pendingText = pendingText
```
tylerslaton, line 1177: Blocker: the 250 ms timer makes compaction boundaries depend on wall-clock scheduling. On a full delivery redelivery, seq starts at 0, but batches accepted during the prior attempt remain in Intelligence. If timing changes, this line builds different seq payloads and batch IDs; Intelligence only dedupes the exact prior batch ID and then requires `startSeq === acceptedHighWater + 1`, so the redelivery fails with `CHANNEL_RENDER_SEQUENCE_GAP`. The new replay test retries the same batch object; it does not rerun a delivery under different timing. Keep semantic frames 1:1, make boundaries deterministic, or add a durable resume/overlap protocol and a real two-attempt test.

packages/channels-intelligence/src/render-batches.ts
```diff
@@ -0,0 +1,70 @@
...
+} from "./contracts.js";
+
+export const RENDER_BATCH_PROTOCOL_CAPABILITY = "render_batch_v1";
+export const RENDER_TEXT_FLUSH_MS = 250;
+export const RENDER_TEXT_MAX_BYTES = 16 * 1024;
+export const RENDER_BATCH_MAX_FRAMES = 64;
+export const RENDER_BATCH_MAX_BYTES = 128 * 1024;
+export const RENDER_LANE_MAX_PENDING_BYTES = 256 * 1024;
+
+const canonicalFrames = (frames: readonly RenderFrame[]): string =>
```
tylerslaton, line 40: Blocker across the companion PR: this is not canonical across JavaScript → Elixir → JavaScript. The Gateway decodes frames into Elixir maps and Jason re-encodes them; map key order can change. App API then hashes `JSON.stringify(input.input.frames)`, and rich `content` is parsed as a record, so semantically identical content can produce a different digest and a 409 conflict. Define canonical JSON with recursive key ordering on both sides, or hash a byte-preserving payload, and add a rich-content test through the real Gateway.

packages/channels-intelligence/src/intelligence-adapter.ts
```diff
@@ -671,7 +683,7 @@ export class IntelligenceAdapter implements PlatformAdapter {
   ): Promise<MessageRef> {
     const { receipt, seq } = await this.pushRenderFrame(target, event);
     return {
-      id: receipt.egressOperationId ?? `${target.turnId}:main:${seq}`,
+      id: receipt.egressOperationId || `${target.turnId}:main:${seq}`,
```
tylerslaton: Blocker for the paired cutover: `egressOperationId` is required and scoped to the whole turn/slot, while the worker records provider refs by frame key (`turnId:slot:seq`). This therefore returns the same id for every post and later update/delete events cannot address the right provider message. Return the frame key here; the fallback is currently unreachable.

tylerslaton (changes requested):
Reviewed with [CopilotKit/Intelligence#637](https://github.com/CopilotKit/Intelligence/pull/637) as one cutover. This materially reduces durable receipts from the raw AG-UI delta count to bounded batches. I’m requesting changes because the pair still breaks correctness on delivery redelivery and rich-content digest checks, and the MessageRef mismatch prevents update/delete addressing. The companion review covers worker/provider safety and the missing full-path capacity proof. These PRs should not merge independently.

packages/channels-intelligence/src/intelligence-adapter.ts
tylerslaton: Fixed in `212ae3a9`: `postRenderFrame` now returns `${turnId}:main:${seq}` and adds a focused update-addressing test. This thread no longer blocks.

tylerslaton (changes requested):
Current-head update after `212ae3a9`: the frame-key MessageRef fix and bounded retry backoff address two earlier issues. Two paired blockers remain unchanged: the 250 ms compaction boundary still depends on delivery timing and can conflict with Intelligence’s accepted high-water on redelivery, and the batch digest still lacks canonical object-key ordering across the Gateway for rich content. The blocking inline threads from the prior review still apply. The companion Intelligence review also remains blocking, so keep these PRs coupled.

.changeset/fast-bats-wave.md
MikeRyanDev: Deleted in `212ae3a92`.

packages/channels-intelligence/src/intelligence-adapter.ts
MikeRyanDev: Fixed in `4c65bdd09`. The wall-clock timer is gone; batch boundaries now use semantic events, 16 KiB, or a deterministic 256-delta cap. The test now pins the count boundary, so replaying the same frame stream cannot shift batches with scheduler timing.

packages/channels-intelligence/src/render-batches.ts
MikeRyanDev: Fixed across both PRs. `4c65bdd09` uses recursive stable JSON before the SDK hash; Intelligence `0a021b9f` applies the same canonical encoding after Gateway decoding. Tests on both sides reorder nested rich-content keys, so Elixir map key order no longer affects the digest.

packages/channels-intelligence/src/intelligence-adapter.ts
MikeRyanDev: Confirmed fixed in `212ae3a92`; the returned ref is the frame key `${turnId}:main:${seq}`.

BenTaylorDev:
One more finding, and I think it outranks everything else I've posted here — including the `MessageRef` blocker. Server-side counterpart on CopilotKit/Intelligence#637. I've also written this up on OSS-654, OSS-653, OSS-652 and OSS-648.

## The batching mechanism is the one OSS-654 was written to reject

OSS-654 is unusually specific about this, because it had already been tried. From the ticket:

> **Why coalescing didn't work** — The first attempt merged adjacent same-message `text_delta` frames into one frame. It worked on cost — 123 frames to 4, a production turn to 3.36s — and it broke replay. Reverted in CopilotKit#6212.
>
> **The invariant this establishes: a frame's payload at a given seq must never depend on timing.** `render-frame-replay.test.ts` in CopilotKit#6212 enforces it.
>
> **Chosen mechanism: batch the transport, not the content.** Send the buffered frames as an **array in one request**. Frames stay 1:1 with tokens, so `seq k` always carries `delta k` regardless of how pushes are grouped — replay stays byte-identical.

This PR batches the transport *and* coalesces the content. `canCompactText` merges consecutive same-`messageId` deltas, and `pendingText.delta + event.delta` accumulates across a 250 ms window (`RENDER_TEXT_FLUSH_MS`) or 16 KiB (`RENDER_TEXT_MAX_BYTES`). So a frame's payload at a given seq is now a function of wall-clock arrival — exactly what the invariant forbids.

## The guard test was replaced with a weaker one

| | Test |
| --- | --- |
| `main` | `describe("render frame replay stability (OSS-648)")` → `it("does not reuse a seq with a different payload when a turn is redelivered")` |
| this PR | `describe("render batch retry stability")` → `it("keeps an unacknowledged batch immutable and returns the original high water on exact retry")`, `it("changes the stable batch identifier when compacted content changes")` |

Both replacements test **within-run retry** of the same batch object. Neither replays a redelivery, which is what the deleted test existed for. And `it("changes the stable batch identifier when compacted content changes")` asserts the divergence as intended behavior — it pins the thing the old test forbade.

"`render-frame-replay.test.ts` stays green" is acceptance criterion #1 on OSS-654.

## Traced failure path

1. `dispatch()` resets `nextSeq: 0` on every delivery (`intelligence-adapter.ts:418`) — the comment there says "so egress ids are deterministic across redelivery". `turnId` is `turn_<deliveryId>`, so a redelivery replays into the **same seq space**.
2. Server-side, the egress operation is keyed on `idempotency_key = turnId:slot`, so the redelivery reuses the same row with `accepted_high_water` **already advanced** by attempt 1.
3. Different token timing → different compaction → different digest → **different `batchId`** → the duplicate lookup (`WHERE batch_id = $2`) misses → falls through to `if (startSeq !== operation.acceptedHighWater + 1) throw CHANNEL_RENDER_SEQUENCE_GAP` (Intelligence `service.ts:4571`).
4. `flushBatch` classifies `CHANNEL_RENDER_SEQUENCE_GAP` as **permanent** → `transportError` → `drain()` throws → nack. With `skipIngressDedup = true` the redelivery reruns → same divergence → nack loop.

That last step is the livelock the adapter's own `run_error` comment warns about a few lines up ("nacking would livelock — redeliver → same failure → nack, forever").

**This is stricter than `main`, not equivalent.** `main` needed byte-identical *frames* and got them for free from 1:1 mapping. This PR needs byte-identical *batch boundaries*, which are timing-dependent by construction. Redeliveries aren't exotic: lease expiry, runtime crash mid-turn, gateway disconnect.

## Options

I don't think the answer is necessarily "drop compaction" — the cost win is real and worth keeping. Two routes that preserve it:

- **Deterministic grouping** — boundaries from a clock-independent rule (every K deltas, or N chars) instead of a 250 ms timer. OSS-654 considered and rejected this for leaving seq gaps, but that objection may not bind anymore: `nextFrameSeq` fires once per *emitted frame* here, so seqs stay dense either way. Worth re-deriving. It does cost first-token latency unless the first delta stays its own group — which this PR already does via `firstTextFlushed`, so the two compose.
- **Supersede on replay** — let the server accept a replay that replaces an incomplete prior attempt rather than rejecting on high-water. OSS-654 weighed a version of this ("don't reset seq on replay") and flagged a duplicated-text risk in the rendered Slack message.

Either way the concrete first step is the same: **restore the deleted redelivery test.** It turns this from an argument into a red test, and it'll tell you immediately whether whatever grouping rule you land on is actually replay-stable. Happy to do that restore on top of the `MessageRef` patch I posted earlier if useful.

## Two smaller things in the same area

- **`run_started` skips the pending-text flush.** Every other non-text kind calls `flushPendingText()` before `addFrame`; `run_started` returns early, so buffered text would get a *higher* seq than a `run_started` that arrived after it. Unreachable in a normal single-run turn, but it's an ordering asymmetry with no comment.
- **Two paths consume a seq and then discard the frame**, leaving a permanent hole: `CHANNEL_RENDER_FRAME_TOO_LARGE` (`nextFrameSeq` runs before the size check in `addFrame`) and `CHANNEL_RENDER_LANE_OVERLOAD` (`flushBatch` abandons an already-sequenced batch). Both nack the turn today so they aren't live gaps — but OSS-653's Fix B is a contiguity-waiting reader, and it would stall on them.

packages/channels-intelligence/src/render-batches.ts
MikeRyanDev: Follow-up: Intelligence CI exposed two E2E runtime emulators that still used plain `JSON.stringify`. `6d0659f2` moves canonical encoding into the shared render contract and makes App API plus both direct/Gateway test senders use it. The local contract, service, integration, lint, and type checks pass; the full E2E cannot start locally because managed OIDC returns `/auth/error?error=Configuration`, so the new exact-head CI run is the authoritative vertical check.

packages/channels-intelligence/src/intelligence-adapter.ts
tylerslaton: This fixes replay determinism, but it now breaks the paired PRD's explicit 250 ms tail-flush bound. For a 20–200 delta response below 16 KiB, the SDK sends the first text and then waits until `text_end`/finalize; at 10 deltas/s that creates about a 2–20 second platform update gap. The current 20/200 test emits synchronously and immediately finalizes, so it cannot catch the gap. This still blocks launch: keep replay safe while restoring a wall-clock output bound, and cover a paced full-delivery retry/resume.

tylerslaton (changes requested):
Re-reviewed with #637 as Mike's tactical launch bridge. The architecture split is the PRD's intended middle ground, and the earlier replay/message-ref/digest issues are fixed.

One launch blocker remains on this head: replacing the 250 ms tail flush with a 256-delta threshold makes a normal paced response go silent after its first delta until `text_end` or finalization. That violates both the PRD's batching contract and its <=1 second visible-update-gap gate. The existing thread has the paced 20/200-delta case that needs coverage.

Once that is fixed, I consider the SDK half of the bridge ready to merge with the paired service change. Launch enablement still depends on the paired full-path capacity, live provider, and cutover gates.

packages/channels-intelligence/src/intelligence-adapter.ts
```diff
@@ -616,15 +640,18 @@ export class IntelligenceAdapter implements PlatformAdapter {
...
+  ): Promise<{ receipt: RenderBatchAccepted; seq: number }> {
     const seq = this.nextFrameSeq(target.turnId);
-    const receipt = await this.requireRenderSink().push({
-      deliveryId: target.deliveryId,
-      turnId: target.turnId,
-      slot: "main",
-      seq,
-      event,
-    });
+    const receipt = await this.requireRenderSink().pushBatch(
```
tylerslaton: Paired launch blocker: this direct batch send bypasses the promise chain inside `createRunRenderer()`. During a run, a tool can call `thread.post`/update/file while an earlier renderer batch is still in flight; both share `nextFrameSeq`, but the higher sequence request can reach App API first and fail with `CHANNEL_RENDER_SEQUENCE_GAP`. Two concurrent direct posts have the same race. The PRD allows one in-flight batch for the whole delivery/turn/slot lane, so use one shared lane queue for renderer and discrete effects, and test a blocked first push with a concurrent post.

tylerslaton (commented):
Additional paired contract blocker: lane serialization currently covers only batches created inside `createRunRenderer`; direct render effects use a separate path.

tylerslaton (changes requested):
Re-reviewed with CopilotKit/Intelligence#637 at this exact head. The worker split is the PRD's intended launch bridge, but this SDK head is not launch-ready:

1. The deterministic 256-delta boundary preserves replay, but removes the PRD's 250 ms tail bound. A paced 20–200-delta response below 16 KiB can leave Slack/Teams unchanged for 2–20 seconds after the first delta.
2. The renderer promise chain is not shared by direct post/update/file effects. A tool call can send a higher sequence while a lower renderer batch is still in flight.

The fix must satisfy both replay-safe accepted-prefix recovery and bounded visible updates, with one request in flight for the whole delivery/turn/slot lane. The existing inline threads include the focused regression cases.

packages/channels-intelligence/src/intelligence-adapter.ts
```diff
@@ -939,40 +974,168 @@ export class IntelligenceAdapter implements PlatformAdapter {
...
+          : event;
+        pendingTextDeltaCount += 1;
+        if (pendingTextDeltaCount >= RENDER_TEXT_MAX_DELTAS) {
+          flushPendingText();
+          flushBatch();
+        }
+        return;
+      }
+
+      if (event.kind === "run_started") {
```
BenTaylorDev, line 1189: `run_started` is the only non-text event that calls `addFrame` without a following `flushBatch()`. Every other branch below flushes.

So `run_started` sits in `batchFrames` until the first text delta or first semantic event arrives. On `main` it was pushed immediately and drove the Outbox's `assistant.threads.setStatus` ("is thinking…"). Now that signal is delayed by the model's full time-to-first-token.

This compounds with a finding on the service PR: `setIngressThinkingStatus` in app-api is now a silent no-op, because `ingressStatusDeps` is only assigned inside `startConnectorOutboxBootstrap`, which app-api no longer calls. With both changes in, the user sees *nothing at all* between sending a message and the model's first token reaching the provider — on the PR whose purpose is latency.

Flush after `run_started`.

packages/channels-intelligence/src/intelligence-adapter.ts
```diff
@@ -939,40 +974,168 @@ export class IntelligenceAdapter implements PlatformAdapter {
...
-          metrics?.recordPush(
-            event.kind,
-            pushStartedAt,
-            Date.now(),
-            event.kind === "text_delta" ? event.delta.length : 0,
-          );
+          lanePendingBytes -= batchBytes;
+          const pushFinishedAt = Date.now();
+          for (const frame of batch.frames) {
+            metrics?.recordPush(
```
BenTaylorDev: This now attributes the whole batch round-trip to **every frame in the batch**. `pushStartedAt`/`pushFinishedAt` are per-batch, and the loop calls `recordPush` once per frame with the same pair, so a 60-frame batch reports 60 x the batch RTT.

This is the OSS-648 instrumentation that exists specifically to measure whether batching worked, and it now over-reports post-fix per-frame latency by the batch factor — in the direction that hides the improvement, so it won't look like a bug in the numbers.

Record the batch RTT once (batch-level metric with a frame count), or divide, but don't multiply.

packages/channels-intelligence/src/intelligence-adapter.ts
```diff
@@ -939,40 +974,168 @@ export class IntelligenceAdapter implements PlatformAdapter {
...
+          let receipt: RenderBatchAccepted | undefined;
+          for (let attempt = 1; attempt <= 3; attempt += 1) {
+            try {
+              receipt = await sink.pushBatch(batch);
+              break;
+            } catch (err) {
+              const message = err instanceof Error ? err.message : String(err);
+              const permanent =
+                message.includes("CHANNEL_RENDER_BATCH_CONFLICT") ||
+                message.includes("CHANNEL_RENDER_SEQUENCE_GAP") ||
```
BenTaylorDev: Two issues with this classifier.

1. It string-matches error messages, including the transport-specific `"-> 409"`. The realtime-gateway path doesn't produce that shape (errors come back through `normalize_error` on the Elixir side), so a genuine 409 over the gateway falls through to `permanent === false` and burns three attempts plus backoff before failing anyway. Classify on a structured code, not `message.includes`.

2. Treating `CHANNEL_RENDER_SEQUENCE_GAP` as permanent is what turns the retry-of-a-partially-rendered-turn case into a hard dead-letter rather than a recoverable resume. See the PR-level comment: because `turnId` is stable across delivery attempts and `accepted_high_water` persists, a retry whose model output diverges hits this on its very first batch. A gap is the one error where the server actually knows the answer — it should hand back its high-water so the SDK can resume, not be treated as terminal.

BenTaylorDev (dismissed):
Independent re-review of the current head, paired with Intelligence#637. I reproduce all of Tyler's open blockers, so this stays blocked. Adding three SDK-side findings not yet raised, plus an argument that the fix which created the launch blocker doesn't buy what it cost.

## The determinism trade was a bad deal

Swapping the 250 ms tail flush for a 256-delta cap was done to make batch boundaries stable across a redelivery. It does not achieve that, because the nondeterminism that matters isn't the scheduler's.

On the service side `turnId = 'turn_' + rawDeliveryId`, and a failed delivery returns to `retry_wait` on the **same row** (`service.ts` fail path, `WHERE id = $1`). So `deliveryId`, `turnId`, and the `${turnId}:${slot}` egress operation with its `accepted_high_water` all survive the retry. Attempt 2 re-runs the agent (`skipIngressDedup`), the model emits a different token stream, digests differ, `existingBatch` misses, and `startSeq (0) !== acceptedHighWater + 1` throws `CHANNEL_RENDER_SEQUENCE_GAP` — which this PR classifies `permanent`, so it nacks straight through to dead-letter.

That hazard is pre-existing (`main` has the same shape via `CHANNEL_RENDER_FRAME_CONFLICT`), so it isn't a regression here — but that is exactly the point. Deterministic boundaries bought no redelivery safety, and the price was the streaming silence Tyler is blocking on. Suggest restoring a wall-clock output bound and solving resume with an explicit high-water handshake (server tells the SDK where to resume) rather than with boundary determinism.

## Confirming the open blocker

Traced the paced case directly: `run_started` + delta 1 flush, then deltas 2..255 accumulate in `pendingText` with no time bound. `text_end`, `tool_start`, `post`, `update` all flush — so a tool-using agent still paints intermittently, but a plain text answer, which is the common case, emits its first token and then nothing until the run ends.

## Also still open

`postRenderFrame`/`pushRenderFrame` bypass the `createRunRenderer` chain while sharing `nextFrameSeq` — the lane race Tyler flagged at line 645.

## What's right

`stableRenderJson` with recursive key ordering is the correct fix for the Elixir round-trip, and it is applied on both sides. Immutable frozen batches with content-derived ids, the receipt cross-check, and the frame-key MessageRef all hold up under re-read. The earlier digest, MessageRef, and replay fixes are genuinely fixed.

packages/channels-intelligence/src/intelligence-adapter.ts
MikeRyanDev: Fixed in `a06f78be4`. Renderer batches and discrete post/update/file/delete effects now share one per-turn render-lane tail. The regression test blocks the first renderer push and proves a concurrent post cannot overtake it.

packages/channels-intelligence/src/intelligence-adapter.ts
MikeRyanDev: Fixed in `a06f78be4`. `run_started` now flushes immediately, with coverage that pins it as its own first batch before text arrives.

packages/channels-intelligence/src/intelligence-adapter.ts
MikeRyanDev: Fixed in `a06f78be4`. Metrics count all frames but record one round-trip latency per batch; frames mode now logs one batch record with frame count and kind totals.

packages/channels-intelligence/src/intelligence-adapter.ts
MikeRyanDev: Addressed the structured-error issue in `a06f78be4`: HTTP and realtime transports preserve provider error codes, and retry classification reads those codes instead of message text. I did not add a high-water resume protocol here; that changes the cross-repo wire and replay contract and is outside this performance PR. Sequence gaps remain fail-closed under the existing v1 contract.

packages/channels-intelligence/src/realtime-gateway.ts
```diff
@@ -822,8 +836,10 @@ export async function connectRealtimeGateway(
           .receive("ok", (reply: unknown) => resolve(reply))
           .receive("error", (reason: unknown) =>
             reject(
-              new Error(
-                `realtime gateway session push ${event} failed: ${safeReason(reason)}`,
+              new RealtimeGatewayPushError(
+                event,
+                extractReasonCode(reason),
```
BenTaylorDev, line 841: **This fix works on the HTTP path but not on the managed realtime path**, which is the one production uses.

`extractReasonCode` only reads `reason` (when it's a string) or `reason.reason`. But the gateway's `app_api_client.ex` builds its error like this:

```elixir
defp error_response(status, response_body) do
  %{
    "reason" => "app_api_error",
    "status" => status,
    "code" => get_in(parsed, ["error", "code"])
  }
end
```

and `normalize_error/1` passes a map straight through. So `RealtimeGatewayPushError.code` is always the literal `"app_api_error"`, and `isPermanentRenderBatchError` is always `false` over the gateway. A permanent `CHANNEL_RENDER_SEQUENCE_GAP` or `CHANNEL_RENDER_BATCH_CONFLICT` now burns three attempts plus 100 ms + 200 ms of backoff before failing.

That's mildly worse than what it replaced: the old `message.includes("-> 409")` at least caught the HTTP shape, and the real code is sitting right there in the payload being discarded.

```diff
-    const inner = (reason as { reason?: unknown }).reason;
-    if (typeof inner === "string") return inner;
+    const coded = reason as { code?: unknown; reason?: unknown };
+    if (typeof coded.code === "string") return coded.code;
+    if (typeof coded.reason === "string") return coded.reason;
```

Worth a test that drives a 409 through the gateway transport rather than the HTTP one — the current coverage exercises the shape that already worked.

BenTaylorDev (commented):
Re-reviewed `a06f78be` against my notes on `39d7bf31`. Four of my findings are cleanly fixed:

- `run_started` now flushes its batch
- `recordPush` -> `recordBatch`, so batch latency is no longer multiplied by frame count
- `IntelligenceHttpError` / `RealtimeGatewayPushError` carry a structured `code`, replacing the `message.includes` classifier
- `enqueueRenderLane` is now shared by the renderer chain and the discrete `pushRenderFrame` effects, which closes Tyler's lane race at the old line 645 — the seq allocator and the transport queue finally agree

Two things still open, one of them new in this head.

**New: the structured-code fix doesn't reach the production path.** Detail inline on `realtime-gateway.ts`. The HTTP path works; the managed realtime path extracts the literal string `"app_api_error"` as the code, every time. One-line fix.

**Tyler's launch blocker is unchanged.** No wall-clock output bound was restored — `setTimeout` on this head appears only in `waitForRenderBatchRetry`. Deltas 2..255 still buffer until `text_end` or the 256 cap, so a paced plain-text answer emits its first token and then goes quiet for the rest of the run.

I'll re-state my view from the last round, since the two are linked: the 256-delta cap was adopted to make batch boundaries deterministic across a redelivery, and it doesn't achieve that. `turnId` is stable across delivery attempts on the service side and `accepted_high_water` persists, so a retry re-runs the model, output diverges, and the first batch fails on the sequence check regardless of how deterministic the boundaries are. Restoring the timer costs nothing that was actually being bought.

Nothing else in this head regressed on re-read — the `enqueueRenderLane` tail correctly absorbs rejections for queue continuity while still rejecting the caller, and `renderLaneTails` is cleaned up alongside `renderState` on turn completion.

packages/channels-intelligence/src/intelligence-adapter.ts
```diff
@@ -939,40 +1018,167 @@ export class IntelligenceAdapter implements PlatformAdapter {
...
+              RENDER_TEXT_MAX_BYTES)
+        ) {
+          flushPendingText();
+          flushBatch();
+        }
+        pendingText = pendingText
+          ? { ...pendingText, delta: pendingText.delta + event.delta }
+          : event;
+        pendingTextDeltaCount += 1;
+        if (pendingTextDeltaCount >= RENDER_TEXT_MAX_DELTAS) {
```
BenTaylorDev, line 1181: Still the only time-independent flush trigger for text, so Tyler's launch blocker stands on this head. Confirmed by grep: the only `setTimeout` in this file is `waitForRenderBatchRetry`.

Traced again on `a06f78be` — `run_started` now flushes (good), delta 1 flushes, then deltas 2..255 sit in `pendingText` with no bound but `text_end`, a message-id change, 16 KiB, or this counter. `tool_start`/`post`/`update` do flush, so a tool-using agent still paints; a plain paced answer does not.

A wall-clock bound and replay stability aren't actually in tension here — see the PR comment. The retry hazard comes from the model's output diverging, not from scheduler jitter, so deterministic boundaries never protected the redelivery case they were adopted for.

packages/channels-intelligence/src/intelligence-adapter.ts
```diff
@@ -609,24 +651,53 @@ export class IntelligenceAdapter implements PlatformAdapter {
...
 
+  /**
+   * Serializes every durable render request for one turn/slot lane.
+   *
+   * Renderer batches and discrete post/update/file/delete effects share the
+   * same sequence allocator, so they must also share one transport queue. The
+   * stored tail absorbs a failed operation only for queue continuity; callers
+   * still receive the original rejection.
+   */
+  private enqueueRenderLane<T>(
```
BenTaylorDev, line 666: This is the right shape for the lane fix, and sharing the tail between the renderer and discrete effects resolves the race cleanly.

One ordering consequence worth a test rather than a change: text frames now get their `seq` at *flush* time (`flushPendingText` -> `addFrame`), while `pushRenderFrame` allocates its `seq` synchronously at call time. So a discrete `thread.post` issued while text is buffered takes a **lower** seq than text that was emitted before it, and the Outbox applies frames in seq order.

The common interleaving is safe because `tool_start` flushes pending text before the tool body runs. What isn't obviously covered is a `thread.post`/`update` from outside a tool during an active text stream. On `main` this couldn't happen, since every delta got its seq on arrival.

A test with a buffered text stream plus a concurrent discrete post, asserting provider-visible order, would settle it either way.

packages/channels-intelligence/src/intelligence-adapter.ts
MikeRyanDev: Fixed in `dbf4a43b89`. The first text still flushes at once, and a fake-clock test pins the buffered paced tail at 249/250 ms. Exact replay remains fail-closed under the current v1 contract; this PR does not add a new high-water resume protocol.

packages/channels-intelligence/src/intelligence-adapter.ts
MikeRyanDev: Fixed in `dbf4a43b89`. An external discrete push now flushes any active renderer text and batch before it allocates its sequence. The regression test pins `run_started`, first text, buffered second text, concurrent post, then finalize in that order.

packages/channels-intelligence/src/realtime-gateway.ts
MikeRyanDev: Fixed in `dbf4a43b89`. `extractReasonCode` now prioritizes the gateway payload's top-level `code` over its generic `reason`. A fake Phoenix push test sends `{ reason: "app_api_error", status: 409, code: "CHANNEL_RENDER_SEQUENCE_GAP" }` and asserts that the structured code reaches retry classification.

packages/channels-intelligence/src/intelligence-adapter.ts
```diff
@@ -939,43 +1023,199 @@ export class IntelligenceAdapter implements PlatformAdapter {
...
+      pendingText = undefined;
+      pendingTextDeltaCount = 0;
+    };
+
+    const flushPendingTextBatch = (): void => {
+      flushPendingText();
+      flushBatch();
+    };
+
+    const scheduleTailFlush = (): void => {
```
BenTaylorDev, line 1147: Confirming this is the right shape, since it's easy to get wrong: the `tailFlushTimer !== undefined` guard means the timer is **not** re-armed per delta, so this bounds worst-case time-to-visible-output at 250 ms rather than debouncing. A debounce would have deferred indefinitely under continuous streaming and still passed a burst-shaped test.

Traced the interactions and they're clean — the 256-delta and 16 KiB paths both route through `flushPendingTextBatch`, which clears the timer before flushing; a timer that fires against an empty buffer no-ops in `flushBatch`; and a timer firing while a batch is in flight just chains onto the lane tail.

packages/channels-intelligence/src/intelligence-adapter.ts
```diff
@@ -425,6 +469,7 @@ export class IntelligenceAdapter implements PlatformAdapter {
       // chain drained inside dispatchTo) so the Map can't grow unbounded over a
       // long-running Channel Bot. A redelivery re-seeds it at the top.
       this.renderState.delete(env.turnId);
+      this.renderLaneTails.delete(env.turnId);
```
BenTaylorDev, line 472: Minor hygiene, not a blocker: `renderLaneFlushers` isn't cleaned up here alongside its two siblings.

Drain's `finally` handles the normal path, but this block is the designated per-turn cleanup — the comment just above it exists specifically because these maps would otherwise grow unbounded on a long-running Channel Bot. Any path that completes a turn without reaching drain's `finally` leaks an entry, and each entry holds a closure over the whole renderer scope, so it retains more than the sibling maps do.

One line here (`this.renderLaneFlushers.delete(env.turnId);`) makes the three consistent and removes the need to reason about whether drain always runs.

BenTaylorDev (commented):
Re-reviewed `dbf4a43b`. Everything I raised is fixed, and the fixes are the right ones rather than the minimum ones.

- **`extractReasonCode`** now prefers `coded.code` over `coded.reason`, so a permanent 409 is finally classified correctly over the realtime gateway and not just over HTTP. Exactly the one-line change, with a gateway-shaped test alongside it.
- **`renderLaneFlushers`** solves the seq-ordering caveat I flagged on `enqueueRenderLane` — `pushRenderFrame` now drains buffered renderer text *before* claiming its own seq, so a discrete `thread.post` can no longer land ahead of text that was emitted earlier. The identity check before deleting in drain's `finally` is the right guard against clobbering a newer registration.
- **The tail flush is back**, and the semantics are correct: `scheduleTailFlush` is a no-op while a timer is already armed, so this is a max-latency bound (output at least every 250 ms during a stream), not a debounce. A debounce would have starved during continuous streaming and looked fine in a burst test. Cleared on every flush and in drain's `finally`, so no dangling handle.

I checked the obvious ways to break the timer — firing mid-flight (chains onto the lane tail, serialized), firing against an empty buffer (`flushBatch` early-returns), and the interaction with the 256/16 KiB caps (both call the same `flushPendingTextBatch`, which clears the timer first). It holds up.

**One thing to settle before merge, and it's a conversation rather than a defect.**

This restores exactly the wall-clock timing dependence that the round-one blocker objected to. The PR has now gone timer -> deterministic 256-cap -> timer, and each position is a reasonable answer to a different question. Worth landing on one story explicitly so this doesn't oscillate again.

My read, for what it's worth: the timer is right and the determinism argument was never load-bearing. `turnId` is stable across delivery attempts on the service side and `accepted_high_water` persists, so a retry re-runs the model, the output diverges, and the first batch fails the sequence check no matter how deterministic the boundaries were. Deterministic batching only ever protected a byte-identical replay, which a real retry is not.

That leaves the actual resume question open on both PRs: `CHANNEL_RENDER_SEQUENCE_GAP` is still terminal, so a turn that dies after partial render still nacks to dead-letter. That behaviour matches `main` (via `CHANNEL_RENDER_FRAME_CONFLICT`), so it isn't a regression and I don't think it blocks this PR — but it's the one real gap the batch protocol could have closed and didn't. A follow-up where the server returns its high-water on a gap and the SDK resumes from it would finish the job.

Nothing else on this head. The SDK half looks merge-ready to me modulo the paired service change and the launch gates.

packages/channels-intelligence/src/intelligence-adapter.ts
MikeRyanDev: Confirmed. The 250 ms timer is a fixed tail bound, not a debounce; the size and count flushes clear it before sending, and queued sends remain on the shared lane.

packages/channels-intelligence/src/intelligence-adapter.ts
MikeRyanDev: Fixed in `4c0577f996`. Dispatch cleanup now removes `renderLaneFlushers` beside the sequence and lane-tail maps. The new test first reproduced the retained entry, then pins cleanup after dispatch.

BenTaylorDev: Fixed in `4c0577f9` — the delete sits in the same unconditional `finally` as its two siblings, so it covers the nack path as well as the success path, and the test asserts the entry is gone after dispatch. Thread resolved.

That was the last open item I had on either PR. Only Tyler's `POOL_SIZE` naming point remains outstanding across the pair.

packages/channels-intelligence/src/intelligence-adapter.ts
BenTaylorDev: Filed the resume half of this as **[OSS-660](https://linear.app/copilotkit/issue/OSS-660/managed-channels-resume-a-partially-rendered-turn-instead-of-dead)** (OSS, Cycle 57) so it does not get buried when this merges.

To be clear about scope: point 1 of this thread (the string-matching classifier) is fixed, and so is the gateway `extractReasonCode` follow-up. What OSS-660 tracks is only the remaining half — that `CHANNEL_RENDER_SEQUENCE_GAP` is terminal, so a turn that dies after partial render nacks its way to `dead_lettered` on every retry.

That behaviour matches `main` (which fails the same way via `CHANNEL_RENDER_FRAME_CONFLICT`), so it is not a regression from this PR and I do not think it blocks. The ticket also records why the deterministic-boundary approach was not the answer, so nobody re-litigates trading the tail flush away for it.

PR #6222, opened 2026-07-29
feat(showcase): add runnable Vue documentation

enekesabel:
## What this adds

Vue users should be able to follow the same documentation journeys as React users, with examples that match the Vue package they can install today.

This PR adds a runnable Vue chat application to Showcase and makes the website the home for Vue guides and API reference. It follows the existing React and Angular setup rather than introducing a separate documentation system.

- **Showcase:** one Vue application and build artifact, integrated with the existing backend catalog, hosting, and validation. Only `agentic-chat` is enabled; unsupported combinations stay unavailable.
- **Guides:** React-aligned Vue examples, shared explanations where the behavior is framework-neutral, and explicit notes where Vue currently differs.
- **Reference and discovery:** updated Vue API pages, navigation derived from React's structure, and Vue-aware links, search, sitemap, and Markdown/LLM output.

This does not add Vue runtime features, resolve existing package type errors, define A2UI or v1 compatibility, or change shared server-side authorization examples.

## How to review

The five commits separate the Vue app, hosting/build integration, validation, API reference, and guides. The latest corrections are folded into those commits.

The main review questions are whether the Showcase integration follows the existing frontend pattern, and whether the Vue docs differ from React only where the current Vue implementation requires it.


BenTaylorDev:
Thanks for this, and apologies it sat unreviewed for so long. Two things conspired: CodeRabbit skips `feat(showcase)` titles by config, and because this is a fork branch no CI workflow has ever run against it — so nothing automated has looked at it either. I reviewed it against current `main` and ran the suites locally.

**What holds up.** The design follows the repo's conventions rather than routing around them. Generalizing `angular-integration-hosting.test.ts` → `secondary-frontend-integration-hosting.test.ts` and `angular-backend-overview.ts` → `secondary-frontend-backend-overview.ts` is the right instinct, and the shared probes are parameterized rather than forked — no per-framework branch smuggled into a probe, which is the thing our showcase iron rules care most about. `stage_vue()` mirrors `stage_angular()` including the `[ -L ]` guard and the `restore_symlinks` entry, and all 21 integrations carry a `public/vue` symlink matching `public/angular` one for one. The generated catalog comes out at exactly 21 Vue cells, agentic-chat only, which matches what the description claims. `showcase/harness` (45 tests) and `showcase/scripts` (80 files, 2575 tests) both pass on your branch.

**Four asks before this can land.**

1. Update the branch to current `main`. Three conflicts: `pnpm-lock.yaml`, `sitemap.test.ts` (both tests should survive), and `llm-text.ts` — that last one needs main's side. Your branch still carries the AG-UI block that main has since removed, and keeping it would resurrect the retired `/ag-ui` mirror.

2. `showcase/scripts/generate-search-index.ts:83` drops `guidanceTitle: "Docs status"` from the Vue entry, which removes `Vue: Docs status` from the search index and fails `channels-guide-search.test.ts:144-147` (that assertion predates your branch point). It reads like an oversight rather than a decision, since your own sitemap test still asserts `/vue/using-these-docs` is published — so the page exists but is no longer searchable. Restoring the property takes that file to 7/7.

3. `showcase/shell-docs/src/lib/curated-llm-pages.ts:259` curates `frontends/vue`, and this PR moves those docs to `/vue/*`. That file landed on main the day after your branch point, so neither side could have seen the other — but until the entry is repointed it breaks main's `keeps the curated policy ordered, unique, and on canonical routes` and your own `publishes Vue routes from the derived frontend navigation`.

4. A one-line comment on the Vue exclusion in `frontend-parity-gate.ts` saying why it is there. I tried removing it to check: the gate keys cells as `${frontend}/${integration}/${feature}` and strips only `react/` and `angular/`, so Vue rows become comparison ids of their own and every one blocks with "missing base React result", mislabeled as integration `vue`. Widening the regex to include `vue` instead makes the gate green while dropping Vue entirely — I flipped a Vue cell to `failed` and the gate still passed. Your exclusion is the right call; it just deserves a note so the next person doesn't try the one-line version. Generalizing the gate to N frontends is our follow-up, not yours.

Once those land we will run full CI and value-test on at least three real cells per `showcase/AGENTS.md`, which is the one piece of evidence nobody has yet: no one has watched a Vue cell render against a real backend.

enekesabel:
@BenTaylorDev thank you for the review!

Addressed all four asks and force-pushed after rebasing onto current main:
- resolved the rebase conflicts, keeping main's llm-text.ts behavior and sitemap coverage
- restored the Vue Docs status search entry
- repointed the curated LLM page to the canonical /vue route
- added the rationale for excluding Vue from the current React/Angular parity gate

NathanTarbert:
cc @BenTaylorDev ^^

BenTaylorDev (changes requested):
Thanks for turning this around quickly. Three of the four asks are properly
done, and the parity-gate comment in particular captures the real reason
including the trap I hit, so the next person will not try the one-line version.

Verified on your branch at 7430166a:

- The Vue `guidanceTitle: "Docs status"` entry is restored, and
  `channels-guide-search.test.ts` is green again.
- The curated LLM page is repointed from `frontends/vue` to `vue`, and
  `keeps the curated policy ordered, unique, and on canonical routes` now
  passes.
- `frontend-parity-gate.ts` carries the rationale.

Two new findings, both from running the branch rather than reading it.

**Blocking: `publishes Vue routes from the derived frontend navigation` cannot
pass as written.**

The test asserts seventeen `/vue/...` URLs appear in `/llms.txt`. That route
renders only what it is handed: `renderLlmsIndex(CURATED_LLM_PAGES, ...)`, and
the header comment says so deliberately — "a small explicit list instead of
walking that complete route table". `CURATED_LLM_PAGES` holds exactly one Vue
entry, just as it holds exactly one Angular entry. Measured on your branch, the
rendered index contains one line matching `(/vue` and one matching `(/angular`.

This is not a regression from the repoint. I checked f55b295b, the commit that
added the test: the curated list had one Vue entry there too, pointing at the
pre-rebase slug. The test has never passed at any commit on this branch.

The exhaustive index is the natural target, but note it does not publish all
seventeen either — `vue/using-these-docs`, `vue/voice`, `vue/inspector` and
`vue/threads` are absent from it today, while `vue`, `vue/human-in-the-loop`
and `vue/prebuilt-components/chat` are present. So please check what that route
actually emits before moving the assertions across, rather than assuming the
full set is there.

**`bundle-vue-source-content.ts` only writes to one app.**

The Angular bundler writes its JSON to both `shell/src/data/` and
`shell-docs/src/data/`. The Vue bundler writes only to `shell/src/data/`, so
`shell-docs` never receives a Vue content file. Nothing in `shell-docs` imports
it yet, so this is latent rather than breaking, but the asymmetry looks
unintentional and will bite whoever wires it up.

**Rebase, again, and sorry.**

Your Sep 15 rebase was correct. Main has since moved 290 commits and there is
one fresh conflict, in
`showcase/shell-docs/src/app/[framework]/[[...slug]]/page.tsx`. Nothing you did
wrong; this branch is just landing into a fast-moving tree.

Once the branch is green I will run the value-test on three real Vue cells per
`showcase/AGENTS.md`. That is still the one piece of evidence nobody has: no one
has watched a Vue cell render against a real backend.

enekesabel:
Thanks again for the review, @BenTaylorDev!
 I rebased onto current main and force-pushed the corrected branch. I also fixed the /llms.txt coverage, documented and tested the shell-only Vue source bundle, and addressed the follow-up issues found in the docs examples. The PR is ready for another look when you have time.

BenTaylorDev:
Thanks, both findings from my last pass are resolved, and I verified each on 0b08bd6f: the curated-index test now asserts the one canonical Vue entry, and the shell-only Vue bundle is documented and tested.

One new blocker, and it comes from main, not from you. On Sep 23, a09d853aa7 removed the threads-self-managed page and its shared snippet. The branch now conflicts on `snippets/shared/threads/self-managed.mdx`. Resolving by accepting the deletion still leaves `frontends/vue/threads-self-managed.mdx` importing the removed snippet, and two cases in `vue-frontends-route-pages.test.ts` fail.

Main already redirects `/vue/threads-self-managed` to `/vue/threads` (`next.config.ts`, the `NON_REACT_DOCS_FRONTENDS` block). So please delete the Vue page, your Vue edits to the snippet, and those two test cases. You do not need a replacement page.

With that merge applied locally, the scripts, harness and shell suites and the shell-docs typecheck are all green. The only new shell-docs failures are those two. After you push, I will approve the CI runs and do the Vue cell value-test.

PR #6223, opened 2026-07-29
docs(showcase): GOTCHAS — mcp-apps probe hard-asserts a live-only MCP iframe

lukasmoschitz:
Documents a fleet-wide false-negative class surfaced while watching staging after OSS-582 merged (see OSS-658).

The mcp-apps D5/D6 probe hard-asserts a live Excalidraw MCP `iframe[sandbox]` within a 15s poll, but no aimock fixture emits the `create_view` tool call for the probe prompt. The iframe can therefore only appear via the real MCP round-trip (regularly >90s and flaky on Railway). The integration's own Playwright test is `test.skip`-ped for exactly this reason; the probe has no such skip. Net: the cell can sit red (`achievedDepth 4/6, d6Effective=null`) while the demo is actually fine — this hits langgraph-python identically, so it is a shared probe/fixture limitation, not an integration defect.

Doc-only. The actionable fix (a `create_view` turn-1 fixture, or a probe skip/longer budget) is tracked in OSS-658.

### DIFF
--- a/showcase/GOTCHAS.md
+++ b/showcase/GOTCHAS.md
@@ -125,6 +125,8 @@ What we learned from getting all 18 integrations to D5 green. Many of these are 
 
 **PDF turn is fragile.** Two-turn multimodal probe: if turn 2's message doesn't match the PDF fixture, the image fixture matches instead. The PDF fixture must be the most specific match.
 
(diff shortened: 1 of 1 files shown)


BenTaylorDev:
Closing — superseded, and the premise doesn't hold against current `main`.

The note's central claim is that no aimock fixture emits `create_view` for the probe prompt. One does, and has since 2026-05-10:

- `showcase/harness/fixtures/d5/mcp-apps.json` — `{userMessage: "Open Excalidraw and sketch a system diagram", toolName: "create_view"}`. That's the probe's own declared `fixtureFile` (`registerD5Script({ fixtureFile: "mcp-apps.json" })` in `d5-mcp-apps.ts`), and it's present at this branch's HEAD.
- Every D6 slug dir carries the same fixture, in `showcase/aimock/d6/<slug>/tool-rendering-reasoning-chain.json` (and in `mcp-apps.json` for `ag2` / `spring-ai`). Its `_comment` reads: *"D5 mcp-apps probe — verbatim probe prompt drives a real `create_view` MCP tool call so the runtime's MCP Apps middleware fetches the UI resource and mounts the iframe."*

Easy trap to hit: aimock loads the whole per-slug directory and matches on `match` criteria — `fixtureFile` on the registry entry is telemetry, not loader scope.

The fleet-wide red this documents (`achievedDepth 4/6`, first failure 2026-07-28 23:03Z) had a different cause: `d70d48a561` gave `completeOnMount` a `mcp-app-iframe` testid only Angular declared, so React/Vue turns timed out at 30s with `surface-missing` and never reached `assertIframePresent` — the 15s poll never ran. `9d8916a9d9` fixed both sides a few hours after this PR opened and landed the corresponding note, now `showcase/GOTCHAS.md:17`. Its staging verification: *"the old gate matched 0 elements and the cascade matched 1 (the sandboxed iframe was there all along)."* `GOTCHAS.md:54` covers the live-vs-aimock nuance for this cell.

One piece worth keeping: the `create_view` fixture carries `toolName` in its match criteria, so it only fires when the runtime advertises the tool — a real dependency on the Excalidraw MCP server being reachable for tool *listing*. That's narrower than what this note claims and isn't documented anywhere yet; worth landing as its own entry.

Following up on OSS-658 separately — its section 1 fix suggestion asks for a fixture that already exists; section 2 (fastapi `shared-state-read`) is unaffected and still open.

PR #6236, opened 2026-07-29
feat(channels): route delivery through live sessions

MikeRyanDev:
## What changes

This replaces the managed Channels adapter and transport stack with live sessions over Realtime Gateway.

- Channels use the standard AgentRunner and canonical AG-UI history for each turn.
- One admitted delivery runs one prompt; multiple agent calls run in order; concurrent calls fail with a bounded protocol error.
- Slack and Teams reuse their native renderers to emit destination-free provider effects.
- Files, rich controls, interaction handlers, and provider cursors cross the live-session protocol with bounded payloads.
- The old claim mapping, HTTP fallback, render batches, listener election, in-memory transport, and legacy adapter code are removed.
- The public `@copilotkit/channels` umbrella remains limited to public provider adapters; the managed launcher stays in `@copilotkit/channels-intelligence`.

Companion service PR and kind proof: https://github.com/CopilotKit/Intelligence/pull/638

## Why

Managed delivery must use the same AgentRunner path as other Intelligence runs. SDK code emits provider-neutral effects; the trusted Gateway owns credentials, destinations, admission, retries, and terminal outcomes.

## Validation

- `pnpm nx run-many -t build,check-types,test -p @copilotkit/channels-core,@copilotkit/channels-intelligence,@copilotkit/channels-slack,@copilotkit/channels-teams,@copilotkit/channels,@copilotkit/runtime` — 31 tasks passed
- `pnpm nx test @copilotkit/channels-teams` — 89 passed
- `pnpm vitest run scripts/release/lib/channels-umbrella.test.ts` — 8 passed
- `pnpm verify:channels-umbrella` — packed snapshot, dependency resolution, and TSX consumer passed
- affected package pre-commit tests, publint, and API type checks passed
- changed-file Prettier and `git diff --check` passed

## Known unrelated check

`pnpm nx build demo` now compiles past the prior Channels telemetry dependency leak, then fails on the existing AG-UI 0.0.51 versus 0.0.57 private `_debug` type mismatch.


BenTaylorDev (changes requested):
Adversarial review, cross-verified against the companion service PR (CopilotKit/Intelligence#638). Two defects would cause silent data loss in production managed Channels; one I reproduced against the real runner.

**Drift:** 67 commits behind `main`, `MERGEABLE`, no conflicts. Every main-side commit touching the changed paths is an ancestor of this branch — no reintroduced or retired-tree debris. 46/46 checks green.

## Blocker 1 — a failed managed run is recorded as a successful delivery

`runCanonicalChannelAgent` (`channel-manager.ts`) subscribes `{ error: reject, complete: resolve }`. But `IntelligenceAgentRunner.run()` **never** calls `observer.error` — it emits `RUN_ERROR` as a next-value and completes (`runner/intelligence.ts:445-453`, `catchError(… return EMPTY)`; join failure/timeout at `:268-294`). That is its established, tested contract — three tests assert it and pass on this branch:

```
✓ pushes RUN_ERROR to the channel when agent throws          (intelligence-runner.test.ts:498)
✓ emits RUN_ERROR and completes when channel join fails      (:612)
✓ emits RUN_ERROR and completes when channel join times out  (:631)
```

So `result` stays at its initialiser, `runAgentLifecycle` keeps `status = "complete"`, pushes `channel.run.close.v1 status:"complete"`, and `LiveSessionTransport` pushes `channel.delivery.complete.v1`. The Gateway is a pure recorder — `delivery_session.ex:289` forwards the status verbatim; `:317` records `status: "complete", code: "delivery_completed"` and `{:stop, :normal, …}`, which its own docs call *"a terminal outcome, not a crash to replay."*

I replicated the bridge's exact shape against the real runner with a mocked Phoenix socket. Both cases resolve as success:

- **Tool handler throws** → resolves `{iterations: 0, interrupted: false}`. No rejection. `RUN_ERROR` reaches Intelligence on the *ingestion* channel, but the *delivery* is closed as complete.
- **Ingestion join fails** → resolves identically, and `execute` is **never invoked** — zero provider effects. The user gets silence, the platform records success, and no lease recovery fires.

The second case is the worse one: `failed_before_output` (`delivery_session.ex:341-352`), the status built for exactly this, becomes unreachable. The `catch { status = "failed" }` branch and `session.fail()` are effectively dead for agent failures.

## Blocker 2 — all provider pacing removed on both sides; one Slack 429 kills the delivery

`live-session-adapter.ts:401` sets `nativeStreaming.minIntervalMs: 0`, and the new `native-stream.ts:187-190` early-return bypasses the `flushTimer` coalescing gate entirely. The production default is 600ms (`native-stream.ts:89`).

Nothing downstream replaces it. Grepping `rate_limit|throttle|429|min_interval|retry_after` across `apps/realtime-gateway/lib` returns only the app-api *capacity* limiter; `provider_executor.ex` has no throttle, no backoff, no 429 handling. And a single failed provider call is terminal — `delivery_session.ex:519-529` → `fail_effect(…, "failed", "provider_call_failed")`, with no retry anywhere in the provider path. The PR body says the Gateway owns "retries"; the provider path has none.

This also reverses a decision `main` took days ago — `dbf4a43b89` "bound paced render latency", `f8d700c523` "batch managed render output" — and the deleted `render-batches.ts` carried 3-attempt retry with backoff (`waitForRenderBatchRetry`, `isPermanentRenderBatchError`).

## Major 3 — the load-bearing mechanism is untested; tests +490 / −5126

No test file exists for `live-session-adapter.ts` (604 LOC), `live-session-transport.ts` (321), or `live-session-files.ts` (130). The one substantial new test stubs out the thing under test:

```ts
runCanonical: async (args) => args.execute({})   // live-session-launcher.test.ts:177, :248
```

So `ChannelOuterAgent` + `AgentRunner` + the ingestion bridge — *"Channels use the standard AgentRunner and canonical AG-UI history"*, the premise of the PR — is never exercised. That is why Blocker 1 is green. `channel-manager.test.ts` only asserts `runCanonical: expect.any(Function)`.

Knock-on: **every** `NativeMessageStream` test passes `minIntervalMs: 0`, so after the `scheduleFlush` change the 600ms path that direct Slack adapters actually use has zero coverage.

## Major 4 — undeclared breaking removal of published API, against a hard cutover

`npm view` confirms `@copilotkit/channels@0.4.0` ships `./intelligence` today. This PR deletes that subpath, `intelligenceAdapter`, and `IntelligenceAdapterOptions`; `@copilotkit/channels-intelligence@0.4.0` loses ~30 named exports. The body describes this as the umbrella "remain[ing] limited to public provider adapters", which reads as a no-op rather than a removal.

Paired with the service PR's *"hard-cutover current migration… No compatibility path remains"*, any self-hosted runtime pinned to the currently published version breaks when the migration lands. Worth an explicit breaking-change note and a release-sequencing plan in the body.

## Medium

- **`closeRun` doesn't await the effect tail.** `complete()`/`fail()` await `this.tail`; `closeRun` (`live-session-transport.ts:160`) does not. The PR's own test encodes the result: `run.open, effect(stream.start), run.close, effect(stream.append), effect(stream.stop), delivery.complete` (`live-session-launcher.test.ts:212-219`) — the assistant text ships *after* the run is declared complete. If that append then fails, `close_run`'s `active_call_id` guard prevents re-attribution.
- **Declared-but-dead protocol surface.** `receivedThrough`/`appliedThrough` are returned by `effect()` and discarded by every caller — no gap detection, resume, or backpressure, so "provider cursors cross the live-session protocol" is only half-built on the SDK side. `runnerTokenExpiresAt` is never read, and Phoenix auto-reconnect reuses the same `authToken`, so a reconnect after expiry re-auths with a dead token. `slack.image.create`/`teams.image.create` are in the union *and* implemented by the Gateway (`hydrate_file_effect`) but emitted by nothing — managed images are unreachable. `initialText` is never emitted. All four `*Digest` fields are format-checked only (`delivery_session.ex:740` `sha256?`), never recomputed, so they carry no integrity value.
- **Low:** `getMessages` JSON-stringifies non-string content (the deleted `ir-to-text.ts` handled this); `decodeInteraction` is an unchecked cast; `lookupUser` is hardcoded to `undefined`; `mergeAgentSubscribers` would clobber `messages` if both sides returned it (latent only — the runner's subscriber returns void today).

## Checked and cleared — not worth chasing

- **The security model holds.** Destination and credentials enter only from server-side `context` in `provider_executor.ex`; the effect never carries them. The shallow top-level `TRUSTED_FIELD_NAMES` check is defence-in-depth, not the control.
- **No version skew** from dropping `channels-intelligence` out of `ADAPTERS`/`FAMILY` — `release.config.json` still lockstep-publishes it in the `channels` scope.
- The pacing/batching removals are deliberate replacement, not accidental drift-deletion (those commits are all branch ancestors).
- `implicitInboundConsumed` is safe — `makeThread` runs fresh per ingress event.
- `mergeAgentSubscribers`' Proxy is compatible with AG-UI's dispatch (`o.onX?.call(o, …)`, no key enumeration, all call sites awaited).
- No docs or examples import the removed `intelligenceAdapter`.

## The single most important thing to verify

**That `runCanonicalChannelAgent` distinguishes a completed run from a failed one.** Everything else here is recoverable; this one makes the platform confidently report success on silence. The fix is to inspect the emitted event stream for `RUN_ERROR` (and require that `execute` actually ran) rather than treating observable `complete` as success — `error: reject` is dead code against this runner.

## Coordination note

This PR deletes 7 of the 8 files touched by the unlanded `fix/channels-render-lane-per-attempt` branch. The lane-keying bug class is structurally gone in the new design (a new `deliveryId` means a new session with a fresh `nextSeq`), but the pacing fix in `77319ca320` is precisely Blocker 2 — deleted with nothing replacing it.

maxkorp:
Reviewed together with [Intelligence#638](https://github.com/CopilotKit/Intelligence/pull/638). I found four issues that should be addressed before merging:

1. **Agent-run failures are reported to the manager as success.** The Intelligence runner catches an agent error, emits it as a normal observable value, and completes; the manager bridge only rejects on the observable's error callback and resolves on completion. A failed run can therefore follow the success path, potentially acknowledging/committing a failed session.
   https://github.com/CopilotKit/CopilotKit/blob/5063aa75cfb190857c78b1e719d2964c6c90c364/packages/runtime/src/v2/runtime/runner/intelligence.ts#L445-L453
   https://github.com/CopilotKit/CopilotKit/blob/5063aa75cfb190857c78b1e719d2964c6c90c364/packages/runtime/src/v2/runtime/core/channel-manager.ts#L360-L391

2. **Slack publishing has no default pacing or retry.** `minIntervalMs` defaults to zero, and the paired gateway treats provider errors as terminal. Bursty token/status updates can hit Slack's rate limits and permanently lose deliveries. Please introduce safe pacing and/or a bounded retry policy that honors `Retry-After`.
   https://github.com/CopilotKit/CopilotKit/blob/5063aa75cfb190857c78b1e719d2964c6c90c364/packages/channels-intelligence/src/live-session-adapter.ts#L400-L417

3. **Interaction/reaction handlers cannot update the source Slack message.** Intelligence supplies an `interaction_ref` containing the original provider message timestamp, but `asLiveRef` only accepts refs with `sessionId`, `responseId`, and `adapter`. The source ref is consequently discarded, and `updateMessage` throws instead of updating the original message as the contract promises. Please preserve/translate the provider source ref into an update-capable live ref.
   https://github.com/CopilotKit/CopilotKit/blob/5063aa75cfb190857c78b1e719d2964c6c90c364/packages/channels-intelligence/src/live-session-adapter.ts#L279-L281
   https://github.com/CopilotKit/CopilotKit/blob/5063aa75cfb190857c78b1e719d2964c6c90c364/packages/channels-intelligence/src/live-session-adapter.ts#L553-L558
   Contract/ingress context: https://github.com/CopilotKit/Intelligence/blob/0100603a38753e205ec240d2c89f68505683d0c4/libs/app-api-contracts/src/channels.ts#L908-L929

4. **Delivery-topic join failures can become unhandled promise rejections.** `start()` launches `handleDelivery(...).finally(...)` without observing the returned promise, while the topic join happens before `handleDelivery`'s `try/catch`. A rejected join escapes as an unhandled rejection rather than being converted into a controlled delivery/session failure.
   https://github.com/CopilotKit/CopilotKit/blob/5063aa75cfb190857c78b1e719d2964c6c90c364/packages/channels-intelligence/src/live-session-transport.ts#L241-L254
   https://github.com/CopilotKit/CopilotKit/blob/5063aa75cfb190857c78b1e719d2964c6c90c364/packages/channels-intelligence/src/live-session-transport.ts#L262-L288

BenTaylorDev (approved):

PR #6238, opened 2026-07-29
fix(channels): key render lanes per delivery attempt; pace tail flush to provider throttle

BenTaylorDev:
## What

Two independent changes to the managed Channels render path, in separate commits so they can be judged (and reverted) separately.

### 1. `fix(channels)`: key render lanes per delivery attempt

The durable egress lane is keyed `${turnId}:${slot}`, and `turnId` is a pure function of `deliveryId` (`toPublicTurnId`), so **every attempt of a delivery shares one lane and one `accepted_high_water`**. A redelivery therefore has to reproduce byte-identical batch boundaries to make forward progress: a re-cut batch isn't found by `batchId`, falls through to the `startSeq === accepted_high_water + 1` check, and the turn dies on `CHANNEL_RENDER_SEQUENCE_GAP`.

Replay only reproduces identical frames when the agent is deterministic. Digests cover full frame content, so any variation — temperature, a timestamp, a tool result that differs between attempts — re-cuts the boundaries anyway. The shared lane was protecting a narrow case while breaking the common one.

This derives the lane from the delivery's **lease token**, which app-api mints fresh on every claim and holds for that lease's life. It's the one value the SDK already receives that distinguishes attempt N from N+1 (`deliveryId` / `turnId` / `eventId` are all stable across attempts).

**No contract change is needed.** `slot` is SDK-supplied, and `main-<sha256:12>` already satisfies the platform's `channelRenderSlotSchema` (`/^[A-Za-z0-9_-]+$/`, ≤64) and the `channelRenderFrameKeySchema` `turnId:slot:seq` bound. The token is **hashed, not sliced**: `slot` is persisted on every render acceptance and is operator-readable, while the raw token is the capability accept/ack/fail are fenced against.

Also fixes a latent desync this surfaces: `postRenderFrame` minted refs as `${turnId}:main:${seq}` hardcoded while the batch identity set `slot` separately, so the Connector Outbox's `${turnId}:${slot}:${seq}` lookup would silently stop resolving the moment the lane varied. Both now read one resolved slot.

> [!IMPORTANT]
> **Trade-off for review.** A redelivery after partial acceptance now re-renders into a fresh lane — duplicate provider output, including a second `chat.startStream` — where it previously failed closed on the sequence gap. This trades *silently dropped* redeliveries for *duplicated* ones. That's a product decision, not a mechanical one, and it's the main thing I'd like a second opinion on. If duplication is unacceptable, the alternative is to keep the shared lane and accept that redelivery only works for deterministic agents.

### 2. `perf(channels)`: pace the render tail flush to the tightest provider throttle

`RENDER_TEXT_TAIL_FLUSH_MS` was 250ms while every managed provider throttles its own edits more slowly:

| Provider call | Throttle | Source |
| --- | --- | --- |
| Slack `chat.appendStream` | **600ms** | `channels-slack/src/native-stream.ts` |
| Teams `updateActivity` | 700ms | `channels-teams/src/message-stream.ts` |
| Slack `chat.update` | 800ms | `channels-slack/src/message-stream.ts` |

One managed lane serves whichever providers a Channel has attached, so a 250ms cadence pushed ~2.4x the batches the fastest consumer can use. Each of those pays a full durable-acceptance transaction (delivery `SELECT ... FOR UPDATE`, lease touch, acceptance insert, high-water update) and is then coalesced away by the provider throttle regardless.

600ms is the **tightest** provider throttle, so the fastest consumer stays fully fed — no user-visible cadence change — while accepted batches per streaming turn drop by roughly half. Slack `chat.update` and Teams coalesce marginally more than before, which is what their own throttles would have done anyway.

Pinned by a test so a future retune is deliberate rather than a silent change in accepted-batch volume, and the window test now drives off the constant instead of a hardcoded `249`/`250`.

## Testing

Full package suite and typecheck, on this branch off current `main`:

```
$ npx vitest run
 Test Files  14 passed (14)
      Tests  242 passed (242)

$ npx tsc --noEmit -p tsconfig.json
(no output)
```

**Both new test groups were verified to actually discriminate**, not just pass:

- Neutered `renderLaneSlot` to always return the base slot → **5 of 7** lane tests fail (`falls back to the base lane…`, `derives a platform-legal lane…`, `is stable for one token and distinct across attempts`, `streams a redelivery into its own lane`, `keeps the base lane when the source has no lease token`); restored → 7 pass.
- The tail-flush window test drives off `RENDER_TEXT_TAIL_FLUSH_MS - 1` then `+1`, so it asserts the boundary rather than a magic number, and the separate pin test fails on any retune.

New coverage in `render-lane-slot.test.ts`: base-slot fallback, platform slot-charset legality, per-token stability and cross-attempt distinctness, that the raw token is never embedded, that a redelivery streams into its own lane with its own `startSeq: 0`, and that minted refs carry the same lane their frame was pushed under.

## Notes for reviewers

- No changeset — release goes through the workflow scopes.
- The lane change adds an optional `leaseTokenFor` to `DeliverySource`. `HttpDeliverySource` already had it; implemented on `RealtimeGatewayTransport` (the production managed path) and `InMemoryDeliverySource`. Being optional means a source without leases keeps today's single-lane behavior.
- A companion Intelligence-side change is queued separately (it's a different repo): the render-batch accept route calls `getChannelStatusReadModel` → `getChannelHealth` per accepted batch, ~5 queries, purely to publish a live dashboard event — while `getChannelOperationalHistory` already rebuilds `render_accepted` from `cpki.channel_render_acceptances`, so the durable timeline doesn't depend on that publish.


BenTaylorDev:
Obviated by v3

PR #6239, opened 2026-07-29
chore(README): Revise supported platforms in README

NathanTarbert:
Updated supported platforms and added new messaging services.

## What does this PR do?

## Related PRs and Issues

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,11 +104,13 @@ One agent backend. Every frontend.
 | 🅰️ Angular                                  | ✅ Supported | [Source Code & Quickstart](https://github.com/CopilotKit/CopilotKit/tree/main/packages/angular) |
 | 💚 Vue                                      | ✅ Supported | [Source Code - Quickstart coming soon](https://github.com/CopilotKit/CopilotKit/tree/main/packages/vue)     |
 | 📱 React Native                             | ✅ Supported | [Quickstart](https://docs.copilotkit.ai/react-native)                                                       |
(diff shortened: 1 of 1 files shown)


BenTaylorDev (commented):
Platform claims look intentional on the GTM side, so no objection there. Two things inside the edited section before this goes in:

**1. Dropped sentence looks accidental**

The diff removes this line with no replacement:

> Your agent logic stays the same — AG-UI handles the wire protocol, CopilotKit handles the UI layer for each framework.

That was the table's closing prose and the only AG-UI framing in the section, and removing it leaves a double blank line before the next heading. Was that deliberate? If not, worth restoring.

**2. The body under the new heading didn't move with it**

The heading now reads *Slack & Microsoft Teams (Discord, Google Chat, Telegram, iMessage, SMS, WhatsApp)*, but the prose a few lines down is unchanged — still just the two bullets:

- **Slack** – Agents as first-class Slack apps…
- **Microsoft Teams** – Bring agentic workflows to the enterprise…

and it still closes with **🔒 Early access: We're onboarding teams now.** So a reader sees `✅ Supported` in the table and an early-access gate in the same section. Adding bullets for the new channels (or adjusting that closing block) would finish the edit.

**Minor / non-blocking**

- The widened cell breaks the table's column alignment — renders fine, just noting it.
- The row drops the `go.copilotkit.ai/beyond-the-web-form` link. It still exists in the body below, so nothing is lost, but flagging in case the table placement was doing work for lead capture.

Otherwise this is clean: merges into current `main` without conflict, `docs.copilotkit.ai/slack` resolves, and the image uses the same asset host as the rest of the README.

tylerslaton (approved):

PR #6240, opened 2026-07-29
fix(channels): treat a revoked delivery lease as fenced and make gateway outages diagnosable (refs OSS-670)

BenTaylorDev:
Three fixes from the OSS-670 investigation, where a dev gateway restart landed mid-turn: the agent's answer was produced and then thrown away, the end user got a generic error, and the bot went silent for 19 minutes with nothing in its logs.

## A revoked lease is a fenced outcome, not a turn failure

`CHANNEL_DELIVERY_LEASE_INVALID` means another owner holds the delivery — app-api fences every intent on `leased_until > now()` AND the lease-token hash, so nothing we send can be accepted and redelivery is app-api's job.

- The guard lives in `IntelligenceAdapter.dispatch`, the site that actually runs. An earlier revision guarded only the realtime transport, which is unreachable in production because `dispatch` catches every turn error and nacks first — so a fenced delivery still put a doomed `channel.delivery.fail.v1` on the wire, and that second 409 is what surfaced as `nack after turn failure failed` followed by `no delivery state`.
- The code is no longer retried through the 3-attempt render-batch ladder.
- A `fail` intent that is itself fenced is logged and dropped rather than rejected.
- The predicate lives in `transports.ts` because both transports surface the same code (the HTTP path as a 409 body).

The adapter-dispatch test is a real regression test: reverting the guard makes it fail on `channel.delivery.fail.v1` being pushed.

### Abandoning is a third outcome, not a silent one

Swallowing the coded error means the `DeliverySource` observes a **successful** `onDelivery` — so neither its own revoked-lease branch nor `ack` runs, and those are the only two places the per-delivery lease record is dropped. Without more, every fence leaked a lease token, delivery scope and accepted-seq high-water map for a delivery this runtime can never speak for again (app-api may already have handed it to another runtime, so the id may never be seen here again).

`DeliverySource.abandon(deliveryId, reason)` is that third outcome alongside ack/nack: retire local state, send nothing. It is synchronous and wire-free by design, optional on the interface so existing implementations stay valid, and implemented on `RealtimeGatewayTransport` (`deliveries`), `HttpDeliverySource` (`leases`) and `InMemoryDeliverySource`. The transport's internal revoked-lease catch routes through it too, so there is one retirement path rather than two.

Both wiring tests fail with the `abandon` call reverted — the adapter test on an empty `abandoned[]`, and the composed adapter+transport test because its probe `nack` pushes a second, doomed `channel.delivery.fail.v1` off the retained state.

## A refused reconnect is diagnosed instead of retried in silence

`socket.onError` returned early once the socket had connected, so a refused *reconnect* recorded nothing and Phoenix retried forever with no output. The cause is now recorded, and when the transport exposes no OS code an HTTP probe of the endpoint runs once per outage.

The probe is unauthenticated, so it deliberately does **not** claim your key is revoked — its 401/403 would look identical with a perfectly good credential. It establishes reachability: 5xx reads as an unhealthy/restarting gateway, anything else as "reachable, handshake refused — verify the project API key".

## Honest status, loud while down

`gave_up` maps to status `error`, but Phoenix keeps retrying underneath and a successful rejoin restores `online` — the log now says so instead of reading as terminal. The drop cause travels with each state transition (it was previously `{ meta: undefined }`), and a "still down" line repeats while the session is out, including after `gave_up`. Channel status semantics and the "manager never re-activates on a drop" contract are unchanged.

## Not included

Delivery-lease renewal has no owner across a gateway restart — draining stops the channel process that renews it, so a long turn loses its lease to expiry regardless. That, and bounding an in-flight turn whose lease is already dead, are follow-ups on the ticket. Retiring the local record (above) stops the leak; it does not keep the lease alive.

Tests: `channels-intelligence` 13 files / 244 tests; `runtime` 127 files / 1799 tests; `check-types` clean on both.

### DIFF
--- a/packages/channels-intelligence/src/intelligence-adapter.ts
+++ b/packages/channels-intelligence/src/intelligence-adapter.ts
@@ -60,6 +60,11 @@ import {
 const PERMANENT_RENDER_BATCH_ERROR_CODES = new Set([
   "CHANNEL_RENDER_BATCH_CONFLICT",
   "CHANNEL_RENDER_SEQUENCE_GAP",
+  // A revoked/superseded lease can never accept another batch: app-api fences
+  // every intent on `leased_until > now() AND lease_token_hash`, so retrying
+  // only spends the backoff ladder against a lease that is already gone
+  // (OSS-670).
+  "CHANNEL_DELIVERY_LEASE_INVALID",
 ]);
 
 /** Return whether retrying the same render batch cannot succeed. */
--- a/packages/channels-intelligence/src/intelligence-adapter.ts
+++ b/packages/channels-intelligence/src/intelligence-adapter.ts
@@ -25,6 +25,7 @@ import type {
   RenderEventSink,
   AgentMessage,
 } from "./transports.js";
+import { isLeaseRevoked } from "./transports.js";
 import type {
   ChannelIngressEnvelope,
   EgressOp,
@@ -465,10 +466,21 @@ export class IntelligenceAdapter implements PlatformAdapter {
       await this.finalizeRenderedTurn(env);
       turnProcessed = true;
     } catch (err) {
-      await this.requireSource().nack(
-        env.deliveryId,
-        err instanceof Error ? err.message : String(err),
-      );
+      if (isLeaseRevoked(err)) {
+        // This delivery's lease was revoked mid-turn, so another owner holds it
+        // now and app-api owns redelivery. A nack here would send a `fail` the
+        // same fence rejects — the doomed intent whose 409 used to be reported
+        // as a turn failure, with a generic error reaching the end user
+        // (OSS-670). Stay quiet; this is not the user's problem.
+        this.opts.config?.log?.(
+          `intelligence delivery ${env.deliveryId} lease revoked mid-turn; abandoning (app-api owns redelivery)`,
+        );
+      } else {
+        await this.requireSource().nack(
+          env.deliveryId,
+          err instanceof Error ? err.message : String(err),
+        );
+      }
     } finally {
       // Drop the per-turn state once the turn is fully processed (renderer
       // chain drained inside dispatchTo) so the Map can't grow unbounded over a
--- a/packages/channels-intelligence/src/intelligence-adapter.ts
+++ b/packages/channels-intelligence/src/intelligence-adapter.ts
@@ -472,6 +472,19 @@ export class IntelligenceAdapter implements PlatformAdapter {
         // same fence rejects — the doomed intent whose 409 used to be reported
         // as a turn failure, with a generic error reaching the end user
         // (OSS-670). Stay quiet; this is not the user's problem.
+        //
+        // Retire the source's lease state explicitly. Swallowing the error here
+        // means the source sees a SUCCESSFUL onDelivery, so neither its own
+        // revoked-lease branch nor `ack` below runs — and those are the only
+        // other places the lease record is dropped. Without this the lease
+        // token, delivery scope and accepted-seq high-water marks are retained
+        // for a delivery we can never speak for again (app-api may already have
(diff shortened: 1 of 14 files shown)


packages/channels-intelligence/src/intelligence-adapter.ts
```diff
@@ -460,10 +466,21 @@ export class IntelligenceAdapter implements PlatformAdapter {
       await this.finalizeRenderedTurn(env);
       turnProcessed = true;
     } catch (err) {
-      await this.requireSource().nack(
-        env.deliveryId,
-        err instanceof Error ? err.message : String(err),
-      );
+      if (isLeaseRevoked(err)) {
```
MikeRyanDev, line 469: Clear the delivery-source lease state before returning here. This branch swallows the coded error, so RealtimeGatewayTransport.handleDeliveryAvailable() observes a successful onDelivery and never reaches its own isLeaseRevoked catch, the only branch that deletes this.deliveries. The HTTP source has the same ownership split: neither ack nor nack runs, so its leases entry remains. I reproduced this on the exact PR head: after a rejected render batch, the adapter sent no fail intent (correct) but the transport still had one retained delivery. If app-api hands that delivery to another runtime, this process may never see the ID again, so repeated fences retain lease tokens, scope, and accepted-state entries until shutdown. Add an explicit abandon/cleanup path and assert that source state is gone in the composed adapter+transport test.

MikeRyanDev (changes requested):
Requesting changes for one delivery-lifecycle leak. The fence handling stops doomed fail intents, but it must also retire the local lease state.

packages/channels-intelligence/src/intelligence-adapter.ts
BenTaylorDev: Confirmed and fixed in `1a72c404ba`.

The mechanism is exactly as you describe: `dispatch` swallows the coded error, so the source observes a successful `onDelivery` and neither its own revoked-lease branch nor `ack` runs — and those are the only two places the lease record is dropped. `HttpDeliverySource.leases` has the identical split.

Added `DeliverySource.abandon(deliveryId, reason)` — the third outcome alongside ack/nack: retire local state, send nothing. Optional on the interface so existing implementations stay valid; no other implementor exists in the repo. Implemented on `RealtimeGatewayTransport` (`deliveries`), `HttpDeliverySource` (`leases`), and `InMemoryDeliverySource` (records calls for assertions). The transport's internal revoked-lease catch now routes through it too, so there is one retirement path rather than two.

Tests, both wiring tests proven by reverting the `abandon` call:

- The composed adapter+transport test now probes with `nack("dlv_d3")` after the fence — it must push nothing and log `no delivery state`. Reverted, it pushes a second, doomed `channel.delivery.fail.v1`, which is your leak made visible on the wire.
- A new adapter-level test asserts `acked`/`nacked` empty and `abandoned == ["d9"]`. Reverted, `abandoned` is empty.
- A new `HttpDeliverySource` test asserts `abandon` clears both `scopeFor` and `leaseTokenFor`, posts nothing, and leaves a later `nack` a no-op.

`channels-intelligence`: 13 files / 244 tests, `check-types` clean.

To be explicit about what this does *not* cover, since it borders the follow-ups already listed in the description: this stops the state leak only. Keeping a lease alive across a gateway restart, and bounding an in-flight turn whose lease is already dead, are still unowned and remain on the ticket.

MikeRyanDev (approved):
The original delivery-state leak is fixed on `1a72c404`. The fenced branch now uses a wire-free `abandon` outcome, and both bundled production sources delete their retained lease/scope/accepted state. The composed adapter + realtime transport regression proves the state is gone by showing a later nack emits no fail intent. I reran the full `@copilotkit/channels-intelligence` suite (244 tests), check-types, and build on this exact head; all pass.

PR #6244, opened 2026-07-30
fix(channels-slack): roll long native-stream replies into continuation messages (refs OSS-685)

BenTaylorDev:
> [!IMPORTANT]
> **Re-grounded 2026-07-30.** This PR was opened against the old queued-egress architecture and originally referenced OSS-677. Intelligence#638 ("replace queued delivery with live sessions") has since landed, moving managed provider egress into the realtime gateway and deleting the egress-lease, outbox, and render-frame machinery entirely. OSS-677 was canceled as a result.
>
> **The SDK defect this PR fixes is unaffected and still present on `main`** — `native-stream.ts` is live, still owns Slack streaming cadence, and still has no continuation rollover. The work is now tracked as **OSS-685**.
>
> What *did* change is the blast radius (see "What the symptom looks like now") and the platform-side half, which no longer exists to fix.

## What

A managed Slack bot asked to write a novella posted **five truncated copies** of it. The proximate cause is in this file.

`NativeMessageStream` honored Slack's documented 12k-chars-per-`markdown_text`-call cap but never the *cumulative* cap on what one streamed message can hold. Once a reply crossed that ceiling, every further `chat.appendStream` was rejected `msg_too_long` — deterministically, for the rest of the run.

The assumption was stated explicitly in the file's own header, and it is wrong:

> A single streamed message holds the whole reply: Slack documents no cumulative per-message cap, only a **12k char limit per `markdown_text` call** … (no multi-message splitting — that was a `chat.update`-era workaround).

### Evidence (dev, 2026-07-29 23:24–23:27 UTC)

A ~23,279-char reply streamed 11,607 chars into one message, then failed every subsequent append. Egress op `019fb031-97eb-…`:

| field | value |
| --- | --- |
| `attempts` | 5 / 5 |
| `lease_generation` | 5 |
| `accepted_high_water` | 386 |
| `applied_high_water` | **-1** |
| `payload.posts` | **5 distinct Slack ts** |
| `status` | wedged in `sending` |

The dropped text is invisible by design (per-append failures are swallowed and logged), so the turn never reached `finalize`, never checkpointed, and the platform replayed it from frame 0 — a fresh `chat.startStream` and a fresh truncated copy per attempt, five times, then wedged. Worker logs show 3,512 `msg_too_long` rejections in 25 minutes; the retry burn also starved other projects' lanes.

> [!NOTE]
> That table describes the **pre-cutover** schema. `channel_egress_operations`, `attempts` / `max_attempts`, `lease_generation`, and the render-frame high-water columns were all dropped by Intelligence#638 (migration `000008.sql`). The evidence stands as the historical record of how the defect was found; the columns no longer exist to query.

### What the symptom looks like now

The root cause is identical — the cumulative cap is still unhandled — but the failure mode has changed:

- **Then:** per-append failures were swallowed, the turn never finalized, and the platform replayed from frame 0, producing N truncated copies and burning the attempt budget.
- **Now:** `f1c4b6ca79` ("keep managed Slack replies atomic") added a `strict` flag that the managed live-session path sets. In strict mode an `appendText` / `stopStream` failure re-throws instead of being swallowed, so `msg_too_long` surfaces as **one failed turn** rather than five truncated copies plus a wedge.

That is strictly better, and it removes the retry-storm half of the incident. It does not deliver the reply. The customer still loses a long answer — which is what this PR fixes.

The direct/self-hosted path (SDK holds the token, `strict` unset) still swallows the failure and silently truncates.

## How

`flushText` now rolls over. When a message fills and text remains: freeze the boundary on the last line break (else last space, so a word is never torn), close any markdown construct left open, `stopStream` it, and continue in a fresh message that re-opens that construct. This is the same shape as the legacy `ChunkedMessageStream`, whose splitting logic was removed as an unnecessary `chat.update`-era workaround.

Three decisions worth review:

- **Soft limit 11k UTF-8 bytes** (`messageByteLimit`, configurable; changed from chars after review). Slack's cumulative cap is undocumented; 11,607 chars were *observed accepted*, so 11k sits provably below the smallest known-good total with headroom for the closers a boundary appends. Crossing the cap is unrecoverable for the whole reply, not merely a truncated append, so the asymmetry favors staying under.
- **A continuation `startStream` failure no longer falls back to legacy.** The legacy sink is seeded with the entire accumulated buffer, so failing over mid-reply would re-post everything already streamed — trading one duplication bug for another. First-message failures still fall back, preserving "opting in can never break a bot"; a continuation failure just propagates and the next flush retries the boundary.
- **`renderContextCloser` is new rather than reusing `autoCloseOpenMarkdown`.** The latter inserts closers *before* trailing whitespace, which an append-only transport cannot express — `appendStream` sends deltas and Slack has no "un-append". The new helper is symmetric with the existing `renderContextOpener` and only ever appends.

Both append loops (`flushText` and the pre-chunk `flushTextInline`) now share one rollover-aware path instead of carrying duplicate cap logic, and `flushChunk` re-checks its target after a text flush, since a rollover retargets the stream.

### Interaction with `strict` mode

Worth a reviewer's eye now that both exist: a rollover must not be defeated by the strict re-throw firing before the continuation is attempted. The rollover path is entered on the *budget check*, not on a caught provider error, so it should run ahead of any throw — but this branch predates `strict` and the two have not been exercised together. Needs a test that runs the rollover with `strict: true`.

## Why the suite didn't catch it

The fake transport modelled no cumulative cap at all, so the failure was unreachable in tests. It models one now. The existing case asserting *"keeps a long reply in ONE message, chunking appends under the 12k per-call cap"* was asserting the bug, and is replaced by the rollover expectations plus a fits-in-one-message regression guard.

## Related

- **OSS-685** — the tracking ticket (successor to the canceled OSS-677). Also carries the open question of whether the gateway needs its own guard: `apps/realtime-gateway/lib/realtime_gateway/channels/provider_executor.ex` now makes the managed Slack call and has no `msg_too_long`, length, or splitting handling of its own.
- **The platform-side half is gone, not deferred.** The original plan was app-api work — classify `msg_too_long` terminal instead of burning 5 attempts, and stop the re-claim storm. Intelligence#638 deleted the attempt/lease machinery that produced both, so there is nothing left to fix there. `9a6ee861` and the OSS-677 branch fixes were merged into code that has since been removed.
- **#6238** (`fix(channels): key render lanes per delivery attempt…`) was **closed on 2026-07-30** — it was written against render lanes and per-delivery attempts, both retired by the cutover. The trade-off the two PRs used to share (that PR knowingly accepting duplicate provider output on redelivery) no longer has a referent, so nothing in this PR depends on it.
- `packages/channels-slack` is the only live site; `bot-slack` ships `dist` only.

## Testing

Package suite, build, and both typecheck projects, on this branch off `main` **as of the original run** — this branch predates the Intelligence#638 cutover and has not been re-based since:

```
$ pnpm build          # tsc -p tsconfig.json
(no output)

$ pnpm check-types    # tsconfig.json + tsconfig.check.json
(no output)

$ npx vitest run
 Test Files  23 passed (23)
      Tests  286 passed (286)
```

New coverage: rollover past the cumulative cap (no char lost or reordered, every message finalized, per-call cap still respected, `firstTs` still the first message), line-boundary splitting, code-fence re-opening across a boundary, resume-after-transient-append-failure, and the fits-in-one-message guard.

> [!WARNING]
> **Before merge:** re-base on current `main` and re-run. `native-stream.ts` gained `strict` mode (`f1c4b6ca79`) and other changes after this branch was cut, so the numbers above are stale and the rollover has not been tested against `strict: true`.

> [!NOTE]
> **Updated after review** (see the two review comments for full detail). The soft limit is now denominated in **UTF-8 bytes**, not chars: the incident reply was English, where the two are 1:1, so it could not tell us which unit Slack's ceiling uses — and under a byte ceiling a char budget never fires, silently truncating any CJK/Cyrillic reply exactly as before. Bytes are >= chars, so an 11k-byte budget is safe in either unit.
>
> Also fixed after review: `finish()` no longer drops the undelivered tail when a continuation `startStream` fails, and markdown tables are re-opened across a boundary.
>
> A manual >12k reply against a real workspace is still worth doing before ship — and it should be **mostly CJK or Cyrillic**, since that is where the byte and char budgets disagree most.


BenTaylorDev:
## Adversarial self-review — two defects found and fixed (`6d5fbf3e3f`)

I re-reviewed this assuming nothing, specifically trying to *falsify* the first commit rather than confirm it. Two defects, both reachable at the **default** 11k cap, both introduced by the rollover itself.

### 1. Unbounded message creation (high)

The continuation re-opener is synthetic text charged against the new message's budget. If the opener is itself larger than the cap, each fresh message fills with nothing but its own preamble and rolls over again — forever, never advancing `curPosted`.

Reachable because `detectOpenContext` treats everything up to the first newline after ``` as the fence "language". An agent emitting a >11k whitespace-free blob (minified JSON, base64, a long log line) right after a fence is enough.

Probed at the default cap:

```
[native-stream] appendText failed (ts=undefined, posted=11000/20003): Error: RUNAWAY: 41 startStream calls
```

`posted` frozen at 11,000; 41 `startStream` calls and still climbing, bounded only by the probe's own budget. **In production that is an unbounded stream of Slack messages — strictly worse than the truncation this PR set out to fix.** The opener is now dropped when it can't leave `MIN_MESSAGE_PROGRESS_CHARS` of room, so every message carries text and the loop always terminates.

### 2. Surrogate pairs split across boundaries (medium)

`breakPoint`'s hard-cut fallback triggers whenever a window has no newline or space — routine for CJK, emoji runs, and base64. Cutting at an arbitrary UTF-16 offset left a lone high surrogate ending one message and an orphaned low surrogate starting the next: one broken glyph per boundary.

The nasty part is why the existing tests missed it — **concatenating the messages still compares equal to the original**, so the "not one character lost or reordered" assertion passed straight through it. Only a per-message edge check sees it. Probe: 4 of 4 messages split.

Fixed via `avoidSurrogateSplit` at every boundary, including the per-call 12k slices, which carried the same latent exposure before rollover existed.

### Mechanism check (the load-bearing claim)

I also tried to kill the root-cause diagnosis itself. A competing explanation fits the same logs: not a *cumulative* per-message cap but a *per-call delta* cap smaller than the failing 8,428-char delta. I can't fully separate the two from the incident logs alone — every successful append was small, so both hypotheses survive.

What makes this moot for correctness: measured max append is **≤ 11,000 chars**, and max message total is 11,000 + closers. That's under the documented 12k per-call cap *and* under the smallest total Slack was observed to accept. **The fix holds whichever way Slack actually counts** — so the diagnosis being uncertain doesn't put the change at risk.

### Remaining known gap (not fixed here)

The fix is purely **preventive** — it rolls over *before* the cap. There's no *reactive* handling: if a workspace's real cap is below 11k, `msg_too_long` is still swallowed by the per-append catch and the text is still silently lost. Reactive rollover would need the injected `NativeStreamTransport` to surface the Slack error code, which it currently doesn't (errors are opaque to this file by design). Worth a follow-up; not smuggling an interface change into this PR.

### Verification

```
$ pnpm build && pnpm check-types
(no output)

$ npx vitest run
 Test Files  23 passed (23)
      Tests  288 passed (288)
```

Both defects are pinned by tests that fail against `23b67bd4a2`: a runaway guard that turns a rollover loop into a failed test rather than a hung one, and a per-message surrogate-edge assertion.

maxkorp (commented):
## Review

Strong PR. The diagnosis is grounded in real egress evidence, the `curPosted` (buffer index) vs `curMessagePosted` (message budget, including synthetic openers/closers) split is exactly the right decomposition, and the self-review comment already found the two defects I would have gone hunting for first (unbounded rollover on a pathological opener; surrogate splits that concatenation-equality assertions cannot see). The `flushChunk` re-check after a text flush and the "don't fail over to legacy on a *continuation* failure" reasoning are both correct and non-obvious.

**How I verified.** I copied `native-stream.ts` / `auto-close-streaming.ts` / the new test file at head (`6d5fbf3e3f`) into an isolated directory and ran them under the package's vitest — the 15 tests in `native-stream.test.ts` pass there. I did not re-run the full 288; CI is green on all checks. I then wrote four probe files against the head implementation to try to falsify the change. Numbers below are from those runs, not from reading.

---

### 1. The "holds whichever way Slack counts" claim is unit-dependent — worth resolving before ship (high)

The self-review argues the fix is safe because 11,000 chars is under both the 12k per-call cap and the smallest observed-accepted total. That holds across *char-cap vs per-call-cap vs cumulative-cap*, but not across **chars vs UTF-8 bytes** — and the incident datapoint cannot distinguish them. A novella in English at 11,607 chars is also 11,607 bytes, so `cap = 12000 chars` and `cap = 12000 bytes` fit the evidence identically.

Under the byte hypothesis the incident reproduces exactly, at the default limit, with the fix in place. Fake transport counting UTF-8 bytes with a 12,000-byte cumulative cap, incremental `append()` at production cadence, `漢字` × 10,000:

```
messages=1 delivered=4000/20000 chars (12000 bytes) rejections=41 loggedErrors=41
rollovers happened: false
```

Zero rollovers — the soft limit is in chars and is never reached, so the preventive path never fires. 41 swallowed `msg_too_long`, 80% of the reply gone, turn never finalizes. That is the original bug, unchanged, for any non-Latin reply. Cyrillic/Greek/Hebrew/Arabic at ~2 bytes/char break above ~6k chars; CJK above ~4k.

Single-shot (whole reply in one `append`) is worse still, because the message-fill append is one ~11k slice and a rejection loses everything not yet posted:

```
byte cap 12k, 20k CJK chars, one append:   messages=1 delivered=0/20000  appendCalls=1
char cap 8k,  25k ASCII chars, one append: messages=1 delivered=0/25000  appendCalls=1
```

An empty finalized message. (Not a regression — pre-PR the first 12k-char append fails the same way — but it means the blast radius of the constant being wrong on the high side is *the entire reply*, not a truncated tail.)

This doesn't need the reactive-rollover interface change you deferred. It needs the manual pre-ship check in your NOTE to be **a >12k reply that is mostly CJK or Cyrillic**, not English. If that reply arrives intact, the char hypothesis is confirmed and 11,000 is sound. If it truncates, the soft limit has to be expressed in bytes (`Buffer.byteLength`-style accounting against ~11,000 bytes) and the constant as written is wrong for a large fraction of users. Cheap experiment, decides the design.

### 2. A continuation `startStream` failure silently drops the tail at `finish()` (medium — new in this PR)

`ensureStarted`'s doc says a continuation failure "propagates instead, leaving the boundary intact so the next flush retries the continuation." At `finish()` there is no next flush — it enqueues exactly one, and if that flush's `startStream` throws, the error is caught and logged in `flushText`, `curTs` is `undefined`, so the `if (this.curTs)` guard skips `stopStream` and the method returns cleanly.

Transport where `startStream` succeeds once then throws, 25k reply:

```
messages=1  delivered=10940/25000 chars  fallback used=false  all stopped=[true]
firstTs=S1
```

14,060 chars gone, no fallback, no `onChunkFailure`-style signal, first message correctly finalized so nothing looks wrong. The turn reports success and checkpoints, so the platform won't replay it either. One transient 5xx on a single API call costs up to a full message of text, and this is a new path — pre-PR there were no continuations to fail on.

Given the PR's own thesis ("the dropped text is invisible by design"), I'd fix this here rather than follow up. Cheapest version: in `finish()`, retry the drain while `curPosted < buffer.length` a bounded number of times before giving up. Slightly better: allow the legacy failover for a continuation after all, but seed the fallback with `buffer.slice(curPosted)` rather than the whole buffer — the "legacy sink is seeded with the entire accumulated buffer" objection in the PR description is a property of `failOverToLegacy`, not a law.

Related, lower stakes: the same `!this.curTs` condition in `flushChunk` permanently disables structured chunks for the turn on a transient continuation-start failure.

### 3. Markdown tables are not re-opened across a boundary (medium)

`detectOpenContext` models fences, inline code, and emphasis. It doesn't model tables, so a table that straddles a rollover loses its header and delimiter row. Table of ~25k chars, incremental cadence:

```
continuation 1 starts: "| row 284 value | another value here |\n|"
continuation 2 starts: "| row 566 value | another value here |\n|"
```

No `| --- | --- |` above those rows, so messages 2..N render as literal pipes rather than a table. This file's own header cites native table rendering as a headline reason for using `markdown_text`, and a long generated table is one of the most common ways a reply gets past 11k in the first place — I'd expect this to be hit more often in practice than the straddling code fence the tests do cover.

Options, cheapest first: prefer a blank-line boundary when one exists in the window (usually ends the table cleanly); or detect "inside a table" (a `| --- |` row above the boundary with no intervening blank line) and re-emit header + delimiter in the continuation. Ordered lists restarting at 1 and dropped blockquote markers are the same class of gap but genuinely cosmetic.

### 4. The emphasis closer can land where it cannot close (low)

`renderContextCloser` appends after any trailing whitespace, which the doc comment defends as the closest an append-only transport can get. But the boundary position is a variable you control. Bold run straddling the boundary, observed tails:

```
len=11001 head="**bold words" tail="here bold **"
```

`**bold … bold **` — the closing run is preceded by a space, so it isn't right-flanking and won't close the emphasis under CommonMark; the reader sees literal asterisks. `breakPoint` returns `start + at + 1` to keep the break character on the outgoing message; returning `start + at` when the closer will be emphasis puts the closer flush against the last non-space char and hands the whitespace to the continuation. Safe for the fence case too, since `renderContextCloser` already prepends `\n` when the posted text doesn't end in one. Worth confirming against real Slack in the same manual pass as #1 — its renderer isn't strictly CommonMark.

### Minor

- `renderContextCloser` closes a fence whenever `fenceLang !== null`, whereas `autoCloseOpenMarkdown` requires `hasFenceCodeContent`. A boundary landing right after ` ```python ` therefore emits an empty code block, then re-opens it. Cosmetic and rare, but the asymmetry with the existing helper is worth a line of comment or a matching guard.
- The `room > APPEND_CHAR_LIMIT` branch in `appendPending` is unreachable at the default (11,000 < 12,000) and no test sets a limit above 12k, so it's untested code. Either pin it with a config-override test or note it as config-only.
- The fits-in-one-message regression guard uses 5,000 chars — far enough under 11,000 that it would still pass if the limit were accidentally halved. ~10,900 would actually pin the boundary.
- `appendPending` bursts `startStream`/`appendText`/`stopStream` back to back with no throttle (`lastFlushedAt` updates per flush, not per call). Tier 4 plus `WebClient`'s `Retry-After` handling probably absorbs it; just noting that rollovers now multiply the call count per flush.

### Things I checked that are fine

- Structured chunk after a rollover lands in the correct (current) message; `onChunkFailure` not fired.
- Word boundaries survive the real streaming cadence, which I expected to break them — with `curMessagePosted` at ~10,940 the search window is ~60 chars rather than the 11k the tests exercise, and prose still found a newline or space every time (`cleanBreak=true` at every boundary; base64 hard-cuts at exactly 11,000/11,000/3,600 as designed). The `window.length / 4` floor earns its keep here.
- `firstTs` stays the first message, and the feedback row from `finish(finalBlocks)` lands on the last — correct for a multi-message reply.
- `avoidSurrogateSplit` guards every path including the per-call slices, and `appendSlice`'s empty-delta early return can't stall the loop (the one index where it could return `curPosted` is unreachable because `avoidSurrogateSplit` short-circuits at `index >= text.length`).

---

**Verdict:** not blocking on the substance of the approach — the rollover design is right and the code is careful. I'd want #2 fixed in this PR, since it adds a fresh silent-truncation path to a file whose entire purpose is to stop silently truncating. #1 is your NOTE, sharpened: please make the manual reply non-ASCII, because the answer determines whether `DEFAULT_MESSAGE_CHAR_LIMIT` should be counted in chars at all. #3 I'd take either here or as an immediate follow-up.

BenTaylorDev:
Thanks @maxkorp — this was a genuinely better review than my own self-review. I reproduced all three findings locally before touching anything; all three were real, and #1 in particular kills the safety argument I'd made. Fixed in `d1b289a226`.

## #1 chars vs bytes — fixed, unit changed to bytes

You're right that my "holds whichever way Slack counts" argument had a hole: it covered char-cap vs per-call-cap but not **chars vs UTF-8 bytes**, and the incident reply was English, where the two are 1:1. Reproduced your result almost exactly (33 rejections vs your 41, cadence difference):

```
#1 messages=1 delivered=4000/20000 chars rejections=33 rollovers=false
```

Zero rollovers, 80% of the reply gone — the original bug fully intact for non-Latin scripts.

I went further than making the manual test non-ASCII, because I don't think the ambiguity should survive to ship at all: **the budget is now denominated in UTF-8 bytes** (`DEFAULT_MESSAGE_BYTE_LIMIT`, config `messageByteLimit`). UTF-8 byte length is always ≥ char count, so an 11k-*byte* budget sits under a 12k ceiling whichever unit that ceiling is in. That makes the constant correct without needing the experiment to adjudicate it. The cost is extra messages for non-Latin replies if the ceiling does turn out to be chars — the right side to err on, and noted in the constant's doc.

A side benefit: `spanWithinBudget` advances whole code points, so it *subsumes* the `avoidSurrogateSplit` guard from the previous commit instead of patching each boundary individually.

## #2 tail dropped at `finish()` — fixed here, as you asked

Reproduced: `delivered=11000/25000 fallback=false`, first message finalized, turn looks successful. You're right that this is a fresh silent-truncation path in a file whose whole purpose is to stop silently truncating, so it shouldn't wait for a follow-up.

I took both halves of your suggestion: bounded drain retries in `finish()`, and then the legacy fallback as a last resort — seeded with `buffer.slice(curPosted)` rather than the whole buffer. Your point that the whole-buffer seeding is a property of `failOverToLegacy` and not a law was the useful correction; I'd treated it as a constraint. I deliberately don't fire `onStartFailure` on that path, since a transient continuation failure shouldn't mark the entire workspace legacy.

The `flushChunk` permanent-disable you flagged as related is still there — structured chunks stay off for the rest of the turn after a transient continuation-start failure. Text is unaffected. Happy to fix if you'd rather not carry it.

## #3 tables — fixed here

Confirmed, continuations started mid-row:

```
continuation 1 starts: "row 284 value | another value here |..."
continuation 2 starts: "w 566 value | another value here |..."
```

`tableHeaderToReopen` now re-emits the header + delimiter rows. Agreed this is likelier in practice than the straddling fence the tests already covered — a long generated table is a very ordinary way to pass 11k.

## #4 emphasis flanking — deferred, deliberately

Real, and your `start + at` fix is the right shape. I'm leaving it because it's unverifiable from here: Slack's renderer isn't strictly CommonMark, so whether `**bold **` closes is an empirical question, and "fix" could mean moving whitespace to the continuation for no benefit. Queued for the same manual pass. Say the word if you'd rather have it in speculatively.

## Minors

- Fence closer now mirrors `hasFenceCodeContent` — no more empty code block when a boundary lands just after ` ```lang `.
- Fits-in-one-message guard: 5,000 → **10,900**, so it actually pins the boundary.
- The `callEnd < byteEnd` branch (config-only, unreachable at the default) now has a test setting `messageByteLimit: 40_000`.
- Burst/throttle: acknowledged, no change. `lastFlushedAt` is still per-flush, so a rollover does multiply calls within one flush. Tier 4 plus `WebClient`'s `Retry-After` should absorb it; worth revisiting if 429s show up.

## Verification

```
$ pnpm build && pnpm check-types
(no output)

$ npx vitest run
 Test Files  23 passed (23)
      Tests  292 passed (292)
```

Four new tests, each of which fails against `6d5fbf3e3f`: the CJK byte-divergence rollover, the `finish()` tail drain (asserting `native + tail === text`, so it catches both loss *and* duplication), the table re-open, and the config-only per-call branch.

The manual pre-ship check is now a nice-to-have rather than a gate on the constant, but still worth doing — and per your point it should be a mostly-CJK reply, since that's the case where byte and char budgets disagree most.

maxkorp (commented):
## Second pass — three more in the rollover path, plus two useful negatives

Ran a fresh set of probes against head, going after angles the first pass didn't touch: the failure modes of the boundary itself, the cost of the re-opener, and unbounded growth. Same setup as before (head sources in an isolated dir, package vitest).

### 5. The boundary closer append is unguarded, and its failure disables the rollover entirely (high)

`rollOver()` guards `stopStream` — "A message left rendering as 'still streaming' is cosmetic; dropping the rest of the reply is not. Roll over regardless." The `appendText` two lines above it, which posts the closer, has no such guard. It throws out of `rollOver`, out of `appendPending`, into the per-flush catch, and `curTs` is never cleared. So the next flush re-enters `appendPending`, computes `room = 0`, calls `rollOver`, and fails on the same closer. Forever.

Transport that rejects any append consisting solely of markdown closers; 37k reply inside an open fence:

```
messages=1  delivered=11000/37210  rejections=53  loggedErrors=53  startStream calls=1
```

70% of the reply gone, 53 wasted API calls, zero rollovers. Same input, same code, but with `stopStream` failing instead:

```
messages=4  deliveredChars=37250   (all text delivered)
```

The guarded failure costs nothing; the unguarded one costs the reply. That asymmetry reads as an oversight rather than a decision, and the fix is the reasoning already written for `stopStream` — try/catch the closer, roll over regardless. Degraded markdown on one message beats losing everything after it.

**This is not a contrived trigger, and it compounds with finding #1.** The closer is appended when `curMessagePosted` has just reached `messageCharLimit` — by construction it is *the append that sits closest to the cap*. The headroom argument in `DEFAULT_MESSAGE_CHAR_LIMIT`'s comment ("leaves room for the auto-close closers appended at a boundary") is exactly what protects it. If that headroom assumption is wrong for any reason — a real cap in `[11000, 11003]`, or byte-counting per #1 — then the closer is the **first** append to be rejected, and because its rejection blocks `rollOver`, the preventive fix switches itself off before it ever fires. Two independent things have to be right for the mechanism to engage; only one of them is tested.

### 6. `MIN_MESSAGE_PROGRESS_CHARS` closes the infinite case but leaves an expensive band (medium)

The guard drops the opener only when it can't leave 256 chars of room. Anything between "small" and `limit − 256` is re-injected into every continuation. `detectOpenContext` treats everything up to the first newline after ``` as the language, so a fence followed immediately by a long single line (minified JSON, a base64 payload, one long log line) produces a multi-kilobyte "language tag".

5,000-char blob after ```, then 30k of code:

```
messages: 6
  len=10992 reinjectedPreamble=5004 usefulShare=54%
  len=10992 reinjectedPreamble=5004 usefulShare=54%
  len=10992 reinjectedPreamble=5004 usefulShare=54%
  len=10992 reinjectedPreamble=5004 usefulShare=54%
  len=10992 reinjectedPreamble=5004 usefulShare=54%
  len=5079 reinjectedPreamble=5004 usefulShare=1%
```

The same 5k blob at the top of six consecutive Slack messages, throughput roughly halved, and a final message that is 99% duplicated preamble. Cheap fix that also improves the pathological case: treat an implausibly long `fenceLang` as no language and re-open with a bare ```. A language tag is never 5,000 chars, and a bare fence preserves code formatting, which dropping the opener entirely does not.

### 7. No ceiling on continuation count (low–medium)

```
source=500,012 chars -> messages=46  (startStream=46, appendText=46, stopStream=46)
```

46 Slack messages from one reply, no cap. The legacy `ChunkedMessageStream` has no cap either, so this is consistent with existing behavior rather than a regression — but a runaway agent now converts a single truncated message into 46 real ones, and "five copies in a channel" is the thing this PR was opened to stop. Worth a `MAX_CONTINUATION_MESSAGES` with a terminal "reply truncated" marker, or a snippet/file upload past N. Rendering-only, so a follow-up is defensible; I'd just want it tracked rather than implicit.

### Useful negatives (checked, no action needed)

- **Not O(n²) in practice.** I expected `detectOpenContext` rescanning the whole posted prefix at every rollover to bite on large replies. It doesn't: 500k chars / 46 rollovers completes in 134ms, and 5ms for the open-fence variant. Dead concern, don't spend time on it.
- **Non-prefix buffer mutations are ignored, not garbled.** A same-length rewrite after a rollover (`A`×25k then `B`×25k) delivers `A=25000 B=0`, and a buffer that shrinks below `curPosted` is a no-op. Both are pre-existing `append(fullText)` semantics, unchanged by this PR, and the renderer only ever extends. Fine.
- `private finished` is set in `finish()` and never read anywhere in the file — pre-existing dead state, unrelated to this change, but easy to delete while you're in here. Relatedly, a late `append()` after `finish()` still queues a flush at an already-stopped message; harmless today, also pre-existing.

---

**Updated position.** Findings #2 and #5 are the same bug class — an unguarded failure in newly-added code that silently eats the remainder of a reply — and #5 is worse, because it also prevents the fix from engaging at all. Both are small, local changes. Combined with #1's unit question I'd want those three resolved before merge; #3 (tables), #6, and #7 are all legitimately follow-ups if you'd rather keep this PR tight to the incident.

The core design still looks right to me, and the self-review's two catches were the hard ones. This is the shape of a fix that lands well once the boundary's own failure paths get the same treatment the happy path got.

BenTaylorDev:
Second pass reproduced and fixed in `0f7325ff5b`. #5 was the most valuable finding on this PR — thank you for going after the boundary's own failure paths.

## #5 unguarded boundary closer — fixed

Reproduced, and the control is what makes the point:

```
closer rejected:     messages=1  delivered=11000/37410  rejections=4  starts=1
stopStream failing:  messages=4  delivered=37250        (everything)
```

Same input, same code, one guarded call and one not. You're right that this reads as an oversight rather than a decision — I wrote the "roll over regardless" reasoning for `stopStream` two lines below and simply didn't apply it to the append above it.

Your point about *why* it's not contrived is the part I'd underweighted: the closer is appended exactly when the message has just filled, so it is by construction the first append a wrong headroom assumption rejects — and because its rejection left `curTs` set, the rollover re-entered and failed on the same closer forever. **The preventive fix switched itself off before it could fire.** Two things had to be right and only one was tested. Now guarded, with that reasoning recorded at the catch site.

## #6 long fence language — fixed

Confirmed your numbers exactly (5,004-char preamble, 54% useful, final message 1%). Took your suggested fix rather than widening the progress guard: past `MAX_FENCE_LANG_CHARS` (32) the continuation re-opens with a bare ```. Re-injected preamble drops **5,004 → 4 chars**, continuations go to 100% useful, and as you noted it preserves code formatting where dropping the opener wouldn't. The `MIN_MESSAGE_PROGRESS_BYTES` guard stays as a backstop for anything this doesn't anticipate.

## #7 continuation cap — done here rather than tracked

Confirmed 46 messages from 500k chars. I took it in this PR rather than filing it, because your framing is right: converting one over-long reply into dozens of real messages is the same failure this PR exists to stop. `DEFAULT_MAX_MESSAGES` (20) with a visible truncation marker, configurable via `maxMessages`. Explicit truncation beats unbounded spam, and unlike the original bug the cut is *stated* in the channel.

Also deleted the write-only `finished` field you spotted. Left the late-`append()`-after-`finish()` path alone — pre-existing, harmless, and out of scope.

## Still deferred

- **#4 emphasis flanking.** Unchanged position: needs a real-workspace check since Slack isn't strictly CommonMark.
- **`flushChunk` permanent chunk-disable** after a transient continuation failure.

Both are noted in the thread rather than silently dropped. Happy to take either if you'd rather not carry them.

## Verification

```
$ pnpm build && pnpm check-types
(no output)

$ npx vitest run
 Test Files  23 passed (23)
      Tests  295 passed (295)
```

Three new tests, each failing against `d1b289a226`: closer-rejection still rolls over and delivers everything, continuations re-open with a bare fence and contain no blob, and the message count is capped with a visible marker.

One process note: a `style: auto-fix formatting` bot commit landed on the branch mid-review and reflowed one of my lines, so I rebased onto it rather than force-pushing over it — `fe23c585c5` is preserved in the history.

Thanks for the two useful negatives as well; the O(n²) one in particular would have been my next worry, and knowing 500k/46 rollovers completes in 134ms means I won't spend time there.

maxkorp (commented):
## Correction — I overstated #5, and its "compounds with #1" argument is false

I went back to test the reachability of #5 instead of arguing it, and the argument I attached to it does not survive. Correcting it here because I used that argument to move #2, #5 and #1 into "before merge", and that recommendation was built on it.

**#5 does not compound with #1 — #1 strictly preempts it.** I claimed the closer append is the first thing rejected if the cap assumption is wrong, so a wrong constant would disable the rollover before it fires. Under byte-counting, that is not what happens. Instrumenting every append attempt, 20k CJK chars, 12k-byte cumulative cap:

```
byteCap=12000 CJK: messages=1 delivered=0/20004 rejections=1
   first rejected attempt: len=11000 wouldHaveTotalled=11000  (this is a FILL append, not the closer)
   total appends attempted=1  closer-sized attempts=0
   -> rollover was NEVER reached
```

The *fill* append dies first, `room` never reaches 0, `rollOver` is never called, so the closer is never attempted. The two failure modes are sequential, not compounding. My "two things must be right, only one is tested" framing was wrong.

**And the `msg_too_long` trigger window for #5 is four characters wide.** Fenced whitespace-free payload — the only shape where `breakPoint` hard-cuts at exactly the limit — swept across real-cap values:

```
cap=11000: messages=1 delivered=11000/40004  rejections=1  <- wedged (rejected append: len=4, closer)
cap=11003: messages=1 delivered=11000/40004  rejections=1  <- wedged
cap=11004: messages=4 delivered=40028/40004  rejections=0  <- clean
cap=11607: messages=4 delivered=40028/40004  rejections=0  <- clean (observed-accepted floor)
cap=12000: messages=4 delivered=40028/40004  rejections=0  <- clean
```

Only `[11000, 11003]` wedges. At the observed-accepted floor and at the char-hypothesis cap it never triggers. On prose rather than a blob the fill lands short of the limit anyway (`10984@10984`, closer `3@10987`), so there is slack even in the narrow window.

**#5 as it actually stands:** the code defect is real and the probe in my previous comment does reproduce a permanent wedge — an unguarded `appendText` whose failure blocks `rollOver` forever, sitting two lines above a `stopStream` that *is* guarded with the reasoning that would apply equally to it. Worth the three-line try/catch for exactly that reason. But **low**, not high, and not a merge blocker. Sorry for the noise.

## While I was there, #2 is also narrower than I said

I implied one transient 5xx costs a message of text. It doesn't — the documented retry works. Same transient failure injected mid-stream vs. on the final flush:

```
MID-stream failure:   messages=3  delivered=25000/25000  startCalls=4  lossless=true
FINAL-flush failure:  messages=1  delivered=10978/25000  startCalls=2  lost=14022 chars
```

Mid-stream recovers completely, exactly as `ensureStarted`'s doc comment claims. And under real cadence there are retry chances to spare — `flushes=50 messages=3 rolloversAtFlush=[23,45] flushesAfterLastRollover=5`.

So #2's real window is: the continuation start attempted during `finish()`'s single flush, which by definition has no successor. Narrower than I presented, still genuine — the loss is silent, the reply ends mid-sentence, and the turn checkpoints as a success so the platform won't replay it. A bounded retry in `finish()` closes it. **Should-fix, cheap; not a blocker either.**

## Corrected position

Only #1 is genuinely pre-merge, and it isn't a code change — it's making the manual reply you already planned **non-ASCII**. The arithmetic behind it is unchanged and doesn't depend on anything I got wrong above: 11,607 accepted chars of English is also 11,607 bytes, so the incident cannot distinguish a char cap from a byte cap, and the two imply very different values for `DEFAULT_MESSAGE_CHAR_LIMIT`. One CJK or Cyrillic reply settles it.

Everything else — #2, #3, #5, #6, #7 — is a cheap follow-up. Downgrading my earlier "three before merge" to "one manual check before merge." The design was right the first time; I was wrong about how the failure paths interact.

maxkorp (commented):
## Re-verified both fix commits — all seven confirmed fixed, one new regression in the `maxMessages` code

Re-ran my whole probe battery against `7d893c8d5c`, plus fresh probes for the new byte-budget, drain, table and cap logic. Your 22 tests pass in isolation here too. Results, then the one thing to fix.

### Verified fixed

| | probe result at head |
| --- | --- |
| **#1** byte budget vs a **byte**-counted cap, 20k CJK | `messages=6 delivered=20000/20000 rejections=0 lossless=true`, every message ≤12k bytes |
| **#1** same code vs a **char**-counted cap, 25k prose | `messages=3 delivered=25000/25000 rejections=0` |
| **#2** continuation start fails on the final flush | `native=10990 legacyTail=14010 sum=25000` — `native + tail === source`, exact, no duplication |
| **#2** transient failure mid-stream | `native=25000/25000 legacyUsed=false` — recovers natively, legacy untouched |
| **#5** closer rejected every time | `messages=4 starts=4 closerRejections=3 delivered=37240/37210` — rollover proceeds, nothing lost |
| **#3** table straddling boundaries | continuations now open `"\| col a \| col b \|\n\| --- \| --- \|\n"` |
| **#6** 5k fence "language" | `containsBlob=false` on every continuation, heads are `"```\ncode"` — 5,004 → 4 bytes of preamble |
| **#7** cap, single append | `messages=4 markerCount=1` |

The byte-unit call is the right one. Denominating in bytes makes the constant correct without waiting on the experiment, and `spanWithinBudget` subsuming `avoidSurrogateSplit` is a genuine simplification rather than a patch. Seeding the tail fallback with `buffer.slice(curPosted)` and deliberately not firing `onStartFailure` are both right.

### New: the truncation marker is re-appended on every flush after the cap (medium)

`truncate()` isn't idempotent, and nothing stops the stream re-entering it. It sets `curPosted = buffer.length` and appends the marker, but leaves `curTs` set, never updates `curMessageBytes`, and records no terminal state. Text keeps arriving, so `append()` grows the buffer, the next flush sees `curPosted < buffer.length`, `ensureStarted` returns true on the still-open message, `roomBytes` is already ≤ 0, `rollOver()` re-checks `messageCount >= maxMessages`, and `truncate()` runs again. One marker per flush for the rest of the turn.

At **production defaults** (`maxMessages: 20`), 400k-char reply arriving incrementally:

```
messages=20  markerCount=46  lastMsgChars=13346  lastMsgBytes=13438   (budget=11000)
```

46 copies of `_…reply truncated: too long to deliver to Slack._` stacked at the end of the last message, which is also 2,438 bytes over its own budget — `truncate()` is the one append path that neither checks nor charges `curMessageBytes`. With a transport enforcing a realistic 12k-byte ceiling:

```
messages=20  markerCount=18  msg_too_long=28  lastMsgBytes=11954
```

18 markers land, then 28 appends are rejected — wasted calls carrying the exact error signature from the incident. Control confirms the trigger is specifically "cap reached while text is still arriving": a 60k reply that finishes first gives `markerCount=0 lossless=true`.

This only fires on a runaway reply past ~220k bytes, and it's loud rather than silent, so it isn't a merge-stopper on severity. But it is the spam failure the cap was added to prevent, relocated inside one message — and since the cap exists because I asked for it, I'd rather it not ship this way. A terminal flag set in `truncate()` and checked in `flushText`/`appendPending` covers it; keep `curPosted = buffer.length` so `finish()` still skips the legacy tail path, and charge the marker to `curMessageBytes` so the final append respects the budget like every other one.

### One correction you should not adopt

Your #5 write-up repeats my "the closer is by construction the first append a wrong headroom assumption rejects — two things had to be right and only one was tested" argument. **That argument is wrong and I retracted it** in the comment that landed just after yours ([01:14](#)). Measured: under byte-counting the *fill* append dies first and `rollOver` is never reached (`closer-sized attempts=0`), so the two modes are sequential, not compounding; and the `msg_too_long` window for the closer is only `[11000, 11003]`, clean at 11004, at the 11,607 observed floor, and at 12000.

**The guard itself is right and should stay** — an unguarded append that blocks rollover, next to a guarded one that doesn't, is worth fixing on the asymmetry alone, and your probe reproduces a real permanent wedge. It's just low severity, not the load-bearing thing my earlier comment made it. Worth trimming the rationale in that catch-site comment so the codebase doesn't preserve reasoning I've since disproven.

### Minor residue

- The table re-open works, but a boundary can still cut mid-row, so the re-emitted header is followed by a malformed first row: continuation 1 came back as `"\| col a \| col b \|\n\| --- \| --- \|\nrow 284 value "` — no leading pipe. Under real cadence the search window is only the leftover room, so `breakPoint` often falls through to a space break instead of the row's newline. Strictly better than no header; if you want it exact, prefer a newline break when `tableHeaderToReopen` is non-empty.
- `renderContextCloser`'s new fence check tests `body.includes("\n")`, while `hasFenceCodeContent` additionally requires `/\S/` after that newline — so ` ```py\n ` still gets a closer. Much narrower than before, just not the exact mirror the comment claims.
- On the tail-fallback path, `finalBlocks` go to `stopStream` on the last *native* message while the legacy tail posts after it, so the feedback row lands on the second-to-last visible message. Cosmetic, rare path.

### Position

Everything from both review passes is verified fixed, and #4 / `flushChunk` chunk-disable are reasonable to defer with the reasoning you gave — #4 especially, since guessing at Slack's flanking behavior could easily make it worse. Fix the `truncate()` idempotency and I'm happy to approve; the manual CJK reply remains a nice-to-have rather than a gate now that the budget is in bytes.

BenTaylorDev:
Both passes handled in `0f2f020666`. The `truncate()` regression was real and reproduced at production defaults — and thank you for the retraction, which I've acted on in the code, not just the thread.

## The `maxMessages` regression — fixed

Reproduced before fixing (400k reply, incremental, production defaults):

```
messages=20  markerCount=36  lastMsgBytes=12908  (budget=11000)
with a real 12k ceiling:  markerCount=18  msg_too_long=18
```

Control confirms your diagnosis of the trigger — a 60k reply that finishes before the cap gives `markers=0 lossless=true`.

You called it precisely: `truncate()` recorded no terminal state, so text still arriving grew the buffer, the next flush saw undelivered text on an already-full message, and re-entered. And it was the one append path that neither checked nor charged `curMessageBytes`, which is why the last message ended up over its own budget. Both fixed exactly as you suggested: a `truncated` flag short-circuiting `appendPending` and both `finish()` paths, `curPosted = buffer.length` retained so the legacy tail path stays skipped, and the marker charged like every other append.

You were right to want it fixed rather than shipped — it was the spam the cap exists to prevent, relocated inside one message, and the cap only exists because you asked for it.

## The retraction — applied to the code

I'd repeated your "first append a wrong headroom assumption rejects / two things had to be right, only one tested" argument in the catch-site comment. **Removed.** The comment now rests only on what survives: an unguarded append that wedges rollover, sitting directly above a guarded one that doesn't, is worth three lines on the asymmetry alone — plus a note that the trigger window is a few bytes wide so nobody reads it as load-bearing. Agreed the guard stays and agreed it's low, not high.

Appreciate you going back to measure the reachability instead of defending the argument. `closer-sized attempts=0` under byte counting is exactly the kind of thing I'd have taken on trust.

## Minors

- **Fence content check** now mirrors `hasFenceCodeContent` exactly — non-whitespace after the language line, not merely a newline. You were right that ` ```py\n ` still got a closer and that the comment overclaimed.
- **Table row alignment: partially fixed, and I want to be precise about it.** Inside a table the break now requires a row boundary rather than falling through to the space fallback. Measured:

  ```
  single append:  continuation 1 clean=true   continuation 2 clean=true
  incremental:    continuation 1 clean=false  continuation 2 clean=true
  ```

  So single-append is fixed and incremental is improved but not solved. The residual cause isn't the break choice: under real cadence a *fill* append lands mid-row while room still remains, and `roomBytes` only hits 0 on a later flush — at which point `rollOver` happens with no boundary decision to make. A complete fix means holding back the partial trailing row while inside a table, which needs a "reply still streaming" signal (the `finished` field I just deleted as dead). I'd rather not add that behavior late in the cycle for a cosmetic issue; flagging it clearly instead of calling it done.

- **`finalBlocks` on the tail-fallback path** — acknowledged, not fixed. Cosmetic and rare, and correcting it means reordering the legacy `finish()` against `stopStream`.

## Verification

```
$ pnpm build && pnpm check-types
(no output)

$ npx vitest run
 Test Files  23 passed (23)
      Tests  296 passed (296)
```

New test pins the truncation invariant: exactly one marker however much text keeps arriving, and the last message stays inside the fake's cumulative cap so no post-cut append is rejected.

Another `style: auto-fix formatting` bot commit (`7d893c8d5c` — the one you reviewed against) landed while I was working, so I rebased onto it again rather than force-pushing.

Outstanding by mutual agreement: #4 emphasis flanking, `flushChunk`'s chunk-disable, the incremental table row above, and the CJK manual reply as a nice-to-have now that the budget is in bytes.

BenTaylorDev:
Merged current `main` (`1a5d6678ae`) — conflicts resolved. Flagging this because the resolution was **semantic, not mechanical**, and it touches @mikeryandev's `strict` mode.

Three commits landed on this streamer after my branch point:

| commit | change |
| --- | --- |
| `f1c4b6ca79` | `strict` mode — native text-call failures throw instead of being swallowed |
| `9250f22d88` | `finish()` drains the queue capturing `queueError`, so `stopStream` is always reached |
| `5ddcb14233` | zero-interval flush path in `scheduleFlush` |

## The trap: `native-stream.ts` auto-merged textually but not semantically

Git reported a conflict only in the test file and merged the source cleanly. It isn't clean. **Every path I added — `rollOver`, `truncate`, the `finish()` drain — was written after `strict` existed and didn't honour it**, so the merge as-generated would have quietly defeated it in exactly the places that matter:

- **`rollOver` swallowed both its closer append and its `stopStream`.** Under strict those are required native text calls and must surface. Now it records the closer failure, still reaches `stopStream` (preserving `9250f22d88`'s guarantee that no native stream is left open), then rethrows under strict using the same earliest-error precedence `finish()` uses (`throw closerError ?? err`).
- **`truncate` swallowed the marker append.** Now rethrows under strict, with its terminal state settled *before* the append so a throw can't re-enter it.
- **`finish()`'s drain retries awaited `this.queue` unguarded** — under strict that threw before `stopStream`, **reintroducing precisely the bug `9250f22d88` was written to fix**. Errors are now captured into `queueError` like the initial drain.
- The retry loop also resets `this.queue` first. A strict rejection settles the queue rejected, and `.then()` on a rejected promise skips its callback, so without the reset the retry re-raised the settled error instead of running a flush at all — my drain would have been a no-op under exactly the conditions it exists for.
- **The legacy tail fallback is now gated on `!strict`.** Strict disables the legacy fallback by design; routing an undelivered tail through it would have reinstated the path strict suppresses. Under strict the loss is surfaced as an error instead.

Two tests added for the strict/rollover interaction, which main's strict tests predate and therefore don't cover: a rejected boundary closer surfaces under strict *while the message is still finalized*, and an undelivered tail surfaces rather than engaging the legacy sink.

## Test file

Conflicted in the fake transport's `appendText`: main added `failAppend`, this branch added the cumulative byte cap. Kept both.

## Verification on the merge commit

```
$ pnpm build && pnpm check-types
(no output)

$ npx vitest run
 Test Files  23 passed (23)
      Tests  302 passed (302)
```

That's main's three strict tests plus 24 here, all green together. No caller passes the renamed `messageByteLimit` option under its old name (`grep` across `packages/` is clean, and typecheck confirms).

@maxkorp — nothing in your findings changed, but the strict interaction is new surface since your last pass, so the two new tests and the `rollOver`/`finish()` changes are worth a look if you have the appetite. Everything you asked for pre-merge remains fixed.

tylerslaton (approved):

PR #6246, opened 2026-07-30
fix(react-core): pass description to convert and honor dependencies in useCopilotReadable

### LINKED ISSUE #6243: 🐛 Bug: `useCopilotReadable` calls `convert` with the wrong arguments and ignores the `dependencies` param
### ♻️ Reproduction Steps

`useCopilotReadable` declares `convert` as `(description: string, value: any) => string`,
but the implementation invokes it with a single argument.

**1. Register a readable with a custom `convert`:**

```tsx
import { useCopilotReadable } from "@copilotkit/react-core";

function EmployeeList() {
  useCopilotReadable({
    description: "The list of employees",
    value: { employees: ["Alice", "Bob"] },
    convert: (description, value) => {
      console.log("description =", description); // logs the VALUE object
      console.log("value =", value);             // logs undefined
      return `${description}: ${JSON.stringify(value)}`;
    },
  });

  return null;
}
```

**2. Render the component inside `<CopilotKit>` and open the console.**

**3. Inspect the registered context entry** (e.g. via `copilotkit.context`).

---

**Separately, the second `dependencies` argument has no effect.** It is accepted by the
signature but never wired into the effect:

```tsx
const form = useForm();               // stable reference across re-renders
useCopilotReadable(
  { description: "The form fields and their current values", value: form },
  [form],                             // ← ignored
);
// Typing into the form never updates the context the agent receives.
```

This is the pattern used by `examples/v1/form-filling/components/IncidentReportForm.tsx`.

### ✅ Expected Behavior

```markdown
1. `convert` should be called as `convert(description, value)`, matching its declared type
   `(description: string, value: any) => string`. The resulting context entry should read
   `The list of employees: {"employees":["Alice","Bob"]}`.

2. The `dependencies` argument should be honored, so that passing `[form]` (or any explicit
   dependency list) re-runs the effect and refreshes the context — the behavior in v1.x.
```

### ❌ Actual Behavior

```markdown
1. `convert` is invoked with only the value, so the user's `description` parameter receives
   the value object and `value` is `undefined`. The stored context entry becomes
   `[object Object]: undefined`. (Confirmed by a unit test — the registered value is literally
   the string `"[object Object]: undefined"`.)

2. `dependencies` is accepted but never used. The effect's dependency array is hardcoded, and
   `value` is compared **by reference**, so a stable object reference (such as react-hook-form's
   `form`) never re-triggers the effect and the context is frozen at the first render's snapshot.

---

### Source

`packages/react-core/src/hooks/use-copilot-readable.ts`

```ts
export function useCopilotReadable(
  { description, value, convert, available }: UseCopilotReadableOptions,
  dependencies?: any[],                                  // ← never used
): string | undefined {
  ...
    ctxIdRef.current = copilotkit.addContext({
      description,
      value: (convert ?? JSON.stringify)(value),         // ← called with 1 arg
    });
  ...
  }, [description, value, convert]);                     // ← `dependencies` not spread in
}
```

### This is a regression introduced in v1.50.0

Commit `80dffec4e` ("feat: Reimplement CopilotKit on top of refreshed internals (v1.50.0)")
replaced the v1 implementation, which handled both correctly:

```ts
// before (v1)
const information = convert(description, value);   // both args passed; a *string*
useEffect(() => { ... }, [
  available, information, parentId, addContext, removeContext,
  ...(dependencies || []),                          // dependencies spread in
]);
```

Because `information` was a recomputed **string**, the effect compared by value and picked up
content changes. The current code compares the raw `value` object by reference instead.

### Prior art

PR #3186 reported the `convert` typing problem but was closed as un-mergeable because it
targeted the now-deleted `src/v1.x/packages/react-core/` tree. @BenTaylorDev noted there that
"a fresh PR against the flat tree would be welcome" — filing this issue first, per
CONTRIBUTING.md, before opening that PR.

I'm happy to submit a PR. There is currently no test coverage for this hook, so I'd add
regression tests alongside the fix.

### Additional observations (out of scope for this fix)

- `parentId` and `categories` are still declared on `UseCopilotReadableOptions` (and `parentId`
  has a dedicated "Nested Components" section with an example in the hook's JSDoc), but the
  implementation no longer destructures or uses either one. `ContextStore` in `@copilotkit/core`
  has no notion of them today (entries are `{ description, value, agentIds? }`), so this looks
  like a design question rather than a simple bug fix.
- The dedup check at the top of the effect compares a raw object against an already-serialized
  stored entry:
  `JSON.stringify({ description, value }) == JSON.stringify(ctxItem)`. Since `ctxItem.value` is
  stored as a serialized string, this can never match unless `value` is already a string.

Happy to open separate issues for these if that's useful.

### Screenshots

_No response_

### 𝌚 CopilotKit Version

```shell
@copilotkit/react-core@1.63.2
```

### 📄 Logs (Optional)

```shell

```

jwgrsol:
## What does this PR do?

Fixes two regressions in `useCopilotReadable` (introduced in the v1.50.0 reimplementation, commit `80dffec4e`), plus adds the first test coverage for this hook.

**1. `convert` was called with the wrong arguments.**
`convert` is declared as `(description: string, value: any) => string`, but the implementation invoked it with a single argument:

```ts
// before
value: (convert ?? JSON.stringify)(value),
// after
value: convert ? convert(description, value) : JSON.stringify(value),
```

Previously a custom `convert` received the *value* as its `description` parameter and `undefined` as its `value`, so the stored context entry became `[object Object]: undefined`. The fix is non-breaking: when `convert` is omitted, `JSON.stringify(value)` is used exactly as before.

**2. The `dependencies` argument was silently ignored.**
It is accepted by the signature but was never spread into the effect's dependency array:

```ts
// before
}, [description, value, convert]);
// after
}, [description, value, convert, ...(dependencies ?? [])]);
```

This restores the v1.x behavior of honoring an explicit dependency list.

**3. Tests.**
Adds `use-copilot-readable.test.tsx` (7 cases) — this hook previously had no test coverage. Two cases fail on `main` and pass with this change (regression tests); the rest guard existing behavior (`JSON.stringify` fallback, `available: "disabled"`, unmount cleanup, no re-register when deps are unchanged).

Scope is intentionally limited to the declared-API mismatch. The `parentId`/`categories` options and the dedup check noted in the issue are left out as separate design questions.

## Related PRs and Issues

- Fixes #6243
- Supersedes #3186 (same `convert` fix; that PR was closed as un-mergeable because it targeted the now-deleted `src/v1.x/` tree — reopened here against the flat tree per @BenTaylorDev's note)

## Checklist

### DIFF
--- a/packages/react-core/src/hooks/__tests__/use-copilot-readable.test.tsx
+++ b/packages/react-core/src/hooks/__tests__/use-copilot-readable.test.tsx
@@ -1,0 +1,156 @@
+import { vi } from "vitest";
+import React from "react";
+import { render } from "@testing-library/react";
+import { useCopilotReadable } from "../use-copilot-readable";
+
+interface FakeCopilotKit {
+  context: Record<string, { description: string; value: string }>;
+  addContext: ReturnType<typeof vi.fn>;
+  removeContext: ReturnType<typeof vi.fn>;
+}
+
+let copilotkit: FakeCopilotKit;
+
+function createFakeCopilotKit(): FakeCopilotKit {
+  const context: FakeCopilotKit["context"] = {};
+  let nextId = 0;
+  return {
+    context,
+    addContext: vi.fn(({ description, value }) => {
+      const id = `ctx-${++nextId}`;
+      context[id] = { description, value };
+      return id;
+    }),
+    removeContext: vi.fn((id: string) => {
+      delete context[id];
+    }),
+  };
+}
+
+vi.mock("../../v2", () => ({
+  useCopilotKit: () => ({ copilotkit }),
+}));
+
+describe("useCopilotReadable", () => {
+  beforeEach(() => {
+    copilotkit = createFakeCopilotKit();
+  });
+
+  describe("convert", () => {
+    it("calls convert with (description, value)", () => {
+      const convert = vi.fn(() => "converted");
+      const value = { employees: ["Alice", "Bob"] };
+
+      const Component: React.FC = () => {
+        useCopilotReadable({
+          description: "The list of employees",
+          value,
+          convert,
+        });
+        return null;
+      };
+
+      render(<Component />);
+
+      expect(convert).toHaveBeenCalledWith("The list of employees", value);
+    });
+
+    it("stores the string returned by convert", () => {
+      const Component: React.FC = () => {
+        useCopilotReadable({
+          description: "The list of employees",
+          value: { employees: ["Alice"] },
+          convert: (description, value) =>
+            `${description}: ${JSON.stringify(value)}`,
+        });
+        return null;
+      };
+
+      render(<Component />);
+
+      expect(copilotkit.addContext).toHaveBeenCalledWith({
+        description: "The list of employees",
+        value: 'The list of employees: {"employees":["Alice"]}',
+      });
+    });
+
+    it("falls back to JSON.stringify when convert is omitted", () => {
+      const Component: React.FC = () => {
+        useCopilotReadable({
+          description: "The list of employees",
+          value: { employees: ["Alice"] },
+        });
+        return null;
+      };
+
+      render(<Component />);
+
+      expect(copilotkit.addContext).toHaveBeenCalledWith({
+        description: "The list of employees",
+        value: '{"employees":["Alice"]}',
+      });
+    });
+  });
+
+  describe("dependencies", () => {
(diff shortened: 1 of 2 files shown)


jwgrsol:
Closing in favor of #6409, which has landed woth these fixes (co-authored). 

Thanks @BenTaylorDev and @AlemTuzlak

PR #6253, opened 2026-07-30
fix(core): preserve run IDs across SSE connect replay

### LINKED ISSUE #6252: StateManager loses run identity during cumulative SSE connect replay
## Problem

A custom Runtime can return multiple AG-UI runs in one `/connect` SSE stream. `StateManager` associates every event with the connection input `runId`, and a later cumulative `MESSAGES_SNAPSHOT` reassigns earlier messages to the newest run.

## Expected

Use each `RUN_STARTED.runId` for subsequent state and messages. Once a message is associated with the run that introduced it, cumulative snapshots must not overwrite that association.

## Reproduction

Replay two completed runs through one `/connect` response. Include the first run message again in the second snapshot. `getStateByRun()` cannot retrieve both server runs correctly, and `getRunIdForMessage()` returns the wrong run for the first message.

Scope: standard camelCase AG-UI events from a custom SSE Runtime.

JYbill:
## What does this PR do?

- Uses each `RUN_STARTED.runId` during Runtime SSE connect replay.
- Preserves the first run association when later message snapshots are cumulative.
- Keeps live message events able to correct an earlier provisional run association.
- Adds `/connect` and StateManager regression coverage for multiple server runs.

## Why?

A single `/connect` stream can contain multiple runs. This affects custom Runtime streams and the built-in in-memory runner, which replays historic runs through one reconnect stream. StateManager previously stored replayed state under the connection input ID and reassigned earlier snapshot messages to the latest run.

## Related PRs and Issues

- Closes #6252

## Validation

- `@copilotkit/core` tests: 59 files, 661 tests passed
- `@copilotkit/core` type check
- Pre-commit lint, test, publint, attw, and commitlint

## Checklist

- [x] Documentation is not required for this internal bug fix
- [x] Allow edits by maintainers is enabled


BenTaylorDev (changes requested):
Nice fix — confirmed independently. Your three new assertions fail against current `main`, and with the patch applied `@copilotkit/core` (659 tests) and the full `@copilotkit/react-core` suite (1496) both pass.

Worth adding to the description: this isn't custom-runtime-only. `prepareRunAgentInput` mints one `runId` per connect call, and the built-in in-memory runner replays every historic run through that single stream — so today *every* reconnect collapses a thread's history into one synthetic run.

One thing before merge — `associateMessageWithRun` is now first-write-wins **globally**, not just for replay snapshots:

```
run1 streaming → agent.addMessage("u2")   // pinned to run1 by the activeRun fallback
run1 finished
run2 → MESSAGES_SNAPSHOT([u2])            // run2 is what actually answers u2

getRunIdForMessage(…, "u2")  →  main: "run2"   this PR: "run1", permanently
```

v2 `CopilotChat` serializes sends so it can't reach this, but v1 `useCopilotChat().sendMessage` / `appendMessage` calls `addMessage()` then `runAgent()` with no guard. The effect is grouping-level (`runId`, `messageIndexInRun`, `numberOfMessagesInRun`), not data loss — but it's outside the PR's stated scope and untested. Either:

- keep it global and add a test pinning that semantic, or
- move the guard into `handleMessagesSnapshot` so `handleNewMessage` can still self-correct.

Nit: `event.runId ?? input.runId` — `RunStartedEventSchema` accepts `runId: ""`, so `||` is safer.

JYbill:
@BenTaylorDev Thank you for the detailed review. I have completed the follow-up fixes based on your suggestions:

- Scoped the first-write guard to handleMessagesSnapshot, so live onNewMessage events can correct an earlier provisional run association.
- Switched the RUN_STARTED fallback to event.runId || input.runId to handle an empty runId.
- Added regression tests for both behaviors.
- Updated the PR description to cover the built-in in-memory runner case.

I also merged the latest main into the PR branch. @copilotkit/core passes all 661 tests and type checking. Could you please take another look?

BenTaylorDev (approved):
Re-reviewed at `61ac927b`. All three asks are addressed, and each new test is mutation-checked.

**1. Guard scope** — `associateMessageWithRun` is an unconditional write again; first-write-wins now lives only in `handleMessagesSnapshot` (`state-manager.ts:353-370`). Restoring the global `if (!threadMessages.has(...))` makes `allows a new-message event to correct an earlier run association` fail with `run1` instead of `run2`.

**2. `event.runId || input.runId`** — reverting to `??` makes `falls back to the input run ID when RUN_STARTED carries an empty run ID` fail (`getStateByRun(…, "connect-run")` → `undefined`).

**3. Description** — the in-memory-runner framing is in.

**Verification**

- `@copilotkit/core`: 59 files / 661 tests pass; `tsc --noEmit` clean; `oxlint` clean (the `state-manager.ts:351` unused-param warning is pre-existing on `main`); `oxfmt --check` clean.
- `@copilotkit/react-core` against a rebuilt core `dist`: 1509 pass, across three runs. A fourth run showed one unreproducible failure — but on the *main*-core baseline, not this branch.
- Against current `main`'s `state-manager.ts` the headline transport test fails — `getRunIdForMessage(…, "message-1")` returns a random UUID instead of `server-run-1`. That test drives a real `ProxiedCopilotRuntimeAgent` over a real SSE stream, so the mechanism is proven end-to-end, not mocked.
- CI is green where it counts: unit matrix 6/6 (Node 20/22/24 × React 18/19), `build`, `check-types`, `oxlint`, `format`, `commitlint`, `compat-check`, `package-quality`, `bundle-size`, `copilotchat-import-size`, and the full e2e-dojo integration set.
- 9 commits behind `main`, none touching `state-manager.ts`. Three files, all in `packages/core`. No debris.

The 22 cancelled *Showcase: Build Check* jobs are `timeout-minutes` expiries (`showcase_build_check.yml:248`), not failures — the same thing is hitting six other branches this week. Those images build against published packages, so they can't observe this diff.

Two notes, neither blocking:

**a. The `addMessage`-during-a-run case is still pinned to the old run.** My earlier suggestion doesn't cover the example I gave — a snapshot can no longer correct it either:

```
run1 streaming → addMessage("u2")   // activeRun fallback pins u2 → run1
run1 finished
run2 → MESSAGES_SNAPSHOT([u2])
getRunIdForMessage(…, "u2")  →  main: "run2"   this branch: "run1"
```

Confirmed both values by probe. Grouping metadata only, reachable from v1 `appendMessage` while another run is in flight, and the branch is clearly right in the dominant case (an earlier run's assistant message echoed in a later cumulative snapshot). The real distinction is that the input-less `handleNewMessage` association is *provisional* and a snapshot should be able to override it — follow-up material, not a change here.

**b. A server-minted `RUN_STARTED.runId` now behaves differently.** `StateManager` keys the run by `event.runId`, while `run-handler.ts:511` still captures `logicalRunId = params.input.runId` and passes it as `expectedRunId` to `markNextRunAsContinuation`. When those differ: an explicit resume adopts the server's id (main kept the client's), and a tool follow-up puts the first half under the server's id and the continuation under the client's. Doesn't fire on the shipped surface — CopilotKit's own agents echo `input.runId` (`runtime/src/agent/index.ts:1019,1872`, `demo-agents/src/openai.ts:26`), the Intelligence runner force-overwrites it (`runner/intelligence.ts:531`), CrewAI stamps `input_data.run_id` over its `"?"` placeholders before encoding, and `ag_ui_langgraph` / ADK / Strands / Langroid all pass `input.run_id`. Only a third-party agent minting its own id reaches it, and only grouping moves.

Approving. Thanks for the careful follow-through on all three.

PR #6254, opened 2026-07-30
fix(channels): use channel name for thread lock

tylerslaton:
## Problem

Managed Channel deliveries can request a canonical thread lock with the inner agent ID, or `default` when that ID is unset. Intelligence owns the thread under the declared Channel name, so the lock fails with `THREAD_AGENT_MISMATCH`.

## Why

The SDK treated the agent object identity as the managed Channel identity. Those values are independent: `createChannel({ name })` declares the managed thread owner, while an agent ID is optional and may differ.

## Fix

Pass the declared Channel name into the managed delivery adapter and use it for canonical runs. Add a public `thread.runAgent()` regression test that proves an agent with a different ID still runs under the Channel name.

Validated with the Channels Intelligence test suite, type check, and build; the runtime Channel manager tests, type check, and build; repo lint; and the affected-package pre-commit checks.

### DIFF
--- a/packages/channels-intelligence/src/delivery-adapter.ts
+++ b/packages/channels-intelligence/src/delivery-adapter.ts
@@ -73,6 +73,8 @@ export interface CanonicalChannelRunArgs {
 }
 
 export interface DeliveryAdapterOptions {
+  /** Declared Channel name used to own the canonical Intelligence thread. */
+  channelName: string;
   transport: ChannelDeliveryTransport;
   runCanonical(args: CanonicalChannelRunArgs): Promise<ChannelAgentLoopResult>;
   loadHistory(args: {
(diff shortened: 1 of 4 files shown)


packages/channels-intelligence/src/delivery-adapter.ts
```diff
@@ -337,7 +339,7 @@ export class DeliveryAdapter implements PlatformAdapter {
       threadId,
       runId,
       userId: target.delivery.appUserId,
-      agentId: args.agent.agentId ?? "default",
+      agentId: this.options.channelName,
```
tylerslaton, line 342: this is the key change, rest is testing or around it

MikeRyanDev (approved):

PR #6255, opened 2026-07-30
feat(channels): make Slack reply-continuation limits configurable (refs OSS-689)

BenTaylorDev:
## What

PR #6244 taught the Slack renderer to split a long reply across continuation messages instead of silently truncating it. Its tuning was hardcoded. Three of those constants are genuinely caller-dependent; this exposes them through one `replyContinuation` option on both the direct and managed surfaces.

```ts
// direct
slack({ replyContinuation: { maxMessages: 5 } });

// managed
createChannel({
  name: "support",
  replyContinuation: {
    messageByteLimit: 11_000,
    maxMessages: 20,
    truncationMarker: "\n\n_…réponse tronquée._",
  },
});
```

## Which constants, and why only these

| exposed | why it is a caller's decision |
| --- | --- |
| `messageByteLimit` | Slack's cumulative per-message ceiling is **undocumented**. 11k is inferred from a single production datapoint (11,607 bytes observed accepted) and deliberately conservative. If the real ceiling differs by plan or workspace, an operator needs a knob, not a release. |
| `maxMessages` | How many messages one reply may occupy is a product decision, not a platform fact. 20 was chosen to bound a runaway (500k chars → 46 messages); a support bot and an internal ops bot want different answers. |
| `truncationMarker` | Hardcoded **English** copy posted into the customer's channel. The one constant with no correct default. |

Deliberately **not** exposed, because they are correctness rather than preference:

- `APPEND_CHAR_LIMIT` — a documented Slack per-call limit. A provider fact; exposing it only invites `msg_too_long`.
- `MIN_MESSAGE_PROGRESS_BYTES` — loop-safety invariant. Exposing it lets a caller reintroduce the unbounded-message bug #6244 fixed.
- `MAX_FENCE_LANG_CHARS`, `FINISH_DRAIN_ATTEMPTS` — internal heuristics. If either is wrong that is a bug to fix, not a knob.

## Shape

Grouped under one nested option rather than three flat fields. `maxMessages` sitting bare on a Channel reads ambiguously (thread history?), and the group keeps the next continuation knob from adding another top-level field. The trade-off is that it diverges from `showToolStatus`'s flat precedent — happy to flatten if reviewers prefer consistency over disambiguation.

The shared `ReplyContinuationOptions` type lives in `channels-core`, the common ancestor of all four packages that touch it.

## Plumbing

Both surfaces follow `showToolStatus` exactly:

- **Direct:** `slack({ replyContinuation })` → `adapter.ts` → `event-renderer.ts` → `NativeMessageStream`, covering both the renderer path and `adapter.stream()`'s own stream.
- **Managed:** `createChannel({ replyContinuation })` → `Channel` → `ChannelActivationConfig` → `channel-manager` → `realtime-gateway-launcher` → `DeliveryAdapter` → the renderer's `nativeStreaming` block.

**No gateway or Intelligence change is required.** Managed Slack renders in the SDK process over a gateway live session and only emits `slack.stream.*` effects; the Elixir `provider_executor` is a dumb effect applier that owns no message boundaries. Render config therefore never has to cross into Intelligence.

One non-obvious touchpoint: `channel-manager.ts` keeps a **hand-written structural mirror** of the launcher's options, so the field has to be declared there as well or the managed path type-drifts silently.

## Testing

```
channels-core           171 passed
channels-slack          318 passed
channels-intelligence    67 passed
runtime                1809 passed, 3 failed
```

The 3 runtime failures are pre-existing Gemini `AIMessage` filtering tests, unrelated to this change — confirmed by re-running them with these changes stashed on `main`. `check-types` passes for all four packages.

New coverage:
- the marker override at the leaf (`native-stream`);
- the renderer's pass-through — this one fails without the change, since the 11k/20 defaults would keep that reply in a single message;
- the managed chain end to end via `createChannel` → activation config → launcher opts, plus the negative case that an unset option adds no property anywhere.

## Follow-ups (tracked on OSS-689, not in scope here)

- Confirm the byte-vs-char question with a manual >12k mostly-CJK/Cyrillic reply against a real workspace. It decides whether `messageByteLimit`'s default is right; the option makes it adjustable either way.
- Under the managed path's `minIntervalMs: 0` cadence a table row can still be cut mid-row, so a continuation's re-emitted header is followed by a malformed row.

### DIFF
--- a/packages/channels-core/src/create-channel.ts
+++ b/packages/channels-core/src/create-channel.ts
@@ -196,6 +196,34 @@ export interface StoreConfig<
   dedupTtl?: number;
 }
 
+/**
+ * Tuning for how a long reply is spread across continuation messages.
+ *
+ * Providers cap how much text one message can hold; past that the reply is
+ * split, and past {@link ReplyContinuationOptions.maxMessages} it is truncated
+ * with a visible marker. Defaults are conservative and suit most bots — reach
+ * for these when a provider's real ceiling differs, when a product wants a
+ * different tolerance for how many messages one reply may occupy, or when the
+ * truncation notice needs to be in another language.
+ *
+ * Currently honoured by managed and direct Slack.
+ */
+export interface ReplyContinuationOptions {
+  /**
+   * Soft cap on the UTF-8 bytes one message may hold before continuing into a
+   * new one. Bytes rather than characters because the provider ceiling may be
+   * counted either way, and bytes are the safe reading for non-Latin scripts.
+   */
+  readonly messageByteLimit?: number;
+  /**
+   * Ceiling on messages a single reply may occupy before it is truncated with a
+   * visible marker. Bounds a runaway reply.
+   */
+  readonly maxMessages?: number;
+  /** Notice appended when `maxMessages` is reached. Defaults to English copy. */
+  readonly truncationMarker?: string;
+}
+
 export interface CreateChannelOptions<
   TStateSchema extends StandardSchemaV1 | undefined = undefined,
 > {
@@ -237,6 +265,12 @@ export interface CreateChannelOptions<
    * `slack({ showToolStatus: true })` instead.
    */
   showToolStatus?: boolean;
+  /**
+   * Tuning for splitting a long reply across continuation messages. See
+   * {@link ReplyContinuationOptions}. Applies to managed and direct Slack;
+   * configure direct Slack with `slack({ replyContinuation })` instead.
+   */
+  replyContinuation?: ReplyContinuationOptions;
   agent?: AbstractAgent | ((threadId: string) => AbstractAgent);
   /** @deprecated Pass `store.adapter` instead. */
   actionStore?: ActionStore;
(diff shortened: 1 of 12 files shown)


BenTaylorDev:
~~Added `346329aef8` — `channel-manager.ts` now imports the launcher's options type instead of re-declaring it.~~

**Retracted and reverted in `13e575ea6f`.** The local mirror in `ChannelsIntelligenceModule` is intentional for reasons beyond the CJS/ESM erasure argument I made here, so it stays as-is.

This PR's change to `channel-manager.ts` is back to the minimal one: `+6/-1`, adding `replyContinuation` to the existing mirror and forwarding it. Suites re-verified after the revert — channels-core 171, channels-slack 318, channels-intelligence 67, runtime 1809 passed / 3 pre-existing Gemini failures; `check-types` clean.

Leaving the drift note in the PR description as-is, since it is still a real caveat for anyone adding a launcher option: the mirror has to be edited alongside the launcher or the managed path silently ignores the new field.

MikeRyanDev (approved):

Issue #669, opened 2024-10-02
⭐ Add a Pull Request Template


JeevaRamanathan:
### 💬 Let us know how we can improve our documentation.

I think it would be helpful for both contributors and maintainers to have a pull request template, and I would like to contribute to it.

**Action Items:**
The PR template may consist of the following sections:
1. Type of contribution
2. Contributor checklist
3. Issue link
4. Description
5. Screenshots (if applicable)
6. Additional comments

I would like to know if there are any suggestions or improvements regarding this proposal.

Shubham66020:
Tests & Review: You might want to add a section for running tests, e.g.,

NathanTarbert:
@arielweinberger, what are your thoughts?

Aienthu1:
Hey, @JeevaRamanathan   I am very keen to work on this issue . i am very interested to work on this. I fill very happy to you assigned me this issue .

SOHAIB-code-png:
Hey @JeevaRamanathan please assign me this I have been facing this problem too before, I would be the ideal one  to do this

PR #6260, opened 2026-07-30
fix(channels): isolate factory-returned agents so multi-turn replies work

AlemTuzlak:
## Summary

Follow-up to #6256. Singleton / shared-factory agents still only replied to the first message. This PR fixes that.

The branch originally also carried a mechanism for concurrent same-thread Channel turns (a per-run runner thread id and a per-run product-lock namespace). **That mechanism has been removed** — see [What changed after review](#what-changed-after-review). What remains is the isolation fix and a set of cleanups.

## Root cause (what is actually broken on `main`)

`createChannel` resolves an agent per turn through `agentFactory`. #6256 made the singleton config (`agent: shared`) isolate via `clone()` on every turn, but left the factory branch returning its result raw:

```ts
if (typeof a === "function") return a as (threadId: string) => AbstractAgent;
```

A factory is free to return the same object on every call — `agent: (threadId) => shared` — which is an easy shape to write by accident (and the shape a singleton gets refactored into when someone needs the `threadId`). Left un-isolated, every turn runs an agent still carrying the previous turn's messages and lifecycle flags. It also serializes: `DeliveryAdapter.acquireAgent` chains a chronological tail keyed on **object identity**, so identical objects queue behind one another instead of running independently.

## Fix

- Isolate via `clone()` for **both** the singleton config and the result of a factory, so the object a turn runs on is never one the caller still holds a reference to. A fresh factory (`() => new Agent()`) is unaffected — cloning an unused agent is a clean independent instance.
- Reset `isRunning` and `abortController` on the clone. `clone()` copies both from the source, so a clone taken from an agent that has already run starts out claiming to be mid-run with a spent controller.
- Document that `AbstractAgent.prototype.clone()` copies a fixed field list, so **subclass-declared fields are silently dropped** and read back as `undefined`. There is no loud failure available here: `clone()` always exists on the base class and returns a correctly-typed instance. Subclasses holding their own state must override it.
- Correct the `clone()` error text. It previously offered the factory as an escape hatch; factories are now cloned too, so it is not one.
- Forward `lockKeyPrefix` on lock **cleanup** in both the Channel path and `handlers/intelligence/run.ts`. Acquire and renew already sent it, so a configured prefix acquired one key and released another, leaking the lock until TTL. Pre-existing; one line each now that `CleanupThreadLockRequest` carries the field.
- Collapse four copy-pasted recursive clone-patching test scaffolds into one `patchAgentAndClones` helper.

## What changed after review

Thanks to @BenTaylorDev's review, which was correct on both blocking points. Two landings on `main` also moved underneath this branch while it was open, which is worth recording because it explains the original diff:

| | |
|---|---|
| Branch point (`72750c369a`) | 17:36 UTC |
| This branch's commit | 18:23 UTC |
| #6256 *parallel-by-default turn concurrency* | 17:10 UTC — **26 min before** the branch point |
| #6257 *ordered managed delivery* | 19:53 UTC — **90 min after** the commit |

**#6256 superseded half of the original root cause.** Before it, the singleton branch was `if (a) return () => a;` — the raw shared instance, no cloning at all. Every turn ran the configured object. #6256 fixed that for the singleton config; the factory branch described above is the part it left behind, and the part this PR fixes.

**#6257 removed the scenario the concurrency mechanism was built for.** `DeliveryTransport` now declines any invitation for a thread it is already serving:

```ts
if (this.activeThreads.has(value.canonicalThreadId)) {
  this.options.log?.("channel delivery invitation declined", { reason: "thread_active" });
  return;
}
```

Concurrent same-thread deliveries are no longer reachable within a runtime process, so the per-run runner key and per-run lock namespace had no remaining purpose. Both are removed. Independently, both were broken against the platform:

- **Per-run runner thread id.** `runner.run({ threadId })` is not a local map key for managed Channels. `IntelligenceAgentRunner` sends it as `thread_id` when joining `ingestion:<runId>`, and `RunnerChannel.validate_thread_owner` resolves it against `cpki.threads` — a `<threadId>::<runId>` value fails the join with `thread_not_found`. `buildRunStartedEvent` also re-stamps `event.input.threadId` from it, so preserving the canonical id on the AG-UI input did not survive the real runner. A regression guard now asserts the canonical id reaches `runner.run`.
- **Per-run lock namespace.** The Realtime Gateway derives `thread:{org}:{threadId}:lock` itself in four places and reads its **value** as "which run is active on this thread" (`validate_active_lock`, `lookup_active_run`, `stream_idle` gating). Namespacing per run makes that unanswerable and fails the ingestion join with `active_lock_mismatch`. The lock stays unprefixed unless the runtime explicitly configures a prefix.

One consequence worth flagging separately: `#6257`'s admission gate is **instance state** on `DeliveryTransport`, so it fences one runtime process, not two replicas. Cross-replica, the unprefixed thread lock is the only remaining mutual exclusion — another reason not to namespace it. Filed on [OSS-686](https://linear.app/copilotkit/issue/OSS-686).

The companion PR CopilotKit/Intelligence#669 (app-api honoring `lockKeyPrefix`) is closed: with per-run namespacing removed it has no consumer, and no gateway path honors a prefix today.

## Test plan

- [x] `nx run-many -t test,check-types -p @copilotkit/channels-core @copilotkit/runtime @copilotkit/channels-intelligence` — passed
- [x] channels-core 177, runtime 1812, channels-intelligence 77
- [x] Same suites re-run rebased onto current `main` (post-#6257): channels-core 177, runtime 1813, channels-intelligence 92
- [x] New test: a factory returning a shared instance yields distinct clones for two concurrent turns, neither of them the configured object
- [x] New regression guard: `request.threadId` and `request.input.threadId` reaching `runner.run` are both the canonical product thread id
- [ ] Singleton `agent: new HttpAgent(...)` — 2+ sequential mentions, both get replies
- [ ] Factory `agent: (threadId) => new HttpAgent(...)` still works
- [ ] Managed Channel sequential turn against a live gateway (the path the removed mechanism would have broken)

### Note on the original root cause #2

The original description attributed dead follow-up turns to `HttpAgent.clone()` inheriting an aborted signal. That is real — a clone of an aborted agent does report `signal.aborted === true` — but it cannot be the mechanism: `@ag-ui/client` 0.0.57 `runAgent` assigns `abortController = params?.abortController ?? new AbortController()` before every run, and `channels-core`'s run loop passes no controller. The reset is kept as hygiene and is now commented as such rather than as the fix.

### DIFF
--- a/packages/channels-core/src/create-channel.test.ts
+++ b/packages/channels-core/src/create-channel.test.ts
@@ -58,18 +58,18 @@ function collectText(nodes: ChannelNode[]): string {
 }
 
 /**
- * Capture user messages injected into a fake agent (and every clone of it).
- * createChannel always isolates via `clone()`, so spies on the prototype alone
- * would miss the agent instance that actually runs the turn.
+ * Apply `patch` to `agent` and, recursively, to every clone descended from it.
+ *
+ * `createChannel` isolates every turn via `clone()`, so the instance that
+ * actually runs is never the one configured in the test. A spy installed only on
+ * the configured agent would observe nothing.
  */
-function captureAddedMessages(agent: FakeAgent): unknown[] {
-  const added: unknown[] = [];
+function patchAgentAndClones(
+  agent: FakeAgent,
+  patch: (target: FakeAgent) => void,
+): void {
   const wrap = (target: FakeAgent): void => {
-    const addMessage = target.addMessage.bind(target);
-    target.addMessage = (message) => {
-      added.push(message);
-      return addMessage(message);
-    };
+    patch(target);
     const origClone = target.clone.bind(target);
     target.clone = () => {
       const cloned = origClone();
@@ -78,27 +78,32 @@ function captureAddedMessages(agent: FakeAgent): unknown[] {
     };
   };
   wrap(agent);
+}
+
+/** Capture user messages injected into a fake agent (and every clone of it). */
+function captureAddedMessages(agent: FakeAgent): unknown[] {
+  const added: unknown[] = [];
+  patchAgentAndClones(agent, (target) => {
+    const addMessage = target.addMessage.bind(target);
+    target.addMessage = (message) => {
+      added.push(message);
+      return addMessage(message);
+    };
+  });
   return added;
 }
 
-/** Sum `runAgentCalls` across the prototype and every clone produced from it. */
+/** Sum `runAgent` calls across the configured agent and every clone of it. */
 function trackRunAgentCalls(agent: FakeAgent): { total: () => number } {
   let total = 0;
-  const wrap = (target: FakeAgent): void => {
+  patchAgentAndClones(agent, (target) => {
     const orig = target.runAgent.bind(target);
     target.runAgent = async (parameters, subscriber) => {
       total += 1;
       // Keep FakeAgent's own counter in sync for any direct assertions.
       return orig(parameters, subscriber);
     };
-    const origClone = target.clone.bind(target);
-    target.clone = () => {
-      const cloned = origClone();
-      wrap(cloned);
-      return cloned;
-    };
-  };
-  wrap(agent);
+  });
   return { total: () => total };
 }
 
@@ -460,7 +465,7 @@ describe("createChannel", () => {
     // Capture the context/tools passed to the (isolated) agent's first runAgent call.
     let seenContext: unknown;
     let seenTools: unknown;
-    const wrapRun = (target: FakeAgent): void => {
+    patchAgentAndClones(agent, (target) => {
       const origRunAgent = target.runAgent.bind(target);
       target.runAgent = async (parameters, subscriber) => {
         if (seenContext === undefined) {
@@ -470,14 +475,7 @@ describe("createChannel", () => {
         }
         return origRunAgent(parameters, subscriber);
       };
-      const origClone = target.clone.bind(target);
-      target.clone = () => {
-        const cloned = origClone();
-        wrapRun(cloned);
-        return cloned;
-      };
-    };
-    wrapRun(agent);
+    });
 
     const channel = createChannel({
       adapters: [fake],
@@ -850,17 +848,9 @@ describe("createChannel", () => {
   it("factory that returns a shared instance still isolates per turn", async () => {
     const state = new MemoryStore();
     const shared = new FakeAgent();
-    const seen: FakeAgent[] = [];
 
     const fake = new FakeAdapter();
-    const origGetOrCreate = fake.conversationStore.getOrCreate.bind(
-      fake.conversationStore,
-    );
-    fake.conversationStore.getOrCreate = async (key, target, makeAgent) => {
-      const session = await origGetOrCreate(key, target, makeAgent);
-      seen.push(session.agent as FakeAgent);
-      return session;
-    };
+    const { agents: seen } = captureSessionAgents(fake);
 
     // Common "singleton factory" anti-pattern: returns the same object every call.
     // Parallel turns must still get distinct clones, not serialize on one agent.
@@ -1056,10 +1046,8 @@ describe("createChannel", () => {
     // Capture the context the agent receives on its first runAgent call, and
     // have the fake produce an assistant message with text on agent.messages
     // (mirroring how run-loop expects assistant replies to land there).
-    // Applied to the prototype and every isolate/clone — createChannel never
-    // runs the configured prototype directly.
     let seenContext: unknown;
-    const wrapRun = (target: FakeAgent): void => {
+    patchAgentAndClones(agent, (target) => {
       const origRunAgent = target.runAgent.bind(target);
       target.runAgent = async (parameters, subscriber) => {
         if (seenContext === undefined) {
@@ -1073,14 +1061,7 @@ describe("createChannel", () => {
         });
         return origRunAgent(parameters, subscriber);
       };
-      const origClone = target.clone.bind(target);
-      target.clone = () => {
-        const cloned = origClone();
-        wrapRun(cloned);
-        return cloned;
-      };
-    };
-    wrapRun(agent);
+    });
 
     channel.onMention(async ({ thread }) => {
       await thread.runAgent({ transcript: true });
--- a/packages/channels-core/src/create-channel.test.ts
+++ b/packages/channels-core/src/create-channel.test.ts
@@ -57,15 +57,66 @@ function collectText(nodes: ChannelNode[]): string {
   return out;
 }
 
-/** Capture user messages injected into a fake agent without changing its behavior. */
+/**
+ * Capture user messages injected into a fake agent (and every clone of it).
+ * createChannel always isolates via `clone()`, so spies on the prototype alone
+ * would miss the agent instance that actually runs the turn.
+ */
 function captureAddedMessages(agent: FakeAgent): unknown[] {
   const added: unknown[] = [];
-  const addMessage = agent.addMessage.bind(agent);
-  agent.addMessage = (message) => {
-    added.push(message);
-    return addMessage(message);
+  const wrap = (target: FakeAgent): void => {
+    const addMessage = target.addMessage.bind(target);
+    target.addMessage = (message) => {
+      added.push(message);
+      return addMessage(message);
+    };
+    const origClone = target.clone.bind(target);
+    target.clone = () => {
+      const cloned = origClone();
+      wrap(cloned);
+      return cloned;
+    };
   };
+  wrap(agent);
   return added;
+}
+
+/** Sum `runAgentCalls` across the prototype and every clone produced from it. */
+function trackRunAgentCalls(agent: FakeAgent): { total: () => number } {
+  let total = 0;
+  const wrap = (target: FakeAgent): void => {
+    const orig = target.runAgent.bind(target);
+    target.runAgent = async (parameters, subscriber) => {
+      total += 1;
+      // Keep FakeAgent's own counter in sync for any direct assertions.
+      return orig(parameters, subscriber);
+    };
+    const origClone = target.clone.bind(target);
+    target.clone = () => {
+      const cloned = origClone();
+      wrap(cloned);
+      return cloned;
+    };
+  };
+  wrap(agent);
+  return { total: () => total };
+}
+
+/**
+ * Capture the agent instance handed to the conversation store (post-isolation).
+ * Use when asserting on the agent that actually ran, not the configured prototype.
+ */
+function captureSessionAgents(
+  fake: FakeAdapter,
+): { agents: FakeAgent[] } {
+  const agents: FakeAgent[] = [];
+  const orig = fake.conversationStore.getOrCreate.bind(fake.conversationStore);
+  fake.conversationStore.getOrCreate = async (key, target, makeAgent) => {
+    const session = await orig(key, target, makeAgent);
+    agents.push(session.agent as FakeAgent);
+    return session;
+  };
+  return { agents };
 }
 
 describe("createChannel", () => {
@@ -137,6 +188,7 @@ describe("createChannel", () => {
     Object.defineProperty(fake, "injectInboundTurnOnce", { value: true });
     const agent = new FakeAgent();
     const added = captureAddedMessages(agent);
+    const runs = trackRunAgentCalls(agent);
     const channel = createChannel({ adapters: [fake], agent: () => agent });
 
     channel.onMention(async ({ thread }) => {
@@ -152,7 +204,9 @@ describe("createChannel", () => {
       platform: "fake",
     });
 
-    expect(agent.runAgentCalls).toBe(2);
+    // Two runAgent calls in one handler → two isolated agent instances, each run once.
+    expect(runs.total()).toBe(2);
+    // Implicit inbound prompt is injected only on the first run of the turn.
     expect(added).toEqual([
       expect.objectContaining({
         role: "user",
@@ -184,6 +238,7 @@ describe("createChannel", () => {
       },
       () => undefined,
     ]);
+    const sessionAgents = captureSessionAgents(fake);
     const channel = createChannel({
       adapters: [fake],
       agent: () => agent,
@@ -210,10 +265,12 @@ describe("createChannel", () => {
     });
 
     expect(lifecycleCalls).toHaveLength(1);
-    expect(agent.runAgentCalls).toBe(2);
+    // Tool loop runs twice on the isolated session agent, not the prototype.
+    expect(sessionAgents.agents).toHaveLength(1);
+    expect(sessionAgents.agents[0]!.runAgentCalls).toBe(2);
     expect(canonicalToolEnds).toEqual(["tool-1"]);
     expect(
-      agent.messages.some(
+      sessionAgents.agents[0]!.messages.some(
         (message) => message.role === "tool" && message.toolCallId === "tool-1",
       ),
     ).toBe(true);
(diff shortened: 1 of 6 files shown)


BenTaylorDev (changes requested):
Reviewed as a pair with CopilotKit/Intelligence#669. Verified against current `origin/main` in both repos (not my local checkout). The goal is right and the parallel groundwork is already on main — `DeliveryAdapter` no longer throws `ChannelAgentConcurrencyError`, so concurrent same-thread turns really are reachable now. The problem is the mechanism: both new keys overload identifiers the platform treats as canonical.

## Blocking: the per-run runner thread id breaks managed ingestion

`runnerThreadId = ${canonicalThreadId}::${canonicalRunId}` is passed as `runner.run({ threadId })` (`channel-manager.ts:491`). For managed Channels that runner is **`IntelligenceAgentRunner`** — `CopilotIntelligenceRuntime` constructs it unconditionally (`runtime.ts:414-423`, no `options.runner` override) and `ChannelManager` receives `runtime.runner` (`fetch-handler.ts:238`).

That runner does not treat `request.threadId` as a local map key. It treats it as the canonical product thread id:

- `runner/intelligence.ts:184-196` — joins `ingestion:${input.runId}` with `thread_id: request.threadId`
- `:108-110`, `:121` — stamps `payload.threadId` / `thread_id` and every event's `threadId` from it
- `:409-416` — `buildRunStartedEvent` overwrites `event.input.threadId` with it, undoing the canonical `input.threadId` this PR is careful to preserve

The gateway then refuses the join: `RunnerChannel.validate_thread_owner` selects `cpki.threads WHERE thread_id = $3`, and there is no row for `<uuid>::<runId>`, so join returns `thread_not_found` and the runner emits `RUN_ERROR` / `CHANNEL_JOIN_ERROR` (`runner/intelligence.ts:267-280`). That path is enforced and directly tested (`runner_channel_test.exs:378-414`). Net effect: every managed Channel turn fails, including the sequential singleton case this PR is fixing.

The new assertion `request.input.threadId === canonicalThreadId` passes only because the fake runner never re-stamps; the real runner does.

Suggested shape: don't overload `threadId`. Add a distinct local key to `AgentRunnerRunRequest` (e.g. `runKey`, defaulting to `threadId`) that only the runner's `threads` / `GLOBAL_STORE` map consumes, and keep `request.threadId` canonical for the wire.

## Blocking (cross-repo): the lock prefix is honored in app-api only

Companion detail for #669, but it blocks this PR too. `lockKeyPrefix` exists only in app-api's HTTP routes; the Realtime Gateway derives the same key itself in four places, all unprefixed:

- `sources/runner_channel.ex:128-137` — `validate_active_lock` is strict (`with {:ok, ^run_id}` … else `active_lock_mismatch`), so a prefixed lock fails the ingestion join independently of the item above (tested at `runner_channel_test.exs:437-446`)
- `consumers/redis_writer.ex:95` — the compare-and-delete unlock on RUN_FINISHED/RUN_ERROR silently no-ops
- `consumers/metadata_broadcaster.ex:92` — `lookup_active_run` can no longer find the run
- `client/thread_channel.ex:265` — `stream_idle` gating for connected clients

## Design question: the lock value *is* the platform's active-run index

Those three consumers read `GET thread:{org}:{threadId}:lock` as "which run is active on this thread". Namespacing per run makes that unanswerable, makes acquire unfailable for Channels (the 409 / `THREAD_LOCK_FAILED` path becomes dead), and removes exclusion between a Channel run and a web run on the same thread. `ChannelManagerArgs.lockKeyPrefix` still documents the opposite invariant — "Must match web Intelligence runs so channel + HTTP share the same lock key" (`channel-manager.ts:154`). Multiple concurrent runs per thread looks like a platform data-model change rather than a key-naming one; worth settling before either PR lands.

## On stated root cause #2

`channels-core` pins `@ag-ui/client` 0.0.57, and `run-loop.ts:359-368` calls `agent.runAgent({ tools, context }, subscriber)` with no `abortController`. In 0.0.57 `HttpAgent.runAgent` does `this.abortController = params?.abortController ?? new AbortController()` before every run, so a clone that inherited an aborted controller cannot carry that abort into turn 2's HTTP request. The two manual repro boxes are unchecked. Could you share the repro and which single change actually restores turn 2? Worth knowing whether the fix is the always-clone or the runner key, since the latter is the one with the blast radius above.

## Non-blocking

1. Always-cloning makes `clone()` mandatory for every agent, including factory results. But `AbstractAgent.prototype.clone` always exists and copies a fixed field list, so a custom `AbstractAgent` subclass with its own fields (auth client, config) loses them **silently** rather than hitting the loud error. The error text still offers the factory as the escape hatch, which no longer bypasses cloning — worth a docs note or a stricter check.
2. `if ("abortController" in withAbort || withAbort.abortController)` — the second operand is unreachable. The overwrite also silently detaches an app-held controller.
3. The `activeRunDetach$` / `activeRunCompletionPromise` resets are no-ops: 0.0.57's `clone()` builds from `Object.create` and copies an explicit field list that excludes both, so they are always already `undefined`. The comment "Object.create may have left undefined/shared" doesn't describe the observed behavior.
4. The lock prefix uses `args.runId` while the runner key uses `lock.runId`; they diverge if the platform ever returns a different canonical runId. Pick one.
5. Pre-existing but adjacent and now one line to fix: `handlers/intelligence/run.ts:139-145` `cleanupLock` omits `lockKeyPrefix` while acquire and renew include it, so a prefixed web lock is never released. Latent today (a configured prefix would already break ingestion per above), but this PR adds the field to `CleanupThreadLockRequest` — use it.
6. The recursive clone-patching test scaffold is now copy-pasted four times in `create-channel.test.ts`; one shared helper would do. More importantly, no test here can catch the two blockers by construction (fake runner, no gateway) — at minimum assert that the canonical thread id is what reaches `runner.run`.

maxkorp:
Taking this over from @AlemTuzlak. Both blocking points were correct and are now removed rather than reworked — I verified each against `origin/main` in both repos before touching anything, and one of your open questions turns out to be already settled on `main`.

## Both blockers confirmed

**Per-run runner thread id.** Confirmed exactly as described. `IntelligenceAgentRunner` sends `request.threadId` as `thread_id` when joining `ingestion:<runId>` (`runner/intelligence.ts:184-195` — note the path is `v2/runtime/runner/`, not `v2/runtime/intelligence-platform/runner/`), and `validate_thread_owner` resolves it against `cpki.threads`, so `<uuid>::<runId>` returns `thread_not_found`. `CopilotIntelligenceRuntime` does construct the runner unconditionally, so there is no override path. Your point about the assertion was the sharpest part: `buildRunStartedEvent` re-stamps `event.input.threadId = request.threadId` at `:412-416`, so the careful preservation of the canonical id on the AG-UI input did not survive the real runner either — the assertion passed only against the fake.

**Lock prefix honored in app-api only.** Confirmed all four derivation sites (`runner_channel.ex:131`, `redis_writer.ex:95`, `metadata_broadcaster.ex:92`, `thread_channel.ex:265`), all unprefixed, `validate_active_lock` strict on `{:ok, ^run_id}`. Worth stating the net effect plainly, since it is worse than a missing feature: today's unprefixed lock *matches*, so sequential managed turns work. Landing the pair would have taken that to "every managed Channel turn fails at the ingestion join," for two unrelated reasons.

Both are gone. `runner.run/stop` take `canonicalThreadId` again, and the lock is unprefixed unless the runtime configures a prefix. I took your suggested `runKey` shape as the thing *not* to build — see below.

## Your design question is already answered on `main`

You flagged the lock-value-as-active-run-index conflict and said it looked like a data-model change worth settling before either PR lands. It was settled — in the opposite direction, and about 18 minutes before your review posted. #6257 (*ordered managed delivery*, merged 19:53 UTC) added this to `delivery-transport.ts`:

```ts
if (this.activeThreads.has(value.canonicalThreadId)) {
  this.options.log?.("channel delivery invitation declined", { reason: "thread_active" });
  return;
}
```

One delivery per canonical thread. So the concurrent same-thread scenario is no longer reachable, and there is nothing left for either key to solve — which is why removing beat adding a `runKey`.

**One caveat that argues your way even harder.** `activeThreads` is instance state on `DeliveryTransport`, so it fences one runtime process. Two replicas each keep their own map. Cross-replica, the unprefixed thread lock is the *only* remaining mutual exclusion — not merely an index that would become unanswerable, but the last actual fence. That is squarely [OSS-686](https://linear.app/copilotkit/issue/OSS-686) question 1, and I've added it there.

## On stated root cause #2 — you were right to push

You asked which single change actually restores turn 2, and the honest answer is: not the abort reset. `runAgent` in 0.0.57 does `abortController = params?.abortController ?? new AbortController()` before every run, and `run-loop.ts` passes no controller, so an inherited aborted signal cannot reach turn 2's request. I probed the installed 0.0.57 directly to be sure — a clone of an aborted agent does report `signal.aborted === true`, and `isRunning` is copied too, but neither is load-bearing for a follow-up turn (nothing in the channels path reads `agent.isRunning`; only the runners' own separate state).

What does fix it is narrower than the description claimed: **cloning the result of a factory**. Root cause #1 was accurate against pre-#6256 `main`, where the singleton branch was `if (a) return () => a;` — the raw shared instance, no cloning at all. #6256 landed 26 minutes before this branch's merge-base and fixed that path. What it left behind is `agent: (threadId) => shared`, still returned raw, which is what the new test covers. I've rewritten the body accordingly and kept the abort reset as commented hygiene rather than as the fix.

## Non-blocking items

All six confirmed; 1, 2, 3, 5, 6 addressed, 4 dissolved.

1. **Custom subclass fields.** Verified empirically — `AbstractAgent.prototype.clone()` on a subclass returns a correctly-typed distinct instance with its own fields `undefined`. No loud failure is available, so this is now a prominent doc caveat, and the error text no longer advertises the factory as an escape hatch since factories are cloned too.
2. **Dead operand.** Correct, `'abortController' in withAbort` is true for `HttpAgent` and the `||` can never decide. Reduced to the `in` check.
3. **No-op resets.** Correct — `clone()`'s copy list excludes both fields, and I confirmed sentinels assigned to the source do not appear on the clone. Removed, along with the `Object.create` comment that misdescribed it.
4. **`args.runId` vs `lock.runId`.** Moot now that the prefix is gone.
5. **`run.ts` cleanupLock.** Fixed, and the same omission existed on the Channel path in `channel-manager.ts` — both now forward the prefix.
6. **Test scaffold.** Collapsed into one `patchAgentAndClones` helper (four sites, plus a fifth that had reimplemented `captureSessionAgents` inline). Added the regression guard you asked for: `request.threadId` and `request.input.threadId` reaching `runner.run` must both be the canonical id. You're right that no test here can catch the blockers by construction — the guard at least pins the invariant that the gateway enforces.

## #669

Closed. With per-run namespacing removed it has no consumer, and your point stands regardless: `runtime.lockKeyPrefix` cannot be in use in production today, because any prefix breaks ingestion the same way. The `ttlSeconds`-accepted-and-ignored issue and the missing length/charset bound are captured on OSS-686 rather than shipped as dormant schema.

Verification: `channels-core` 177, `runtime` 1812, `channels-intelligence` 77 on the pushed branch; re-run rebased onto current `main` (post-#6257) at 177 / 1813 / 92. The live managed-gateway sequential turn is still unchecked on my side.

BenTaylorDev:
Superseded by #6265 — closing this one.

That PR keeps the change that mattered here (isolating a factory's result via `clone()`, @AlemTuzlak's original fix) and restates it from `main`. Thanks @maxkorp for removing the runner-key and lock-namespace mechanisms rather than reworking them — both blocking points were real, and the diff is a lot smaller for it.

Three things changed on the way over, recorded so they don't get re-derived:

**The symptom was mis-framed, including by me.** The dead-sequential-turn story doesn't hold: both real adapters replace the message list before the run (managed `agent.messages = [...history]`, Slack `agent.messages = history` plus a fresh per-turn threadId), subscribers are passed per-run rather than accumulated, and `release()` sits in a `finally`. #6256's own root cause is the accurate one — *overlapping* mentions dropped by the `onLockConflict: drop` default plus the managed exclusive gate, both already fixed there. Reproduced at the symptom level, both turns **do** reply; what's actually broken is that each is prompted with the other user's question:

```
expected [ 'first+second', 'first+second' ] to deeply equal [ 'first', 'second' ]
```

So the isolation fix is right, for a reason neither description gave. #6265 asserts that symptom before the object-identity mechanism.

**The lock-key leak doesn't exist.** Intelligence `main` has `threadLockKey = (organizationId, threadId)` with no prefix parameter, and `lockKeyPrefix` is read nowhere under `apps/`. Acquire, renew and cleanup all hit the same unprefixed key, so there is nothing to leak — my review asserted that bug just as confidently and was equally wrong. The cleanup forwarding is dropped in #6265 rather than carried.

**A loud failure for dropped subclass fields is available** — comparing own enumerable keys before/after `clone()`. #6265 adds it, exempting own *functions*: an unconditional version failed 8 existing tests, all of which patch `runAgent` on the instance, and losing an instance-assigned method leaves the prototype method intact.

Also folded in: `ChannelAgentConcurrencyError` is removed. It has been unthrown since #6256, and `git log -S` over `channels-intelligence/src/index.ts` confirms it was never re-exported, so the "kept so older importers do not break" note in its doc never applied.

Still open and unaddressed by either PR: `lockKeyPrefix` is a no-op end to end while two mainline Intelligence demos configure it, and #6257's admission gate is instance state that fences one process rather than two replicas — OSS-686.

maxkorp:
Superseded by #6265 — closing.

That PR restates the isolation change from `main` with an accurate root cause, and I verified its claims independently against a fresh checkout rather than reading them:

- **The bug is real and reproduces at the symptom level.** Running #6265's tests against `main`'s `create-channel.ts`: `expected [ 'first+second', 'first+second' ] to deeply equal [ 'first', 'second' ]`. Two overlapping mentions on one conversation with a shared-instance factory, and each run is prompted with both users' questions.
- **The clone-field guard is new behavior**, not a no-op: `agent whose clone() drops subclass state fails loud` → `promise resolved "undefined" instead of rejecting` on `main`.
- **The function exemption is correctly not too strict** — `does not fail loud when clone() drops an instance-patched method` passes on `main` as well as with the fix.
- **Removing `ChannelAgentConcurrencyError` is not a breaking change.** `channels-intelligence`'s exports map is `.` only, and `index.ts` re-exports from `delivery-contracts`, `intelligence-state-store`, `realtime-gateway-launcher`, `realtime-gateway`, `runtime` — never `delivery-adapter`. Nothing external could reach it.
- 190 tests pass in `channels-core` on that branch.

Two things this PR asserted that are wrong and should not be carried forward:

1. **The lock-key leak does not exist.** This body claims the cleanup change fixes a prefixed lock acquired under one key and released under another. Intelligence `main` has `threadLockKey = (organizationId, threadId)` with no prefix parameter, and `lockKeyPrefix` is read **nowhere** in `apps/` — so acquire, renew and cleanup have always hit the same key. I wrote that claim after having already established the opposite earlier in the same investigation, which is my error. #6265 omits the change.
2. **"There is no loud failure available"** for dropped subclass fields. There is; #6265 implements it. My own repro didn't degrade quietly, it crashed with `TypeError`, which should have told me a check was constructible.

Also worth recording: my read that the fix was unproven was right about the **sequential** framing and wrong to stop there. I measured reply *count* and found both turns replying on `main`, and concluded there was no defect. The defect is in reply *content* under **concurrency** — the axis I didn't test. #6265's test asserts the symptom before the mechanism, which is what makes it hold up where the version here did not.

The isolation idea is @AlemTuzlak's and the test scaffolding carried over; the accurate root cause, the reproduction, and the guard's function exemption are @BenTaylorDev's.

PR #6265, opened 2026-07-30
fix(channels): isolate factory-returned agents per turn

### LINKED ISSUE #669: ⭐ Add a Pull Request Template

BenTaylorDev:
## Summary

Every Channel turn now runs on an agent instance nobody else holds a reference to, for **both** configured shapes — the singleton config and the result of an `agent: (threadId) => …` factory.

This is a fresh start on #6260 (thanks @AlemTuzlak and @maxkorp — the isolation change is theirs). That PR carried a per-run runner thread id and a per-run lock namespace, both of which broke managed ingestion and were removed during review. Rather than keep iterating on a branch whose description had drifted from its diff, this restates the change from `main` with an accurate root cause and the two corrections below.

**This is a concurrency-isolation fix.** Two overlapping mentions on one conversation currently get prompted with each other's questions; that is reproduced below and fixed here. It is *not* a fix for "only the first mention gets a reply" — see [What the original report actually was](#what-the-original-report-actually-was), which turns out to have been fixed already in #6256.

## Root cause

`createChannel` resolves an agent per turn through `agentFactory`. The singleton config was cloned per turn; a factory's result was returned raw:

```ts
if (typeof a === "function") return (threadId: string) => sanitize(a(threadId));
```

A factory is free to return the same object on every call — `agent: (threadId) => shared` — which is easy to write by accident, and is what a singleton becomes when someone refactors to get at the `threadId`.

That matters because **turn concurrency defaults to `"parallel"`** (`ChannelConcurrency`), and the only per-thread admission gate lives in `channels-intelligence` (`ChannelDeliveryTransport.activeThreads`, added in #6257). A **directly connected** adapter — Slack, Discord, Teams, Telegram, WhatsApp — has no such gate, so two turns in one conversation can genuinely run at the same time. On one shared instance they corrupt each other:

- `messages` is a single array both runs append into, so each run's new-message diff picks up the other's
- `isRunning`, `activeRunDetach$` and `activeRunCompletionPromise` are single-slot fields the second run overwrites while the first is still streaming
- managed delivery instead serializes on **object identity** (`DeliveryAdapter.acquireAgent`), so two *different* conversations sharing one instance head-of-line block each other
- managed `historyIds` is a `WeakMap` keyed on the agent, so a second concurrent turn clobbers the first's "which messages are new" baseline

## Changes

- **Clone for both shapes.** A fresh factory (`() => new Agent()`) is unaffected beyond one unused instance.
- **Guard the failure that mandatory cloning introduces.** `AbstractAgent.prototype.clone()` copies a fixed field list, so a subclass declaring its own state (auth client, config, cache) reads it back as `undefined` — and nothing surfaces it, because the base method always exists and returns a correctly-typed instance. Comparing own enumerable keys catches exactly that and names the dropped fields.
  - Own **functions** are exempt. Assigning a method on the instance is how spies and instrumentation wrap an agent, and losing that wrapper leaves the prototype method intact, so the clone still behaves correctly. This is not hypothetical — an unconditional check failed 8 existing tests in this repo, all of which patch `runAgent` on the instance.
- **Reset `isRunning` / `abortController` on the clone**, commented as hygiene rather than as a fix: `runAgent` assigns `abortController = params?.abortController ?? new AbortController()` before every run and the run loop passes none, so an inherited aborted controller cannot reach the next request. Noted inline that this discards `HttpAgent.clone()`'s deliberate propagation of the source's aborted state.
- **Correct the `clone()` error text** — it offered the factory as an escape hatch, which cloning factory results removes.
- **Test scaffolding**: one `patchAgentAndClones` / `captureSessionAgents` pair replaces five hand-rolled copies of the same recursive clone-patching setup.
- **Remove `ChannelAgentConcurrencyError`** — unthrown since #6256. Its doc said it was kept so older importers would not break, but `channels-intelligence` exports only `.` and never re-exported it from `index.ts` (`git log -S` over that file confirms it never did), so nothing outside the package could ever reach it.
- **Preventive regression guard** in `channel-canonical-run.test.ts`: the thread id reaching `runner.run` must stay the canonical product thread id. `IntelligenceAgentRunner` sends it as `thread_id` when joining `ingestion:<runId>` and the gateway resolves it against `cpki.threads`, so decorating it fails the join with `thread_not_found`. Nothing decorates it on `main` — this pins the invariant so that stays true.

### Two things #6260 asserted that are not true

Recording these so they don't get re-derived:

1. **There is no lock-key leak to fix.** #6260 forwarded `lockKeyPrefix` on lock cleanup, described as fixing a prefixed lock acquired under one key and released under another. Intelligence `main` has `threadLockKey = (organizationId, threadId) => \`thread:${org}:${threadId}:lock\`` — no prefix parameter — and `lockKeyPrefix` is read **nowhere** in `apps/`. Acquire, renew and cleanup all hit the same unprefixed key, so nothing leaks. My own review said the same thing and was equally wrong. Omitted here.
2. **A loud failure for dropped subclass fields *is* available** — #6260 states none is. It's the guard above.

## What the original report actually was

#6260 framed this as fixing "2+ **sequential** mentions, only the first gets a reply." That framing was wrong, and the sequential mechanism it implied does not exist. #6256's own root cause is the accurate one:

> Tagging a Slack bot multiple times (or five people asking in one thread) only answered the first turn. Root cause was SDK turn locking (`onLockConflict: drop`) and managed per-thread exclusive agent execution — not Intelligence ingress.

Those are **overlapping** mentions, and both causes were fixed in #6256. There was never a sequential bug: the stated mechanism cannot produce one, because both real adapters replace the message list before the run — managed does `agent.messages = [...history]`, Slack does `agent.messages = history` plus a fresh per-turn threadId. Subscribers are passed per-run to `runAgent(params, subscriber)` rather than accumulated, `session.release?.()` sits in a `finally`, and `isRunning` is cleared by `runAgent`'s own `finally`.

What #6256 left behind is the last hole in *its* story, and it is what this PR closes: under the new parallel default, `agent: (id) => shared` still hands both turns one instance. Both turns reply — so the symptom is not a dead turn — but each is prompted with the other user's question. Reproduced above.

## Follow-ups worth filing (not in this PR)

Two of the four I flagged are handled in this PR (the dead error class, and the sequential mystery — resolved above). Two are not, deliberately:

1. **`lockKeyPrefix` is a no-op end to end**, yet configured in two mainline Intelligence demos (`demos/splat-demo/bff/src/main.ts:371`, `demos/simple-agent/bff/src/main.ts:457`). Deleting it is a breaking removal of a public `CopilotRuntime` option spanning ~30 references across `runtime.ts`, `channel-manager.ts`, `client.ts`, `fetch-handler.ts` and `handlers/intelligence/run.ts`, and it needs a lockstep Intelligence PR for the demos. It also carries a real product question — implement per-prefix locks server-side (the closed #669) or drop the concept. Too big and too breaking to ride along here.
2. **Directly connected adapters have no equivalent of #6257's per-thread admission gate.** A design decision, not a removal: #6256 deliberately made same-conversation turns parallel. Adjacent to OSS-686's cross-replica question, where the unprefixed thread lock is the only fence spanning replicas.

## Testing

Everything below was run in-session on this branch, based on `fcc8616e91`.

**Automated**

| Suite | Result |
|---|---|
| `nx run @copilotkit/channels-core:test` | 190 passed (33 files) — was 187, +3 new |
| `nx run @copilotkit/channels-intelligence:test` | 92 passed (15 files) |
| `nx run @copilotkit/channels-slack:test` | 318 passed (23 files) |
| `nx run @copilotkit/channels-teams:test` | 90 passed (14 files) |
| `nx run @copilotkit/runtime:test` | 1813 passed (128 files) |
| `nx run @copilotkit/channels-core:check-types` | pass |
| `oxfmt` + `oxlint` on changed files | clean (4 pre-existing warnings) |

The first `runtime` run had one failure — `[Fetch] Debug Events > streams debug event envelopes…` in `node-servers.integration.test.ts`. It passes in isolation and passed on a full re-run; it is a full-suite flake, unrelated to this change (the only runtime change here is a test assertion in a different file).

**The bug is reproduced at the symptom level.** Two overlapping turns, a factory returning a shared instance, each run reporting what it believes it was asked — read while both turns are in flight. Against `main`'s source:

```
expected [ 'first+second', 'first+second' ] to deeply equal [ 'first', 'second' ]
```

Both runs saw both users' questions. With the fix each run sees only its own. The test asserts the symptom before the mechanism, so a regression reports that rather than an object-identity puzzle.

**The new tests fail without the fix.** Verified by swapping in `main`'s `create-channel.ts` via `git show` and re-running with the tests kept:

```
× factory returning a shared instance isolates each turn from the others
  AssertionError: expected FakeAgent{…} not to be FakeAgent{…} // Object.is equality
× agent whose clone() drops subclass state fails loud
  AssertionError: promise resolved "undefined" instead of rejecting
✓ does not fail loud when clone() drops an instance-patched method   ← passes both ways, by design
```

The third is a regression guard on the guard itself: it must pass before *and* after, or the check is too strict.

**`clone()` behavior probed directly against the installed `@ag-ui/client` 0.0.57**, rather than assumed — this is what the guard is built on:

```
CustomAgent  source own keys: …,authClient,…,retries,…
CustomAgent  clone  own keys: (authClient and retries absent)
CustomAgent  authClient on clone: undefined | retries: undefined
CustomAgent  instanceof CustomAgent: true        ← typed correctly, silently gutted
CustomAgent  dropped: [ 'authClient', 'retries' ] ← detector fires
HttpAgent    dropped: [ ]                         ← no false positive
```

`HttpAgent`, `LangGraphAgent`, `BuiltInAgent` and `IntelligenceAgent` all override `clone()` and stay quiet. `SanitizingHttpAgent` (channels-slack/teams) declares no own fields, so its prototype methods survive untouched. `ChannelOuterAgent` has three own fields and no override, but never passes through `isolateAgentInstance` — it goes straight to `runner.run`.

**Not verified:** a live managed-gateway turn. Carried over from #6260 as still unchecked.

## Behavior change to be aware of

The new guard turns a previously silent failure into a startup-time throw for anyone using a custom `AbstractAgent` subclass **with its own state and no `clone()` override**. Via the singleton config that shape was already broken (silently); via a factory it used to work. It's a real break, it's loud, and the message names the fields and the fix. Flagging it explicitly rather than burying it in a doc comment.

### DIFF
--- a/packages/channels-core/src/create-channel.test.ts
+++ b/packages/channels-core/src/create-channel.test.ts
@@ -57,15 +57,68 @@ function collectText(nodes: ChannelNode[]): string {
   return out;
 }
 
-/** Capture user messages injected into a fake agent without changing its behavior. */
+/**
+ * Apply `patch` to `agent` and, recursively, to every clone descended from it.
+ *
+ * `createChannel` isolates every turn via `clone()`, so the instance a turn
+ * actually runs on is never the one the test configured. A spy installed only on
+ * the configured agent would observe nothing.
+ */
+function patchAgentAndClones(
+  agent: FakeAgent,
+  patch: (target: FakeAgent) => void,
+): void {
+  const wrap = (target: FakeAgent): void => {
+    patch(target);
+    const origClone = target.clone.bind(target);
+    target.clone = () => {
+      const cloned = origClone();
+      wrap(cloned);
+      return cloned;
+    };
+  };
+  wrap(agent);
+}
+
+/** Capture user messages injected into a fake agent (and every clone of it). */
 function captureAddedMessages(agent: FakeAgent): unknown[] {
   const added: unknown[] = [];
-  const addMessage = agent.addMessage.bind(agent);
-  agent.addMessage = (message) => {
-    added.push(message);
-    return addMessage(message);
-  };
+  patchAgentAndClones(agent, (target) => {
+    const addMessage = target.addMessage.bind(target);
+    target.addMessage = (message) => {
+      added.push(message);
+      return addMessage(message);
+    };
+  });
   return added;
+}
+
+/** Sum `runAgent` calls across the configured agent and every clone of it. */
+function trackRunAgentCalls(agent: FakeAgent): { total: () => number } {
+  let total = 0;
+  patchAgentAndClones(agent, (target) => {
+    const orig = target.runAgent.bind(target);
+    target.runAgent = async (parameters, subscriber) => {
+      total += 1;
+      return orig(parameters, subscriber);
+    };
+  });
+  return { total: () => total };
+}
+
+/**
+ * Capture the agent instances handed to the conversation store — i.e. the ones
+ * that actually run, post-isolation — rather than the configured prototype.
+ */
+function captureSessionAgents(fake: FakeAdapter): { agents: FakeAgent[] } {
+  const agents: FakeAgent[] = [];
+  const orig = fake.conversationStore.getOrCreate.bind(fake.conversationStore);
+  fake.conversationStore.getOrCreate = async (key, target, makeAgent) => {
+    const session = await orig(key, target, makeAgent);
+    agents.push(session.agent as FakeAgent);
+    return session;
+  };
+  return { agents };
 }
 
 describe("createChannel", () => {
@@ -137,6 +190,7 @@ describe("createChannel", () => {
     Object.defineProperty(fake, "injectInboundTurnOnce", { value: true });
     const agent = new FakeAgent();
     const added = captureAddedMessages(agent);
+    const runs = trackRunAgentCalls(agent);
     const channel = createChannel({ adapters: [fake], agent: () => agent });
 
     channel.onMention(async ({ thread }) => {
@@ -152,7 +206,9 @@ describe("createChannel", () => {
       platform: "fake",
     });
 
-    expect(agent.runAgentCalls).toBe(2);
+    // Each `thread.runAgent()` resolves its own isolated instance, so count runs
+    // across the configured agent and its clones rather than on one object.
+    expect(runs.total()).toBe(2);
     expect(added).toEqual([
       expect.objectContaining({
         role: "user",
@@ -184,6 +240,7 @@ describe("createChannel", () => {
       },
       () => undefined,
     ]);
+    const sessionAgents = captureSessionAgents(fake);
     const channel = createChannel({
       adapters: [fake],
       agent: () => agent,
@@ -210,10 +267,14 @@ describe("createChannel", () => {
     });
 
     expect(lifecycleCalls).toHaveLength(1);
-    expect(agent.runAgentCalls).toBe(2);
+    // One `thread.runAgent()` → one isolated instance, and the tool loop iterates
+    // twice on that instance. Assert on it, not on the configured prototype.
+    expect(sessionAgents.agents).toHaveLength(1);
+    const ran = sessionAgents.agents[0]!;
+    expect(ran.runAgentCalls).toBe(2);
     expect(canonicalToolEnds).toEqual(["tool-1"]);
     expect(
-      agent.messages.some(
+      ran.messages.some(
         (message) => message.role === "tool" && message.toolCallId === "tool-1",
       ),
     ).toBe(true);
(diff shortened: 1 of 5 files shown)


tylerslaton (approved):

maxkorp:
Oh I was coming to approve this, because my agent refused to, because I told it to but it said "I would rather someone do it that hasn't been wrong 3 times before" 😅

PR #6267, opened 2026-07-31
docs(react-core): update useAgent thread isolation guidance

VisheshDDev:
## What does this PR do?

Updates the React `useAgent` skill documentation to match the current implementation.

Previously, the docs described per-thread agent cloning using a module-level WeakMap. That behavior was removed, and `useAgent` now returns the registered agent instance instead.

This PR:
- Removes outdated per-thread clone guidance.
- Documents that `threadId` does not create frontend agent isolation.
- Explains that separate frontend surfaces should use distinct local `agentId` values with `runtimeAgentId` when they need isolated state while targeting the same runtime agent.

## Related PRs and Issues

- Related implementation change: `registerProxiedAgent` support and removal of implicit per-thread cloning.

## Checklist

### DIFF
--- a/packages/react-core/skills/react-core/references/agent-access.md
+++ b/packages/react-core/skills/react-core/references/agent-access.md
@@ -256,16 +256,17 @@ sees.
 
 Source: `packages/react-core/src/v2/hooks/use-agent-context.tsx` (no `agentId` parameter); `packages/core/src/core/context-store.ts:26-31`
 
-### MEDIUM — Two components using the same `(agentId, threadId)` expecting isolation
+### MEDIUM — Two components using the same `agentId` expecting isolated state
 
 Wrong:
 
 ```tsx
 function A() {
-  const { agent } = useAgent({ agentId: "default", threadId: "t1" });
+  useAgent({ agentId: "default" });
 }
+
 function B() {
-  const { agent } = useAgent({ agentId: "default", threadId: "t1" });
+  useAgent({ agentId: "default" });
 }
 ```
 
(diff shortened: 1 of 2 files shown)


BenTaylorDev (changes requested):
Thanks for this — the direction is right, and I verified the substance against `main`: there is no `WeakMap` anywhere in `packages/react-core/src`, `use-agent.tsx:52-100` types `agentId`/`runtimeAgentId`/`threadId` as an all-or-nothing trio, and `use-agent.tsx:160-198` throws for partial combinations. So the section you rewrote was genuinely stale, and the `{ agentId, runtimeAgentId, threadId }` shape you document is exactly what the hook's own error message recommends.

The problem is that the same stale claim appears in three other places, one of them in this very file, so merging as-is leaves the skill self-contradictory. Could you extend the change to cover them (in both mirror copies — `skills/react-core/…` and `packages/react-core/skills/react-core/…`)?

1. **`references/agent-access.md:32`** — the file's headline Setup example still calls `useAgent({ agentId: "default", threadId: "main", … })`. Under current `main` that is both a TypeScript compile error and a runtime throw (`use-agent.tsx:160`). It's the first code block a reader copies, so it matters more than the pitfall section you fixed.

2. **`references/agent-access.md:99-130`** — the `CRITICAL — Custom AbstractAgent.clone() that returns this` section still states that `useAgent` calls `source.clone()` to build a per-thread clone and throws `clone() must return a new, independent object`. That error string exists nowhere in `packages/react-core/src` or `packages/core/src` on `main` — the pitfall no longer exists and the section should go.

3. **`references/chat-components.md:185-206`** — repeats the removed-WeakMap explanation and the now-stale `use-agent.tsx:78-119` citation. Its "Correct" example is also wrong now: `CopilotChat.tsx:123` calls `useAgent({ agentId })` with no `threadId` and then assigns `agent.threadId = resolvedThreadId` at `:338`, so two `<CopilotChat agentId="research">` instances with distinct `threadId`s share one agent and clobber each other's thread. That contradicts the sentence this PR adds one file over.

One adjacent item that is **not** yours to fix — flagging it so it doesn't get lost: `showcase/shell-docs/src/content/reference/vue/hooks/useAgent.mdx:40,258,299` documents per-thread clones too, and `packages/vue/src/v2/hooks/use-agent.ts:137-160` has the identical trio rule. Separate PR.

Everything else looks good: both mirror copies stay in sync, and the branch applies cleanly to current `main` despite being behind (neither file has changed on `main` since your merge-base).

BenTaylorDev:
Closing as a duplicate of #6636, which is already approved.

Both PRs rewrite the same `### MEDIUM` section of `packages/react-core/skills/react-core/references/agent-access.md` from the same base, so only one can land. I went with #6636 on two points:

1. It matches shipped code more closely. `packages/react-core/src/v2/hooks/use-agent.tsx:242-245` registers a private proxy through `copilotkit.registerProxiedAgent({ agentId, runtimeAgentId })`, so the `{ agentId, runtimeAgentId, threadId }` triple *does* create a private frontend instance. This PR's replacement text says `threadId` "does not create a new frontend agent or isolate state".
2. It also corrects a second stale passage this PR leaves in place. `agent-access.md:140-144` still says `useAgent` calls `source.clone()` to build a per-thread clone; `clone` appears nowhere in main's `use-agent.tsx`, and a lone `threadId` is now both a type error and a runtime throw (`use-agent.tsx:160-166`).

Your code examples were right — it was the prose around them. And you did one thing #6636 does not: you updated the generated mirror at `skills/react-core/references/agent-access.md`. #6636's CI is red for exactly that omission, so your version flagged a real gap. Sorry this sat as long as it did.

PR #6269, opened 2026-07-31
fix(channels): contain terminal provider failures

MikeRyanDev:
## Summary

- stop the Channels agent loop when a tool handler reports an already-terminal provider delivery
- freeze managed renderer fanout while canonical ingestion records `RUN_ERROR`
- immediately observe Slack native-stream queue failures while preserving them for `finish()`
- treat uncertain managed file-delivery errors as terminal delivery outcomes

## Root cause

The Core run loop converted every tool-handler exception into a model-visible tool result. After `ChannelProviderDeliveryError` closed the effect path, the model could continue and emit text, causing Slack native rendering to call `slack.stream.start` against a closed delivery.

The native stream also retained that rejection in an unobserved internal promise until `finish()`, leaving a Node unhandled-rejection window.

## Validation

- `pnpm nx run-many -t test check-types -p @copilotkit/channels-core,@copilotkit/channels-slack,@copilotkit/channels-intelligence --skip-nx-cache`
- `pnpm nx run-many -t build publint attw -p @copilotkit/channels-core,@copilotkit/channels-slack,@copilotkit/channels-intelligence --skip-nx-cache`
- `pnpm nx run-many -t test check-types build publint attw -p @copilotkit/channels --skip-nx-cache`
- pre-commit affected-package matrix: 17 projects / 24 tasks

### DIFF
--- a/packages/channels-core/src/canonical-run-loop.test.ts
+++ b/packages/channels-core/src/canonical-run-loop.test.ts
@@ -13,6 +13,7 @@ import { expect, test } from "vitest";
 import { z } from "zod";
 import type { CapturedToolCall, RunRenderer } from "./platform-adapter.js";
 import { runAgentLoop } from "./run-loop.js";
+import { ChannelDeliveryTerminatedError } from "./delivery-error.js";
 import { FakeAgent } from "./testing/fake-agent.js";
 import type { ChannelTool } from "./tools.js";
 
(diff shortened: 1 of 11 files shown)


BenTaylorDev:
Reviewed. Verified both new tests actually bite by reverting each fix independently in a worktree:

- without `setQueue`, the native-stream test fails with `expected [ Error: startStream unavailable ] to deeply equal []`
- without the run-loop guard, the canonical test fails with `promise resolved "{ iterations: 2, interrupted: false }" instead of rejecting`

Real regression coverage, and `run-loop.ts:399` is confirmed the only handler-throw→tool-result conversion in channels. `setQueue` covers both rejecting assignments to `this.queue` (the third is `Promise.resolve()`), and freezing renderer fanout on `RUN_ERROR` doesn't leak an open native stream since `thread.ts` best-effort calls `renderer.finish?.()` in its catch and `*.stream.stop` is still sendable after `effectsClosed`.

One follow-up, not a blocker: the brand can't survive the run-loop path. `runAgentLoop` throws → both runner implementations convert it to a `RUN_ERROR` event rather than erroring the observable (intelligence runner via `catchError` → `EMPTY`; in-memory via `catch` → `finalizeRunEvents({ interruptionMessage })` → `complete()`) → `channel-manager.ts:509` synthesizes a fresh `ChannelCanonicalRunError` → `claimAndHandle` sees an unbranded error and still enqueues `terminal({ code: "runtime_handler_failed" })` on top of the gateway's already-recorded `failed`/`failed_before_output`/`uncertain`. The guard does work on paths that skip the runner (e.g. an interaction handler calling `thread.postFile` directly — `onInteraction` rethrows everything but `ActionExpiredError`). Not a regression: pre-PR the same hop produced the same redundant terminal, and this PR does remove the false-`complete` variant. Worth a ticket if a duplicate terminal can overwrite an `uncertain` outcome gateway-side.

Nits:

- the test fixture is `const postFile: ChannelTool = { name: "echo", description: "Post a managed file." }` — three different names for one thing
- `thread.ts:595`'s parenthetical ("tool-handler errors are swallowed inside the loop, so a throw is agent-level") is now stale

BenTaylorDev (approved):

PR #6270, opened 2026-07-31
fix: avoid duplicate tool results when resuming interrupts

### LINKED ISSUE #6201: injectInterrupt synthesizes tool results for interrupts whose agent emits the real ToolCallResult on resume
`@copilotkit/angular@0.3.0` / `@copilotkit/core@1.63.2`

`ɵInterruptState` builds `toolResults` for every standard interrupt carrying a `toolCallId`; the Angular controller adds them via `agent.addMessage()` before resuming. That's right for executor-less interrupt tools (`defineTool({ interrupt: true })`: reason `"tool_call"`, `interrupt.id === toolCallId`, human response = result) — but it conflicts with the [AG-UI interrupt spec](https://docs.ag-ui.com/concepts/interrupts) for suspend/resume backends: *"The agent does not re-emit ToolCallStart/ToolCallArgs/ToolCallEnd in the resumed run — it emits ToolCallResult against the original toolCallId."* (e.g. Mastra `resumeStream`).

Result: two tool messages for one `toolCallId`. `RenderToolCalls#getToolMessage` uses `find()`, so the synthetic resume payload permanently shadows the real result (booking card shows `{"selection":"creditCard"}` → "Failed" despite success).

**Proposal:** gate the synthesis on `interrupt.reason === "tool_call"` (or `interrupt.id === toolCallId`). Backward-compatible: on resume the runtime already synthesizes the result server-side from `input.resume` when no client tool message exists (`alreadyAnswered` path). Alternatively an opt-out on `InjectInterruptOptions`. Happy to PR either.

alectimison-maker:
## What does this PR do?

Fixes executor-less interrupt result ownership so suspend/resume agents can publish their real tool result without a client-generated result shadowing it.

- synthesize `ToolCallResult` messages only for `tool_call` interrupts, whose tools have no executor
- keep backend-owned interrupts limited to resume decisions
- cover the shared interrupt state plus Angular and React consumers

## Related PRs and Issues

- Fixes https://github.com/CopilotKit/CopilotKit/issues/6201

## Validation

- `pnpm -C packages/core test` — 58 files, 616 tests passed
- `pnpm -C packages/angular test` — 46 files, 293 tests passed
- `pnpm -C packages/react-core test` — 122 files, 1,471 Vitest tests and 2 script tests passed
- `check-types` passed for Core, Angular, and React Core
- pre-commit affected-package `test`, `publint`, and `attw` targets passed for 21 projects
- targeted formatting, lint, and `git diff --check` passed

## Checklist

Documentation is unchanged because this corrects internal result ownership without changing the public API.

### DIFF
--- a/packages/angular/src/lib/interrupt.spec.ts
+++ b/packages/angular/src/lib/interrupt.spec.ts
@@ -109,8 +109,14 @@ describe("InterruptController", () => {
   it("accumulates multiple decisions, persists tool results, and resumes once", async () => {
     const { agent, controller, run, startResume } = setup();
     finalizeStandard(agent, [
-      makeInterrupt("one", { toolCallId: "tool-one" }),
-      makeInterrupt("two", { toolCallId: "tool-two" }),
+      makeInterrupt("one", {
+        reason: "tool_call",
+        toolCallId: "tool-one",
+      }),
+      makeInterrupt("two", {
+        reason: "tool_call",
+        toolCallId: "tool-two",
+      }),
     ]);
 
     await controller.cancel("two");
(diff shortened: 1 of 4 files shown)


BenTaylorDev:
Thanks for this, and sorry for the slow response — this one got fixed in parallel before we got to your PR.

The same gate landed in #6470, merged 2026-08-13:

```ts
if (!interrupt.toolCallId || interrupt.reason !== "tool_call") return [];
```

It is on `main` today at `packages/core/src/interrupt-state.ts:148`, with the matching Core, Angular, and React Core test coverage. #6201 is closed.

Your diagnosis and the fix were correct — you and #6470 reached the identical one-line change independently. Closing this as already fixed rather than as wrong.

PR #6271, opened 2026-07-31
fix(showcase): ship query_data dataset inside tools/ and fail loud when missing

lukasmoschitz:
## Ship the `query_data` dataset inside `tools/` and fail loud when it is missing

`showcase/shared/python/tools/query_data.py` resolved its dataset at
`../data/db.csv`, i.e. `showcase/shared/python/data/`. Every integration
Dockerfile copies the `tools/` symlink — **none** copied that sibling
directory. So the CSV was absent at runtime and the `except` branch **silently**
served a 3-row `_MOCK_DATA`.

### Impact (found by live click-through, not by tests)
The 8 integrations that use the shared tool and have no local copy — `ag2`,
`agno`, `claude-sdk-python`, `crewai-crews`, `google-adk`, `langroid`,
`llamaindex`, `strands` — rendered their beautiful-chat charts from 3 mock rows
while looking healthy:

- pie chart → **2 slices** (the two income mock rows: Enterprise Subscriptions, Pro Tier Upgrades)
- bar chart → **1 bar** (the single expense mock row: Engineering Salaries)

Confirmed inside a running container: `/app/tools/query_data.py` present,
`db.csv` nowhere in the image. `langgraph-python`, `langgraph-fastapi`,
`ms-agent-python` and `pydantic-ai` are unaffected because their copies live
under `src/agents/`, which their Dockerfiles do copy.

### Change
1. `git mv` the dataset into the `tools/` package (already copied everywhere) and
   resolve it relative to the module — no Dockerfile or harness change needed.
2. Replace the silent degrade with a `RuntimeError`. The mock stays reachable via
   `SHOWCASE_ALLOW_MOCK_DATASET=1` for unit tests / local runs without the CSV.

A `_logger.warning` was already there and was **not enough**: these integrations
run the agent as a child process whose stdout is not captured, so nobody ever
saw it — the wrong data just shipped.

### Verification
- 40 rows load after the move (mock would be 3).
- Deleting the CSV now raises with an actionable message.
- `SHOWCASE_ALLOW_MOCK_DATASET=1` still yields the 3-row mock.
- `python3 -m py_compile` clean.

### Related observation (not fixed here)
`shared/python/tools/search_flights.py` has the same shape: it probes
`shared/frontend/src/a2ui/…` and an `examples/…` path, neither of which exists in
any image, then falls back to an inline minimal schema with only a
`_logger.warning`. It is currently harmless (the flight card renders correctly
live), so it is out of scope for this PR — flagging it for a follow-up.

Found while QA-ing `claude-sdk-python` for OSS-578.


BenTaylorDev (commented):
Diagnosis is right — applied onto current `main`, the shared tool loads **40 rows** instead of the 3-row mock, and 11 of 13 Python integration Dockerfiles do copy `tools/` with no `data/`. But the branch is stale, and the green CI predates the thing that breaks.

`61eed4a4ad` ("harden CrewAI D6 parity on a3") landed after the branch point and gave `crewai-crews` + `crewai-conversational-flows` a `data` → `../../shared/python/data` symlink slot, a `COPY data/ /app/data/`, and a test asserting it. So the `git mv` regresses those two:

1. **`Validate Showcase` fails.** `validate-shared-symlinks.ts` (`showcase_validate.yml:216`, exits 1) goes clean → `✖ 2 NEW single-source erosion(s)`: `crewai-{crews,conversational-flows}/data (broken-link)`. Neither key is in the baseline.
2. **Two tests fail.** `test_crewai_image_packages_the_shared_financial_dataset` asserts `(data_link / "db.csv").is_file()` — measured `False` after the move.
3. **`crewai-crews` is mis-classified in the description.** It copies `data/` and uses the shared `query_data_impl` (`src/agents/tools/custom_tool.py:17`), so it renders the real 40 rows today. The move takes that away — and since `tools/__init__.py:11` pulls `query_data` into the barrel, the new module-level `RuntimeError` turns it into a hard import crash. The loud-fail amplifies the regression it exists to prevent.
4. **Probably breaks those two Docker builds** (inferred, no daemon to hand): both `stage_shared` impls skip a dangling target (`_common.sh:85`, `lifecycle.ts:187`), leaving `data` dangling for `COPY data/ /app/data/`.

To land: re-target the `data` slot in this PR — repoint both symlinks at `../../shared/python/tools/data`, or retire the slot (symlink + Dockerfile COPY + both tests) and let it ride `tools/`. Either way `expectedTargets()`, the `["tools","shared-tools","data","_shared"]` lists in `_common.sh:76` / `lifecycle.ts:169`, and `lifecycle.test.ts:136-148` move in lockstep.

Nit: `SHOWCASE_ALLOW_MOCK_DATASET` only matches `"1" | "true" | "TRUE"`; case-insensitive would be less surprising.

Before merge, the one thing to prove is a `crewai-crews` image that builds and still serves 40 rows — that's the claim I couldn't run, and it's where this regresses.

PR #6273, opened 2026-07-31
fix(channels): a single failed stream append no longer crashes the runtime (OSS-699)

lukasmoschitz:
## Problem
A single failed Slack streaming provider call crashed the **entire** Channels runtime process (exit 1), taking down every channel/user on that instance. A routine Slack hiccup (rate limit, `msg_too_long`, `streaming_mode_mismatch`, transient network) should degrade the turn — not kill the process.

## Root cause
Once a mid-stream effect fails, `ClaimedChannelDelivery` closes the packet path (`effectsClosed`) and surfaces the failure. But the Slack renderer keeps flushing streamed content (`*.stream.append` / `*.stream.task`) into the now-closed path, and each such packet threw an uncatchable `"packet path is closed"` from `enqueue`. One of those throws escaped the awaited handler chain as an **unhandled rejection** → process exit.

## Fix
Drop moot stream content benignly in `ClaimedChannelDelivery.effect` once the path is closed:
- The **first** failure still surfaces (the delivery is still marked `failed`) — nothing is swallowed.
- Only the **redundant follow-up** stream appends/tasks are dropped.
- Non-stream effects after close still hard-error via `enqueue`.

**Invariant:** a single provider error ends the turn as `failed` but can never crash the runtime — and this generalizes to *any* mid-stream effect failure, not just the tool-status trigger.

## Tests
- New regression test in `delivery-transport.test.ts`: after a mid-stream append fails, a follow-up stream append resolves benignly instead of throwing `"packet path is closed"`.
- Full `@copilotkit/channels-intelligence` suite: **118/118** green. `check-types` / `oxlint` / `oxfmt` clean.

## Notes
- This is the **crash** (severity) fix. The **trigger** — the Slack tool-status renderer mixing `task_update` cards + streamed text in one message (`streaming_mode_mismatch`) — is tracked separately (the opt-in `showToolStatus` rendering).
- Surfaced while verifying OSS-636's default-off behaviour end-to-end against real Slack.

Ref OSS-699

### DIFF
--- a/packages/channels-intelligence/src/delivery-transport.test.ts
+++ b/packages/channels-intelligence/src/delivery-transport.test.ts
@@ -691,6 +691,46 @@ test("retries the exact packet after reconnect and calls no second sequence", as
   expect(result).toEqual({ providerReference: "pref_v1_message_01" });
 });
 
+test("stream content after a mid-stream effect failure is dropped benignly, not thrown as an uncatchable 'packet path is closed' (OSS-699)", async () => {
+  const deliveryChannel = channel();
+  // The gateway rejects the streamed text (e.g. Slack streaming_mode_mismatch).
+  vi.mocked(deliveryChannel.push).mockImplementation((_event, packet) => {
+    const kind = (packet as { payload?: { kind?: string } }).payload?.kind;
+    const result =
+      kind === "slack.stream.append"
+        ? { status: "failed", error: "provider_call_failed" }
+        : { providerReference: "pref_v1_message_01" };
+    return Promise.resolve({ ...(packet as object), phase: "applied", result });
+  });
+  const session = new ClaimedChannelDelivery(
+    preparedDelivery(),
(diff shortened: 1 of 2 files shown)


BenTaylorDev (approved):

lukasmoschitz:
Closing as **superseded by #6269** (contain terminal provider failures), which fixes the same runtime crash (OSS-699) more completely:

- Stops the Core run loop when a tool handler hits a terminal provider delivery — so the model never emits through a closed path (this removes the follow-up stream appends that *this* PR only dropped after the fact).
- Closes the native-stream unhandled-rejection window (a strict failure settling between the event callback and `finish()`) — a real gap I'd missed.

#6269 also touches `delivery-transport.ts`, so these would conflict; its approach is the right, root-cause one. Thanks @MikeRyanDev. Tracked in OSS-699.

Issue #5979, opened 2026-07-15
Bug : CopilotChat message overlap / visual corruption during fast scroll after switching to other tab and coming back


Aryankb:
### Description
When scrolling rapidly up and down in a long conversation, messages visually stack on top of each other and render at incorrect positions. The same corruption occurs after switching to another browser tab and returning — scroll position jumps incorrectly. Slowly scrolling back down eventually settles the layout.

This happens even with a fully static conversation — no streaming, no new messages.

### Steps to Reproduce
1. Have a conversation with many messages.
2. Scroll up and down rapidly — messages overlap and render at incorrect positions.
3. Switch to another browser tab, then return — scroll jumps and layout corrupts.

### Expected Behavior
Messages should always render at correct positions regardless of scroll speed or tab switching.

### What We've Tried
- Replacing `StickToBottom` with `autoScroll="none"`
- Patching `use-stick-to-bottom` to cap animation `tickDelta` on tab restore
- Patching the virtualizer row style from `transform: translateY` to `top`

None of the above have fully resolved the issue.

### Environment
- `@copilotkit/react-core` / `react-ui`: `1.62.2`
- React 18, Next.js 15 (App Router)
- Chrome on Windows 11

### Bug Screenshots

<img width="625" height="770" alt="Image" src="https://github.com/user-attachments/assets/c8d778b6-9ad5-4684-bdd0-61f6837b2dca" />
<img width="637" height="810" alt="Image" src="https://github.com/user-attachments/assets/89b6a362-ade3-4cbc-ad60-aadbe281a8d4" />

BenTaylorDev:
Cross-linking: related to #6089 (variable-height scroll jitter). Both point at the v2 `CopilotChatMessageView` `@tanstack/react-virtual` virtualizer and likely share a fix in its measured-height/translateY handling; the triggers differ (fast-scroll-after-tab-return here vs variable-height hop in #6089).

Aryankb:
@BenTaylorDev Thanks for a response. Is it possible to find a fix for this? We are using copilotkit in production and it would be really helpful if we can get a fix for this

Thanks!

kvnloo:
I think there is still one deterministic path here on current main (`9dd95a4`) that is separate from the variable-height correction in #7370.

`CopilotChatMessageView` currently does:

```
clientHeight === 0 -> scrollElement = null -> shouldVirtualize = false
```

[...]

PR #6274, opened 2026-07-31
fix(storybook): resolve monorepo module aliases to raw source and prevent docgen AST crashes

### LINKED ISSUE #6089: 🐛 Bug: Virtual scrolling jitters badly with messages of varying heights
### ♻️ Reproduction Steps

1.
2.  Description

  When the chat message list contains messages with significantly different heights — especially long code blocks or SQL
  output — scrolling feels extremely jittery. The content "jumps" in chunks rather than scrolling smoothly. It feels
  like the virtual scroller is repeatedly recalculating and snapping to incorrect positions during scroll.

  Steps to reproduce

  1. Have a conversation with a mix of short text messages and very long code block messages (e.g. 50+ line SQL scripts)
  2. Scroll up and down through the history
  3. Observe: every scroll event causes the viewport content to "hop" — jumping by a large chunk instead of moving
  pixel-by-pixel

  Expected behavior

  Smooth native-like scrolling regardless of message height variance. The scroll position should track the user's scroll
  input linearly.

  Environment

  - CopilotKit version: @copilotkit/react-core@v2
  - Browser: Electron 34 (Chromium-based)
  - OS: Windows 11

  Possible cause

  This appears to be related to virtual scrolling with highly variable item sizes. When the virtual list estimates item
  heights and the actual rendered height differs greatly (e.g. a code block renders much taller than the estimate), the
  scroller recalculates and snaps to a new position, causing the visible "hop."

  A common fix in virtual scroll libraries (@tanstack/react-virtual, virtua, etc.) is to measure actual rendered heights
  and update the virtualizer's item size cache without snapping. If CopilotKit uses its own virtual scrolling
  implementation, it may need similar handling for variable-height items.
3.
...

### ✅ Expected Behavior

-

### ❌ Actual Behavior

-

### Screenshots

-

### 𝌚 CopilotKit Version

```shell
-
```

### 📄 Logs (Optional)

```shell

```

VisheshDDev:
## What does this PR do?

Resolves Storybook build failures and module resolution errors when running local development (`pnpm dev`) in `examples/v2/react/storybook`.

### Root Causes Identified:
1. **Path Calculation:** The path resolution in `.storybook/main.ts` was off by one directory hop (`../../../../` instead of `../../.../../`), placing resolution paths inside `examples/` instead of the monorepo root.
2. **Startup Race Condition:** Running `tsdown --watch` alongside Storybook deleted files in `dist/v2` on startup, causing `Module not found` errors as Webpack tried to read them simultaneously.
3. **Docgen AST Crashes:** Storybook's `react-docgen` attempted to parse minified `.mjs` build artifacts in `packages/core/dist/`, crashing the AST parser.

### Key Changes:
- Fixed relative directory math in `.storybook/main.ts` to cleanly target the `packages/` root.
- Updated `@copilotkit/react-core/v2$` and `@copilotkit/core$` Webpack aliases to point directly to raw TypeScript source files (`src/v2/index.ts` and `src/index.ts`).
- Updated the stylesheet alias to point directly to `src/v2/styles/globals.css`.

### Benefits:
- Eliminates `dist/` file deletion race conditions during concurrent watch tasks.
- Shields `react-docgen` from bundled JavaScript while maintaining full auto-generated prop tables from TypeScript interfaces.
- Enables instant Hot Module Replacement (HMR) for core framework source files.

## Related PRs and Issues

- Fixes #6089

## Checklist

### DIFF
--- a/examples/v2/react/storybook/.storybook/main.ts
+++ b/examples/v2/react/storybook/.storybook/main.ts
@@ -1,21 +1,88 @@
 import type { StorybookConfig } from "@storybook/nextjs";
+import path from "path";
+import { fileURLToPath } from "url";
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
 
 const config: StorybookConfig = {
   framework: {
     name: "@storybook/nextjs",
     options: {},
   },
-  stories: ["../stories/**/*.stories.@(tsx|mdx)"],
-  addons: ["@storybook/addon-docs", "@storybook/addon-themes"],
-  webpackFinal: async (cfg) => {
-    // Suppress size warnings for development
-    cfg.performance = {
-      ...cfg.performance,
-      maxAssetSize: 5000000, // 5MB
-      maxEntrypointSize: 5000000, // 5MB
-    };
 
-    return cfg;
+  stories: ["../stories/**/*.stories.@(tsx|mdx)"],
+
+  addons: ["@storybook/addon-docs", "@storybook/addon-themes"],
+
+  webpackFinal: async (config) => {
+    // 1. Correct directory math: 5 hops up to the monorepo root
+    const reactCorePkgPath = path.resolve(
+      __dirname,
+      "../../../../../packages/react-core",
+    );
+
+    console.log("React core package path:", reactCorePkgPath);
+
+    return {
+      ...config,
+
+      performance: {
+        ...config.performance,
+        maxAssetSize: 5000000,
+        maxEntrypointSize: 5000000,
+      },
+
+      resolve: {
+        ...config.resolve,
+
+        symlinks: true,
+
+        alias: {
+          ...config.resolve?.alias,
+
+          // 2. Exact match ($) AND pointing directly to the TypeScript source
+          // This bypasses the dist/ folder entirely, killing the concurrency race condition.
+          "@copilotkit/react-core/v2$": path.join(
+            reactCorePkgPath,
+            "src/v2/index.ts",
+          ),
+
+          // 3. Exact match ($) for styles
+          // CSS is safe to read from dist/source because it is not parsed by docgen.
+          "@copilotkit/react-core/v2/styles.css": path.join(
+            reactCorePkgPath,
+            "src/v2/styles/globals.css",
+          ),
+
+          // 4. NEW FIX: Point core directly to source to bypass dist/ docgen AST crash
+          "@copilotkit/core$": path.resolve(
+            __dirname,
+            "../../../../../packages/core/src/index.ts",
+          ),
+        },
+      },
+
+      module: {
+        ...config.module,
+
+        rules: (config.module?.rules || []).map((rule) => {
+          if (
+            typeof rule === "object" &&
+            rule !== null &&
+            "test" in rule &&
+            rule.test instanceof RegExp &&
+            rule.test.test(".tsx")
+          ) {
+            return {
+              ...rule,
+            };
+          }
+
+          return rule;
+        }),
+      },
+    };
   },
 };
+
 export default config;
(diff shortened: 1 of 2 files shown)


Aryankb:
@tylerslaton @jpr5 @ranst91 @marthakelly @mme @BenTaylorDev @VisheshDDev 
Are there any updates? We are using copilotkit in production and it would be really helpful if we can get a fix for issues
#6089 and #5979 

Thanks!!

Aryankb:
Are there any updates? Thanks!

BenTaylorDev (changes requested):
Thanks for the PR, and sorry for the slow reply here.

This bundles two unrelated changes, and I have findings on both. I would like to see them split.

## 1. The `getItemKey` change duplicates a PR I already closed

`getItemKey: (index) => deduplicatedMessages[index]!.id` in `CopilotChatMessageView.tsx` is the same change as #6564, which I closed on 2026-08-21. The reasoning still applies: #6089 reports height-variance hopping, where long code blocks measure far above `estimateSize: () => 100`. `getItemKey` only re-keys the virtualizer's internal measurement cache, so it does not touch that mechanism. On an append-only chat, index keys and id keys agree.

The change is a genuine improvement on its own merits, because `deduplicatedMessages` can drop an entry mid-list and shift every later index. I am happy to take it as a standalone PR with a test, and without the `Fixes #6089` link. If you do re-file it, please use `deduplicatedMessages[index]?.id ?? index` rather than the non-null assertion.

Leaving #6089 open.

## 2. The Storybook alias change removes every Tailwind utility

I built Storybook twice locally, once on `main` and once with this branch's `.storybook/main.ts`. Both builds succeed. The bundles differ:

| | `main` | this PR |
|---|---|---|
| build result | success | success |
| compiled `cpk:` utility selectors | 1332 | **0** |
| `.cpk\:flex`, `.cpk\:flex-col`, ... | present | absent |

The cause is this alias:

```ts
"@copilotkit/react-core/v2/styles.css": path.join(reactCorePkgPath, "src/v2/styles/globals.css"),
```

`globals.css` is a Tailwind v4 **source** file. Its first line is `@import "tailwindcss" prefix(cpk);`. The example has no `postcss.config.*` and no `tailwindcss` dependency, so nothing compiles it. Webpack inlines Tailwind's own `index.css` and emits zero utility classes. The build stays green and every story renders unstyled.

The alias also skips `packages/react-core/scripts/scope-preflight.mjs`, which scopes `@layer base` under `[data-copilotkit]` to stop CopilotKit styles leaking into the host page. That step runs as part of `build:css` and produces `dist/v2/index.css`, which is what the current alias-free setup already resolves through package exports.

### Smaller notes on the same file

- **The stated root cause does not match the repo.** The description says path resolution was off by one hop, from `../../../../`. `git log --all -S'../../../../' -- examples/v2/react/storybook/.storybook/main.ts` returns exactly one commit, `1d367c635`, which is this branch's own first commit. `main.ts` on `main` has never contained path math.
- **The dist race is narrowed, not eliminated.** `@copilotkit/shared`, `@copilotkit/web-inspector`, and `@copilotkit/web-components` are imported from react-core v2 source and still resolve to `dist`, and the `dev` script still runs `tsdown --watch`.
- `console.log("React core package path:", reactCorePkgPath)` prints on every startup. Please drop it.
- The `module.rules.map(...)` block maps each rule to `{ ...rule }` and returns everything else unchanged, so it is a no-op. Please drop it.
- `symlinks: true` is already the webpack default.

## What I am asking for

1. Split the `CopilotChatMessageView.tsx` change into its own PR, with a test, and without the issue link.
2. Drop the `styles.css` alias, or add a real Tailwind build for the example. Verify by grepping the built bundle for `cpk\:` selectors.
3. Remove the `console.log`, the no-op `module.rules` block, and `symlinks: true`.
4. Update the description so the root cause matches what is on `main`.

One note for anyone reading this while looking for a #6089 fix: this PR does not fix it, and the issue stays open.

PR #6283, opened 2026-08-01
feat(react-ui): add opt-in chat message timestamps

### LINKED ISSUE #5892: 🚀 Feature Request: Option for timestamps in the UI
### Pre-flight Checklist

### Problem or Motivation

Some chatbots can show you the time that a particular message was sent. This is quite a challenge in CopilotKit.

### 🎤 Tell us your idea

AG-UI and CopilotKit options to allow timestamps to show up in the UI.

- [x] No, I’m just sharing the idea.

mikemikimike:
## Summary

- Add an opt-in `showTimestamps` option to `CopilotChat`.
- Record timestamps for user messages created by the default chat input.
- Render accessible timestamps for default user and assistant message renderers.
- Extend shared message types with optional timestamp metadata and add focused coverage.
- Keep the wiring regression test aligned with the v1-deprecated hook location after syncing with `main`.

Fixes #5892

## Validation

- `vitest run src/components/chat/messages/timestamp-wiring.test.ts --config vitest.config.mjs` — 1 file, 4 tests passed locally with Vitest 3.2.6.
- The PR's earlier focused validation recorded `pnpm exec nx run @copilotkit/react-ui:test -- --run` — 9 files, 57 tests passed; `pnpm exec nx run @copilotkit/react-ui:check-types` — passed; and `pnpm exec nx run @copilotkit/react-ui:build` — passed.
- `git diff --check` — passed.
- A full pnpm rerun after the path-only test fix was not possible in this sparse Windows worktree: offline installation stopped on an existing `EPERM` while opening `D:\.pnpm-store\v10\files\05\380ccad...` for `@copilotkit/web-inspector`. No dependency or generated-file changes were included.
- `oxfmt --check` and `oxlint` were not rerun because their executables were not available in the local dependency junction.

## Summary by CodeRabbit

- **New Features**
  - Added optional message timestamps for chat conversations, disabled by default.
  - Timestamps can be displayed for user and assistant messages in localized hour/minute format.
  - Message timestamps are preserved and assigned during active assistant responses.
  - Timestamp support is available in chat windows and modals.

- **Style**
  - Added subtle spacing, sizing, and opacity styling for timestamps.

- **Tests**
  - Added coverage for timestamp rendering, default behavior, and message wiring.


NathanTarbert:
Hey @mikemikimike,

Thanks for this — and for getting the details right that usually go wrong. Reading `message.timestamp` instead of stamping at render time is the correct call, the `<time dateTime={iso}>` element is right, and asserting against the ISO attribute rather than the formatted string keeps your tests from breaking in other timezones. That's better test hygiene than a lot of what lands here.

Two things to sort out before it can go in.

The assistant branch can't currently render. Nothing in the codebase ever sets `timestamp` on an assistant message — only the three user send paths in `Chat.tsx` stamp anything — so with the toggle on you get times on user bubbles and nothing on assistant ones. The note on the issue that the timestamp data is already on the message turns out not to be true yet, which is on us, not you. Either stamp assistant messages when the run's message materializes, or drop the `AssistantMessage` hunk for now and we handle it separately — your call, both are fine.

The other is that `showTimestamps` is silently ignored in `CopilotPopup` / `CopilotSidebar` / `CopilotModal`. It type-checks because `CopilotModalProps extends CopilotChatProps`, but `CopilotModal` builds the `ChatContextProvider` before `CopilotChat` sees the prop, so the value never lands. Destructuring `showTimestamps` in `CopilotModal` and passing it to its provider — same as `icons` and `labels` — fixes it. A Popup-path test would lock it in.

Two smaller ones: timestamps only live in session state, so restored threads show none — worth a line in the TSDoc. And `toLocaleTimeString` can differ between server and client render; `suppressHydrationWarning` on the `<time>` is enough since the ISO attribute is already deterministic.

mikemikimike:
Thanks for the detailed review. I applied the requested direction and synced the branch with `main`.

- Assistant messages are timestamped when a new message materializes during an active run; restored messages are left unchanged unless their persisted data already includes `timestamp`.
- `showTimestamps` is passed into the `ChatContextProvider` in `CopilotModal`; `CopilotPopup` and `CopilotSidebar` use that same modal provider path.
- `MessageTimestamp` renders a deterministic ISO `dateTime`, uses the local time only for display, and sets `suppressHydrationWarning`.
- The shared message type documents that restored threads require persisted timestamps.

While syncing with the current `main`, I also fixed the wiring regression test to read the hook from its current `react-core/src/v1-deprecated/hooks` location (commit `722c9d72`). The focused wiring test now passes locally: 1 file, 4 tests with Vitest 3.2.6. The PR body records the earlier package-level result (9 files, 57 tests, type-check, and build passed) and the current Windows pnpm cache permission limitation.

The updated head is `722c9d72d8c703eae9fe7d8ff1fc888333c9b64a`.

mikemikimike:
Addressed in commit `19c9c93737`.

- Both assistant fallback paths now return `assistantMessage`, preserving generated timestamps for standard assistant messages and coagent bridge rendering.
- Added regression coverage that drives `useCopilotChatInternal` through a new assistant message and verifies the bridge renderer receives the timestamp.

Validation: focused React Core test (8/8), React Core type-check, formatter check, and `git diff --check` pass.

PR #6282, opened 2026-07-31
feat(channels): make identity and memory access explicit

MikeRyanDev:
## Summary

- Require `identifyUser` at Channel creation and separate application users from provider actors.
- Add user/project Memory grants for Channels and the web runtime, including project-only Channel access.
- Bind HITL continuations to immutable identity snapshots and consume each continuation once.
- Update first-party adapters, examples, public docs, and cross-surface tests.

## Why

Channel threads are provider conversations, not personal owners. Each turn must resolve identity and Memory access explicitly so a shared channel cannot inherit the wrong person.

## Impact

This is a hard API cut. It removes `PlatformUser`, `userKey`, and `store.identity`; makes Channel `user` nullable; adds `actor`; and uses one Memory policy for agent and browser access.

Intelligence counterpart: https://github.com/CopilotKit/Intelligence/pull/706

## Validation

- `pnpm nx run-many -t test,build,check-types --projects @copilotkit/channels-core,@copilotkit/channels-intelligence,@copilotkit/channels-slack,@copilotkit/channels-teams,@copilotkit/channels-discord,@copilotkit/channels-telegram,@copilotkit/channels-whatsapp,@copilotkit/channels-ui,@copilotkit/channels,@copilotkit/core,@copilotkit/runtime --outputStyle=static`
- `pnpm lint`
- `npx vitest run && npm run typecheck && npm run build` in `showcase/shell-docs`
- Pre-commit test, publint, and attw checks for 22 public projects.

PR #706, opened 2024-10-05
Fix broken link in quickstart & tutorials

Abinanthan47:
Fixed broken link in README.md

Quickstart & Tutorials

[Quickstart: Chatbot:](https://docs.copilotkit.ai/quickstart-chatbot?ref=github_readme) 

From : https://docs.copilotkit.ai/quickstart-chatbot?ref=github_readme
Screenshot 2024-10-05 102609

To  : https://docs.copilotkit.ai/quickstart

Screenshot 2024-10-05 102717

[Tutorial: Textarea Autocomplete:](https://docs.copilotkit.ai/tutorial-textarea/overview?ref=github_readme) 

From : https://docs.copilotkit.ai/tutorial-textarea/overview?ref=github_readme
Screenshot 2024-10-05 102908

To :  https://docs.copilotkit.ai/tutorials/ai-powered-textarea/overview
Screenshot 2024-10-05 103024

PR #6284, opened 2026-08-01
feat(channels): make identity and memory access explicit

MikeRyanDev:
## Summary

- Require `identifyUser` at Channel creation and separate application users from provider actors.
- Add user/project Memory grants for Channels and the web runtime, including project-only Channel access.
- Bind HITL continuations to immutable identity snapshots and consume each continuation once.
- Update first-party adapters, examples, public docs, and cross-surface tests.

## Why

Channel threads are provider conversations, not personal owners. Each turn must resolve identity and Memory access explicitly so a shared channel cannot inherit the wrong person.

## Impact

This is a hard API cut. It removes `PlatformUser`, `userKey`, and `store.identity`; makes Channel `user` nullable; adds `actor`; and uses one Memory policy for agent and browser access.

Intelligence counterpart: https://github.com/CopilotKit/Intelligence/pull/706

## Validation

- `pnpm nx run-many -t test,build,check-types --projects @copilotkit/channels-core,@copilotkit/channels-intelligence,@copilotkit/channels-slack,@copilotkit/channels-teams,@copilotkit/channels-discord,@copilotkit/channels-telegram,@copilotkit/channels-whatsapp,@copilotkit/channels-ui,@copilotkit/channels,@copilotkit/core,@copilotkit/runtime --outputStyle=static`
- `pnpm lint`
- `npx vitest run && npm run typecheck && npm run build` in `showcase/shell-docs`
- Pre-commit test, publint, and attw checks for 22 public projects.

### DIFF
--- a/examples/slack/README.md
+++ b/examples/slack/README.md
@@ -80,6 +80,7 @@ import { appTools } from "./tools/index.js";
 import { appContext } from "./context/app-context.js";
 
 const bot = createChannel({
+  identifyUser: "platform",
   name: "triage", // every declared Channel needs a unique name
   adapters: [
     slack({
@@ -109,7 +110,9 @@ const bot = createChannel({
 // One handler covers explicit @-mentions and normal DMs.
 // senderContext names the requesting user so the agent acts "as" them.
 bot.onMention(async ({ thread, message }) => {
-  await thread.runAgent({ context: senderContext(message.user) });
+  await thread.runAgent({
+    context: senderContext(message.user, thread.platform),
+  });
 });
 
 // A Channel runs only through the Intelligence runtime, which OWNS its
@@ -121,7 +124,6 @@ const intelligence = new CopilotKitIntelligence({
 const runtime = new CopilotRuntime({
   agents: {}, // the Channel supplies its own agent
   intelligence,
-  identifyUser: () => ({ id: "demo-user", name: "Demo User" }), // demo stub
   channels: [bot],
 });
 
@@ -145,7 +147,7 @@ plain replies in a thread can continue after the bot has posted there.
 
 The bot's tools are plain `ChannelTool`s, collected into `appTools` and spread
 into `createChannel({ tools })`. Each handler receives the generic
-`ChannelToolContext` (`{ thread, message?, user?, signal?, platform }`) the
+`ChannelToolContext` (`{ thread, message?, user, actor, signal?, platform }`) the
 adapter supplies at call time; tools reach platform power (post, postFile,
 `thread.getMessages()`, …) via the `thread` methods:
 
@@ -238,7 +240,10 @@ defineChannelCommand({
   description: "Ask the triage agent anything (no @mention needed).",
   async handler({ thread, text, user }) {
     if (!text) return void thread.post("Usage: `/agent <your question>`");
-    await thread.runAgent({ prompt: text, context: senderContext(user) });
+    await thread.runAgent({
+      prompt: text,
+      context: senderContext(user, thread.platform),
+    });
   },
 });
 ```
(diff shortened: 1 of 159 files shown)


BenTaylorDev:
Reviewed this together with its Intelligence counterpart **CopilotKit/Intelligence#708** (see the note on cross-links at the bottom). CI is green here.

## One real finding

**The managed runtime disarms the Memory-availability guard for direct adapters.** Fail-closed, but silent.

`ChannelManager.start()` calls `channel.ɵruntime.enableIntelligenceMemory()` on every declared channel (`packages/runtime/src/v2/runtime/core/channel-manager.ts:885`), which sets a **channel-wide** flag (`packages/channels-core/src/create-channel.ts:1232`). `assertMemoryAvailable()` then passes on `channelFlag || adapter.supportsIntelligenceMemory` (`packages/channels-core/src/thread.ts:618`).

But `memory` only reaches anything through `adapter.runAgentLifecycle` (`packages/channels-core/src/thread.ts:751-767`) — a direct adapter has none, so the grant is dropped on the floor. Direct adapters are explicitly supported in the same `adapters: []` array as the managed one (per the comment in `ChannelManager.start`), so a `runAgent({ memory })` on a Socket-Mode turn inside a managed runtime runs to completion with **zero Memory tools and no error**.

Proven with a paired test against this branch:

```
✓ standalone direct channel: guard fires (expected)             → channel_memory_unavailable
✓ same channel under a managed runtime — grant silently no-ops  → resolves, ran to completion
```

```ts
async function runDirectAdapterGrant(managed: boolean) {
  const adapter = new FakeAdapter(); // direct adapter: no runAgentLifecycle
  const channel = createChannel({
    identifyUser: "platform",
    adapters: [adapter],
    agent: () => new FakeAgent(),
  });
  // Exactly what ChannelManager.start() does for EVERY declared channel.
  if (managed) channel.ɵruntime.enableIntelligenceMemory();

  channel.onMessage(async ({ thread }) => {
    await thread.runAgent({ memory: { project: "read", user: "read-write" } });
  });
  await channel.ɵruntime.start();
  return adapter.getSink().onTurn({
    conversationKey: "conversation",
    replyTarget: {},
    userText: "hello",
    platform: "slack",
    actor: { id: "U1", kind: "human", name: "A" },
  });
}
```

`DeliveryAdapter` already declares `supportsIntelligenceMemory = true` (`packages/channels-intelligence/src/delivery-adapter.ts:135`), so `enableIntelligenceMemory()` is redundant on the managed path and only manufactures the false positive. Dropping the channel-wide flag and letting `assertMemoryAvailable` consult `this.deps.adapter` alone looks like the whole fix.

No leak — no tools attached means no access. It's just the one case the guard exists to make loud.

## Smaller

- **Transcript continuity isn't in Impact.** `transcript:user:${id}` keeps its prefix, but the id source moved from `store.identity` → `userKey` to `identifyUser` → `ApplicationUser.id` (`packages/channels-core/src/transcripts.ts:18`). Deployments where those two resolvers disagree silently orphan existing transcripts. Small blast radius (Channels are ~a week old and Intelligence-only), but it's a one-line addition to a section that already lists the cut.
- **`resolveWebMemory` bare `catch {}`** (`packages/runtime/src/v2/runtime/handlers/shared/memory-policy.ts:39`) returns 500 with no log, while its sibling `resolveIntelligenceUser` `console.error`s. An operator debugging their own `memory.access` callback gets nothing back.
- `packages/channels-core/src/identity.ts:85` — the `value === undefined` subclause is unreachable behind `typeof value !== "object"`.

## Verified correct (so you know what was covered)

- **Grant header contract matches in both directions.** The SDK always sends `Required<MemoryGrant>`; Intelligence's zod is `.strict()` with both keys required. Project-only correctly omits `x-cpki-user-id`, and `resolveMcpMemoryIdentity` accepts that only when `user: "none"`.
- **`kv/consume`**: `{ value: T | null }` → `undefined` normalization lines up with app-api, and `isRuntimeChannelsRequest`'s regex was updated to admit the new path — an easy one to forget.
- **Legacy path preserved**: no `memory` config → no grant header → Intelligence defaults to `FULL_MEMORY_GRANT`, matching pre-PR behavior. Same for `exposeMemoryRoutes` / `enableEnterpriseLearning`.
- **Project-only realtime** (`packages/core/src/memory.ts`): the `primaryTopic` swap avoids double-joining the project topic, and `credentialsSucceeded` is only dispatched when at least one cred pair validates, so the `!` assertions hold.
- **Continuation one-use**: `claimContinuation`'s `consume` is atomic in all three store impls, and the conformance suite covers it. Adding `kv.consume` as a required `StateStore` member only affects third-party custom stores — there are no first-party Postgres/Redis impls in the tree.

## Nit

The PR body links "Intelligence counterpart: #706" — that's the `main`-based twin. The counterpart of this stacked PR is **#708**. (#708 has the mirror-image mistake, pointing at #6282.) Worth fixing so a reviewer doesn't land on a variant with a different base.

BenTaylorDev (approved):
Approving. Full pass over the identity/Memory cut, the continuation binding, and the cross-repo contract against CopilotKit/Intelligence#708 — the grant header, kv/consume, and legacy-default paths all line up in both directions, and the one-use continuation holds across all three store impls.

Taking the direct-adapter Memory guard finding above as a non-blocking follow-up: it's fail-closed (no tools attached means no access), so it costs a confusing silence, not a leak. Worth the one-line fix on the way through.

Issue #2716, opened 2025-11-13
🐛 Bug: Error handlers missing request headers (and more)


hejtmii:
### ♻️ Reproduction Steps

1. Setup CopilotKit with local runtime
2. Configure `onError` in `CopilotRuntime` instance
3. Add some `headers` to the request
4. Simulate a situation when an error is thrown
5. `onError` doesn't provide request headers in `errorEvent.context.request.headers`

We need it to pass our own context data (mainly user ID) to the error handler, to be able to connect the error log to actions of particular user

### ✅ Expected Behavior

`onError` gets all possible data (or at least `headers`)  it exposes in its contract from the request

### ❌ Actual Behavior

`onError` handler is missing request data, primarily `headers`

### 𝌚 CopilotKit Version

```shell
1.10.6
```

### 📄 Logs (Optional)

```shell

```

BenTaylorDev:
The type slot for this already exists — `CopilotRequestContext.request.headers` is defined — but it's only populated on the client path; the runtime `onError` path never fills it in, so you see it empty. So this is a small wire-up rather than a net-new API. Keeping open, scoped to populating `request.headers` in the runtime `onError` context.

BenTaylorDev:
This is fixed — closing.

`request.headers` is now populated in the runtime error context:

- `23a108a38a` (2026-08-01) — *populate request headers in runtime error context*
- `3a9eda8441` (2026-08-05) — *redact sensitive headers from runtime error context*

On current `main`, `runtime-error-reporter.ts` fills in `headers`, `method`, `url`, `path`, and `startTime` on `errorEvent.context.request`. **You're on 1.10.6 and this covers you** — the same change wires the reporter into the v1 `CopilotRuntime` constructor, so a v1 `onError` goes through the same path.

One thing to know before you wire up your user-ID correlation. Headers are filtered through a **denylist**, so your own custom headers come through untouched. These seven are redacted:

`authorization` · `proxy-authorization` · `cookie` · `set-cookie` · `x-api-key` · `api-key` · `x-copilotcloud-public-api-key`

So a header like `x-user-id` reaches your handler fine. But if you were planning to derive the user from a JWT in `Authorization`, that one is stripped — pass the identifier in a dedicated header instead.

Thanks for the clear report, @hejtmii, and for the patience — this sat longer than it should have. Credit to @rodboev for the fix.

PR #6287, opened 2026-08-01
fix(runtime): populate request headers in runtime error context

rodboev:
## Summary

`CopilotRuntime` exposes an `onError` callback, but the v1-to-v2 delegation drops it before server request processing. Runtime failures therefore can't provide the incoming request headers that applications use to correlate errors with a user or tenant.

## Root cause

The legacy constructor retains the callback type, while delegated runtime options omit it. Common, SSE, and Intelligence handlers consume failures inside their own boundaries, before a generic endpoint hook can reconstruct the legacy event.

## Changes

- Add one internal runtime reporter that snapshots incoming Fetch request headers.
- Attach the configured legacy callback to the delegated runtime instance.
- Route common, SSE, and Intelligence agent-run failures, including standard `RUN_ERROR` events, through that reporter exactly once.
- Redact sensitive request headers (authorization, proxy-authorization, cookie, set-cookie, x-api-key, api-key, and the CopilotKit public-key header) before they reach the `onError` event.
- Preserve responses, stream close behavior, telemetry, cleanup, logging, endpoint hooks, and callback isolation.
- Add production-path regressions.

## Compatibility

The existing `CopilotErrorEvent` type and optional `context.request.headers` field remain unchanged. Header names and values come from the failing request's Fetch `Headers` object, with sensitive credentials stripped before the event is emitted. The callback receives a fresh record, so mutation cannot alter the request or a later event.

## Out of scope

React provider propagation, Chat, CopilotMessages, the deprecated runtime-client hook, redaction policy, HTTP response exposure, v2 endpoint-hook semantics, and non-agent runtime routes remain outside this slice.

## Related PRs and Issues

Addresses #2716.

Scope follows https://github.com/CopilotKit/CopilotKit/issues/2716#issuecomment-5086936254. Runtime error-routing precedent: https://github.com/CopilotKit/CopilotKit/pull/2143.

## Test plan

- [x] Runtime error regression and reporter tests, 12 + 4 tests passed. Covers public routing, sanitized header propagation with credential redaction, callback rejection containment, snapshots, mutation isolation, and malformed requests.
- [x] SSE and Intelligence telemetry tests, 7 + 14 tests passed. Covers agent-run failures, `RUN_ERROR`, setup boundaries, exact-once reporting, telemetry, cleanup, and response preservation.
- [x] Runtime preservation suites, 57 + 129 + 56 tests passed. Existing request, endpoint-hook, and runtime-library behavior remains intact.
- [x] Typecheck, formatting, and lint passed on changed runtime files; lint reported four pre-existing warnings.

### DIFF
--- a/.changeset/populate-runtime-error-request-headers.md
+++ b/.changeset/populate-runtime-error-request-headers.md
@@ -1,5 +1,0 @@
----
-"@copilotkit/runtime": patch
----
-
-fix(runtime): populate request headers in runtime error context
--- a/.changeset/populate-runtime-error-request-headers.md
+++ b/.changeset/populate-runtime-error-request-headers.md
@@ -1,0 +1,5 @@
+---
+"@copilotkit/runtime": patch
+---
+
+fix(runtime): populate request headers in runtime error context
--- a/packages/runtime/src/lib/runtime/__tests__/copilot-runtime-error.test.ts
+++ b/packages/runtime/src/lib/runtime/__tests__/copilot-runtime-error.test.ts
@@ -243,9 +243,10 @@ describe("CopilotRuntime onError types", () => {
   it("preserves the response when the rejecting runtime error handler runs", async () => {
     const errorSpy = vi.spyOn(console, "error").mockImplementation(() => {});
     try {
+      const onError = vi.fn().mockRejectedValue(new Error("callback failed"));
       const runtime = new CopilotRuntime({
         agents: Promise.reject(new Error("agent lookup failed")),
-        onError: vi.fn().mockRejectedValue(new Error("callback failed")),
+        onError,
       });
       const handler = createCopilotRuntimeHandler({
         runtime: runtime.instance,
@@ -263,6 +264,13 @@ describe("CopilotRuntime onError types", () => {
         error: "Failed to run agent",
         message: "agent lookup failed",
       });
+      expect(onError).toHaveBeenCalledOnce();
+      await vi.waitFor(() =>
+        expect(errorSpy).toHaveBeenCalledWith(
+          "CopilotRuntime onError reporting failed:",
+          expect.any(Error),
+        ),
+      );
     } finally {
       errorSpy.mockRestore();
     }
--- a/packages/runtime/src/lib/runtime/__tests__/copilot-runtime-error.test.ts
+++ b/packages/runtime/src/lib/runtime/__tests__/copilot-runtime-error.test.ts
@@ -1,8 +1,13 @@
-import {
+import type {
   CopilotErrorEvent,
   CopilotRequestContext,
   CopilotErrorHandler,
 } from "@copilotkit/shared";
+import { Observable } from "rxjs";
+import { HttpAgent } from "@ag-ui/client";
+import type { BaseEvent } from "@ag-ui/client";
+import { createCopilotRuntimeHandler } from "../../../v2/runtime";
+import { CopilotRuntime } from "../copilot-runtime";
 
 describe("CopilotRuntime onError types", () => {
   it("should have correct CopilotTraceEvent type structure", () => {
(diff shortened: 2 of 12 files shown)


BenTaylorDev (changes requested):
The reporter design is good — the `Symbol`-keyed attachment threads the handler without widening the public options type, `report()` is defensively wrapped so a throwing or rejecting handler can't take down a run, and the phase enum gives real triage value. Runtime-side error reporting is worth having. One change needed before it lands.

## Request headers are captured unredacted

`runtime-error-reporter.ts`:

```js
headers: Object.fromEntries(request.headers.entries()),
```

That puts every inbound header into the `CopilotErrorEvent` — `Authorization: Bearer …`, `Cookie`, `x-api-key`, and the CopilotKit public-key header included — and `report()` hands the event straight to the caller's `onError` with no sanitization in between.

To be precise about the exposure: this doesn't leave the process on its own. It goes to the application's handler, and where it ends up depends on what the developer does with it. But `onError` is commonly wired to Sentry, Datadog, or PostHog, and a framework can't assume the sink scrubs — so handing raw credentials across that boundary is an unsafe default even when it's often harmless in practice.

What convinced me this isn't just my preference: `context.request.headers` has existed in `CopilotRequestContext` for a while, and **nothing in the repo populates it.** Two places build this exact structure and both deliberately stop short of headers:

- `packages/react-core/src/hooks/use-copilot-runtime-client.ts:44-48` — sets `operation`, `url`, `startTime`
- `packages/runtime/src/lib/runtime/__tests__/copilot-runtime-error.test.ts:36-41` — the canonical type-structure test, sets `operation`, `method`, `url`, `startTime`

There's also no redaction helper anywhere to reuse (`lib/logger.ts` redacts only `pid`/`hostname`), so this would be establishing the pattern rather than following one.

**The ask:** filter before assigning — either an allow-list of useful diagnostic headers (`content-type`, `user-agent`, `x-request-id`, and similar) or a deny-list covering at minimum `authorization`, `cookie`, `set-cookie`, `x-api-key`. Either is fine by me; the allow-list fails safe as new header names appear.

One knock-on: `runtime-error-reporter.test.ts:55` asserts `context.request.headers` `toEqual` the full supplied map, so it currently pins the unredacted behavior. That assertion should flip to proving redaction — otherwise a later attempt to add filtering will look like a regression.

## Also

Please drop `.changeset/populate-runtime-error-request-headers.md` — releases go through `nx release` and there are no files under `.changeset/` on `main`. (Same note I left on #6291; your #6294 gets this right and says so in its description.)

Nothing else blocking. Happy to re-review as soon as the header filtering is in.

rodboev:
Thanks for the close read, and for the note that `context.request.headers` had never been populated anywhere. Filtering it before it crosses the `onError` boundary is the right call.

Handled both:

1. **Header redaction.** `runtime-error-reporter.ts` now routes the request headers through a `sanitizeRequestHeaders` helper backed by a deny-list (authorization, cookie, set-cookie, x-api-key). Authorization and cookie values stay in the request for the handler's own use, but never reach the `CopilotErrorEvent`. I went with a deny-list over an allow-list so `x-request-id`, `user-agent`, and `content-type` still surface for error correlation. The reporter test now carries `Authorization` and `Cookie` on the request and asserts they are absent from the emitted event while the safe headers come through, and the mutation-isolation assertion still holds.
2. **Changeset removed.** Dropped `.changeset/populate-runtime-error-request-headers.md`. Also removed files under `.changeset/` from all other open PRs.

Re-review whenever you're ready.

BenTaylorDev (approved):
Verified the redaction and the changeset removal. Both land as asked.

`runtime-error-reporter.ts:37-54` strips `authorization`, `proxy-authorization`, `cookie`, `set-cookie`, `x-api-key`, `api-key`, and the CopilotKit public-key header before the event is built, and the test now proves it rather than pinning the old behavior — the `toEqual` fixes the surviving set to exactly the two safe headers, the per-name `not.toHaveProperty` assertions cover each credential, and asserting `request.headers.get("authorization")` still resolves shows the redaction is a copy, not a mutation of the request. Deny-list over allow-list is the right call given you want `x-request-id` and `user-agent` for correlation.

CI is green across the board now that it's been kicked off — full unit matrix, `check-types`, `oxlint`, `format`, `package-quality`, and every integration suite. I also merged the branch onto current `main` locally (clean, no conflicts) and ran the `packages/runtime` suite there: 2017 passed, and the only failures were 10 that fail identically on unmodified `main` in my environment. The end-to-end tests in `copilot-runtime-error.test.ts` are what sold me: they drive a real `Request` through `createCopilotRuntimeHandler`, so the header population is demonstrated on the production path rather than against a stubbed reporter.

Two follow-ups, neither blocking:

1. A correction to my earlier review — I said there was no redaction helper to reuse. There is one: `handlers/header-utils.ts` carries `DEFAULT_DENY_HEADER_NAMES` / `DEFAULT_DENY_HEADER_PREFIXES` for inbound-header *forwarding*. It isn't reusable wholesale here, and shouldn't be: it forwards `authorization` by design and denies `x-request-id`, which is the opposite of what an error context wants. But its `x-copilotcloud-` *prefix* rule is worth mirroring in your set, so a future platform header can't leak into `onError` just because the exact name isn't listed. Fine as a later change.
2. `sse/connect.ts` and `handle-suggest.ts` also build responses through `createSseEventResponse` without a reporter, so `agent/connect` and `agent/suggest` failures still bypass `onError`. That matches the scope you declared, just noting where the next slice would go.

Issue #4634, opened 2026-05-04
🐛 Bug: RUN_STARTED / AG-UI: no standard way to correlate server trace (OTel, Langfuse) with browser for user feedback


mustafaasif2:
### ♻️ Reproduction Steps

1. Integrate a CopilotKit + AG-UI agent (e.g. Mastra) with **server-side** OpenTelemetry / Langfuse so each run has a **trace id** in the backend.
2. In the **browser**, use the standard AG-UI / CopilotKit v2 path (`useAgent` + `agent.subscribe` or equivalent) and handle lifecycle events.
3. Observe `RUN_STARTED` (or the equivalent lifecycle): the payload includes **`threadId`** and **`runId`**, but **no field** that matches the **Langfuse / OTel root trace id** used for that run (unless the host app invents a convention).
4. To submit **client-side** feedback (e.g. Langfuse Web SDK `score({ traceId })`) against the **same** trace as the server, the app must:
   - **Derive** a trace id from `runId` (e.g. UUID → 32-char hex) **and** force the **same** id on the server (`tracingOptions.traceId`), **or**
   - Patch transport/events (e.g. SSE / extended payloads), **or**
   - Give up and send feedback only from the server.
There is no single documented, portable mechanism in the AG-UI event contract for “trace id used by observability for this run.”

### ✅ Expected Behavior

- AG-UI documents **how** browser-side tools should correlate with **server** traces when both sides participate (Langfuse, OTel).
- Preferably, **`RUN_STARTED`** (or another lifecycle event) carries an **optional, standardized field** for this — e.g. `observabilityTraceId` / `externalTraceId` — when the runtime can provide it; OR the spec explicitly recommends **`CUSTOM`** events / passthrough fields with a **registered name** for observability metadata.
- CopilotKit runtime (when applicable) **populates** that field when trace context is known, so apps do not rely on **private** `runId`→hex hacks across stacks.

### ❌ Actual Behavior

- **`RUN_STARTED`** exposes **`threadId`** and **`runId`** but **not** the exporter’s trace id as part of the **public** AG-UI contract.
- Application authors use **ad-hoc conventions**: derive a hex id from `runId`, monkey-patch agent **`tracingOptions.traceId`**, or middleware on SSE — none of which are **standardized** or **portable** across AG-UI hosts (Mastra, LangGraph, etc.).
- Products like Langfuse expect **`traceId`** on the client for **`score()`**; without a shared contract, every integration reinvents the same glue code.

[...]

Issue #3039, opened 2026-01-20
🚀 Feature Request: Expose AG-UI rawEvent field to frontend Message objects


xegulon:
Related to #2678 - I'm facing a similar challenge when implementing user feedback (thumbs up/down) with Langfuse observability.

## The problem

AG-UI events have a `rawEvent` field (defined in `@ag-ui/core` BaseEventSchema) that allows passing custom metadata:

```typescript
// @ag-ui/core - BaseEventSchema
{
    type: EventType;
    timestamp?: number;
    rawEvent?: any;  // ← This field exists on AG-UI events
}
```

However, when CopilotKit converts AG-UI events into frontend `Message` objects, this `rawEvent` field is **not propagated**. The GraphQL schema for `BaseMessageOutput` only includes:

```typescript
type BaseMessageOutput = {
    createdAt: string;
    id: string;
    status: MessageStatus;
    // ← rawEvent is lost!
};
```

## My use case

I'm using Google ADK with Langfuse for LLM observability. When a user clicks thumbs up/down via `onThumbsUp`/`onThumbsDown`, I need to correlate the message to the Langfuse trace ID to submit feedback scores.

Currently, I inject the Langfuse `trace_id` into the AG-UI event's `rawEvent` field on the backend:

```python
# Backend (Python)
event = event.model_copy(update={
    "raw_event": {"langfuse_trace_id": trace_id}
})
```

But on the frontend, I cannot access this data because `message.rawEvent` doesn't exist.

## Proposed solution

Expose the `rawEvent` field from AG-UI events to the frontend `Message` type, so it can be accessed in callbacks like `onThumbsUp`:

```typescript
onThumbsUp={(message) => {
    const traceId = message.rawEvent?.langfuse_trace_id;
    // Submit feedback to Langfuse
}}
```

## Workaround I'm currently using

I'm forced to replace the `message_id` with the `trace_id` on the backend, which works but feels hacky and could cause issues if CopilotKit expects unique message IDs.

## Benefits

- Enables custom metadata propagation from backend to frontend
- Solves the feedback/observability correlation problem mentioned in #2678
- Maintains backward compatibility (rawEvent is optional)
- No breaking changes to existing code

Thank you for considering this!

BenTaylorDev:
[...]

PR #6289, opened 2026-08-01
feat(react-core): expose AG-UI raw event to feedback callbacks

rodboev:
## Summary

React v2 feedback callbacks receive an assistant message without the trace metadata carried by the direct AG-UI event that created it. This slice exposes that metadata to thumbs callbacks without changing canonical messages or future run inputs.

## Root cause

AG-UI keeps `rawEvent` on events while reducer-created assistant messages remain protocol-clean. `StateManager` sees the direct start event but previously discarded its correlation before `CopilotChatMessageView` forwarded the message to feedback callbacks.

## Changes

- Store defined direct `TEXT_MESSAGE_START.rawEvent` metadata by agent, thread, and message.
- Replace repeated scoped entries and prune them with message removal and lifecycle cleanup.
- Return a cloned sidecar value through `CopilotKitCore.getRawEventForMessage`.
- Enrich only thumbs-up and thumbs-down callback arguments at click time across flat and virtualized rendering.
- Add production-path regressions and document the callback-only type.

## Out of scope

Canonical messages, future `RunAgentInput.messages`, render props, message identity, stream ordering, snapshots, transformed chunks, persistence, GraphQL, legacy React, Vue, Angular, and standardized trace semantics remain outside this slice.

## Related PRs and Issues

Addresses #3039.

The callback-only scope follows https://github.com/CopilotKit/CopilotKit/issues/3039#issuecomment-5086936452. Related trace-correlation contract: #4634.

## Test plan

- [x] StateManager sidecar tests, 10 passed. Covers direct capture, falsey values, replacement, scope isolation, cleanup, snapshots, and chunks.
- [x] React v2 feedback tests, 4 passed. Covers real callback routing, canonical and outbound cleanliness, render identity, and flat/virtual paths.
- [x] Full package suites, 625 core tests, 1,475 React Core tests, and 2 script tests passed.
- [x] Typecheck, formatting, lint, and whitespace validation passed; lint reported five pre-existing warnings.

## Notes

The clean-base behavioral half of the reproduction remains unproved because temporary worktree setup hung behind unrelated Git processes. The PR makes no base execution claim for that half.

### DIFF
--- a/.changeset/expose-feedback-raw-event.md
+++ b/.changeset/expose-feedback-raw-event.md
@@ -1,0 +1,6 @@
+---
+"@copilotkit/core": minor
+"@copilotkit/react-core": minor
+---
+
+feat(react-core): expose raw event metadata to feedback callbacks
--- a/packages/core/src/__tests__/state-manager-raw-event.test.ts
+++ b/packages/core/src/__tests__/state-manager-raw-event.test.ts
@@ -116,7 +116,7 @@ describe("StateManager direct text-start raw event sidecar", () => {
   });
 
   it("captures replacements and preserves state and message mappings on unsubscribe", async () => {
-    let rawEvent = { version: 1 };
+    let rawEvent: unknown = { version: 1 };
     const agent = new RawEventAgent((input) => {
       const events = startAndFinish(input, "replace-message", rawEvent);
       return [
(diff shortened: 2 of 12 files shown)


BenTaylorDev (approved):
Approving. Checked the retention behavior in particular, since holding provider payloads per message is the part that could go wrong.

It's tightly scoped: only `TEXT_MESSAGE_START` is recorded, and only when the provider actually supplied `event.rawEvent` — when it doesn't, any prior entry is deleted rather than left behind. `pruneRawEvents` diffs the retained ids against the live message list, deletes anything absent, and collapses empty thread and agent maps; `clearAgentState` and `unsubscribeFromAgent` both drop the agent's map. So nothing accumulates beyond messages currently in history.

`unknown` is also the right type here. `rawEvent` is a provider-shaped passthrough — a raw OpenAI or Anthropic chunk — and there's no honest type to give it.

**Verified:** `@copilotkit/core` at 59 files / 626 tests (baseline 58 / 616, so +1 file and +10 tests), `@copilotkit/react-core` at 123 / 1475, and `core:build` clean. Fork PRs get no automatic CI here, so this was the first run.

**One interaction worth a follow-up,** not blocking: pruning runs on `onMessagesSnapshotEvent`, so a snapshot that omits a message drops its raw event permanently — including the stale-snapshot case #6294 is fixing, where the message itself comes back but its raw event won't.

**One small ask:** the new `onMessagesChanged` handler prunes only when `input` is absent. The reasoning isn't obvious from the code — a one-line comment on why pruning is skipped mid-run would help whoever reads it next.

Issue #5930, opened 2026-07-12
Feature: MCP Apps per-tool include/exclude filter (beyond agentId/server scoping)


likebean:
## Summary

`mcpApps` can scope servers by `agentId`, but there is no way to expose only a **subset of tools** from an attached MCP Apps server. For multi-tenant agent platforms this is a real gap.

## Problem

Today:

```ts
mcpApps: {
  servers: [
    {
      type: "http",
      url: "https://mcp.example.com/mcp",
      serverId: "crm",
      agentId: "support-agent", // ✅ filters by agent
      // ❌ no includeTools / excludeTools
    },
  ],
}
```

Runtime wiring (`configureAgentForRequest`) only filters servers by `agentId`, then `MCPAppsMiddleware` does `tools/list` and injects **all** tools that have `_meta["ui/resourceUri"]`.

Relevant shapes today:

- `McpAppsServerConfig = MCPClientConfig & { agentId?: string }`
- `MCPAppsMiddlewareConfig = { mcpServers?: MCPClientConfig[] }`
- `MCPClientConfig` = `{ type, url, serverId?, headers? }` — no tool allowlist

So if one MCP server exposes both `generate_qr` and `admin_dashboard` UI tools, attaching the server for an agent that should only use `generate_qr` still makes **both** visible to the LLM.

## Why this matters

Agent platforms (Admin UI → agent selects tools → CopilotKit Runtime) typically:

1. Register one MCP server that may mix UI and non-UI tools
2. Let each agent select **specific** tools
3. Route UI tools through CopilotKit MCP Apps, plain tools through the agent backend (e.g. Agno `MCPTools.include_tools`)

Backend toolkits already support include/exclude (e.g. Agno). CopilotKit can scope by **agent × server**, but not **agent × tool**, so the Chat/Runtime side cannot match the backend permission model.

Codex-style hosts already have `enabled_tools` / allowlists per MCP server; this would align CopilotKit with that expectation.

## Proposed API

Extend server config, e.g.:

```ts
type McpAppsServerConfig = MCPClientConfig & {
  agentId?: string;
  /** If set, only these tool names are injected (allowlist). */
  includeTools?: string[];
  /** Optional denylist applied after includeTools (or after all UI tools). */
  excludeTools?: string[];
};
```

Behavior in `MCPAppsMiddleware.fetchToolsFromServer` (conceptual):

1. Keep current filter: only tools with UI resource meta
2. If `includeTools` is set → keep only those names
[...]

PR #6292, opened 2026-08-01
fix(runtime): reject unenforceable mcpApps tool policy instead of silently ignoring it

rodboev:
## Summary

`mcpApps.servers` entries that carry `includeTools` or `excludeTools` are currently accepted even though the pinned `@ag-ui/mcp-apps-middleware` package has no option for them. The runtime then ignores the keys, so tools an operator intended to restrict remain available. This change rejects that configuration instead of allowing a silent no-op.

## What CopilotKit owns

- `mcpApps.servers` configuration and `agentId` scoping.
- Projection of selected servers into `MCPAppsMiddleware`.
- Reporting unsupported configuration before middleware construction.

Discovery, model-emitted tool execution, frontend-proxied execution, server identity, and tool provenance belong to `@ag-ui/mcp-apps-middleware`.

## Changes

- Extract the server projection into `resolveMcpAppsServers`, which scans all configured entries for defined policy keys, filters by `agentId`, strips only `agentId`, and forwards other fields unchanged.
- Return a configuration error naming the unsupported key, server, pinned middleware version, owning package, and issue when a policy key is supplied.
- Add tests for agent scoping, field forwarding, malformed and empty values, undefined spread values, constructor avoidance, and the existing HTTP error path.
- Document the ownership boundary and add a runtime changeset.

## Why the filter stays external

The pinned package is version `0.0.3`. It owns the private server maps, UI-tool discovery, model-emitted execution, and frontend proxy execution. A CopilotKit middleware could observe only one of those paths and would have to duplicate private server identity and tool provenance. The complete `includeTools` and `excludeTools` implementation belongs in the external package, where one predicate can cover discovery and both execution paths.

## Current behavior

Plain JavaScript or JSON configuration can supply `excludeTools: ["delete_account"]` without a TypeScript excess-property check. The runtime currently accepts the configuration, constructs `MCPAppsMiddleware`, and leaves the tool available. The new behavior returns an HTTP 500 through the existing runtime error path, names the unsupported key and dependency, and does not construct the middleware.

## Follow-up

The counterpart change in `@ag-ui/mcp-apps-middleware` should add the fields to the per-server configuration, preserve absent versus empty include lists, resolve server identity through its existing maps, and apply one predicate after UI-resource discovery and before model-emitted and proxied tool execution. Once that version is released, CopilotKit can remove the rejection and pass the fields through unchanged.

## Related issue

Refs #5930.

The cross-repository ownership split follows the proposal in https://github.com/CopilotKit/CopilotKit/issues/5930#issuecomment-5128722524. This PR does not close the issue.

## Test plan

- [x] `pnpm -C packages/runtime exec vitest run src/v2/runtime/__tests__/mcp-apps-servers.test.ts src/v2/runtime/__tests__/mcp-apps-middleware-integration.test.ts` passed, 2 files and 20 tests
- [x] `pnpm -C packages/runtime exec vitest run` passed, 129 files and 1,836 tests
- [x] `pnpm exec nx run @copilotkit/runtime:check-types` passed
- [x] `pnpm exec oxlint` and `pnpm exec oxfmt --check` passed on changed TypeScript files
- [x] `pnpm check:plugin-skills` passed
- [ ] `CI green for static / quality and test / unit on Node 20, 22, and 24`

### DIFF
--- a/.changeset/enable-mcp-apps-tool-filters.md
+++ b/.changeset/enable-mcp-apps-tool-filters.md
@@ -1,0 +1,7 @@
+---
+"@copilotkit/runtime": minor
+---
+
+fix(runtime): reject unenforceable mcpApps tool policy instead of silently ignoring it
+
+Reject MCP Apps per-tool policy keys until the external middleware supports them, preventing silent configuration no-ops.
--- a/packages/runtime/skills/runtime/references/wiring-mcp-apps-middleware.md
+++ b/packages/runtime/skills/runtime/references/wiring-mcp-apps-middleware.md
@@ -49,6 +49,11 @@ export default { fetch: handler };
 Each server entry accepts an optional `agentId`. When set, the server's tools are only
 exposed to that agent. Omit it to expose to all agents.
 
+Per-tool filtering belongs to `@ag-ui/mcp-apps-middleware`. The runtime currently pins
+version `0.0.3`, which does not support `includeTools` or `excludeTools`. Supplying either
+key raises a configuration error instead of silently ignoring it; use a compatible
+middleware release when that upstream policy contract is available.
+
 ## Gotcha — do NOT put MCP under agents
 
 ```typescript
--- a/packages/runtime/src/v2/runtime/__tests__/mcp-apps-middleware-integration.test.ts
+++ b/packages/runtime/src/v2/runtime/__tests__/mcp-apps-middleware-integration.test.ts
@@ -1,13 +1,30 @@
 import { describe, it, expect, afterEach, vi } from "vitest";
-import {
-  AbstractAgent,
-  RunAgentInput,
-  BaseEvent,
-  EventType,
-} from "@ag-ui/client";
+import { AbstractAgent, EventType } from "@ag-ui/client";
+import type { BaseEvent, RunAgentInput } from "@ag-ui/client";
 import { Observable } from "rxjs";
 import { LLMock, MCPMock } from "@copilotkit/aimock";
 import { MCPAppsMiddleware, getServerHash } from "@ag-ui/mcp-apps-middleware";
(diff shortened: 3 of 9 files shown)


BenTaylorDev (approved):
Approving. Good change — it converts a silent, security-relevant gap into a loud one.

**Premise verified:** the installed `@ag-ui/mcp-apps-middleware@0.0.3` has zero occurrences of `includeTools`/`excludeTools` anywhere in its type surface, so those keys really are dropped on the floor today.

I also checked how risky the throw is, since turning a previously-working config into an error is the kind of thing that bites on upgrade. It's low: the keys were never documented (no hits anywhere in `showcase/` or `skills/` on `main`) and `McpAppsServerConfig` never declared them — which is why your code needs the `as unknown as Record<string, unknown>` cast to see them at all. So anyone carrying them guessed, or copied from another MCP wrapper, or was told to by an LLM. Small population — but for them the current behavior is that they believe tools are filtered and nothing is filtering. Rare and severe is exactly when throwing beats warning.

**Verified:** `packages/runtime` at 131 files / 1850 tests (baseline 130/1835). Fork PRs get no automatic CI here, so this was the first run.

**One note, and it's a simplification rather than an addition:** the validation runs inside `configureAgentForRequest`, so it re-checks on every request and a misconfigured runtime fails on first traffic rather than at startup. Validating once when the runtime is constructed would be less work per request and would surface the mistake at deploy time. Worth doing whenever this area is next touched — not a condition of merge.

Please drop `.changeset/enable-mcp-apps-tool-filters.md` before merge — releases go through `nx release` and there are no files under `.changeset/` on `main`. (Also, the filename says "enable" while the change rejects; worth a glance that the description matches the final intent.)

PR #6293, opened 2026-08-01
fix(runtime): initialize arrays before /- append in AGUISendStateDelta

### LINKED ISSUE #5998: 🐛 Bug:  OPERATION_PATH_CANNOT_ADD when LLM emits AGUISendStateDelta with /- append on uninitialized array

rodboev:
## Summary

`AGUISendStateDelta` can emit an array append against a state where the target array has not been initialized. The emitted patch then fails during event compaction with `OPERATION_PATH_CANNOT_ADD`.

This change keeps the authoritative state represented in emitted events, initializes a missing array immediately before its first `/-` append, and applies the same contract across the generic, AI SDK, and TanStack state-delta paths. Existing arrays and valid deltas retain their current contents and operation order.

Closes https://github.com/CopilotKit/CopilotKit/issues/5998

## Changes

- Emit the input state before the first delta when no earlier state event represents it
- Initialize a missing array before normalized classic, AI SDK, and TanStack state deltas append through `/-`; custom raw-event mode remains outside this normalizer
- Preserve populated arrays and already-valid patch operations
- Preserve caller-owned state when structured cloning falls back
- Cover missing-array reconstruction, sibling converters, failure guards, and existing-array preservation
- Follow the current `release.config.json` and Nx release convention; no Changeset file is added.

## Test plan

- [x] `pnpm -C packages/runtime exec vitest run src/agent/__tests__/state-tools.test.ts src/agent/__tests__/converter-aisdk.test.ts src/agent/__tests__/converter-tanstack.test.ts`
- [x] `pnpm -C packages/runtime exec vitest run`
- [x] `pnpm exec nx run @copilotkit/runtime:check-types`
- [x] `pnpm exec oxfmt --check` on changed runtime files
- [x] `pnpm exec oxlint` on changed runtime files, zero errors with two pre-existing no-shadow warnings
- [x] Verify no `.changeset` file is added because current main uses Nx release
- [x] Verify the final diff contains only the private helper, runtime implementation, sibling converters, and focused tests

### DIFF
--- a/packages/runtime/src/agent/__tests__/converter-aisdk.test.ts
+++ b/packages/runtime/src/agent/__tests__/converter-aisdk.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect } from "vitest";
-import { EventType } from "@ag-ui/client";
+import { compactEvents, EventType } from "@ag-ui/client";
 import {
   createAgent,
   createDefaultInput,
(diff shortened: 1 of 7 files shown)


BenTaylorDev (approved):
Approving. I verified the bug, the fix, and the surrounding behavior rather than taking the description's word for it.

**The bug is real and it's silent**, which is the part that makes this worth fixing. Scripting a `STATE_DELTA` against the pinned `@ag-ui/client` 0.0.57:

```
add /items/- with no /items    -> OPERATION_PATH_CANNOT_ADD, state stays {}, no RUN_ERROR, no throw
add /items/- with /items: []   -> works
your normalized form           -> works: {"items":["first"]}
```

A failed patch logs to console and the run continues with state quietly unchanged. That's worse than an error, and an LLM appending to a not-yet-existing array is not an exotic case.

**Completeness is right:** all three producers are wired identically — `converters/aisdk.ts`, `converters/tanstack.ts`, and `BuiltInAgent`. That's the usual place this kind of fix goes half-done.

**Verified:** `packages/runtime` at 130 files / 1850 tests (baseline 1835, so +15 from your additions), and `runtime:build` adds no type errors — 19 exist before and after, all pre-existing in `delivery-adapter`, none in `agent/`. `cloneState` uses `structuredClone` with a deep fallback, so the shadow state never mutates the caller's `input.state`, which was my main worry. Fork PRs don't get automatic CI, so this was the first run.

One thing for the description: `createStateEventNormalizer` also injects a synthetic `STATE_SNAPSHOT` before the first `STATE_DELTA` when `initialState` is defined. Defensible, since deltas need a base — but it's a bigger behavior change than the title implies and deserves a line so it isn't a surprise later.

PR #6296, opened 2026-08-01
fix(react-core): preserve logical run identity across HITL resolve

### LINKED ISSUE #3456: 🐛 Bug: runId changes after HITL resolve

rodboev:
## Summary

Preserve the logical run ID when a legacy `useCopilotAction({ renderAndWaitForResponse })` frontend tool resolves and `processAgentResult` starts its recursive follow-up.

The run handler binds each internal continuation handoff to the exact follow-up invocation, cancels it when setup fails or no run starts, and keeps the handoff out of the public `CopilotKitCore.runAgent` contract. The public regression drives the legacy hook through its `useHumanInTheLoop` and `useFrontendTool` path, renders the approval control, resolves it, and verifies both agent calls use the same generated ID.

Closes https://github.com/CopilotKit/CopilotKit/issues/3456

## Changes

- Preserve the originating ID across recursive frontend-tool follow-up runs
- Keep the existing legacy HITL registration and response behavior unchanged
- Add core follow-up coverage and a public `useCopilotAction` regression
- Retain the existing standard/legacy interrupt and StateManager coverage from the earlier fix

## Test plan

- [x] `pnpm -C packages/react-core exec vitest run src/hooks/__tests__/use-copilot-action.e2e.test.tsx`
- [x] `pnpm -C packages/core exec vitest run src/__tests__/core-follow-up.test.ts`
- [x] `pnpm -C packages/react-core exec vitest run src/v2/hooks/__tests__/use-interrupt.test.tsx`
- [x] `pnpm -C packages/core exec vitest run src/__tests__/state-manager.test.ts`, 39 tests passed
- [x] `pnpm -C packages/react-core exec vitest run`, 123 files and 1475 tests passed
- [x] `pnpm -C packages/core exec vitest run`, 58 files and 625 tests passed
- [x] `pnpm -C packages/core run check-types`
- [x] `pnpm -C packages/react-core run check-types`
- [x] `pnpm exec oxfmt --check` on all eight changed source/test files
- [x] `pnpm exec oxlint` on all eight changed source/test files, 5 pre-existing warnings and 0 errors

### DIFF
--- a/packages/core/src/__tests__/core-follow-up.test.ts
+++ b/packages/core/src/__tests__/core-follow-up.test.ts
@@ -1,6 +1,6 @@
 import { describe, it, expect, beforeEach, vi, afterEach } from "vitest";
 import { CopilotKitCore } from "../core";
-import { FrontendTool } from "../types";
+import type { FrontendTool } from "../types";
 import {
   MockAgent,
   createAssistantMessage,
@@ -51,6 +51,43 @@ describe("CopilotKitCore.runAgent - Follow-up Logic", () => {
 
     expect(agent.runAgentCalls).toHaveLength(2);
     expect(result.newMessages).toContain(followUpMessage);
+  });
+
+  it("should preserve the originating run ID through a frontend follow-up", async () => {
+    const tool = createTool({
+      name: "runIdTool",
+      handler: vi.fn(async () => "Result"),
+      followUp: true,
+    });
+    copilotKitCore.addTool(tool);
+
+    const agent = new MockAgent({
+      newMessages: [createToolCallMessage("runIdTool")],
+    });
+    copilotKitCore.addAgent__unsafe_dev_only({
+      id: "test",
+      agent: agent as any,
+    });
+    agent.runAgentCallback = (input) => {
+      if (agent.runAgentCalls.length === 2) {
+        agent.setNewMessages([createAssistantMessage({ content: "Done" })]);
+        expect(input.forwardedProps).toMatchObject({
+          __copilotkit_follow_up: true,
+        });
+      }
+      expect(input.runId).toBe("logical-hitl-run");
+    };
+
+    await copilotKitCore.runAgent({
+      agent: agent as any,
+      runId: "logical-hitl-run",
+    });
+
+    expect(agent.runAgentCalls).toHaveLength(2);
+    expect(agent.runAgentCalls.map((input) => input.runId)).toEqual([
+      "logical-hitl-run",
+      "logical-hitl-run",
+    ]);
   });
 
   it("should not trigger recursive call when tool.followUp is false", async () => {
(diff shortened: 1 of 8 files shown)


BenTaylorDev (approved):
Approving. I verified the mechanism rather than taking the description's word for it, and this holds up well.

**The design call is the right one.** Keeping the continuation marker in an in-memory `WeakMap` keyed by agent — rather than smuggling it through `forwardedProps` — means it never reaches a runtime and can't collide with user data. The comment says as much, and it's the part I'd have pushed back on had it gone the other way.

**What I checked:**

- **Lifecycle ordering.** `bind()` depends on `onRunInitialized` firing before `RUN_STARTED`, so I scripted an agent and confirmed the order is `onRunInitialized -> onRunStartedEvent -> onRunFinishedEvent -> onRunFinalized`. `expectedInput` is therefore always set before the marker is looked up, so the pending entry always gets matched and cancelled — no orphaned markers.
- **Cancellation coverage.** `cancel()` is wired into every exit: the `detachActiveRun` throw, the `!started` path, and the catch block. I couldn't find a route that leaves a marker active.
- **`isContinuation` misclassification.** I looked for a genuinely-new run that would be misread as a continuation. It would need to carry `forwardedProps.command.resume`, which is already semantically a resume, so I don't think the case exists.
- **Stale `interruptRunIdsRef`.** Cleared on both `RUN_STARTED` and `onRunFailed`; a miss degrades to "no runId", which is exactly current `main` behavior. No hang, no collision.
- **Tests.** `@copilotkit/core` 58 files / 626 tests pass; `use-interrupt.test.tsx` + `use-copilot-action.e2e.test.tsx` 35/35 pass with both halves applied and core rebuilt. `@copilotkit/core:build` succeeds. Worth noting fork PRs don't get automatic CI here, so this hadn't been exercised before now.

**One non-blocking nit for a follow-up:** `markNextRunAsContinuation(agent, _expectedRunId?)` never reads its second parameter, and it's declared in `CopilotKitCoreFriendsAccess` as `expectedRunId?: string`. Either use it to validate the handoff or drop it from both signatures before it settles into the API.

**Sequencing note for anyone reading this later:** this touches the same `runFinished && input.runId === subRunId` heuristic that #6294 rewrites, so those two conflict in `run-handler.ts` and `state-manager.ts`. Landing this first means #6294 gets rebased onto it — which is the right order, since #6294 has a separate revision pending.

PR #6298, opened 2026-08-01
feat(skills): make managed Intelligence the default setup path, add a Channels skill (refs OSS-705)

BenTaylorDev:
Companion to CopilotKit/Intelligence#714 (OSS-705). That PR builds the `copilotkit channels` CLI; this one covers the skills and docs half of the same [PRD](https://app.notion.com/p/3af3aa381852810b8254ec0cbb5be6af).

## Managed Intelligence becomes the default path in `copilotkit-setup`

The most-used "add CopilotKit to your project" path walked every new user into the self-hosted SSE runtime and never offered the managed one. `CopilotIntelligenceRuntime`, `CopilotKitIntelligence`, the required `identifyUser`, and the hosted environment values all appeared in this skill's *reference* files but were wired by **no step** — so the skill could describe managed Intelligence without ever producing it.

- **Step 2 now chooses the runtime mode before any runtime code is written**, because the mode changes how the runtime is constructed and retrofitting it means rewriting the file. Managed is the recommended default and has real wiring.
- **Self-hosted SSE stays fully documented** as a deliberate opt-out, with its prerequisites and its tradeoff stated at the point of choice. The OSS packages are published and MIT-licensed, so obscuring the alternative would not prevent its use and would cost credibility on everything around it.
- **Step 6 becomes the actual Intelligence step** rather than a telemetry aside, and separates the two credentials that setup mistakes conflate: the server-side project API key (a secret, and never `NEXT_PUBLIC_`-prefixed) and the public license key (a project identifier meant to reach the client).
- **Fixes a command that does not exist.** Both the skill and `references/telemetry-setup.md` instructed `npx copilotkit auth`. The command is `login`, and `project select` is what provisions the project.

## New `copilotkit-channels` skill

Covers the code half: the Channel declaration, the long-running host requirement, and which mounts start activation on their own versus which wait for an explicit `channels.ready()`.

It leads with the managed-versus-self-hosted boundary, because both product families use the words "channels" and "Slack" and the OSS demos ship their own Slack manifest and Teams package.

## Docs

- `channels/intelligence.mdx` — the CLI as a **peer** path to the wizard, with the tradeoff stated. The wizard walkthrough is untouched, and the docs say explicitly that either path can finish what the other started.
- `packages/channels-{slack,teams}/README.md` — directional pointers: lead with what managed provides, route there, state plainly that the self-hosted adapter remains supported.

## Two deviations from the PRD, both deliberate

1. **Pointers are scoped to Slack and Teams.** The PRD asks for all eight `packages/channels-*/README.md`. Managed Channels supports only those two providers, so the same pointer in `channels-discord`, `-telegram`, or `-whatsapp` would route a reader to something that cannot serve them.
2. **`frontends/{slack,teams}.mdx` need no pointer.** The PRD lists them as describing the OSS adapter product; both already document the *managed* path ("CopilotKit Intelligence holds the Slack credentials") and both already link to the configuration page, which now offers the CLI.

## A correction worth reviewing

The Channels skill first asserted that activation is lazy on every host and that `await listener.channels.ready()` is always required — wrong, and wrong in its own Step 4 example, which uses `createCopilotNodeListener`. OSS-641 split the behavior: node and express start activation at creation; hono and the generic fetch handler still defer. The skill now carries the table and says to check the mount, because telling a Node host to add a call it does not need is as unhelpful as omitting one that is required.

`showcase/.../deploy-and-operate.mdx` already described this correctly and is unchanged.

## Notes for review

- A standalone skill **must** be registered in `RESERVED_LIFECYCLE_SLUGS` (`scripts/sync-plugin-skills.ts`) or `pnpm sync:plugin-skills` deletes it as an orphan. Registered, and the test now pins that requirement with the reason.
- `pnpm check:plugin-skills` passes; `scripts/__tests__/sync-plugin-skills.test.ts` passes (50 files / 483 tests).
- The last commit used `--no-verify`, disclosed in its message: touching `packages/*/README.md` makes 13 projects affected and the pre-commit hook aborts that run with `exit 130`. The same target set (`test`, `publint`, `attw` for `@copilotkit/channels-intelligence`) passes standalone. That commit is two markdown files.

### DIFF
--- a/packages/channels-slack/README.md
+++ b/packages/channels-slack/README.md
@@ -14,6 +14,20 @@ Channel itself only runs inside a CopilotKit Intelligence-configured
 standalone / DIY runner and no `channel.start()`; the runtime starts and owns
 the channel because Intelligence is configured.
 
+## Managed Channels: the alternative to holding your own credentials
+
+This adapter is the **self-hosted** path: your process holds the Slack credentials, runs the Slack ingress, and talks to Slack directly.
+
+**Managed Intelligence Channels** is the alternative. Intelligence owns the provider edge — signed ingress, egress, and encrypted credential storage — so your process holds no Slack credentials and exposes no public Slack endpoint. You also get durable threads, the Channels dashboard with per-Channel health and transcripts, and guided provider setup from either the browser wizard or the CLI:
+
+```bash
+npx copilotkit channels add support
+```
+
+Your bot code is otherwise identical — the agent, tools, context, commands, and turn handlers do not change. Only the transport does. See `examples/slack/app/managed.ts` for the same bot wired both ways, and the **copilotkit-channels** skill for the runtime wiring.
+
+This self-hosted adapter remains fully supported. Choose it when you want the provider connection inside your own infrastructure.
+
 ## Install
 
 ```sh
--- a/packages/channels-teams/README.md
+++ b/packages/channels-teams/README.md
@@ -16,6 +16,20 @@ Channel itself only runs inside a CopilotKit Intelligence-configured
 standalone / DIY runner and no `channel.start()`; the runtime starts and owns
 the channel because Intelligence is configured.
 
+## Managed Channels: the alternative to holding your own credentials
+
+This adapter is the **self-hosted** path: your process holds the Microsoft Teams credentials, runs the Microsoft Teams ingress, and talks to Microsoft Teams directly.
+
+**Managed Intelligence Channels** is the alternative. Intelligence owns the provider edge — signed ingress, egress, and encrypted credential storage — so your process holds no Microsoft Teams credentials and exposes no public Microsoft Teams endpoint. You also get durable threads, the Channels dashboard with per-Channel health and transcripts, and guided provider setup from either the browser wizard or the CLI:
+
+```bash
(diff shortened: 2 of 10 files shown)


BenTaylorDev:
Context update — this PR still ships as-is.

The `init` flow is being reshaped since this was opened: instead of offering a Channel as an add-on inside the web-app flow, `init` will branch early on "web app or channel". Design: https://app.notion.com/p/3b03aa38185281ab8183c7fa2abd564f (amends §13 of the CLI Channel Setup PRD).

That does not change anything here. Two follow-ups that touch this branch specifically:

1. **`skills/copilotkit-channels/SKILL.md` needs rescoping.** #6315 makes every starter ship a working `channel-host.mts`, so for a scaffolded project the skill's job becomes *customising* a Channel (tools, context, commands, per-provider extras) rather than hand-writing the host. The hand-wiring path should stay for projects not built from a starter. That file is introduced by this PR, so the change belongs here rather than in #6315 — a rescoped draft was prepared and set aside rather than committed to the other branch, precisely to avoid a merge conflict on a brand-new file.
2. Nothing else — the mode choice and Step 6 rewrite are unaffected.

AlemTuzlak (approved):

PR #6304, opened 2026-08-02

deepshekhardas:
Port of #5433.


BenTaylorDev:
Closing — this fix is already on `main` as `packages/core/src/agent.ts:176-182`, landed in July 2026 as `aad69ca12e` (#5433). The added `requestInit` override is identical to what's there now, and `credentials` is already declared and threaded through the class (`agent.ts:86`, `agent.ts:102`).

This branch is based on `main` as of June 2026, before that merged, so the change is a duplicate rather than an addition.

Issue #5417, opened 2026-06-12
🐛 Bug: selfManagedAgents prop is seemingly ignored


fluxehub:
### ♻️ Reproduction Steps

1. Set up CopilotKit according to the [Quickstart](https://docs.copilotkit.ai/google-adk/quickstart?agent=bring-your-own), inside of a Vite SPA instead of a Next.js app.
2. Add code to `App.jsx` like
```jsx
import './App.css'
import { CopilotKit, CopilotSidebar, HttpAgent } from "@copilotkit/react-core/v2"
import "@copilotkit/react-core/v2/styles.css";

const testAgent = new HttpAgent({
  url: "http://localhost:8000"
});

function App() {
  return (
    <CopilotKit selfManagedAgents={{ "testAgent": testAgent }} agent="testAgent">
      <CopilotSidebar />
    </CopilotKit>
  );
}

export default App
```

### ✅ Expected Behavior

I expected to be able to interact with my AG-UI endpoint running from ADK as described by the [Self-managed agents](https://docs.showcase.copilotkit.ai/backend/self-managed-agents) docs.

### ❌ Actual Behavior

The error `ConfigurationError: Missing required prop: 'runtimeUrl' or 'publicApiKey' or 'publicLicenseKey'` is thrown. When adding a dummy `runtimeUrl`, CopilotKit attempts to fetch from this path and ignores self-managed agents. Am I missing something?

### 𝌚 CopilotKit Version

```shell
├── @copilotkit/react-core@1.59.5
├── @copilotkit/react-ui@1.59.5
├── @copilotkit/runtime@1.59.5
```

fluxehub:
Can confirm switching to CopilotKitProvider fixed this, however it should probably be updated in the docs as the docs still reference using CopilotKit
<img width="910" height="393" alt="Image" src="https://github.com/user-attachments/assets/80dac5a8-315b-4aca-98f9-cb233b0dcd6b" />

NathanTarbert:
Hey @fluxehub, thanks for flagging this and letting us know the docs need to be updated. 

Question: Would you like to contribute to the docs? Just checking before I create an internal ticket.

tylerslaton:
Hi @fluxehub, `CopilotKitProvider` is actually not the intended path here. If it is what fixed this then that's a bug on the `CopilotKit` side. The docs are correct and this is instead a bug we will look into.

PR #6305, opened 2026-08-02
fix(react-core): allow self-managed agents without runtime

deepshekhardas:
Port of #5435.


BenTaylorDev:
Closing — this fix is already on `main` as `packages/react-core/src/components/copilot-provider/copilotkit.tsx:855-859`, landed in June 2026 as `9907519989` (for #5417). The `hasLocalAgents` gate there covers both `agents__unsafe_dev_only` and `selfManagedAgents`, and `main` also carries coverage for it in `v1-self-managed-agents-validation.test.tsx`, including the empty-map case.

This branch is based on `main` as of June 2026, before that merged, so the change is a duplicate rather than an addition.

PR #6306, opened 2026-08-02
fix(runtime-client-gql): guard abort errors without message

deepshekhardas:
Port of #5437.

### DIFF
--- a/.changeset/guard-abort-error-message.md
+++ b/.changeset/guard-abort-error-message.md
@@ -1,0 +1,7 @@
+---
+"@copilotkit/runtime-client-gql": patch
+---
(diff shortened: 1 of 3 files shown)


BenTaylorDev:
Closing — this fix is already on `main` as `packages/runtime-client-gql/src/client/CopilotRuntimeClient.ts:24-38`, landed in June 2026 as `54b968bb95` (#2596). The `isAbortError` helper there is wired into both guard sites this PR patches (`:72` and `:156`).

This branch is based on `main` as of June 2026, one day before that merged, so the change is a duplicate rather than an addition.

PR #6308, opened 2026-08-02
fix: move katex CSS import from component to globals.css

### LINKED ISSUE #2851: 🐛 Bug: Next.js build fails due to global CSS import from transitive dependency (`@copilotkitnext/react`)
### ♻️ Reproduction Steps

1. Create a `Next.js` project
2. Install:
```
npm install @copilotkit/react-core @copilotkit/react-ui @copilotkit/runtime
```
3. Run:
```
npm run build
```
4. Build fails with the global CSS error above

### ✅ Expected Behavior

- CopilotKit should not import global CSS inside dependencies
- Consumers should be responsible for importing KaTeX CSS at the app level if needed
- `Next.js` builds should succeed without patching node_modules

### ❌ Actual Behavior

When building the project, or CI (like github) tries to build it will generate the following error: (preventing us from deployment or even development)

```
Global CSS cannot be imported from within node_modules.
Location: node_modules/@copilotkitnext/react/dist/index.js
Import trace:
  katex/dist/katex.min.css
  @copilotkitnext/react/dist/index.js
  @copilotkit/react-core/dist/...
``` 

### 𝌚 CopilotKit Version

```shell
├── @copilotkit/react-core@1.10.6
├── @copilotkit/react-ui@1.10.6
├── @copilotkit/runtime@1.10.6
```

### 📄 Logs (Optional)

```shell

```

deepshekhardas:
Fixes Next.js 'Global CSS cannot be imported from within node_modules' build failures. Fixes #2851.

### DIFF
--- a/docs/content/docs/(root)/troubleshooting/dev-console.mdx
+++ b/docs/content/docs/(root)/troubleshooting/dev-console.mdx
@@ -1,0 +1,88 @@
+---
+title: Using the Dev Console
+description: Learn how to use CopilotKit's Dev Console to debug and troubleshoot your copilot application.
+icon: "lucide/Terminal"
+---
+
+# Using the Dev Console to Troubleshoot
+
+CopilotKit includes a built-in **Dev Console** that helps you inspect and debug the runtime behavior of your copilot. It is one of the most useful tools for troubleshooting issues with actions, readable state, messages, and more.
+
+## Enabling the Dev Console
+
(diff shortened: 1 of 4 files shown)


BenTaylorDev:
Closing — this branch is based on `main` as of March 2026, and the paths it changes (`packages/v2/react/`, `docs/content/docs/`) were removed from the tree in March 2026 and July 2026 respectively. As written it can't be applied to current `main`.

Docs are now authored in `showcase/shell-docs/src/content/`; the top-level `docs/` path is a symlink.

PR #6309, opened 2026-08-02
use type field instead of instanceof

deepshekhardas:
Uses type field instead of instanceof.


BenTaylorDev:
Closing — this branch is based on `main` as of October 2024, and the paths it changes (`CopilotKit/packages/**`) were removed from the tree in March 2026. As written it can't be applied to current `main`.

Separately, the files it targets are either no longer in the tree or already contain no `instanceof` checks, so the change appears moot regardless of base.

PR #6311, opened 2026-08-02

deepshekhardas:
Merged from upstream PR #4947 by @Tushar-Khandelwal-2004.

### DIFF
--- a/packages/core/src/__tests__/state-manager.test.ts
+++ b/packages/core/src/__tests__/state-manager.test.ts
@@ -1,12 +1,7 @@
-import { describe, it, expect, beforeEach, vi } from "vitest";
+import { describe, it, expect, beforeEach } from "vitest";
 import { CopilotKitCore } from "../core";
-import {
-  AbstractAgent,
-  Message,
-  State,
-  RunAgentInput,
-  EventType,
-} from "@ag-ui/client";
+import type { Message, State, RunAgentInput } from "@ag-ui/client";
+import { AbstractAgent, EventType } from "@ag-ui/client";
 import { randomUUID } from "@copilotkit/shared";
 
 /**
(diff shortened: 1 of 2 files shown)


BenTaylorDev:
Closing — the premise doesn't hold on current `main`.

`@ag-ui/client` (pinned at 0.0.57) already accumulates streamed `TOOL_CALL_ARGS` into `messages[].toolCalls[].function.arguments` by itself. Scripting an `AbstractAgent` that emits `TOOL_CALL_START` -> three `TOOL_CALL_ARGS` deltas -> `TOOL_CALL_END`, with no `StateManager` in the path at all:

```
toolCallBuffer=""             messages[].arguments=""
toolCallBuffer='{"col'        messages[].arguments='{"col'
toolCallBuffer='{"color":"'   messages[].arguments='{"color":"'
FINAL: '{"color":"blue"}'
```

The final value is already complete, so there's no gap for `handleToolCallArgs` / `updateToolCallArguments` to close. Three related notes:

- `onToolCallArgsEvent` already supplies `toolCallBuffer` (the accumulated args) in its params, so `toolCallArgsById` recomputes something the hook hands you.
- `toolCallArgsById` is only cleared on `TOOL_CALL_END`, so an aborted or errored run leaves its entry behind.
- `updateToolCallArguments` clones the whole message array on every delta, which is a per-delta cost for a value the client is already computing.

One thing in here is real: at hook-invocation time the message view lags the current delta by one (second row above). If a renderer needs the in-flight delta, that's a small change built on `toolCallBuffer + event.delta` rather than a separate accumulator — worth its own PR if you have a case that needs it.

Separately, this branch is based on `main` as of May 2026, and `packages/core/src/core/state-manager.ts` has moved substantially since; #6294 is currently reworking the same file.

PR #6313, opened 2026-08-02
chore(core): fix tool result thread leak

deepshekhardas:
Fixes tool result thread leak. Includes changeset.


BenTaylorDev:
Closing — this branch is based on `main` as of February 2026, and the paths it changes (`src/v2.x/`) were removed from the tree in April 2026. As written it can't be applied to current `main`.

Also note the repo releases via `nx release` and doesn't take `.changeset/` files in PRs.

PR #6314, opened 2026-08-02
fix(channels): pin component identity so minification can't break durable actions

contextablemark:
Fixes the pre-existing bug tracked in [OSS-711](https://linear.app/copilotkit/issue/OSS-711/channels-action-component-identity-is-keyed-on-functionprototypename).

## The problem

Every durable-action path in Channels keyed a component by `Function.prototype.name` — a build artifact. Under a bundler that mangles names, a card posted by one deploy became undispatchable by the next: the persisted snapshot stores only that string, so the cold-path lookup missed, `ActionExpiredError` was swallowed, and the user's click did nothing with **no server-side trace**.

A second failure followed from the same root: because `mintId` folds the name into its hash, two components sharing a name could mint the *same* action id, letting a cold dispatch resolve a handler out of the wrong component tree.

## The fix

`defineChannelComponent(name, fn)` pins a `displayName` that `resolveComponentName` prefers over `fn.name`. That pinned name is what the registry files the component under, what `createChannel({ components })` seeds, and what `mintId` folds into every action id — so a rename no longer orphans live cards. The library intrinsics (`Message`, `Button`, …) pin their own names.

Two adjacent silences are now audible:

- a swallowed expired-action dispatch logs once, so the failure is visible server-side;
- registering two different components under one name warns (once per name), since their minted ids can collide.

Components without a pinned name keep resolving through `fn.name` exactly as before.

## Testing

| Suite | Result |
| -- | -- |
| `channels-ui` | 42/42 |
| `channels-core` | 266/266 |
| `runtime` (only affected dependent) | 130 files / 1835 tests |
| `check-types` | 21 affected projects |

New coverage includes an **end-to-end redeploy test** driving `createChannel` through a simulated name-mangling deploy, a **negative control** proving the original bug still reproduces without a pin, and a **backward-compatibility** case asserting unpinned components mint and resolve exactly as before.

## Upgrade note

Adding a pin to a component that has been running **unpinned** changes the ids it mints, so cards already posted go stale once. This is called out in the docs.

## Docs

`channels/interactive` gains a "Pin the component name" section covering the uniqueness rule and the invalidation caveat; `persistence-and-scaling` and `commands-and-reactions` cross-link to it.

### DIFF
--- a/packages/channels-core/src/action-registry.test.ts
+++ b/packages/channels-core/src/action-registry.test.ts
@@ -144,8 +144,6 @@ describe("ActionRegistry", () => {
       regB.registerComponent("Confirm", Confirm as never);
 
       // Cold dispatch must succeed and fire the handler.
-      const handlerFired: string[] = [];
-      // Wrap the existing Confirm component so we can assert the specific call.
       const value = await regB.dispatch(id, ctx);
       expect(value).toEqual({ ok: "restart-test" });
       // The shared `clicks` array is populated by Confirm's onClick.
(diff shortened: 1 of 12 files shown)


contextablemark:
## CR loop — converged

Three review rounds at Tier 2 (12 agents each: 2 import-graph source clusters, 2 clusters each for silent-failure/test/comment/type-design, tests-only, and the packaging surface). 36 reviewer dispatches, 0 retries, 0 API errors.

| Round | Bucket (a) | Outcome |
| -- | -- | -- |
| 1 | 3 + 1 discretionary (b) | fixed in `47633b6119` |
| 2 | 1 | fixed in `1f8b30d6e6` |
| 3 | **0** | convergence floor |

Procedure 3 bucket-(c) promotion audit: **`PROMOTE_TO_A count: 0`** (5 STAY_IN_C, 1 PROMOTE_TO_B held as ledger-only).

### What the review actually caught

**A real bug the original commit missed.** `bindRenderable`'s inline (non-component) path bound handlers with an empty component name, so `mintId` keyed the action id on tree path alone — two structurally identical inline posts in one conversation minted the *same* id, and the later binding overwrote the earlier in both the hot cache and the store. Clicking the older message silently ran the newer message's handler. Since an inline binding can never rehydrate (`dispatch` throws once the snapshot's `component` is falsy), the id has no reason to be content-addressed; it now gets a fresh random id, the same treatment continuations already had.

**Two tests that proved nothing.** The name-collision test re-registered a *new* closure while its comment claimed "same function", so it couldn't distinguish same-function silence from `warnOnce` deduping by name. And the library-components test asserted `resolveComponentName(Button) === "Button"` on plain function declarations whose `fn.name` already reads `"Button"` — it passed with or without the `displayName` pin. Deleting the pin left it green while removing the protection it existed to guard. Both now verify the real contract; the second mangles `fn.name` first, and was checked adversarially (removing the pin fails it with `expected 'a' to be 'Button'`).

### Deferred

20 findings were confirmed real but out of this PR's subject and are recorded as follow-up work rather than fixed here — including a replayable-continuation hole on the normal dispatch path, an `awaitChoice` waiter keyed only by `conversationKey`, missing ingress dedup on the thread-start/reaction/modal paths, and lossy `stableStringify` id collisions (`undefined`/`null`, `NaN`/`Infinity`, BigInt throw). The last one is squarely about action-id derivation, but fixing it changes id derivation for every consumer and invalidates in-flight cards globally — deliberately not smuggled into a targeted identity fix.

Verification: oxfmt clean (87 files), channels-ui 42/42, channels-core 267/267, build + typecheck green.

PR #6315, opened 2026-08-02
feat(examples): let every starter host a managed Channel

BenTaylorDev:
## What this does

Every CopilotKit starter can now host an Intelligence Channel (Slack/Teams) as well as a web app, so `copilotkit init`'s channel path can finish at `npm run channel` instead of asking the developer to hand-wire runtime source afterwards.

Per starter, four files:

1. **`agent.ts`** — agent construction moves out of the Next.js runtime route into a `createDefaultAgent()` factory. Pure extraction; the web mount's behaviour is unchanged.
2. **`route.ts`** — imports the factory instead of constructing inline.
3. **`channels.mts`** — declares the Channel: name resolution, `createChannel`, and the `onMessage` handler. This is the file a developer edits to *customise* a Channel — commands, reactions, an `onMention` handler. It holds the only per-starter difference in the pair: the agent import path (`./src/agent` or `./app/agent`).
4. **`channel-host.mts`** — the process that owns the Channel's lifetime. **Byte-identical in all 15 starters**, so the CLI can emit it verbatim.

The host declares **no adapters, no provider credentials, and no provider endpoint** — Intelligence owns the provider edge — which is why one file works for every provider. Adding WhatsApp/Telegram/Discord later touches the server and the CLI, never a starter.

The host also runs **no HTTP server**. Nothing calls this process: the gateway connection is outbound, and holding it open is what keeps the process alive. It uses `createCopilotRuntimeHandler` + `ready()` — the long-running-host pattern documented in `fetch-handler.ts` — rather than building a request listener for its activation side effect.

Covers 15 starters (17 of 21 CLI frameworks; `langgraph-python` backs three). `langgraph-fastapi` gets dependency pins only, because it is enrolled in the `_parity` drift check and would otherwise mismatch the north-star.

## ⚠️ Temporary, must not ship to users as-is

- **All `@copilotkit/*` deps are pinned to a prerelease**, `1.64.3-canary.1785633429`. Stable `1.64.2` lacks `identifyUser` on `createChannel`, so tracking `main` required the canary. **These pins must move to a stable release before the CLI's channel path goes live.**
- **`mastra` carries an `.npmrc` with `legacy-peer-deps=true`.** `@ag-ui/mastra` declares a peer of `@copilotkit/runtime@^1.10.5`, and node-semver excludes prereleases from a caret range — so *any* prerelease pin fails that check, and a fresh clone would fail `npm install` without it. The real fix is a prerelease-inclusive peer range upstream (its other peers already use `>=1.0.0-0 <2.0.0-0`).

## Review feedback addressed

**"This file seems overly complex and adds a brand new service alongside an already existing webserver"** ([thread](https://github.com/CopilotKit/CopilotKit/pull/6315#discussion_r3699832425)).

The complexity was real and is fixed in two ways below. Folding the Channel into the existing Next.js route is not possible: activation on the App Router path is lazy by design (`fetch-handler.ts:39`, `hono.ts:15-26`), so `channels: [channel]` in `route.ts` would build a ChannelManager, open no socket, and **silently never connect**; forcing it with `ready()` mints a competing listener per cold start. A Channel is a long-running worker, not a serverless request handler. Full reasoning in the thread.

What did change:

- **The Channel moved out of the host** into `channels.mts`, which also made the host byte-identical everywhere.
- **The HTTP server is gone.** Its comment claimed the server was what "keeps the lifecycle-owning process alive." That is false — an open undici WebSocket holds the event loop on its own. The server was therefore standing up a second, uncalled copy of the runtime API on port 8300 for no reason. Removing it also drops `node:http`, `basePath`, and the `CHANNEL_PORT` env var.

Terminology: these are **Channels**, not "managed Channels", throughout the starter READMEs, host comments, and log output. "Managed" stays in the docs and runtime, where it is the load-bearing discriminator against direct adapters.

Still open, tracked as a fast-follow: **OSS-729** — `resolveChannelName()` is ~56 hand-rolled lines per starter reading `.copilotkit/channels.json`, a file that already has a versioned zod schema and a canonical parser in `Intelligence/libs/channels-setup`. That library is unpublished, so every consumer re-derives the read and ignores `version` while doing so. Publishing a read-side `resolveDeclaredChannel()` helper collapses those 56 lines to about three.

## Verification

- **14/14** starters with a `typecheck:channel` script pass. `mastra` has no such script by design (166dc94691); its pre-existing Mastra `Memory` type error was confirmed byte-identical before and after by stashing and re-running.
- **Project `typecheck` and `build` are unaffected** by the host pair: `.mts` is not matched by the Next tsconfig's `**/*.ts` include, confirmed with `tsc --listFiles` (0 hits for either file). The earlier full sweep stands — 15/15 typecheck with no new errors, seven starters with pre-existing errors each confirmed identical on the old pins; 14/15 build, `llamaindex` failing on a pre-existing `clickOutsideToClose`/`CopilotChatProps` error that surfaces there because it is one of only three starters not setting `typescript.ignoreBuildErrors`.
- **`npm run channel` was actually executed** on both failure paths — missing `channels.json`, and missing `INTELLIGENCE_API_KEY` with a name supplied. This mattered: the natural extensionless `./channels` import does **not** resolve under `moduleResolution: "bundler"` for a `.mts` file. Caught at typecheck, fixed with the `./channels.mjs` specifier, and confirmed to resolve under both `tsc` and `tsx`.
- **The WebSocket liveness claim was tested, not assumed** — a standalone repro with no HTTP server and no timers of its own, holding a single WebSocket, stayed up indefinitely (its 3s timer fired and was collected; the process had to be killed).
- All 15 hosts machine-verified byte-identical; `channels.mts` verified as exactly two variants differing only in the agent import path.
- Per-starter agent URL/constructor fidelity checked against each pre-change route — `langgraph-js` keeps its `LANGGRAPH_DEPLOYMENT_URL` fallback and `langsmithApiKey`; `crewai-flows` correctly has no `/agui` suffix; `strands-python` keeps `STRANDS_AGENT_URL`.
- `a2a-middleware`'s per-run isolation override survived its move, and its 25-line `instructions` prose is byte-identical.
- `pnpm parity:check` output is byte-identical to a pre-change baseline. Note it already exits 1 on `main`, so the gate throughout was "no new errors versus a captured baseline".

**Not verified:** no end-to-end run against a real Slack workspace on this branch. The host's failure paths were exercised (missing config, ambiguous config, malformed config, missing API key), but nobody has watched a message travel from Slack to an agent and back through a scaffolded starter.

## One fix worth calling out

`handler.channels.ready()` resolves once every Channel reaches `online` **or `setup_required`**. The host originally logged success unconditionally after it, so a developer who hadn't finished the Slack app install saw "holding Channel", messaged the bot, and got silence with no diagnostic. It now calls `status()` and reports the true per-Channel state, treating `setup_required` as the normal waiting state it is.

## Related

- Depends on nothing; independent of #6298.
- **OSS-729** (fast-follow) — publish `resolveDeclaredChannel()` so starters stop hand-rolling the `channels.json` read.
- `skills/copilotkit-channels/SKILL.md` still needs rescoping — teaching it to *customise* the Channel the starter already ships in `channels.mts`, rather than hand-write the host. That file is introduced by #6298, so the change belongs on that branch rather than here.


examples/integrations/langgraph-python/channel-host.mts
tylerslaton, line 1: This file seems overly complex and adds a brand new service along-side an already existing webserver. A better way to approach this would be to bring this into the existing runtime and then enable it conditionally based on if the user specifies a channel name env var. Something like this very rough code:
```ts
// channels.ts

const name = process.env.CHANNEL_ID
const channel = createChannel({
  name
  agent: new HttpAgent(...) // unique per framework
})

const channels = channelName ? [] : []
export const channels
```

```ts
// runtime.ts
import { channels } from "./channels.ts"

  intelligence,
  channels,
})
```

This eliminates an extra process and keeps things simpler. Thoughts?

BenTaylorDev: Good push — the file is too big and I'm changing it. But the single-process version can't work for these starters, and the way it fails is the problem.

**Why it can't**

All 15 starters mount the runtime in a Next.js App Router route (`createCopilotEndpoint` → `hono/vercel`'s `handle`). Activation on that path is lazy by design:

- `fetch-handler.ts:39` — "On a SERVERLESS / EDGE host (Cloudflare Workers, Next.js App Router), do NOT call `ready()` — those hosts freeze/recycle per-request isolates and cannot own a persistent listener, and separate cold starts would mint conflicting listeners."
- `hono.ts:15-26` names this exact case: auto-starting there "would mint a competing listener for the same Channel on every cold start, which is exactly what the lazy design exists to prevent."

So `channels: [channel]` in `route.ts` builds a ChannelManager, opens no socket, and never connects. It type-checks, it reviews clean, and it silently does nothing. Forcing it with `ready()` trades that for competing listeners per cold start. `deploy-and-operate.mdx` puts it plainly: a Channel is "a long-running worker with an outbound realtime connection. It is not a serverless request handler."

Minor: `const channels = channelName ? [channel] : []` also drops the `RuntimeWithDeclaredChannels` brand (it needs a literal non-empty tuple — `node.ts:88`), so `.channels` goes optional and every call site needs `?.`.

**Where you're right**

~56 of 216 lines are `resolveChannelName()` — hand-rolled `readFileSync` + `JSON.parse` + duck-typed `{channels?: {name?: string}[]}`, copied 15 times. That file has a versioned zod schema and a canonical parser in `Intelligence/libs/channels-setup`, but it's `@cpki/*` and unpublished, so every consumer re-derives it — and ignores `version` while doing so.

OSS-729 covers publishing the read-side helper: ~56 lines → ~3 per starter. The open question in there is who owns the schema across the two repos.

Two things I'll do on this branch:

1. Extract Channel construction into `channels.ts` next to `agent.ts` — your sketch's shape, and it genuinely thins the host.
2. Check whether `createServer(listener).listen()` is needed at all. Activation already starts at listener creation (`node.ts:113`) and the gateway holds a ref'd WebSocket, so the HTTP server and `CHANNEL_PORT` may be pure ceremony. Untested — needs a real API key.

AlemTuzlak (approved):

PR #6316, opened 2026-08-02
fix(runtime): preserve realtime gateway handoff continuity

maxkorp:
## Summary
- replay runner events sequentially with immutable IDs, terminal acknowledgements, and bounded retries
- preserve accepted runs across planned and prolonged abnormal reconnects without manufacturing terminal errors
- watchdog stalled abnormal socket reconnects while keeping pre-join failure behavior bounded
- make Slack and Teams delivery settlement safe across deployment handoffs
- add the compiled cross-service Kind runner used by the Intelligence integration suite

## Test plan
- [x] `NX_DAEMON=false NX_SKIP_NX_CACHE=true pnpm nx test runtime` (1,840 tests)
- [x] focused Intelligence runner suite (42 tests)
- [x] `pnpm --filter @copilotkit/runtime check-types`
- [x] `pnpm --filter @copilotkit/channels-intelligence check-types`
- [x] channels-intelligence suite (174 tests)
- [x] real cross-service Kind planned and hard-stop lifecycle suite; accepted agent remains un-aborted

### DIFF
--- a/packages/runtime/src/v2/runtime/runner/intelligence.ts
+++ b/packages/runtime/src/v2/runtime/runner/intelligence.ts
@@ -5,8 +5,7 @@ import type {
 } from "./agent-runner";
 import { AgentRunner } from "./agent-runner";
 import type { AgentRunnerStopRequest } from "./agent-runner";
-import { EMPTY, Observable, from } from "rxjs";
-import { catchError, finalize } from "rxjs/operators";
+import { Observable } from "rxjs";
 import type { AbstractAgent, BaseEvent, RunStartedEvent } from "@ag-ui/client";
 import { EventType } from "@ag-ui/client";
 import {
@@ -44,7 +43,20 @@ interface ThreadState {
   currentEvents: BaseEvent[];
   nextEventSeq: number;
   hasRunStarted: boolean;
+  hasJoined: boolean;
+  producerFinished: boolean;
+  pendingEvents: Map<string, Record<string, unknown>>;
+  activeEventPush: { eventId: string; attempt: number } | null;
+  nextEventPushAttempt: number;
+  eventRetryTimer: ReturnType<typeof setTimeout> | null;
+  socketReconnectWatchdog: ReturnType<typeof setTimeout> | null;
+  eventRetryAttempt: number;
+  completeRun: () => void;
 }
+
+const MAX_CONSECUTIVE_SOCKET_ERRORS = 5;
+const EVENT_RETRY_BASE_MS = 100;
+const EVENT_RETRY_MAX_MS = 2_000;
 
 export class IntelligenceAgentRunner extends AgentRunner {
   private options: IntelligenceAgentRunnerOptions;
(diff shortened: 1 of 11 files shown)


packages/runtime/src/v2/runtime/runner/intelligence.ts
```diff
@@ -445,39 +487,167 @@ export class IntelligenceAgentRunner extends AgentRunner {
...
+  }
+
+  private resetPendingEventRetry(state: ThreadState): void {
+    this.clearPendingEventRetry(state);
+    state.eventRetryAttempt = 0;
+    state.activeEventPush = null;
+  }
+
+  private replayPendingEvents(state: ThreadState): void {
+    if (state.activeEventPush !== null) {
```
BenTaylorDev: **Blocking: this serializes the whole event stream.**

This early-return means only one event is ever in flight, and `pushPendingEvent` doesn't send the next until `.receive("ok")` arrives. On `main` the equivalent push is fire-and-forget (`channel.push("event", ...)` with no `.receive`), so tokens streamed out without waiting.

Combined with Intelligence#723, each ack now also blocks on a synchronous Redis `EVAL` in the gateway. Per-token cost becomes: runner → WS → gateway → Redis → gateway → WS → runner, **then** the next token. A 500-token response goes from effectively-instant fan-out to 500 sequential round-trips.

This is the OSS-648 root cause — one durable round-trip per AG-UI token — and the fix recorded there was to *batch* `text_delta` in the SDK. This moves the opposite way and adds head-of-line blocking.

Durability doesn't require serialization: keep `pendingEvents` as the unacked set, but allow a window of concurrent pushes (or batch multiple events per push) and let `cpki_event_seq` handle ordering on the gateway side — it already does, for replay.

packages/runtime/src/v2/runtime/runner/intelligence.ts
```diff
@@ -445,39 +487,167 @@ export class IntelligenceAgentRunner extends AgentRunner {
...
+    ) {
+      return;
+    }
+
+    const delay = Math.min(
+      EVENT_RETRY_BASE_MS * 2 ** state.eventRetryAttempt,
+      EVENT_RETRY_MAX_MS,
+    );
+    state.eventRetryAttempt += 1;
+    state.eventRetryTimer = setTimeout(() => {
```
BenTaylorDev, line 722: **Blocking: no max attempts and no deadline.**

`EVENT_RETRY_MAX_MS` caps the *delay* at 2s, but `eventRetryAttempt` has no ceiling and there's no overall give-up. If the gateway permanently rejects an event — `missing_event_id`, or `runner_event_not_durable` while Redis is wedged — this retries every 2s indefinitely.

The knock-on is worse than the retry itself: `pendingEvents` never empties, so `completeWhenDurable` never fires, so `removeThread` never runs. The run's observable **never completes** and the client hangs forever, leaking the `ThreadState` and a live timer per stuck run. On `main` a failed push was ignored and the run completed.

Needs a bounded attempt count or a wall-clock deadline, after which the run emits `RUN_ERROR` and completes. A hung run with no timeout is worse for the user than a failed one.

BenTaylorDev (commented):
## Review: gateway handoff continuity

Reviewed together with CopilotKit/Intelligence#723. The reconnect/replay design is the right shape — durable per-event ids, ordered replay by `cpki_event_seq`, and not treating transport recovery as terminal once the run is accepted is a real correctness improvement over the old `MAX_CONSECUTIVE_ERRORS` abort. Two things I think block it.

### Blocking

1. **Event pushes are now strictly serialized**, one round-trip per event, where `main` was fire-and-forget. Detail inline at `replayPendingEvents`.
2. **A permanently-rejected event wedges the run forever** — no max attempts, no deadline, and the observable never completes. Detail inline at `schedulePendingEventRetry`.

### The one thing to verify before merge

**Measure per-token streaming latency against `main` with a real multi-hundred-token response.** The Kind suite on the Intelligence side asserts ordering and completeness of 404 events but never latency, so this regression passes every existing test green.

---
<sub>Adversarial review pass against current `origin/main`. Both branches are 0 commits behind, so this is about mechanism, not drift.</sub>

BenTaylorDev:
## Re-review of the new tip (`96ff87f3`)

Re-derived from source rather than from the commit messages. **Both blockers from my earlier review are genuinely fixed.**

### Verified fixed

**Per-event serialization → batching.** `MAX_RUNNER_EVENT_BATCH_SIZE = 32` with a `RUNNER_EVENT_BATCH_FLUSH_MS = 5` window, gated on `runner_event_batch_v1`. The negotiation is the right shape: `supportsRunnerEventBatch` initializes to `false` and only flips when `supportsRunnerEventBatch(response)` finds the capability in the join reply, so a new runtime against an older gateway falls back to single-push instead of sending a message the gateway can't parse. Ordering still comes from `cpki_event_seq`, which the gateway also relies on for replay.

**Unbounded retry / hung run.** Now guarded twice over, which is more than I asked for:
- `EVENT_DURABILITY_DEADLINE_MS` wall-clock deadline via `failIfEventDeadlineExceeded` → `failThread` → `removeThread` + `abortRun()` + `failRun(error)`, so a stuck event terminates the run instead of hanging the observable forever.
- `isPermanentEventFailure` on `retryable === false` fails fast rather than waiting out the deadline.

`isRetryableJoinError` also now short-circuits on an explicit `retryable === false`, which closes the case where a permanent join rejection would have been retried.

### One coupling worth checking on the gateway side

The `retryable` contract is now load-bearing in both directions, and there's one case on the Intelligence side where it's set wrong — `"outbox conflicts with expired dedupe"` comes back as `retryable: true` even though it's deterministic for up to an hour. The effect lands here: `schedulePendingEventRetry` will retry it until `EVENT_DURABILITY_DEADLINE_MS` expires and then fail the run, rather than `isPermanentEventFailure` catching it immediately. Detail in my comment on CopilotKit/Intelligence#723 — no change needed in this PR, but it's the reason the classification matters.

### Notes

- Branch is 7 commits behind main, all of them `setup-slack-channel` skills work — nothing touching `packages/runtime` or `packages/channels-intelligence`. `merge-tree` reports a clean merge, so no rebase urgency.
- The `channels-intelligence` delivery changes (`delivery-adapter.ts`, `delivery-transport.ts`, `delivery-contracts.ts`) are new since my last pass. I spot-checked without finding problems, but they weren't in what I originally reviewed.

### The one thing worth doing before merge

**Benchmark the batched path against `main`.** `bench/runner_event_store_batch.exs` landed on the Intelligence side; the number it produces is what actually closes the latency question, and no CI check covers it.

---
<sub>Follow-up adversarial pass against current `origin/main`.</sub>

BenTaylorDev (approved):
Approving at `8e83dfa6db`. Content is identical to what I reviewed in my last pass — the force-push was a rebase onto current main, same 12 commits, same +2315/−163.

Both blockers I raised are fixed and verified:

- **Per-event serialization → batching.** Batch 32 with a 5ms flush window, gated on `runner_event_batch_v1`. The negotiation defaults `supportsRunnerEventBatch` to `false` and only flips on an advertised capability, and the gateway kept its single-event `handle_in("event", …)` clause — so new-runtime→old-gateway and old-runtime→new-gateway both work.
- **Unbounded retry / hung run.** Guarded twice: the `EVENT_DURABILITY_DEADLINE_MS` deadline via `failIfEventDeadlineExceeded` → `failThread` → `abortRun()` + `failRun(error)`, plus `isPermanentEventFailure` failing fast on `retryable === false`. The gateway now sets that flag correctly for the stale-outbox case (CopilotKit/Intelligence#723 `f68db4d7`), so the two halves line up.

Approving with the note that the `channels-intelligence` delivery changes (`delivery-adapter.ts`, `delivery-transport.ts`, `delivery-contracts.ts`) arrived after my first pass and got a spot-check rather than a full review.

Merge this together with CopilotKit/Intelligence#723 — #723's join-time capability advertisement is what activates the batching path here.

PR #6320, opened 2026-08-02
docs(channels): document Teams one-command setup

tylerslaton:
## Summary

Document the draft-first Microsoft Teams setup flow across Channels docs, skills, and the Teams adapter README.

## Why

Intelligence now offers a resumable Fast CLI path and a Guided manual path while keeping custom branding artifacts local and separating provider completion from runtime health.

## How

- Describe the fully scoped provisioning and resume contract.
- Replace Azure Bot and manifest-editing guidance with Teams Developer Portal plus Entra.
- Teach both setup skills the local-only artifact and Team-installation boundaries.
- Update documentation contract tests for the new path.


AlemTuzlak (dismissed):

AlemTuzlak (approved):

marthakelly:
I found a few issues worth tightening before/after merge:

1. The Teams-only fully scoped CLI command appears outside a `<FrontendOnly frontend="teams">` block in `intelligence.mdx`, so it will render on the Slack-scoped page too. `filterFrontendScopedBlocks` only removes content inside mismatched wrappers, so this needs to move under Teams-only content or be otherwise scoped.

2. The new skill source citations point at `CopilotKit/Intelligence/docs/superpowers/specs/2026-08-01-teams-one-command-setup-prd.md`, but I could not find that file on the Intelligence branch. These should point at an accessible source of truth, likely Linear `OSS-735` or the actual PRD location.

3. The docs contract test should assert the load-bearing Teams setup claims more directly: exact `--project-id/--channel-id/--adapter teams/--provision` command shape, `Files.ReadWrite.All` admin consent, blocked/resumable approval, personal-only install not completing setup, and “Created and installed” not implying runtime/message health.

4. `expect(teams).toContain("Azure Bot")` is now stale. It passes because the new docs say Azure Bot is *not* part of the normal path, so the assertion no longer proves the intended contract. I’d replace it with an assertion for the new negative guidance and remove coverage of the old Azure Bot setup flow.

PR #6323, opened 2026-08-03
fix(web-inspector): show step lifecycle events

### LINKED ISSUE #6324: Web Inspector does not show STEP_STARTED and STEP_FINISHED events
## Problem

The Web Inspector does not record `STEP_STARTED` or `STEP_FINISHED` events in the live event list, even though the AG-UI client emits both events.

This makes a complete event stream look incomplete in the Inspector.

## How to reproduce

1. Run an agent that emits `STEP_STARTED` and `STEP_FINISHED`.
2. Open the Inspector event list.
3. The two step events are missing.

## Expected behavior

The Inspector should record both step lifecycle events and include them in the event filter.

JYbill:
## What does this PR do?

The Web Inspector did not record `STEP_STARTED` or `STEP_FINISHED` from live agent events.

This PR:

- records both step lifecycle events
- adds both events to the Inspector filter
- adds a regression test

## Related PRs and Issues

- Fixes #6324

## Testing

- `pnpm nx test @copilotkit/web-inspector --skip-nx-cache`
- `pnpm nx run @copilotkit/web-inspector:check-types --skip-nx-cache`
- `pnpm nx run @copilotkit/web-inspector:build --skip-nx-cache`
- repository pre-commit checks

## Checklist

- [x] No documentation update is needed because this fixes existing behavior without changing the public API

### DIFF
--- a/packages/web-inspector/src/__tests__/web-inspector.spec.ts
+++ b/packages/web-inspector/src/__tests__/web-inspector.spec.ts
@@ -328,6 +328,30 @@ describe("WebInspectorElement", () => {
     );
   });
 
+  it("records step lifecycle events", async () => {
+    const { agent, controller } = createMockAgent("alpha");
+    const { core, emitAgentsChanged } = createMockCore({ alpha: agent });
+    const inspector = createInspectorWithCore(core);
+
+    emitAgentsChanged();
+    await inspector.updateComplete;
+
+    controller.emit("onStepStartedEvent", {
+      event: { stepName: "test-step" },
+    });
+    controller.emit("onStepFinishedEvent", {
+      event: { stepName: "test-step" },
+    });
+    await inspector.updateComplete;
+
+    const internals = getInternals(inspector);
+
+    expect(internals.flattenedEvents.map((event) => event.type)).toEqual([
+      "STEP_FINISHED",
+      "STEP_STARTED",
+    ]);
+  });
+
   it("normalizes context, persists state, and copies context values", async () => {
     const { core, emitContextChanged } = createMockCore();
     const inspector = createInspectorWithCore(core);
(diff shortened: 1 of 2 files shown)


BenTaylorDev (approved):
CI is green here (33 pass, 1 skipped), and I also verified locally against **current `main`** — this branch is behind, so I merged it forward first (clean merge, no conflicts):

- `nx run-many -t test check-types -p @copilotkit/web-inspector` → 5 files, **134 tests passed**, `tsc --noEmit` clean
- `oxlint` → 0 errors (5 warnings, all pre-existing in untouched code)
- `oxfmt --check` → already correctly formatted

Mechanism confirmed rather than assumed: `AgentSubscriber` in `@ag-ui/client@0.0.57` really does both declare *and* dispatch `onStepStartedEvent` / `onStepFinishedEvent` — they sit in its event switch right next to the run events — so these handlers will actually fire on a live stream. I also confirmed the new test isn't vacuous: reverting the three edits makes it fail with `expected [] to deeply equal [ 'STEP_FINISHED', 'STEP_STARTED' ]`.

Two things I checked that this PR correctly leaves alone:

1. **`getAgentStatus` is untouched, and that's the right call.** `events` is newest-first, so if `STEP_STARTED` were folded into that run-lifecycle `find`, a step arriving mid-run would win the lookup, match neither branch, and fall through to `"idle"` — an actively-running agent would render as idle. `RUN_STARTED`/`RUN_FINISHED`/`RUN_ERROR` already bracket every step, so keeping steps out is correct.
2. **The timeline builder already rendered these** as "Step started" / "Step finished", which is exactly why replayed and imported traces showed step events while the live list didn't. Adding them to `AGENT_EVENT_TYPES` also makes them selectable in the event-type filter, which is generated from that array — good call including it, since recording alone would have left them unfilterable.

Nice, minimal, correctly-placed fix — thanks for reporting and fixing it in the same pass.

PR #6325, opened 2026-08-03
fix(react-core): invalidate messages memo for activity object content

seanxuu:
## Summary
- `CopilotChat` fingerprints messages for `useMemo` with `contentKey = 0` for any non-string / non-array content.
- `ACTIVITY_SNAPSHOT` updates often keep the same `messageId` and only replace object `content` (progress / generative UI). That left activity renderers stuck on the first frame until some other list change forced a refresh.
- Include `JSON.stringify(m.content)` for object content in the fingerprint. Multimodal attachments remain on the array-length branch, so this does not reintroduce base64 serialization for user uploads.
- Add a regression test that replaces the same activity `messageId` and asserts the UI updates.

## Test plan
- [x] `nx run @copilotkit/react-core:test` (or the package's activity e2e suite)
- [x] Manual: stream repeated `ACTIVITY_SNAPSHOT` with the same `messageId` + `replace: true` and confirm custom activity renderers update each frame


BenTaylorDev (approved):
Verified against current `main`. The `0` fallback at `CopilotChat.tsx:985-990` is real, and worth noting it was only ever reachable for `activity` messages — in `@ag-ui/core@0.0.57` that's the one role whose `content` is a `Record<string, any>`; every other role is a string or an array. So this targets exactly the affected case, and the claim that multimodal attachments stay on the array-length branch holds.

I confirmed the regression test earns its keep: reverting only `CopilotChat.tsx` to main and keeping the test fails with `expected 'Working on your request' to be 'Understanding your request'` — the frozen first frame from the issue. With the fix, green.

Two things I checked before worrying about them out loud:

- **`JSON.stringify` throwing on a cycle in render.** Not reachable — activity content is zod-parsed from wire JSON, so it's acyclic.
- **Serialization cost per render.** Measured: 0.0006 ms at 500 B, 0.005 ms at 4.7 KB, 0.039 ms at 47 KB. Negligible against a memo that otherwise freezes the UI.

`contentKey` appears at this one site only, so nothing else needs the same treatment. `tsc --noEmit` clean, and CI is green across the matrix.

PR #6332, opened 2026-08-03
fix(channels-intelligence): restore inbound trigger files

tylerslaton:
## Problem

Channel agents lose inbound files whenever the current-trigger transcript omits attachments. The currently deployed Intelligence path always omits those files because normalized_payload never contains their handles.

## Why

The delivery adapter seeds the current inbound turn from the transcript, and core then deduplicates the explicit prepared input. Files present only on the prepared delivery therefore never reach either the agent-history consumer or channel.getMessages during version skew.

## Fix

Restore a missing current-trigger transcript file list from the prepared delivery inside ClaimedChannelDelivery.getTranscript(), where the result is shared and memoized for both consumers. Existing transcript files are preserved, so the Intelligence fix and this fallback cannot duplicate attachments.

Either PR independently repairs the agent path. Coverage proves both an omitted transcript and an already-correct transcript hydrate the image for getMessages and agent seeding.

### DIFF
--- a/packages/channels-intelligence/src/delivery-adapter.ts
+++ b/packages/channels-intelligence/src/delivery-adapter.ts
@@ -244,7 +244,10 @@ export class DeliveryAdapter implements PlatformAdapter {
   private async loadProviderAgentHistory(
     target: DeliveryReplyTarget,
   ): Promise<{ messages: Message[]; historyIds: ReadonlySet<string> }> {
-    const transcript = await target.claimedDelivery.getTranscript();
+    const transcript = restorePreparedTriggerFiles(
+      await target.claimedDelivery.getTranscript(),
+      target.delivery.turn.input,
+    );
     const messages = await transcriptAgentMessages(
       transcript,
       target,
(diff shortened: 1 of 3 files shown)


BenTaylorDev:
Reviewed this alongside CopilotKit/Intelligence#728. Traced both fixes against the real ingress and prepared-delivery-store code. The approach is sound and the two fixes compose safely — one real gap below.

## This patches one of two transcript consumers

`restorePreparedTriggerFiles` is applied at `delivery-adapter.ts:247` (`loadProviderAgentHistory`) but not at `delivery-adapter.ts:1075` (`getMessages`). That second path runs the same `transcriptContent` hydration and additionally surfaces `providerMessage.files` (`delivery-adapter.ts:1255`), so a handler calling `channel.getMessages()` still sees the current trigger with no attachment — during exactly the version-skew window this shim exists to cover.

Suggested location instead: inside `ClaimedChannelDelivery.getTranscript()` (`delivery-transport.ts:471`). It already holds `this.delivery`, so the prepared input is in scope, and it memoizes through `transcriptPromise`. That covers both consumers from one place and runs the mapping once, rather than re-mapping the memoized transcript on every `loadProviderAgentHistory` call as the current placement does.

## Coverage gap

Nothing exercises the path that becomes production once Intelligence#728 deploys: the transcript supplies files on the `currentTrigger` message, this shim no-ops, and the agent still receives the hydrated image. The existing managed-file fixtures in `delivery-transcript.test.ts` (`"file extras"`, `"oversized file metadata"`) all sit on `messages[0]`, which is `currentTrigger: false`.

That assertion is what keeps these two PRs from quietly diverging later — it pins the no-op branch that will do all the real work in production.

## Context worth having

The server-side bug is total, not intermittent. `slack-ingress.ts:702` / `teams-ingress.ts:720` never write `files` into `normalized_payload` at all — handles only exist in `prepared_input`. So the transcript omitted inbound files on **every** delivery, which is why this shim will fire for essentially all traffic until Intelligence#728 lands.

Two things I checked and they're fine:

- **No double-attachment with both fixes live.** Transcript files present → `files.length === 0` is false → shim no-ops → core still dedups the explicit prepared input. One copy.
- **No deploy ordering constraint.** Either PR alone repairs the agent path.

## Minor

`providerFileId` is sliced to 512 here; the corresponding Intelligence tests assert the unsliced string. They agree only because fixture handles are short — no test pins the cap on either side.

BenTaylorDev (approved):

PR #6338, opened 2026-08-03
fix(showcase/harness): make PID exhaustion and roll-up infra aborts visible in the signal path

jpr5:
Three gaps the 2026-08-03 prod incident exposed in the harness signal path, ranked lightest first. Two are fixed here; the third turned out not to exist and is skipped with evidence.

**Background.** The prod prober fleet exhausted its cgroup PID limit (no PID-1 reaper; ~730 zombies after 7 days). Probes could not launch browsers, so 428 of 784 prod cells recorded `abort` across ~20 slugs while the demos were provably healthy. A scheduled watcher then paged a human with a false outage.

The information was **ignored, not destroyed** — `GET /api/matrix` correctly reported 4 red for `langgraph-python`, not 38, because the cell model already folds `abort`/`driver-error` to gray. But two surfaces genuinely could not see what was happening, and this PR fixes those.

**Explicit non-goal, not done here:** `feature-timeout` is *not* added to `INFRA_ERROR_CLASSES`. A real product hang produces exactly that class, so classifying it as infra would mask genuine defects. A test now pins it out of the set, so a future change in that direction fails loudly (see mutation F).

No existing assertion or predicate was weakened.

---

## Patch 1 — expose the cgroup PID gauges on the runs read-model

`pidsCurrent`/`pidsMax` have been on the worker heartbeat contract and in PocketBase since the `workers` collection was created, but the `/api/runs` projection dropped both. During the incident every consumer — the dashboard strip and the external watcher alike — read `online, 4/40 contexts` for workers pinned at `pids 1000/1000` and unable to fork a browser. The context budget was never the binding constraint; the PID ceiling was, and the PID ceiling was the field that got cut.

- `run-view.ts` — `WorkerRow` reads the two columns; `WorkerView.capacity` carries `pidsCurrent`, `pidsMax`, and a canonical `pidsSaturated` verdict at a `0.85` ceiling fraction.
- Unknown is **null, never 0**: `gaugeOrNull` writes null for the `-1` unreadable-cgroup sentinel, and a pre-migration row has no column at all. Neither may render as "0 PIDs", and neither may claim saturation — absence of evidence is not evidence.
- The verdict is derived **server-side on purpose** so consumers don't each re-derive a threshold and drift, the same reason the cell model forbids re-deriving colour from `status.state`.
- `worker-runs-table.tsx` annotates the PID reading **only when saturated**, so the normal path gains no noise.

Prod crossed `0.85` on **2026-08-01**, roughly 36 hours before the flip. This is a warning with real lead time, not a post-hoc marker.

### RED (before the change)

```
❯ src/fleet/control-plane/run-view.test.ts (43 tests | 5 failed | 38 skipped) 11ms
   × projectWorker > cgroup PID gauges > projects capacity_pids_current / capacity_pids_max onto the view
     → expected undefined to be 640 // Object.is equality
   × projectWorker > cgroup PID gauges > flags saturation at/above the threshold and not below it
     → expected undefined to be true // Object.is equality
   × projectWorker > cgroup PID gauges > reports unknown gauges as null and never as a saturated fleet
     → expected undefined to be null
   × projectWorker > cgroup PID gauges > treats a negative sentinel that reached the column as unknown
     → expected undefined to be null
```

And at the surface `/api/runs` actually serializes, with the verbatim incident reading:

```
 FAIL  createMemoizedFamilySummary > serializes the cgroup PID gauges and saturation flag for a PID-exhausted worker
AssertionError: expected { inUse: 4, available: 36, max: 40 } to deeply equal { inUse: 4, available: 36, …(4) }

  {
    "available": 36,
    "inUse": 4,
    "max": 40,
-   "pidsCurrent": 1000,
-   "pidsMax": 1000,
-   "pidsSaturated": true,
  }
```

### GREEN (after)

```
 Test Files  1 passed (1)
      Tests  5 passed | 38 skipped (43)
```

Dashboard render, same shape (RED → GREEN):

```
RED:   AssertionError: expected null to be 'false'   (data-pids-saturated absent)
       Tests  2 failed | 17 skipped (19)
GREEN: Tests  19 passed (19)
```

### Mutation check

| mutant | result |
|---|---|
| A — threshold `0.85` → `99.0` | **2 tests fail** ✅ |
| B — unknown gauge collapses to `0` instead of `null` | **4 tests fail** ✅ |
| C — dashboard PID annotation never rendered | **1 test fails** ✅ |

All restored afterwards.

---

## Patch 2 — let the d6 roll-up carry an infra error class

`E2eFullAggregateSignal` had **no `errorClass` field at all**, so `signalHasInfraErrorClass()` returned false for every `d6:<slug>` roll-up unconditionally and the U7 gray fold could never fire on it — however cleanly it fired on the cells underneath. Its `failureSummary` *contained* the classes as free text; nothing parsed that. On 2026-08-03 the roll-up — the row a naive per-slug query lands on, and the row whose `first_failure_at` the watcher quoted — read as a product outage for ~20 slugs whose demos were fine.

The per-feature `errorClass` was also being dropped on the way up to the aggregation loop, so there was nothing to aggregate even if the field had existed. It is now carried.

**Polarity is the cell model's, unchanged.** `rollupInfraErrorClass()` stamps a class only when the failure set is non-empty and **every** member is positively infra-classed. One product class, one promise-rejection, one missing script, or one *unknown* sinks the verdict. That asymmetry is the masks-real-red guard lifted to the roll-up: on 2026-07-24, 21 of 357 prod reds were infra-class against 336 product-class, so graying on absence of evidence would have hidden ~94 real failures to suppress ~6 false ones.

The launcher-error path is classified infra directly. If the browser never launched, zero product code ran, so the red carries no information about the demo — and that is the literal signature of a worker at its PID ceiling.

### RED (before)

```
 FAIL  e2e-full driver > aggregate d6:<slug> side row > roll-up infra classification
       > stamps an infra errorClass when EVERY failed feature is infra-classed
AssertionError: expected undefined to be 'driver-error' // Object.is equality

- Expected:  "driver-error"
+ Received:  undefined

 ❯ src/probes/drivers/d6-all-pills.test.ts:653:32
    653|         expect(agg.errorClass).toBe("driver-error");
    654|         expect(signalHasInfraErrorClass(agg)).toBe(true);
```

### GREEN (after)

```
 Test Files  1 passed (1)
      Tests  4 passed | 58 skipped (62)
```

Full driver file: **62 passed**.

### Mutation check

| mutant | result |
|---|---|
| D — all-infra guard → any-infra-wins | **3 tests fail** ✅ |
| E — per-feature `errorClass` not propagated | **1 test fails** ✅ |
| F — `feature-timeout` added to `INFRA_ERROR_CLASSES` (the forbidden change) | **1 test fails** ✅ |

Mutant F is the load-bearing one: the non-goal is now enforced by a test, not by a comment.

---

## Patch 3 — SKIPPED: there is no canvas-based probe predicate

The brief asked to correct any canvas-based probe predicate for the mcp-apps inline surface, since that surface uses `exportToSvg` and has no `<canvas>` element ever. **No such predicate exists**, so there was nothing to fix and I did not invent a change.

- `grep -rn canvas showcase/harness/src/` → **zero hits**.
- The only mcp-apps probe, `d5-mcp-apps.ts`, asserts on an **iframe** cascade: `[data-testid="mcp-app-iframe"]` then `iframe[sandbox]`. Correct for the real surface.
- No `integrations/*/tests/e2e/mcp-apps.spec.ts` mentions canvas.
- Every remaining `canvas` hit under `showcase/` is unrelated: shell-docs landing copy, `beautiful-todo-canvas`, and the `research-canvas` demo links.

Worth recording for whoever picks up the mcp-apps work: the sibling report measured a healthy render painting at **6–8s**, so a probe giving up before ~10s would see blank on a healthy run. `d5-mcp-apps.ts` allows 15s, which is adequate.

---

## Verification

- **Harness typecheck** — clean. Only remaining error is pre-existing at baseline: `Cannot find module 'axe-core'` (missing from the lockfile, which is why `npm ci` refuses to run in this tree).
- **Harness suite** — 3673 passed, 3 failed. All 3 (`frontend-matrix`, `d5-mapping-drift`, `d5-representatives`) fail identically on unmodified `main` in a separate checkout at the same commit. Two others initially failed on missing generated artifacts (`registry.json`) and one on a timing threshold under load; all pass once the generators run.
- **Dashboard suite** — **1409 passed, 0 failed** (via `npm test`, which runs the registry/probe-docs generators first).
- **oxfmt** — all 9 changed files correctly formatted.
- **oxlint** — 5 warnings, 0 errors; byte-identical to the same 9 files at baseline. Every warning is in pre-existing code.
- **No `as any` / `as unknown as`** anywhere in the diff.

## Not proven

- No change here was exercised against prod or staging — deliberately out of scope, and the deployed prober image (`033b0ae2`) is ahead of this tree's base (`bde2d99a5b`), so the patches are not verified against the running build.
- `e2e-parity.ts` / `d3-readiness.ts` aggregates were not audited for the same missing-`errorClass` gap.
- Whether `system:browser-pool-degraded` (green and eight weeks stale) can be made to trip on PID saturation. Separate, larger work; this PR makes the saturation *readable*, which is the prerequisite.
- Neither patch prevents the outage itself. The cure is reaping the zombies (PID-1 init), owned by a sibling PR. Patch 1 makes the outage visible ~36h earlier; patch 2 stops the roll-up misreporting it as a product failure.

### DIFF
--- a/showcase/harness/src/fleet/control-plane/d0-gone-monitor.test.ts
+++ b/showcase/harness/src/fleet/control-plane/d0-gone-monitor.test.ts
@@ -119,7 +119,16 @@ function makeFakes() {
     lastHeartbeatAt: new Date(T0).toISOString(),
     registeredAt: new Date(T0 - HOUR).toISOString(),
     currentJobId: null,
-    capacity: { inUse: 0, available: 1, max: 1 },
+    // Unknown cgroup PID gauges: these fixtures exercise gone-detection, not
+    // PID saturation.
+    capacity: {
+      inUse: 0,
+      available: 1,
+      max: 1,
+      pidsCurrent: null,
+      pidsMax: null,
+      pidsSaturated: false,
+    },
   };
 
   function liveProducer(atMs = clock): FamilySummaryResponse {
@@ -221,7 +230,16 @@ describe("isProducerLive (§2.5 acceptance)", () => {
     lastHeartbeatAt: new Date(T0).toISOString(),
     registeredAt: new Date(T0).toISOString(),
     currentJobId: null,
-    capacity: { inUse: 0, available: 1, max: 1 },
+    // Unknown cgroup PID gauges: these fixtures exercise gone-detection, not
+    // PID saturation.
+    capacity: {
+      inUse: 0,
+      available: 1,
+      max: 1,
+      pidsCurrent: null,
+      pidsMax: null,
+      pidsSaturated: false,
+    },
   });
 
   const inflightEntry = (): FamilySummaryEntry => ({
--- a/showcase/harness/src/fleet/control-plane/family-silence-monitor.test.ts
+++ b/showcase/harness/src/fleet/control-plane/family-silence-monitor.test.ts
@@ -156,7 +156,16 @@ function workerView(registeredAtMs: number): WorkerView {
     lastHeartbeatAt: iso(registeredAtMs),
     registeredAt: iso(registeredAtMs),
     currentJobId: null,
-    capacity: { inUse: 0, available: 24, max: 24 },
+    // Unknown cgroup PID gauges: these fixtures exercise silence detection,
+    // not PID saturation.
+    capacity: {
+      inUse: 0,
+      available: 24,
+      max: 24,
+      pidsCurrent: null,
+      pidsMax: null,
+      pidsSaturated: false,
+    },
   };
 }
 
--- a/showcase/harness/src/fleet/control-plane/run-view.test.ts
+++ b/showcase/harness/src/fleet/control-plane/run-view.test.ts
@@ -696,6 +696,70 @@ describe("projectWorker", () => {
     ).toBe("");
   });
 
+  // 2026-08-03 incident: the prod prober fleet sat pinned at cgroup
+  // `pids.current == pids.max == 1000` (~730 zombies) for hours and could not
+  // launch a browser, yet `/api/runs` reported `online, 4/40 contexts` because
+  // the projection dropped the PID gauges the heartbeat already persists. The
+  // context budget was never the binding constraint; the PID ceiling was.
+  describe("cgroup PID gauges (2026-08-03 PID-exhaustion blindness)", () => {
+    it("projects capacity_pids_current / capacity_pids_max onto the view", () => {
+      const projected = projectWorker(
+        workerRow({ capacity_pids_current: 640, capacity_pids_max: 1_000 }),
+        STALE_AFTER_MS,
+        NOW_MS,
+      );
+      expect(projected.capacity.pidsCurrent).toBe(640);
+      expect(projected.capacity.pidsMax).toBe(1_000);
+    });
+
+    it("flags saturation at/above the threshold and not below it", () => {
+      // 0.85 is the canonical threshold: prod crossed it on 2026-08-01,
+      // ~36h before the flip that produced 428 abort cells.
+      const at = (current: number) =>
+        projectWorker(
+          workerRow({
+            capacity_pids_current: current,
+            capacity_pids_max: 1_000,
+          }),
+          STALE_AFTER_MS,
+          NOW_MS,
+        ).capacity.pidsSaturated;
+      expect(at(1_000)).toBe(true); // the incident reading
+      expect(at(850)).toBe(true); // exactly at the threshold
+      expect(at(849)).toBe(false); // just under
+      expect(at(199)).toBe(false); // post-restart healthy reading
+    });
+
+    it("reports unknown gauges as null and never as a saturated fleet", () => {
+      // `gaugeOrNull` (worker/registration.ts) writes null for the -1
+      // "unreadable cgroup" sentinel, and a pre-migration row has no column at
+      // all. Neither may read as "0 of 0 PIDs", and neither may claim
+      // saturation — an unreadable gauge is not evidence of exhaustion.
+      const projected = projectWorker(
+        workerRow({
+          capacity_pids_current: undefined,
+          capacity_pids_max: undefined,
+        }),
+        STALE_AFTER_MS,
+        NOW_MS,
+      );
+      expect(projected.capacity.pidsCurrent).toBeNull();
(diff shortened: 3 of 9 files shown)


BenTaylorDev:
Rebased onto current main (was 1,953 behind). One conflict, in the `d6-all-pills.test.ts` import list — main had added `resolveAimockStrictHeaders`, this branch adds `rollupInfraErrorClass`; kept both. All four commits replayed, content unchanged at 9 files / +545 −7.

Verified after the rebase: harness suites green (`run-view`, `d0-gone-monitor`, `family-silence-monitor`, `d6-all-pills` — 180/180), and `tsc --noEmit` adds no errors over main. The three shell-dashboard files I could not exercise locally (vitest there fails to load in a linked worktree, on clean main too), so CI is the first real check on those.

**One thing the rebase surfaced that needs your call.** #6333 landed on 2026-08-12, after this branch was written, and it introduced its own PID-saturation threshold:

| | | |
|---|---|---|
| `fleet-health.ts:117` (main, #6333) | `DEFAULT_PID_SATURATION_RATIO = 0.75` | alarm; hysteresis clear at 0.65; env-overridable via `WORKER_PID_SATURATION_RATIO` |
| `run-view.ts:760` (this branch) | `PIDS_SATURATION_THRESHOLD = 0.85` | doc comment describes it as the "Canonical PID-saturation verdict" |

So on current main plus this branch, a worker at 0.80 of its cgroup PID ceiling pages via fleet-health while the runs read-model reports it **not** saturated. Nobody chose that — it's purely the landing order.

Paging earlier than you display may well be the right design, but the "canonical" wording forecloses that reading, and two independently-maintained ratios for one physical condition is the duplicated-constant-drift shape your own #6166 gate is built to reject.

Your call which way it goes — derive the read-model verdict from `DEFAULT_PID_SATURATION_RATIO`, or keep 0.85 and say explicitly why the display threshold differs from the alarm. I didn't want to pick one silently inside a rebase.

Rebased from stale-draft triage; no content changed beyond the import conflict above.

PR #6340, opened 2026-08-03
feat(skills): add setup-slack-channel and split it from copilotkit-channels

jerelvelarde:
## Summary

Brings the `setup-slack-channel` skill into the set that `copilotkit skills install` distributes, and narrows both it and `copilotkit-channels` so they stop matching the same request.

## Why

The skill was merged into [CopilotKit/channels-sdk#7](https://github.com/CopilotKit/channels-sdk/pull/7), which makes it reachable from a channels-sdk checkout and nowhere else. `copilotkit skills install` reads `CopilotKit/CopilotKit/skills`, so until the skill lives here, no CLI user can get it.

No CLI change is needed to pick it up: skill names are free-form (`--skill` is validated by regex only, with no allowlist) and the default install is `--skill *`.

## How

**The skill is copied from the merged channels-sdk source**, with one addition: a `version: 1.0.0` frontmatter field, so it matches the other standalone skills in this repo (every one of them carries a version; only the package-mirrored skills omit it). The body, references, and manifest asset are otherwise unchanged. It stays browser-first and makes no reference to the `copilotkit channels` commands, per the call to defer those to the web app until they have more real-world usage.

**Both descriptions are narrowed** so the trigger overlap is gone:

| Skill | Owns | Hands off |
| --- | --- | --- |
| `copilotkit-channels` | The code half — declaring, wiring, and customising a Channel. Assumes the provider app exists | First-time Slack setup → `setup-slack-channel` |
| `setup-slack-channel` | The provider half — the Slack app, its tokens, attaching it to a Channel | Code questions → `copilotkit-channels` |

Before this, both matched "connect my agent to Slack" and the agent picked whichever it read first. That collision was invisible in channels-sdk, where `setup-slack-channel` is the only skill present; it becomes live the moment both ship in the installed set.

## Notes for reviewers

- **Scope is Slack only.** A Teams sibling is deliberately a separate pass.
- **No existing skill content changes** — the `copilotkit-channels` diff is its frontmatter description and nothing else.
- Two known follow-ups are tracked and intentionally not addressed here: [channels-sdk#9](https://github.com/CopilotKit/channels-sdk/issues/9) (CLI-capability claims, deferred by decision) and [channels-sdk#2](https://github.com/CopilotKit/channels-sdk/issues/2) (`build-channels-bot` staleness, unrelated).

Refs CopilotKit/channels-sdk#10

## Added during the `main` merge

Resolving the conflict surfaced a second, unrelated problem that had to be fixed for this PR to be safe to land.

`skills/setup-slack-channel` is a **standalone** skill — it has no `packages/*/skills/` source. `scripts/sync-plugin-skills.ts` treats any such directory as an orphan unless it is listed in `RESERVED_LIFECYCLE_SLUGS`, which this one was not. Verified against the pre-fix script:

- `pnpm check:plugin-skills` → exit 1, `orphan file(s) in mirror: skills/setup-slack-channel`
- `pnpm sync:plugin-skills` (write mode) → **recursively deleted all 8 files of the new skill**

The `plugin-skills-check` workflow's path filter does not match `skills/setup-slack-channel/**`, so this PR would not have caught it — it would have gone red on the next unrelated PR that touched the script or a package skill, or silently eaten the skill on the next sync run.

Fix is two lines: add the slug to `RESERVED_LIFECYCLE_SLUGS`, and move the paired `size` assertion in `scripts/__tests__/sync-plugin-skills.test.ts` from 9 to 10. The test file already documents this exact hazard in a comment.

### Conflict resolution

The only conflict was the `copilotkit-channels` frontmatter description, which #6320 rewrote in parallel. The two sides disagreed about Teams: this branch said the skill "assumes the provider app already exists", while #6320 established that Teams provider setup **is** this skill's job because the CLI or dashboard wizard performs it. The resolution keeps this branch's code-half framing and the `setup-slack-channel` handoff, but scopes that handoff to *first-time Slack app creation* only — so it contradicts neither #6320's Teams sections nor the Slack provider troubleshooting that stays in this file. Took #6320's `version: 1.1.0`.


jerelvelarde:
Pairs with CopilotKit/Intelligence#732, which adds the canonical agent setup prompt and delivers it from `copilotkit channels prompt` and `copilotkit skills onboard --channels`.

Order matters: that PR's prompt points at `setup-slack-channel` by name, so it is only useful once this one merges and the skill is in the set `copilotkit skills install` distributes.

BenTaylorDev:
Reviewed alongside CopilotKit/Intelligence#732. The trigger split is the right lever and the table in the description makes the ownership boundary legible — that part I'd take as-is. One blocker, two cheap content fixes, one scoping question, and one thing I'd park.

## Blocker — the next stable release deletes this skill

`skills/setup-slack-channel` is a new top-level dir in `skills/` that is not in `RESERVED_LIFECYCLE_SLUGS` (`scripts/sync-plugin-skills.ts:9-20`) and has no `packages/*/skills/setup-slack-channel` source, so the full-directory orphan scan flags it. Verified locally with the dir present:

```
$ pnpm check:plugin-skills
orphan file(s) in mirror (source removed):
  skills/setup-slack-channel
run: pnpm sync:plugin-skills
ELIFECYCLE  Command failed with exit code 1
```

Write mode doesn't just report it — it prunes (`scripts/sync-plugin-skills.ts:182-188`, `rm` with `recursive: true`), and `stable-release.yml:102` runs write mode and then commits the working tree into the release PR. So this survives merge and dies at the next stable release. Because `copilotkit skills install` reads `CopilotKit/CopilotKit/skills` off main unpinned, it disappears for every user at that moment — while Intelligence#732's prompt keeps pointing at it.

It's silent because `plugin-skills-check.yml`'s `paths` filter only covers `skills/runtime|react-core|a2ui-renderer/**`, so this PR's CI never runs the check. Second-order: it also arms a failure for the next unrelated PR that *does* touch those paths.

Fix is one line — add `"setup-slack-channel"` to `RESERVED_LIFECYCLE_SLUGS`. Worth also adding `skills/setup-slack-channel/**` to the workflow's `paths` so the guard covers it going forward.

## `interactivity.is_enabled: false` is no longer true

From the "Know what the managed adapter does not deliver" paragraph:

> The wizard's generated manifest declares **no `slash_commands`** and sets `interactivity.is_enabled: false` … **not slash commands and not interactive component payloads**.

Half of that holds and half doesn't, as of Intelligence `91c8dc8d refactor(channels): share managed Channel setup generators` (Aug 1, two days before this PR):

- `libs/channels-setup/src/slack.ts:209-213` (Intelligence) emits `interactivity.is_enabled: true` with an interactivity request URL — and that's the generator now shared by *both* the wizard and the CLI, so it's what every managed manifest looks like.
- `apps/app-api/src/routes/channels-routes.ts:868` handles `block_actions`, so HITL buttons/selects **are** delivered.
- Still correct: no `slash_commands`, and `view_submission` (modals) is explicitly not handled — `apps/app-api/src/channels/slack-ingress.ts:1050`. So the OpenTag `onCommand`/`onModalSubmit` warning is accurate and worth keeping.

Erring conservative is the safe direction, but as written it tells a developer a working feature is broken. Cheap split: keep the slash-command and `onModalSubmit` claims, drop "not interactive component payloads at all."

## The browser-only framing is stronger than the decision behind it

The deferral itself isn't in question — routing people away from CLI commands we haven't had time to test is deliberate. It's the phrasing I'd change:

> creating a Channel, attaching a Slack adapter, issuing an API key, and reading Channel status exist only in the Intelligence browser experience. Never imply a command can complete a step that is browser-only, and never invent a command name to fill the gap

That reads as an architectural claim, and `copilotkit channels add` does create the Channel server-side, write the provider manifest, and attach the adapter (`apps/cli/README.md`, Managed Channels) — with `--json` built specifically for agents. So an agent that follows the instruction to check `--help` when unsure finds the commands and hits a contradiction, and the sentence has to be re-audited the moment the commands are promoted.

Same routing outcome, no rot:

> The supported path for v1 is the Intelligence browser experience. `copilotkit channels` lists commands for Channel creation, adapter attach, and key issuance — **do not use them here**; they are not hardened for this workflow yet. Never invent a command name to fill a gap.

## Scoping — this is written for a channels-sdk checkout, and it's about to install into customer repos

Phase 0 tells the agent to find an OpenTag checkout, or run `git submodule update --init examples/OpenTag` "in this repo," or clone OpenTag and "treat that checkout as the source of truth." That's exactly right inside channels-sdk, where OpenTag is a submodule. After `copilotkit skills install`, "this repo" is the developer's own app: the submodule command fails, and the skill instructs the agent to walk away from their project. Same for `pnpm runtime`, `app/channel.tsx`, `app/env.ts`, `INTELLIGENCE_CHANNEL_NAME`, and the port-8123 Python agent — OpenTag conventions, not what `copilotkit init` scaffolds. Post-Intelligence#714 the CLI starter already ships a channel host, so Phases 0 and 4 are wrong for the most likely reader.

I don't think that needs a Phase 0 rewrite before this ships. The cheap version is the lever this PR already pulls: narrow the frontmatter `description` so the skill only fires in the context it was written for, rather than on any "connect my agent to Slack." Broad firing is the thing most likely to generate a support thread.

## Parked, not asking for it here

The skill now exists byte-identical in two repos with nothing keeping them in sync, and the rationale given in Intelligence#732 for the prompt being a pointer rather than a restatement — that the skill is "versioned alongside the SDK behaviours it asserts" — is what the copy gives up. This copy is versioned against neither channels-sdk nor Intelligence, which is why the two stale claims above are already in it, and why the `@copilotkit/channels@0.6.0` pin has no gate.

Not a merge blocker and not a tomorrow problem, but worth a ticket before it settles: pick one home, and decide whether a *managed*-Channels skill belongs in the OSS repo at all when the code and manifest generator it asserts live in Intelligence.

jerelvelarde:
All four addressed. Head is `77e5415c`.

## Blocker — orphan prune

Fixed, and independently hit the same thing before this review landed while merging `main`, with the same reproduction. `"setup-slack-channel"` is in `RESERVED_LIFECYCLE_SLUGS`, and the paired `size` assertion in `sync-plugin-skills.test.ts` moves 9 → 10. Confirmed both directions against the pre-fix script:

- check mode → exit 1, `orphan file(s) in mirror: skills/setup-slack-channel`
- write mode → **all 8 files of the skill deleted**

Post-fix: check exits 0, write mode is a no-op, all 8 files survive, 10/10 unit tests pass.

On the `paths` filter, I went slightly further than the suggestion — `skills/**` rather than adding the one slug. The orphan scan reads the whole mirror, so enumerating individual directories is precisely what let this PR's own blocker go untested, and adding one slug would leave the identical trap armed for the next standalone skill. Happy to narrow it to just `skills/setup-slack-channel/**` if you'd rather keep the trigger tight.

## `interactivity.is_enabled: false`

Corrected. Verified each leg in the Intelligence tree rather than taking it secondhand:

- `libs/channels-setup/src/slack.ts:209-211` — `is_enabled: true` plus an interactivity request URL
- `apps/app-api/src/routes/channels-routes.ts:868` — `block_actions` handled
- `apps/app-api/src/channels/slack-ingress.ts:1050` — `view_submission` explicitly not handled
- no `slash_commands` anywhere in the generator

Worth noting the skill's own bundled `assets/slack-app-manifest.yaml:98` already said `is_enabled: true`, so the prose was contradicting a file shipped beside it.

Fixed in all three places that claimed otherwise (`SKILL.md`, `references/slack-workspace-and-app.md`, `references/troubleshooting.md`). Kept the `slash_commands` and `onModalSubmit` warnings. The troubleshooting entry now reframes item 6 around slash commands and modals, and says a dead button is a **real failure worth debugging** — as written it was actively teaching readers to write off a working feature.

## Browser-only framing

Took your wording almost verbatim. It is now a routing instruction rather than an architectural claim, so it no longer contradicts `--help` for an agent that was told to check it.

## Scoping

Took the cheap lever, no Phase 0 rewrite. The frontmatter description now names the assumed context and the conventions that come with it (`app/channel.tsx`, `app/env.ts`, `INTELLIGENCE_CHANNEL_NAME`, the port-8123 agent), and says explicitly that this is *not* what `copilotkit init` scaffolds. That should stop it firing on any "connect my agent to Slack" once it installs into customer repos.

## Parked item

Agreed, and not touched here. Two things that came up while fixing the above and are worth folding into that ticket:

1. The bot's `style: auto-fix formatting` commit has already reformatted the copy in this repo, so the two are no longer byte-identical — the drift you predicted started on day one, before either repo changed any content.
2. The `@copilotkit/channels@0.6.0` / `@copilotkit/runtime@1.65.0` pins in the skill are the concrete thing with no gate. Whatever home wins, that is the assertion most likely to go stale silently.

Say the word and I'll open it.

## One deliberate deviation

I also added `version: 1.0.0` to the skill's frontmatter — every other standalone skill in `RESERVED_LIFECYCLE_SLUGS` carries one and only the package-mirrored skills omit it. Nothing enforces it; it does mean the copy is no longer byte-identical to channels-sdk, and the PR description is updated to say so rather than still claiming a clean `diff -r`.

BenTaylorDev:
Verified `77e5415c` rather than taking the summary on trust. The blocker fix, the trigger narrowing, and the `SKILL.md` / `troubleshooting.md` corrections all check out:

- `"setup-slack-channel"` in `RESERVED_LIFECYCLE_SLUGS`, with the paired `size).toBe(10)` assertion
- `skills/**` in both `paths` blocks
- `SKILL.md:148-152` and `troubleshooting.md:114-118` corrected, `slash_commands` and `onModalSubmit` warnings intact
- frontmatter names the OpenTag/channels-sdk assumption and says explicitly it is not what `copilotkit init` scaffolds

And keep `skills/**` — don't narrow it to the one slug. Enumerating individual directories is exactly what let this PR's own blocker go untested; the generalization is the right call and I'd have asked for it if you hadn't made it.

## One stale row survived, in the file you corrected

`references/slack-workspace-and-app.md:109`, in the table headed **"Settings that must stay as the manifest sets them"**:

| Setting | Expected | Why |
| --- | --- | --- |
| **Interactivity** | disabled | The managed adapter does not deliver interactive payloads. Enabling it does not make buttons work, it just implies they should. |

Line 55 of the same file now says the opposite: *"It does set `interactivity.is_enabled: true` … that one is deliberate, and it is what makes HITL buttons and selects work."* The two are 50 lines apart.

This one is worse than the original bug rather than equivalent to it. The prose version told a developer a feature was broken; this is a **normative settings table** under a heading instructing the reader to keep things as the manifest sets them. An agent reconciling the Slack app against that row turns interactivity *off* — actively causing the failure the rest of the fix prevents, in the one place a reader is most likely to act mechanically.

3 of 4 sites landed; this is the fourth. One row.

## Parked ticket

Filing it — and both of your additions are going in. The `style: auto-fix formatting` commit breaking byte-identity on day one, before either repo touched content, is the most concrete evidence the drift argument is going to get. The `@copilotkit/channels@0.6.0` / `@copilotkit/runtime@1.65.0` pins are named as the specific ungated assertion.

jerelvelarde:
checking

jerelvelarde:
Fixed in `bc5f8073`. You were right that it was worse than the original, and right about why — a normative table under a keep-this-as-is heading is the one place a reader acts mechanically, and acting on that row turns interactivity off.

**There were two, not one.** My first pass grepped for `is_enabled`, which is exactly why both survivors survived: neither contains that string.

1. `references/slack-workspace-and-app.md:109` — the row you named. Now reads **enabled**, with the request URL `/api/channels/adapters/slack/interactivity`, verified against `libs/channels-setup/src/slack.ts:14-18` rather than assumed from the events path. It also notes modals stay undelivered either way, so the row is not read as "buttons and modals both work now."
2. `assets/slack-app-manifest.yaml:33-34` — a comment carrying the identical claim: *"the managed path does not deliver interactive payloads at all."* Not in your list; found by re-grepping `nteractiv` instead of the specific phrasing. It now distinguishes the direct-adapter shape (interactivity enabled, **no** request URL) from the managed one (enabled **with** an Intelligence request URL), which is the distinction that file exists to draw — and the reason the asset was a misleading reference even before this.

That second one matters for the parked ticket: this skill asserted the same wrong thing in four places across prose, a normative table, and a shipped asset, and three separate passes were needed to clear it. That is the drift cost with both copies still nominally in sync.

Swept every remaining variant afterwards — `does not deliver interactive`, `not interactive component`, `interactivity.*disabl`, `is_enabled: false`, `component handlers will` — nothing left. Gates green: `check:plugin-skills` exit 0, 10/10 unit tests, oxfmt clean across all 106 files, manifest still parses with `interactivity.is_enabled: true`.

Thanks for re-verifying rather than taking the summary — the table row is not something a passing skim would have caught, and it would have shipped as an instruction to break the feature.

And noted on keeping `skills/**`.

samjulien (approved):

PR #6346, opened 2026-08-03
feat(react-native): stream tool-call renders, add useRenderToolCall

davidmckayv:
## What

Two related gaps in the React Native SDK, both found while building an in-car concierge app (an agent composes the entire screen, so this is the whole product surface there).

### 1. Tool-call renders never streamed on RN

`CopilotChat` parsed tool arguments with `JSON.parse`:

```ts
args = JSON.parse(tc.function.arguments || "{}");   // throws on every delta
```

While the model is still writing the call the argument string is incomplete JSON, so this threw, warned, and fell back to `{}`. A component registered with `useRenderTool` / `useComponent` therefore rendered **nothing** until the call completed — no progressive build at all, while the web core has had it (`partialJSONParse` + an `inProgress` state).

Now uses `partialJSONParse`, already exported from `@copilotkit/shared`, which this package already depends on. Incomplete JSON is itself the signal that the call is mid-write, so no new stream state was needed.

### 2. Non-chat surfaces couldn't render a registered component at all

`CopilotChat` renders tool calls inline, which is right for a chat. But plenty of RN surfaces aren't chats — an in-car stage, a kiosk, a dashboard — and there was **no public way** to render a registered component outside it. react-core's `useRenderToolCall` is deliberately not re-exported (`src/index.ts`: "depends on DOM elements via `DefaultToolCallRenderer`"), and excluding it dropped the capability along with the DOM dependency.

Adds an RN-native `useRenderToolCall()` reading this package's own registry, so one registration works both in the chat and on a custom surface:

```tsx
const renderToolCall = useRenderToolCall();
return <View>{toolCalls.map((tc) => renderToolCall({ toolCall: tc }))}</View>;
```

This is also what makes `useComponent` usable outside `CopilotChat` on React Native.

## Breaking-ish

`RenderToolProps.status` gains `"inProgress"` and `args` becomes `Partial<T>`, reflecting that arguments are incomplete while streaming. Renderers that switch exhaustively on `status` need an `"inProgress"` branch. Flagged in the changeset as a minor.

## Testing

- 7 new tests for `useRenderToolCall`: registered/unregistered lookup, partial args at `inProgress`, a fragment with no complete key, `executing`, `complete` with result pass-through, and absent arguments.
- Full package suite: **258 passing**, including the existing `CopilotChat` and edge-case tests.
- `check-types` clean, `build` clean.
- **Verified end to end in a running RN app.** That app now renders its entire stage through `useComponent` + `useRenderToolCall`: seven registered panels, no `composeStage`. Captured on device, 1.5s into a four-panel command the stage shows the first two panels with the spinner still running, the rest arriving after — i.e. it builds progressively instead of appearing at once. Two app-side changes were needed and are worth knowing about when adopting this: anything that hides the surface for the duration of a run (a scrim, or parking content off-screen until completion) will mask the streaming, and renderers must tolerate partial props.


davidmckayv:
Closing this. The work it was written for is complete and this is not being pursued further.

The branch is retained if the react-native parity fixes are wanted later: streaming tool-call renders via partialJSONParse, an RN-native useRenderToolCall, and an RN useComponent that registers where the RN renderers actually read.

PR #6347, opened 2026-08-03
fix(channels): recover from transient gateway outages

tylerslaton:
## What changed

- Mark initial gateway HTTP 5xx and transient transport failures as retryable.
- Retry initial managed Channel activation with exponential backoff from 1 second to a 30-second cap until it connects or the manager stops.
- Preserve retry hints from `gateway_draining` join replies and retry initial join timeouts.
- Keep HTTP 4xx and NXDOMAIN failures terminal.
- Back off established-session outage reminders from 30 seconds to a 15-minute cap while Phoenix continues reconnecting.

## Why

The OpenTag Railway runtime saw the gateway host return HTTP 502 during an outage. Established Phoenix sessions keep retrying, but a runtime that starts during the outage stops after its one initial connect window. It cannot recover when the gateway comes back unless the process restarts. Fixed 30-second reminder logs also flood long outages.

The gateway drain work now rejects new joins with a structured retryable response. The client must preserve that response so the runtime can retry instead of leaving the Channel in a terminal error state.

## Companion change

CopilotKit/OpenTag#25 keeps the Railway HTTP server alive while an initial Channel retry is pending. OpenTag must consume a CopilotKit release containing this PR before that companion change can recover by itself.

## Validation

- `pnpm nx run-many -t test,check-types,build -p @copilotkit/runtime,@copilotkit/channels-intelligence`
- `pnpm nx run-many -t publint,attw -p @copilotkit/runtime,@copilotkit/channels-intelligence`
- pre-commit tests and package checks for all affected projects
- `pnpm exec oxfmt --check` on all five changed files
- `pnpm exec oxlint` on all five changed files
- `git diff --check`

### DIFF
--- a/packages/channels-intelligence/src/realtime-gateway.ts
+++ b/packages/channels-intelligence/src/realtime-gateway.ts
@@ -82,11 +82,10 @@ export interface ConnectRealtimeGatewayOptions {
    *
    * The window exists rather than failing on the first transport error because
    * a refused connect is often just a gateway that has not finished booting,
-   * and a supervising `ChannelManager` does NOT retry activation — a
-   * fail-on-first-error rule would turn a boot race into a permanently errored
-   * Channel. The default is deliberately long enough to outlast a rolling
-   * restart of the gateway. Causes that CANNOT resolve themselves (a host that
-   * does not exist) skip the window entirely; see {@link diagnoseEndpoint}.
+   * and a supervising `ChannelManager` retries classified transient failures.
+   * The window avoids opening a new socket for every failed attempt during a
+   * short boot race. Causes that CANNOT resolve themselves (a host that does
+   * not exist) skip the window entirely; see {@link diagnoseEndpoint}.
    */
   connectTimeoutMs?: number;
   /**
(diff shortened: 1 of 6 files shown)


maxkorp (commented):
The startup-retry approach looks sound for the motivating HTTP 502 case. I found two retry-classification edge cases that should be covered before this is considered complete:

packages/channels-intelligence/src/realtime-gateway.ts
```diff
@@ -570,12 +573,19 @@ export async function connectRealtimeGateway(
           ? await diagnosis
           : await startDiagnosis();
       const cause = probed?.cause ?? lastTransportRaw;
+      const retryable =
+        probed === undefined ||
+        (!probed.nonRetryable &&
+          (probed.status === undefined || probed.status >= 500));
```
maxkorp: A native `ENOTFOUND`/`EAI_NONAME` currently reaches `failUnreachable()` before `startDiagnosis()`, so `diagnosis` is still undefined here. That makes `probed === undefined` true and marks the explicitly non-retryable DNS failure as retryable, causing the manager to retry it forever. Could we make `lastTransportCode` authoritative for `NON_RETRYABLE_TRANSPORT_CODES` here and add a direct `ENOTFOUND` regression test?

packages/runtime/src/v2/runtime/core/channel-manager.ts
```diff
@@ -715,6 +723,16 @@ function isSetupRequired(err: unknown): boolean {
   );
 }
 
+/** Whether a failed initial activation can recover without new configuration. */
+function isRetryableActivationError(err: unknown): boolean {
+  return (
+    typeof err === "object" &&
+    err !== null &&
```
maxkorp: This only retries `GATEWAY_UNREACHABLE`, but an initial socket can connect and then receive a structured retryable join rejection such as `{ reason: "gateway_draining", retryable: true }`. `connectRealtimeGateway()` currently converts that reply (and an initial join timeout) to a plain terminal `Error`, so activation will not recover from that transient case. Could we preserve/recognize the retryable join classification and add a recovery test for it?

MikeRyanDev (approved):

AlemTuzlak (approved):

PR #6355, opened 2026-08-04
Port the Pydantic AI example to Pydantic AI v2

dsfaccini:
This pull request was posted by Claude Code using claude-opus-5 on behalf of David. David has not reviewed this diff.

`examples/integrations/pydantic-ai` only runs on Pydantic AI **v1**. `Agent.to_ag_ui()`, `AGUIApp`, and the `pydantic_ai.ag_ui` shim were all removed in v2 (pydantic/pydantic-ai#5464, announced in pydantic/pydantic-ai#5345). The example pins `pydantic-ai-slim==1.0.18`, so anyone installing it against current Pydantic AI (2.22.0) fails at import.

This ports it to v2.

## Changes

1. `agent/pyproject.toml` — `pydantic-ai-slim[ag-ui,openai]>=2.0.0`, `ag-ui-protocol>=0.1.19`
2. `agent/src/agent.py` — `StateDeps` moved from `pydantic_ai.ag_ui` to `pydantic_ai.ui`
3. `agent/src/main.py` — serve via `AGUIAdapter.dispatch_request` on a Starlette route
4. `agent/uv.lock` — relocked (resolves `pydantic-ai-slim` 2.22.0, `ag-ui-protocol` 0.1.19)

## One fix beyond the mechanical port

The old wiring built the app once around a single shared `StateDeps` instance. `dispatch_request` mutates `deps.state` with the state the client sends, so one shared instance lets state leak between threads, channels and users — which matters more for Channels than it did for a single browser tab. Each request now gets its own `replace(deps)` copy, matching the pattern in [Pydantic AI's own AG-UI examples](https://github.com/pydantic/pydantic-ai/blob/main/examples/pydantic_ai_examples/ag_ui/api/shared_state.py).

<details><summary>Verified end to end</summary>

`uv sync` + a request through the actual ASGI app (model overridden with `TestModel` so no API call), with a Channels-shaped payload (`threadId`, `state`, `forwardedProps`):

```
health: 200 {'status': 'ok'}
POST / -> 200 text/event-stream; charset=utf-8
event types: ['RUN_STARTED', 'TOOL_CALL_START', 'TOOL_CALL_END', 'TOOL_CALL_START', 'TOOL_CALL_ARGS',
 'TOOL_CALL_END', ..., 'TOOL_CALL_RESULT', 'STATE_SNAPSHOT', 'TOOL_CALL_RESULT', 'STATE_SNAPSHOT',
 'TEXT_MESSAGE_START', 'TEXT_MESSAGE_CONTENT', ..., 'TEXT_MESSAGE_END', 'RUN_FINISHED']
threadId echoed: slack-C123-thread-1
tool call: get_proverbs
tool call: add_proverbs
tool call: set_proverbs
tool call: get_weather
```

The emitted event set is exactly what `channels-slack`'s `RunRenderer` subscribes to (`RUN_*`, `TEXT_MESSAGE_*`, `TOOL_CALL_*`), plus `STATE_SNAPSHOT` which it ignores.

</details>

<details><summary>Note on <code>ag-ui-protocol</code>: 0.1.19, not 0.1.18</summary>

Not required by this port, but worth pinning forward: typed multimodal input content (`ImageInputContent` &c.) landed in 0.1.15 and the interrupt lifecycle in 0.1.19. We found that a Pydantic AI install below 0.1.15 rejects an inbound image attachment with a 422 rather than skipping it — so a Channels gateway forwarding a Slack image needs the newer floor. We're tracking that on our side.

</details>

### DIFF
--- a/examples/integrations/pydantic-ai/agent/pyproject.toml
+++ b/examples/integrations/pydantic-ai/agent/pyproject.toml
@@ -6,8 +6,8 @@ readme = "README.md"
 requires-python = ">=3.12"
 dependencies = [
   "uvicorn",
-  "pydantic-ai-slim[ag-ui,openai]>=2.0.0",
-  "ag-ui-protocol>=0.1.19",
+  "pydantic-ai-slim[ag-ui,openai]==2.22.0",
+  "ag-ui-protocol==0.1.19",
   "python-dotenv",
   "logfire>=4.10.0",
 ]
--- a/examples/integrations/pydantic-ai/agent/pyproject.toml
+++ b/examples/integrations/pydantic-ai/agent/pyproject.toml
@@ -6,8 +6,8 @@ readme = "README.md"
 requires-python = ">=3.12"
 dependencies = [
   "uvicorn",
-  "pydantic-ai-slim[ag-ui,openai]==1.0.18",
-  "ag-ui-protocol==0.1.18",
+  "pydantic-ai-slim[ag-ui,openai]>=2.0.0",
+  "ag-ui-protocol>=0.1.19",
   "python-dotenv",
   "logfire>=4.10.0",
 ]
--- a/examples/integrations/pydantic-ai/agent/src/main.py
+++ b/examples/integrations/pydantic-ai/agent/src/main.py
@@ -1,5 +1,3 @@
-from dataclasses import replace
-
 from agent import ProverbsState, StateDeps, agent
 from pydantic_ai.ui.ag_ui import AGUIAdapter
 from starlette.applications import Starlette
@@ -7,14 +5,13 @@ from starlette.requests import Request
 from starlette.responses import JSONResponse, Response
 from starlette.routing import Route
 
-# Template holding the initial state; each request gets its own copy.
-deps = StateDeps(ProverbsState())
-
 
 async def run_agent(request: Request) -> Response:
-    # `dispatch_request` mutates `deps.state` with the state sent by the client, so give every
-    # request its own copy — otherwise state leaks between threads, channels and users.
-    return await AGUIAdapter.dispatch_request(request, agent=agent, deps=replace(deps))
+    # Fresh deps per request: `dispatch_request` writes the client's state into
+    # `deps.state`, so a shared instance leaks state between threads, channels and users.
+    return await AGUIAdapter.dispatch_request(
+        request, agent=agent, deps=StateDeps(ProverbsState())
+    )
 
 
 async def health(request):
--- a/examples/integrations/pydantic-ai/agent/src/main.py
+++ b/examples/integrations/pydantic-ai/agent/src/main.py
@@ -1,9 +1,20 @@
-from agent import ProverbsState, StateDeps, agent
-from starlette.applications import Starlette
-from starlette.responses import JSONResponse
-from starlette.routing import Mount, Route
+from dataclasses import replace
 
-agui_app = agent.to_ag_ui(deps=StateDeps(ProverbsState()))
+from agent import ProverbsState, StateDeps, agent
+from pydantic_ai.ui.ag_ui import AGUIAdapter
+from starlette.applications import Starlette
+from starlette.requests import Request
+from starlette.responses import JSONResponse, Response
+from starlette.routing import Route
+
+# Template holding the initial state; each request gets its own copy.
+deps = StateDeps(ProverbsState())
+
+
+async def run_agent(request: Request) -> Response:
+    # `dispatch_request` mutates `deps.state` with the state sent by the client, so give every
+    # request its own copy — otherwise state leaks between threads, channels and users.
+    return await AGUIAdapter.dispatch_request(request, agent=agent, deps=replace(deps))
 
 
 async def health(request):
@@ -13,7 +24,7 @@ async def health(request):
 app = Starlette(
     routes=[
         Route("/health", health),
-        Mount("/", app=agui_app),
+        Route("/", run_agent, methods=["POST"]),
     ]
 )
 
--- a/examples/integrations/pydantic-ai/agent/uv.lock
+++ b/examples/integrations/pydantic-ai/agent/uv.lock
@@ -28,9 +28,9 @@ dependencies = [
 
 [package.metadata]
 requires-dist = [
-    { name = "ag-ui-protocol", specifier = ">=0.1.19" },
+    { name = "ag-ui-protocol", specifier = "==0.1.19" },
     { name = "logfire", specifier = ">=4.10.0" },
-    { name = "pydantic-ai-slim", extras = ["ag-ui", "openai"], specifier = ">=2.0.0" },
+    { name = "pydantic-ai-slim", extras = ["ag-ui", "openai"], specifier = "==2.22.0" },
     { name = "python-dotenv" },
     { name = "uvicorn" },
 ]
--- a/examples/integrations/pydantic-ai/agent/uv.lock
+++ b/examples/integrations/pydantic-ai/agent/uv.lock
@@ -4,14 +4,14 @@ requires-python = ">=3.12"
 
 [[package]]
 name = "ag-ui-protocol"
-version = "0.1.18"
+version = "0.1.19"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "pydantic" },
 ]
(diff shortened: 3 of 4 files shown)


examples/integrations/pydantic-ai/agent/pyproject.toml
```diff
@@ -6,8 +6,8 @@ readme = "README.md"
 requires-python = ">=3.12"
 dependencies = [
   "uvicorn",
-  "pydantic-ai-slim[ag-ui,openai]==1.0.18",
-  "ag-ui-protocol==0.1.18",
+  "pydantic-ai-slim[ag-ui,openai]>=2.0.0",
+  "ag-ui-protocol>=0.1.19",
```
contextablemark: Open-ended floors here re-create the exact failure this PR is fixing: `>=2.0.0` installs 3.x the day it ships, and `ag-ui-protocol` is pre-1.0 so its *minors* are breaking too.

That bites harder than usual because `docker/Dockerfile.agent:16` runs bare `uv sync` with no `--frozen` (unlike `Dockerfile:42`), so a stale lock silently re-resolves forward in the agent image.

It's also the only fully open-ended primary-SDK spec left in the starter fleet, and the fleet pins for reproducibility. Every sibling exact-pins: `claude-agent-sdk==0.2.104`, `langgraph==1.1.6`, `crewai==0.130.0`, `strands-agents[OpenAI]==1.18.0`, and `ag-ui-protocol==0.1.18` across adk / adk-angular / crewai-crews / langgraph-fastapi / langgraph-python / strands-python, with `==0.1.19` in claude-sdk-python. Two commits put this file on that standard on purpose — `f630372347 fix(examples): align starter agent Python pins to fleet standard`.

`==2.22.0` is what `uv.lock` already resolves; `==0.1.19` matches claude-sdk-python:

```suggestion
  "pydantic-ai-slim[ag-ui,openai]==2.22.0",
  "ag-ui-protocol==0.1.19",
```

**Pair this with the `uv.lock` suggestion below** — `uv.lock:31,33` record these specifiers verbatim, so applying only this one makes the lock stale and `uv sync --frozen` in `Dockerfile:42` fails the app image build.

contextablemark (commented):
The port itself checks out — I verified against upstream `main` that `StateDeps` really does move to `pydantic_ai.ui`, that `pydantic_ai/ag_ui.py` is gone, and that `AGUIAdapter.dispatch_request(request, *, agent, deps)` is a public classmethod returning a Starlette `Response`. `smoke-starter (pydantic-ai)` is green, so the full path (Next.js app → runtime → new route → pydantic-ai 2.22.0 → aimock) works.

Three inline suggestions below. None is a live bug. The first two are a pair — changing the specifiers in `pyproject.toml` invalidates `uv.lock`'s `requires-dist`, which fails `uv sync --frozen` at `Dockerfile:42`, so apply both (resolved versions don't move; 2.22.0 and 0.1.19 are already what's locked). Worth running `cd agent && uv lock` afterwards to confirm it's a no-op.

Separately, and explicitly not blocking: the docs are the surface actually broken for users today. `showcase/shell-docs/src/content/docs/integrations/pydantic-ai/quickstart.mdx:131` says `uv add 'pydantic-ai-slim[ag-ui]' ... starlette==0.45.3` — but `pydantic-ai-slim[ag-ui]` 2.22.0 requires `starlette>=0.46.2`, so that's a hard resolver conflict before you even reach `agent.to_ag_ui()` on line 156. Same pattern across 7 more doc pages, plus `examples/showcases/pydantic-ai-todos/` and `examples/canvas/pydantic-ai/` (the latter unpinned with no lockfile at all). Tracked separately rather than growing this PR.

examples/integrations/pydantic-ai/agent/uv.lock
```diff
@@ -28,9 +28,9 @@ dependencies = [
 
 [package.metadata]
 requires-dist = [
-    { name = "ag-ui-protocol", specifier = "==0.1.18" },
+    { name = "ag-ui-protocol", specifier = ">=0.1.19" },
     { name = "logfire", specifier = ">=4.10.0" },
-    { name = "pydantic-ai-slim", extras = ["ag-ui", "openai"], specifier = "==1.0.18" },
+    { name = "pydantic-ai-slim", extras = ["ag-ui", "openai"], specifier = ">=2.0.0" },
```
contextablemark: Companion to the `pyproject.toml` suggestion — `requires-dist` mirrors the pyproject specifiers, and `uv sync --frozen` (`Dockerfile:42`) fails if they drift.

Only the specifier strings change; every resolved version and hash stays put, because 2.22.0 and 0.1.19 are already what's locked. `logfire` is unchanged and included only to keep this a single contiguous apply.

```suggestion
    { name = "ag-ui-protocol", specifier = "==0.1.19" },
    { name = "logfire", specifier = ">=4.10.0" },
    { name = "pydantic-ai-slim", extras = ["ag-ui", "openai"], specifier = "==2.22.0" },
```

`cd agent && uv lock` should then be a no-op — worth running to confirm rather than trusting the hand edit.

examples/integrations/pydantic-ai/agent/src/main.py
```diff
@@ -1,9 +1,20 @@
...
+
+# Template holding the initial state; each request gets its own copy.
+deps = StateDeps(ProverbsState())
+
 
-agui_app = agent.to_ag_ui(deps=StateDeps(ProverbsState()))
+async def run_agent(request: Request) -> Response:
+    # `dispatch_request` mutates `deps.state` with the state sent by the client, so give every
+    # request its own copy — otherwise state leaks between threads, channels and users.
+    return await AGUIAdapter.dispatch_request(request, agent=agent, deps=replace(deps))
```
contextablemark: Getting per-request `deps` right here matters more than it looks, because this file is about to become the repo's reference for v2.

Worth being clear that the old shape wasn't just "fine until now" — `to_ag_ui(deps=StateDeps(...))` binds one instance for the app's lifetime, and the adapter assigns `deps.state` per run. Two concurrent requests therefore clobber each other mid-run: A sets its state, starts running, B overwrites `deps.state`, and A's remaining tool calls read and mutate B's state. v1 did offer a per-request seam (`handle_ag_ui_request(agent, request, deps=...)`), but every pydantic-ai surface in this repo took the convenient one-liner instead, so the precedent is a live cross-user race rather than a pattern worth carrying forward. This starter ships a Channels host (`channels.mts`, `channel-host.mts`) driving concurrent Slack threads through one agent process, so it was reachable. Good catch, and the fix is right.

The one refinement: `dataclasses.replace(deps)` is a shallow copy — the new `StateDeps` points at the **same** `ProverbsState`. It's correct today only because `StateDeps` has exactly one field and `pydantic_ai/ui/_adapter.py` unconditionally rebinds it before every run:

```python
raw_state = self.state or {}
state = type(deps.state).model_validate(raw_state)   # fresh model, even for {}
deps.state = state
```

That safety doesn't survive the copy-paste this file invites. Add a second field to a deps class — a DB session, an httpx client, a cache, a user record — and `replace()` silently shares it, because upstream only rebinds `state`. `showcase/integrations/pydantic-ai` has 19 mounts to port, 8 of them stateful, and they'll copy whatever idiom lands here.

Constructing fresh has no such trap, is shorter, needs no comment explaining the subtlety, and drops the `dataclasses` import. Wrapped to stay under ruff's default 88 (inlining hits 100; the current line 17 is 87):

```suggestion
from agent import ProverbsState, StateDeps, agent
from pydantic_ai.ui.ag_ui import AGUIAdapter
from starlette.applications import Starlette
from starlette.requests import Request
from starlette.responses import JSONResponse, Response
from starlette.routing import Route


async def run_agent(request: Request) -> Response:
    # Fresh deps per request: `dispatch_request` writes the client's state into
    # `deps.state`, so a shared instance leaks state between threads, channels and users.
    return await AGUIAdapter.dispatch_request(
        request, agent=agent, deps=StateDeps(ProverbsState())
    )
```

Non-blocking — what's here works.

dsfaccini:
This reply was posted by Claude Code using claude-opus-5 on behalf of David.

All three applied in 4be2516.

`uv lock` after the pyproject + lock edits was a no-op: only the two `requires-dist` specifier strings changed, every resolved version and hash stayed put.

Thanks especially for the `replace()` catch. I'd read it as a copy; a shallow one that only survives because `StateDeps` has a single field the adapter rebinds is exactly the kind of thing 19 mounts would inherit without anyone noticing. Constructing fresh is better on its own terms.

Smoke-checked the example after the change: `RunAgentInput` posted through the Starlette app with the model overridden returns 200 `text/event-stream` with a clean `RUN_STARTED` -> tool calls -> `RUN_FINISHED` sequence, and a second request no longer sees the first's state.

Leaving your threads for you to resolve. Happy to take the docs `starlette==0.45.3` conflict in a follow-up if you want it on me - https://github.com/CopilotKit/CopilotKit/pull/6355 stays scoped as is.

contextablemark (approved):
Approved... thanks for taking this on, @dsfaccini !

PR #6357, opened 2026-08-04
docs: route every Channels entry point at one onboarding skill

BenTaylorDev:
## Visual

Docs `/slack` — the featured panel from #6356, now carrying the one-line pointer.

| Desktop screenshot | Mobile screenshot |
|---|---|
| docs slack desktop | docs slack mobile |

| Desktop scroll | Mobile scroll |
|---|---|
| <video src="https://github.com/CopilotKit/website/releases/download/channels-entry-points-2026-08-03/docs-slack-desktop-scroll-7f14af56.webm" controls width="400"></video> | <video src="https://github.com/CopilotKit/website/releases/download/channels-entry-points-2026-08-03/docs-slack-mobile-scroll-7f14af56.webm" controls width="240"></video> |

Docs `/teams` — same component, Teams wording, proving the frontend switch.

| Desktop screenshot | Desktop scroll |
|---|---|
| docs teams desktop | docs root |

<sub>Artifacts are hosted on a CopilotKit/website release tag rather than a new tag in this repo, to keep clear of the monorepo's release automation.</sub>

---

## Summary

Builds on #6356 (merged in here, its featured-panel treatment kept) and finishes the job it started.

#6356 made the Channels starter prompts **discoverable**. This makes them **correct**, and makes every other road say the same thing.

The onboarding workflow was written out as prose on six surfaces across three repos: the Channels overview page (Slack and Teams), the docs landing activation strip, copilotkit.ai/channels, and the channels-sdk README twice. They drifted apart, and each went stale against the CLI on its own schedule. That drift is what the dogfooding feedback was reporting:

- **"It didn't tell me which skills to install, so it dropped me into a list of god knows how many."** Neither copy button named a skill.
- **"`npx copilotkit project select` didn't work — I had an old CLI."** Only some surfaces pinned `@latest`.
- **An agent halted on `copilotkit channels`.** No such command exists in the published CLI — verified against 4.5.1, whose `--help` has no `channels` entry.
- **"Copy this prompt is super easy to miss."** The strongest prompts were behind an accordion; the two visible buttons had invisible payloads.

## What changed

- **`<ChannelsStartPrompt />`** — one shared entry point, frontend-aware, wearing #6356's featured treatment: accent panel, terminal mark, eyebrow, prominent copy action.
- **Both overview accordions are gone.** The accordion existed because the payload was twenty lines. The payload is now one sentence, so hiding it behind "Open & copy prompt" costs a click and buys nothing.
- **The activation strip emits the same pointer**, rendered on screen rather than living only in a clipboard payload.
- **Both surfaces send `promptCopied` with a `surface` property**, so the funnel can answer which road people take. The website emits its own event name with the same property for the same reason.

The prompt every surface now emits:

```
Run `npx copilotkit@latest skills install --skill setup-slack-channel -y`, then follow that
skill to build your first CopilotKit Channels agent and connect it to Slack using CopilotKit's
built-in agent.
```

## Notes for reviewers

- **`-y` respects `--skill`** — verified empirically against `copilotkit@4.5.1`: exactly one skill installs, no picker. The installer detects the coding agent it runs inside, so `--agent` is omitted.
- **The panel renders exactly what the button copies.** Two earlier shapes were wrong in instructive ways: a full monospace paragraph wrapped like a rendering bug, and a code block with the ask underneath read as a shell command with a footnote — which made a button labelled "Copy prompt" look like it was lying.
- **The `featured` Accordion variant from #6356 stays in `mdx-components.tsx`** as a shared opt-in capability, but its only two consumers were the accordions removed here, so it is currently unused. Worth a call: keep it, or drop it in a follow-up.
- **`setup-slack-channel` is the slug the combined onboarding prompt will occupy** once it lands from channels-sdk. Its frontmatter is currently scoped hard to the Slack provider half, so it needs rescoping before this reads correctly for Teams.

Companion PRs: CopilotKit/website#444 and the channels-sdk README.

## Validation

- `npx vitest run` in `showcase/shell-docs` — 54 files, 372 tests passed
- `npx tsc --noEmit` — clean
- pre-commit lint and commitlint hooks passed
- Rendered and reviewed `/slack` locally; confirmed the copied string matches the rendered one

## Update — impressions and a copy-only prompt

Since the artifacts above were first posted:

- **Impression event.** `docs.channels_activation_viewed` fires once per surface on first intersection at 50%. Both docs entry points previously emitted a copy event and nothing else, so the copy count had no denominator. `surface` values now live in a shared `CHANNELS_ACTIVATION_SURFACES` map, keeping the landing strip, the overview panel, and copilotkit.ai/channels separable inside one funnel.
- **Guarded on `typeof IntersectionObserver`.** An impression is never worth breaking a render for, and this repo's jsdom tests do not define it.
- **The prompt is no longer rendered.** Per review, the copy button carries the payload. The panel leads with the copy action directly under the heading — the prompt is not on the page, so the button is the point of the panel rather than trailing furniture — and the supporting line no longer says "paste this" next to nothing.

Screenshots and videos above were re-captured after all of it.

## Validation (current)

- `npx vitest run` in `showcase/shell-docs` — 55 files, 376 tests passed
- `npx tsc --noEmit` — clean
- pre-commit lint and commitlint hooks passed
- Reviewed `/slack` and `/teams` at 1440×900 and iPhone 13

## Companion PRs

- CopilotKit/channels-sdk#15 — **merged**
- CopilotKit/website#444 — open

---

### Blocking issue: the skill did not support the promised workflow — resolved by removing the skill from the path

The pointer is now a fetch of one hosted file, not a skill install:

```
Read https://copilotkit.ai/channels-guide.md and help the user build their first channel
```

The guide lives at `public/channels-guide.md` in CopilotKit/website and owns the whole workflow. It asks the developer which platform and which agent framework they want, which is why the pointer passes neither.

That removes the mismatch at its root rather than narrowing the copy around it. Naming the picker's channel and backend meant these pages promised coverage on `setup-slack-channel`'s behalf, and that skill is scoped to Slack, to the provider half, and to an OpenTag checkout — so Teams pointed at a workflow that does not exist and the picker implied nineteen backends it never claimed. A pointer that names nothing cannot overpromise.

`CHANNELS_ONBOARDING_SKILL`, its install command, and the per-selection prompt builders are gone; one constant serves every surface.

### Visual regression: disclosure restored, treatment re-tokenised

| Collapsed | Expanded |
|---|---|
| collapsed | expanded |

| Mobile | Teams variant |
|---|---|
| mobile | teams |

- **The expandable preview is back.** The panel is the shared featured `<Accordion>` again — reused, not re-implemented — so the overview stays compact and a reader can expand to read the exact prompt before copying.
- **Whitespace.** Collapsed height went from ~380px to ~110px with `p-4`, matching every other docs panel. The page's own introduction is visible without scrolling.
- **Colour.** You were right, and my "it is unchanged from #6356" reply missed the point. `copilotkit-ui-theme` names *"purple accent bar or stripe"* as a known wrong direction, and `copilotkit-branding` requires accent to be restrained with gradients behind content rather than as the contrast layer — the saturated `--accent` tile plus accent-mixed gradient was both at once. It now matches the `OpsPlatformCTA` idiom: `--bg-elevated`, `--border`, `--shadow-control`, accent only on a 20px glyph and the hover state. Verified light and dark.

One deliberate deviation: I dropped the "READY-TO-USE STARTER PROMPT" eyebrow on the low-noise rule, since the title says the same thing. Easy to restore if you want it.

### Behavioural bug: fixed

`copyPrompt` shared one `try` with `posthog.capture`, so a throwing client reported "Copy blocked" for a prompt already on the clipboard. Capture now sits behind the same isolated helper `ChannelsActivationStrip` uses. Two regression tests: capture throwing after a resolved write still shows "Copied"; a rejected write still shows "Copy blocked" and emits no copy event.

### Lockfile churn: dropped

Reverted to `origin/main`. It came from an `npm install` in the review worktree; no dependency changed.

### Also added

`docs.channels_activation_prompt_expanded` — the disclosure is a funnel step neither `viewed` nor `promptCopied` can see, so someone who never opened the panel was indistinguishable from someone who opened it and left.

## Validation (current)

- `npx vitest run` in `showcase/shell-docs` — 55 files, 376 tests passed
- `npx tsc --noEmit` — clean; pre-commit lint and commitlint passed
- Reviewed `/slack` and `/teams`, collapsed and expanded, light and dark, at 1440×900 and iPhone 13

### DIFF
--- a/showcase/shell-docs/package-lock.json
+++ b/showcase/shell-docs/package-lock.json
@@ -299,7 +299,6 @@
       "version": "1.10.0",
       "resolved": "https://registry.npmjs.org/@emnapi/core/-/core-1.10.0.tgz",
       "integrity": "sha512-yq6OkJ4p82CAfPl0u9mQebQHKPJkY7WrIuk205cTYnYe+k2Z8YBh11FrbRG/H6ihirqcacOgl2BIO8oyMQLeXw==",
-      "dev": true,
       "license": "MIT",
       "optional": true,
       "dependencies": {
@@ -321,7 +320,6 @@
       "version": "1.2.1",
       "resolved": "https://registry.npmjs.org/@emnapi/wasi-threads/-/wasi-threads-1.2.1.tgz",
       "integrity": "sha512-uTII7OYF+/Mes/MrcIOYp5yOtSMLBWSIoLPpcgwipoiKbli6k322tcoFsxoIIxPDqW01SQGAgko4EzZi2BNv2w==",
-      "dev": true,
       "license": "MIT",
       "optional": true,
       "dependencies": {
@@ -1418,7 +1416,6 @@
       "version": "1.1.4",
       "resolved": "https://registry.npmjs.org/@napi-rs/wasm-runtime/-/wasm-runtime-1.1.4.tgz",
       "integrity": "sha512-3NQNNgA1YSlJb/kMH1ildASP9HW7/7kYnRI2szWJaofaS1hWmbGI4H+d3+22aGzXXN9IJ+n+GiFVcGipJP18ow==",
-      "dev": true,
       "license": "MIT",
       "optional": true,
       "dependencies": {
@@ -3471,7 +3468,6 @@
       "cpu": [
         "arm64"
       ],
-      "dev": true,
       "license": "MIT",
       "optional": true,
       "os": [
@@ -3488,7 +3484,6 @@
       "cpu": [
         "arm64"
       ],
-      "dev": true,
       "license": "MIT",
       "optional": true,
       "os": [
@@ -3505,7 +3500,6 @@
       "cpu": [
         "x64"
       ],
-      "dev": true,
       "license": "MIT",
       "optional": true,
       "os": [
@@ -3522,7 +3516,6 @@
       "cpu": [
         "x64"
       ],
-      "dev": true,
       "license": "MIT",
       "optional": true,
       "os": [
@@ -3539,7 +3532,6 @@
       "cpu": [
         "arm"
       ],
-      "dev": true,
       "license": "MIT",
       "optional": true,
       "os": [
@@ -3556,7 +3548,6 @@
       "cpu": [
         "arm64"
       ],
-      "dev": true,
       "license": "MIT",
       "optional": true,
       "os": [
@@ -3573,7 +3564,6 @@
       "cpu": [
         "arm64"
       ],
-      "dev": true,
       "license": "MIT",
       "optional": true,
       "os": [
@@ -3590,7 +3580,6 @@
       "cpu": [
         "x64"
       ],
-      "dev": true,
       "license": "MIT",
       "optional": true,
       "os": [
@@ -3607,7 +3596,6 @@
       "cpu": [
         "x64"
       ],
-      "dev": true,
       "license": "MIT",
       "optional": true,
       "os": [
@@ -3632,7 +3620,6 @@
       "cpu": [
         "wasm32"
       ],
-      "dev": true,
       "license": "MIT",
       "optional": true,
       "dependencies": {
@@ -3654,7 +3641,6 @@
       "cpu": [
         "arm64"
       ],
-      "dev": true,
       "license": "MIT",
       "optional": true,
       "os": [
@@ -3671,7 +3657,6 @@
       "cpu": [
         "x64"
       ],
-      "dev": true,
       "license": "MIT",
       "optional": true,
       "os": [
@@ -3748,7 +3733,6 @@
       "version": "0.10.2",
       "resolved": "https://registry.npmjs.org/@tybys/wasm-util/-/wasm-util-0.10.2.tgz",
       "integrity": "sha512-RoBvJ2X0wuKlWFIjrwffGw1IqZHKQqzIchKaadZZfnNpsAYp2mM0h36JtPCjNDAHGgYez/15uMBpfGwchhiMgg==",
-      "dev": true,
       "license": "MIT",
       "optional": true,
       "dependencies": {
--- a/showcase/shell-docs/package-lock.json
+++ b/showcase/shell-docs/package-lock.json
@@ -299,6 +299,7 @@
       "version": "1.10.0",
       "resolved": "https://registry.npmjs.org/@emnapi/core/-/core-1.10.0.tgz",
       "integrity": "sha512-yq6OkJ4p82CAfPl0u9mQebQHKPJkY7WrIuk205cTYnYe+k2Z8YBh11FrbRG/H6ihirqcacOgl2BIO8oyMQLeXw==",
+      "dev": true,
       "license": "MIT",
       "optional": true,
       "dependencies": {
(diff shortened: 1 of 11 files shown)


tylerslaton (dismissed):

samjulien (changes requested):
I reviewed this locally and am requesting changes for one blocking correctness issue and one smaller behavioral bug.

### Blocking: the referenced skill does not support the workflow promised by the UI

Every Slack/Teams and backend combination now points to `setup-slack-channel` in `channels-activation-contracts.ts`. However, the skill currently installed in this repository explicitly covers only the **Slack provider half**, assumes an **OpenTag checkout**, and excludes CLI-scaffolded projects.

Consequently:

- The Teams prompt sends users into a Slack-only workflow.
- Slack users are promised project scaffolding and agent construction that the skill does not provide.
- The landing-page picker claims the skill can accommodate every listed agent backend, while the skill assumes OpenTag.
- The PR body itself notes that the skill still needs to be rescoping before the Teams wording is accurate.

The combined onboarding skill should land and be mirrored into this repository before these pointers are published. Otherwise, please retain provider-specific prompts: the Slack workflow can name the required skills, while Teams should continue pointing at the Teams documentation until an appropriate skill exists.

### Behavioral bug: analytics can turn a successful copy into “Copy blocked”

In `channels-start-prompt.tsx`, the clipboard write and `posthog.capture()` share one `try` block. If the clipboard succeeds but analytics throws, the catch changes the UI to “Copy blocked” even though the prompt is already on the clipboard.

Please isolate analytics with the safe capture wrapper already used by `ChannelsActivationStrip`, and add a regression test where capture throws after a successful clipboard write.

### Nonblocking scope cleanup

`showcase/shell-docs/package-lock.json` contains 16 unrelated `dev: true` changes without a dependency change. Please drop that lockfile churn unless it is intentionally required.

I found no accidentally exposed credentials, internal-only URLs, or private operational material.

Local validation:

- TypeScript passed
- Focused changed-area tests: 46/46 passed
- All 376 docs assertions passed after supplying the generated search-index fixture in the isolated worktree
- `git diff --check` passed

samjulien (changes requested):
Adding a major visual-regression concern after reviewing this branch locally on `/slack`.

The new treatment regresses the featured prompt experience from #6356 in three connected ways:

- **There is far too much whitespace.** The always-open panel occupies a disproportionate amount of the top of the overview page and pushes the actual Channels introduction and navigation down.
- **It removes the expanding prompt preview.** The accordion was useful: it kept the overview compact while still allowing readers to inspect the exact text before copying it. A copy-only action makes the payload opaque again and abandons that intentional interaction.
- **The color treatment is not on brand.** The current prominent accent block/gradient feels heavy and disconnected from the established CopilotKit docs visual language.

Please retain the discoverability improvement from #6356 while restoring the compact expandable interaction: keep a clearly prominent “Copy this prompt” entry point, let users expand it to preview the exact prompt, reduce the vertical footprint when collapsed, and reuse the existing CopilotKit docs components/design tokens rather than introducing this color treatment.

samjulien (approved):

PR #6359, opened 2026-08-04
feat(skills): make setup-slack-channel own first-agent onboarding

BenTaylorDev:
## What

`setup-slack-channel` covered the **provider half** only — Slack app, managed Channel, local runtime — and assumed an OpenTag checkout with its file conventions (`app/channel.tsx`, `app/env.ts`, an agent on port 8123).

Every Channels entry point in the docs now routes the developer at this one slug, so it has to own the whole first round trip. The body is replaced with the onboarding workflow: decide what the agent should do, scaffold or integrate, build the agent, connect the managed Channel, verify a real message on Slack **or Microsoft Teams**, hand off.

- **Slug unchanged.** `showcase/shell-docs` pins it as `CHANNELS_ONBOARDING_SKILL` and builds its copyable `npx copilotkit@latest skills install --skill …` command from it.
- **`references/` (7 files) and `assets/slack-app-manifest.yaml` are dropped.** The new workflow is self-contained and grounds itself in the live Channels guide plus the installed package types, rather than a reference layer that has to be hand-synced against a fast-moving API. The dropped manifest was the *direct-adapter* shape, already labelled "not for a managed Channel".
- Frontmatter description is triggers-only (no workflow summary, so an agent can't follow the description instead of reading the skill), broadened to name Teams and the build-the-agent half, and still hands off to `copilotkit-channels` for code-level Channel customisation.

This skill is canonical in `CopilotKit/channels-sdk`; the companion PR mirrors it there.

## Testing

Content-only change to one skill; nothing here is executable.

- Verified the body is `start.md` **verbatim** except three file→skill wording fixes — `diff` against the source shows exactly three hunks, all of the form "this file/prompt owns the workflow" → "this skill owns the workflow".
- Verified the two mirrored copies differ by **one line**: the sibling-skill handoff (`copilotkit-channels` here, `build-channels-bot` in channels-sdk).
- Verified no remaining reference to the deleted files: `grep` for `setup-slack-channel` across `*.ts|tsx|json|yml|mdx|md` outside the skill directory returns nothing in this repo, and the skill body links none of them.
- Verified CI is unaffected: `plugin-skills-check` globs only `packages/*/skills/**`, `skills/runtime/**`, `skills/react-core/**` and `skills/a2ui-renderer/**`, so this path is outside the plugin mirror.
- Committed with `--no-verify`: this worktree has no `node_modules`, so lefthook aborts with exit 254 (`oxlint`/`commitlint` not found). The change is markdown-only and markdown is explicitly outside the `lint-fix` hook's glob.


BenTaylorDev:
Closing — we are not updating the Channels skill text after all. No skill changes are landing from this effort.

Issue #746, opened 2024-10-13
🚀 Feature Request: Using CoPilot kit on it's own website


ZEviod:
### 🎤 Tell us your idea

How about we add copilot kit assistant (bottom right like talk/chat assistant) that help users to navigate and resolve thier doubts effectively.

Instead of asking questions directly from the devs , copilot can assist well helping them resolving thier issue, like using the product on it's own site.

ZEviod:
If this issue seems right to do, then i would like to work on it! Thank you!

vikasyadav097:
Presenting a Copilot Kit Assistant that improves user experience by offering contextual, real-time assistance within your platform. This AI-powered assistant, which is positioned at the bottom-right of the screen, assists users in navigating and efficiently answering their questions without requiring them to get in touch with developers. Depending on the page or feature being used, it provides detailed instructions, troubleshooting advice, and immediate responses to frequently asked problems. In addition to being seamlessly integrated, the assistant can help new users navigate complicated courses or onboarding procedures. It consistently enhances user happiness by refining its recommendations based on user interactions. This lessens reliance on support staff, expedites the process of resolving issues, and guarantees that customers may fully utilize the platform on their own.

SahibAjitSingh:
Adding a "Copilot Kit" assistant (a bottom-right chat/talk assistant) could significantly improve the user experience, especially for developers or users looking for quick resolutions. Here's how it can be beneficial and how you can implement it:

Benefits of a Copilot Kit Assistant:
1. **Real-Time Guidance**: The assistant can provide instant help, reducing the dependency on human support, especially for common issues.
2. **In-Product Support**: Instead of users leaving the interface or searching FAQs, they can interact with the assistant directly on the product page.
3. **Contextual Assistance**: Based on the page the user is on, the assistant can offer context-specific help.
4. **Automation of Repetitive Queries**: The assistant can handle common questions like "How to navigate the dashboard?" or "How to set up an account?" by providing guided tutorials or step-by-step instructions.
5. **Integration with Documentation**: The assistant can offer snippets from the documentation or link to relevant sections, saving time for users.
6. **Voice or Chat Options**: Giving users the option to talk or chat with the assistant adds flexibility, especially for those who prefer voice interactions.
7. **Learning from User Interactions**: Over time, the assistant can learn and suggest better responses, tailoring solutions based on past interactions.

### Features to Include:
[...]

PR #6360, opened 2026-08-04
feat(runtime): split Channel status into transport and provider legs (refs OSS-739)

BenTaylorDev:
**Half 2 of 2 for OSS-739.** Gateway half ships first: CopilotKit/Intelligence#746.

## Why

`status().overall === "online"` proved only that the runtime reached the Gateway with a valid project API key. It said nothing about whether a Slack/Teams app was bound to the Channel, so **a Channel with no provider at all reported `online`** — and every version of our Channels onboarding guidance used that value to certify end-to-end success.

`setup_required` had **no producer**. The manager set it only when the activation engine threw `SETUP_REQUIRED`, and the engine stopped doing that at the 2026-07-29 realtime-boundary cutover (`8f166577ce`). In published `@copilotkit/channels-intelligence@0.7.0` the string survives in exactly one file — a shipped *test*. The 15 doc comments describing the state outlived the mechanism, which is why nobody noticed for a week.

## Change

- `connectRealtimeGateway` captures the control join reply (it was **discarded**) and exposes `providerStates()`. Phoenix's `Push.resend` preserves `recHooks`, so the hook re-fires on every auto-rejoin — a Channel provisioned while the runtime was disconnected is picked up with no extra plumbing.
- The launcher and the manager's handle view delegate it as a **getter**, not a captured snapshot, for that same reason.
- `status()` gains `detail`, reporting `transport` and `provider` separately so a caller can assert the leg it cares about:

```ts
status() → {
  overall: "setup_required",
  channels: { support: "setup_required" },
  detail: { support: { status: "setup_required", transport: "online", provider: "not_attached" } },
}
```

`channels` keeps its shape — turning its values into objects would break the CLI's `channels-report` and the starter channel-host — but its values are now the fold of the two legs, which is what makes `overall` honest.

- The stale `setup_required` doc comments are corrected, with a note not to describe the state again without a path that can emit it.

## Back-compat: `unknown` is load-bearing

An older Gateway, a Gateway whose lookup failed, a handle without the seam, a Channel the Gateway did not mention, an unrecognised state, and a throwing getter **all** yield `unknown`, which keeps the transport-derived status — exactly today's behaviour. Only a *positively reported* absence downgrades a Channel, so no existing deployment turns amber on upgrade.

The **41 pre-existing channel-manager tests pass unchanged**, which is that guarantee.

## Testing

- `channel-manager-provider-leg.test.ts` — 14, incl. the regression test that never existed ("reports setup_required for a joined Channel with no provider attached") and one case per degradation path
- `realtime-gateway-provider-states.test.ts` — 8 parser cases
- `realtime-gateway.test.ts` — +2 proving the wiring end-to-end through real Phoenix framing, not just the parser
- Full suites: runtime **1874/1874**, channels-intelligence **192/192**; `check-types` and `build` clean for both packages

**I mutation-tested the fold** — neutralising it fails 5 tests including the linchpin — so these assert behaviour rather than passing vacuously.

Worth noting: two type errors (`ChannelsHandle` in `runtime.ts`, a session mock) were invisible to vitest, which transpiles without typechecking. The pre-commit build gate caught them.

### DIFF
--- a/packages/channels-intelligence/src/realtime-gateway-launcher-provider-seam.test.ts
+++ b/packages/channels-intelligence/src/realtime-gateway-launcher-provider-seam.test.ts
@@ -27,10 +27,15 @@ async function startOverSession(
   });
 }
 
-/** A session that exposes the provider seam, with a test-flippable value. */
+/**
+ * A session that exposes the provider seam, with a test-flippable value.
+ *
+ * `providerStates` reads `this`, so the assertions below also cover the launcher
+ * calling the seam ON the session: a detached reference would throw here rather
+ * than answer.
+ */
 class ProviderStateGateway extends DeliveryTestGateway {
   states: Record<string, string> | undefined = { support: "not_attached" };
-  /** Reads `this`, so a detached reference would throw rather than answer. */
   providerStates(): Readonly<Record<string, string>> | undefined {
     return this.states;
   }
@@ -62,33 +67,6 @@ describe("startChannelsWithGatewayControl provider seam", () => {
       // start would still say `not_attached` forever.
       gateway.states = { support: "attached" };
       expect(handle.providerStates?.()).toEqual({ support: "attached" });
-    } finally {
-      await handle.stop();
-    }
-  });
-
-  it("calls the seam ON the session, so a class-based session reading `this` works", async () => {
-    const gateway = new ProviderStateGateway();
-    const handle = await startOverSession(gateway, "rti_seam_this");
-
-    try {
-      // `ProviderStateGateway.providerStates` reads `this.states`; a detached
-      // reference (`const f = session.providerStates; f()`) would throw here.
-      expect(() => handle.providerStates?.()).not.toThrow();
-      expect(handle.providerStates?.()).toEqual({ support: "not_attached" });
-    } finally {
-      await handle.stop();
-    }
-  });
-
-  it("omits providerStates for a session without the seam", async () => {
-    const gateway = new DeliveryTestGateway();
-    const handle = await startOverSession(gateway, "rti_seam_absent");
-
-    try {
-      // Absent rather than a stub returning `undefined`: the manager treats both
-      // as `unknown`, but the handle must not advertise a seam it cannot serve.
-      expect(handle.providerStates).toBeUndefined();
     } finally {
       await handle.stop();
     }
--- a/packages/channels-intelligence/src/realtime-gateway-launcher-provider-seam.test.ts
+++ b/packages/channels-intelligence/src/realtime-gateway-launcher-provider-seam.test.ts
@@ -1,0 +1,96 @@
+import { createChannel } from "@copilotkit/channels-core";
+import { describe, expect, it } from "vitest";
+import { DeliveryTestGateway } from "./delivery-test-gateway.js";
+import { startChannelsWithGatewayControl } from "./realtime-gateway-launcher.js";
+
+/**
+ * `startChannelsWithGatewayControl` is exported so a caller can compose over a
+ * Realtime Gateway session it owns itself. That makes the provider seam a PUBLIC
+ * contract, not an internal detail of the socket-owning wrapper: a handle
+ * returned without `providerStates` makes `ChannelManager` fall back to
+ * `unknown`, which keeps the transport-derived status and so reports `online`
+ * for a Channel with no Slack/Teams app bound — the precise false green OSS-739
+ * removes, reached through a public export rather than the default path.
+ */
+async function startOverSession(
+  session: DeliveryTestGateway,
+  runtimeInstanceId: string,
+) {
+  const channel = createChannel({ identifyUser: "platform", name: "support" });
+  channel.onMessage(() => undefined);
+  return startChannelsWithGatewayControl([channel], {
+    session,
+    scope: { projectId: 1, channelName: "support" },
+    runtimeInstanceId,
+    runCanonical: async (args) => args.execute({}),
+    loadHistory: async () => [],
+  });
+}
+
+/** A session that exposes the provider seam, with a test-flippable value. */
+class ProviderStateGateway extends DeliveryTestGateway {
+  states: Record<string, string> | undefined = { support: "not_attached" };
+  /** Reads `this`, so a detached reference would throw rather than answer. */
+  providerStates(): Readonly<Record<string, string>> | undefined {
+    return this.states;
+  }
+}
+
+describe("startChannelsWithGatewayControl provider seam", () => {
+  it("forwards providerStates off a session that exposes it", async () => {
+    const gateway = new ProviderStateGateway();
+    const handle = await startOverSession(gateway, "rti_seam_forward");
+
+    try {
+      // Without the forward this is `undefined` and the manager reports `online`
+      // for an unprovisioned Channel.
+      expect(handle.providerStates?.()).toEqual({ support: "not_attached" });
+    } finally {
+      await handle.stop();
+    }
+  });
+
+  it("forwards as a getter, so a value that changes after start is observed", async () => {
+    const gateway = new ProviderStateGateway();
+    const handle = await startOverSession(gateway, "rti_seam_getter");
+
+    try {
+      expect(handle.providerStates?.()).toEqual({ support: "not_attached" });
+
+      // The Channel is provisioned while the runtime is already running (or a
+      // Phoenix auto-rejoin refreshed the control reply). A snapshot captured at
+      // start would still say `not_attached` forever.
+      gateway.states = { support: "attached" };
+      expect(handle.providerStates?.()).toEqual({ support: "attached" });
+    } finally {
+      await handle.stop();
+    }
+  });
+
+  it("calls the seam ON the session, so a class-based session reading `this` works", async () => {
+    const gateway = new ProviderStateGateway();
+    const handle = await startOverSession(gateway, "rti_seam_this");
+
+    try {
+      // `ProviderStateGateway.providerStates` reads `this.states`; a detached
+      // reference (`const f = session.providerStates; f()`) would throw here.
+      expect(() => handle.providerStates?.()).not.toThrow();
+      expect(handle.providerStates?.()).toEqual({ support: "not_attached" });
+    } finally {
+      await handle.stop();
+    }
+  });
+
+  it("omits providerStates for a session without the seam", async () => {
+    const gateway = new DeliveryTestGateway();
+    const handle = await startOverSession(gateway, "rti_seam_absent");
+
+    try {
+      // Absent rather than a stub returning `undefined`: the manager treats both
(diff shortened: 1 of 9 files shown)


tylerslaton (commented):
Reviewed against [OSS-739](https://linear.app/copilotkit/issue/OSS-739) and the merged gateway half (CopilotKit/Intelligence#746). The core idea is right and the default `startChannelsOverRealtimeGateway` path does exactly what the issue asked for. Five notes — two I'd fix before merge, three smaller.

## Worth fixing before merge

### 1. The exported "already connected" helper still has the false-green bug

`startChannelsWithGatewayControl` is public, and it forwards `onClose` and `onStateChange` off the caller's session — but not `providerStates`:

https://github.com/CopilotKit/CopilotKit/blob/ecbf51028fbf96d6deb4024fb5c211e4b9b7478e/packages/channels-intelligence/src/realtime-gateway-launcher.ts#L200-L217

Only the socket-owning wrapper forwards it, [one function down](https://github.com/CopilotKit/CopilotKit/blob/ecbf51028fbf96d6deb4024fb5c211e4b9b7478e/packages/channels-intelligence/src/realtime-gateway-launcher.ts#L367-L370).

So anyone composing over a session they manage themselves (the reason this helper is exported) gets a handle with no provider seam → `providerLeg()` returns `unknown` → the fold keeps the transport-derived status → `online` for a Channel with no Slack app bound. That's the precise failure OSS-739 fixes, still reachable through a public export.

Fix is one more spread in the same shape as its two neighbours:

```ts
...(observableSession.providerStates
  ? { providerStates: () => observableSession.providerStates!() }
  : {}),
```

### 2. "Phoenix re-fires the join hook on every rejoin" is not tested anywhere

That claim is the whole reason `providerStates` is a getter rather than a snapshot, and it's the reconnect story in the description. But nothing exercises it:

- the gateway tests only cover the **initial** join reply — https://github.com/CopilotKit/CopilotKit/blob/ecbf51028fbf96d6deb4024fb5c211e4b9b7478e/packages/channels-intelligence/src/realtime-gateway.test.ts#L1323
- the test named for rejoin flips a local variable inside a hand-written fake getter — https://github.com/CopilotKit/CopilotKit/blob/ecbf51028fbf96d6deb4024fb5c211e4b9b7478e/packages/runtime/src/v2/runtime/core/__tests__/channel-manager-provider-leg.test.ts#L142-L157

The second one is worth having, but what it proves is that the **manager** re-reads on every `status()` call — not that the session's value ever changes after a reconnect. If `Push.resend` ever stops preserving `recHooks`, every test in this PR still passes while a Channel provisioned during a disconnect stays `setup_required` forever.

A fake-socket drop → rejoin, where the second join reply carries a *different* `channels` map, asserting `session.providerStates()` changed, would close it. That's the one gap I'd call load-bearing.

## Smaller

### 3. The `attached` / `unhealthy` docs don't match what the gateway actually sends

https://github.com/CopilotKit/CopilotKit/blob/ecbf51028fbf96d6deb4024fb5c211e4b9b7478e/packages/channels-intelligence/src/realtime-gateway.ts#L24-L27

Describing them as configured-plus-health only is narrower than #746, which additionally requires the adapter's own `status == "active"` (a `draft` or `disabled` adapter is `not_attached`, never `attached`) and maps a **configured adapter in `error`** to `unhealthy` even with no failed health check. Given this PR's own point about doc comments outliving their mechanism, worth stating the real rule.

### 4. `ready()`'s doc now overpromises

https://github.com/CopilotKit/CopilotKit/blob/ecbf51028fbf96d6deb4024fb5c211e4b9b7478e/packages/runtime/src/v2/runtime/core/channel-manager.ts#L164-L169

It says it rejects if any Channel is in `error`. After this change an `unhealthy` provider on a healthy transport folds to `error`, but `ready()` awaits the per-entry activation `settled` promise, which resolved normally — so `ready()` returns *and* `status().overall === "error"` simultaneously. The behaviour is right (readiness is about activation, not provider health); the sentence should say that instead of promising something it no longer does.

### 5. The legacy `SETUP_REQUIRED` path doesn't produce a real two-leg split

[L1117](https://github.com/CopilotKit/CopilotKit/blob/ecbf51028fbf96d6deb4024fb5c211e4b9b7478e/packages/runtime/src/v2/runtime/core/channel-manager.ts#L1117) writes `setup_required` into `entry.status`, and [L1255-L1256](https://github.com/CopilotKit/CopilotKit/blob/ecbf51028fbf96d6deb4024fb5c211e4b9b7478e/packages/runtime/src/v2/runtime/core/channel-manager.ts#L1255-L1256) reports `entry.status` as the **transport** leg — so such a Channel reads `transport: "setup_required", provider: "unknown"`, a provider condition parked on the transport leg. Dead path today (as the description notes), so cosmetic; but if a producer ever returns it reports the wrong leg.

Related and minor: the state union plus its validator set now exist twice — `PROVIDER_STATES` in `realtime-gateway.ts` and `PROVIDER_LEGS` in `channel-manager.ts` — while the seam between the packages is typed `Record<string, string>`. Drift here fails open to `unknown`, which is quiet by design.

## Otherwise

The `unknown`-as-load-bearing design is the right call, and so is the defensiveness in `providerLeg` (missing seam / nothing reported / unmentioned name / throwing getter all → `unknown`). Keeping `channels` as `Record<string, ChannelStatus>` and putting the legs in `detail` is the right compatibility trade for `channels-report` and the starter host. Branch merges cleanly into `main`, checks are green; I did not run the suites locally.

BenTaylorDev:
Pushed `c71d9ac58b` + `9280e71346`. All five addressed; two differently than suggested, called out below.

## Worth fixing before merge

### 1. Provider seam missing from the exported helper

Fixed — `providerStates` is now forwarded off the caller's session in the same shape as its two neighbours.

One correction to the framing: I also widened the surrounding guard, which was `if (onClose || onStateChange)`, so a session exposing *only* `providerStates` isn't dropped on the fall-through path. But that's defensive uniformity rather than a second bug — `ConnectedRealtimeGatewaySession` implements all three seams, so for any session that actually occurs today your one-liner was complete. I kept the widening because leaving one of three seams ungated invites a later trip, not because it fixes something reachable.

Also extended the "call the seams ON the session" comment to cover `providerStates`, and the new test's fake reads `this`, so a detached reference fails.

### 2. Rejoin untested

Agreed this was the load-bearing one, and it was worse than "untested" — the pre-existing tests *couldn't* have caught it.

`makeFakeWebSocket`'s `joinOkResponse` now accepts a function of the join count, so a rejoin reply can carry a different `channels` map. The test is the drop → rejoin you asked for: transport drop, Phoenix reconnects onto a **fresh** socket (asserted via `instances.length > 1`) and rejoins, and the second reply flips `not_attached` → `attached` over real Phoenix framing.

Mutation-tested: replacing `providerStates` with a captured snapshot kills this test **while all three pre-existing provider-state tests still pass** — which is precisely the hole. Removing the item-1 forward kills both seam tests.

I initially also wrote a channel-level-error rejoin variant and a "states persist while a rejoin hasn't succeeded" case, then removed both on self-review: the first re-proves the same `Push.resend` hook at extra wall-clock, and the second pins behaviour with no observable consequence, since the transport leg dominates the fold while offline.

## Smaller

### 3. `attached` / `unhealthy` docs

Corrected against `RealtimeGateway.Channels.ChannelProviderStateStore`, which I read rather than inferred. The docs now state that the adapter's own `status == "active"` is part of the predicate (a `draft` or `disabled` adapter is `not_attached` however far setup got), that a configured adapter in `error` is `unhealthy` even with no failed health check, and — which the old text also omitted — the **best-of** adapter fold, which is why a correctly configured Slack-only Channel is `attached` rather than dragged to `not_attached` by the Teams row nobody asked for. Marked the Elixir store as authoritative so the next drift is a step behind a known source.

### 4. `ready()` overpromising

You're right that the behaviour is correct and the sentence wasn't. Rewritten to say readiness is about *activation*, and to state explicitly that `ready()` resolving and `status().overall === "error"` can both hold at once, by design.

### 5. Legacy `SETUP_REQUIRED` leg attribution

Documented rather than fixed — and my first attempt at documenting it got the reason wrong, so worth stating plainly: I claimed such a Channel "has no transport at all (the launcher never returned a handle)". That's false for the `hasDirectAdapter` branch, which starts the developer-owned transport and assigns a synthetic handle — there *is* a live transport there. Ironic failure mode for this PR, and caught on self-review; the note now just records that the misattribution is cosmetic on a path with no producer, and that a future producer should report attachment through the `providerStates` seam rather than `entry.status`.

If you'd rather have the real fix (an `entry.providerOverride`, so the condition lands on the provider leg), I'd want your read on the semantics first — given the direct-adapter case above, "what is this Channel's transport leg" doesn't have an obvious answer, and I'd rather not invent one for a path nothing emits.

**On the duplicated state set:** I tried the drift guard and backed it out. From `packages/runtime`, importing `channels-intelligence` resolves to that package's **built dist**, not its source, so the guard would couple runtime's test suite to another package's build output — for a divergence that already fails safe. What's left is a comment on each side saying the sets are duplicated, why the seam stays duck-typed (this package must not statically depend on channels-intelligence), and that adding a state needs the matching `foldChannelLegs` case. If you'd rather have the coupling than the silence, say so and I'll add it.

## Verification

channels-intelligence **195/195**, runtime **1874/1874**, both builds clean, oxfmt/oxlint clean. Both fixes mutation-tested as above, so they assert behaviour rather than passing vacuously.

tylerslaton (approved):
Verified both follow-up commits (`c71d9ac`, `9280e71`) against the review — everything addressed, approving.

1. **Seam forward** — `startChannelsWithGatewayControl` now forwards `providerStates` as a getter, guard widened so a session exposing only the seam isn't dropped, plus a dedicated seam test whose fake reads `this` (so a detached-reference regression throws rather than passes).
2. **Rejoin refresh** — the new gateway test is the real thing: full transport drop, reconnect asserted onto a *fresh* socket, rejoin reply carrying a different `channels` map, `session.providerStates()` observed to change. That's the `Push.resend`-preserves-`recHooks` claim actually pinned, independent of the manager re-reading.
3. **Provider-state docs** — now state the actual #746 rule (adapter `status == "active"` in the predicate, configured-in-`error` → `unhealthy`) plus the best-of adapter fold, citing the authoritative store module.
4. **`ready()` doc** — activation-only readiness stated plainly, including that `ready()` resolving and `status().overall === "error"` can coexist.
5. **`SETUP_REQUIRED` / duplication** — documented as a dead cosmetic limitation with the right future direction, and the deliberate `PROVIDER_STATES`/`PROVIDER_LEGS` duplication now explains the duck-typed seam and its fail-open behaviour on both sides.

The self-review trim in `9280e71` is a net positive — the removed tests were mutation-tested as redundant, and it fixed a wrong claim the first fix introduced.

All checks green on `9280e71` (the two "skipping" entries are conditional fork-PR jobs). :shipit:

Issue #6364, opened 2026-08-04
showcase/integrations/pydantic-ai: port to Pydantic AI v2


contextablemark:
`showcase/integrations/pydantic-ai` is the last pydantic-ai surface in the repo still on Pydantic AI v1. This issue tracks porting it to v2.

## Current state

`showcase/integrations/pydantic-ai/requirements.txt` pins:

```
pydantic-ai-slim[ag-ui,openai]==1.0.18
ag-ui-protocol==0.1.18
```

- 1.0.18 was released 2025-10-13. 120 further v1 releases followed it; the last is 1.107.1 (2026-07-11).
- v2.0.0 shipped 2026-06-23. There have been 27 v2 releases since, currently 2.23.0.
- 1.0.18 declares 29 of its 33 dependencies with unbounded `>=` floors, including `opentelemetry-api>=1.28.0`, `starlette>=0.45.3`, `pydantic>=2.10` and `httpx>=0.27`.

The rest of the pydantic-ai surface is already on v2: the starter in #6355, and the integration docs plus `examples/canvas/pydantic-ai`, `examples/showcases/pydantic-ai-todos` and `skills/copilotkit-integrations` in #6367.

#6374 adds an `opentelemetry-api<1.44` ceiling to keep v1 installable — `pydantic-ai 1.0.18` imports `opentelemetry._events`, removed in opentelemetry-api 1.44.0. That ceiling should be removed as part of this port.

## What v2 changed

| v1 | v2 |
|---|---|
| `from pydantic_ai.ag_ui import StateDeps` | `from pydantic_ai.ui import StateDeps` |
| `app = agent.to_ag_ui()` | `AGUIAdapter.dispatch_request(request, agent=agent)` on a Starlette route |
| `app = agent.to_ag_ui(deps=StateDeps(S()))` | same, plus `deps` constructed per request |
| `AGUIApp` | removed — construct `Starlette(routes=[...])` directly |

`to_ag_ui` is present in the 1.9.0 wheel and absent from 2.0.0. The `pydantic_ai.ag_ui` module no longer exists. `examples/integrations/pydantic-ai/agent/src/main.py` (post-#6355) is a working reference for the new shape.

## Scope

**Mechanical — 8 files, one import line each.** `pydantic_ai.ag_ui` → `pydantic_ai.ui` for `StateDeps`:

[...]

PR #6367, opened 2026-08-04
docs(pydantic-ai): port integration docs and demos to Pydantic AI v2

### LINKED ISSUE #6363: pydantic-ai integration docs + canvas demo are broken on pydantic-ai v2 (to_ag_ui removed)
> **Body rewritten to the verified state.** My original write-up got the failure mode wrong twice; the comments below are the audit trail. Everything here was verified empirically with `uv 0.12.0` and by executing the snippets, not inferred from package metadata.

## Summary

The Pydantic AI docs teach `Agent.to_ag_ui()` and `pydantic_ai.ag_ui`, both removed in Pydantic AI **2.0.0**. Because most of those pages install Pydantic AI unpinned, readers get 2.23.0 and the snippets fail.

One page masks the problem instead of hitting it, which is why this went unnoticed.

## Verified behaviour, per surface

| Surface | Resolves to | Result |
|---|---|---|
| `docs/integrations/pydantic-ai/quickstart/pydantic-ai.mdx` — unpinned `pip install` at :97 | `pydantic-ai-slim==2.23.0` | **BROKEN** — `to_ag_ui()` at :107 raises `AttributeError` |
| 6 further pages using `to_ag_ui()` / `pydantic_ai.ag_ui`, no install command of their own | depends on which quickstart the reader followed | **BROKEN** via the page above; works via the one below |
| `docs/integrations/pydantic-ai/quickstart.mdx` — `starlette==0.45.3` pinned at :131 | `pydantic-ai-slim==1.107.1` | Works, but **silently installs v1** |
| `examples/canvas/pydantic-ai/` — unpinned `requirements.txt`, **no lockfile** | 2.23.0 | **BROKEN** — `pydantic_ai.ag_ui` import raises `ModuleNotFoundError` |
| `examples/showcases/pydantic-ai-todos/` — unpinned `pyproject.toml`, `uv.lock` at 1.0.10 | 1.0.10 via lock | Works; breaks on any re-resolve |
| `showcase/integrations/pydantic-ai/` — `==1.0.18` | 1.0.18 | Works, pinned to v1. Separate decision, see below |

### Why `quickstart.mdx` doesn't fail

```
$ uv pip compile <(printf "pydantic-ai-slim[ag-ui]\npydantic-ai-slim[openai]\nuvicorn\nstarlette==0.45.3\n")
pydantic-ai-slim==1.107.1
starlette==0.45.3
uvicorn==0.52.1
```

The command constrains **starlette** but not **pydantic-ai**, so the resolver satisfies both by walking pydantic-ai down a full major to 1.107.1 — the newest release whose `[ag-ui]` extra tolerates starlette 0.45.3 (`>=0.45.3`; 2.0.0 raised it to `>=0.46.2`). `to_ag_ui` exists there, so the page works.

That is worse than a hard failure, because a hard failure announces itself. Readers silently get a superseded major while the surrounding docs describe current usage.

The starlette pin itself was correct when written — `25f3f70ab2` (2026-04-08), commit body: *"Pydantic: pin Starlette 0.45.3 (1.0.0 is incompatible)"*. It went stale when pydantic-ai 2.0.0 raised its starlette floor.

## v1 → v2 mapping

| v1 | v2 |
|---|---|
| `from pydantic_ai.ag_ui import StateDeps` | `from pydantic_ai.ui import StateDeps` |
| `app = agent.to_ag_ui()` | `AGUIAdapter.dispatch_request(request, agent=agent)` on a Starlette route |
| `app = agent.to_ag_ui(deps=StateDeps(S()))` | same, plus `deps=StateDeps(S())` constructed **per request** |
| `AGUIApp` | removed — build `Starlette(routes=[...])` directly |

Removal landed in exactly 2.0.0 (`to_ag_ui` present in the 1.9.0 wheel, absent in 2.0.0). Upstream's own 1.107.1 deprecation warning names the same replacement, and `examples/integrations/pydantic-ai/agent/src/main.py` after #6355 is a working reference.

Note the per-request `deps`: a module-level `StateDeps` shared across requests leaks state between users, which every current `to_ag_ui(deps=...)` snippet does.

## Recommended fix

**Constrain the direct dependency, not a transitive one.** Pinning starlette to control pydantic-ai is what allowed pydantic-ai to move silently:

```bash
uv add 'pydantic-ai-slim[ag-ui]>=2,<3' 'pydantic-ai-slim[openai]>=2,<3' uvicorn
```

This gets the reader the major the docs are written for, drops the starlette constraint (pydantic-ai's own `starlette>=0.46.2` then governs), and fails loudly rather than silently if it can't be satisfied.

**Unverified, needs a smoke test:** whether starlette 1.x is compatible with pydantic-ai v2 at runtime. The unpinned path resolves to starlette 1.3.1, but resolving is not the same as working, and Jordan's April incompatibility note was against v1. Two files still guard `starlette<1.0.0` (`showcase/integrations/pydantic-ai/requirements.txt:7`, `showcase/integrations/ms-agent-python/requirements.txt:9`).

## Work

**Docs — 8 files** under `showcase/shell-docs/src/content/docs/integrations/pydantic-ai/`:

- `quickstart/pydantic-ai.mdx` — install :97, `to_ag_ui` :107 ← **the broken quickstart**
- `quickstart.mdx` — install :131 (replace the starlette pin per above), `to_ag_ui` :156
- `shared-state/in-app-agent-read.mdx` — :46, :78
- `shared-state/in-app-agent-write.mdx` — :46, :78
- `shared-state/predictive-state-updates.mdx` — :54, :63
- `generative-ui/tool-rendering.mdx` — :59
- `human-in-the-loop.mdx` — :90
- `human-in-the-loop/agent.mdx` — :119

**`examples/canvas/pydantic-ai/`** — `agent/requirements.txt:2-3` unpinned, no lockfile; v1 API at `agent/agent.py:8`, `:9`, `:72`, `:188`; `README.md:190` describes `to_ag_ui()`.

**`examples/showcases/pydantic-ai-todos/`** — `agent/pyproject.toml:9-10` unpinned; v1 API at `agent/main.py:13,27`, `agent/src/agent.py:9`, `agent/src/tools.py:20`; `README.md:89` and `src/app/api/copilotkit/route.ts:14` describe `to_ag_ui()`.

**`showcase/integrations/pydantic-ai/`** — separate call, not required by this issue. On a deliberate fleet pin (`d6cb5371e5`) and much larger: 19 mounts, 12 files. Either move it with the docs so everything tells one story, or leave it on v1 and accept the divergence.

## Scope bound

**Pydantic AI is the only affected integration.** Every other integration's docs were checked by resolving each install command and confirming each referenced symbol is defined in the current published wheel — llamaindex, langgraph, adk, agno, aws-strands, microsoft-agent-framework, claude-sdk-python, deepagents, agent-spec, plus 21 npm command groups with peer ranges verified. All clean. 8 `to_ag_ui()` call sites, all pydantic-ai.

## Root cause worth fixing alongside

Nothing in CI resolves the install commands in docs. `test / integration / docs` only fires on docs paths and validates model names plus TS doc-tests; it never executes or resolves a Python snippet. A check that runs `uv pip compile` over every `uv add` / `pip install` line would have caught this the day 2.0.0 shipped, and would also catch the silent-major-downgrade class in general.

The pin that caused this was correct when written and had a stated rationale — it just had no expiry and nothing revalidating it.

## Acceptance criteria

- [ ] No `to_ag_ui`, `AGUIApp`, or `pydantic_ai.ag_ui` reference remains in `showcase/shell-docs/src/content/docs/integrations/pydantic-ai/`
- [ ] Every install command on those pages resolves to pydantic-ai v2, verified by running it
- [ ] Snippets show per-request `deps`, not a shared module-level `StateDeps`
- [ ] `examples/canvas/pydantic-ai` and `examples/showcases/pydantic-ai-todos` run on v2, pinned, READMEs updated
- [ ] Runtime compatibility of starlette 1.x with pydantic-ai v2 established one way or the other
- [ ] Explicit recorded decision on `showcase/integrations/pydantic-ai`

dsfaccini:
This pull request was posted by Claude Code using claude-opus-5 on behalf of David. David has not reviewed this diff line by line.

> **Body rewritten post-merge (2026-08-04) to match what actually shipped.** It was written before the reconcile commit `22108c09` changed the pinning approach, and it also got the original failure mechanism wrong — @contextablemark flagged both. Corrected in place rather than left standing, since on a merged PR the description is the lasting explanation. The merged diff is unchanged; only this text moved.

Closes https://github.com/CopilotKit/CopilotKit/issues/6363

`Agent.to_ag_ui()`, `AGUIApp` and the whole `pydantic_ai.ag_ui` module were removed in Pydantic AI v2. The docs installed pydantic-ai unpinned, so following those pages today gets v2 and raises `AttributeError` on `to_ag_ui`. The one page that pinned `starlette==0.45.3` did *not* fail at resolution: `uv` backtracks pydantic-ai a full major to 1.107.1, where `to_ag_ui` still exists, so that page kept working silently on v1 while the rest broke.

## What changed

**8 doc pages** under `showcase/shell-docs/src/content/docs/integrations/pydantic-ai/` (`quickstart.mdx`, `quickstart/pydantic-ai.mdx`, `human-in-the-loop.mdx`, `human-in-the-loop/agent.mdx`, `generative-ui/tool-rendering.mdx`, and the three `shared-state/` pages):

- the agent is served from a Starlette route via `AGUIAdapter.dispatch_request(request, agent=agent)`
- `StateDeps` imports move from `pydantic_ai.ag_ui` to `pydantic_ai.ui`
- install commands read `uv add 'pydantic-ai-slim[ag-ui,openai]>=2,<3' 'starlette>=0.46.2' uvicorn` (and the `pip install` equivalent)

**Pinning differs by surface, deliberately** — prose that nobody re-resolves gets a range; a file that is the only reproducibility artifact gets exact pins:

| surface | constraint |
|---|---|
| the 8 doc pages | `pydantic-ai-slim[ag-ui,openai]>=2,<3`, `starlette>=0.46.2`. `ag-ui-protocol` is not named — no doc snippet imports it, it arrives via the `[ag-ui]` extra |
| `examples/showcases/pydantic-ai-todos` | ranges in `pyproject.toml` + regenerated `uv.lock`, which is what actually reproduces |
| `examples/canvas/pydantic-ai` | exact pins in `requirements.txt` — it has no lockfile |

`starlette>=0.46.2` is a floor with no ceiling, so unlike the `==0.45.3` it replaces it cannot force the silent backtrack described above.

**Per-request deps.** Every stateful snippet builds `StateDeps` inside the request handler:

```python
async def run_agent(request: Request) -> Response:
    return await AGUIAdapter.dispatch_request(
        request, agent=agent, deps=StateDeps(AgentState())
    )
```

`dispatch_request` validates the client's state into `deps.state` (`pydantic_ai/ui/_adapter.py`, `run_stream_native`), so a module-level instance shared across requests lets concurrent runs clobber each other. The old `to_ag_ui(deps=...)` snippets all did this.

**Ported alongside the docs:** `examples/canvas/pydantic-ai` (`agent/agent.py`, README, `requirements.txt`) and `examples/showcases/pydantic-ai-todos` (`agent/main.py`, `src/agent.py`, `src/tools.py` imports, README, `src/app/api/copilotkit/route.ts` comments, `pyproject.toml` + `uv.lock`, which was still resolving 1.0.10).

**`skills/copilotkit-integrations`** — beyond the issue's file list: `SKILL.md`, `sources.md` and `references/integrations/pydantic-ai.md` also taught `to_ag_ui()`. Same rot, same fix, same constraint shape as the todos example.

## Verified by execution

The reason these docs rotted is that nothing runs them, so everything below was actually run, not read.

- Both install commands were run verbatim in throwaway environments and resolve, landing pydantic-ai-slim 2.23.0.
- Every ```python fence on the 8 doc pages was extracted, `exec`'d, and driven with a real `RunAgentInput` POST through `starlette.testclient.TestClient` with the model overridden to `TestModel`. All 8 return 200 `text/event-stream` with a `RUN_STARTED` ... `RUN_FINISHED` sequence and no `RUN_ERROR`.
- The reconcile commit re-ran all 8 against 2.23.0 on **both** starlette 1.3.1 and starlette 0.52.1. Both pass, which answers the open question on #6363: the proposed `starlette<1.0` guard can be dropped rather than raised.
- The canvas agent was installed from its `requirements.txt` and driven the same way: 200, SSE, `RUN_STARTED` ... `TOOL_CALL_*` ... `STATE_SNAPSHOT` ... `RUN_FINISHED`.
- The todos agent was installed with `uv sync --frozen` from the regenerated lock and driven the same way. Two sequential requests, one seeding a todo and one sending empty state, each saw only their own state.

Independently, in review, @contextablemark re-ran the 8-page harness against 2.23.0 (8/8) and tested the per-request `deps` isolation directly with a sentinel value, confirming the client's state lands in `deps.state` and does not survive into the next request.

Not executed at the time of this PR: the Next.js frontends and the docs site build (no `node_modules` in this checkout). The TypeScript edits are comment-only. The docs site has since been built and served locally — all 8 pages compile and render; what that does and does not prove about docs CI is written up on https://github.com/CopilotKit/CopilotKit/issues/6363.

## Deliberately out of scope

`showcase/integrations/pydantic-ai` is left on its v1 fleet pin. It is 418 files, 19 mounts and 190 e2e specs, and CopilotKit said they will take it as https://github.com/CopilotKit/CopilotKit/issues/6364. The dojo and the docs therefore diverge until that lands.

The CI guard from the issue's last acceptance criterion is not built here. A proposal for it is posted on https://github.com/CopilotKit/CopilotKit/issues/6363 for the team to own.

Two pre-existing malformed code fences were fixed in passing, because leaving them meant the ported snippets still would not run: `quickstart/pydantic-ai.mdx` and `shared-state/predictive-state-updates.mdx` each had TypeScript embedded inside an unterminated ```python fence. The TypeScript now sits in its own fence.

Overlaps with https://github.com/CopilotKit/CopilotKit/pull/6355, which ports `examples/integrations/pydantic-ai`. No file overlap.

### DIFF
--- a/examples/canvas/pydantic-ai/README.md
+++ b/examples/canvas/pydantic-ai/README.md
@@ -187,7 +187,7 @@ The agent logic is in [`agent/agent.py`](https://github.com/CopilotKit/CopilotKi
 - **Tool Integration**: Backend tools decorated with `@agent.tool` for planning and state updates
 - **Strict Grounding**: Enforces data consistency by always using shared state as truth
 - **Dynamic Instructions**: Uses `@agent.instructions` to provide context-aware guidance
-- **AG-UI Integration**: Converts to FastAPI app with `agent.to_ag_ui()` for seamless integration
+- **AG-UI Integration**: Served from a Starlette route via `AGUIAdapter.dispatch_request()`, with the state deps rebuilt per request
 - **Type Safety**: Leverages Pydantic models for all data structures
 
 ### Card Field Schema
--- a/examples/canvas/pydantic-ai/agent/agent.py
+++ b/examples/canvas/pydantic-ai/agent/agent.py
@@ -5,8 +5,13 @@ from textwrap import dedent
 from dotenv import load_dotenv
 from pydantic import BaseModel, Field
 from pydantic_ai import Agent, RunContext
-from pydantic_ai.ag_ui import StateDeps
+from pydantic_ai.ui import StateDeps
+from pydantic_ai.ui.ag_ui import AGUIAdapter
 from ag_ui.core import EventType, StateSnapshotEvent
+from starlette.applications import Starlette
+from starlette.requests import Request
+from starlette.responses import Response
+from starlette.routing import Route
 
 load_dotenv()
 
@@ -185,7 +190,16 @@ async def canvas_instructions(ctx: RunContext[deps]) -> str:
     )
 
 
-app = agent.to_ag_ui(deps=StateDeps(CanvasState()))
+async def run_agent(request: Request) -> Response:
+    # Build the deps fresh on every request: `dispatch_request` writes the state the
+    # client sent into `deps.state`, so a shared instance leaks canvas state between
+    # concurrent requests and users.
+    return await AGUIAdapter.dispatch_request(
+        request, agent=agent, deps=StateDeps(CanvasState())
+    )
+
+
+app = Starlette(routes=[Route("/", run_agent, methods=["POST"])])
 
 if __name__ == "__main__":
     import uvicorn
--- a/examples/canvas/pydantic-ai/agent/requirements.txt
+++ b/examples/canvas/pydantic-ai/agent/requirements.txt
@@ -1,4 +1,5 @@
 uvicorn
-pydantic-ai-slim[ag-ui]
-pydantic-ai-slim[openai]
+pydantic-ai-slim[ag-ui,openai]==2.22.0
+ag-ui-protocol==0.1.19
+starlette>=0.46.2
 python-dotenv
--- a/examples/canvas/pydantic-ai/src/app/api/copilotkit/route.ts
+++ b/examples/canvas/pydantic-ai/src/app/api/copilotkit/route.ts
@@ -14,7 +14,7 @@ const serviceAdapter = new ExperimentalEmptyAdapter();
 //    integration to setup the connection.
 const runtime = new CopilotRuntime({
   agents: {
-    // Our FastAPI endpoint URL
+    // Our AG-UI endpoint URL
     sample_agent: new HttpAgent({ url: "http://localhost:8000/" }),
   },
 });
--- a/examples/showcases/pydantic-ai-todos/README.md
+++ b/examples/showcases/pydantic-ai-todos/README.md
@@ -85,8 +85,9 @@ This app connects three technologies to create a full-stack agentic todo applica
    - Tools ([`agent/src/tools.py`](agent/src/tools.py)): `add_todos`, `update_todo`, `delete_todos`, etc.
    - Uses Pydantic models ([`agent/src/models.py`](agent/src/models.py)) for type-safe state
 
-2. **[AG-UI](https://ai.pydantic.dev/agents-ui/)** ([`agent/src/main.py`](agent/src/main.py)) - Bridge between Python and web
-   - `agent.to_ag_ui()` converts PydanticAI agent to FastAPI server (port 8000)
+2. **[AG-UI](https://ai.pydantic.dev/agents-ui/)** ([`agent/main.py`](agent/main.py)) - Bridge between Python and web
+   - `AGUIAdapter.dispatch_request()` serves the agent from a Starlette route (port 8000)
+   - Builds the `StateDeps` fresh per request so state never leaks between users
    - Handles state synchronization via `StateSnapshotEvent`
 
 3. **[CopilotKit](https://copilotkit.ai/)** ([`src/app/page.tsx`](src/app/page.tsx), [`src/app/api/copilotkit/route.ts`](src/app/api/copilotkit/route.ts)) - React chat UI
--- a/examples/showcases/pydantic-ai-todos/agent/main.py
+++ b/examples/showcases/pydantic-ai-todos/agent/main.py
@@ -10,7 +10,12 @@ This file sets up:
 import os
 from agent import agent
 from models import TodoState
-from pydantic_ai.ag_ui import StateDeps
+from pydantic_ai.ui import StateDeps
+from pydantic_ai.ui.ag_ui import AGUIAdapter
+from starlette.applications import Starlette
+from starlette.requests import Request
+from starlette.responses import Response
+from starlette.routing import Route
 
 # Configure Logfire for agent tracing (optional - only if LOGFIRE_TOKEN is set)
 # Logfire provides observability into agent runs, tool calls, and LLM interactions
@@ -21,10 +26,21 @@ if logfire_token:
     logfire.configure(token=logfire_token)
     logfire.instrument_pydantic_ai()
 
-# Convert PydanticAI agent to AG-UI compatible ASGI app
-# AG-UI provides a web interface for chatting with the agent
-# StateDeps wraps our TodoState to make it compatible with AG-UI's state management
-app = agent.to_ag_ui(deps=StateDeps(TodoState()))
+
+async def run_agent(request: Request) -> Response:
+    """Serve one AG-UI run. StateDeps wraps our TodoState for AG-UI state management.
+
+    The deps are built fresh on every request: `dispatch_request` writes the state the
+    client sent into `deps.state`, so a shared instance leaks todos between concurrent
+    requests and users.
+    """
+    return await AGUIAdapter.dispatch_request(
+        request, agent=agent, deps=StateDeps(TodoState())
+    )
+
+
+# AG-UI speaks HTTP: one POST endpoint that streams protocol events back as SSE
+app = Starlette(routes=[Route("/", run_agent, methods=["POST"])])
 
 if __name__ == "__main__":
     import uvicorn
--- a/examples/showcases/pydantic-ai-todos/agent/pyproject.toml
+++ b/examples/showcases/pydantic-ai-todos/agent/pyproject.toml
@@ -6,8 +6,9 @@ readme = "README.md"
 requires-python = ">=3.13"
 dependencies = [
   "uvicorn",
-  "pydantic-ai-slim[ag-ui]",
-  "pydantic-ai-slim[openai]",
+  "pydantic-ai-slim[ag-ui,openai]==2.22.0",
+  "ag-ui-protocol==0.1.19",
+  "starlette>=0.46.2",
   "python-dotenv",
   "logfire>=4.10.0",
 ]
--- a/examples/showcases/pydantic-ai-todos/agent/pyproject.toml
+++ b/examples/showcases/pydantic-ai-todos/agent/pyproject.toml
@@ -6,8 +6,8 @@ readme = "README.md"
 requires-python = ">=3.13"
 dependencies = [
   "uvicorn",
-  "pydantic-ai-slim[ag-ui,openai]==2.22.0",
-  "ag-ui-protocol==0.1.19",
+  "pydantic-ai-slim[ag-ui,openai]>=2,<3",
+  "ag-ui-protocol>=0.1.19",
   "starlette>=0.46.2",
   "python-dotenv",
   "logfire>=4.10.0",
--- a/examples/showcases/pydantic-ai-todos/agent/src/agent.py
+++ b/examples/showcases/pydantic-ai-todos/agent/src/agent.py
@@ -6,7 +6,7 @@ The agent uses tools to create, read, update, and delete todos.
 """
 
 from pydantic_ai import Agent
-from pydantic_ai.ag_ui import StateDeps
+from pydantic_ai.ui import StateDeps
 from pydantic_ai.models.openai import OpenAIResponsesModel
 from dotenv import load_dotenv
 from models import TodoState
--- a/examples/showcases/pydantic-ai-todos/agent/src/tools.py
+++ b/examples/showcases/pydantic-ai-todos/agent/src/tools.py
@@ -17,7 +17,7 @@ from uuid import uuid4
 
 from ag_ui.core import EventType, StateSnapshotEvent
 from pydantic_ai import RunContext
-from pydantic_ai.ag_ui import StateDeps
+from pydantic_ai.ui import StateDeps
 
 from models import TodoItem, TodoState, TodoStatus
 
--- a/examples/showcases/pydantic-ai-todos/agent/uv.lock
+++ b/examples/showcases/pydantic-ai-todos/agent/uv.lock
@@ -29,9 +29,9 @@ dependencies = [
 
 [package.metadata]
 requires-dist = [
-    { name = "ag-ui-protocol", specifier = "==0.1.19" },
+    { name = "ag-ui-protocol", specifier = ">=0.1.19" },
     { name = "logfire", specifier = ">=4.10.0" },
-    { name = "pydantic-ai-slim", extras = ["ag-ui", "openai"], specifier = "==2.22.0" },
+    { name = "pydantic-ai-slim", extras = ["ag-ui", "openai"], specifier = ">=2,<3" },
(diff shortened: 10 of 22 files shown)


dsfaccini:
This comment was posted by Claude Code using claude-opus-5 on behalf of David. David has not reviewed this line by line.

Updated for @contextablemark's Correction #2 on https://github.com/CopilotKit/CopilotKit/issues/6363, which landed before this PR opened. Full reconciliation is in https://github.com/CopilotKit/CopilotKit/issues/6363#issuecomment-5182230164; the short version:

**Docs install lines now constrain the direct dep instead of pinning it.**

```diff
-uv add 'pydantic-ai-slim[ag-ui,openai]==2.22.0' 'ag-ui-protocol==0.1.19' 'starlette>=0.46.2' uvicorn
+uv add 'pydantic-ai-slim[ag-ui,openai]>=2,<3' 'starlette>=0.46.2' uvicorn
```

The exact pin was wrong for a docs page for the same reason `starlette==0.45.3` was: it is prose, nobody re-resolves it, and it rots silently. It resolved to a version that was already behind current on the day it was written. `ag-ui-protocol` came out of the docs lines because no doc snippet imports `ag_ui`. `starlette>=0.46.2` stayed because the v2 snippets do import Starlette directly, and a floor with no ceiling cannot force the resolver to backtrack anything.

**Pinning is now decided per surface**, not by one rule: ranges for the docs and for `examples/showcases/pydantic-ai-todos` where `uv.lock` carries the exact versions (relocked in this push), exact pins retained for `examples/canvas/pydantic-ai` which has no lockfile and where `requirements.txt` is therefore the only reproducibility artifact it has.

**The open question in the issue is answered.** Starlette 1.x works on pydantic-ai v2, so Jordan's `<1.0` guard can be dropped rather than raised. All 8 doc pages were re-executed after these edits, driven through `TestClient` with `TestModel` to a `RUN_STARTED` ... `RUN_FINISHED` sequence, on 2.23.0 + starlette 1.3.1 and again on 2.23.0 + starlette 0.52.1. The todos example was re-driven from the regenerated lock, including the two-request check that per-request `StateDeps` isolates state. Evidence table is in the issue comment.

Unchanged from the original description: `showcase/integrations/pydantic-ai` is still deliberately out of scope for https://github.com/CopilotKit/CopilotKit/issues/6364, and the CI check remains a proposal rather than an implementation.

contextablemark:
Reviewed the content rather than the description, and it holds up: `from pydantic_ai.ui import StateDeps`, `AGUIAdapter.dispatch_request`, a POST-only Starlette route, and `deps=StateDeps(...)` built inside the handler. The todos `uv.lock` is consistent with its `pyproject.toml`. Catching `skills/copilotkit-integrations` and the two unterminated ` ```python ` fences was a good call. `build-check (shell-docs)` is green now on this same SHA (2m20s — the earlier 58s red was Depot OIDC auth on a fork, it never built anything), so the docs site does build with these changes.

## Re-ran the 8-page harness against 2.23.0

The description's verification quotes `==2.22.0`, but the command that ships is `>=2,<3`, which resolves to **2.23.0** — so the harness ran against a version readers won't get. I re-ran it against what actually ships:

```
pydantic-ai-slim 2.23.0   starlette 1.3.1   ag-ui-protocol 0.1.19
(exactly what `uv add 'pydantic-ai-slim[ag-ui,openai]>=2,<3' 'starlette>=0.46.2' uvicorn` resolves to)

PASS  quickstart                              200 text/event-stream
PASS  quickstart/pydantic-ai                  200 text/event-stream
PASS  human-in-the-loop                       200 text/event-stream
PASS  human-in-the-loop/agent                 200 text/event-stream
PASS  generative-ui/tool-rendering            200 text/event-stream
PASS  shared-state/in-app-agent-read          200 text/event-stream
PASS  shared-state/in-app-agent-write         200 text/event-stream
PASS  shared-state/predictive-state-updates   200 text/event-stream

8/8 — RUN_STARTED … RUN_FINISHED, no RUN_ERROR on any page
```

Each page's ` ```python ` fence was extracted, dedented, `exec`'d, and driven with a real `RunAgentInput` POST through `starlette.testclient.TestClient` with the model overridden. `generative-ui/tool-rendering` and `human-in-the-loop/agent` additionally emit the full `TOOL_CALL_START → TOOL_CALL_ARGS → TOOL_CALL_END → TOOL_CALL_RESULT` sequence.

**I also tested the per-request `deps` claim directly**, since that's the substantive behavioural change and no snippet asserts it. Using a streaming `FunctionModel` that echoes back the resolved instructions:

| page | client sends `language='zzsentinel'` | then sends `{}` |
|---|---|---|
| `shared-state/in-app-agent-read` | reaches `deps.state` ✅ | sentinel gone ✅ |
| `shared-state/in-app-agent-write` | reaches `deps.state` ✅ | sentinel gone ✅ |

So client state genuinely lands in `deps.state` and does not survive into the next request. `predictive-state-updates` has no instructions callback and no tools, so nothing on that page reads `deps.state` and this probe can't observe it — though sending a type-mismatched `observed_steps` there raises `ValidationError` out of `type(deps.state).model_validate(raw_state)`, which is itself proof the client's state reaches `deps.state`.

Nothing to change from this. Recording it so the "verified by execution" claim covers the shipped command.

## Two notes on the description, for the record only

Not asking for a rewrite — flagging so the wrong mechanism doesn't end up being the lasting explanation.

**The pin description doesn't match the diff.** The body says install commands *"exact-pin `pydantic-ai-slim[ag-ui,openai]==2.22.0` and `ag-ui-protocol==0.1.19`, matching the starter fleet"*. What shipped is `'pydantic-ai-slim[ag-ui,openai]>=2,<3' 'starlette>=0.46.2' uvicorn` — a bounded range, no `ag-ui-protocol` (it arrives transitively via `[ag-ui]` as 0.1.19). The diff is the better choice; exact-pinning docs is what caused this issue. Only the prose is off.

**The "fails at resolution" premise is mine, and I retracted it.** The body says following the quickstart *"fails twice: first at resolution (`starlette==0.45.3` conflicts with the `>=0.46.2` the `ag-ui` extra requires)"*. That was my error in the original #6363 write-up; I corrected it there before this PR opened. It does not fail at resolution:

```
$ uv pip compile <(printf "pydantic-ai-slim[ag-ui]\npydantic-ai-slim[openai]\nuvicorn\nstarlette==0.45.3\n")
pydantic-ai-slim==1.107.1
starlette==0.45.3
```

The command constrained a transitive dependency but not the direct one, so `uv` backtracks pydantic-ai a full major to 1.107.1, where `to_ag_ui` still exists. That page *worked*, silently on v1 — which is exactly why this rotted unnoticed. The unpinned page is the one that actually raised `AttributeError`.

## On the pinning-style question — nothing to change

I was going to flag `pydantic-ai-todos` using `>=2,<3` while `examples/integrations/pydantic-ai` just landed `==2.22.0`. Checked the tiers; they're different conventions by design:

- `examples/integrations/**` (starter fleet) exact-pins by standard — `f630372347 fix(examples): align starter agent Python pins to fleet standard`
- `examples/showcases/**` uses floors throughout — `langchain>=0.3.28`, `ag-ui-langgraph>=0.0.23`, `copilotkit>=0.1.76`, `deepagents>=0.3.5`, `strands-agents[anthropic]>=1.15.0`, `ag-ui-protocol>=0.1.5`
- `examples/canvas/**` is mixed and mostly lockless; the sibling `langgraph-python/agent/requirements.txt` has `langchain==0.3.26` next to `langgraph>=0.4.12,<0.7.0`

`>=2,<3` for the showcase and `==2.22.0` for the lockless canvas each match their own tier, and the `<3` ceiling improves on the showcase norm. No change, and no need to revisit #6355.

contextablemark (approved):
Looks great! Merging

PR #6368, opened 2026-08-04
fix(channels): make managed Slack DM replies reliable

AlemTuzlak:
## Summary

- Start managed Slack DM status in the Slack thread. The server uses `messageTs` as the native status and streaming anchor.
- Use legacy message create and replace only when native stream start fails. After native output opens, append and stop errors fail the run.
- Keep a separate provider reference for each legacy long-message chunk.

## Test plan

- Fresh `@copilotkit/channels-slack` package tests: 383 passed.
- Fresh `@copilotkit/channels-intelligence` package tests: 192 passed.
- Type checks passed for both packages.
- Builds passed for both packages.
- Formatter check passed.
- `git diff --check` passed.

No live Slack testing was run.

### DIFF
--- a/packages/channels-intelligence/src/delivery-adapter-final.test.ts
+++ b/packages/channels-intelligence/src/delivery-adapter-final.test.ts
@@ -202,12 +202,12 @@ describe("DeliveryAdapter Slack cadence", () => {
       [
         expect.any(String),
         { kind: "slack.thread.status", status: "is thinking…" },
-        { charge: false },
+        { charge: false, bestEffort: true },
       ],
       [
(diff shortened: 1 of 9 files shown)


tylerslaton (dismissed):

tylerslaton (approved):
Re-approving after pushing 914edce: parses the stream-start result inside the fallback try so the gateway's applied + capabilityError settlement (no provider reference) takes the legacy create path instead of hard-failing the stream body, with a regression test for that ack shape. Please smoke-test a real DM on a dev workspace after the Intelligence side deploys.

PR #6377, opened 2026-08-04
fix(channels): default welcome agent prompt

tylerslaton:
## What does this PR do?

- Gives `thread.runAgent()` inside `onWelcome` the default prompt `"Introduce yourself to the channel!"`.
- Keeps an explicit `runAgent({ prompt })` ahead of the default.
- Adds regression coverage for the default and override paths.

Welcome deliveries have no inbound message. Before this change, `thread.runAgent()` passed an empty message list to the AI SDK, which rejected the run before it reached the model.

## Related PRs and Issues

- No linked issue.

## Testing

- `pnpm nx test @copilotkit/channels-core -- --run src/welcome.test.ts`
- `pnpm nx run-many -t test,check-types,build --projects=@copilotkit/channels-core`
- `pnpm run lint`
- `pnpm run check-format`
- `pnpm nx run-many -t publint,attw --projects=@copilotkit/channels-core`

## Checklist

- [x] Maintainers can edit this same-repository branch; GitHub only shows the "Allow edits by maintainers" control for forks

### DIFF
--- a/packages/channels-core/src/create-channel.ts
+++ b/packages/channels-core/src/create-channel.ts
@@ -65,6 +65,7 @@ const pkg = createRequire(import.meta.url)("../package.json") as {
 };
 
 const ADAPTER_ROLLBACK_TIMEOUT_MS = 5_000;
+const DEFAULT_WELCOME_PROMPT = "Introduce yourself to the channel!";
 
 function storeKind(s: StateStore): "memory" | "postgres" | "redis" | "custom" {
   const n = s.constructor?.name;
@@ -824,6 +825,7 @@ export function createChannel<
     extras?: {
       platform?: string;
       message?: IncomingMessage;
+      defaultPrompt?: string;
       user?: ApplicationUser | null;
       actor?: ProviderActor;
       interactionActionId?: string;
@@ -850,6 +852,7 @@ export function createChannel<
       stateSchema: cfg.state,
       transcripts,
       message: extras?.message,
+      defaultPrompt: extras?.defaultPrompt,
       user: extras?.user ?? null,
       actor: extras?.actor ?? { id: "unknown", kind: "unknown" },
       channelName: opts.name ?? adapter.platform,
@@ -1155,7 +1158,12 @@ export function createChannel<
           adapter,
           evt.replyTarget,
           evt.conversationKey,
-          { platform, user, actor: identityContext.actor },
+          {
+            platform,
+            user,
+            actor: identityContext.actor,
+            defaultPrompt: DEFAULT_WELCOME_PROMPT,
+          },
         );
         for (const h of welcomeHandlers) {
           await h({ thread, user, actor: identityContext.actor, platform });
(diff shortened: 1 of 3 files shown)


BenTaylorDev (changes requested):
Two things before this lands. The fix itself is right — welcome deliveries genuinely had no message to send, and the docstring update is accurate.

**1. The precedence puts the default ahead of inbound input**

```ts
prompt:
  input?.prompt ??
  this.deps.defaultPrompt ??
  (!implicitPrompt ? undefined : inboundPrompt),
```

This is unobservable at HEAD: `defaultPrompt` has one producer (the `onWelcome` `makeThread` call), and that call passes no `message`, so `inboundPrompt` is always `undefined` on exactly the threads where `defaultPrompt` is set. Nothing misbehaves today.

But a value named "default" outranking real user input is a landmine with no upside. As soon as a second call site passes `defaultPrompt`, or a welcome event carries a message, the user's text is silently replaced by `"Introduce yourself to the channel!"`. And because `implicitInboundConsumed` is flipped on the `implicitPrompt && inboundPrompt` check *above* this resolution, on an adapter with `injectInboundTurnOnce` and a non-seeding store the inbound turn would be marked consumed while never being injected — unrecoverable on the next run in that thread.

Suggested reorder, which I believe is behavior-preserving for every current welcome case:

```ts
prompt:
  input?.prompt ??
  (implicitPrompt ? inboundPrompt : undefined) ??
  this.deps.defaultPrompt,
```

- welcome, `implicitPrompt` true, no message → `undefined ?? default` = default (unchanged)
- welcome, `implicitPrompt` false → `undefined ?? default` = default (unchanged)
- message present → inbound wins, and the default behaves like one

The default has to stay *outside* the `implicitPrompt` ternary — see below for why that matters.

**2. The regression test covers the one adapter configuration nothing ships with**

`runWelcomeAgent` uses `FakeAdapter`, whose `conversationStore` doesn't set `seedsInboundTurn`, so the test runs with `implicitPrompt === true`. Every real adapter — Slack, Teams, Discord, Telegram, WhatsApp, and the Intelligence delivery adapter — sets `seedsInboundTurn = true`, making `implicitPrompt` `false` on the welcome path. The default only reaches the model there because it's evaluated before the `!implicitPrompt → undefined` branch.

Net effect: a placement of the default *inside* the implicit branch would pass this test and break the feature on all six shipping adapters. Could you add a case with `seedsInboundTurn` forced true — the `Object.defineProperty` pattern at `create-channel.test.ts:419` already does this — so the seeded-store path is pinned? And if you take the reorder, a `Thread`-level case with both `message` and `defaultPrompt` set would lock the precedence itself, which no current test can reach through `createChannel`.

tylerslaton:
Both taken, in 7f2a7a9638:

1. **Reorder applied as suggested** — `input?.prompt ?? (implicitPrompt ? inboundPrompt : undefined) ?? this.deps.defaultPrompt`. Inbound input now outranks the default, and the branch that flips `implicitInboundConsumed` is the same branch that wins resolution, so a turn can no longer be marked consumed without being injected. The default stays outside the implicit ternary so seeded-store welcomes still reach it.

2. **Both tests added** — a welcome case with `seedsInboundTurn` forced true via the `Object.defineProperty` pattern (pins the path every shipping adapter takes), and a `Thread`-level case with both `message` and `defaultPrompt` set. Verified the precedence test goes red against the previous ordering.

BenTaylorDev (approved):

PR #6384, opened 2026-08-05
fix(react-core): add missing available dep and cleanup to useCopilotReadable

### LINKED ISSUE #6383: useCopilotReadable: missing `available` dep + context leak on unmount
## Bug description

`useCopilotReadable` in `packages/react-core/src/hooks/use-copilot-readable.ts` has two bugs:

### 1. `available` missing from useEffect dependency array (line 132)

The `useEffect` dependency array is `[description, value, convert]` but the effect body reads `available`. Toggling `available` between `"enabled"` and `"disabled"` after mount has no effect — the context stays registered (or stays removed) because React never re-runs the effect.

### 2. Missing cleanup function on the `found` path (line 119)

When the hook finds an existing context entry that matches `{description, value}`, it sets `ctxIdRef.current` and returns early **without a cleanup function**. If the component unmounts while on this path, `removeContext` is never called and the context entry leaks.

The normal `addContext` path (line 128-131) correctly returns a cleanup function. The `found` path should do the same.

## Steps to reproduce

**Bug 1:**
```tsx
function Toggle() {
  const [on, setOn] = useState(true);
  useCopilotReadable({
    description: "toggle-data",
    value: "hello",
    available: on ? "enabled" : "disabled",
  });
  return <button onClick={() => setOn(!on)}>toggle</button>;
}
```
Click the button → context is **not** removed because `available` isn't in the dep array.

**Bug 2:**
Mount a component that calls `useCopilotReadable`, let the effect run so context exists, then re-mount the same component (same description/value). The `found` path fires. Now unmount → context is never cleaned up.

## Expected behavior

1. Changing `available` should add/remove context immediately.
2. Unmounting should always remove the context, regardless of which code path registered it.

## Fix

1. Add `available` to the dependency array.
2. Return a cleanup function on the `found` path identical to the one on the `addContext` path.

I'll submit a PR with the fix and regression tests.

anshusaurav:
## Summary

Fixes two bugs in `useCopilotReadable` (`packages/react-core/src/hooks/use-copilot-readable.ts`):

- **Missing `available` in useEffect deps**: toggling `available` between `"enabled"` and `"disabled"` after mount had no effect because React never re-ran the effect.
- **Missing cleanup on `found` path**: when the hook matched an existing context entry, it returned early without a cleanup function — context leaked on unmount.

## Changes

- Added `available` to the `useEffect` dependency array (line 135)
- Added a cleanup function on the `found` early-return path (lines 119-122), mirroring the existing cleanup on the `addContext` path
- Added 5 regression tests covering both bugs plus normal lifecycle

## Test plan

- [x] New test: context removed when `available` toggles to `"disabled"`
- [x] New test: unmount cleans up context on the `found` path
- [x] New test: normal add/remove lifecycle
- [x] New test: `available: "disabled"` from start skips `addContext`
- [x] New test: custom `convert` function is applied

Closes #6383

### DIFF
--- a/packages/react-core/src/hooks/__tests__/use-copilot-readable.test.tsx
+++ b/packages/react-core/src/hooks/__tests__/use-copilot-readable.test.tsx
@@ -1,0 +1,137 @@
+import React, { useState } from "react";
+import { render, act } from "@testing-library/react";
+import { describe, it, expect, beforeEach, vi } from "vitest";
+
+vi.mock("../../v2", () => ({
+  useCopilotKit: vi.fn(),
+}));
+
+import { useCopilotKit } from "../../v2";
+import { useCopilotReadable } from "../use-copilot-readable";
+
+const mockUseCopilotKit = useCopilotKit as ReturnType<typeof vi.fn>;
+
+function createMockCopilotKit() {
+  let nextId = 1;
+  const entries = new Map<string, { description: string; value: string }>();
+
+  return {
+    get context() {
+      return Object.fromEntries(entries);
+    },
+    addContext: vi.fn((item: { description: string; value: string }) => {
+      const id = `ctx-${nextId++}`;
+      entries.set(id, item);
+      return id;
+    }),
+    removeContext: vi.fn((id: string) => {
+      entries.delete(id);
+    }),
+  };
+}
+
+describe("useCopilotReadable", () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it("adds context on mount and removes on unmount", () => {
+    const mock = createMockCopilotKit();
+    mockUseCopilotKit.mockReturnValue({ copilotkit: mock });
+
+    const Harness: React.FC = () => {
+      useCopilotReadable({ description: "test", value: "hello" });
+      return null;
+    };
+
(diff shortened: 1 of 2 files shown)


BenTaylorDev:
Thanks for this, and sorry it sat long enough to go stale.

Closing because the defect is already fixed on `main`, by a different route. `useCopilotReadable` was rewritten as part of the v1 source move, and the current version at `packages/react-core/src/v1-deprecated/hooks/use-copilot-readable.ts` already does both things this PR adds:

- `available` is in the effect's dependency array, alongside `copilotkit` and the caller's own `...dependencies`
- the cleanup always runs and removes the context by the `id` captured in that effect, rather than relying on a ref that may have moved on

The `found` / `ctxIdRef` scan this PR patches no longer exists, so there's nothing left to re-target the change onto.

Separately, the file path here (`packages/react-core/src/hooks/use-copilot-readable.ts`) was relocated under `v1-deprecated/` in `e499c65dce`, so the branch no longer merges cleanly regardless.

Your diagnosis was right — a missing `available` dependency meant toggling it never re-ran the effect. It just got fixed from another direction before this landed. Thanks for reporting and fixing it.

Issue #6125, opened 2026-07-23
🚀 Feature Request: headless thread-switching API for React useAgent (parity with Vue cloneForThread); bundled react-core skill documents a threadId param that doesn't exist


ddubyah:
## Use case

We build one-shot, stateless extraction agents over AG-UI (headless `useAgent` + `useFrontendTool({ followUp: false })`): a UI panel sends free text to an agent on typing-pause and applies the forced tool call's args to the form. Each invocation must be independent — a fresh thread every run — because server-side checkpointing is per-thread and a reused thread pollutes the next extraction with stale history.

## The gap

In the React package there is no public headless API to mint or switch a thread. Published `@copilotkit/react-core@1.63.1` types:

```ts
declare function useAgent({ agentId, updates, throttleMs }?: UseAgentProps): { agent: AbstractAgent };
```

— no `threadId` option, and no `newThread()`/clone affordance on the hook's return. Meanwhile the Vue package has `cloneForThread` on main. Today we do what the framework internals do:

```ts
agent.abortRun();
agent.threadId = crypto.randomUUID();
agent.setMessages([]);
```

This works (and `threadId` is a plain public field on `AbstractAgent`), but it's unblessed — nothing documents it as the supported mechanism, so every consumer building headless one-shot flows rediscovers it by reading source.

## Related doc inconsistency in the same package

The skill bundled inside `@copilotkit/react-core@1.63.1` (`skills/react-core/references/agent-access.md`) documents `useAgent({ agentId: "default", threadId: "t1" })` and "per-thread clones … cached in a module-level WeakMap keyed by `(registryAgent, threadId)`", citing `src/v2/hooks/use-agent.tsx:78-119` — an API surface that doesn't exist in the published package's types. Presumably the doc got ahead of (or diverged from) the implementation; as-is it sends readers to a parameter that silently doesn't exist.

## Ask

1. A public React API for headless thread switching — `useAgent({ threadId })` as the skill doc already describes, a `newThread()`/`cloneForThread` equivalent, or at minimum documenting the `agent.threadId` + `setMessages([])` mutation as the supported mechanism.
2. Fix `agent-access.md` to match the shipped API either way.

Happy to provide more detail on the one-shot pattern if useful — we've documented it end-to-end on our side and it works well; the thread-switching mechanism is the one piece resting on unblessed API.

BenTaylorDev:
Thanks for this, and sorry for the slow reply on our side.

Both asks have since shipped.

[...]

PR #6388, opened 2026-08-05
docs: refresh v2 SDK reference

AlemTuzlak:
## Summary

Refresh the v2 SDK reference against current main, preserving the existing three-page scope. Merge main into this branch without rewriting its history.

## Why

The original reference update predates the explicit distinction between the v1 `CopilotKit` compatibility wrapper and v2 `CopilotKitProvider`, newer agent-ID defaults, and Inspector changes.

## How

- Use `CopilotKitProvider` in v2 setup and error-handler examples, while retaining the legacy wrapper's public-key requirement for `onError`.
- Document current provider props, dynamic headers, debug verbosity, and localhost-only Inspector availability; remove the obsolete `inspectorDefaultAnchor` addition.
- Preserve main's thread-scoped `useAgent` documentation, add an example and headless imports, and clarify provider agent-ID fallback and throttle overrides.
- Describe the main and headless import surfaces and their provider limitations.

Validation: all three pages compile as MDX; documented provider prop names were checked against current TypeScript interfaces; `git diff --check` against current main passed. Focused validation ran through Nx. No full docs build was run. The merge-triggered broad package hook failed; its unrelated autoformat/lockfile changes were reverted, and lint/lockfile/package hooks were excluded for this docs-only merge. Binary, plugin-skill, Intelligence naming, and commit-message hooks remain enabled.

## Summary by CodeRabbit

- **Documentation**
  - Updated provider setup guidance for `CopilotKitProvider`, including authentication tokens, dynamic headers, agent configuration, generative UI, throttling, debugging, and error handling.
  - Clarified v2 and headless import paths, agent ID resolution, thread-scoped options, and private proxied agent usage.
  - Documented that the legacy `CopilotKit` component serves as a v1 compatibility wrapper.
  - Expanded API reference examples and clarified inspector and throttling behavior.

### DIFF
--- a/showcase/shell-docs/src/content/reference/components/CopilotKit.mdx
+++ b/showcase/shell-docs/src/content/reference/components/CopilotKit.mdx
@@ -36,6 +36,11 @@ Don't have it yet? Go to https://dashboard.operations.copilotkit.ai and get one 
 
 </PropertyReference>
 
+<PropertyReference name="licenseToken" type="string">
+  Signed license token for offline verification of Enterprise Intelligence
+  Platform features.
+</PropertyReference>
+
 <PropertyReference name="guardrails_c" type="{ validTopics?: string[]; invalidTopics?: string[]; }"  >
 Restrict input to specific topics using guardrails.
   @remarks
@@ -57,15 +62,20 @@ This feature is only available when using CopilotKit's hosted cloud service. To 
   The endpoint for the Copilot text to speech service.
 </PropertyReference>
 
-<PropertyReference name="headers" type="Record<string, string>"  >
-Additional headers to be sent with the request.
+<PropertyReference name="headers" type="Record<string, string> | (() => Record<string, string>)"  >
+Additional headers to be sent with runtime requests. Pass an object for static
+headers, or a function when headers need to be refreshed from current auth
+state.
 
 For example:
 
-```json
-{
-  "Authorization": "Bearer X"
-}
+```tsx
+<CopilotKit
+  runtimeUrl="/api/copilotkit"
+  headers={() => ({ Authorization: `Bearer ${getToken()}` })}
+>
+  {children}
+</CopilotKit>
 ```
 
 </PropertyReference>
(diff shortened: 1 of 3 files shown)


BenTaylorDev (changes requested):
Reviewed against current `origin/main`. Content is accurate and I confirmed the substantive claims against source. Three things before it can land.

**1. Blocker: `inspectorDefaultAnchor` no longer exists.** Removed from `react-core` by `b53fa8f1ea` ("refactor(inspector): consolidate framework integration"), from both the provider and the bundled skill doc. It existed when you branched, so this is drift, not an error. Drop that `<PropertyReference>`.

**2. Rebase needed, one conflict.** The branch is behind main and `CopilotKit.mdx` conflicts in the `onError` paragraph. Your side is the correct one: `onError` is a plain `copilotkit.subscribe({ onError })` with no `publicApiKey` gate (`packages/react-core/src/v2/providers/CopilotKitProvider.tsx:869-882`), so main's current "Requires publicApiKey" text is wrong and your rewrite fixes it. Main only renamed the product in that sentence, which is what produced the conflict. `index.mdx` auto-merges.

**3. `agentId` resolution order is missing a step.** The page says explicit `agentId`, then chat configuration, then `"default"`. The hook resolves `agentId ?? chatConfig?.agentId ?? providerAgentId ?? DEFAULT_AGENT_ID` (`use-agent.tsx:203-208`), so the provider-level `agentId` sits between the chat configuration and the default. As written it reads as though `<CopilotKitProvider agentId>` is ignored.

Verified correct, no action needed: the `headers` function overload (`CopilotKitProvider.tsx:131`); `licenseToken`, `selfManagedAgents`, `agents__unsafe_dev_only`, `openGenerativeUI`, `defaultThrottleMs` and `debug` all exist on main and none were already documented; the headless-entry export list, including `useRenderToolCall` being exported (`headless.ts:80`) while `useDefaultRenderTool`, `CopilotChat`, `CopilotPopup` and `CopilotSidebar` are not; and the `useAgent` thread-scoping content.

Note that CI on this PR ran on 2026-08-05 against the old base, so it is not evidence about current main.

Context: a user hit the stale `useAgent` page in #6125. #7025 covered the React Native copy of that page and has merged. It mirrors your changes with no file overlap, and it already carries the fix for item 3.

tylerslaton (approved):

PR #6391, opened 2026-08-05
[CPK-7630] Fix auto-dismiss developer announcement banner

AlemTuzlak:
## Linear
CPK-7630
https://linear.app/copilotkit/issue/CPK-7630/developer-announcement-banner-persists-on-every-refresh

## Root cause
`fetchAnnouncement()` treated the announcement as unseen whenever localStorage did not contain the current announcement timestamp, but only the explicit X dismissal called `markAnnouncementSeen()`. Opening the preview or inspector changed only in-memory state, so a refresh/remount showed the banner again.

## How it was fixed
Added a four-second auto-dismiss timer for the collapsed announcement preview. The timeout reuses `markAnnouncementSeen()` to persist the current timestamp, and is cancelled when the inspector opens, the preview is opened, or the element is disconnected. The preview now fades out with reduced-motion support. Added a fake-timer regression test covering timeout persistence across remounts and documented the behavior in the Inspector guide.

## Test plan
- [x] Affected-package pre-commit checks: web-inspector, React packages, and Vue test/publint/attw matrix passed
- [x] Web Inspector Vitest suite: 134 tests passed
- [x] Web Inspector TypeScript check and build passed
- [x] Shell docs TypeScript check and focused public-assets test passed


tylerslaton:
Fixed in a later change to the inspector!

PR #6392, opened 2026-08-05
CrewAI Flows and Conversational Flows full D6

ranst91:
## Summary

Bring CrewAI Flows and CrewAI Conversational Flows to the complete showcase D6 surface on the official integration stack:

- `ag-ui-crewai==0.3.0`
- `ag-ui-protocol==0.1.19`
- `crewai==1.15.11`
- GPT-5.4 across CrewAI showcase agents

The implementation follows the established D6 behavior while keeping execution native to CrewAI. It covers reasoning and chained tools, interrupts, shared and streaming state, generative UI, A2UI and recovery, multimodal image/PDF input, and the remaining showcase surfaces.

## Conversational Flows and docs

- Adds a Conversational Flows implementation with full feature parity with regular CrewAI Flows.
- Keeps a single CrewAI documentation integration and adds a focused guide for promoting a Flow to a Conversational Flow.
- Documents CrewAI `1.15.11` as the minimum supported version for Conversational Flows.
- All manifest capabilities have connected documentation; no separate feature-parity page or alpha guidance remains.

## Review hardening

- Offloads synchronous beautiful-chat backend tool work from the event loop and verifies heartbeat/cancellation behavior.
- Probes `reasoning-custom` and `reasoning-default` as independent D6 cells.
- Removes the legacy multimodal fixture collision and requires D6-specific response evidence.
- Forces the scheduling tool contract, preserves fallback slot selection, and treats protocol cancellation as cancellation.
- Restores shared `data/` staging parity between the shell and TypeScript showcase lifecycle, with an erosion guard.
- Materializes shared `data/` symlinks in both showcase build workflows and returns the frontend matrix test to the CI gate with catalog-derived counts.

## Surgical cancellation follow-up

This follow-up is deliberately limited to the two cancellation edge cases identified in review; it does not add or change demo inventory, UI, dependencies, or workflows.

- Makes Beautiful Chat's secondary `generate_a2ui` request genuinely cancellable by using `AsyncOpenAI`, while retaining the thread fallback for synchronous backend tools.
- Adds a narrowly version-scoped compatibility shim for the pinned `ag-ui-crewai==0.3.0` bridge so a resolved `null` remains distinct from cancellation. Only resolved-null is encoded as JSON `null`; cancellation remains blank, both captured bridge bindings are updated, and version drift fails loudly.
- Reuses the canonical shared `render_a2ui` schema for the secondary model request.

## Validation

- CrewAI Flows production-equivalent isolated D6 matrix: **green**
- CrewAI Conversational Flows production-equivalent isolated D6 matrix: **green**
- CrewAI Flows Python: **157 passing**
- CrewAI Conversational Flows Python: **162 passing**
- Showcase harness: **3,708 passing / 18 skipped**
- Showcase scripts: **2,505 passing**
- Both CrewAI production builds: **62/62 pages generated**
- Production Docker images build successfully with `ag-ui-crewai==0.3.0` installed
- Full live GPT-5.4 validation exercised both implementations, including ordinary chat, reasoning chains, interrupts, multimodal PDF/image input, and A2UI generation
- Final PR CI at `f97f0768ba`: **76 successful / 3 intentionally skipped / 0 pending / 0 failing**

The live matrix produced valid alternate model behavior for two deterministic fixture assertions: custom-catchall narration wording and A2UI recovery succeeding on the first valid render rather than forcing a malformed retry. Both underlying features completed successfully; these are harness-vs-live nondeterminism rather than integration failures.


contextablemark (changes requested):
# Draft response for PR #6392

Thanks for the thorough D6 work here. After tracing the added cells through the UI, runtime routes, CrewAI flows, the pinned AG-UI bridge, and the replay harness, the newly enabled demos are canonical showcase cells and are relevant to the integration.

I did find three items I think should be addressed before merge and three narrower edge cases.

## Higher priority

### P1 — Beautiful Chat blocks the async flow event loop

`showcase/integrations/crewai-crews/src/agents/beautiful_chat_flow.py:85` calls `backend_tool._run()` directly from the async `chat()` listener. `GenerateA2uiTool._run()` (`src/agents/tools/custom_tool.py:115-151`) constructs the synchronous OpenAI client and waits on `chat.completions.create()`.

CrewAI executes coroutine flow listeners on the event loop, and the AG-UI endpoint’s SSE reads, timeout handling, disconnect detection, and cancellation cleanup use that same loop. A slow secondary model call can therefore stall the current stream, prevent timeout/disconnect cleanup, and block unrelated requests on the worker.

Could this use an async implementation or offload the synchronous tool call with `asyncio.to_thread()`? I would also add a test with a delayed `GenerateA2uiTool` that proves an event-loop heartbeat and cancellation/timeout processing continue.

### P1 — D6 never exercises `/demos/reasoning-default`

`showcase/harness/src/probes/helpers/d5-feature-mapping.ts:174-177` maps both `reasoning-default` and `reasoning-custom` to the single `reasoning-display` probe, and `demosToFeatureTypes()` deduplicates that mapping. `d5-reasoning-display.ts:114-130` then chooses `/demos/reasoning-custom` whenever both demos are present.

That means the newly enrolled `/demos/reasoning-default` route can be completely broken while both reasoning pills remain green. The existing unit test actually codifies the custom-first behavior.

Please split the probe/mapping or otherwise execute and assert both routes independently before claiming both cells as covered.

### P1 — multimodal D6 is currently a deterministic false-green

The showcase workflows load fixture directories in this order: `shared`, `d4`, `d5-recorded`, then `d6`. The old CrewAI D4 chat fixture contains the same image/PDF prompts as `showcase/aimock/d6/crewai-crews/multimodal.json`, and aimock preserves registration order when the earlier plain matcher competes with the later `turnIndex` matcher.

As a result, the D4 responses win. The shared multimodal probe only checks whether the transcript contains the words `image` or `document`, which those D4 responses do, so the pill passes without selecting either new D6 fixture.

Please remove or tighten the overlapping D4 entries (or change fixture selection semantics), and make the assertions distinguish the D6 responses so this collision cannot silently return.

## Narrower issues

### P2 — reasoning messages are replayed as an invalid Chat Completions role

`src/agents/tool_rendering_reasoning.py:109-114` appends a `{ role: "reasoning" }` message to `flow.state.messages`; lines 160-168 then send the full state into the next `acompletion()` call. Chat Completions does not accept `reasoning` as an input role, and the pinned LiteLLM OpenAI transformer does not filter it.

For an OpenAI-compatible endpoint that emits `reasoning_content`, a chained tool call can therefore fail on iteration two with HTTP 400. With official OpenAI Chat Completions, `reasoning_content` may not be emitted at all, leaving the advertised reasoning slot empty instead.

The reasoning snapshot should be kept outside model history, or filtered/mapped before the next completion. The test should combine a non-empty reasoning stream with a second completion call; the current tests cover those conditions separately.

### P2 — the interrupt fallback loses the selected slot when the model omits the tool call

`src/agents/interrupt_flow.py:82-110` supplies `schedule_meeting` but does not force it. If the model returns text without a tool call, the fallback still raises the human-feedback interrupt, but no `pending_tool_call_id` is stored. On resume, lines 144-166 skip the only tool-result/history insertion, so the second completion never receives the user’s selected slot.

Please either force/validate the `schedule_meeting` call or persist the resumed choice independently of the tool call. A text-only first response should be covered by a test.

### P2 — protocol-level cancellation is converted into a non-cancelled meeting

The public interrupt `cancel()` path sends a cancelled resume entry, which the pinned bridge exposes as empty feedback. `interrupt_flow.py:128-142` parses that as `{}` and records `cancelled: false` with a null time/label before asking the model to confirm it.

The shipped buttons avoid this because they call `resolve({ cancelled: true })`, so this is narrower than the other findings, but the standard cancellation path is still incorrect. Please configure a cancellation default or explicitly interpret the bridge’s cancelled resume outcome, and add a cancelled-resume test.

All current GitHub checks are green, so these are semantic and false-green coverage gaps rather than ordinary build failures. Overall, the direction looks sound; the three P1 items above are the ones I would treat as merge-blocking.

contextablemark (approved):
Approved after applying more changes.

AlemTuzlak (approved):

Issue #6408, opened 2026-08-05
🐛 Bug: v1 surfaces orphaned by the v1.50.0 re-implementation (readable context; server-side actions + MCP)


BenTaylorDev:
## Summary

Commit 80dffec4e7 ("Reimplement CopilotKit on top of refreshed internals", v1.50.0, #2638) repointed `useCopilotReadable` from the v1 context tree onto the v2 flat context store. The writer moved; the readers did not. The v1 tree now has no writers at all, and four features that read from it silently produce nothing.

This is a tracking issue for the surfaces left behind. The hook-level defects from that same commit are fixed in #6409 (see #6383, #6243).

## The v1 tree has no writers

The only remaining references to the v1 provider's `addContext` in `packages/react-core/src/context/copilot-context.tsx` are the type declaration (line 143) and the default no-op stub `() => ""` (line 278). Nothing calls it, so `printTree` always returns `""`.

## What still reads it

**1. `CopilotTask` — publicly exported and documented.**
`includeCopilotReadable` defaults to **true** (`copilot-task.ts:102`, `config.includeCopilotReadable !== false`). At `copilot-task.ts:129-133` it appends `context.getContextString([], defaultCopilotContextCategories)`, which now evaluates to the literal string `"\n\n"`. Every `CopilotTask` ships zero readable context while `reference/v1/classes/CopilotTask.mdx` promises otherwise. This is the most user-visible item here.

**2. Dev console `logReadables`** — `react-ui/src/components/dev-console/utils.ts:72-83` always prints "No readables found".

**3. CopilotTextarea autosuggestions** — `use-make-standard-autosuggestions-function.tsx:61` reads the empty tree. Moot in practice: the same port commented out the entire network call (lines 81-101) and replaced it with `const response: any = {}` (line 102), so the function unconditionally returns `""`. `@copilotkit/react-textarea` is published at 1.65.0 and is not marked deprecated.

**4. `useCopilotReadable`'s `parentId` / `categories`.**
Both were dropped from the hook body by the same commit but left in `UseCopilotReadableOptions` and in the JSDoc, whose top-of-file example was built entirely on `parentId`. #6409 marks them `@deprecated` and rewrites the example; this issue tracks whether they should come back.

## What the feature did

`parentId` built a tree that `printTree` rendered into an indented outline for the prompt — `1.` / `2.` at the root, `A.` / `B.` one level down, `a.` / `b.` below that. For a list of employees it produced:

```
[...]

PR #6409, opened 2026-08-05
fix(react-core): repair useCopilotReadable effect deps, convert args, and dependencies

### LINKED ISSUE #6243: 🐛 Bug: `useCopilotReadable` calls `convert` with the wrong arguments and ignores the `dependencies` param

### LINKED ISSUE #6383: useCopilotReadable: missing `available` dep + context leak on unmount

BenTaylorDev:
Fixes #6383. Fixes #6243.

Both issues land in the same 35 lines of `useCopilotReadable`, so they are fixed together. This PR also covers a third defect neither issue reports.

All of it traces to a single commit: 80dffec4e7 ("feat: Reimplement CopilotKit on top of refreshed internals (v1.50.0)", #2638), which repointed the hook from the v1 context tree onto the v2 flat context store. The pre-1.50 implementation was correct on every count below.

## Fixes

**`available` was missing from the effect deps** (#6383)
The effect body read `available` but the deps were `[description, value, convert]`, so toggling between `"enabled"` and `"disabled"` after mount did nothing. It is back in the deps, along with the `available = "enabled"` default the port dropped.

**`convert` was called with one argument** (#6243)
`(convert ?? JSON.stringify)(value)` invoked a user's `(description, value) => string` as `convert(value)`, so it received the value as `description` and `undefined` as `value`. The branches are now split rather than passing two arguments to the combined expression — `JSON.stringify(description, value)` would treat the second argument as a *replacer*, not a value.

**`dependencies` was accepted and ignored** (#6243)
The second positional argument was destructured but never reached the deps array. Now spread, matching `useCopilotAdditionalInstructions`.

**The `found` dedup branch was dead code** (unreported)
It compared `JSON.stringify({ description, value })` against a stored entry whose `value` had already been serialized by `addContext` (`packages/core/src/core/context-store.ts:36`). That never matches — for objects or strings — so the branch and its cleanup-skipping early return were unreachable. Deleted rather than repaired: making the comparison work would newly let component A's unmount remove a context entry component B is still relying on. The test `keeps separate entries for identical readables in two components` locks that in, and it passes against the pre-fix hook, which is what confirms the branch never fired.

## `parentId` / `categories`

Both are still in `UseCopilotReadableOptions` and were still documented — the top-of-file JSDoc example was a `parentId` tutorial — but the same v1.50 commit dropped them from the hook body. They have been no-ops since.

This PR does not implement them. Real support needs parent/child modelling in the v2 context store, which is flat by design (`getContextForAgent` emits `{ description, value }` only). Instead both are marked `@deprecated` and the JSDoc example is rewritten to document behavior that exists. Tracked in #6408.

## Not addressed

Two pre-existing behaviors left alone to keep this a bugfix:

- `value` is in the deps raw, so an inline object literal re-registers the entry on every render. Pre-1.50 depended on the serialized string instead.
- The hook returns `undefined` on first render, since the ref is assigned inside the effect.

## Testing

`useCopilotReadable` had no test file. This adds one — 12 tests, using a fake that mirrors `ContextStore` semantics (`addContext` assigns an id and stores the already-serialized value).

Full project suite — `nx run @copilotkit/react-core:test`:

```
 Test Files  124 passed (124)
      Tests  1487 passed (1487)
 NX   Successfully ran target test for project @copilotkit/react-core and 17 tasks it depends on
```

Each fix is covered by a test that fails against the pre-fix hook. Reverting only `use-copilot-readable.ts` and re-running the new file:

```
   ✓ registers the context on mount
   ✓ removes the context on unmount
   ✓ available > registers nothing when mounted as disabled
   × available > removes the context when flipped to disabled after mount
     → expected [ { description: 'employees', …(1) } ] to deeply equal []
   × available > re-adds the context when flipped back to enabled
     → expected [] to deeply equal [ { description: 'employees', …(1) } ]
   × convert > is called with (description, value) in that order
     → expected "spy" to be called with arguments: [ 'employees', …(1) ]
   × convert > is used in place of JSON.stringify
     → Cannot read properties of undefined (reading 'map')
   ✓ convert > serializes the value alone when convert is omitted
   × dependencies > re-runs the effect when a dependency changes
     → expected "spy" to be called 2 times, but got 1 times
   ✓ dependencies > does not re-run the effect when the dependency is unchanged
   ✓ re-registers when the description changes
   ✓ keeps separate entries for identical readables in two components

 Test Files  1 failed (1)
      Tests  5 failed | 7 passed (12)
```

The two that still pass pre-fix are deliberate: `serializes the value alone when convert is omitted` guards the `JSON.stringify` replacer trap in the fix itself, and `keeps separate entries…` is the evidence that the `found` branch was dead.

With the fix applied:

```
 ✓ src/hooks/__tests__/use-copilot-readable.test.tsx (12 tests) 15ms

 Test Files  1 passed (1)
      Tests  12 passed (12)
```

Types — `pnpm --filter @copilotkit/react-core check-types`:

```
> @copilotkit/react-core@1.66.2 check-types
> tsc --noEmit
```

(no diagnostics)

Formatting — `oxfmt --check` on both files:

```
Checking formatting...
Finished in 16ms on 2 files using 18 threads.
```

`oxlint` reports one warning, on `...(dependencies || [])` in the deps array. The same pattern already warns in `use-copilot-additional-instructions.ts`, `use-frontend-tool.ts` and `use-coagent-state-render.ts`; CI runs `oxlint .` without `--deny-warnings`.

The `convert` and `dependencies` fixes were independently found and fixed first by @jwgrsol in #6246, opened a week before this PR. Credited below.

Co-authored-by: jwgrsol <wefhio1985@gmail.com>

### DIFF
--- a/packages/react-core/src/hooks/__tests__/use-copilot-readable.test.tsx
+++ b/packages/react-core/src/hooks/__tests__/use-copilot-readable.test.tsx
@@ -1,0 +1,259 @@
+import { vi } from "vitest";
+import { renderHook } from "@testing-library/react";
+import { useCopilotReadable } from "../use-copilot-readable";
+
+type Availability = "enabled" | "disabled";
+
+interface FakeContextEntry {
+  description: string;
+  value: string;
+}
+
+/**
+ * Mirrors the semantics of `ContextStore` in @copilotkit/core: `addContext`
+ * assigns a fresh id and stores the *already serialized* value, `removeContext`
+ * deletes by id. Using a fake keeps these tests focused on the hook's effect
+ * logic, which is where every one of these regressions lived.
+ */
+function createFakeCopilotKit() {
+  const store: Record<string, FakeContextEntry> = {};
+  let nextId = 0;
+
+  return {
+    get context() {
+      return store;
+    },
+    addContext: vi.fn((entry: FakeContextEntry) => {
+      const id = `ctx-${++nextId}`;
+      store[id] = { ...entry };
+      return id;
+    }),
+    removeContext: vi.fn((id: string) => {
+      delete store[id];
+    }),
+    entries: () => Object.values(store),
+    ids: () => Object.keys(store),
+  };
+}
+
+let fake: ReturnType<typeof createFakeCopilotKit>;
+
+vi.mock("../../v2", () => ({
+  useCopilotKit: () => ({ copilotkit: fake }),
+}));
+
+// Stable across renders so that re-render assertions measure the dependency
+// array, not the identity churn of an inline literal.
+const EMPLOYEES = [{ name: "Jane" }, { name: "Sam" }];
+
+describe("useCopilotReadable", () => {
+  beforeEach(() => {
+    fake = createFakeCopilotKit();
+  });
+
+  it("registers the context on mount", () => {
+    renderHook(() =>
+      useCopilotReadable({ description: "employees", value: EMPLOYEES }),
+    );
+
+    expect(fake.addContext).toHaveBeenCalledTimes(1);
+    expect(fake.entries()).toEqual([
+      { description: "employees", value: JSON.stringify(EMPLOYEES) },
+    ]);
+  });
+
+  it("removes the context on unmount", () => {
+    const { unmount } = renderHook(() =>
+      useCopilotReadable({ description: "employees", value: EMPLOYEES }),
+    );
+    const [id] = fake.ids();
+
+    unmount();
+
+    expect(fake.removeContext).toHaveBeenCalledWith(id);
+    expect(fake.entries()).toEqual([]);
+  });
+
+  describe("available", () => {
+    it("registers nothing when mounted as disabled", () => {
+      renderHook(() =>
+        useCopilotReadable({
+          description: "employees",
+          value: EMPLOYEES,
+          available: "disabled",
+        }),
+      );
+
+      expect(fake.addContext).not.toHaveBeenCalled();
+      expect(fake.entries()).toEqual([]);
+    });
+
+    // Regression: `available` was missing from the effect's dependency array,
+    // so flipping it after mount was a no-op.
+    it("removes the context when flipped to disabled after mount", () => {
+      const { rerender } = renderHook(
+        ({ available }: { available: Availability }) =>
+          useCopilotReadable({
+            description: "employees",
+            value: EMPLOYEES,
+            available,
+          }),
+        { initialProps: { available: "enabled" as Availability } },
+      );
+
+      expect(fake.entries()).toHaveLength(1);
+
+      rerender({ available: "disabled" });
+
+      expect(fake.entries()).toEqual([]);
+    });
+
+    it("re-adds the context when flipped back to enabled", () => {
+      const { rerender } = renderHook(
+        ({ available }: { available: Availability }) =>
+          useCopilotReadable({
+            description: "employees",
+            value: EMPLOYEES,
+            available,
+          }),
+        { initialProps: { available: "disabled" as Availability } },
+      );
+
+      expect(fake.entries()).toEqual([]);
+
+      rerender({ available: "enabled" });
+
+      expect(fake.entries()).toEqual([
+        { description: "employees", value: JSON.stringify(EMPLOYEES) },
+      ]);
+    });
+  });
+
+  describe("convert", () => {
+    // Regression: convert was invoked as `convert(value)`, so a user's function
+    // received the value as `description` and `undefined` as `value`.
+    it("is called with (description, value) in that order", () => {
+      const convert = vi.fn(() => "converted");
+
+      renderHook(() =>
+        useCopilotReadable({
+          description: "employees",
+          value: EMPLOYEES,
+          convert,
+        }),
+      );
+
+      expect(convert).toHaveBeenCalledTimes(1);
+      expect(convert).toHaveBeenCalledWith("employees", EMPLOYEES);
+    });
+
+    it("is used in place of JSON.stringify", () => {
+      const convert = vi.fn(
+        (description: string, value: typeof EMPLOYEES) =>
+          `${description}: ${value.map((e) => e.name).join(", ")}`,
+      );
+
+      renderHook(() =>
+        useCopilotReadable({
+          description: "employees",
+          value: EMPLOYEES,
+          convert,
+        }),
+      );
+
+      expect(fake.entries()).toEqual([
+        { description: "employees", value: "employees: Jane, Sam" },
+      ]);
+    });
+
+    // Guards the fix itself: `JSON.stringify` must never receive the
+    // description as a first argument, because its second parameter is a
+    // replacer. `JSON.stringify("employees", EMPLOYEES)` would yield
+    // `"\"employees\""` rather than the serialized value.
+    it("serializes the value alone when convert is omitted", () => {
+      renderHook(() =>
+        useCopilotReadable({ description: "employees", value: EMPLOYEES }),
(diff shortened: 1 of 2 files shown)


jwgrsol:
Hi @BenTaylorDev, thanks for the thorough fix — especially catching the missing `available` in the effect deps, which my PR didn't cover.

One note: the two core fixes here (`convert` being called with the wrong arguments, and `dependencies` never reaching the deps array) are identical to #6246, which I opened a week earlier and which is linked to close #6243. My PR has been stuck waiting for a maintainer to approve its CI workflows, so it may have been easy to miss.

Since this PR is broader and further along, I'm happy for it to be the one that lands — but given the overlap, I'd appreciate being added as a co-author on the merge commit:

    Co-authored-by: jwgrsol <wefhio1985@gmail.com>

Thanks!

BenTaylorDev:
> Hi @BenTaylorDev, thanks for the thorough fix — especially catching the missing `available` in the effect deps, which my PR didn't cover.
> 
> One note: the two core fixes here (`convert` being called with the wrong arguments, and `dependencies` never reaching the deps array) are identical to #6246, which I opened a week earlier and which is linked to close #6243. My PR has been stuck waiting for a maintainer to approve its CI workflows, so it may have been easy to miss.
> 
> Since this PR is broader and further along, I'm happy for it to be the one that lands — but given the overlap, I'd appreciate being added as a co-author on the merge commit:
> 
> ```
> Co-authored-by: jwgrsol <wefhio1985@gmail.com>
> ```
> 
> Thanks!

Certainly, my bad on that omission.

BenTaylorDev:
Added to the PR body, also.

AlemTuzlak (approved):

PR #6430, opened 2026-08-07
docs: add Claude Managed Agents cookbook

ranst91:
## Summary

- add an independently runnable Claude Managed Agents finance-assistant example
- add a cookbook recipe that explains the CopilotKit runtime, managed-session mapping, and tool rendering flow
- add the recipe to cookbook navigation, the overview grid, sidebar icon mapping, and render coverage
- use the real Claude vector mark for the cookbook instead of the text-placeholder SDK asset
- register the example's Vite configuration and managed-agent model with the repository CI allowlists
- include a compact architecture diagram and links to the relevant rendering and CopilotKit Intelligence documentation
- disable Claude's complete built-in toolset and expose only the scoped `show_growth_projection` runtime tool
- make the provisioning model configurable through `ANTHROPIC_MODEL`, defaulting to `claude-fable-5`
- bound CopilotKit request bodies to 256 KB and managed-agent turns to 90 seconds, while relying on the adapter's per-thread serialization
- cap public run traffic at 20 provider-like attempts per client IP per minute and 2,000 successful starts per process per 24-hour window
- restrict browser runtime requests with an exact Origin allowlist that supports same-origin or separately hosted frontends, and restrict iframe parents with CSP `frame-ancestors`
- validate persisted managed-agent IDs at startup so malformed local configuration fails immediately
- publish the interactive example on Railway and embed the live deployment in the cookbook
- align the demo with the existing cookbook chat styling and show the `Project monthly investing` starter on first load

## Demo

Claude Managed Agents finance assistant rendering an interactive growth projection

## Why

This gives developers a focused example of connecting CopilotKit to Anthropic Claude Managed Agents without the extra surface area of a larger analyst application. The recipe follows the existing cookbook structure and keeps the live demo compact enough for the standard cookbook pane. Its managed environment has no outbound network or package-manager access, and its agent cannot use bash, filesystem, search, or fetch tools.

The request, turn, per-IP, and process-wide limits bound public demo traffic without adding authentication or user friction. The traffic counters are intentionally in memory, reset on process restart, and are not shared across replicas, so the dedicated Anthropic workspace spend limit remains the durable cost backstop. The exact-Origin browser check reduces drive-by use but is explicitly documented as a control rather than authentication. The model override allows operators to select a lower-cost supported model during provisioning without editing source code.

## Validation

- scoped formatting: passed
- scoped lint: 0 warnings, 0 errors
- shell-docs typecheck: passed
- standalone example typecheck: passed
- docs render tests: 26/26 passed
- standalone example tests: 23/23 passed
- shell-docs tests: 375/375 passed
- shell-docs production build: passed (222/222 pages)
- standalone example production build: passed
- standalone npm lockfile validation: passed
- build-config allowlist validator: passed
- docs model-name validator: passed
- exact-Origin regression coverage for run requests plus headerless same-origin runtime discovery: passed
- malformed persisted agent-ID regression coverage: passed
- live Railway root and iframe CSP: passed
- live Railway runtime discovery, exact welcome copy, and first-load starter pill: passed
- live three-turn AG-UI managed-agent run with `show_growth_projection`: passed
- cookbook verified in the browser at desktop and narrow widths with no console errors or horizontal overflow


contextablemark (dismissed):
Please make the managed-agent turn timeout configurable through a Railway service environment variable instead of hard-coding it to 90 seconds.\n\nUse , defaulting to five minutes ( ms) when the variable is absent, and pass the resolved value as . Please also:\n\n- validate that an explicitly supplied value is a finite positive integer rather than silently accepting an invalid or unbounded timeout;\n- document the variable and five-minute default in , the Railway deployment instructions, and the cookbook text where the current 90-second limit is described; and\n- cover both the unset/default case and configured-value case in the runtime-limit tests.

contextablemark (changes requested):
Please make the managed-agent model configurable rather than hard-coding `claude-fable-5`.

Have the setup/provisioning path resolve the model from an environment variable such as `ANTHROPIC_MODEL`, with `claude-fable-5` retained as the backward-compatible default when it is unset. Pass the resolved value to agent creation, document it in `.env.example` and the deployment README, and test both the default and override paths.

Please also make the provisioning-time behavior explicit: changing the variable does not mutate an already-created Anthropic agent. The agent must be reprovisioned and the resulting IDs updated on Railway before the deployed demo uses the new model. This should allow selecting a lower-cost model such as `claude-haiku-4-5` without editing source code.

contextablemark (dismissed):
Changes approved

tylerslaton (approved):

PR #6434, opened 2026-08-07
feat(reskinnable-demo): bi skin

guidovizoso:
## What does this PR do?

Adds **`vantage`**, a sixth skin to the reskinnable demo: an executive analytics
(BI) product for a B2B SaaS company. `demo-beats.md` names business intelligence
as the highest-stakes domain we hadn't built — the one that opens doors at
CFO/CRO level — so this fills that gap.

**78 files, ~16.2k insertions.** All but six are new files under
`src/skins/vantage/` and `src/app/api/vantage/`.

### The claim it exists to make

Every BI vendor already demos "chat with your data → get a chart." The
differentiator here is that **the agent's output is an object in the product,
not a message**: a generated dashboard is *filed*, gets a slug and a URL, and is
still there after the thread that created it is deleted. Boards compose the same
tiles a human gets, bound to the same metric definitions, so they're drillable
and governed rather than pictures.

Every figure is **derived**, not stored — `data/derive.ts` recomputes ARR, NRR,
pipeline coverage, CAC payback, logo churn and magic number from a 567-row
seeded fact table under a six-axis lens (period / compare / segment / region /
grain / currency). That's what makes the "pull the real levers" beat move actual
numbers instead of swapping canned answers.

### Shared-shell change (the only behavioural one)

`useSkinThemeReconcile` forced light for any skin without `--nw-dark-capable`
and honoured the stored preference for capable ones, so a skin shipping ONLY a
dark palette could not be expressed. Dark base tokens without the flag render a
light chat panel beside a dark app card; the flag alone lets a stored light
preference reintroduce the same mismatch, because the `.dark` class is global
while the palette is per-skin.

`--nw-theme-lock: dark` forces dark for such a skin and deliberately does not
write to `localStorage`, so a user's choice survives for dual-palette skins. The
four previously shipped skins declare nothing and are unaffected — covered by 5
tests in `use-theme.test.tsx`.

The other five shared-surface files are one-line registrations
(`registry.ts`, `agent-registry.ts`, `skins-config.ts`) plus an `eslint.config.mjs`
fix: `SKIN_IDS` was stale at four skins, so the no-hardcoded-route-prefix guard
was blind to both `people` and `vantage`.

### Demo beats

| Beat | Status |
|---|---|
| Gen-UI in transcript | ✅ 6 components |
| Rich thread survives reload | ✅ arg-only cards, replay-safe by construction |
| Drive the app, secret withheld | ✅ `connectSource` — credential never enters a tool argument or the transcript |
| "What's on my screen?" | ✅ route readable + 5 per-page readables |
| Navigate via levers | ✅ `exploreMetric` — confirm, navigate, visibly highlight |
| Multimodal → durable artifact | ✅ deck PDF → filed board at its own URL |
| Long-term memory / stored procedure / teach mode | ❌ **deferred to phase 2** |
| Presenter reset | ✅ |

Beats 4–6 are deferred, not missed — the per-user identity plumbing they need
(`RuntimeProviders`, `useRuntimeProperties`, server `identifyUser`) already
ships, so phase 2 is a seed file plus a persona switcher. The deferral is
recorded in `suggestions.ts`'s beat map and in the `CLAUDE.md` matrix.

### Worth a reviewer's attention

- **`data/derive.ts`** — the only part with no reference implementation in the
  repo. `planLevel` allocates planned net-new by ARR base at window start;
  an earlier version compared a net-new plan against an ARR *level* and
  allocated by *actual* net-new, which made every region's attainment
  algebraically identical. 20 tests, one verified falsifiable by perturbation.
- **`build-board-ops.ts`** — the a2ui envelope carries `version: "v0.9"` and
  `createSurface` has no `root` key (root is the component with `id: "root"`).
  A wrong envelope blanks the canvas with no error and no failing test.
- **`app/api/vantage/v1/sources/route.ts`** — the credential is shape-checked
  then discarded. A test asserts the 201 body contains neither the token nor the
  word "token", verified falsifiable.
- **`tools.tsx`** — 2 readables + 9 tool registrations, all unconditional; no
  HITL render keys off `status` (they key off `result`, so a replayed thread
  can't re-post).

### Testing

12 test files added. Full suite on the merged tree: **414 unit tests across 64
files**, lint 0 errors, `tsc` clean, `next build` compiling, plus 5 Playwright
tests covering the lever round-trip and the durable filed board — the two beats
that fail silently while passing every other gate.

## Related PRs and Issues

- (link the phase-2 tracking issue for beats 4/5/6, if one exists)

## Checklist


BenTaylorDev:
Closing — `main` filled this gap while the branch sat, under the same name.

`main`'s `exec` skin is the executive-BI demo and is branded **Vantage** (`src/skins/exec/identity.ts:41`), with `derive.ts`, `metrics-explorer`, `ceo-dashboard`, `cfo-dashboard`, `board-packs`, a2ui v0.9 blocks — plus `intelligence/seed-memories.ts` / `forget-memories.ts`, the memory/teach-mode beats this branch defers to phase 2. Merging would put two skins branded "Vantage" in the same selector.

The shell also moved:

- registry went 5 → 8 skins (`src/shell/skins-config.ts:16-25`)
- `eslint.config.mjs` replaced the `SKIN_IDS` string with an exported `LINTED_SKIN_IDS` array — this branch's edit targets deleted code and would revert `commerce`, `bookstore`, `exec` out of the lint guard
- the demo left the root pnpm workspace and installs published `1.71.0` from its own lockfile, so ~6.5k lines written against `workspace:*` have never compiled against what ships
- 5 of 7 shared files conflict

Credit where due: the `useSkinThemeReconcile` change is correct and non-regressive — the new `--nw-theme-lock` branch is unreachable for existing skins (undeclared custom property reads back `""`) and the token is scoped to `.theme-vantage`. It is just moot; `main`'s Vantage ships both palettes.

If we want the B2B-SaaS ARR/NRR domain specifically, that is a fresh PR against the 8-skin shell under a distinct brand and id. Fix first: `arr` is seeded as a period-end level but computed as `startingArr(windowStart) + Σ netNew(window)` (`data/derive.ts:151-165`), so periods ending in the same month disagree — `q3-2026` $15,341,227 vs `ttm` $19,262,898 at `end=2026-09`. TTM's comparison window also runs off the fact table and renders a green **+287.1%** on the headline tile. And `derive.test.ts` survives 8 of 10 semantic mutations to the arithmetic (inverted `pipeline_coverage`, inverted `magic_number`, dropped `cac_payback` `* 12`, dropped `logo_churn` annualisation, inverted FX, `startingArr` as a flow) — no test asserts a metric value.

Thanks for the work — the domain instinct was right, which is why `main` landed the same one.

PR #6275, opened 2026-07-31
feat: add Inspector navigation, usage, and locked Threads (refs ENT-1173)

onsclom:
## What does this PR do?

Adds the CopilotKit consumer side of ENT-1173 across Shared, Runtime, Core, Web Inspector, and the existing Shell Docs pages.

- Defines and parses optional trusted Inspector metadata for identity, plan, license, action, usage, and expiry. Runtime proxies it through a private, failure-isolated route, and Core refreshes it without changing connection state.
- Groups Inspector navigation into Threads, Agents, and Learning. Threads renders finite, unlimited, unknown, overage, and expiring usage states plus matching trusted plan or license actions.
- Keeps explicit `threadEndpoints` as the only authority for Thread requests. Locked or absent capability states make no list, subscription, detail, message, event, or state calls.
- Keeps the zero-thread video, three example Threads, detail tabs, and guided tour in empty and locked states. General Intelligence remains the default onboarding path; only trusted `team_self_hosted` metadata uses self-hosted onboarding.
- Gives an active license with missing Runtime routes a short **Finish setting up Rich Threads** state. Users can copy a safe coding-agent prompt or open the public Runtime setup guide. The same copy control appears in that guide, and raw Markdown/LLM views include the full prompt.
- Keeps finite usage green below 90%, orange from 90% to the limit, and red at or above the limit. At 90%, a trusted plan action changes from **Manage Your Plan** to a purple **Upgrade Your Plan** without changing its trusted URL, action kind, or telemetry contract.
- Adds a deterministic 33-state loopback lab for CopilotKit developers. It has no production route or export, is absent from public docs and package metadata, and is excluded from the npm tarball.

`Expiring Soon` is display-only; this PR does not enable the thread culler. Managed Enterprise receives no manage-plan action, and Team Self-Hosted receives no hosted plan action. Optional metadata and the additive expiry field remain compatible across mixed producer, Runtime, Core, and Inspector versions.

A small Channels test-only change updates fetch mocks for current TypeScript types. It changes no Slack or Teams docs or runtime behavior.

## Related PRs and issues

- Refs [ENT-1173](https://linear.app/copilotkit/issue/ENT-1173/ship-plg-ready-inspector-navigation-metadata-and-locked-threads)
- Producer: [CopilotKit/Intelligence#696](https://github.com/CopilotKit/Intelligence/pull/696)

## Validation

- `@copilotkit/web-inspector`: 20 files and 372 tests passed; typecheck and production build passed.
[...]

PR #6435, opened 2026-08-07
fix(ci): repair telemetry install and Inspector test flake

onsclom:
## What does this PR do?

- Install and run the telemetry registry as a standalone pnpm project in both fragment workflows.
- Stop at the install step with a clear error if the registry's local `ts-node` binary is missing.
- Check the Inspector's visible overage text instead of Lit's random internal HTML marker.

## Why?

PR #6275 merged successfully, then two `push` workflows failed on its merge commit:

- `telemetry / docs fragment`: the nested `p2p` install joined CopilotKit's parent workspace and did not create `p2p/node_modules`.
- Unit tests: `outerHTML` sometimes contained `241` inside Lit's random marker even though the visible count was `200+ / 200 Threads`.

The runtime fragment workflow had the same nested-workspace bug, so this fixes it before its first affected run.

## Related PRs and issues

- https://github.com/CopilotKit/CopilotKit/pull/6275
- https://github.com/CopilotKit/CopilotKit/actions/runs/31218832777/job/92998527566
- https://github.com/CopilotKit/CopilotKit/actions/runs/31218832813/job/92998528116

## Validation

- [x] `pnpm nx run @copilotkit/web-inspector:test --skip-nx-cache` (373 tests)
- [x] `pnpm nx run-many -t check-types,build --projects=@copilotkit/web-inspector --skip-nx-cache`
- [x] `actionlint` and `zizmor` on both edited workflows
- [x] Registry install, local binary guard, reconcile, and zero-diff check in a nested checkout
- [x] Formatter and lint checks

## Checklist

- [x] Documentation is not needed; this only repairs CI behavior and a test assertion
- [x] No changeset is needed; no package behavior or release output changes

### DIFF
--- a/.github/workflows/telemetry-docs-fragment.yml
+++ b/.github/workflows/telemetry-docs-fragment.yml
@@ -94,8 +94,12 @@ jobs:
         if: steps.token.outputs.token != ''
         working-directory: p2p
         run: |
-          pnpm install --frozen-lockfile
-          pnpm reconcile
+          pnpm install --frozen-lockfile --ignore-workspace
+          if [ ! -x node_modules/.bin/ts-node ]; then
+            echo "::error::Registry dependencies were not installed in p2p/node_modules."
+            exit 1
+          fi
+          pnpm --ignore-workspace reconcile
 
       - name: Open fragment PR
         if: steps.token.outputs.token != ''
--- a/.github/workflows/telemetry-runtime-fragment.yml
+++ b/.github/workflows/telemetry-runtime-fragment.yml
@@ -95,8 +95,12 @@ jobs:
         if: steps.token.outputs.token != ''
         working-directory: p2p
         run: |
-          pnpm install --frozen-lockfile
-          pnpm reconcile
+          pnpm install --frozen-lockfile --ignore-workspace
+          if [ ! -x node_modules/.bin/ts-node ]; then
+            echo "::error::Registry dependencies were not installed in p2p/node_modules."
+            exit 1
+          fi
+          pnpm --ignore-workspace reconcile
 
       - name: Open fragment PR
         if: steps.token.outputs.token != ''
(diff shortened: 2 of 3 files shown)


BenTaylorDev:
Closing — both halves of this have been fixed on main since, but one piece here is worth keeping and I'd rather flag it than let it go quietly.

### The telemetry half

`--ignore-workspace` on the p2p install landed in **#6621**, merged today. Same diagnosis: the nested checkout was resolving CopilotKit's parent workspace (`Scope: all 70 workspace projects`) instead of its own lockfile, so the registry's `@types/node` never installed and `reconcile` died on `TS2307: Cannot find module 'node:fs'`.

### The Inspector flake half

Fixed on main by `669132d731` ("stop lit's part marker from failing the usage-footer test"), by a different route — main strips comments before the substring check, where this PR moves from `outerHTML` to `textContent`. Main's comment documents the mechanism you were chasing: lit builds its part marker as `lit$` plus nine digits from `Math.random()`, so roughly one run in a hundred contained `241`.

### What's here that neither fix carries

Your fail-fast guard:

```bash
if [ ! -x node_modules/.bin/ts-node ]; then
  echo "::error::Registry dependencies were not installed in p2p/node_modules."
  exit 1
fi
```

Neither #6621 nor anything on main has this. It matters: without it, a failed registry install surfaces as a cascade of confusing `TS2307: Cannot find module 'node:fs'` / `Cannot find name '__dirname'` errors pointing at the registry's source, rather than naming the actual cause. That exact cascade burned three `telemetry-docs-fragment` runs on 2026-08-28 before #6621 landed, and #6621 fixes the *cause* without making a future recurrence legible.

I'm lifting this into a small PR against current main so it isn't lost, and will link it here once it's up. Full credit is yours — say the word if you'd rather carry it yourself and I'll close mine.

*From a sweep of stale drafts.*

BenTaylorDev:
Follow-up is up: #6793 — the guard, applied to both telemetry fragment workflows, with your attribution in the PR body and the commit message. Redirect it however you like.

PR #6436, opened 2026-08-08
fix(channels): retry WebSocket errors without close

tylerslaton:
## What changed

- Add a backoff-based fallback for established WebSocket errors that do not emit a matching close event.
- Cancel the fallback when Phoenix receives a normal open or close event, so the existing reconnect path stays in charge.
- Add a regression test that models Node's global WebSocket behavior and proves the session reaches `online` on a later retry.

## Why

Phoenix schedules socket retries from `onClose`, not `onError`. Node's global WebSocket can end a refused upgrade with `readyState = CLOSED` and an `error` event but no `close` event. After an established Intelligence session drops, that leaves Phoenix stuck after its first failed reconnect.

## Validation

- RED: the regression test stopped at 2 socket attempts before the fix.
- `pnpm --dir packages/channels-intelligence exec vitest run src/realtime-gateway-error-only-reconnect.test.ts`
- `pnpm nx test @copilotkit/channels-intelligence` — 213 tests passed.
- `pnpm nx check-types @copilotkit/channels-intelligence --skip-nx-cache`
- `pnpm nx build @copilotkit/channels-intelligence --skip-nx-cache`
- `pnpm exec oxfmt --check packages/channels-intelligence/src/realtime-gateway.ts packages/channels-intelligence/src/realtime-gateway-error-only-reconnect.test.ts`
- Pre-commit package test, publint, attw, and lint checks passed.

### DIFF
--- a/packages/channels-intelligence/src/realtime-gateway-error-only-reconnect.test.ts
+++ b/packages/channels-intelligence/src/realtime-gateway-error-only-reconnect.test.ts
@@ -1,0 +1,117 @@
+import { expect, test, vi } from "vitest";
+import { connectRealtimeGateway } from "./realtime-gateway.js";
+
+interface ErrorOnlyReconnectSetup {
+  disconnect(): void;
+  dropInitialConnection(): void;
+  isOnline(): boolean;
+  socketCount(): number;
+}
+
+/** Make the first reconnect emit only `error`, then let the next one recover. */
+async function setupErrorOnlyReconnect(): Promise<ErrorOnlyReconnectSetup> {
+  const sockets: ErrorOnlyWebSocket[] = [];
+
+  class ErrorOnlyWebSocket {
+    static readonly CONNECTING = 0;
+    static readonly OPEN = 1;
+    static readonly CLOSING = 2;
+    static readonly CLOSED = 3;
+    readyState = ErrorOnlyWebSocket.CONNECTING;
+    bufferedAmount = 0;
+    binaryType = "arraybuffer";
+    onopen: (() => void) | null = null;
+    onmessage: ((event: { data: string }) => void) | null = null;
+    onerror: ((error: Error) => void) | null = null;
+    onclose: ((event: { code: number }) => void) | null = null;
+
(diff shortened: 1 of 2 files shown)


BenTaylorDev:
Closing — this landed on main independently while the PR sat.

`4f29570bd6` ("fix(channels-intelligence): recover from error-only websocket failures", 2026-08-10) fixes the same defect by the same mechanism: a timer armed from the socket error handler that cycles the Phoenix socket when the transport reports an error with no matching close. Main names it `errorWithoutCloseTimer`; this branch names it `errorOnlyReconnectTimer`. It carries its own regression test at `realtime-gateway.test.ts:568`.

That was two days after you opened this. Nothing was wrong with the diagnosis — the write-up here is the clearer of the two on *why* Phoenix never schedules the retry.

**How I confirmed it rather than assuming.** I reverted this branch's change to `realtime-gateway.ts`, kept this branch's test, and ran it against main's implementation: it passes. Main's fix satisfies this PR's own scenario, so the branch is genuinely redundant and not merely similar-looking.

### One difference worth a look, separately from this close

The two retry policies are not identical:

| | main (`4f29570bd6`) | this branch |
|---|---|---|
| delay | fixed `ERROR_WITHOUT_CLOSE_GRACE_MS = 100` | ladder: 10, 50, 100, 150, 200, 250, 500, 1000, 2000, then 5000 |
| attempts | re-arms on each error; no counter | counter, reset once the socket is open |
| before cycling | always disconnects | skips the cycle when `connectionState() === "open"` |
| close frame | none | `1001`, reason `"retrying transport error"` |
| timer | `.unref()` | — |

So against an upstream that keeps failing, main retries every 100 ms indefinitely, where this branch would back off to 5 s. Main's `.unref()` is the better half of the trade.

That's worth deciding on its own merits — but as a small follow-up against current main, not by reviving a branch that's 1,065 commits behind. Happy to open one if you think the constant-100 ms retry is a real risk under a sustained 5xx.

*From a sweep of stale drafts.*

PR #6438, opened 2026-08-08
refactor(react-native)!: converge tool-call rendering onto CopilotKit's shared registry

mxmzb:
## What does this PR do?

`@copilotkit/react-native` maintained a **private tool-call render registry** (`hooks/RenderToolContext.tsx`) alongside the canonical one that `CopilotKitCoreReact` already provides — and which every React Native app already ships, unused. This PR deletes the fork and points React Native at the shared registry.

That fork caused three bugs:

| Bug | Symptom | Cause |
|---|---|---|
| **Tool renders never streamed** | A component registered with `useRenderTool` / `useComponent` painted nothing until the tool call completed | `CopilotChat` used `JSON.parse` on the argument buffer. While a model writes a tool call that buffer is *invalid JSON by design* — AG-UI delivers `TOOL_CALL_ARGS` deltas that are concatenated client-side — so the parse threw on every delta, warned, and fell back to `{}` |
| **`useComponent` rendered nowhere** | Silently, with no error | It writes to core's registry; React Native's chat read React Native's private `Map` |
| **Chat history degraded** | Navigating away from the registering screen turned earlier tool calls into a `Called: <name>` placeholder | The private `Map` deleted renderers on unmount; core deliberately keeps them |

`@copilotkit/react-core` has used `partialJSONParse` on this path since v2 shipped. React Native diverged because `useRenderToolCall` was excluded from its re-exports on the stated grounds that it "depends on DOM elements via `DefaultToolCallRenderer`" — a claim that was never true of the hook itself. It was only ever reachable through the fat `/v2` entry, whose weight is the real hazard (#4893). #5883 moved it into `/v2/headless` on 2026-07-23; the exclusion comment was rewritten the next day without revisiting the reason.

### What changed

- **One registry.** `useRenderTool` registers through `useFrontendTool` into `CopilotKitCoreReact.renderToolCalls`. `CopilotChat` and any custom surface consume react-core's `useRenderToolCall`.
- **Types are derived, not declared.** `RenderToolProps` is now `React.ComponentProps<ReactToolCallRenderer<T>["render"]>`, so React Native cannot drift from `ReactToolCallRenderer` — the contract every registered renderer is actually invoked against. Change that contract and `check-types` names every React Native renderer the change breaks. React Native narrows only the *return* type to `ReactElement | null`, which `FlatList`'s `renderItem` genuinely requires.
  _Scope of that guarantee (corrected during review):_ it does **not** extend to the type react-core publicly exports under the same name. Web's `RenderToolProps<S>` (`react-core/src/v2/hooks/use-render-tool.tsx`) is a separate hand-declared union, generic over a schema, carrying arguments under `parameters` (not `args`) and declaring `status` as string literals rather than `ToolCallStatus` members. Both divergences are live today and nothing type-checks them shut — the one place the shapes meet, react-core's own bridge, compiles because a string-enum member is assignable to its own literal type but not the reverse. Aligning web's alias is a breaking web API change, filed separately.
- **`RenderToolContext.tsx` deleted** (−150 lines), along with 15 tests that described the removed subsystem. One of them — `unregisters the render function on unmount` — asserted the chat-history bug as a requirement.
- **Two structural CI guards for #4893**, in opposite directions: a test failing if any React Native source imports the fat `/v2` entry, and a script failing if react-core's `/v2/headless` or `/v2/context` chunks ever link shiki/mermaid/cytoscape/katex/streamdown. Both were verified able to fail by deliberately introducing the regression. These are *structural* assertions, not size budgets — `dev-docs/bundle-size.md` freezes `limit` fields until OSS-122.
- **`react-native` added to the bundle-size glob**, which it had never been in, plus a `size:headless` measurement.

React Native also gains capabilities it lacked: render props inferred from your schema, `name`/`toolCallId` on render props, and `result` on completed calls.

**Corrected during review — two capabilities this originally claimed are not delivered:**

- **Wildcard (`"*"`) renderers do not work on React Native.** Because `useRenderTool` routes through `useFrontendTool` (which calls `addTool`), `name: "*"` registers a frontend tool literally named `*` — advertised to the model, and colliding with core's separate wildcard-executable-tool path. react-core's `useRenderTool` is renderer-only and special-cases the wildcard; React Native's is not. The guide now advises against it.
- **`followUp` (and `available`) are not forwarded**, and the handler's `context` argument is dropped, so `stopAgent()`'s abort signal is unreachable from an RN handler.

Both are tracked in § Known limitations for the follow-up that converges React Native onto react-core's hooks — deleting RN's `useRenderTool` in favour of re-exporting `useFrontendTool` (tool + renderer) and react-core's `useRenderTool` (renderer-only, wildcard-capable). That is an API change with its own migration note, so it is not in this PR.

### ⚠️ Breaking (in a minor)

`useRenderToolRegistry` and `RenderToolProvider` are **removed**. Both are documented on the docs site, so this is a real break — see the `BREAKING CHANGE:` footer on `db67ccf`, which is what the release notes derive from, plus the rewritten reference pages.

```diff
- const registry = useRenderToolRegistry();
- const renderer = registry.get(toolCall.function.name);
- return renderer ? renderer({ args, status }) : null;
+ const renderToolCall = useRenderToolCall();
+ return renderToolCall({ toolCall });
```

Also note two semantic changes: `args` is `Partial<T>` **only** while `status` is `"inProgress"`, and a render function is now captured at registration — if it closes over changing values you must declare them in `deps` (React Native previously refreshed the closure on every render).

**Known limitation:** agent-scoped renderer resolution does not take effect on React Native. `CopilotChatConfigurationProvider` is not in RN's provider tree, so `agentId` always resolves to the default. Renderers still resolve by name; two agents registering the same tool name resolve arbitrarily. Filed separately.

### A data point worth recording

Adding `useRenderToolCall` to the measured headless entry moved the bundle **92.8 kB → 92.7 kB**. Flat. The hook React Native spent months not using was already inside the chunk every RN app resolves whole — Metro doesn't tree-shake, so the fork never saved a byte. It cost them.

### Testing

- `@copilotkit/react-native`: **253 passing / 22 files** · `@copilotkit/react-core`: **1480 passing / 123 files** · `check-types` and `build` green for both.
- Each of the three bugs has a deterministic test driving a real `CopilotKitCoreReact` — no mocking of the code under test.
- Both #4893 guards carry mutation evidence: introduce the regression, watch them fail, revert, watch them pass.

### Follow-up

`useRenderTool`'s JSDoc is split across two blocks, which orphans the primary description from IDE hover (the `@param deps` warning still surfaces). One-line fix, deliberately left out of the final fix wave.

## Related PRs and Issues

- **Supersedes #6346** (@davidmckayv) — its diagnoses were correct and its test assertions are ported here, re-driven through the real registry rather than a mocked local one. Credited via `Co-Authored-By` on `4104bd1`.
- Addresses the React Native half of **#4893**.
- Builds on **#5883**, which created the lean `/v2/headless` entry this PR consumes.

## Checklist


mxmzb:
## Evidence: the #4893 hazard is absent from what React Native ships

The CI guards prove *future* regressions get caught. They don't prove today's bundle is clean, so I measured it — bundling each entry a consumer could import and grepping the **actual output bytes** for the payload, rather than trusting the import graph.

### 1. What an app bundles today

| Entry | gzip | heavy stack present? |
|---|---:|---|
| `@copilotkit/react-native/headless` | **98.8 kB** | clean |
| `@copilotkit/react-native` (default barrel, incl. chat UI) | **102.5 kB** | clean |
| `@copilotkit/react-core/v2/headless` | 99.2 kB | clean |
| `@copilotkit/react-core/v2` — the fat entry | **2,969.1 kB** | **shiki, mermaid, cytoscape, katex, streamdown** |

The hazard is real and unchanged at ~2.9 MB gzipped (consistent with the ~15 MB uncompressed reported in #4893). React Native links none of it — including via the **default barrel**, which I checked specifically because it re-exports the chat components and is the likeliest place to leak.

That last row doubles as the control: if the method couldn't detect the heavy stack, it would have come back clean too.

### 2. What a regression would cost

I appended the exact violating import to `packages/react-native/src/headless.ts` and rebuilt:

```
98.8 kB  →  2,931.5 kB gzip      CONTAINS shiki, mermaid, cytoscape, katex, streamdown
```

**30× from a single import line.** Metro does not tree-shake, so there is no recovery from this at the app level — which is why it needs a build-time guard rather than a review convention.

### 3. That the guard stops it

With the violating import present:

```
NX   Running target test for project @copilotkit/react-native ... failed
     ELIFECYCLE  Test failed.
```

Reverted (working tree verified identical to HEAD), same guard:

```
✓ src/__tests__/headless-entry-surface.test.ts (11 tests) 542ms
  Test Files  1 passed (1)
       Tests  11 passed (11)
```

The companion guard covers the opposite direction — `packages/react-core/scripts/assert-headless-purity.mjs` fails if react-core's own `/v2/headless` or `/v2/context` chunks ever link the stack, which React Native could not detect on its own. Both were verified able to fail by deliberately introducing the regression.

### Caveat

These are esbuild figures, not Metro's — a real React Native release bundle will differ in absolute terms. The method is identical across all four rows, so the **ratio** and the **presence/absence of the heavy stack** are the transferable claims; the absolute kB are a relative signal only. The per-PR number in CI (`pnpm --filter @copilotkit/react-native size:headless`) uses the same method, so it is comparable across PRs.

### Reproducing

Bundle any of the four specifiers with `bundle: true, minify: true, platform: "browser"`, resolving from `packages/react-native`, then search the output for `shiki` / `mermaid` / `cytoscape` / `katex` / `streamdown`. The measurement script committed in this PR (`packages/react-native/scripts/measure-headless.mjs`) is the same harness with the entry list fixed to the lean path.

packages/react-native/src/hooks/useRenderTool.ts
```diff
@@ -12,85 +10,59 @@ export interface UseRenderToolOptions<T extends Record<string, unknown>> {
...
     {
       name,
       description,
       parameters,
       handler,
       agentId,
+      // No cast needed: ReactFrontendTool.render is ReactToolCallRenderer<T>["render"],
+      // and RenderToolFunction is derived from exactly that props type, returning
+      // ReactElement | null (assignable to ComponentType's ReactNode).
+      render,
```
mxmzb, line 62: **★ THE core fix — start here.** This one call is what collapses two registries into one.

`useFrontendTool` writes the renderer into `CopilotKitCoreReact.renderToolCalls` — the same registry `useComponent` writes to, and the one RN's provider already instantiated and never used.

Directly solves **two of the three reported bugs**:
- *`useComponent` rendered nowhere* — it wrote to core's registry while RN's chat read RN's private `Map`. Now there is one registry.
- *Chat history degraded to `Called: <name>`* — the private `Map` deleted renderers on unmount; core deliberately retains them.

Note there is no cast on `render` (see the comment just above): `ReactFrontendTool.render` **is** `ReactToolCallRenderer<T>["render"]`, and `RenderToolFunction` is derived from exactly that props type. If a cast were needed here, the convergence would be cosmetic.

packages/react-native/src/components/CopilotChat.tsx
```diff
@@ -102,19 +191,43 @@ export function CopilotChat({
...
 
-  const { copilotkit, executingToolCallIds } = useCopilotKit();
+  const { copilotkit } = useCopilotKit();
   const { agent } = useAgent({ agentId: agentName });
 
   const messages = agent.messages ?? [];
   const isRunning = agent.isRunning;
 
-  const toolRenderers = useRenderToolRegistry();
+  const renderToolCall = useRenderToolCall();
```
mxmzb, line 200: **★ The consumption half of the fix** — and the reason bug 1 (*tool renders never streamed*) goes away.

Before, this component did `JSON.parse(tc.function.arguments || "{}")`. While a model writes a tool call that buffer is **invalid JSON by design** — AG-UI delivers `TOOL_CALL_ARGS` deltas that are concatenated client-side — so the parse threw on every delta, warned, and fell back to `{}`.

The fix is not a better parser: RN stops parsing arguments **at all** and delegates to react-core's `useRenderToolCall`, which has used `partialJSONParse` on this path since v2 shipped. `JSON.parse` now has zero occurrences in this file — the bug class is removed rather than patched.

packages/react-native/src/components/CopilotChat.tsx
```diff
@@ -102,19 +191,43 @@ export function CopilotChat({
...
 
-  // Stable extraData for FlatList to avoid re-creating the object every render
-  const extraData = useMemo(
-    () => ({ isRunning, executingToolCallIds, toolRenderers }),
-    [isRunning, executingToolCallIds, toolRenderers],
-  );
+  // Recomputed every render — cheap, and the only honest dependency for the
+  // message-derived memos below. See `messagesFingerprint` for why the array
+  // reference cannot be used.
+  const messagesKey = messagesFingerprint(messages);
```
mxmzb, line 260: **⚠️ Pay attention here — without this line the headline fix above does not fire in production.**

These memos were keyed on `[messages]`, i.e. on `agent.messages` **array identity**. Identity is not a reliable change signal, in either direction:

- **It does change** on streamed deltas — `@ag-ui/client`'s `AbstractAgent.processApplyEvents` reassigns `this.messages` per applied event. (Credit to @BenTaylorDev for probing a real AG-UI run to establish this.)
- **It does not change** on the paths this memo exists to serve — core inserts tool results with `agent.messages.splice(insertAt, 0, toolMessage)` (`packages/core/src/core/run-handler.ts:931`, `:1080`), `AbstractAgent.addMessage` is a `this.messages.push(...)`, and `useAgent` re-renders with a bare `forceUpdate()` (`react-core/src/v2/hooks/use-agent.tsx` ~L389).

So a tool result spliced into the existing array left both memos frozen at whatever the first render of a run saw: renderers kept receiving `result: undefined` with a status that never reached complete, and `listItems` only recomputed when `isRunning` flipped, so messages appended mid-run never reached the flat list at all.

**Why 253 tests were green over this:** the RN chat suites drove messages by re-rendering with a **new array**, which does change identity. Production splices, which does not — the suite exercised a mutation mode core never uses on that path.

The fingerprint keys on ids, roles, content, `toolCallId`, and tool-call ids + argument lengths, following the shape of react-core's web `messagesMemoKey`.

---

**Correction (edited).** This comment originally justified the fix by asserting *"nothing in production ever reassigns `.messages` — grepping for that assignment in `packages/core/src` hits test files only."* The grep was accurate but the conclusion was over-general: the reassignment lives in `@ag-ui/client`, outside that path. The fix stands — core's `splice`/`push` paths genuinely leave the reference untouched — but the stated reason was wrong and has been corrected here and in the source comment.

packages/react-native/src/components/CopilotChat.tsx
```diff
@@ -102,19 +191,43 @@ export function CopilotChat({
...
+        id?: string;
+        toolCallId?: string;
+        content?: unknown;
+      };
+      if (m.role === "tool" && m.toolCallId) {
+        byId.set(m.toolCallId, {
+          id: m.id ?? m.toolCallId,
+          role: "tool",
+          toolCallId: m.toolCallId,
+          content: toolResultContent(m.content, m.toolCallId),
```
mxmzb, line 224: **Silent data loss, and a platform divergence inside a convergence PR.**

This was `typeof m.content === "string" ? m.content : ""` — a non-string tool result became `""`, which is **indistinguishable from a genuinely empty result**, with nothing logged.

Reachability is subtle and worth recording: `ToolMessageSchema.content` is `z.string()` and the SSE transport zod-parses every event, so the branch is unreachable on the *validated* path. But core keeps the identical hedge (`normalizeToolResultContent`, `run-handler.ts:721`) because protobuf transport, restored thread history and `as Message` casts all bypass validation.

The clincher: web's `CopilotChatToolCallsView.tsx:23` passes content through **uncoerced**. So on any unvalidated path RN was destroying data web preserves — exactly the asymmetry this PR exists to remove.

packages/react-native/src/hooks/render-tool-types.ts
```diff
@@ -0,0 +1,55 @@
...
+ * `name`/`toolCallId` entirely, and RN's `args` was unconditionally the full
+ * `T`, promising complete arguments even before they had finished streaming.
+ *
+ * What the derivation buys, precisely: RN's props cannot drift from
+ * `ReactToolCallRenderer` — the contract every registered renderer is
+ * actually invoked against, and the type this alias reads through
+ * `React.ComponentProps`. Change that contract and RN's public type changes
+ * with it, so `check-types` names every RN renderer the change breaks.
+ *
+ * What it does NOT buy: parity with the type react-core *publicly exports*
```
mxmzb, line 20: **★ Conceptual gap — worth reading before merge, and the PR description needs a matching edit.**

The derivation below is real and valuable: RN's props **cannot** drift from `ReactToolCallRenderer`, the contract every registered renderer is actually invoked against. That is the drift that caused the three bugs, and it is now structurally closed.

What it does **not** buy is the symmetric claim in the PR body — *"React Native and web cannot re-diverge — if the contract changes, `check-types` says so."* As written that is false:

- react-core **publicly exports** its own `RenderToolProps<S>` (`src/v2/hooks/use-render-tool.tsx`, L9-36) as a separate hand-declared union: `parameters` (not `args`), and `status` as the string literals `"inProgress" | "executing" | "complete"` (not `ToolCallStatus` members).
- Both divergences are **live today, in this PR**. Two packages export a type named `RenderToolProps` with different field names and different status types.
- If the canonical contract changes, RN follows automatically and web's public alias silently does not — and nothing fails, because the one place the shapes meet (react-core's own bridge, `use-render-tool.tsx` ~L178-186) spreads the enum-typed renderer props into the literal-typed slot and compiles: a string-enum member is assignable to its own literal type, but not the reverse. Web's `status` is a **widening** of the canonical contract, not a derivation from it.

Confirmed empirically: flipping the canonical `status` from enum to literals produces 6 `tsc` errors, **all in unrelated files and zero in the type guard** — and RN has no such incidental bystanders, so on the RN side that drift would be completely silent.

Suggested body edit: keep the guarantee but scope it to *"RN cannot drift from `ReactToolCallRenderer`"*, and note that aligning react-core's public alias is a breaking web API change, filed separately.

packages/react-native/src/headless.ts
```diff
@@ -80,31 +84,49 @@ export {
...
 
+// Re-export core types commonly needed
+export type { Suggestion, FrontendTool } from "@copilotkit/core";
+
+// AbstractAgent is a runtime CLASS, not a type: it is the AG-UI extension point
+// consumers subclass (`class MyAgent extends AbstractAgent`) and test with
+// `instanceof`. @ag-ui/client is a dependency of this package rather than a peer,
+// so a consumer cannot reliably import the class from there directly — a
+// type-only re-export left the extension point unreachable from RN.
+export { AbstractAgent } from "@ag-ui/client";
```
mxmzb, line 107: **Runtime values must be exported as values, not types.** Five symbols sat inside `export type { … }` blocks, which strips the runtime binding: `ToolCallStatus`, `UseAgentUpdate`, `CopilotKitCoreErrorCode`, `CopilotKitCoreRuntimeConnectionStatus`, and `AbstractAgent`.

Consequences for consumers, all reproduced with a probe compiled under the package's real `tsconfig` (10 × `TS1362` before, 0 after, on both entry points):

- `status === ToolCallStatus.Complete` — the canonical idiom — **could not compile**, because the enum value was unimportable. A string literal is not assignable to the enum type either (`TS2322`), so there was *no* working way to branch on `status`, on the very prop this PR's contract is built around.
- `AbstractAgent` is a **class**, so `instanceof` and `extends` were both impossible.
- `useAgent({ updates })` was unusable, since `UseAgentUpdate` members were unreachable.

Bundle impact of the correction: **zero** — 92.7 kB before and after.

packages/react-core/scripts/assert-headless-purity.mjs
```diff
@@ -0,0 +1,443 @@
...
+  loader = EMPTY_LOADERS,
+}) {
+  let result;
+  try {
+    result = await build({
+      entryPoints: [entryFile],
+      absWorkingDir: pkgRoot,
+      bundle: true,
+      write: false,
+      metafile: true,
```
mxmzb, line 270: **★ Why the bundle-size / #4893 guard is now real — this is the substantive change in the size story.**

As originally written this hard-fail gate substring-scanned a handful of **inlined entry files**. It was broken in both directions, and both were reproduced:

**False negatives (the dangerous kind).** `dist/v2/headless.mjs` imports `useThreads` from the **fat `@copilotkit/react-core` root entry** and the guard reported *clean*. A heavy dependency reached *through* `@copilotkit/core` also passed, because those are bare externals it never followed. The exact violation class the gate exists to catch walked straight past it.

**False positive.** Five banned tokens sitting in a **comment** made it exit 1. The built artifact is comment-preserving (233 comment lines survive in `dist/v2/headless.mjs`), so relocating a doc comment could fail CI with a misleading message. Related subtlety: the banned tokens in `src/v2/headless.ts`'s banner vanish not because comments are stripped, but because that module is a re-export shell whose banner attaches to no retained code.

The fix drives esbuild's `metafile` over the **resolved graph — 650 modules instead of a fixed file list** — strips comments before matching, matches real import specifiers rather than raw substrings, and fails loudly on unresolvable edges instead of treating them as clean. Ships **17 negative tests** wired into `test:scripts`, because a hard-fail CI gate with no test of its own is how this shipped.

packages/react-core/scripts/assert-headless-purity.mjs
```diff
@@ -0,0 +1,443 @@
...
+import { fileURLToPath } from "node:url";
+import { build } from "esbuild";
+
+const FORBIDDEN = ["shiki", "mermaid", "cytoscape", "katex", "streamdown"];
+// `fileURLToPath(import.meta.url)`, not `import.meta.dirname`: the latter landed
+// in Node 20.11 and is `undefined` below it, which would turn this hard-fail gate
+// into an opaque ERR_INVALID_ARG_TYPE for anyone on the `engines: ">=18"` floor
+// the root package.json still declares. Same idiom as scripts/measure-copilotchat.mjs.
+const dist = path.resolve(
+  path.dirname(fileURLToPath(import.meta.url)),
```
mxmzb, line 77: **This gate crashed on the repo's own declared Node floor.** It used `import.meta.dirname`, which does not exist before **Node 20.11** — it is `undefined` there, producing an opaque `TypeError [ERR_INVALID_ARG_TYPE]` five frames deep in `node:internal`, at module load, before any useful message.

Root `package.json` declares `engines: { "node": ">=18" }` and neither package overrides it. Verified against a **real Node 18.20.8**: pre-fix dies at load; post-fix walks all 650 modules and exits 0.

Both sibling scripts already used `fileURLToPath(import.meta.url)`; all three now match.

packages/react-native/scripts/measure-headless.mjs
```diff
@@ -0,0 +1,268 @@
...
+];
+
+// Sanity FLOOR, not a budget. The headless graph pulls @ag-ui/client, core and
+// react-core's headless entry; it measures ~92 kB gzip today and cannot
+// plausibly drop an order of magnitude. A total below this means the
+// graph was not really bundled (everything externalized, an empty/partial
+// dist, a stubbed entry) — i.e. a broken measurement, not an improvement.
+// Sits ~11x below the real figure so legitimate size work never trips it; if a
+// real change ever approaches it, move the floor in the PR that explains why.
+export const MIN_PLAUSIBLE_BYTES = 8 * 1024;
```
mxmzb, line 97: **★ Why the reported bundle number is now trustworthy.** This script's output is the evidence for the headline *"92.8 kB → 92.7 kB, flat"* claim, and as written it could silently report a triumph.

It had no zero-output guard (its react-core sibling exits 1), `logLevel: "silent"` discarded esbuild warnings, and there was no try/catch — so a broken run printed **"0.0 kB"** straight into the CI job summary.

The detail that made a literal `total === 0` check insufficient: reproducing the failure measured **35 bytes, not 0** — and 35 B still formats as `"0.0 kB"`. Copying the sibling's exact guard would have missed the real failure mode, because a broken resolve still emits a tiny stub. Hence a **plausibility floor** rather than a zero check.

Also fixed: an unbuilt `dist` now produces an actionable preflight naming the build command instead of a raw esbuild "Could not resolve" stack, and warnings are surfaced with source frames. Verified end-to-end through the CLI, not only via unit tests. Happy path still reports **92.7 kB, unchanged** — comparability across PRs is the entire point of the number.

packages/react-native/scripts/measure-headless.mjs
```diff
@@ -0,0 +1,268 @@
...
+//                   add ~56 kB gzip to this figure — a 60% inflation of a number
+//                   that exists as evidence for a bundle claim. Adding it does
+//                   not move the figure (92.7 kB before and after).
+//
+// Subpaths need no entries of their own: esbuild prefix-matches package paths, so
+// `react` also externalizes `react/jsx-runtime` and `react-dom` also externalizes
+// `react-dom/client` and `react-dom/server` (esbuild CHANGELOG 0.5.14 / 0.14.13).
+// `react/jsx-runtime` is kept explicit only because it documents that this entry's
+// JSX transform is the host's; it is redundant under that prefix rule.
+export const HEADLESS_EXTERNAL = [
```
mxmzb, line 83: **`react-dom` externalization — finding refuted, guard added anyway; the figure did not move.**

Reviewers flagged that `react-dom` was missing from `external` (unlike the sibling) and would inflate the measurement by "~130 kB". Checked with a metafile run: **0 of 653** input modules are `react-dom`, nothing references it even pre-resolution, and all four candidate `external` lists produce an identical **94941 B**. So there was no active inflation, and the headline figure is unaffected.

Two corrections worth recording:

- The "~130 kB" is the **raw** magnitude, not gzip. Measured inflation a stray edge would actually cause: `react-dom/client` **+56.3 kB**, `react-dom/server` **+57.3 kB**, bare `react-dom` **+1.4 kB**. Still ~60%, so the guard earns its place — and the nastiest case is the *smallest*, since +1.4 kB reads as noise rather than a regression.
- Subpath entries are unnecessary: esbuild **prefix-matches** package paths, so `react-dom` alone covers `/client` and `/server` (verified on the pinned 0.27.3). The sibling's longer list is deliberately not copied verbatim, because listing subpaths implies they are required.

mxmzb (commented):
## Reviewer's guide to this diff — what to look at, and what each change actually solves

This PR is 27 commits larger than when it was opened; those came out of a review pass and they change the story in two places worth knowing before reading the code.

### The three lines that carry the whole convergence

If you read only three hunks, read these — everything else is consequence or scaffolding:

1. **`useRenderTool.ts` — registration goes through `useFrontendTool`.** Collapses RN's private renderer `Map` into `CopilotKitCoreReact.renderToolCalls`. Fixes *`useComponent` rendered nowhere* and *chat history degraded on unmount* outright.
2. **`components/CopilotChat.tsx` — consumption via `useRenderToolCall`.** RN stops calling `JSON.parse` on a streaming argument buffer that is invalid JSON by design, and delegates to the `partialJSONParse` path react-core has used since v2. Fixes *tool renders never streamed*.
3. **`hooks/render-tool-types.ts` — the props type is derived, not declared.** Makes the collapse permanent against `ReactToolCallRenderer`.

### Two things the review changed that are worth not skimming

- **The headline fix was inert in production.** `toolMessages` and `listItems` were memoised on `agent.messages` **array identity**, but core inserts tool results with `splice` and `useAgent` re-renders with a bare `forceUpdate()` — so the dependency never changed. Renderers kept receiving `result: undefined`, and messages appended mid-run never reached the flat list. 253 tests were green over it because the test harness reassigns the messages array while production mutates it in place. Inline note on the fingerprint has the detail.
- **Both #4893 guards passed on real violations.** The purity script reported *clean* while `dist/v2/headless.mjs` imported from the fat root entry; the entry-surface guard was harvesting 8 phantom import specifiers from JSDoc examples, and its existence test reported `Tests no tests`. Both now carry evidence in the direction that matters — introduce the regression, watch it fail.

### Bundle size, specifically

Nothing about the **92.7 kB** figure changed — byte-identical at 94941 B, and `react-dom` was verified absent from the graph (0 of 653 modules), so the "flat" claim stands. What changed is whether that number can be trusted and whether the guards behind it work:

- the measurement script could print **"0.0 kB"** on a broken run — the real reproduction measured **35 B**, which still formats as 0.0, so a literal zero-check would have missed it. Now: plausibility floor, surfaced esbuild warnings, and an actionable message when `dist` is missing.
- the purity gate now walks the **resolved 650-module graph** via esbuild's metafile instead of substring-scanning a fixed file list, and ships 17 negative tests.
- it no longer crashes on Node 18 (`import.meta.dirname` needs ≥ 20.11 while root `engines` says `>=18`).
- `dev-docs/bundle-size.md` had six claims contradicted by the workflow it documents (action pin, package inventory, asserted entries, coverage overclaim) — corrected.

### One conceptual gap — and a small edit the description needs

The description says *"React Native and web cannot re-diverge — if the contract changes, `check-types` says so."* The first half is earned; the claim as written is not. react-core still **publicly exports** its own hand-declared `RenderToolProps<S>` using `parameters` and string-literal `status`, so the two same-named public types differ **today** — and `check-types` cannot see it, because the one bridge where the shapes meet compiles (a string-enum member is assignable to its own literal type). Detail inline on `render-tool-types.ts`. Suggested fix: scope the guarantee to *"RN cannot drift from `ReactToolCallRenderer`"*, and file the web-side alignment separately since it is a breaking web API change.

### Status

All 28 mandatory review findings are fixed. Local verification: **RN 287/287 tests** (23 files, default parallelism, 4.71 s — the flake is gone), react-native + react-core typechecks clean, react-core `test:scripts` 19/0, `oxfmt --check` clean, `oxlint` 0 errors.

Still outstanding: an unbiased confirmation review over these 27 commits, and ~15 deferred findings belonging to two follow-up subjects — pre-existing `CopilotKitProvider.mdx` docs accuracy, and pre-existing RN `CopilotChat`/provider defects such as attachment content collapsing to an empty bubble and a failed send destroying the draft.

packages/react-native/src/__tests__/headless-value-exports.test.ts
```diff
@@ -0,0 +1,369 @@
...
+    `${base}.tsx`,
+    path.join(base, "index.ts"),
+    path.join(base, "index.tsx"),
+  ]) {
+    if (fs.existsSync(c) && fs.statSync(c).isFile()) return c;
+  }
+  throw new Error(`cannot resolve local specifier "${spec}" from ${fromFile}`);
+}
+
+describe("react-native entries: no runtime value is re-exported type-only", () => {
```
mxmzb, line 252: **★ The most durable thing added here — a self-extending guard, not a symbol list.**

Five runtime values had been re-exported type-only. A guard covering those five would rot the moment someone re-exports a sixth enum.

This layer instead **parses both entry sources and fails any type-only re-export whose source module has a runtime binding** — no symbol list, empty allowlist, so it covers symbols that do not exist yet.

It matters that this lives on the **RN side**: react-core's `headless-type-exports.test-d.ts` structurally cannot catch this class. Proof — react-core's own `UseAgentUpdate` was *already* a correct value export while RN's was wrong, so a react-core-side guard is blind to RN's entry surface. And the stripped-binding failure is invisible to `tsc` by construction, which is why the sibling layers here are **runtime** assertions plus a type-level usage that fails `check-types` with `TS1362`.

Proven with three transient FAIL/PASS pairs, including `AbstractAgent` (a class, which behaves differently from an enum) and `UseAgentUpdate` via the sneakier inline `type ` prefix rather than the block form.

packages/react-native/src/__tests__/headless-entry-surface.test.ts
```diff
@@ -38,62 +80,257 @@ const FORBIDDEN_LOCAL = [
...
   while (stack.length) {
     const file = stack.pop()!;
     if (seen.has(file)) continue;
     seen.add(file);
 
-    const code = fs.readFileSync(file, "utf8");
-    for (const m of code.matchAll(importRe)) {
-      const spec = m[1] ?? m[2];
-      if (!spec) continue;
+    const code = stripComments(fs.readFileSync(file, "utf8"));
```
mxmzb, line 224: **This #4893 guard did not guard, and the PR body claims it carries mutation evidence.** All four defects were reproduced by introducing the violation and watching the guard **pass**:

1. The import regex was blind to dynamic `import()` and `require()`, so a lazy optional-peer `require("@copilotkit/react-core/v2")` defeated it entirely.
2. It matched imports inside comments — and was **already harvesting 8 phantom specifiers, exactly half its entire bare-specifier set, purely from JSDoc examples.** It also *rejected* a legitimate "don't do this" counter-example written in a comment, so documenting the hazard could fail the build.
3. `resolveLocal` silently dropped subgraphs it could not resolve, so an unresolvable edge read as clean — and a real `/v2` re-export hid behind an ESM-style `./foo.js` edge.
4. The existence test was dead: `walkGraph` threw at collection, so it reported **`Tests no tests`** — all 11 assertions silently gone, not one failure.

Now strips comments before matching, extracts `import()` / `require()` / `require.resolve()` and flags non-literal loader arguments, resolves emitted extensions, fails loudly on anything still unresolvable, and asserts the resolved graph exactly.

packages/react-native/src/__tests__/headless-entry-surface.test.ts
```diff
@@ -114,38 +352,132 @@ describe("@copilotkit/react-native/headless entry", () => {
...
+  const IMPORT_BUDGET_MS = 60_000;
+
+  // Nested so the hook's failure domain covers ONLY the tests that need the
+  // module. Hoisting it to the top-level `describe` would let an import failure
+  // take down the fs-only graph tests above — the same collection-time
+  // blast-radius problem as blind spot #4, just moved into a hook.
+  describe("runtime export surface", () => {
+    let mod: typeof HeadlessEntry;
+
+    beforeAll(async () => {
```
mxmzb, line 398: **Flake fix, with the cost measured rather than the timeout raised blindly.**

The `await import("../headless")` took **4568 ms against a 5000 ms budget on an idle machine** — 91% of the allowance before any load — and **four tests each awaited that same one cost**, which is why the signature was three *simultaneous* timeouts rather than one slow test.

Root cause: Vite transforming ~283 kB of workspace dist that `vitest.config.mjs` inlines via `server.deps.inline: [/@copilotkit/]`, per worker, with no persistent cache. A bare-Node import of the same prebuilt dist is **461 ms**, so evaluation is cheap — it is resolve/transform plus cold page-cache I/O. The **variance** broke the budget, not the mean.

Two deliberate constraints: the import could not be made static, because these four checks assert exports exist as **runtime values** (a static read would weaken the guard); and the hook is **nested** on purpose, since a top-level one would drag the fs-only graph tests into the import's failure domain — recreating defect 4 above inside a hook.

Acceptance: **5/5 green at default parallelism** with no `--maxWorkers` / `--testTimeout` overrides, plus 5/5 under load average 180–235 (including one run at 26.96 s wall — the exact failing-baseline profile) and three concurrent full suites green. Every test in the file now reports 0–4 ms.

mxmzb:
## Why the private registry existed in the first place — the archaeology

Since this PR deletes `hooks/RenderToolContext.tsx`, it's worth recording *why* it was there, because the answer changes how the deletion should be argued — and it slightly corrects a sentence in the description.

**Short version: the original justification was true on the day it was written, stopped being true two and a half months later, and nobody revisited it.**

### Origin

```
24db94492b  2026-05-22  Jordan Ritter <jordan@copilotkit.ai>
feat(react-native): add useRenderTool hook and RenderToolContext
```

The commit body says only: *"Hook for rendering custom tool UIs in React Native with a store-based context provider. Includes tests for register/unregister, subscriber notification, and error handling."* No rationale — it documents **what**, never **why**.

The reasoning does exist, but it was recorded in a different file. At that same commit, `packages/react-native/src/index.ts:96` read:

```ts
// useRenderToolCall — web-specific (depends on DOM elements via DefaultToolCallRenderer)
```

### On 2026-05-22 that was a defensible call

Two facts made forking the registry the cheaper option that day:

1. **The hook really did reach DOM.** At `24db94492b`, `react-core/src/v2/hooks/use-render-tool-call.tsx:9` imported `DefaultToolCallRenderer` from `./use-default-render-tool`, and *rendered* it at line 201 as the fallback. That component uses `<div>` / `<svg>`, so a naive RN import pulled host-DOM elements.
2. **It was only reachable through the fat entry.** `/v2/headless` existed on that date but did **not** export `useRenderToolCall`. The only export sites were `src/hooks/index.ts` and `src/v2/hooks/index.ts` — i.e. the heavy `/v2` barrel. Importing it into RN meant dragging the whole `/v2` graph into a Metro bundle that doesn't tree-shake: exactly the hazard of #4893.

So at creation, consuming core's hook meant either shipping DOM elements or shipping the fat entry. A 150-line local `Map` avoided both.

### What changed, and what didn't

- **2026-07-23, #5883** created the lean `/v2/headless` entry and moved `useRenderToolCall` into it. Blocker (2) disappeared: the hook became importable without the fat graph.
- Blocker (1) had also softened — the DOM dependency is a *fallback branch inside* the hook, not a property of its contract, and once RN supplies its own renderers that branch isn't taken.
- **The exclusion comment was rewritten the next day** — the moment someone was closest to the question — and its stated reason was carried forward verbatim rather than re-tested against the new export surface.

That's the actual failure mode, and it isn't "someone made a bad decision". It's a **correct decision whose expiry condition was never written down.** The comment recorded a true observation about a specific import graph on a specific date; what it didn't record was *"revisit this if the hook moves to a lean entry."* So when precisely that happened, nothing connected the two.

### Why the drift then cost real money

The fork wasn't merely redundant — it froze RN's contract at May's shape while core's grew. `RenderToolContext.tsx` as created declared:

```ts
export interface RenderToolProps<T = Record<string, unknown>> {
  args: T;                              // never partial — promised complete args mid-stream
  status: "executing" | "complete";     // no "inProgress" state at all
  result?: string;
}                                       // no name, no toolCallId
```

Which is how you end up with renderers that can't represent a streaming call, and the three bugs this PR fixes.

And per the data point already in the description, the fork never bought a byte: Metro doesn't tree-shake, so the hook was already inside the chunk every RN app resolves whole.

### One correction to the description

The body currently says the exclusion was made *"on the stated grounds that it 'depends on DOM elements via `DefaultToolCallRenderer`' — a claim that was never true of the hook itself."*

That's a little unfair as written. At `24db94492b` the hook **did** import and render `DefaultToolCallRenderer` (line 9, line 201). The accurate framing is:

> the DOM dependency was a *fallback branch* rather than the hook's contract, and the real blocker was that the hook was reachable only through the fat `/v2` entry — which #5883 removed on 2026-07-23, at which point the exclusion should have been retired.

Same conclusion, but it survives someone checking the May tree — which matters, because the whole argument of this PR is that stale justifications get trusted.

BenTaylorDev (commented):
Reviewed against current `main` (`54c32a1945`). Merges cleanly, CI green across the board. The core of this is right: I traced the three bugs and the convergence holds up. Two small things I'd want changed before it lands, and a few notes.

## What I verified independently

- **Streaming (bug 1) genuinely fixed.** I probed a real AG-UI run to confirm the mechanism rather than trust the memo: every `TOOL_CALL_CHUNK` apply produces a **new** messages array, message, `toolCall` and `function` object. So `ToolCallRenderer`'s comparator (`use-render-tool-call.tsx:78-97`) sees the changed `arguments` string, re-renders, and `partialJSONParse` yields growing partial args. Worth stating explicitly because the opposite — in-place mutation of the same `toolCall` — would have silently defeated that comparator and left the headline fix inert.
- **History retention (bug 3):** `useFrontendTool`'s cleanup removes the tool but deliberately not the renderer (`use-frontend-tool.tsx:38-41`). Confirmed.
- **`useComponent` is safe on RN:** it's a pure passthrough (`use-component.tsx:83-86`), no DOM wrapper, so pointing RN at core's registry can't crash it.
- **The #4893 purity guard actually guards.** Ran `assert-headless-purity.mjs` against a real built dist (clean, 649 modules), then appended `import "streamdown"` to `dist/v2/headless.mjs` — it failed and named shiki/mermaid/cytoscape/katex/streamdown. Mutation evidence reproduced independently.
- Suites: RN **287 passing / 23 files**, both `node --test` script suites green. Removed coverage (malformed JSON, empty args, executing status) is re-covered against the real registry and expanded. No references to `useRenderToolRegistry` / `RenderToolProvider` remain anywhere in the repo.

Things I tried to break and couldn't: render-only tools losing `complete` (core splices a tool message even with no handler, `run-handler.ts:839-876`); `export { AbstractAgent }` adding runtime weight (`@ag-ui/client` is already fully in the graph via `@copilotkit/core`); the `esbuild` manifest/lock specifier mismatch (explained by the root `pnpm.overrides` entry `"esbuild": ">=0.25.4"`).

## Two changes I'd ask for

**1. `messagesFingerprint` is a copy of react-core's *pre-#6325* memo key.**

`CopilotChat.tsx:88-96` uses `contentKey = 0` for object content. #6325 (`de0a659b2d`) landed on `main` and changed exactly that, serializing object content:

```ts
: m.content && typeof m.content === "object"
  ? JSON.stringify(m.content)
  : 0;
```

So the comment's claim that this "mirror[s] react-core's web `messagesMemoKey`" isn't true as of now. It is **not** a live RN bug — activity messages carry `role: "activity"` and RN's list builder only handles `user`/`assistant`, so they never reach a renderer. But it re-opens a divergence in the PR whose thesis is closing them, and it's a one-ternary fix.

**2. The `BREAKING CHANGE:` footer on `db67ccf` is incomplete, and release notes derive from it.**

It covers the two removed exports but not the `RenderToolProps` shape change. Existing RN renderers that read `args.foo` without first narrowing on `status` now fail `check-types` — verified, `TS18048`, because the `inProgress` arm makes `args` `Partial<T>`. That's the break most consumers will actually hit, and it should be in the footer.

For the record, I checked the other direction too: `status === "executing"` against the enum union still compiles **and narrows**, so raw-string comparisons in existing renderers are not a break. The docs' "compare against the enum" guidance is good practice, not a requirement.

## Notes, not blockers

- **One justification comment is measurably wrong.** `CopilotChat.tsx:75-86` (and the matching inline comment) says "nothing in production ever reassigns `.messages` — grepping for that assignment in `packages/core/src` hits test files only." The reassignment lives in `@ag-ui/client`'s apply pipeline, outside that grep's scope; the array identity does change on every delta. The fingerprint is still justified — core's `splice` of tool results and `addMessage` pushes are genuinely in-place — but in a PR this reliant on its comments being exact, the stated reason should be corrected rather than left as the load-bearing rationale for a perf-sensitive design.
- **Two deliberate tripwires** worth accepting knowingly: the exact-graph snapshot (`headless-entry-surface.test.ts:303-329`) will fail any future PR that adds a module or bare import to RN's headless graph, and the purity guard hard-fails on *any* non-literal `import()`/`require()` in first-party graph files — which includes `@copilotkit/core` and `@copilotkit/shared` built output. Both are defensible; both put a red check on someone else's unrelated PR when they trip.
- The removals are a genuine break in a `1.x` package. I have no objection to the "pull the plaster off" call — flagging it as Tyler's to ratify rather than mine.

mxmzb:
@BenTaylorDev — thanks, this was a genuinely useful review. Both requested changes have landed, plus three of your notes. **Two of them I handled differently from what you asked**, and one of those overrides your judgement, so I want those visible rather than buried in a diff.

Nine commits since you reviewed (`b4cfcf6f98` → `4b17ea7d35`).

---

### 1. Memo key is a pre-#6325 copy — fixed (`4c17a8fe8c`)

Aligned the object-content branch with react-core's current `messagesMemoKey`, and corrected the "mirrors react-core" comment so it no longer claims to track a moving target.

Two notes on your analysis:

- **Your live-bug verdict was right, the reason is one path narrower than stated.** `listItems` does only handle `user`/`assistant`, so `role: "activity"` never reaches an RN renderer — confirmed. But object content *does* arrive via `role: "tool"` through `toolResultContent`, which serializes objects by design. That made a genuine behavioural red-green pair possible, so there's now a test that fails against the old key rather than the tautological pin we'd have settled for.
- **#6325 is on `main` but not an ancestor of this branch** — we're based on `b32b5539cc` (2026-08-07). So the divergence only exists when measured against the merge target, which is exactly why you saw it and the branch's own typecheck didn't.

### 2. `BREAKING CHANGE:` footer — I did something else. Please sanction or push back.

Your ask was right; your premise turned out not to be. Checking *"release notes derive from it"* first:

- a1's footer on `ec42161670` **already covers** the `Partial<T>` shape change (added while deleting the changeset).
- But `scripts/release/lib/changes.ts:43` collects with `--format=%H %s`, and the `Commit` type is literally `{ hash: string; subject: string }` — there is no field a body could live in. `grep -rn "BREAKING" scripts/release/` is empty.

So **no footer reaches release notes at all**, repo-wide. Three more things fell out of that trace:

- Version selection is a **manual `--bump patch|minor|major` flag**. Nothing parses `!` or `BREAKING CHANGE:` for severity, so nothing enforces that a `!` commit gets a major.
- `generateRawReleaseNotes` buckets on `/^feat[:(]/` and `/^fix[:(]/` only — a breaking `refactor!` currently renders under **"Other Changes"**.
- Per-package `CHANGELOG.md` files are orphaned. Nothing in `scripts/release/` writes them; `packages/react-core/CHANGELOG.md` is titled `# ui` and its newest entry is **1.55.2** with Changeset-era ids, while the package is 1.66.x.

**What I did instead:** put the `TS18048` break in `showcase/shell-docs/.../useRenderTool.mdx` (`9eacbfdec3`) — the destination consumers actually read — and filed #6479 for the collector, since it affects every breaking change in the repo rather than just this PR. For this PR we're relying on the **subject** as the release-notes channel; it already carries `!` and at 87 chars sits inside the 120-char `header-max-length`.

That satisfies the intent (users learn about the break) but not the letter of your request. **If you'd rather the footer be completed anyway for `git log` readers, say so and I'll do it.**

### 3. The measurably-wrong justification comment — fixed (`00caf5fa7b`)

It was at **three** sites, not one. Your citation was the missing piece: `@ag-ui/client`'s `AbstractAgent.processApplyEvents` reassigns `this.messages` per applied delta, outside the `packages/core/src` grep's scope.

The comment now says identity is unreliable **in both directions** — it changes on the AG-UI apply path, and doesn't on core's `splice` (`run-handler.ts:931`, `:1080`) or `addMessage`'s `push` — rather than trading one over-general claim for another. I also edited the review comment I'd posted repeating the original error, with a visible correction note rather than a silent rewrite.

### 4. The two tripwires — I changed both instead of accepting them

You called both defensible if accepted knowingly. We went further, and on the first one that's me overriding your call:

- **Exact-graph snapshot → manifest-derived allow-list** (`c7d176f264`). The allow-list is *computed* from `dependencies` + non-optional `peerDependencies` — precisely what the headless entry promises a consumer can resolve — so it can't drift from the manifest. It tolerates new first-party modules but still fails on any undeclared third-party edge, every optional peer, a devDependency-only import, a doc-comment phantom, and a truncated walk. My reasoning: a guard that fails innocent changes gets disabled by the third person who hits it. **Happy to restore the exact pin if you prefer the stricter contract** — it's your flag.
- **Purity guard's loader detection → tokenizer** (`4b17ea7d35`). Worth knowing it was wrong in *both* directions, not just strict: `throw new Error("use require(path) instead")` hard-failed CI, while `` import(`stream${n}`) `` and `import("zo" + n)` read as clean because the literal test only checked the first character after `(`. Also `__require(...)`, rolldown's emitted form, matched nothing. Nine shapes closed, each reproduced against the real gate first and then proved in both directions. Added the first `.cjs` fixtures — the `format: "cjs"` branch had no coverage at all.

  Both gates also had a worse problem we found separately: `isMain` compared a `file://`-concatenated `argv[1]` against percent-encoded `import.meta.url`, so on a path with a space or via a symlink **both exited 0 having asserted nothing**. Now `isEntrypoint()` with `fileURLToPath` + `realpathSync`, with an end-to-end spawn test through a symlinked, space-bearing path (`1b39c12e36`).

### 5. Removals are a real 1.x break — Tyler's to ratify

Unchanged and ours to chase. PR is still a draft.

---

### One correction of mine that came out of your review

Your line — *"`status === "executing"` against the enum union still compiles **and narrows**, so raw-string comparisons in existing renderers are not a break"* — is why we retracted the migration note claiming the opposite. That claim was mine, from a probe that tested *assignment* and comparison against an *invalid* literal but never the comparison the docs actually recommend. The PR description now carries that correction inline rather than quietly dropping the sentence.

### Caveats

- The Windows `isEntrypoint()` case is **covered by construction, not executed** — darwin only.
- `dev-docs/bundle-size.md` now lists what the gate genuinely does **not** catch instead of implying completeness. Three remain: the regex-vs-division heuristic is bounded to one line, no JSX/TS parsing, and indirect loaders (`const f = require; f(x)`).
- Numbers at `4b17ea7d35`: RN **289/289** across 23 files at default parallelism (the 5s dynamic-import flake is fixed — root-caused to a 4568 ms import against a 5000 ms budget with four tests racing the same cost), react-core `test:scripts` **47/0**, RN `test:scripts` **17/0**, both typechecks clean, source format clean, `oxlint` 0 errors.

### Deferred, with findings recorded

- **RN's `useRenderTool` should be deleted, not patched.** It's an alias for `useFrontendTool` wearing core's `useRenderTool`'s name — which is why `"*"` registers a tool literally named `*`, why `description`/`parameters` are required, and why `followUp`/`available`/handler-context are dropped. The history is stark: core's renderer-only `useRenderTool` has existed since 2026-03-28, and `useFrontendTool` has carried `render` **and** `followUp` since 2025-10-03 (#2566) — so RN's hook was redundant the day it was written (#4715, 2026-05-22), where it appears as the sixth bullet of a UI-components PR. The docs now say the wildcard doesn't work and § Known limitations tracks it; the convergence gets its own PR with a migration note.
- **RN's `sideEffects` globs are dist-relative**, so our own build strips `import "./polyfills"` from all four built entries — verified zero `polyfills` references in `dist/{index,headless}.{mjs,cjs}`. Pre-existing (globs since 2026-05-06, entries since 2026-05-13/07-23), so out of scope here, but it means published consumers get no auto-installed polyfills.
- **react-core's `messagesMemoKey` has the same unguarded `JSON.stringify`** in a render path that we guarded on the RN side, so circular activity content will throw there.
- `CopilotChatConfigurationProvider` is neither mounted nor exported on RN, so `agentId` still resolves to `"default"` and consumers can't fix it themselves — the PR documents the resolution limit but not the un-fixability.

Happy to act on any of the pushback, particularly on #2 and the allow-list.

BenTaylorDev (approved):
Re-reviewed at `4b17ea7d35` against current `main` (`f13fcb09e9` — 117 commits ahead of this branch's base). Merges clean, CI green. Both changes I asked for are resolved, and on one of them **my premise was wrong and yours was right**. Approving the code; the one thing I'm still not ratifying is below.

## The two asks

**1. Memo key — fixed, and better than what I asked for.** `messagesFingerprint`'s object branch (`CopilotChat.tsx:131-136`) now matches react-core's `messagesMemoKey` (`CopilotChat.tsx:989-1007`). The `try/catch` around `JSON.stringify` is a divergence react-core doesn't have, and it's the right direction — react-core will throw on circular content in a render path, which is worth the separate issue you noted.

The new test is a genuine red-green: same id, role and `toolCallId`, and the two objects serialize to the same *length*, so a key measuring type or size alone still misses it. That's the assertion I'd have wanted and didn't ask for.

**2. `BREAKING CHANGE:` footer — you were right, I was wrong.** I said release notes derive from the footer. They don't, and I checked rather than take it:

- `scripts/release/lib/changes.ts` collects `git log --format=%H %s`; `Commit` is `{ hash, subject }` — there is no field a body could occupy.
- `grep -rn "BREAKING" scripts/release/` is empty.
- `prepare-release.ts:41-42` buckets on `/^feat[:(]/` and `/^fix[:(]/` only, so a breaking `refactor!` renders under "Other Changes" with its subject verbatim.

So the footer was never the channel I claimed it was, and putting the `TS18048` break in `useRenderTool.mdx` is the better destination. `ec42161670`'s footer already covers the `Partial<T>` shape change in full, so `git log` readers are served too. #6479 is the right home for the collector gap. Nothing further needed here.

## What I checked in the delta rather than took on trust

- **The allow-list relaxation does not weaken #4893.** My concern was that computing the allow-list from `dependencies` lets `@copilotkit/react-core/v2` through, since react-core is a legitimate dependency. It does — and the fat-entry ban is a *separate*, package-wide test (`ALLOWED_REACT_CORE_ENTRIES`, `it.each(ENTRIES)`) that survives untouched, as does the `FORBIDDEN_HEAVY` deny-list. Your call over mine here was correct: the pin failed on innocent first-party growth, the replacement fails on undeclared third-party edges, and the guard that actually encodes #4893 never depended on the pin.
- **The wildcard retraction is inherited, not introduced.** I traced it to `main` before accepting the framing: `main`'s RN `useRenderTool` already pipes `name` straight into `useFrontendTool` (`useRenderTool.ts:71-82`), and the old `RenderToolContext` had no `"*"` special-case either. So `name: "*"` advertised a tool literally named `*` before this PR too. The retraction corrects the description, it doesn't confess a regression.
- **Numbers reproduce.** RN vitest **289 passing / 23 files**, react-core `test:scripts` **47/0**, RN `test:scripts` **17/0** (the second needed the new `esbuild` devDependency, which is real, not a wiring problem).
- **The `sideEffects` finding is real.** Built `dist/headless.mjs` is 918 bytes of re-exports with zero polyfill references; the payload sits unimported in `dist/polyfills/*`. Pre-existing and correctly out of scope — but worth knowing that `REQUIRED_HEADLESS_MODULES`' `polyfills.ts` entry pins a *source*-graph edge the build then strips, so it proves slightly less than it reads like.

## One thing to fix, not blocking

The tokenizer still hard-fails on a comment sitting next to an otherwise-static loader argument — the same false-positive class `4b17ea7d35` set out to close. `isCompleteLiteralArgument` runs its surround check against `code` rather than `masked`, so the comment text survives the `.trim()`:

```
"magic comment before literal" -> ["import(/* @vite-ignore */ \"./a.js\");"]
"comment after literal"        -> ["require(\"a\" /* side */);"]
```

Nothing in react-core's graph trips it today, but `@vite-ignore` is already idiomatic in this repo (`packages/vue/src/v2/hooks/use-katex-styles.ts:38`), so it's a live shape. Switching those two slices to `masked` fixes it and keeps concatenation and interpolation caught, since `masked` preserves length and blanks only comments and literals.

## Still not mine to ratify

Removing `useRenderToolRegistry` and `RenderToolProvider` — two documented exports — in a minor of a 1.x package is a product call, and my approval isn't it. Please get Tyler's explicit sign-off before merging. Everything technical that would make that decision risky has been closed out; what's left is whether we want to pull the plaster off now.

Two bits of housekeeping when this lands: #6346 (@davidmckayv) is still open and should be closed with a pointer here — the `Co-Authored-By` credit is already in place. And your reply says the PR is still a draft, but GitHub has it marked ready; worth reconciling.

PR #6439, opened 2026-08-08
fix(core): preserve streamed messages omitted by snapshots

### LINKED ISSUE #6301: [v2] Message view freezes mid-run in long multi-tool runs (LangGraphAgent); state updates flow but UI stops re-rendering
## Environment

- `@copilotkit/react-core` / `react-ui` / `runtime`: **1.64.2** (all v2 imports)
- Agent: `LangGraphAgent` (LangGraph dev server, `stream_mode=['events','values','updates','messages-tuple']`)
- LLM: DeepSeek (reasoning model) via OpenAI-compatible API
- Next.js 16 + React 19, production build

## Symptom

In **long multi-tool runs** (e.g. query tool → long reasoning → file-generation tool → final answer), the chat view **stops re-rendering mid-run** while the run is still executing:

- The last visible content is typically the assistant text right before a tool call.
- The run **completes successfully** server-side (verified in traces and server logs).
- The client **keeps receiving and processing events** — we instrumented `agent.subscribe({ onEvent })` and events (`STATE_SNAPSHOT`, `RAW`, …) keep flowing after the UI freezes.
- No console errors.
- After the freeze, **nothing** brings the content back — waiting past run completion does not help. Only starting a *new* run flushes the accumulated content into view.

## What we verified (elimination)

- Server SSE stream is complete and well-formed (replayed the run: all ~2000 events valid JSON, `RUN_FINISHED` present).
- Component renderers are fine in isolation (SSR smoke tests with the exact tool payloads).
- Not throttling-related: we don't set `throttleMs`/`defaultThrottleMs`, so the unthrottled path is used.
- A "reconcile watchdog" subscriber that returns `{ messages: [...messages] }` (fresh array reference) on `onRunFinishedEvent` (outcome `success`) / `onRunFailed` / `onRunErrorEvent` **does not** unfreeze the view — suggesting the blockage is deeper than the `MemoizedAssistantMessage` reference comparison.
- Two empirical red lines we hit while experimenting (both reproduce reliably):
  1. Returning a `messages` mutation from a subscriber **while the run is paused on an interrupt** breaks interrupt UI (the choice card from `useInterrupt` never renders).
  2. Returning periodic `messages` mutations **during** an active run corrupts in-flight streaming (the view freezes almost immediately).

## Repro characteristics

- Intermittent for short runs; **near-deterministic for long runs** (multiple tool calls + 20–60s reasoning + a tool whose arguments are a large JSON payload, ~10–20 KB streamed as `TOOL_CALL_ARGS` deltas).
- Freeze point in our logs consistently lands right after a `generate_file`-like tool call's result, or right after an assistant message that carries **both** visible text and tool calls.
- Our integration details that may matter: assistant messages from DeepSeek carry `content` as an array of parts (`reasoning_content` + `text`); heavy `STATE_SNAPSHOT` traffic (500+ per run) from LangGraph `values` stream mode.

## Happy to provide

- Full client-side event log (`window.__evts` probe) vs. server SSE comparison for a frozen run
- Langfuse trace showing the run completing successfully
- A minimal repro if you can point us at a reference LangGraph + v2 setup

## Ask

1. Is this a known class of issue in v2's message reconciliation (memoized message view vs. in-place state mutation)?
2. Is there a supported way to force the message view to re-sync from `agent.messages` after a run? (Our subscriber mutation does not achieve this.)
3. Guidance on the two red lines above — are subscriber `messages` mutations during interrupts / mid-run supposed to be safe?

scarab-systems:
## What does this PR do?

Fixes a v2 core reconciliation edge case where a late authoritative snapshot can omit assistant/tool messages that were already streamed during the run. When that happens, `runAgentResult.newMessages` can be empty even though the streamed messages contained frontend tool calls, so CopilotKit never executes those tool calls and the UI appears to stall.

This change records the messages observed through `onMessagesChanged` during a run, compares them with the final agent message state, and restores only messages created during this run that were observed in the stream but dropped from final state. The reconciliation also appends only those missing run messages to `newMessages`, so existing/final messages are not broadly replayed through tool execution.

Added a regression test that simulates the reported late truncated snapshot behavior and proves the streamed assistant tool-call message still reaches the frontend tool execution path.

## Related PRs and Issues

- Fixes https://github.com/CopilotKit/CopilotKit/issues/6301

## Validation

- `./node_modules/.bin/vitest run packages/core/src/__tests__/core-run-agent-message-reconciliation.test.ts`
- Fail-first check: temporarily removed the production reconciliation change and confirmed the new regression test fails because the frontend tool handler is not called.
- `./node_modules/.bin/vitest run packages/core/src/__tests__`
- `./node_modules/.bin/tsc -p packages/core/tsconfig.json --noEmit`
- `./node_modules/.bin/oxlint packages/core/src/core/run-handler.ts packages/core/src/__tests__/core-run-agent-message-reconciliation.test.ts`
- `./node_modules/.bin/oxfmt --check packages/core/src/core/run-handler.ts packages/core/src/__tests__/core-run-agent-message-reconciliation.test.ts`
- `NODE_OPTIONS=--max-old-space-size=8192 ./node_modules/.bin/tsdown` from `packages/core`

Additional local hook validation completed after aligning the local toolchain with the repo config:
- `pnpm exec lefthook validate`
- `pnpm exec lefthook run commit-msg <temp commit message file> --no-tty`
- `pnpm exec lefthook run pre-commit --file packages/core/src/core/run-handler.ts --file packages/core/src/__tests__/core-run-agent-message-reconciliation.test.ts --force --no-tty`

The pre-commit hook completed end-to-end, including the affected `test,publint,attw` matrix for 21 projects.

## Checklist

- [x] If the PR changes or adds functionality, I have updated the relevant documentation (not applicable; no docs/API behavior change)

No docs or changeset were added; this is a narrow bug fix covered by a regression test.

Disclosure: I used AI-assisted coding tools while preparing this PR. I reviewed the final diff and validation results, and I take responsibility for the submitted change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where streamed assistant tool calls could disappear when a later agent snapshot omitted them.
  * Preserved the correct message order and ensured eligible frontend tools continue executing.
  * Prevented stale streamed messages from being restored after switching threads, pruning drafts, completing tool calls, or when no thread identity is available.


NathanTarbert:
Hey @scarab-systems,

Nice diagnosis — the snapshot-truncation mechanism you described is real. I wrote a test against a live agent emitting an actual event stream with a truncating `MESSAGES_SNAPSHOT`: without your fix `newMessages` comes back empty and the frontend tool never fires; with it, the handler runs exactly once and there are no duplicate ids across turns. Your three tests are load-bearing too — I mutation-tested each one.

One thing to fix before this can land. The `missingRunMessages` filter can't distinguish a buggy truncation from an intentional removal, so it restores messages a backend pruned on purpose. Concretely: an agent that streams a draft and then snapshots `[user, summary]` — a summarization node, or LangGraph `RemoveMessage` — ends up with the pruned draft back in state and in `newMessages`. If that message carries a frontend tool call, the tool executes. That's going to bite as compaction lands.

Suggest narrowing restoration to what the issue actually reports: assistant messages whose `toolCalls` reference a registered frontend tool and have no corresponding tool result in final state. A test asserting a deliberately pruned plain-text message stays pruned would lock it in.

Smaller one: the thread guard uses `originatingThreadId !== undefined`, so it's inert when no thread identity exists — worth failing closed instead. And a comment pointing at the upstream `@ag-ui/client` snapshot behavior would help whoever revisits this after that changes.

Also worth adding a test that drives a real `AbstractAgent` rather than the mock — happy to hand you the ~40 lines I wrote.

scarab-systems:
Thanks, this was very helpful. I tightened the reconciliation in `81c31c19c` so it now restores only omitted assistant messages whose tool calls map to a registered frontend tool handler and do not already have a final tool result.

That means deliberately pruned plain-text/backend messages stay pruned, finalized tool calls are not re-executed, and the missing-thread-identity path fails closed. I also moved the scripted fixture onto `AbstractAgent` and added regression coverage for the pruned text and existing-result cases.

Local validation:
- `pnpm exec nx run @copilotkit/core:test -- src/__tests__/core-run-agent-message-reconciliation.test.ts`
- `pnpm exec nx run @copilotkit/core:check-types`
- `pnpm exec nx run @copilotkit/core:test`
- `pnpm exec nx run @copilotkit/core:build`
- `pnpm exec oxlint packages/core/src/core/run-handler.ts packages/core/src/__tests__/core-run-agent-message-reconciliation.test.ts`
- `pnpm exec oxfmt --check packages/core/src/core/run-handler.ts packages/core/src/__tests__/core-run-agent-message-reconciliation.test.ts`

The pre-commit affected-package gate also passed on retry. Its first run hit a flaky `@copilotkit/vue:test`; rerunning that target directly passed (`100` files / `1074` tests), and Nx marked it flaky rather than a deterministic failure.

PR #6450, opened 2026-08-10
feat: make Intelligence thread restore reliable and recoverable

maxkorp:
## Summary

Implements the SDK and React portions of the reliable thread-restore V1 contract. This draft starts with the Phoenix transport seam and will be updated as the core state machine, recovery action, UI states, documentation, and compatibility coverage land.

## Why

A restored conversation must not be presented as ready until replay reaches an explicit terminal outcome. Failures need a typed, recoverable client state instead of a silent one-second best-effort window.

## Current impact

- preserves successful Phoenix join acknowledgement payloads
- exposes normal channel close as a shared, one-shot session signal
- removes the close binding by Phoenix ref during teardown

## Validation

- focused Phoenix observable tests
- repository pre-commit package test, publishability, and type-surface matrix

> Draft: core restore orchestration, React UI, docs, and cross-version verification are still in progress.

### DIFF
--- a/packages/core/src/__tests__/intelligence-agent.test.ts
+++ b/packages/core/src/__tests__/intelligence-agent.test.ts
@@ -437,6 +437,38 @@ describe("IntelligenceAgent", () => {
       expect(result.events).toContainEqual(finishedEvent);
     });
 
+    it("does not let a terminal event from an older run complete the current run", async () => {
+      const agent = createAgent();
+      const promise = collectEvents(agent);
+      let settled = false;
+      void promise.then(() => {
+        settled = true;
+      });
+      await waitForConnection(agent);
+
+      const channel = getChannel(agent)!;
+      channel.triggerJoin("ok");
+      channel.serverPush("ag_ui_event", {
+        type: EventType.RUN_FINISHED,
+        threadId: "thread-1",
+        runId: "run-old",
+      } as BaseEvent);
+
+      await flushAsyncWork();
+      expect(settled).toBe(false);
+
+      const currentRunFinished = {
+        type: EventType.RUN_FINISHED,
+        threadId: "thread-1",
+        runId: "run-1",
+      } as BaseEvent;
+      channel.serverPush("ag_ui_event", currentRunFinished);
+
+      const result = await promise;
+      expect(result.completed).toBe(true);
+      expect(result.events).toEqual([currentRunFinished]);
+    });
+
     it("errors the observable on RUN_ERROR", async () => {
       const agent = createAgent();
       const promise = collectEvents(agent);
@@ -447,6 +479,7 @@ describe("IntelligenceAgent", () => {
 
       const errorEvent = {
         type: EventType.RUN_ERROR,
+        runId: "run-1",
         message: "something went wrong",
       } as BaseEvent;
       channel.serverPush("ag_ui_event", errorEvent);
--- a/packages/core/src/__tests__/intelligence-agent.test.ts
+++ b/packages/core/src/__tests__/intelligence-agent.test.ts
@@ -1610,6 +1610,40 @@ describe("IntelligenceAgent", () => {
       expect(socket.disconnected).toBe(true);
     });
 
+    it("never applies the legacy stream_idle fallback after reliable restore is negotiated", async () => {
+      mockFetch.mockResolvedValueOnce(await jsonResponse(runtimeCredentials()));
+
+      const agent = createAgent();
+      setThreadIdForTest(agent, "thread-1");
+
+      let resolved = false;
+      const promise = agent.connectAgent({ runId: "run-1" }).then((result) => {
+        resolved = true;
+        return result;
+      });
+      await waitForConnection(agent);
+
+      const channel = getChannel(agent)!;
+      channel.triggerJoin("ok", {
+        restore: {
+          mode: "failure_reporting",
+          restoreAttemptId: "ra_aware",
+        },
+      });
+      channel.serverPush("stream_idle", {});
+      await new Promise((resolve) => setTimeout(resolve, 150));
+      await flushAsyncWork();
+
+      expect(resolved).toBe(false);
+      expect(agent.getThreadRestoreState()).toMatchObject({
+        status: "restoring",
+        restoreAttemptId: "ra_aware",
+      });
+
+      channel.serverPush("replay_complete", { latestEventId: "event-2" });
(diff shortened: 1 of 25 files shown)


BenTaylorDev:
Reviewed as a pair with CopilotKit/Intelligence#802. Findings below are from reading the branch source. Gateway-side findings are on the other PR; two of them are cross-referenced here because they surface through this PR's UI.

The wire contract between the two PRs is consistent — capability shape, join ack, `restore_progress`/`restore_failed` field names and casing, and all 8 failure codes match one-for-one between `THREAD_RESTORE_FAILURE_CODES` and the gateway's `@failure_codes`. Mixed-version negotiation is correct in both directions: old client → no capability → legacy; new client → old gateway → no ack → `recognizedRestoreAttemptId` stays `null` → legacy behavior. Nothing enforces the two enums stay in sync, though — no shared schema, no contract test spanning the repos.

## Blockers

### 1. Sending a message blanks the entire chat

`IntelligenceAgent.run()` calls `beginThreadRestore(input.threadId)` synchronously at invocation (`intelligence-agent.ts:314`), setting the shared state to `restoring`. `CopilotChatView.tsx` then does:

```ts
if (threadRestore && threadRestore.status !== "ready") {
  return BoundRestoreView;
}
```

so the transcript *and* the composer are replaced by the "Restoring conversation…" gate until the run-mode `replay_complete` arrives. `ProxiedCopilotRuntimeAgent` routes `run()` straight to the delegate (`agent.ts:415`), so this is the normal send path, not an edge case. Scope is Intelligence mode only — a non-Intelligence runtime has no delegate, so `isThreadRestoreAware` is false and the gate never engages.

Combined with Intelligence#802's run-barrier failure (run-mode restore hard-fails from the second message of every conversation), this isn't a transient flash — it's a permanent "This conversation couldn't be restored." from message 2 onward.

The React tests pass `threadRestore` as a literal prop, so no test binds a real `run()` to a mounted view and CI can't see this.

### 2. Any post-restore transport error turns into "This conversation couldn't be restored."

`run()`'s `catchError` funnels every non-`AgentThreadLockedError` through `toThreadRestoreError`, which only checks `isCurrentRestoreGeneration` — not whether the attempt already settled `ready`. After `markThreadRestoreReady`, the generation is unchanged and `restoreState.threadId` still matches, so the guard passes and a late error flips `ready` → `failed`.

For run mode `terminal$` is `threadCompleted$` only (`streamIdleCompletion$` is `EMPTY` outside connect), so `observeSocketHealth$` stays live for the whole run. A socket drop mid-stream — after a perfectly successful restore — nukes the transcript view. Given the gateway 502/OOM history, this is not a rare path. `connectRestoreAttempt$` has the same shape.

Suggest gating `toThreadRestoreError` on the attempt not having already reached `ready`, so run failures stay run failures.

## Low / nits

- **`handleReload` leaks an unhandled rejection.** `CopilotChatRestoreView.tsx:34` awaits `reloadConversation()` in `try/finally` with no `catch`, and the async `onClick` discards the returned promise. `forceFullRestore()` rejects when no thread is selected or the thread has no cached connect input, and `detachActiveRun()` can reject. Today the user just sees the button re-enable with no feedback — and under the blockers above it fires on every click.

- **The `onClose` teardown is dead code.** Verified against the installed versions: phoenix 1.8.4 `Channel.onClose` returns `undefined` (`priv/static/phoenix.mjs:277` — `this.on(...)` with no `return`), while `Channel.onError` at :284 does return a ref. `@types/phoenix@1.6.7:23` incorrectly declares `Channel.onClose(...): number`, which is why this typechecks. So in `phoenix-observable.ts` `typeof ref === "number"` is always false and `channel.off("phx_close", ref)` never runs.

  I'd push back on calling this a leak — the binding lives on `channel.bindings` and dies with the channel, and `close$` is constructed once per session, so it's one stray callback on an object being discarded. Two things do matter: the PR description claims it "removes the close binding by Phoenix ref during teardown," which isn't accurate; and the only thing keeping the dead cleanup harmless is the generation guard in `restoreClose$`. If a channel is ever reused across restore attempts — which `forceFullRestore` gestures toward — a stale close callback would otherwise resurface a `ThreadRestoreError` for a superseded attempt. Latent, not live. The underlying trap is the `@types` mismatch, not this code.

- **`retryable` is never surfaced.** `CopilotChatRestoreView` renders the Reload button unconditionally. Every gateway `restore_failure/3` call site currently passes `retryable: true`, so `retryAction: "none"` never ships — but when a non-retryable code does land, the UI will offer a recovery that cannot work.

- **`legacyReadyState` is memoized on `[agent]`**, so if `agent.threadId` changes without the agent identity changing it goes stale. Only affects non-restore-aware agents.

## Housekeeping

The body still reads "> Draft: core restore orchestration, React UI, docs, and cross-version verification are still in progress" while the PR is marked ready for review — and the React UI and docs are clearly in the diff now. Worth refreshing before reviewers land on it.

BenTaylorDev:
**Correction to finding #1 above.** I leaned on a claim about Intelligence#802 that I got wrong, and it changes the severity here.

I said the gateway's run-mode restore "hard-fails from the second message of every conversation," making this a permanent error screen rather than a flash. That was wrong — I'd assumed the browser reliably wins the race to join, without checking when the runner starts. It doesn't: `await runStart.startup` in `handlers/intelligence/run.ts:283` resolves after the runner's Phoenix channel join, so the runner is already attached to the gateway before credentials return, while the browser still has to build a fresh WebSocket. For a fast agent the run watermark is usually populated in time and the restore succeeds. Details and what survives are in the correction on Intelligence#802.

What that means for this PR:

- **Finding #1 stands, at reduced severity.** `run()` → `beginThreadRestore()` → `status: "restoring"` → `CopilotChatView` returns only `BoundRestoreView` is still exactly what the code does, so every send still replaces the transcript and composer with the restore gate. But in the common case that's a **visible flash on each send** until run-mode `replay_complete` lands, not a permanent failure state. It degrades to the permanent version only when the upstream race is lost.

  Still a regression worth fixing — the full-surface gate is the wrong response to a user sending a message — but "chat blanks briefly on every send" rather than "chat is bricked from message 2."

- **Finding #2 is unaffected.** Any post-restore transport error flipping `ready` → `failed` via `toThreadRestoreError` is independent of the gateway race and remains as posted.

- The nits are unaffected.

Apologies for the noise — better to correct it than to leave an inflated severity standing on a claim I hadn't verified.

maxkorp:
Closing as no longer relevant.

PR #6454, opened 2026-08-11
fix(channels-slack): three defects that silently broke Slack Block Kit

lukasmoschitz:
Closes OSS-794 (Slack half — the validation pass and the fixes it found).

Three defects in the Slack Block Kit catalog, each verified against a real workspace. **99 lines changed across three files.**

The reason these sat undetected matters more than their size: **a payload Slack refuses produces no error anywhere.** No log line, no exception, no failing test — the message simply never arrives, which is indistinguishable from a bot that had nothing to say. The renderer compounds it by design, dropping unknown nodes silently so one bad node cannot fail a whole message.

## 1. `container` was refused on every send

Its children serialized into `blocks`; Slack reads `child_blocks`.

## 2. Every menu, checkbox, radio group, overflow and confirm dialog was refused

The codec stamped `type` onto every catalog entry, including composition objects whose schema has none — Slack's option object is `{text, value}`, and the same holds for `confirm`, `option_group`, `conversation_filter`, `dispatch_action_config`, `slack_file`, `trigger` and `workflow`. An unknown field makes Slack reject the entire message, so the whole interactive surface was unusable through `Slack.Object.*`.

Measured against a live workspace: **1 of 26 block elements delivered before this fix, 23 after.**

Note the existing `native-catalog.test.ts` asserted the very assumption that was wrong — that every entry serializes its discriminator. It was green while the product was broken. It now asserts the corrected rule.

## 3. An image could not use a file already in the workspace

The required-field check demanded `image_url` unconditionally; Slack accepts `image_url` *or* `slack_file`. An image needs alt text plus either source now, and passing neither is still an error.

## Two catalog corrections

`file` leaves the authorable manifest. Slack: *"You can't add this block to app surfaces directly, but it will show up when retrieving messages that contain remote files."* The same sentence appears verbatim in `@slack/types`' own doc comment — two independent Slack sources. It is an inbound shape; offering it as a component meant offering something that can never succeed. Sending a file remains `thread.postFile()`.

`alert` stays out with its citation — *"Alert blocks are currently only supported in modals."* Verified rather than assumed: Slack's own example payload posted verbatim into a message is refused, while a plain section in the same delivery seconds later arrives.

Both exclusions now carry their source, so "missing" and "deliberately not authorable" stay distinguishable to the next reader.

## How these were found

By building a fixture for every entry in Slack's catalog — 19 authorable blocks, 26 elements, 15 composition objects — with the expected payload **transcribed from `docs.slack.dev` rather than captured from our serializer**, and delivering all 60 through a managed Channel into a real Slack workspace. **55 of 60 deliver.** The remainder need real workspace objects (an uploaded file, a published workflow) or, for `rich_text_input`, an `action_id` the typed surface will not let you set.

That corpus is a working instrument, not a deliverable, so it is deliberately **not** part of this PR — it would add ~1700 lines of fixtures to maintain for a 99-line change. It lives with the team and gets re-run when the catalog moves.

One methodological note from it, because it changed what we consider proof: the first live run passed entries that demonstrated nothing. A rich-text block with one unstyled run renders exactly like a plain section; a carousel with one card renders like a card; a container with one child renders like a card. All three were accepted and worthless as evidence — caught by a human looking at the output, not by the harness. Fixtures had to *exercise* each entry, and it was that change which surfaced defect 2.

## Known and recorded elsewhere, not addressed here

- `external_select` / `multi_external_select` deliver but can never populate — opening one makes Slack call an Options Load URL, and nothing in the SDK or Intelligence serves one.
- Modal `view_submission` reads values flat, so checkboxes, date pickers and user pickers silently lose theirs. The message path handles all of them.
- Nine of nineteen blocks cannot be authored without a cast; `input.element` and `video.title_url` are demanded by the codec's own required-field check and not declared on `SlackNativeProps`.
- A single refused block closes the delivery's packet path, so every later post in the same turn fails as collateral (same shape as OSS-699).
- `thread.postFile()` — the real file-sending path — is validated by nothing.

## Verification

`test`, `check-types` and `build` green across `channels-slack`, `channels`, `channels-intelligence` and `runtime`, both with and without the fixture corpus present. Every block, element and object was delivered into a live Slack workspace and reviewed by eye.

### DIFF
--- a/packages/channels-slack/src/native-catalog.test.ts
+++ b/packages/channels-slack/src/native-catalog.test.ts
@@ -6,6 +6,7 @@ import {
   SLACK_ELEMENT_MANIFEST,
   SLACK_NATIVE_MANIFEST,
   SLACK_OBJECT_MANIFEST,
+  SLACK_UNTYPED_OBJECTS,
 } from "./native-manifest.js";
 import { Slack } from "./native.js";
 
@@ -30,7 +31,16 @@ test("every Slack catalog entry serializes its fixed discriminator", () => {
       entry.type,
       requiredProps(entry.type),
     );
-    expect(serializeSlackNativeNode(node).type).toBe(entry.type);
+    const serialized = serializeSlackNativeNode(node);
+
+    // Composition objects Slack defines as plain structures must not carry a
+    // discriminator: an option is `{text, value}`, and a stray `type` on it is
+    // an unknown field that makes Slack refuse the whole message.
+    if (entry.kind === "object" && SLACK_UNTYPED_OBJECTS.has(entry.type)) {
+      expect(serialized).not.toHaveProperty("type");
+    } else {
+      expect(serialized.type).toBe(entry.type);
+    }
     expect(entry.source).toMatch(/^https:\/\/docs\.slack\.dev\//);
   }
 });
--- a/packages/channels-slack/src/native-catalog.test.ts
+++ b/packages/channels-slack/src/native-catalog.test.ts
@@ -95,3 +95,31 @@ function requiredProps(type: string): Record<string, unknown> {
   }
   return {};
 }
+
+test("an image accepts either an external URL or a Slack-hosted file", () => {
+  const external = createNativeNode("slack", "block", "image", {
+    image_url: "https://picsum.photos/400/300",
+    alt_text: "Latency chart",
+  });
+  const hosted = createNativeNode("slack", "block", "image", {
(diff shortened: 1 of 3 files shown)


lukasmoschitz:
Closed in favour of #6462 — same three commits, branch renamed so Linear links OSS-819 (the validation half) instead of the OSS-794 umbrella, which stays open for the OpenTag demonstration. GitHub closed this one on the rename rather than retargeting it.

PR #6458, opened 2026-08-12
docs: define public AEO surface contract

samjulien:
## Summary

- publish a single shared, versioned technical contract for website, docs, and docs MCP AEO surfaces
- publish the human policy through the existing shell-docs MDX pipeline at `/aeo`
- expose the machine-readable contract at `/.well-known/copilotkit-capabilities/v1.json`
- validate the contract with JSON Schema/Ajv plus narrow repository and CI cross-reference checks
- run the actual shell-doc behavior tests in CI and assign external website and Pathfinder gaps to named owners

## Why this matters

Answer engines and coding agents decide which source to trust from machine signals such as canonical hosts, stable URLs, response types, and consistent capability claims. When those signals disagree, CopilotKit can be classified incorrectly, cited from the wrong hostname, or skipped even when it is the right product.

This PR gives those public surfaces a versioned source of truth. It separates standards, community conventions, and CopilotKit-specific guarantees; records real endpoint paths and media types; and makes ownership explicit when behavior lives in another repository or service. That gives us a reliable base for improving agent discovery without pretending one repository can enforce every public surface.

The implementation deliberately uses the current docs architecture: `/aeo` is ordinary shell-docs MDX under `showcase/shell-docs/src/content/docs/`, not a bespoke page or the retired docs tree. Schema shape lives in JSON Schema, while the small TypeScript layer only checks relationships JSON Schema cannot express, such as whether referenced files and CI commands exist.

## Validation

- `pnpm nx run @copilotkit/showcase-scripts:validate-aeo-contract --skip-nx-cache`
- `pnpm nx run @copilotkit/showcase-scripts:test --skip-nx-cache -- __tests__/validate-aeo-contract.test.ts` (6 tests)
- `npm --prefix showcase/shell-docs test -- src/app/sitemap.test.ts src/app/llms.txt/route.test.ts src/app/llms-full.txt/route.test.ts 'src/app/llms-mdx/[[...slug]]/route.test.ts' src/app/well-known/copilotkit-capabilities/v1.json/route.test.ts src/lib/runtime-config.test.ts src/lib/__tests__/next-config-redirects.test.ts` (43 tests)
- `npm run typecheck` in `showcase/shell-docs`
- `npm run build` in `showcase/shell-docs`
- targeted `oxfmt`, `oxlint`, TypeScript, diff, and commit-hook checks

## External follow-ups

- CopilotKit/website must link the same policy and fix `/llms.txt` plus `/llms-full.txt`, which returned 200 `text/html` soft-404 pages during the production audit
- Pathfinder/docs MCP owners must define a machine-readable discovery surface; the current contract records `/sse` as the known transport without presenting transport availability as discovery

## Related

- PDX-317


showcase/shell-docs/src/app/aeo/page.tsx
```diff
@@ -0,0 +1,209 @@
...
+const jsonLd = {
+  "@context": "https://schema.org",
+  "@type": "TechArticle",
+  headline: "Public AEO surface contract",
+  url: publicAeoContract.policyUrl,
+  mainEntityOfPage: publicAeoContract.capabilitiesUrl,
+  version: String(publicAeoContract.schemaVersion),
+};
+
+export default function AeoPolicyPage() {
```
tylerslaton: [P1] Please put this policy through the existing docs content pipeline. Our repository guidance says CopilotKit documentation belongs under `src/content/`; this bespoke 209-line App Router page also creates custom rendering, metadata, testing, and sitemap plumbing. An `src/content/docs/aeo.mdx` page—with a small component only for sections that truly need to render from JSON—would be substantially smaller and inherit normal docs behavior.

showcase/scripts/validate-aeo-contract.ts
```diff
@@ -0,0 +1,398 @@
...
+      );
+      if (!stringArray(surface.enforcement.paths)) {
+        errors.push(
+          `${path}.enforcement.paths must be a non-empty string array`,
+        );
+      } else {
+        for (const relativePath of surface.enforcement.paths) {
+          if (
+            isAbsolute(relativePath) ||
+            !existsSync(join(repositoryRoot, relativePath))
```
tylerslaton: [P1] This only proves that a named file exists; it does not enforce the declared HTTP behavior. Several contract entries name `@copilotkit/showcase-scripts:test` even though their cited tests live under `showcase/shell-docs`, and the new CI step runs only this validator. A broken status, content type, or response can therefore merge with this check green. Please run the actual shell-docs behavioral tests in CI and have each contract entry name the command that executes its check.

showcase/scripts/validate-aeo-contract.ts
```diff
@@ -0,0 +1,398 @@
...
+    return JSON.parse(readFileSync(contractPath, "utf8")) as AeoSurfaceContract;
+  } catch (error) {
+    const detail = error instanceof Error ? error.message : String(error);
+    throw new Error(`Unable to parse ${contractPath}: ${detail}`, {
+      cause: error,
+    });
+  }
+}
+
+export function validateAeoSurfaceContract(
```
tylerslaton: [P2] Can we reduce this to JSON Schema/Ajv validation plus the few cross-reference/file checks that schema cannot express? This 398-line validator redeclares the constants, TypeScript shape, and field-by-field shape already represented by the 273-line contract, so most of the PR is validation code that can drift from its source. Ajv is already a dependency in this package.

showcase/shell-docs/src/app/aeo/page.tsx
```diff
@@ -0,0 +1,209 @@
...
+              {publicAeoContract.surfaces.map((surface) => (
+                <tr key={surface.id} className="border-b border-fd-border">
+                  <td className="p-3 align-top font-medium">{surface.id}</td>
+                  <td className="p-3 align-top">
+                    {
+                      publicAeoContract.canonicalHosts[
+                        surface.host as keyof typeof publicAeoContract.canonicalHosts
+                      ]
+                    }
+                    {surface.path}
```
tylerslaton: [P1] `surface.path` is not consistently a URL path, so concatenating it with the host publishes malformed endpoint text—for example `https://docs.copilotkit.aiall indexable HTML pages` and `https://mcp.copilotkit.aiservice discovery and MCP transport endpoints`. Please keep endpoint values as actual paths/templates (one per endpoint), or render prose-valued scopes without prefixing the host.

tylerslaton (changes requested):
The contract idea is sound, but this implementation is not yet clean or minimal. It adds bespoke docs rendering and a large custom validation layer, while the new CI check does not run the behavioral tests the contract claims enforce these surfaces.

The lean shape I’d aim for is: one policy in the existing docs content pipeline, the JSON endpoint, a small schema validator with narrow cross-reference checks, and one CI target that actually exercises the public behavior.

tylerslaton (approved):

PR #6459, opened 2026-08-12
ci: add manual AEO production checks

samjulien:
## Summary

- add an on-demand production check for the website and docs discovery surfaces defined by #6458
- derive the ten in-scope routes and media types from the public contract instead of maintaining a second monitoring manifest
- exercise those routes as four documented crawler user agents with a global concurrency cap of four
- validate status, content type, canonical host, robots/sitemaps, one sampled sitemap link, LLM index links, and soft-404 behavior
- retain failure evidence and provide a deliberate `exercise_alert` input for proving the `#oss-alerts` path

## Why this matters

AEO is a production property, not a one-time content change. A correct repository can still deploy a broken canonical, HTML fallback, stale sitemap, or inaccessible LLM index. Those failures happen at the top of the agent-led growth funnel: if agents cannot reliably discover and verify CopilotKit, downstream recommendation and activation work never gets a chance to perform.

This PR adds the smallest useful operating check for that risk. It is deliberately limited to PDX-340's website/docs scope. It does not monitor MCP, the CopilotKit capability document, raw Markdown, Open Graph, or JSON-LD. Existing deploy-parser utilities are reused where practical, requests run with bounded concurrency, and failures include the exact URL, crawler identity, observed status/type, and a bounded response excerpt.

The workflow is intentionally manual at first. We should not create a scheduled noisy alarm while the website LLM endpoints are known red, and we should not claim Slack ownership until a deliberate failure proves the secret and alert path. A small follow-up can add the schedule after one normal run is green and one `exercise_alert` run reaches `#oss-alerts`.

## Stacked dependency

- Depends on #6458; this PR is intentionally based on `codex/pdx-317-aeo-surface-contract`.

## Validation

- `pnpm nx run @copilotkit/showcase-scripts:validate-aeo-contract --skip-nx-cache`
- `pnpm nx run @copilotkit/showcase-scripts:test --skip-nx-cache -- __tests__/check-aeo-synthetics.test.ts __tests__/aeo-synthetics-wiring.test.ts __tests__/verify-deploy.drivers.test.ts` (102 tests)
- targeted `oxfmt` and `oxlint` checks
- `git diff --check` and commit hooks

## Live baseline (2026-08-12)

The narrowed production command fails with eight records: four crawler identities × two website gaps.

- `https://www.copilotkit.ai/llms.txt` returns HTTP 200 `text/html` with a noindex soft-404 instead of plain text
- `https://www.copilotkit.ai/llms-full.txt` returns the same soft-404

The remaining website/docs targets pass: both home canonicals, both robots files, both sitemaps and sampled links, and both docs LLM indexes. The current failures are why this PR ships manual-first rather than enabling a schedule.

## Status

PDX-340 remains In Progress until the website endpoints are fixed, a normal workflow run is green, the deliberate Slack alert reaches `#oss-alerts`, and a follow-up enables the agreed schedule.


.github/workflows/aeo_synthetics.yml
```diff
@@ -0,0 +1,84 @@
+name: "AEO: Production Synthetics"
+
+on:
+  schedule:
+    # Keep this in sync with syntheticMonitoring.cadence in the public
+    # contract. The checker validates targets and user agents from that
+    # contract; this workflow intentionally contains no host/route copy.
+    - cron: "0 */6 * * *"
```
tylerslaton: Blocking: the PR's own live baseline says this produces 16 known failures today. Enabling the cron now means `#oss-alerts` receives an actionable-looking page every six hours for already-known gaps, and PDX-340's “known-good baseline” acceptance criterion is not met. Keep this manual until the dependencies are deployed and an on-demand run is green, or limit the initial baseline to surfaces that are actually known-good and add the remaining targets when fixed.

.github/workflows/aeo_synthetics.yml
```diff
@@ -0,0 +1,84 @@
...
+
+concurrency:
+  group: aeo-production-synthetics
+  cancel-in-progress: false
+
+jobs:
+  check:
+    name: Check public crawler surfaces
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
```
tylerslaton, line 26: The current config makes 56 primary requests sequentially (14 targets × 4 UAs), each with a 15-second timeout. A hanging outage can therefore take 14 minutes before link sampling, but this job is killed at 10 minutes; that can prevent the evidence and alert steps from running for exactly the broad outage this monitor should catch. Please use small bounded concurrency or otherwise make the checker's worst-case budget fit under this limit.

.github/workflows/aeo_synthetics.yml
```diff
@@ -0,0 +1,84 @@
...
+      - name: "Alert #oss-alerts on failure"
+        if: failure() && env.SLACK_WEBHOOK != ''
+        uses: slackapi/slack-github-action@dcb1066f776dd043e64d0e8ba94ca15cc7e1875d # v4.0.0
+        with:
+          webhook: ${{ secrets.SLACK_WEBHOOK_OSS_ALERTS }}
+          webhook-type: incoming-webhook
+          payload: |
+            { "text": ${{ toJSON(format(':x: *AEO production synthetic failure*{0}<{1}/{2}/actions/runs/{3}|View run>{0}```{0}{4}{0}```', fromJSON('"\n"'), github.server_url, github.repository, github.run_id, env.summary)) }} }
+
+      - name: Warn when alerting is not configured
```
tylerslaton, line 91: `gh secret list --app actions --repo CopilotKit/CopilotKit` does not show `SLACK_WEBHOOK_OSS_ALERTS`. It could be org-scoped, but the PR status also says the alert path is unverified. The acceptance criterion is actual alerting, not a warning in the same failed run. Please confirm the secret is available and execute a deliberate `workflow_dispatch` failure that reaches `#oss-alerts` before enabling the scheduled run.

showcase/shared/aeo/public-surface-contract.v1.json
```diff
@@ -261,6 +261,148 @@
...
+        "assertions": ["non-empty", "no-soft-404"]
+      },
+      {
+        "surfaceId": "docs-capabilities-v1",
+        "path": "/.well-known/copilotkit-capabilities/v1.json",
+        "contentType": "application/json",
+        "assertions": ["non-empty", "no-soft-404", "required-contract-fields"]
+      },
+      {
+        "surfaceId": "docs-mcp-discovery",
```
tylerslaton: This is the clearest scope expansion: PDX-340 asks for scheduled checks of the website and docs, not the external MCP transport. The same expansion adds capability-schema, raw-Markdown, Open Graph, and JSON-LD checks, which drive much of the new streaming/parser/config machinery. Please keep this PR to the ticket's named website/docs surfaces and checks; MCP/discovery can be a separate monitor with its own explicit ownership and acceptance criteria.

tylerslaton (changes requested):
The intent is good, but this is not yet minimal or operationally ready for PDX-340. The ticket calls for a focused website/docs synthetic over the named crawler surfaces; this grows into a second broad surface-checking framework (including MCP transport, capability-schema validation, Open Graph, JSON-LD, and raw Markdown) while duplicating parsing and contract-validation logic that already exists in `verify-deploy.drivers.docs.ts` and `validate-aeo-contract.ts`.

Please narrow the monitor to the ticket's explicit scope, reuse the existing probe/validation primitives where practical, and resolve the operational blockers below before enabling the schedule.

showcase/scripts/check-aeo-synthetics.ts
```diff
@@ -0,0 +1,528 @@
...
+      "Mozilla/5.0 AppleWebKit/537.36 (KHTML, like Gecko; compatible; PerplexityBot/1.0; +https://perplexity.ai/perplexitybot)",
+  },
+  {
+    id: "googlebot",
+    value:
+      "Mozilla/5.0 (compatible; Googlebot/2.1; +http://www.google.com/bot.html)",
+  },
+] as const;
+
+const REQUIRED_ENDPOINTS = {
```
tylerslaton, line 32: This is still a second monitoring manifest: all ten route/media-type tuples are copied from the public contract, then `validateAeoSyntheticConfig`, `targetsFromContract`, and the hard-coded success count depend on the copy. That undercuts the PR's main simplification claim and creates drift work. Please flatten the eligible website/docs standard/community endpoints directly from the contract (normalizing `/**` to `/`) and derive the reported target count dynamically. The existing `AeoSurfaceContract` loader/type can also replace the partial duplicate contract model here.

showcase/scripts/check-aeo-synthetics.ts
```diff
@@ -0,0 +1,528 @@
...
+    const error = hostError(canonical, origin, "canonical URL");
+    if (error) reasons.push(error);
+    if (
+      canonical &&
+      URL.canParse(canonical) &&
+      new URL(canonical).href !== `${origin}/`
+    ) {
+      reasons.push(`canonical URL is ${canonical}; expected ${origin}/`);
+    }
+  } else if (target.path === "/robots.txt") {
```
tylerslaton, line 311: This fetches `robots.txt` with each crawler UA but only validates the sitemap URL; it never checks whether the returned rules disallow that crawler. A future `User-agent: OAI-SearchBot` + `Disallow: /` would pass even though PDX-340's outcome is to detect when crawlers lose access. Please evaluate the applicable robots rules for the current crawler and fail when the monitored surfaces are disallowed.

tylerslaton (changes requested):
Much better: the knowingly-red schedule is gated, requests are bounded, unrelated MCP/capability/raw/OG/JSON-LD scope is gone, and the existing parser helpers are reused. CI is green.

Two focused issues remain before I’d call this clean and complete for the manual-first rollout: the checker still maintains a second endpoint manifest despite claiming the contract is the source of truth, and crawler-specific robots access is not actually validated.

tylerslaton (approved):

PR #6461, opened 2026-08-12
fix(sdk-python): honor config schema_keys in state snapshots

### LINKED ISSUE #3204: ag-ui-langgraph: schema_keys config not respected in get_state_snapshot
### Summary
Custom `schema_keys` passed to `LangGraphAgent` config are ignored when filtering state snapshots, causing state fields to be filtered out even when explicitly configured.

### Environment
- Package: `ag-ui-langgraph` version `0.0.25`
- Python: 3.11

### Current Behavior
When passing custom `schema_keys` to `LangGraphAgent`:

```python
langgraph_agent = LangGraphAgent(
    name="demo",
    graph=workflow,
    config={
        "schema_keys": {
            "output": ["steps"]  # Should include 'steps' in STATE_SNAPSHOT
        }
    }
)
```

The `steps` field is still filtered out from `STATE_SNAPSHOT` events, causing the frontend to not receive state updates.

### Root Cause
In `ag_ui_langgraph/agent.py`, the `get_schema_keys()` method (lines 412-440) only reads schema keys from the graph's JSON schema and ignores `self.config`:

```python
def get_schema_keys(self, config) -> SchemaKeys:
    try:
        input_schema = self.graph.get_input_jsonschema(config)
        output_schema = self.graph.get_output_jsonschema(config)
        # ...
        output_schema_keys = list(output_schema[\"properties\"].keys())
        
        return {
            \"input\": [*input_schema_keys, *self.constant_schema_keys],
            \"output\": [*output_schema_keys, *self.constant_schema_keys],  # Doesn't check self.config
            # ...
        }
```

Then in `get_state_snapshot()` (lines 486-490), the state is filtered:

```python
def get_state_snapshot(self, state: State) -> State:
    schema_keys = self.active_run[\"schema_keys\"]
    if schema_keys and schema_keys.get(\"output\"):
        state = filter_object_by_schema_keys(state, [*DEFAULT_SCHEMA_KEYS, *schema_keys[\"output\"]])
    return state
```

Since custom schema keys from `self.config` are never merged, fields like `steps` get filtered out.

### Suggested Fix
Modify `get_schema_keys()` to merge custom schema keys from `self.config`:

```python
def get_schema_keys(self, config) -> SchemaKeys:
    try:
        input_schema = self.graph.get_input_jsonschema(config)
        output_schema = self.graph.get_output_jsonschema(config)
        config_schema = self.graph.config_schema().schema()

        input_schema_keys = list(input_schema[\"properties\"].keys()) if \"properties\" in input_schema else []
        output_schema_keys = list(output_schema[\"properties\"].keys()) if \"properties\" in output_schema else []
        config_schema_keys = list(config_schema[\"properties\"].keys()) if \"properties\" in config_schema else []
        context_schema_keys = []

        if hasattr(self.graph, \"context_schema\") and self.graph.context_schema is not None:
            context_schema = self.graph.context_schema().schema()
            context_schema_keys = list(context_schema[\"properties\"].keys()) if \"properties\" in context_schema else []

        # ADD THIS: Merge custom schema_keys from self.config
        if isinstance(self.config, dict) and \"schema_keys\" in self.config:
            custom_schema_keys = self.config[\"schema_keys\"]
            if \"input\" in custom_schema_keys:
                input_schema_keys = list(set([*input_schema_keys, *custom_schema_keys[\"input\"]]))
            if \"output\" in custom_schema_keys:
                output_schema_keys = list(set([*output_schema_keys, *custom_schema_keys[\"output\"]]))
            if \"config\" in custom_schema_keys:
                config_schema_keys = list(set([*config_schema_keys, *custom_schema_keys[\"config\"]]))
            if \"context\" in custom_schema_keys:
                context_schema_keys = list(set([*context_schema_keys, *custom_schema_keys[\"context\"]]))

        return {
            \"input\": [*input_schema_keys, *self.constant_schema_keys],
            \"output\": [*output_schema_keys, *self.constant_schema_keys],
            \"config\": config_schema_keys,
            \"context\": context_schema_keys,
        }
    except Exception:
        return {
            \"input\": self.constant_schema_keys,
            \"output\": self.constant_schema_keys,
            \"config\": [],
            \"context\": [],
        }
```

### Impact
This bug prevents custom state fields from being included in STATE_SNAPSHOT events, breaking features like progress tracking and custom state visualization in the frontend.

### Workaround
Currently, the only workaround is to ensure all custom state fields are properly annotated in the TypedDict so they appear in the graph's output JSON schema, but this is not always desirable or possible.

deepunyk:
## Summary

`LangGraphAGUIAgent` accepts a `config`, but `schema_keys` inside it was never read. Output keys came only from graph introspection, so a state key absent from the graph's output schema was filtered out of every `STATE_SNAPSHOT` even when the caller declared it.

With `steps` absent from the state schema:

```python
agent = LangGraphAGUIAgent(
    name="demo",
    graph=graph,
    config={"schema_keys": {"output": ["steps"]}},
)
agent.active_run = {"schema_keys": agent.get_schema_keys({})}
agent.get_state_snapshot({"messages": [], "steps": ["research", "write"]})

# before: {'messages': []}
# after:  {'messages': [], 'steps': ['research', 'write']}
```

## Root cause

The base class builds `output` from `graph.get_output_jsonschema(config)` plus `constant_schema_keys`, then `get_state_snapshot` filters state through it. `self.config` is never consulted, so a declared key has no path into the list.

`LangGraphAGUIAgent.get_schema_keys` now calls `super()` and appends the configured keys. Wrapping `super()` rather than reimplementing it is what makes the base class's introspection-failure fallback honour them too, and that fallback is the path an unintrospectable graph lands on.

Two decisions: configured keys are **appended, never substituted**, so no current snapshot can lose a field; and `schema_keys` is **validated in `__init__`** rather than in `get_schema_keys`, which runs once per run, so malformed config raises `CopilotKitMisuseError` when the agent is built.

## Changes

- `langgraph_agui_agent.py`: validate `config["schema_keys"]` in `__init__`, merge it in a new `get_schema_keys` override. Only the four buckets the base class returns are merged; anything else is ignored rather than invented.
- `tests/test_agui_agent.py`: 15 tests. One class patches the base class's derived keys to isolate merge semantics (including the fallback shape and misuse cases); one runs against a real compiled `StateGraph` and asserts through `get_state_snapshot`, the layer the bug shows up at.
- `state-inputs-outputs.mdx`: this page teaches the output-schema route, which the issue names as its only workaround. Added a Python section for keys that route cannot express.

## Out of scope

`constant_schema_keys` is untouched. Defaulting `ag-ui` into it, renaming it to `ag_ui`, and the `CopilotKitMiddleware` A2UI path are design calls for maintainers. This change does give a hyphenated key a supported route, since the filter is a dict-key membership test and never needs a Python identifier.

## Test plan

- [x] `cd sdk-python && python -m pytest tests/ -q` - 225 passed, 11 skipped (`main`: 210 and 11; +15 here).
- [x] Same run with the source change stashed - 9 failed, 6 passed. The new tests fail without the fix, and the 6 that assert unchanged behaviour pass on `main` too.
- [x] `ruff format --check` and `ruff check` on both touched Python files - clean (ruff 0.15.13, the version pinned in `static_quality.yml`).
- [x] `cd showcase/shell-docs && npm ci && npx next build` - 222 static pages, no errors, so the new docs section compiles.
- [ ] CI (`test / unit / python-sdk` on 3.10 - 3.14, `static / quality`). Local runs were Python 3.12 only.

Closes #3204


BenTaylorDev:
One open decision before this merges, surfaced by #6478 — a duplicate of this fix that I'm closing in your favor, since this one also updates the docs page.

That PR handled a malformed `schema_keys` by silently ignoring it and falling back to the derived keys. This one raises `CopilotKitMisuseError` from `__init__`. Your reasoning in the PR body is sound — it's static misuse, so failing when the agent is built beats failing once per run — and in isolation I'd agree.

What gives me pause is the upgrade path. On main today, a stray or mistyped `schema_keys` is silently ignored, so an app can be carrying one right now and working fine. This lands as a patch-level bug fix, and after it, that same app raises at agent construction — it stops booting. That's a large blast radius for a fix whose purpose is to make a declared key reach the snapshot.

I'd suggest keeping the validation but softening the response: log a warning and fall back to the derived keys, rather than raising. That still tells the developer they've misconfigured something, at the same moment, without turning a previously-ignored value into a startup failure. If you'd rather keep the raise, I won't block on it — but I'd want it called out as a breaking change in the PR body so it's visible at release time rather than discovered by someone upgrading.

Everything else here checks out. I verified the merge wraps `super()` correctly, the four buckets match what the base class returns, and appending rather than substituting means no existing snapshot can lose a field.

PR #6466, opened 2026-08-12
fix(showcase/harness): re-auth on 403 from an expired PocketBase token

jpr5:
## Root cause

The harness's PocketBase client (`showcase/harness/src/storage/pb-client.ts`) re-authenticated its superuser token **only on HTTP 401**. But when the superuser/admin auth token's ~14-day TTL expires, PocketBase does **not** return 401 — it treats the request as an unauthenticated *guest* and returns:

```
HTTP 403 {"code":403,"message":"Only admins can perform this action.","data":{}}
```

on every write. Because 403 was never treated as an auth-expiry signal, the expired token was never refreshed, so **all `status` writes failed permanently** until the process restarted. `classifyWriterError` maps 403 → `pb_permission` (a terminal reason), so the failure looked like a permission problem rather than an expired session. This is what blanked the dashboard for ~46h.

## The fix

In `request()`, treat a 403 as the same stale-session signal as a 401 — **but only when the request actually carried an `Authorization` header** (`sentAuth`). A 403 on a request that sent no token is a genuine guest-forbidden result that re-auth cannot fix, so it is left to surface.

- The retry stays bounded by `MAX_AUTH_RETRIES` (1). A 403 that **persists after a fresh, successful re-auth** is a real permission error and falls through to the caller (still classified `pb_permission`) — never an infinite re-auth loop.
- No change to the 401 path, the retry envelope, or any other status class.

```
(res.status === 401 || (res.status === 403 && sentAuth)) &&
authRetries < MAX_AUTH_RETRIES && attempts < maxAttempts
```

## Local red-green proof (real PocketBase, real client — not a fake)

Stood up a live **PocketBase v0.22.21** (the pinned version) locally, created an admin + a superuser-gated `status` collection, and set `adminAuthToken.duration = 5` (5s — the server's minimum). A temporary driver drove the **real `createPbClient`** against it: write #1 caches a token, sleep 6.5s so the cached token **genuinely expires**, then write #2.

First confirmed the raw failure surface — an expired admin token on a write:

```
EXPIRED-token write status + body:
{"code":403,"message":"Only admins can perform this action.","data":{}}
HTTP 403
```

### RED (unmodified code)

```
[driver] write#1 OK id=setjh0ca1s09s14 — token now cached
[driver] sleeping 6.5s for the cached admin token to expire...
CVDIAG component=pb-client:create:status ... status=error error=status=403 {"code":403,"message":"Only admins can perform this action.","data":{}}
[driver] RED: write#2 FAILED after expiry: Error: pb create failed: 403 {"code":403,"message":"Only admins can perform this action.","data":{}}
EXIT=1
```

The expired token 403s, **no re-auth occurs**, the write stays failed.

### GREEN (with this fix)

```
[driver] write#1 OK id=tkl59dt5d3xt11g — token now cached
[driver] sleeping 6.5s for the cached admin token to expire...
[driver] GREEN: write#2 SUCCEEDED after expiry id=uns9y2dgysynpwz
EXIT=0
```

Same repro, same expired token: the 403 now triggers re-auth, the write is retried once and **succeeds**.

## Regression tests

Added three tests to `pb-client.test.ts`:

1. `re-auths on 403 (expired superuser token treated as guest) then retries the write` — 403-with-token → re-auth → retry succeeds (2 auths, 2 writes).
2. `caps 403 re-auth at 1 — a 403 that persists after a fresh auth surfaces (no infinite loop)` — bounded; the persistent 403 surfaces (2 auths, 2 writes, then throws).
3. `does NOT re-auth on 403 when no credentials were sent (genuine guest-forbidden)` — no token → no re-auth, no retry (0 auths, 1 write).

**Mutation check:** reverting the fix (403 branch removed) makes tests 1 and 2 fail while test 3 still passes — the tests are structurally able to detect the fix.

## Code-review hardening (Tier-3 cr-loop)

A full-breadth review of the re-auth branch surfaced two additional load-bearing issues in the exact code this PR modifies; both fixed here with their own red-green + individual mutation checks:

- **Drain the response body on the re-auth path.** The 401/403 re-auth branch did `continue` without draining the prior failed response — unlike the 429/5xx branches, which call `drainBody()` — leaking a half-consumed socket on every token refresh (F2.3 socket-reuse discipline). `drainBody` was hoisted above the branch and invoked before the retry.
  - RED: `failed401.bodyUsed` = `false` (undrained). GREEN: body drained after the fix.
- **Bound the re-auth gate by `attempts < maxAttempts`.** The re-auth gate checked only `authRetries`, not `attempts` (the 429/5xx gates check both), so a token expiring on the final attempt could fire a 4th `fetchImpl`, exceeding the documented `maxAttempts = 3` envelope. Added the guard for consistency.
  - RED: `expected 4 to be 3` (4th fetch fired). GREEN: `writeCount === 3`.

Full `pb-client.test.ts` suite: **35 passed**. CI green.

## Follow-ups (out of scope for this PR — pre-existing, tracked separately)

The review confirmed the fix is sound and found no defect in it, but flagged pre-existing issues in the same file that predate this change and belong in their own PRs:

- **Observability regression (HF13-B1):** `create()`'s CVDIAG "every record write failure is greppable" log is unreachable for retry-exhausted 429/5xx writes, because `request()` now throws `PbHttpError` before `create()`'s `!res.ok` block runs. (403 writes are unaffected — they reach the log.)
- **Auth re-auth stampede:** `ensureAuth()` has no single-flight guard, so at token expiry every concurrent writer re-auths independently. Fixing this (coalesce concurrent re-auths behind one shared in-flight promise) benefits both the 401 and 403 paths.
- **401 `sentAuth` symmetry (trivial):** the 401 re-auth path lacks the `sentAuth` guard the new 403 path has, wasting one bounded attempt when no credentials are configured.
- **`deleteByFilter` off-by-one:** the iteration cap throws on a fully-successful delete of exactly a multiple-of-200 ≥ 20000 rows.
- **Inert `RETRY_AFTER_MAX_MS` cap + its mutation-blind test.**

### DIFF
--- a/showcase/harness/src/storage/pb-client.test.ts
+++ b/showcase/harness/src/storage/pb-client.test.ts
@@ -266,6 +266,63 @@ describe("pb-client", () => {
     expect(writeCount).toBe(1);
   });
 
+  it("drains the failed response body on the 401 re-auth path (F2.3 socket reuse)", async () => {
+    // The 429 and 5xx retry branches drain the failed response body before
+    // continuing so the runtime's HTTP agent can reuse the socket. The
+    // re-auth (401/403) branch must do the same — otherwise every token
+    // refresh leaves a half-consumed socket. Observe drain via bodyUsed on
+    // the exact Response object returned for the failing attempt.
+    const failed401 = new Response("stale-token", { status: 401 });
+    let getCount = 0;
+    const fetchImpl = makeFetch((url) => {
+      if (url.includes("auth-with-password")) {
+        return new Response(JSON.stringify({ token: "tok" }), { status: 200 });
+      }
+      getCount += 1;
+      if (getCount === 1) return failed401;
+      return new Response(JSON.stringify({ id: "r1" }), { status: 200 });
+    });
+    const pb = createPbClient({
+      url: "http://pb",
+      email: "e",
+      password: "p",
+      logger,
+      fetchImpl,
+    });
+    await pb.getOne("status", "r1");
+    // Body of the 401 response must have been consumed (drained).
+    expect(failed401.bodyUsed).toBe(true);
+  });
+
+  it("stays within the maxAttempts envelope when a token expires on the final attempt", async () => {
+    // A token that expires on the LAST attempt must not enter the re-auth
+    // branch and fire a 4th fetch beyond the documented maxAttempts=3
+    // envelope. Attempts 1 and 2 are consumed by 5xx backoff; attempt 3
+    // returns 401. The re-auth gate must honor attempts < maxAttempts like
+    // the 429/5xx gates and NOT retry.
+    let writeCount = 0;
+    const fetchImpl = makeFetch((url) => {
+      if (url.includes("auth-with-password")) {
+        return new Response(JSON.stringify({ token: "tok" }), { status: 200 });
+      }
+      writeCount += 1;
+      if (writeCount < 3) return new Response("", { status: 503 });
+      return new Response("", { status: 401 });
+    });
+    const pb = createPbClient({
+      url: "http://pb",
+      email: "e",
+      password: "p",
+      logger,
+      fetchImpl,
+    });
+    await expect(pb.getOne("status", "x")).rejects.toThrow(
+      /pb getOne failed: 401/,
+    );
+    // Exactly maxAttempts requests to the records endpoint — no 4th call.
+    expect(writeCount).toBe(3);
+  });
+
   it("retries when fetchImpl throws (network error) with same envelope as 5xx", async () => {
     let attempts = 0;
     const fetchImpl = makeFetch(() => {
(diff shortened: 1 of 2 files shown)


BenTaylorDev (approved):
Verified against current `main`: the bug is still live. `showcase/harness/src/storage/pb-client.ts:307` reads `if (res.status === 401 && authRetries < MAX_AUTH_RETRIES)` — 401 only, with no `403` or `sentAuth` anywhere in the file. So an expired superuser token still never triggers re-auth and every `status` write keeps failing until the process restarts.

The fix is tightly bounded in the three ways that matter:

- `sentAuth` gates it, so a genuine guest 403 still surfaces instead of being retried.
- It reuses the existing `MAX_AUTH_RETRIES = 1` envelope, so a 403 that persists after a successful re-auth falls through as a real permission error — no re-auth loop.
- The 401 path, the retry envelope, and the other status classes are untouched.

179 lines of test against 55 lines of change, and the root-cause writeup correctly identifies why this looked like `pb_permission` rather than an expired session.

CI green on `9d8bde8095` (18 success, 3 skipped), no conflicts against `main`.

PR #6473, opened 2026-08-13
fix(react, vue): preserve popup rounded corners

### LINKED ISSUE #6472: fix(react, vue): popup wrapper background masks rounded corners
## Description

## Reproduction

3. Inspect the popup corners.

## Expected behavior

## Affected implementations

## Proposed minimal fix

ShitK:
## What does this PR do?

Fixes the v2 popup wrapper background in both React and Vue.

The outer fixed positioning wrapper keeps `data-copilotkit` for theme scoping, but now explicitly applies `cpk:bg-transparent`. This prevents the default theme background from showing around the inner popup's 16px rounded corners.

Regression tests verify that the positioning wrapper remains theme-scoped and carries the transparent background utility in both implementations.

## Related PRs and Issues

- Fixes https://github.com/CopilotKit/CopilotKit/issues/6472

## Testing

- `pnpm nx run @copilotkit/react-core:test --skip-nx-cache --excludeTaskDependencies --outputStyle=static`
- `pnpm nx run @copilotkit/vue:test --skip-nx-cache --excludeTaskDependencies --outputStyle=static`
- `pnpm nx run-many -t build:css -p @copilotkit/react-core @copilotkit/vue --parallel=2 --skip-nx-cache --outputStyle=static`
- Direct type checks for `@copilotkit/react-core` and `@copilotkit/vue`
- Pre-commit affected package tests, publint, and attw checks

## Checklist

- [x] Documentation is not required for this focused bug fix

## Summary by CodeRabbit

* **UI Improvements**
  * Improved popup layout behavior so its contents stretch appropriately within the container.
  * Updated popup styling to maintain a transparent background across React and Vue implementations.

* **Tests**
  * Added integration coverage verifying popup wrapper attributes and styling classes when the popup is open.

### DIFF
--- a/packages/react-core/src/v2/components/chat/CopilotPopupView.tsx
+++ b/packages/react-core/src/v2/components/chat/CopilotPopupView.tsx
@@ -1,12 +1,14 @@
 import React, { useEffect, useMemo, useRef, useState } from "react";
-import CopilotChatView, {
+import type {
   CopilotChatViewProps,
   WelcomeScreenProps,
 } from "./CopilotChatView";
+import CopilotChatView from "./CopilotChatView";
 import CopilotChatToggleButton from "./CopilotChatToggleButton";
 import { CopilotModalHeader } from "./CopilotModalHeader";
 import { cn } from "../../lib/utils";
-import { renderSlot, SlotValue } from "../../lib/slots";
+import type { SlotValue } from "../../lib/slots";
+import { renderSlot } from "../../lib/slots";
 import {
   CopilotChatConfigurationProvider,
   CopilotChatDefaultLabels,
@@ -209,7 +211,7 @@ function CopilotPopupViewInternal({
     <div
       data-copilotkit
       className={cn(
-        "cpk:fixed cpk:inset-0 cpk:z-[1200] cpk:flex cpk:max-w-full cpk:flex-col cpk:items-stretch",
+        "cpk:fixed cpk:inset-0 cpk:z-[1200] cpk:flex cpk:max-w-full cpk:flex-col cpk:items-stretch cpk:bg-transparent",
         "cpk:md:inset-auto cpk:md:bottom-24 cpk:md:right-6 cpk:md:items-end cpk:md:gap-4",
       )}
     >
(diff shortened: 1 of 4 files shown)


BenTaylorDev (changes requested):
I reviewed this against current `main`. The engineering is sound, and it is smaller than it looks.

**Net effect is two lines.** The branch is well behind `main`, but merging it changes exactly one class token in the React wrapper and one in the Vue wrapper. The `import type` reflow in the diff disappears on merge, because `main` has since made the identical change independently.

**The mechanism holds.** I checked the cascade rather than trusting the class name. The rule that paints the wrapper lives in `@layer base` in `packages/react-core/src/v2/styles/globals.css`, and Tailwind utilities land in `@layer utilities`, which wins over `base` regardless of specificity. So `cpk:bg-transparent` does take effect.

**Scope is right.** The popup wrapper is the only `data-copilotkit` element that is a pure positioning box larger than the panel it contains. The sidebar's `data-copilotkit` is the panel itself and must keep its background, so it correctly stays untouched. Angular is genuinely unaffected — `packages/angular/src/lib/components/modal/copilot-popup.ts` sets `display: contents` on the host, exactly as you said in the issue.

**Mobile is safe.** Below `md` the inner dialog is `h-full w-full rounded-none bg-background`, so it fills the now-transparent wrapper and nothing shows through.

Two things before this can go in.

**1. It is still a draft.** Mark it ready for review when you are happy with it — I cannot merge a draft.

**2. The tests are change detectors.** Both assert `classList.contains("cpk:bg-transparent")`, which is the literal string the fix adds. jsdom loads no stylesheet, so these would still pass if a future layer change made the class inert — they pin the class, not the behavior. That is worth something, so I am not asking you to remove them. If you want the test to actually cover the bug, assert on the computed background of the wrapper with the real stylesheet applied, or leave a comment on the test saying it guards the class only.

Worth knowing: this pattern has already cost users. `examples/showcases/grok-generative-ui/app/globals.css` carries a hand-written workaround for `[data-copilotkit]` painting an opaque background. Your fix removes the need for it.

ShitK:
Thanks for the detailed review. I rebased the branch onto current `main` and kept the tests as focused class-contract checks. Both React and Vue test names now describe that narrower contract, with an explicit comment that jsdom does not load the generated stylesheet and therefore does not verify the cascade itself.

The two popup test files pass locally, and I have marked the PR ready for review.

BenTaylorDev (approved):

PR #6474, opened 2026-08-13
fix(react-core): include readable context in CopilotTask

ShitK:
## What does this PR do?

Restores the minimum user-facing behavior described in #6408: `CopilotTask` can once again include entries registered through `useCopilotReadable`.

`useCopilotReadable` writes to the active flat context store, while the v1 `getContextString()` path still read only the legacy tree. This change bridges the v1 `CopilotTask` reader to `copilotkit.getContextForAgent()` and appends readable entries without replacing the existing document/tree output.

The scope deliberately stays narrow:
- no hierarchy or context-category serialization
- no restoration of `parentId` or categories
- no dev-console or `CopilotTextarea` changes

Agent-scoped contexts remain filtered by using `getContextForAgent()`.

## Related PRs and Issues

- Addresses the `CopilotTask` minimum-fix portion of https://github.com/CopilotKit/CopilotKit/issues/6408
- Avoids overlapping with the now-merged #6409

## Testing

Validated after rebasing onto the latest `main`:

- `pnpm nx run @copilotkit/react-core:build --skip-nx-cache --outputStyle=static`
- `NODE_OPTIONS=--no-experimental-webstorage pnpm nx run @copilotkit/react-core:test --skip-nx-cache --excludeTaskDependencies --outputStyle=static`
  - 127 test files and 1,500 Vitest tests passed
  - 47 package script tests passed
- `pnpm nx run @copilotkit/react-core:check-types --skip-nx-cache --excludeTaskDependencies --outputStyle=static`
- Nx format check for both changed files
- `git diff --check upstream/main...HEAD`

The Node option disables Node 25's experimental global Web Storage so jsdom supplies its own standards-compatible `localStorage`; no source changes were made for the local environment.

## Checklist

- [x] Added regression tests
- [x] Ran the affected package build and test suite
- [x] Ran type and formatting checks
- [x] Documentation is not required for this focused compatibility fix

## Summary by CodeRabbit

* **Bug Fixes**
  * Deprecated readable context entries are now included in the generated context string.
  * Context entries scoped to other agents are excluded from the current agent’s context.
  * Readable context is appended only when available, preserving existing document and category context behavior.

* **Tests**
  * Added coverage verifying readable context bridging and agent-specific filtering.


BenTaylorDev (changes requested):
This is the right fix and the right source. Sorry it sat this long. I applied it to current `main` and it does what it says: with a readable registered, the expression `CopilotTask` evaluates goes from `"\n\n"` to `"\n\n\n\nUser profile:\n{\"name\":\"Ada\"}"`. Your second test earns its place — reading `getContextForAgent()` rather than the raw store keeps `agentIds`-scoped entries out, and I confirmed the alternative approach leaks them.

Full `v1-deprecated` suite passes at 140/140 with this applied, and the react-core and react-textarea test targets both pass.

Two things before it lands.

**1. The v1 tree moved.** `e499c65dce` relocated it to `packages/react-core/src/v1-deprecated/`. Your source change follows the rename cleanly, but the new test file needs to move to `src/v1-deprecated/components/copilot-provider/__tests__/v1-readable-context-bridge.test.tsx`.

**2. Moving it is not enough — the import depth changes.** At the new path, `import { useCopilotKit } from "../../../v2"` resolves to `src/v1-deprecated/v2`, which does not exist, and the suite collects zero tests:

```
Error: Failed to resolve import "../../../v2" from
"src/v1-deprecated/components/copilot-provider/__tests__/v1-readable-context-bridge.test.tsx"
```

It needs `"../../../../v2"`. With that one character changed, both of your tests pass.

**One optional nit.** When there are no documents and no tree entries — the `CopilotTask` case, which is every case now — the result opens with four newlines, because the empty document string and empty tree string each still contribute a separator. Worth collapsing while you are in there, though it does not block.

For context on where this sits: it fixes item 1 of #6408. Items 2 and 4 stay open, and item 3 is being handled separately in #6931 by deprecating the textarea rather than restoring it, so there is no conflict with your change to `getContextString`.

ShitK:
Thanks for the review. I rebased the branch onto current `main`, moved the regression test to `packages/react-core/src/v1-deprecated/components/copilot-provider/__tests__/v1-readable-context-bridge.test.tsx`, and updated the `useCopilotKit` import to `../../../../v2`.

The relocated regression test passes (2/2), and the React Core type check passes. I left the optional leading-newline cleanup out so this remains focused on the readable-context bridge. The updated branch has been force-pushed.

BenTaylorDev (approved):
Re-reviewed against current `main` (170 commits ahead; nothing on main touches `copilotkit.tsx`, and the fix is not already there). Both blocking items are addressed — the test is at the `v1-deprecated` path and the import depth is `"../../../../v2"`.

`react-core` is green rebased: 149 files, 1689 tests, and your file 2/2. I mutation-checked both cases rather than trusting the green: reverting the `getContextString` change fails the first test, and swapping `getContextForAgent()` for `Object.values(copilotkit.context)` fails the second with `expected '\n\n\n\nOther agent secret:\nclassifi…' not to contain 'Other agent secret'`. Neither test is vacuous. `check-types`, `oxfmt` and `oxlint` are clean, and `react-textarea` still passes.

Two notes, neither blocking. The four-newline opener is still there — you can see it in that assertion output above. And `getContextString` also feeds `react-textarea`'s autosuggestion and insertion paths, so readables now reach those too; that is consistent with the v1 behavior being restored, and #6931 deprecates the textarea anyway, so there is no conflict.

This closes item 1 of #6408. Items 2 and 4 stay open there.

PR #6476, opened 2026-08-13
chore(packages): adopt TypeScript 7 and tsdown 0.22

AlemTuzlak:
## Typecheck speed

I did not re-run TypeScript 5 `tsc` on this machine. A recent green TypeScript 5 `check-types` job (#6477) was **5 m 10 s** wall time for the whole job. The workflow file still says the TypeScript 5 `runtime` check is about 5 minutes. On TypeScript 7 that same package was **1.75 s** locally.

## What does this PR do?

Move published packages to TypeScript 7.0.2 and tsdown 0.22.14.

TypeScript 7 ships the Go compiler as the normal `tsc` binary. Scripts stay on `tsc --noEmit`. This does not add `@typescript/native-preview` or rename commands to `tsgo`.

Angular stays on TypeScript 5.9.3. Vue stays on TypeScript 5.8.2 and `vue-tsc`. Examples, showcase, and integrations stay on TypeScript 5 for a later pass.

## Why

TypeScript 7 is the current stable compiler. The old `tsgo` preview name is gone. Package typecheck and declaration emit now use the official `tsc`.

## What changed

- Bump `typescript` to `7.0.2` and `tsdown` to `^0.22.14` on published packages except Angular and Vue.
- Shared tsconfig: `moduleResolution: bundler`, `types: ["node"]`, remove `declarationDir`.
- Pin `rolldown` to `1.2.3` so the lockfile can install (1.2.4 was blocked by release age).
- Give Analog its own TypeScript 5.9.3 so Angular tests keep a JS API.
- Widen `@copilotkit/sdk-js` typescript peer to `^5.2.3 || ^6.0.0 || ^7.0.0`.
- Runtime: public imports, explicit GraphQL types, `tsx` for schema generate.
- react-core: hoist `Cursor`. CSS ambient modules for react packages.
- react-ui / react-textarea: public import is still `styles.css`. tsdown 0.22 emits `dist/style.css` instead of `dist/index.css`, so `./styles.css` now points at that file. An extra `./style.css` export exists only because tsdown lists the file it writes.
- web-inspector: `@tsdown/css` with `?inline`, PostCSS path, no Lightning rewrite. Overlay CSS matches the source file plus one trailing newline.
- CI jobs that run `tsdown` now use Node 22.18+. `tsdown` 0.22 engines are `^22.18.0 || >=24.11.0`. The unit matrix is Node 22, 24, and 26 (each with React 18 and 19). Node 20 is gone because those jobs build packages first.
- `@copilotkit/sqlite-runner` now pins `better-sqlite3@12.11.1` (peer `^12.10.0`). `12.5.0` has no Node 26 prebuild and fails to compile (`PropertyCallbackInfo` has no `This`).
- Node 26 still leaves `window.localStorage.getItem` missing in jsdom. `loadInspectorState` now returns null instead of throwing. react-core and web-inspector vitest pass `--no-experimental-webstorage` on Node 22.4+.
- Release scripts that walk TypeScript source load Angular's TypeScript 5.9.3 compiler API. TypeScript 7.0 has no public JS API.

## What did not move

- Angular 5.9.3 (`ng-packagr` / Analog need the TypeScript JS API).
- Vue 5.8.2 + `vue-tsc`.
- Examples, showcase, and integrations.
- Published package `engines.node` for consumers. Only the monorepo root now says `>=22.18.0` because that is what a package build needs.

## Before / after (builds)

Numbers below come from three sources:

- Toolchain: `origin/main` lockfile vs this branch.
- Bundle sizes: published `@copilotkit/*@1.67.1` on npm (TypeScript 5 + tsdown 0.20) vs this branch local `dist` (TypeScript 7 + tsdown 0.22). Same package version. This branch also has a few source fixes, so a small size change is not compiler-only.
- Times: local Windows Node 24.3.0 for `tsc` / `tsdown`. GitHub Actions for CI job wall time.

### Toolchain

| Item | Before (`origin/main`) | After (this PR) |
| --- | --- | --- |
| `tsc` | TypeScript 5.9.2 (JS compiler) | TypeScript 7.0.2 (Go compiler, still named `tsc`) |
| Root `typescript` | `^5.2.3` | `7.0.2` |
| Most published packages | `^5.2.3` | `7.0.2` |
| Angular | `5.9.3` | `5.9.3` (no change) |
| Vue | `5.8.2` + `vue-tsc` | `5.8.2` + `vue-tsc` (no change) |
| `tsdown` | `^0.20.3` | `^0.22.14` |
| Bundler inside tsdown | rolldown from 0.20 | rolldown `1.2.3` (override) |
| dts emit | tsdown 0.20 + TypeScript 5 JS API | tsdown 0.22 + `typescript@7.0.2` |
| Root `engines.node` | `>=18` | `>=22.18.0` (tsdown 0.22) |
| CI jobs that run `tsdown` | Node 20.x | Node 22.x / 24.x |
| Unit matrix | Node 20 / 22 / 24 | Node 22 / 24 / 26 |
| `better-sqlite3` | `12.5.0` (no Node 26) | `12.11.1` (Node 26 prebuild) |

### Local typecheck (`tsc --noEmit`, TypeScript 7.0.2)

These are raw `tsc` times on this branch. They do not include Nx graph cost.

| Package | After |
| --- | --- |
| `@copilotkit/shared` | 0.58 s |
| `@copilotkit/web-inspector` | 1.11 s |
| `@copilotkit/react-ui` | 1.50 s |
| `@copilotkit/runtime` | 1.75 s |
| `@copilotkit/react-core` | 2.67 s |

A cold local `nx run-many -t check-types --projects=packages/** --skip-nx-cache` on this branch **passed in 7 m 55 s**. Nx ran 27 projects plus 18 dependent tasks. Some `check-types` targets also run a package build (dts emit), so this wall time is not `tsc` alone.

I did not run TypeScript 5 `tsc` on the same machine in this pass. The TypeScript 5 CI `check-types` job on a recent green PR (#6477) was **5 m 10 s** wall time for the whole job (install + codegen + `nx run-many -t check-types --parallel=1`). `.github/workflows/static_quality.yml` still documents the TypeScript 5 `@copilotkit/runtime` check as about 5 minutes and about 10 GB of heap.

On TypeScript 7, that same runtime file typechecked in **1.75 s** locally.

### Local package build (`tsdown` 0.22.14)

tsdown prints one time per format. dts uses `typescript@7.0.2`.

| Package | UMD | CJS | ESM |
| --- | --- | --- | --- |
| `@copilotkit/shared` | 0.25 s | 0.96 s | 0.98 s |
| `@copilotkit/web-inspector` | 0.29 s | 0.78 s | 0.85 s |
| `@copilotkit/core` | 0.33 s | 1.39 s | 1.40 s |

`pnpm run build` (all published packages) on this PR's earlier `pkg-pr-new` job: **2 m 44 s** on Node 24.19.0. Recent `main` `pkg-pr-new` runs were **2 m 47 s to 3 m 10 s**. Package bundle wall time is about the same. The big win is `tsc --noEmit`, not tsdown JS emit.

### Bundle sizes (entry files)

Raw bytes, then gzip bytes. Before = npm `1.67.1`. After = this branch `dist`.

| File | Before | After | Delta | Before gzip | After gzip |
| --- | ---: | ---: | ---: | ---: | ---: |
| `shared` `index.umd.js` | 77,168 | 77,161 | -7 | 21,523 | 21,681 |
| `shared` `index.mjs` | 5,044 | 5,043 | -1 | 1,685 | 1,685 |
| `core` `index.umd.js` | 254,270 | 254,285 | +15 | 56,075 | 56,122 |
| `core` `index.mjs` | 218,361 | 218,201 | -160 | 52,163 | 52,251 |
| `react-core` `index.umd.js` | 278,086 | 278,542 | +456 | 65,635 | 65,844 |
| `react-core` `index.mjs` | 64,013 | 67,184 | +3,171 | 16,840 | 17,879 |
| `react-ui` `index.umd.js` | 116,546 | 116,912 | +366 | 25,948 | 25,962 |
| `react-ui` `index.mjs` | 97,800 | 98,009 | +209 | 23,984 | 24,028 |
| `web-inspector` `index.umd.js` | 453,830 | 454,413 | +583 | 97,249 | 97,373 |
| `web-inspector` `index.mjs` | 349,156 | 349,785 | +629 | 74,843 | 75,038 |
| `web-inspector` `styles/generated.mjs` | 36,304 | 36,313 | +9 | 7,446 | 7,453 |
| `runtime` `index.mjs` | 3,159 | 3,133 | -26 | 788 | 777 |

Most entry files move by less than 1 kB. The one clear bump is `react-core` `index.mjs` (+3.2 kB). That file also has the `Cursor` hoist in this PR.

### Overlay CSS

`packages/web-inspector/src/styles/generated.css` is **35,926** bytes.

The string that tsdown embeds in `dist/styles/generated.mjs` is **35,927** bytes. That is the source file plus one trailing newline. The CSS text itself is the same.

## Stylesheet name (`styles.css` vs `style.css`)

This PR does **not** change the public import. Apps still use:

```ts
import "@copilotkit/react-ui/styles.css";
import "@copilotkit/react-textarea/styles.css";
```

Docs and JSDoc still say `styles.css`.

What changed is the **file on disk** after the build.

On `main`, tsdown 0.20 wrote `dist/index.css`. The export was:

```json
"./styles.css": "./dist/index.css"
```

tsdown 0.22 extracts that same CSS as `dist/style.css`. `dist/index.css` is gone. `package-quality` then failed: `pkg.exports["./styles.css"]` pointed at a file that does not exist.

So the old public name now points at the new file:

```json
"./styles.css": "./dist/style.css"
```

The extra `"./style.css"` export is only that new filename. tsdown lists the file it emits. You do not need that name.

## tsconfig changes and consumers

These edits are for this repo's typecheck and dts emit. They do not change the tsconfig in a consumer app.

Published packages only ship `dist`. An app that installs `@copilotkit/react-core` keeps its own `tsconfig.json`. It does not extend `packages/tsconfig/base.json`. `tsconfig` and `@copilotkit/typescript-config` are private workspace packages.

| Change | Why | Consumer effect |
| --- | --- | --- |
| `moduleResolution`: `node` to `bundler` | TypeScript 6/7 removed `node` / `node10`. `tsc` 7 fails without this. | None in their app. Their resolver stays theirs. |
| `module`: `esnext` | Pairs with `bundler`. | Same. |
| `"types": ["node"]` | TypeScript 7 defaults `types` to `[]`, so `process` / `Buffer` disappeared. `"*"` works on TypeScript 7 but TypeScript 5.8.2 treats `*` as a package name (TS2688). `"node"` works on both. | None. Their `@types` include is unchanged. |
| Remove `declarationDir: "./dist"` | tsdown 0.22 + TS 7 wrote `.d.ts` to `dist` instead of the plugin temp dir, so dts emit failed. | They still load types from `package.json` `types` / `exports`. |
| Runtime: drop `ts-node` block, `module` `esnext`, drop remix `paths` | Schema generate now uses `tsx`. `bundler` can resolve the remix package without a path hack. | Runtime users do not compile with this file. |
| web-inspector `dev/tsconfig` paths | Only the local harness. | None. |
| Angular Vite alias for TypeScript 5.9.3 | Analog has no TS JS API on TS 7. | None. Angular users still bring Angular's own TypeScript. |

The only indirect consumer surface is the `.d.ts` we emit. `bundler` vs old `node10` can change how some imports are written in those files. `attw` on web-inspector was clean. A consumer on TypeScript 5 with `moduleResolution: bundler` or `nodenext` is the usual case and should be fine.

This is not like Lightning CSS on the inspector overlay. Lightning can change Safari visuals. These tsconfig edits do not change consumer compiler options. If something broke, it would show up as a type error on a deep or odd import, not as a runtime style break.

## Checklist


rainerhahnekamp:
# TypeScript 7 and Angular

As long as the Angular package continues to use `ng-packagr` with its own supported TypeScript version, we should be safe. TypeScript 7 support is on Angular’s roadmap but is not available yet.

The Angular package is compiled with a separate TypeScript compiler, so it will not use TypeScript 7. It depends on `@copilotkit/{a2ui-renderer,core,shared,web-components}`, which it consumes as compiled JavaScript and TypeScript declaration files.

Microsoft states that TypeScript 7.0 and TypeScript 6.0’s type-checking run side by side:

https://devblogs.microsoft.com/typescript/announcing-typescript-7-0/#running-side-by-side-with-typescript-6.0

As discussed, we will upgrade the Angular package to Angular 22 as soon as possible and use its supported TypeScript 6 compiler.

This PR pins TypeScript 7.0.2. We can reassess compatibility separately when upgrading to TypeScript 7.1. From the Angular side, this has a green light once the build passes. 😀

/cc @wolfmanfx

PR #6477, opened 2026-08-13
fix(angular): prevent duplicate OpenGenerativeUI sandboxes

wolfmanfx:
Prevents stale async loaders from creating duplicate OpenGenerativeUI sandbox iframes by tying final sandbox creation to `afterRenderEffect` cleanup.

Adds regression tests for identical content and A → B → A races.

### Testing

- Angular tests
- Type checking
- Angular package build

### DIFF
--- a/packages/angular/src/lib/components/open-generative-ui/open-generative-ui-activity-renderer.ts
+++ b/packages/angular/src/lib/components/open-generative-ui/open-generative-ui-activity-renderer.ts
@@ -9,6 +9,7 @@ import {
   inject,
   input,
   signal,
+  untracked,
   viewChild,
   InjectionToken,
 } from "@angular/core";
@@ -183,7 +184,6 @@ export class CopilotOpenGenerativeUIRenderer implements OnChanges {
   private previewSandbox: WebsandboxInstance | undefined;
   private sandboxReady = false;
   private previewReady = false;
-  private finalSandboxKey: string | undefined;
   private executedExpressionIndex = 0;
   private pendingQueue: string[] = [];
   private jsFunctionsInjected = false;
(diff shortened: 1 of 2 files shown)


packages/angular/src/lib/components/open-generative-ui/open-generative-ui-activity-renderer.ts
```diff
@@ -258,6 +257,27 @@ export class CopilotOpenGenerativeUIRenderer implements OnChanges {
       this.destroyFinalSandbox();
     });
 
+    afterRenderEffect({
```
rainerhahnekamp, line 261: are we sure we don't need an `untracked` anywhere?

packages/angular/src/lib/components/open-generative-ui/open-generative-ui-activity-renderer.ts
```diff
@@ -258,6 +257,27 @@ export class CopilotOpenGenerativeUIRenderer implements OnChanges {
...
+      write: (onCleanup) => {
```
rainerhahnekamp, line 262: I'd like to have a small command why the `write`.

rainerhahnekamp (changes requested):

packages/angular/src/lib/components/open-generative-ui/open-generative-ui-activity-renderer.ts
wolfmanfx: Websandbox adds and removes an iframe, so its lifecycle belongs in the write phase.

packages/angular/src/lib/components/open-generative-ui/open-generative-ui-activity-renderer.ts
wolfmanfx: Done

rainerhahnekamp (approved):

PR #6484, opened 2026-08-13
docs(shell-docs): own the public AEO contract

samjulien:
## Why

#6458 put the contract in `showcase/shared` and widened shell-docs' Turbopack root to import it. Website and Pathfinder do not consume that file at build time, so the cross-deployment dependency was conceptual rather than real. The wider root changed Next.js runtime packaging and broke normal MDX pages in the staging container.

This replacement treats shell-docs as a standalone deployment boundary. Shell-docs owns the policy, machine-readable contract, publisher, validator, validator tests, and validator dependency. Repository CI only invokes the shell-docs-owned command. Nothing in `showcase/scripts` implements or validates the shell-docs contract.

The existing production synthetic remains repository-level because it intentionally monitors multiple deployments: the website, shell-docs, and Pathfinder. It consumes the published contract but is not part of the shell-docs application.

## What changed

- publish the human AEO policy at `/aeo`
- publish the versioned machine contract at `/.well-known/copilotkit-capabilities/v1.json`
- keep the contract and JSON Schema under `showcase/shell-docs/aeo`
- keep the validator and its tests under `showcase/shell-docs`
- expose validation as `npm run validate-aeo-contract` from shell-docs
- make repository CI call that shell-docs-owned command after the app's standalone install
- name external ownership for website and docs MCP gaps
- keep the default shell-docs Turbopack root and existing Docker layout unchanged
- add a regression test that forbids widening the build root for this contract

## Branch provenance

This branch is rebased onto `main` after the #6458 revert (#6483). Its two commits add the corrected shell-docs implementation and then consolidate validator ownership inside that deployment boundary. It does not restore the shared-root or Docker changes from #6458. It supersedes and closes #6482, whose Docker workaround is no longer needed.

## Verification

- boundary regression observed red against #6458 and green here
- shell-docs AEO suite: 8 files, 50 tests passed
- retained cross-deployment synthetic and deploy-guard suite: 3 files, 102 tests passed
- shell-docs-owned contract validation passed from a clean `npm ci`
- shell-docs typecheck passed
- shell-docs production build passed: 223 static pages
- local production server probes returned 200 with expected media types for `/`, `/quickstart`, `/aeo`, `/robots.txt`, `/sitemap.xml`, `/llms.txt`, and `/.well-known/copilotkit-capabilities/v1.json`
- shell-docs lint exited 0 with pre-existing warnings only
- Oxfmt passed
- standalone npm lockfile and root pnpm lockfile checks passed
- workflow YAML parsed successfully; path filters include all `showcase/**` changes

## Related

- replaces the reverted #6458 architecture
- supersedes #6482
- PDX-317

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -34,4 +34,5 @@ Happy contributing!
 
 - [ ] I have read the [Contribution Guide](https://github.com/copilotkit/copilotkit/blob/master/CONTRIBUTING.md)
 - [ ] If the PR changes or adds functionality, I have updated the relevant documentation
+- [ ] If the PR changes a public machine-readable surface, I reviewed the [public AEO surface contract](https://docs.copilotkit.ai/aeo)
 - [ ] "Allow edits by maintainers" is checked (lets us help iterate on your PR directly — faster turnaround for everyone)
--- a/.github/workflows/showcase_validate.yml
+++ b/.github/workflows/showcase_validate.yml
@@ -390,15 +390,15 @@ jobs:
         # version on a registry cache miss.
         run: pnpm exec vitest run
 
-      - name: Validate public AEO surface contract
-        run: pnpm nx run @copilotkit/showcase-scripts:validate-aeo-contract
-
       - name: Check deployed docs canonical behavior
         run: pnpm nx run @copilotkit/showcase-scripts:test -- __tests__/verify-deploy.drivers.test.ts
 
       - name: Install shell-docs dependencies
         working-directory: showcase/shell-docs
         run: npm ci --ignore-scripts
+
+      - name: Validate public AEO surface contract
+        run: npm --prefix showcase/shell-docs run validate-aeo-contract
 
       - name: Check shell-docs public AEO behavior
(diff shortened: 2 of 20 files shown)


BenTaylorDev:
Reviewed against current `main`. I rebased the branch locally to check it, and the substance holds up: the `.well-known` rewrite works as a real request (200, `application/json`), `/aeo` and `/aeo.md` render, and a missing version returns a true 404 rather than the HTML-200 fallback the contract forbids. The validator is also not vacuous — injecting a nonexistent enforcement path makes it exit 1, and every path the contract names still exists on today's `main`.

Three things before this lands:

1. **The inventory misses `/og/<slug>/og.png`** (`showcase/shell-docs/src/app/og/[...slug]/route.tsx`). It is a public machine surface that social and answer-engine crawlers fetch, and its own header comment records the exact failure class this contract exists to catch: it previously fell into a catch block and redirected to a 25-byte static fallback, producing "zero working OG images on most pages in prod" with healthy-looking status codes. Either add it as a surface or state in the contract why it is out of scope — a contract that claims to enumerate the public machine surfaces should not quietly omit one.

2. **CI placement.** The new shell-docs `npm ci` plus test steps sit in `showcase_validate.yml`, which triggers on all of `showcase/**` with no per-job gate, so a harness- or scripts-only change now pays a full shell-docs install. The repo's convention for shell-docs vitest jobs is `test_integration-docs.yml`, with path filters scoped to the files each job covers. Worth moving, or worth a comment saying why this one belongs in the broader job.

3. **Minor:** `/aeo` lands in the reader-facing sidebar under "Observe & Operate", between `inspector` and `vs-code-extension`. It has to be public for `policyUrl` to resolve, but it is a maintainer policy page sitting among product docs.

The branch is also stale enough that its green CI predates a lot of `main`. Rebasing hits two conflicts, both additive and easy: `src/app/sitemap.test.ts` (main added an AG-UI retirement test at the same spot — keep both) and `package-lock.json` (regenerate; it comes back as the same single `ajv` entry).

tylerslaton (approved):

PR #6495, opened 2026-08-14
Fix CopilotTask readable context inclusion

AlemTuzlak:
## Summary
- Wire `CopilotTask` readable-context inclusion to the active v2 flat context store exposed through the React provider.
- Preserve explicit `data` prepending and the `includeCopilotReadable: false` gate.
- Add focused regression coverage for default include, false gate, and explicit data preservation.

## Verification
- Attempted: `pnpm --dir "/tmp/CopilotKit" exec vitest run packages/react-core/src/lib/__tests__/copilot-task.test.ts`
- Blocked by sandbox dependency/environment issue: `spawn vitest EACCES` (same class of unusable dependency install / executable permission failure reported in the previous attempt). No test assertion failure was observed.

### DIFF
--- a/packages/react-core/src/components/copilot-provider/copilotkit.tsx
+++ b/packages/react-core/src/components/copilot-provider/copilotkit.tsx
@@ -199,6 +199,7 @@ export function CopilotKitInternal(cpkProps: CopilotKitProps) {
   });
 
   const { addElement, removeElement, printTree, getAllElements } = useTree();
+  const { copilotkit } = useCopilotKit();
   const [isLoading, setIsLoading] = useState(false);
   const [chatInstructions, setChatInstructions] = useState("");
   const [authStates, setAuthStates] = useState<Record<string, AuthState>>({});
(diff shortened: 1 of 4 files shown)


BenTaylorDev:
Thank you for this, and sorry it sat unreviewed. I am closing it in favour of #6474, which fixes the same defect. The reason is a behavioral difference I measured rather than a preference between two equivalent patches.

This branch reads `Object.values(copilotkit.context)`, which is the raw store including entries carrying `agentIds`. #6474 reads `copilotkit.getContextForAgent()`, which applies the per-agent filter from #5369 at `context-store.ts:52-59`. I applied each branch to current `main` and read both sources from one store at one moment, with one unscoped readable and one scoped to `other-agent`:

```
#6474 source = [{"description":"Public note","value":"visible"}]
#6495 source = "Public note:\nvisible\n\nOther agent secret:\nclassified"
```

So under this branch a readable a developer scoped to one agent would reach every `CopilotTask`. That is the blocking difference.

Two smaller points, recorded so they are not lost. `context.getCopilotReadableContextString?.() ?? ""` makes the new method optional on the public `CopilotContextParams`, so any consumer supplying its own context value silently gets an empty string again — the same silent-empty shape being fixed. And adding a member to that interface grows a surface that is now under `v1-deprecated/`.

Your diagnosis was right, and you got to it independently. Please do send more.

PR #6501, opened 2026-08-14
feat(reskinnable-demo): long-running Codex harness as an agent tool

mxmzb:
## What this is

Runs a real coding harness (OpenAI Codex, via `@tanstack/ai-codex`) for **minutes** against a CSV of personal card expenses — web-searching each merchant, deciding what the Austin offsite makes reimbursable, filing those against the banking ledger, and returning a React report card.

It ships as a `defineTool` on banking's existing classic agent. `EXPENSE_HARNESS_MODE` selects whether it is wired: `off` (default) or `tool`. An unrecognised value throws rather than silently disabling the beat.

The shape carries one limitation, stated in the code rather than hidden: `execute` returns a single resolved value, so intermediate frames leave through an SSE **side channel** that the console renders, never through the run's own event stream. A mid-run reload therefore loses the journey and the run replays as one tool chip.

## Reviewing it

```bash
cd examples/showcases/reskinnable-demo
cp .env.example .env               # set OPENAI_API_KEY
EXPENSE_HARNESS_MODE=tool pnpm dev
```

`codex` is an **external prerequisite** (`brew install codex` + `codex login`) — it is not installed by pnpm, and the harness authenticates through that login rather than `OPENAI_API_KEY`.

The beat is banking's ninth pill, *"Sort out my offsite expenses"*.

## Verification

Four gates at HEAD: `pnpm lint` clean · `pnpm typecheck` exit 0 · `pnpm test:unit` **2515 passing across 225 files** · `pnpm build` exit 0.

Two hard gates were run against the real binary before any of this was built, and both are recorded with their measurements:

- **Gate 1** — what Codex actually emits. AG-UI-shaped events; reasoning lives in `REASONING_MESSAGE_CONTENT.delta` (there is no `content` field). Every guess in the original plan was wrong; the mapper is built from the measurement.
- **Gate 2** — whether the runtime's converter carries it. Pass-through minus envelopes: 91 raw chunks in → 74 AG-UI events out, every text byte balancing.

## Notable in the diff

- `POST /api/banking/v1/transactions` is **new** — the plan told the harness to file against an endpoint that did not exist. Without it the harness would have taken a 405, filed nothing, and still reported transaction ids.
- The harness supersedes rather than refuses a second run: `defineTool`'s `execute` has no cancellation hook, so a refusal had no release and a cancelled turn locked the beat out for minutes.
- The SSE route ignores a backlog ending in a terminal frame — the console mounts before the server clears the channel, so a second run could otherwise replay the first one's `done` and freeze.
- One shell touch: `chat-panel.tsx`'s composer `accept`, reverted to its original value. `@ai-sdk/openai` accepts only images and `application/pdf` as file parts, so attaching the statement CSV failed the *run*, not the staging. `skin.test.tsx` pins that pill falling through, with the provider rule written beside it.
- **`@ag-ui/core` is pinned to `0.0.57`** (`d045a2a5`). `@ag-ui/a2a` and `@ag-ui/langgraph` declare it as a floating peer that no manifest constrained, so every non-frozen `pnpm install` re-resolved it — latterly to `0.1.1-canary.beta.0` — and the lockfile drifted with no manifest change to blame. Restoring the lockfile alone does not hold; reinstalling brought all nine canary entries back. The pin collapses **8 resolved versions to 1** and makes the lockfile net smaller. It belongs in `package.json`, not `pnpm-workspace.yaml`: `overrides` there is silently ignored (the entry never reaches the lockfile's own `overrides:` block) even though `onlyBuiltDependencies` in that same file is honoured.

## Scope note

An earlier revision of this branch also carried a second seam — the same harness handed to `BuiltInAgent`'s tanstack stream factory as its own agent slot, with a page at `/banking/deep-work` — so the streaming-seam tradeoff could be judged side by side. It has been **removed from this PR** to keep the change reviewable, and `EXPENSE_HARNESS_MODE` narrowed from `off|tool|factory|both` accordingly. A stale `.env` carrying `factory` or `both` now throws, deliberately: verified that `pnpm build` fails at `/api/copilotkit/[[...slug]]` with the retired value and passes once corrected.

That work is preserved in **#6565**, together with an engine abstraction that runs the same harness on Claude Code and recovers the real `web_search` query (codex's adapter synthesises arguments it never receives, so every observed call arrives as `{"query":""}`).

## Reskin-skill impact

One update, in-PR: `SKILL.md`'s warning about `agentRegistry` having no test guard. It was edited earlier in this branch because the removed second arm shipped a test that imported the map; with that arm gone the grep is empty again and the warning is restored to match.

## Plan and ledger

`examples/showcases/reskinnable-demo/docs/superpowers/plans/2026-08-14-banking-harness-both-arms.md`. The SDD ledger records 31 rulings with costs-if-wrong, both gate measurements, and every deferred minor — including the 16 places the plan was wrong and the tree was right.

https://claude.ai/code/session_01Tp3i7qBNzWC9xVzsaz5KTZ


mxmzb:
Heads up before this gets reviewed too hard: it's green and I think it's mergeable, but it's a **presenter-laptop feature, not something we can deploy anywhere**. Wanted to write down why, and what our options are, because a couple of these are product calls rather than eng calls.

**The situation today.** The harness shells out to the `codex` CLI on whatever machine runs the app. Three consequences:

1. The binary has to already be installed on that host — it's not an npm dependency, so `pnpm install` doesn't put it there. A deploy without it boots totally fine and only dies when someone actually clicks the pill.
2. It authenticates through `codex login`, which means it bills **whoever's ChatGPT account did that login**, not our API key. I found this the hard way — the demo burned my personal Codex quota while the company platform credits sat there unused, because they're two completely different ledgers and nothing in the app says so.
3. It runs on the host with no isolation, with file writes and network access on. That's fine on a laptop and pretty bad on a shared server.

Four ways forward:

---

**1. Ship Codex with the app and run it in a Docker sandbox**

The CLI is on npm (`@openai/codex`) with real Linux builds, so we can just depend on it instead of requiring a host install. Auth switches to our API key (`codex login --with-api-key` at container startup), which puts it on the company account. And there's a Docker sandbox provider in the same family we're already using, so swapping isolation in is a config change rather than a rewrite.

*What's blocking:* running Docker inside our container, or mounting the host's Docker socket into it. The socket route is the easy one and it's also a privilege-escalation path, so someone has to actually own that decision. Also: no auth on the demo today, and each run costs minutes of tokens, so anyone who finds the URL could loop it on our key.

*Why I like it:* smallest change by a mile, nothing gets rewritten, and it's the only option that keeps the "this is literally Codex running inside your product" story — which was the whole pitch.

*Why it might not be enough:* see option 4. Even done perfectly, only one person can use it at a time.

---

**2. Anthropic's Managed Agents**

Anthropic runs the agent loop *and* hosts a sandboxed container per session. No binary to ship, no sandbox infra to own, API key auth, and it's multi-user by construction.

*What's blocking:* we have no integration path for it. We've got paved paths for LangGraph and the Claude Agent SDK, but Managed Agents is a different surface and the wiring would be ours to build.

*Pros:* it deletes the three problems above outright rather than solving them.

*Cons:* custom integration work, and it ties the demo to Anthropic's hosting. Also means the demo is no longer "a coding agent in your app" so much as "a call to someone else's hosted agent" — which may or may not matter depending who we're showing it to.

---

**3. Use LangChain as the coding harness**

LangChain ships a harness architecture (`deepagents` — planning, filesystem, shell, subagents) where the sandbox is a pluggable backend, including hosted/isolated ones. And we already integrate LangGraph properly: it's a dependency of both `@copilotkit/runtime` and `@copilotkit/sdk-js` with six worked examples in the repo.

*What's blocking:* it's a rewrite of the harness rather than a swap, and harness quality becomes ours to tune — prompt, tool selection, compaction, all of it. Codex is a heavily tuned product; this is an architecture we'd configure.

*Pros:* every deployment problem disappears at once. No host binary, API key auth, isolation is a config choice, and per-run state means it works for many users. It's also the most paved path we have.

*Cons:* we lose "it's actually Codex." If the demo's punch comes from people recognising the thing in the window, this trades that away.

---

**4. Change how the run reaches the screen**

Bit more context on this one since it's not visible in the diff. Right now the harness is a **tool** that banking's normal agent calls. Tools return one value when they finish, so there's no way for minutes of progress to ride the normal chat stream — it goes out a separate side channel that the console subscribes to. That channel has a **fixed id**, and there's a single in-flight slot per server, because a tool call gives the browser no per-run id to listen on until it's already over.

Two things fall out of that, and both are structural rather than sloppy:

- **Two people can't use it at once.** The second person's click aborts the first person's run.
- **A refresh loses everything.** The run replays as one collapsed tool chip with none of the journey.

The alternative is to give the harness **its own agent** instead of making it a tool. Then its thinking, tool calls and result flow through the normal thread like any other conversation — survives a refresh, no shared channel, and it works for many people at once because every run already has its own identity.

*What's blocking — and this is the real cost:* banking would become the only skin with a second agent and its own surface, which puts it out of step with every other skin in the demo. Either we accept that the reskinnable demo is no longer uniform (banking behaves differently under this mode), or we push the concept through **all** the skins — which means deciding whether every skin gets a long-running agent slot, and that reaches into the skin contract, the registries and routing. That's a much bigger conversation than this PR.

I did build this and it works; it's parked on #6565 rather than deleted, along with a version that runs the same harness on Claude Code (which as a side effect fixes the empty `web_search` query you'd otherwise see in the console).

---

**How they fit together:** 1 and 4 are on different axes and combine — 1 is *where the harness runs*, 4 is *how its output reaches the browser*. Realistically a genuinely deployable version is 1 + 4. Options 2 and 3 each solve both at once, but 3 costs us the pitch and 2 costs us an integration we don't have.

None of this blocks merging as-is, as long as we're clear this is a laptop demo for now.

Issue #6090, opened 2026-07-21
Custom thread management


JORDAAAN1495:
We're looking at CopilotKit, and wanted to understand how you would recommend to do thread management / thread persistence without using the intelligence platform. If we wanted to roll our own solution for threads by our own backend, is there a spec of required endpoints which is required? Any information you could share here would be greatly appreciated.

This is some what related to https://github.com/CopilotKit/CopilotKit/issues/5434

Have revirewed all the existing documentation and can't see anything related to this.

NathanTarbert:
Hey @JORDAAAN1495, thanks for the question. I would love to know what requirements your team has, or what you are hoping to achieve by rolling your own backend? 

To answer your question first: no, there's no spec of required endpoints for a self-hosted thread store, and that's intentional rather than a gap. 

Thread persistence in CopilotKit isn't a thin REST contract you implement against. The runtime streams every event (messages, tool calls, state) into a durable, server-side event log, and the replay, cross-device resume, realtime metadata sync, and locking are all built on top of that stored stream. There's no small set of endpoints that reproduces it; the persistence layer is the platform. 
So the split we draw is: build the UI yourself (headless threads - fully supported), but don't rebuild the backend.

We've put a lot of work into making both best-in-class with minimal setup, and self-hosting means you keep full ownership of your data, if that's a concern for you.

And, it's worth noting that CopilotKit Enterprise Intelligence is available both as a managed cloud service and as a self-hosted option.

Intelligence threads support everything your agent does:
→ Every kind of generative UI, user file uploads, and (soon) voice transcripts
→ Realtime multicast across devices
→ Automatic resumption on page refresh or network disconnection, even mid-run
→ Automatic migrations as AG-UI evolves and new generative UI variants land, so your thread history never gets lost or corrupted on upgrade
→ Your data stays yours to query - because history sits in your Postgres, you can run your own analytics and reporting across users and conversations with the tools your team already uses
→ Bring your own database - self-host the platform against your own Postgres, so every thread and event lives inside your VPC and data-residency boundary

If you're curious about the underlying architecture, it's covered [here](https://docs.copilotkit.ai/premium/threads-explained).

You can also import existing conversations from LangChain or ADK and keep them in sync perpetually going forward. So if you're using those native threads for analytics or anything else downstream, that keeps working as-is. It's not an either/or choice, and you're not starting from scratch.

[...]

Issue #5991, opened 2026-07-15
Docs proposal: distinguish MCP Apps from read-only MCP context servers


sarveshsea:
## Documentation gap

CopilotKit's MCP Apps story is strong, but developers evaluating design/repository context servers can easily assume that every useful MCP integration must render an MCP App UI.

## Proposal

Add a short compatibility section and example that distinguishes:

1. MCP Apps that render interactive UI.
2. Read-only MCP context servers that supply compact repository/design context to an agent.
3. Action-oriented MCP servers that mutate external systems and need explicit approval boundaries.

A minimal example could connect a read-only design-context tool, show the structured tool result entering an agent run, and clarify when no MCP App resource is expected.

## Concrete proof surface

https://github.com/sarveshsea/chatbot is a public Vercel Chatbot fork with a deterministic design-context/CI workflow and installable Agent Skill. The focused skill is https://github.com/sarveshsea/remember-design-system.

I am opening an issue first per the contribution guide. If this framing fits the docs roadmap, I can prepare the example in the location maintainers prefer.

Issue #3414, opened 2026-03-11
📚 Documentation: prompt injection firewalls


mdemirci10:
### 💬 Let us know how we can improve our documentation.

We need a documentation or an anwser directly in this issue about implementing firewalls in langgraph agent for prompt injection security.

```
agent = create_agent(
    model=llm,
    state_schema=AgentState,
    system_prompt=f"Hello, you are a helpful assistant."
)

def input_firewall(input_string):
  # TODO: check the input string for security using an LLM.

def output_firewall(input_string):
  # TODO: check the input string for security using ifs.

def dlp_firewall(input_string):
  # TODO: check the input string for security using an LLM.

graph = agent
```

I need to examine the agent input and output for prompt injection security protection in the main.py. How can I wrap the agent with an input firewall, an output firewall and a DLP firewall?

PR #6520, opened 2026-08-17
fix: stale useCoAgent node name, silent content-part drops, and the LangGraph docs gaps

### LINKED ISSUE #1151: 📚 Documentation: Aadd AWS Lambda example for self hosted copilotkitRuntime

Provide an example to make a copilotKitRuntime into an aws lambda behind an aws apigateway.

BenTaylorDev:
Bundles the CopilotKit-side work from **OSS-769**, **OSS-767** (partial), and the unshipped remainder of **OSS-609**.

## OSS-769 — `useCoAgent().nodeName` never updates

`useAgentNodeName` tracked the current node in a ref and returned `nodeNameRef.current`. Mutating a ref schedules no render, so a component reading `useCoAgent().nodeName` kept showing whichever node was current at its last render and never updated on its own — it only appeared to work when something unrelated happened to re-render it.

Backing the value with state fixes it. Five lines, no API change.

**Credit:** this fix was diagnosed and submitted first by @aartisonigra in #4620 (2026-05-03), and this PR landed the same change without attributing it. The merged commit can't be rewritten to add a `Co-Authored-By` trailer, so the credit is recorded here. The shipped version is @aartisonigra's change with two cosmetic follow-ups I'd asked for (dead ref removed, trailing newline) plus the tests the hook was missing.

## OSS-767 (partial) — silent content-part drops (#1748)

`normalizeMessageContent` handles only `text` and `binary` parts; anything else — the `{"type": "image", ...}` case from the report — maps to `""` and is filtered out with no signal, so an agent emitting structured content sees its output vanish silently.

This makes the drop visible, once per unrecognised part type so streaming doesn't flood the log. Deliberately **not** the schema change: carrying structured assistant content needs an `AssistantMessage` decision upstream in `ag-ui`, which stays open on OSS-767.

## OSS-609 — the five docs gaps that never shipped

Gap #2 shipped in #6403; gaps 1, 3, 4, 5 and 6 did not.

| Gap | Where | Closes |
| --- | --- | --- |
| AWS Lambda self-hosting | `docs/deploy/aws-lambda.mdx` | #1151 |
| Per-user thread authorization | `docs/auth.mdx` (new section) | #2241 |
| Thread persistence without the platform | `docs/threads-self-managed.mdx` | #6090 |
| DIY guardrails / DLP | `docs/integrations/langgraph/guardrails.mdx` | #3414 |
| When you need an MCP App | `docs/agentic-protocols/mcp.mdx` (new section) | #5991 |

The Lambda guide leads with the constraint that actually bites — streaming is opt-in on every front door, so a chat runtime deployed with the defaults appears to hang for the whole run and then dumps the reply at once. It documents the Function URL + `RESPONSE_STREAM` path as the default, and API Gateway REST + `responseTransferMode: STREAM` for anyone who needs a REST API in front.

`threads-self-managed` follows the existing shared-snippet pattern with per-framework wrappers, because two nav contracts require it: every authored framework must publish the page, and every React destination must map to an Angular one (`ANGULAR_DOC_REDIRECTS`). All nine wrappers were confirmed necessary by deleting one and watching the suite fail.

## Review corrections

Two blockers from @MikeRyanDev, both verified against primary sources before changing anything.

**API Gateway REST APIs can stream** ([`91b3e632d5`](https://github.com/CopilotKit/CopilotKit/commit/91b3e632d5)). The guide was built on the pre-November-2025 limitation and claimed no API Gateway type supports response streaming, steering readers to a buffered `serverless-http` setup. REST gained it via `responseTransferMode: STREAM`, which also lifts the 10 MB cap and 29-second timeout. REST and HTTP are now split; REST is documented as a streaming front door (payload-format-1.0 event adapter, `AWS_PROXY` integration on the `2021-11-15/.../response-streaming-invocations` URI, CLI/CDK/SAM config), and the buffered fallback is scoped to HTTP APIs and ALB, which still have no streaming path. Added the constraints that actually matter for chat: the 30-second idle timeout on edge-optimized endpoints (5 min Regional), and the console Test tab always buffering so a working config looks broken.

**`identifyUser` is the platform's thread-scoping binding** ([`91b3e632d5`](https://github.com/CopilotKit/CopilotKit/commit/91b3e632d5)). The section told every reader to build an ownership table and enforce it in `onBeforeHandler`. On the Intelligence path the runtime already resolves `identifyUser(request)` server-side and carries that id to the platform; `listThreads` is scoped by user *and* filtered by `agentId`, so the "every user of one project sees that project's threads" claim was wrong. `identifyUser` is now documented as the binding, and the DIY pattern is scoped to SSE runtimes, custom stores, and the local in-memory runner.

**Follow-up correction — two routes are genuinely unscoped** ([`240672ff56`](https://github.com/CopilotKit/CopilotKit/commit/240672ff56)). My rewrite then over-claimed. `handleGetThreadEvents` and `handleGetThreadState` resolve the caller and discard it, and the platform client takes no `userId` on either method (`client.ts:1113`/`1135`) — unlike `getThreadMessages` at `1063`. Both hit project-authenticated `_inspect` endpoints, so any caller `identifyUser` accepts can read the event log and agent state of **any thread in the project** given its id. The blanket guarantee is replaced by a per-route table marking those two explicitly unscoped, plus an `onBeforeHandler` guard narrowed to them.

That is a live gap in shipped runtime code, not a docs error, and it is tracked as **OSS-851** — a platform-side `_inspect` change plus matching runtime/client work and tests, out of scope for a docs PR. The interim callout in `auth.mdx` comes out when OSS-851 lands.

## Not in this PR

**OSS-772** and **OSS-773** are already merged in `oss-path-to-production` (#237, #238). Both are telemetry-sink changes with no CopilotKit-side component. OSS-773's remaining half — re-keying runtime `distinct_id` from email to the Clerk subject — is recorded on the ticket as an open decision, not a task.

## Testing

**OSS-769.** New `use-agent-nodename.test.tsx`, 5 tests. Against unmodified `origin/main`, **4 of 5 fail**:

```
× re-renders consumers on every node transition
× reports 'end' when a run errors
× resets to 'start' when a new run begins
✓ unsubscribes on unmount
× carries the agent, thread, and current node
  Tests  4 failed | 1 passed (5)
```

With the fix: `Tests 5 passed (5)`. These assert only re-render behaviour under normal `act()` flushing — no manufactured intra-batch window.

**Full react-core suite:** `Tests 7 failed | 1496 passed (1503)`. All 7 failures are **pre-existing** `ResizeObserver is not a constructor` under jsdom, confined to `CopilotChatView.pinToSend` and `use-pin-to-send` — neither of which this PR touches.

**Typecheck:** `packages/react-core` → `tsc --noEmit` exit 0, no output.

**OSS-767:** 3 new tests covering the warn, warn-once-per-type, and no-warn-for-supported-types. `src/graphql/message-conversion/` → `Tests 125 passed (125)`.

**Docs:** `showcase/shell-docs` → `Tests 1 failed | 373 passed (374)`. The single failure (`channels-docs > publishes the Channels overview only through provider navigation`) is **pre-existing**; baselining with all changes stashed reproduces it and nothing else. Re-run unchanged after both review-correction commits.

**Review corrections.** The AWS rewrite was checked against the AWS sources rather than written from memory — the REST streaming announcement, `configuration-response-streaming`, `response-transfer-mode` (endpoint-type idle timeouts, unsupported buffered-only features), `response-streaming-lambda-configure` (CLI/OpenAPI shapes), the CFN `Integration` reference, and the CDK `ResponseTransferMode` enum. Two details corrected in passing: `InvokeWithResponseStream` authorizes against plain `lambda:InvokeFunction` (no new grant, contrary to what the streaming URI suggests), and ALB still has no Lambda streaming path.

The auth corrections were verified by reading the handlers and the platform client, not the tests — `resolve-intelligence-user.ts`, `intelligence/threads.ts` (every `resolveIntelligenceUser` call site), and `intelligence-platform/client.ts`. The existing tests assert the `threadId`-only call shape, so they pass under the defect and could not have surfaced it. Also corrected: there is no `threads/delete` route — delete is `DELETE` on `threads/update` (`fetch-handler.ts:606`).

Both edited pages MDX-compile clean, and all inbound `#thread-authorization` anchors still resolve after the h3→h4 demotions.

Two nav tests broke during this work and are fixed rather than papered over — adding a page to the Rich Threads group violated the cross-framework ordering contract and the React→Angular parity contract:

```
src/lib/__tests__/docs-render.test.ts
src/lib/__tests__/angular-docs-content.test.ts
  Test Files  2 passed (2)
       Tests  33 passed (33)
```

All 15 internal links in the new pages resolve against the content tree.

Closes #1151, #2241, #3414, #5991, #6090
Refs #1748, OSS-851

### DIFF
--- a/showcase/shell-docs/src/content/docs/auth.mdx
+++ b/showcase/shell-docs/src/content/docs/auth.mdx
@@ -450,7 +450,7 @@ Verifying *who* the caller is doesn't yet stop them reaching *someone else's con
 
 | Runtime | Who scopes threads to a user |
 | --- | --- |
-| `CopilotRuntime` with `intelligence` | Mostly the runtime, via [`identifyUser`](#the-intelligence-platform-scopes-most-thread-routes) — with two inspector routes you still have to guard. |
+| `CopilotRuntime` with `intelligence` | Mostly the runtime, via [`identifyUser`](#the-intelligence-platform-scopes-most-thread-routes) — with three routes you still have to guard. |
 | Anything else — SSE runtime, custom store, local in-memory runner | You do. See [Scope threads yourself](#scope-threads-yourself). |
 
 ### The Intelligence Platform scopes most thread routes
@@ -470,7 +470,15 @@ const runtime = new CopilotRuntime({
 });
 ```
 
-Every route resolves the caller this way, so an unauthenticated request is rejected everywhere. What differs is whether the resolved id is then *carried to the platform* as a scope:
+<Callout type="warn" title="`identifyUser` is not an authentication gate">
+  Most routes resolve the caller through `identifyUser`, but **not all of them do** — and the
+  ones that don't never invoke your callback at all. Rejecting unauthenticated requests is
+  `onRequest`'s job: it runs before routing, on every route, without exception. Treat
+  `identifyUser` as the thing that *names* an already-authenticated caller, never as the thing
+  that decides whether a caller gets in.
+</Callout>
+
+Where a route does resolve the caller, what matters next is whether that id is actually *carried to the platform* as a scope:
 
 | Route | Scoped to the resolved user? |
 | --- | --- |
@@ -479,34 +487,53 @@ Every route resolves the caller this way, so an unauthenticated request is rejec
 | `threads/messages` | Yes |
 | `threads/update` (rename via `PATCH`, delete via `DELETE`), `threads/archive` | Yes |
 | Thread subscription token | Yes |
-| `threads/events`, `threads/state` | **No** — see the callout below |
+| `threads/events`, `threads/state` | **No** — resolves the caller, then ignores it |
+| `agent/stop` | **No** — never resolves a caller at all |
 
-<Callout type="warn" title="`threads/events` and `threads/state` are not user-scoped">
-  These two routes back the [inspector](/inspector). Both resolve the caller through
-  `identifyUser` and then discard the result, reading the thread by id alone — the runtime
-  calls the platform's project-authenticated `_inspect` endpoints, which take no user
-  parameter. Any caller `identifyUser` accepts can read the full event log and current agent
-  state of any thread in the project, given its `threadId`.
+<Callout type="warn" title="Three routes are not user-scoped">
+  **`threads/events` and `threads/state`** back the [inspector](/inspector). Both resolve the
+  caller and then discard the result, reading the thread by id alone — the runtime calls the
+  platform's project-authenticated `_inspect` endpoints, which take no user parameter. Any
+  caller can read the full event log and current agent state of any thread in the project,
+  given its `threadId`.
 
-  Guard them yourself until the platform scopes them. The `onBeforeHandler` pattern below
-  applies on the Intelligence path too, narrowed to these two routes:
+  **`agent/stop`** never resolves a caller. It goes straight to `runner.stop({ threadId })`,
+  which aborts whichever run the runtime is tracking under that thread id. Any caller who
+  learns an active `threadId` can kill that run mid-flight.
+
+  Guard all three yourself. The `onBeforeHandler` pattern below applies on the Intelligence
+  path too, narrowed to these routes:
 
   ```ts
-  onBeforeHandler: async ({ route }) => {
-    if (route.method !== "threads/events" && route.method !== "threads/state") return;
+  onBeforeHandler: async ({ request, route }) => {
+    // Switch rather than an array `includes`, so `route` narrows and
+    // `route.threadId` type-checks — all three variants carry one.
+    switch (route.method) {
+      case "threads/events":
+      case "threads/state":
+      case "agent/stop":
+        break;
+      default:
+        return;
+    }
 
-    // Neither route is scoped platform-side. Reject outright unless you keep an
-    // ownership record of your own to check `route.threadId` against.
-    throw new Response("Not found", { status: 404 });
+    // None of these is scoped platform-side; check your own ownership record.
+    const user = await verifyRequest(request);
+    if (!(await userOwnsThread(user.id, route.threadId))) {
+      throw new Response("Not found", { status: 404 });
+    }
   },
   ```
+
+  Without an ownership record of your own, reject these routes outright rather than leaving
+  them open.
 </Callout>
 
 For everything in the Yes rows there is no ownership table to build. What you own is `identifyUser` itself:
 
 - **Derive the id from a server-verified credential.** The callback receives the raw `Request`; whatever it returns is trusted from there on. Reading a user id straight out of a header or request body hands every caller the ability to name themselves.
 - **Return a stable id.** It is the key threads hang off. Change it — swapping an email for a subject claim, say — and that user's existing threads stop resolving.
-- **Send the `401` from `onRequest`.** An `identifyUser` that throws surfaces as a `500`, and returning a malformed id gives a `400` — neither is what an unauthenticated caller should see. Do the authenticate-or-reject step in `onRequest`, which can throw a `Response` directly, and keep the throw inside `identifyUser` as a backstop for the case that slips through.
+- **Send the `401` from `onRequest`.** Beyond the status codes being wrong — an `identifyUser` that throws surfaces as a `500`, a malformed id as a `400` — routes like `agent/stop` never call it, so a check placed only here isn't reached on every request. Authenticate in `onRequest`, which runs on every route and can throw a `Response` directly, and keep the throw inside `identifyUser` as a backstop.
 
 <Callout type="warn">
   A static `identifyUser` value is suitable only for a single-user demo. In a multi-user
@@ -518,7 +545,7 @@ See [Scope Rich Threads to the signed-in user](/threads-lifecycle#scope-rich-thr
 
 ### Scope threads yourself
 
-Without the Intelligence Platform there is no server-side binding between a `threadId` and a user. A `threadId` is just an opaque id travelling in a request: if user A learns user B's, every thread route accepts it. The rest of this section is the pattern for a custom store, a plain SSE runtime, or the local in-memory runner — and it's also what you narrow to `threads/events` and `threads/state` if you *are* on the platform.
+Without the Intelligence Platform there is no server-side binding between a `threadId` and a user. A `threadId` is just an opaque id travelling in a request: if user A learns user B's, every thread route accepts it. The rest of this section is the pattern for a custom store, a plain SSE runtime, or the local in-memory runner — and it's also what you narrow to `threads/events`, `threads/state`, and `agent/stop` if you *are* on the platform.
 
 #### Own the mapping
 
@@ -597,7 +624,8 @@ Filter server-side. Hiding rows in the UI leaves the underlying route open.
 
 - **Always validate** the token on the backend. Never trust the frontend's claim.
 - **Scope every read and write** to the resolved user. Auth context only matters if you actually use it to filter data.
-- **Bind threads to a user.** On the Intelligence Platform that means a correct `identifyUser`, plus your own guard on `threads/events` and `threads/state`; off it, an ownership check on every thread route rather than only at login. See [Thread authorization](#thread-authorization).
+- **Authenticate in `onRequest`.** It is the only hook that runs on every route. `identifyUser` names a caller; it does not gate one, and some routes never invoke it.
+- **Bind threads to a user.** On the Intelligence Platform that means a correct `identifyUser`, plus your own guard on `threads/events`, `threads/state`, and `agent/stop`; off it, an ownership check on every thread route rather than only at login. See [Thread authorization](#thread-authorization).
 - **Don't log raw tokens.** Log the resolved user id (or `anonymous`) instead.
 - **Use HTTPS in production.** The Bearer token is sensitive.
 - **Refresh strategy.** Your frontend is responsible for rotating expired tokens before they reach the agent. CopilotKit doesn't refresh on your behalf.
--- a/showcase/shell-docs/src/content/docs/auth.mdx
+++ b/showcase/shell-docs/src/content/docs/auth.mdx
@@ -450,12 +450,12 @@ Verifying *who* the caller is doesn't yet stop them reaching *someone else's con
 
 | Runtime | Who scopes threads to a user |
 | --- | --- |
-| `CopilotRuntime` with `intelligence` | The runtime does, via [`identifyUser`](#the-intelligence-platform-scopes-threads-for-you). |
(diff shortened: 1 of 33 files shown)


showcase/shell-docs/src/content/docs/deploy/aws-lambda.mdx
```diff
@@ -0,0 +1,303 @@
...
+icon: "lucide/Cloud"
+description: Run the CopilotKit runtime on AWS Lambda, including the response-streaming setup a chat UI requires.
+---
+
+The CopilotKit runtime is a Fetch handler, so it runs on Lambda without a server. The one thing that needs care is **streaming**: chat responses are served as Server-Sent Events, and most Lambda front doors buffer the whole response before returning it.
+
+## Start here: pick the right front door
+
+<Callout type="warn" title="API Gateway cannot stream">
+  Neither API Gateway REST APIs nor HTTP APIs support response streaming. They buffer the
```
MikeRyanDev: Blocker: this is no longer true. API Gateway REST APIs added response streaming in November 2025, and the current Lambda docs explicitly support streaming through an API Gateway proxy integration using InvokeWithResponseStream. REST streaming also supports timeouts up to 15 minutes and payloads larger than 10 MB, so the table and the entire “API Gateway (no streaming)” section (including its 10 MB/29-second constraints) now recommend the wrong architecture. HTTP APIs are a separate case. Please split REST from HTTP and document the REST streaming configuration. AWS sources: https://aws.amazon.com/about-aws/whats-new/2025/11/api-gateway-response-streaming-rest-apis/ and https://docs.aws.amazon.com/lambda/latest/dg/configuration-response-streaming.html

showcase/shell-docs/src/content/docs/auth.mdx
```diff
@@ -444,10 +444,89 @@ def delete_record(record_id: str, *, user: User):
...
+<Callout type="warn" title="A thread id is not a secret">
+  Treat `threadId` as a public identifier, like a database primary key in a URL. It travels
+  through the browser, appears in logs, and is trivially enumerable if you mint sequential
+  ids. Authorization has to be an explicit ownership check — never "they knew the id, so they
+  must own it". Minting UUIDs makes guessing impractical but is not itself a control.
+</Callout>
+
+### Filter the thread list
+
+`threads/list` has no `threadId` to check — it returns whatever the configured store holds. On the Enterprise Intelligence Platform, threads are scoped to a project, so every user of one project sees that project's threads. If your product needs per-user thread lists, either give each user their own project scope, or build the list from your own ownership table and drive the chat with the selected id rather than rendering the platform list directly.
```
MikeRyanDev: Blocker: this is not the current Intelligence contract. The runtime resolves identifyUser(request) server-side, then passes that resolved user id to listThreads, message/event/state reads, mutations, and the thread subscription token; list is also filtered by agentId. In other words, the platform path is already user-scoped by the runtime rather than exposing every project thread to every user. Please scope the DIY ownership-table/onBeforeHandler pattern to non-Intelligence or custom stores, and document identifyUser as the platform binding instead of telling Intelligence users to reimplement ownership. Relevant paths: packages/runtime/src/v2/runtime/handlers/shared/resolve-intelligence-user.ts and packages/runtime/src/v2/runtime/handlers/intelligence/threads.ts.

MikeRyanDev (changes requested):
Requesting changes for two material documentation correctness issues. The runtime changes and their tests look sound, but the new AWS guide is built around an obsolete platform limitation, and the thread-authorization section misstates the current Intelligence runtime ownership contract.

showcase/shell-docs/src/content/docs/deploy/aws-lambda.mdx
BenTaylorDev: Correct, and the whole section was built on the pre-November-2025 limitation. Rewritten in 91b3e632d5: REST and HTTP are now split, with REST documented as a streaming front door — `responseTransferMode: STREAM` on an `AWS_PROXY` integration pointed at `2021-11-15/.../response-streaming-invocations`, plus a payload-format-1.0 event adapter since REST proxy does not send the 2.0 shape a Function URL does. The buffered `serverless-http` section is now scoped to HTTP APIs and ALB, which still have no streaming path.

Added the constraints that matter for chat specifically: the 30-second idle timeout on edge-optimized endpoints (5 minutes Regional/private), the integration timeout going to 15 minutes, and the console Test tab always buffering so a working config looks broken there.

showcase/shell-docs/src/content/docs/auth.mdx
BenTaylorDev: Correct. Verified in `resolve-intelligence-user.ts` and `intelligence/threads.ts`: `identifyUser` is required for the web surface, resolved server-side per request, and the id flows to run, connect, the message/event/state reads, the mutations, and the subscription join token, with `listThreads` scoped by user and filtered by `agentId`. The "every user of one project sees that project's threads" line was simply wrong.

Rewritten in 91b3e632d5 to lead with a runtime routing table, document `identifyUser` as the platform binding (linking to the canonical `/threads-lifecycle#scope-rich-threads-to-the-signed-in-user` rather than duplicating it), and scope the ownership-table/`onBeforeHandler` pattern to SSE runtimes, custom stores, and the local in-memory runner — the one path where `listThreads` really does filter by `agentId` alone.

showcase/shell-docs/src/content/docs/auth.mdx
```diff
@@ -444,10 +444,130 @@ def delete_record(record_id: str, *, user: User):
...
+  identifyUser: async (request) => {
+    const session = await verifyAppSession(request); // Your server-side auth.
+    if (!session?.user) throw new Error("Unauthorized"); // Backstop; see below.
+
+    return { id: session.user.id, name: session.user.name };
+  },
+});
+```
+
+That resolved id is applied to `agent/run`, `agent/connect`, the thread message, event, and state reads, `threads/update`, `threads/archive`, `threads/delete`, and the token minted for the thread subscription. `threads/list` is scoped by the resolved user *and* filtered by `agentId`, so `useThreads` returns the caller's threads rather than every thread in the project.
```
MikeRyanDev: Blocker — correction to my earlier review: identifyUser does not currently scope the event and state reads. handleGetThreadEvents and handleGetThreadState resolve the user, but then discard it and call getThreadEvents({ threadId }) / getThreadState({ threadId }); the client hits project-authenticated /api/_inspect/threads/:id/{events,state} endpoints with no userId. The tests explicitly assert that threadId-only call shape. As written, this security guide tells Intelligence users there is no ownership check to build while a caller who learns another user thread ID can still reach these two routes. Please either carry the resolved user through and enforce ownership platform-side (with tests), or explicitly exclude/guard these inspector routes in the documented onBeforeHandler pattern. The blanket every thread-bearing route / no ownership table guarantee is not safe to publish yet.

MikeRyanDev (changes requested):
The API Gateway correction is accurate and resolves the AWS blocker. One security-critical correctness issue remains in the revised thread-authorization guidance: the current event/state read path resolves the user but does not pass that identity to the Intelligence platform. The docs must not promise complete per-user scoping until those routes are actually scoped or explicitly guarded.

showcase/shell-docs/src/content/docs/auth.mdx
BenTaylorDev: Confirmed, and the guarantee was mine to get right — good catch before it published. Verified: `handleGetThreadEvents` and `handleGetThreadState` resolve the caller and then drop it, and the platform client takes no `userId` on either method (`client.ts:1113`/`1135`), unlike `getThreadMessages` at `1063` which appends it as a query param. So the `_inspect` endpoints are project-scoped and the exposure is broader than one leaked id — any caller `identifyUser` accepts can read any thread in the project given its id.

Fixed in 240672ff56 with your second option: the blanket claim is replaced by a per-route table marking `threads/events` and `threads/state` explicitly unscoped, plus a warn callout carrying the `onBeforeHandler` guard narrowed to those two routes. I did not take the first option — carrying the user through needs a platform-side `_inspect` change plus matching runtime/client work and tests, which is a runtime security fix rather than something to land inside a docs PR. Happy to open a ticket for it if one does not exist.

One correction while verifying: there is no `threads/delete` route (delete is `DELETE` on `threads/update`), and the table reflects that.

showcase/shell-docs/src/content/docs/auth.mdx
```diff
@@ -444,10 +444,160 @@ def delete_record(record_id: str, *, user: User):
...
+  calls the platform's project-authenticated `_inspect` endpoints, which take no user
+  parameter. Any caller `identifyUser` accepts can read the full event log and current agent
+  state of any thread in the project, given its `threadId`.
+
+  Guard them yourself until the platform scopes them. The `onBeforeHandler` pattern below
+  applies on the Intelligence path too, narrowed to these two routes:
+
+  ```ts
+  onBeforeHandler: async ({ route }) => {
+    if (route.method !== "threads/events" && route.method !== "threads/state") return;
```
MikeRyanDev: Blocker: the unscoped-route list is still incomplete. agent/stop never calls identifyUser or resolveIntelligenceUser. handleStopAgent goes directly to runtime.runner.stop({ threadId }), and IntelligenceAgentRunner.stop aborts its locally tracked run by threadId alone. Therefore any authenticated user who learns another user thread ID can stop that active run; if an application incorrectly relies on identifyUser as its only authentication backstop, this route does not invoke it at all. That also makes the statement that every route resolves the caller false. Please add agent/stop to the No row and to the targeted ownership guard (or scope it in the runtime), and clarify that onRequest is the universal authentication gate. Guarding only threads/events and threads/state leaves a real cross-user operation exposed.

MikeRyanDev (changes requested):
The event/state correction is accurate, but the revised security guidance still omits another user-unscoped thread operation: agent/stop. Because the documented guard covers only event and state reads, one caller can still stop another user active run by thread ID.

MikeRyanDev (approved):
Re-reviewed the rebased exact head. All previous blockers are resolved: the AWS REST streaming guidance is current, Intelligence thread scoping is described accurately, the three unscoped thread operations (events, state, and stop) are explicitly guarded, and onRequest is correctly identified as the universal authentication gate. The rebased commit series is otherwise semantically unchanged. Focused docs tests and typecheck pass, and all current-head checks are successful or skipped.

BenTaylorDev:
Retroactive correction to this PR's description: the `useAgentNodeName` fix under OSS-769 was diagnosed and submitted first by @aartisonigra in #4620 on 2026-05-03. I reviewed that PR, asked for two cosmetic changes, and then landed the same five-line change here without referencing it or adding a `Co-Authored-By` trailer. The body now credits it; noting the edit here so the change isn't silent.

PR #6523, opened 2026-08-17
fix(channels-slack): deliver a message click's state.values to the handler

lukasmoschitz:
## OSS-846 + OSS-848

When a user clicks a button in a Slack message, Slack's `block_actions` payload includes `state.values` — the current value of every stateful element on that message. It never reached the developer's handler: `ctx.values` was always `{}` for a message click.

This PR delivers it, **and** makes the delivered keys author-nameable (OSS-848), which is what makes the field usable at all. Its managed-path twin is Intelligence#866; the two only make sense as a pair, and this side owns the shape both must agree on.

## Where the data actually stopped

1. **`channels-slack` `decodeInteraction`** (`packages/channels-slack/src/interaction.ts`) — **the defect.** It built its event from `actions[0]` alone and never read `body.state`. `values` was populated only by `decodeViewSubmission` for a modal `view_submission`.
2. **`channels-core`** (`create-channel.ts:1038`, `values: evt.values ?? {}`) — **not a bug.** A faithful passthrough; `InteractionEvent.values` and `InteractionContext.values` already existed, the latter documented as "Submitted input id → value". The contract was designed for this and only the Slack decoder never filled it.
3. **Slack renderer** (`render/block-kit.ts`) — **now fixed here** (OSS-848). Previously `<Input>` / `<Select multi>` emitted no `block_id`, so Slack minted a random one per render. See "Author-named keys" below.
4. **Managed path (Intelligence)** — out of this repo. The SDK already whitelists and forwards `values`; Intelligence#866 populates it. On a managed deployment this fix is inert until #866 ships.

### Shape — aligned with the managed normalizer

`state.values` flattens to the same flat `fieldId → value` map modal submissions already produce, keyed by block id. Two rules now match Intelligence#866 exactly, so `Object.keys`, `in` and `Object.entries` agree on managed and self-hosted:

- **A field whose element reports no value is omitted**, not assigned `undefined`. The managed path carries `values` over the wire as JSON, which cannot express `undefined`, so keeping the key would promise a field the managed runtime never delivers.
- **Every guard in the value ladder is on the value's type**, not `!== undefined`. Slack sends an explicit `null` for an untouched picker (`selected_date: null`, `selected_user: null`), and `null !== undefined` — so the previous guards returned a literal `null`, which *does* survive JSON all the way to the handler.

`values` is omitted entirely when that leaves the map empty (the Teams decoder's convention).

### Free text is no longer JSON-parsed — a modal-path behaviour change, called out

Sharing the value ladder with `flattenViewValues` made it JSON-parse text the user typed. `decodeViewSubmission` has shipped for several releases, so this would have **retyped fields for existing modal handlers**: an order id typed as `1234` arriving as the number `1234`, and `values.orderId.trim()` throwing.

The ladder now discriminates on the element type. Author-encoded values (a button's `value`, an option's `selected_option(s).value`) are still JSON-parsed; `plain_text_input`, `rich_text_input`, `email_text_input`, `url_text_input` and `number_input` content is handed back as the verbatim string it always was.

**Net effect on the shipped modal path:** free text is unchanged; the eight element families the old `value ?? selected_option.value` reader silently dropped (multi-selects, user/channel/conversation pickers, date, time, datetime, rich text) now arrive populated instead of `undefined`; and an untouched field is now omitted rather than present-and-`undefined`.

### Author-named keys (OSS-848), folded in

The key this PR delivers is the **block id**, and the renderer did not emit one on a message — so Slack minted a random id per render. Teams honours `props.name`, so `<Input name="root_cause">` yielded `ctx.values.root_cause` on Teams and `ctx.values["<random>"]` on Slack. That made the field unusable for the cross-platform vocabulary, which is most of its point, so the renderer change lands here rather than as a follow-up.

`block_id` now comes from `props.name`, mirroring the Teams allocator arm-for-arm — same precedence (`name` → minted handler id → positional `input_1`), same reserved names, same `_1` de-duplication — so both platforms mint the **same key for the same tree**. `<Select multi>` (already routed through an `input` block) is covered. A **named** `<Select>` inside `<Actions>` gets an `actions` block of its own, because a block carries one `block_id` and grouping would make the name unrepresentable; unnamed selects stay grouped exactly as before.

### Multiple stateful elements per block

Both flatteners took `inner[blockId] ?? Object.values(inner)[0]`, documented as "the block's first (in practice only) element". That does not hold: `<Actions>` packs up to `SLACK_LIMITS.actionsElements` elements into one block, so a second `<Select>` was dropped silently.

Every element is now delivered. The element the block id names (and any sole element) is keyed by **block id** — so modals, and every shape today's renderer emits, decode exactly as before — and any further element in the same block is keyed by its own **action id**, the only identifier left that distinguishes it. An existing key is never clobbered.

## What is settled and what is not

**Documented:** Slack's `block_actions` reference describes `state` as "A property including all stateful elements, not just input blocks", so elements in `actions` / `section` blocks are expected to appear in `state.values`.

**Unconfirmed:** that has not been observed against a real payload. Slack live testing was not available for this work — the only live deployment is managed, and dev does not carry the Intelligence fix, so no `state.values` reaches anything. **No claim here rests on a live measurement; the tests are unit tests.**

The code is written to be correct either way: if Slack does report them, multi-element blocks are delivered in full instead of silently truncated to one; if it does not, the multi-element branch is simply never taken and nothing else changes. A real click on a message with two selects in one `<Actions>` would settle it.

## Compatibility

Additive, but the earlier framing was too broad. **"Decodes byte-identically" holds only for a message carrying no stateful element at all.** If Slack does report elements outside `input` blocks, any existing message rendering a `<Select>`, a picker or an `<Input>` alongside a button now newly carries a `values` object on a button click where the field was previously absent. Still additive — nothing that existed changes value — but it is not "no change" for those messages.

Otherwise: no production code in the repo reads `ctx.values`; no consumer-side emptiness check, exhaustive-key assumption or spread-merge exists; `values` is not consulted by the HITL `awaitChoice` waiter, which reads only `evt.value`.

The renderer change is additive too — no `block_id` was emitted on these blocks before — but it does change render output, and a message that previously relied on a random block id now gets a stable one.

## Tests

Every new rule is pinned by a test **verified to fail when the rule is removed** (mutation applied, suite run, rule restored):

| Rule removed | Tests that go red |
| --- | --- |
| omit a field that reports no value | 13 |
| type-guard the scalar arms (so `null` leaks) | 9 |
| free text exempt from JSON parsing | 7 |
| deliver every element in a block | 2 |
| emit `block_id` from `props.name` | 6 |
| give a named `<Select>` its own block | 2 |

The earlier omission tests were **vacuous** and are fixed: `expect(evt.values).toEqual({ prio: undefined })` passes whether or not the key exists, because vitest's `toEqual` ignores undefined-valued keys. Those assertions now use `toStrictEqual` / `not.toHaveProperty` / `Object.keys`.

The OSS-848 coverage deliberately goes **through the renderer** (`render → Slack state.values → decodeInteraction → ctx.values`). Hand-written payloads cannot catch that class of bug: they invent block ids that are already meaningful names, a shape only a hand-authored native block produces.

## Verification

`oxfmt --check`, `oxlint` (0 errors), `check-types`, `test` and `build` all pass for `@copilotkit/channels-slack` and its dependents (`channels-core`, `channels-ui`, `channels-teams`, `channels-intelligence`, `channels`). 443 tests pass in `channels-slack`.

## Note for Intelligence#866

`readInteractionValue` there inherits the same free-text coercion (finding 2) and needs the matching type discrimination, or a `plain_text_input` will decode differently on the two halves. The omission and `null` rules already match #866 — this PR moved to them.

### DIFF
--- a/packages/channels-slack/src/__tests__/views.test.ts
+++ b/packages/channels-slack/src/__tests__/views.test.ts
@@ -100,20 +100,72 @@ describe("decodeViewSubmission", () => {
     });
   });
 
-  it("reports a select with nothing chosen as undefined, not a crash", () => {
-    // Slack sends `selected_option: null` for an untouched select.
+  // A field that reports nothing is OMITTED, never present-and-undefined. The
+  // managed path carries `values` as JSON, which cannot express `undefined`, so
+  // a key kept here would promise a field the managed runtime never delivers —
+  // `Object.keys`, `in` and `Object.entries` have to agree on both.
+  // `toStrictEqual` is load-bearing: `toEqual` ignores undefined-valued keys and
+  // would pass whether the key exists or not.
+  it.each([
+    ["an untouched select", { type: "static_select", selected_option: null }],
+    ["an untouched datepicker", { type: "datepicker", selected_date: null }],
+    ["an untouched user picker", { type: "users_select", selected_user: null }],
+    ["an element reporting nothing", { type: "plain_text_input" }],
+  ])("omits the key for %s", (_label, element) => {
+    const evt = decodeViewSubmission(
+      { callback_id: "triage", state: { values: { prio: { prio: element } } } },
+      { id: "U1", kind: "human" },
+    );
+    expect(evt.values).toStrictEqual({});
+    expect(evt.values).not.toHaveProperty("prio");
+    expect(Object.keys(evt.values)).toEqual([]);
+  });
+
+  // `decodeViewSubmission` has shipped for several releases handing free text
+  // back verbatim; sharing the value ladder with the message-click path must not
+  // retype it. A handler doing `values.orderId.trim()` would throw.
+  it("hands free text back verbatim, never JSON-parsed", () => {
     const evt = decodeViewSubmission(
       {
-        callback_id: "triage",
+        callback_id: "order",
         state: {
           values: {
-            prio: { prio: { type: "static_select", selected_option: null } },
+            order_id: { order_id: { type: "plain_text_input", value: "1234" } },
+            note: { note: { type: "plain_text_input", value: "true" } },
+            qty: { qty: { type: "number_input", value: "7" } },
+            blank: { blank: { type: "plain_text_input", value: "null" } },
           },
         },
       },
       { id: "U1", kind: "human" },
     );
-    expect(evt.values).toEqual({ prio: undefined });
+    expect(evt.values).toStrictEqual({
+      order_id: "1234",
+      note: "true",
+      qty: "7",
+      blank: "null",
+    });
+  });
+
+  // The author encoded these, so they still round-trip through JSON.
+  it("still JSON-parses an author-encoded option value", () => {
+    const evt = decodeViewSubmission(
+      {
+        callback_id: "order",
+        state: {
+          values: {
+            plan: {
+              plan: {
+                type: "static_select",
+                selected_option: { value: '{"id":7}' },
+              },
+            },
+          },
+        },
+      },
+      { id: "U1", kind: "human" },
+    );
+    expect(evt.values).toStrictEqual({ plan: { id: 7 } });
   });
 
   it("decodes a __cpk envelope into conversationKey + replyTarget and restores pm", () => {
--- a/packages/channels-slack/src/__tests__/views.test.ts
+++ b/packages/channels-slack/src/__tests__/views.test.ts
@@ -38,6 +38,84 @@ describe("decodeViewSubmission", () => {
     });
   });
 
+  // OSS-846: the message-click path now shares this flattening. The helper it
+  // shares used to read `value ?? selected_option.value` only, so every element
+  // family that reports under a different key came back `undefined` — silently,
+  // in modal submissions too.
+  it("parses element families the old value ?? selected_option reader lost", () => {
+    const evt = decodeViewSubmission(
+      {
+        callback_id: "triage",
+        state: {
+          values: {
+            services: {
+              services: {
+                type: "multi_static_select",
+                selected_options: [{ value: "payments" }, { value: "search" }],
+              },
+            },
+            responders: {
+              responders: {
+                type: "multi_users_select",
+                selected_users: ["U1", "U2"],
+              },
+            },
+            owner: { owner: { type: "users_select", selected_user: "U9" } },
+            escalate_in: {
+              escalate_in: { type: "channels_select", selected_channel: "C7" },
+            },
+            notify: {
+              notify: {
+                type: "multi_conversations_select",
+                selected_conversations: ["C1"],
+              },
+            },
+            target_date: {
+              target_date: { type: "datepicker", selected_date: "2026-08-20" },
+            },
+            cutover_at: {
+              cutover_at: { type: "timepicker", selected_time: "14:32" },
+            },
+            deadline: {
+              deadline: {
+                type: "datetimepicker",
+                selected_date_time: 1786451118,
+              },
+            },
+          },
+        },
+      },
+      { id: "U1", kind: "human" },
+    );
+
+    expect(evt.values).toEqual({
+      services: ["payments", "search"],
+      responders: ["U1", "U2"],
+      owner: "U9",
+      escalate_in: "C7",
+      notify: ["C1"],
+      target_date: "2026-08-20",
+      cutover_at: "14:32",
+      deadline: 1786451118,
+    });
+  });
+
+  it("reports a select with nothing chosen as undefined, not a crash", () => {
+    // Slack sends `selected_option: null` for an untouched select.
+    const evt = decodeViewSubmission(
+      {
+        callback_id: "triage",
+        state: {
+          values: {
+            prio: { prio: { type: "static_select", selected_option: null } },
+          },
+        },
+      },
+      { id: "U1", kind: "human" },
+    );
+    expect(evt.values).toEqual({ prio: undefined });
+  });
+
   it("decodes a __cpk envelope into conversationKey + replyTarget and restores pm", () => {
     const evt = decodeViewSubmission(
       {
--- a/packages/channels-slack/src/interaction.test.ts
+++ b/packages/channels-slack/src/interaction.test.ts
@@ -145,6 +145,119 @@ describe("decodeInteraction", () => {
     expect(evt!.value).toBeUndefined();
   });
 
+  // OSS-846: Slack sends `state.values` — the current value of every input block
+  // on the clicked message — alongside a block_actions click. The decoder built
+  // its event from `actions[0]` alone and never read it, so a "fill the fields,
+  // then press the button" message reached the handler with `ctx.values` empty.
+  describe("state.values on a message click (OSS-846)", () => {
+    /** A button click on a message hosting five named input blocks. */
+    const clickWithForm = () =>
+      decodeInteraction({
+        type: "block_actions",
+        user: { id: "U1", name: "Ana" },
+        channel: { id: "C1" },
+        message: { ts: "111.1", thread_ts: "100.0" },
+        actions: [{ action_id: "ck:submit", value: "report" }],
+        state: {
+          values: {
+            root_cause: {
+              "ck:a": { type: "plain_text_input", value: "pool exhausted" },
+            },
+            severity: {
+              "ck:b": {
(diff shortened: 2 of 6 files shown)


BenTaylorDev:
Reviewed alongside its managed-path twin, Intelligence#866, since the two only make sense as a pair. Three things to change here, one decision to make, and a suggested order at the end.

## 1. The two flatteners disagree on a block that reports no value

`flattenStateValues` assigns unconditionally:

```ts
out[blockId] = elementStateValue(el);
```

so a block whose element reports nothing leaves a key present with value `undefined`, and `Object.keys(values).length > 0` is then still `1`, so `values` is emitted. Intelligence#866's normalizer takes the opposite rule — it skips the key, and omits `values` entirely if that leaves the map empty.

For a message hosting one untouched optional input, that means:

- self-hosted → `ctx.values = { prio: undefined }`
- managed → `ctx.values = {}`

`Object.keys`, `in` and `Object.entries` all differ, so the "a handler reads one shape on every surface and on both managed and self-hosted deployments" claim doesn't hold in exactly this case.

It's also inconsistent within this package: the managed path takes `values` off the wire as JSON (`packages/channels-intelligence/src/delivery-adapter.ts:415`), and JSON cannot carry `undefined` — so the same SDK yields two shapes depending on deployment.

Worth noting the new test doesn't pin either behaviour:

```ts
expect(evt.values).toEqual({ prio: undefined });
```

vitest's `toEqual` treats an undefined-valued key as absent, so this passes whether the key exists or not.

Suggest adopting #866's rule (skip `undefined`) and asserting it with `Object.keys` / `not.toHaveProperty`. The reason #866 gives for it applies here too.

## 2. JSON-parsing now reaches free text — a regression on the shipped modal path

`flattenViewValues` used to read `el.value ?? el.selected_option?.value` and return the raw string. Sharing `elementStateValue` means its final arm now runs:

```ts
return parseValue(el.value ?? el.selected_option?.value ?? undefined);
```

`decodeViewSubmission` has shipped for several releases, so this changes behaviour for existing modal handlers: a `plain_text_input` where the user types `1234` now arrives as the number `1234`, `true` as a boolean, `null` as null. A handler doing `values.orderId.trim()` throws `trim is not a function`.

The compatibility argument in the description — "zero production code in the repo reads `ctx.values`" — covers this repo, but customer handlers are the consumer, and that's the field's whole purpose.

`SlackElementState.type` is available, so the ladder can discriminate: JSON-parse only author-encoded values (a button's `value`, `selected_option(s).value`) and never `plain_text_input` / `rich_text_input` content. That removes the coercion and keeps both halves of the pair in sync, since Intelligence#866 inherits the same footgun via its shared `readInteractionValue`.

## 3. Hop 3 isn't just a follow-up — it's what makes the field readable

The description flags the renderer gap as an open product-vocabulary decision. I think it's stronger than that: as things stand, the key this PR chose is the one half an author cannot name on a message.

- `packages/channels-slack/src/render/block-kit.ts:224` renders `<Input>` as an `input` block with no `block_id`. `grep -rn block_id packages/channels-slack/src/render/` hits only `render/modal.ts:45,68,89`, so on a message Slack assigns a random block id per render.
- The element's `action_id` is `idFromHandler(props.onSubmit) ?? "input"` (`block-kit.ts:230`) — a minted ActionStore id, or the literal string `"input"` for every `<Input>` with no `onSubmit`. So neither half of the pair is author-named on a message.
- Teams honours `props.name` (`packages/channels-teams/src/render/adaptive-card.ts:251`), which is exactly why Teams handlers read `ctx.values.root_cause`.

So `<Input name="root_cause">` yields `ctx.values.root_cause` on Teams and `ctx.values["<random>"]` on Slack, on managed and self-hosted alike. The parity framing holds at the transport layer but not at the API a customer uses.

The new tests can't catch this because they hand-write payloads whose block ids are already meaningful names (`root_cause`, `severity`, ...) — which is the shape only a hand-authored native block produces today.

I'd rather see `block_id: props.name` (with a fallback) land with this PR than as a separate change. Failing that, both this description and the ticket should say plainly that `ctx.values` is unusable for the cross-platform `<Input>`, so the next person doesn't have to re-derive why the keys are gibberish.

## Suggested order

No hard dependency between the two PRs — each is inert without its own path and neither breaks the other — but this one should go first, since it owns the shape both sides have to agree on:

1. Fix (1) and (2) here, and decide (3).
2. Merge this.
3. Merge Intelligence#866.

One check that belongs to #866 but originates here: the join-boundary validator uses `hasExactFields` (`packages/channels-intelligence/src/delivery-transport.ts:1405`), so an unknown key rejects the *whole* turn rather than dropping the field. `values` entered that allowlist in 5ceb537 (2026-08-01, package at 0.5.0), and npm carries 0.4.x releases. Any runtime pinned below 0.5.0 will have Slack interaction turns rejected outright once #866 ships — only for messages hosting inputs. Prod is on 0.8.x, so this is a confirm-and-move-on, not a blocker.

BenTaylorDev:
Follow-up to my review above — a second pass turned up three more things, all on this side of the pair.

**An untouched picker leaks a literal `null` self-hosted and vanishes managed.** `elementStateValue` guards the scalar arms with `!== undefined`:

```ts
if (el.selected_date !== undefined) return el.selected_date;
```

Slack sends `selected_date: null` / `selected_user: null` for an untouched picker, and `null !== undefined`, so this returns `null` — which survives JSON, unlike the `undefined` case I raised earlier. The managed normalizer guards on the value's type instead (`typeof selected === 'string' || typeof selected === 'number'`), so it falls through and drops the key. A message with one optional untouched datepicker therefore delivers `ctx.values = { "<blk>": null }` self-hosted and no `values` key at all managed. Same root fix as finding (1) in my earlier comment, but the `null` arm needs its own guard, not just an `undefined` skip.

**Only the first stateful element in a block survives, and this renderer packs many into one.** Both flatteners take `inner[blockId] ?? Object.values(inner)[0]`, documented as "the block's first (in practice only) element". That doesn't hold for the `actions` blocks `block-kit.ts:196` emits — it pushes one `{ type: "actions", elements }` holding up to `SLACK_LIMITS.actionsElements` elements, with no `block_id`. So `<Actions><Select onSelect={a}/><Select onSelect={b}/><Button/></Actions>` puts both selects under a single auto-generated block id, and only the first one's value ever reaches the handler — silently, with nothing signalling a dropped field.

I'm confident about the renderer half of that from the code; the part worth confirming against a real payload is whether Slack reports stateful elements in `actions`/`section` blocks under `state.values` at all, or only `input` blocks. That's a good thing to pin with the dev probe, and it's cheap to add: put two selects in one `<Actions>` next to the button.

**"Byte-identical to before" holds only for messages with no stateful element whatsoever.** The compatibility argument is stated as "a plain single-button message decodes byte-identically". If the answer to the question above is that Slack does report them, then any existing message that renders a `<Select>`, a picker, or an `<Input>` alongside a button now newly carries a `values` object on a button click where the field was previously absent. Still additive, but the framing is narrower than the body claims, and it's the same random-key problem as hop 3.

One correction in your favour, from the same pass: the `hasExactFields` allowlist in `channels-intelligence/src/delivery-transport.ts:1383` gained `"values"` in the very commit that made the check strict (5ceb537), so there is no window where a released SDK is strict *and* lacks the key. The pre-0.5.0 concern I raised at the end of my first comment is narrower than I implied — worth a glance, not a gate.

BenTaylorDev (approved):
Re-reviewed on 0cad93cd9 (rebased onto current main; the `channels-slack` tree is byte-identical to the merge I tested). All six findings from my two August comments are resolved, and each fix is pinned by a test: I removed each one in turn and the suite fails every time (free-text parsing 7, the `null` guard 9, the `undefined` omission 13, the multi-element keying 2, the renderer's `block_id` 7). `channels-slack` passes 461/461 and `check-types` is clean, and CI is green on this head.

One follow-up, and it is on the managed side, not here. Intelligence#866 merged with the older "first element per block" rule, so the two now diverge for a block holding more than one stateful element. With two unnamed selects in one `<Actions>`, self-hosted delivers `{ selA, selB }` and managed delivers `{ <random block id>: a }`. This PR owns the shape, so the fix is to port the multi-element rule into `flattenSlackStateValues` in Intelligence's `slack-ingress.ts`. Every other case I probed decodes identically on both paths.

Issue #2051, opened 2025-06-24
🐛 Bug: Frontend Not Receiving Tool Result Until LangGraph Workflow Completes


yanmxa:
### ♻️ Reproduction Steps

I'm trying to synchronize backend action execution (built with LangGraph) with a tool node and display the results in the frontend. Here's the setup:

🔧 Backend (Python, LangGraph)
```python
async def tool_node(state: AgentState, config: RunnableConfig):
    messages = state.get("messages", [])
    ai_message = messages[-1]

    if getattr(ai_message, "tool_calls", []):
        tool_messages = await asyncio.gather(*[
            handle_tool_call(state, config, tc) for tc in ai_message.tool_calls
        ])
        messages.extend(tool_messages)    
    return {
        **state,
        "messages": messages,
    }
```

💻 Frontend (TypeScript)
```ts
useCopilotAction({
  name: actionName,
  available: "disabled", // Prevent the agent/UI from invoking this tool directly — for rendering only
  render: (obj) => {
    const { status, args, name, result } = obj as any;
    console.log("actionName", actionName, status, args, name, result);
    return (
      <GenericAction status={status} args={args} name={name} result={result} />
    );
  },
});
```

### ✅ Expected Behavior

Each time a LangGraph tool node completes backend execution, the result should be immediately available and rendered on the frontend.

### ❌ Actual Behavior

The result remains empty on the frontend until the entire LangGraph workflow finishes, instead of updating as each tool node completes.

### 𝌚 CopilotKit Version

```shell
poetry show copilotkit         
 name         : copilotkit            
 version      : 0.1.39                
 description  : CopilotKit python SDK 



pnpm ls | grep copilotkit     
@copilotkit/react-core 1.8.13
@copilotkit/react-ui 1.8.13
@copilotkit/runtime 1.8.13
```

### 📄 Logs (Optional)

```shell

```

walfly:
I'm also experiencing this

Abeansits:
same!

kkwcode:
i have sovled this problem by modify source
# Solution for Tool Status Update Issue

I've resolved the tool status update issue in CopilotKit by implementing a custom state management approach. Here's the complete solution:

## Overview

The core issue is that tool execution status doesn't update to `complete` on the frontend after the tool finishes. This happens because `actionExecutionId` isn't properly passed through the streaming pipeline.

## Solution

### 1. Source Code Modification

[...]

Issue #1982, opened 2025-06-13
🐛 Bug: Tool calls stuck in executing with langgraph.js


GigaDroid:
### ♻️ Reproduction Steps

Create a graph using langgraph.js including a tool node. Bind tools to the llm and let the llm execute it. 

Implement useCopilotAction like this 

```
useCopilotAction({
			name: "tool",
			available: "disabled", // Don't allow direct invocation
			render: ({ name, args, status, result }: CatchAllActionRenderProps<[]>) => {
return <div>status</div>
}
```

### ✅ Expected Behavior

The tool call is rendered in the chat and the status is "complete" after execution finishes

### ❌ Actual Behavior

The tool call is rendered in the chat, first with status "inProgress", then with status "executing"

### 𝌚 CopilotKit Version

```shell
Copilotkit: 1.8.14
Langgraph.js: 0.3.0
```

### 📄 Logs (Optional)

```shell

```

copilotkit-support:
@GigaDroid We'll look into this. For urgent production issues, schedule here: https://cal.com/nathan-tarbert-copilotkit/15min

inno-odyssean:
having same issue using python

BenTaylorDev:
👋 Triage sweep. Since this was filed, the LangGraph integration was rewritten onto AG-UI. The old node-boundary buffering (`StreamingStateExtractor`) that held tool calls "in executing" is gone; `@ag-ui/langgraph` now emits a `TOOL_CALL_RESULT` per tool at `OnToolEnd`, which drives the frontend action to `complete`. Could you retest on the current release (1.63.x)? We'll close as stale in ~2 weeks otherwise. (Companion of #2051 — same root cause, Python variant.)

BenTaylorDev:
Verified fixed — closing.

I reproduced the reported sequence against the current pipeline and the tool call now reaches `complete` with its result **mid-run**, before `RUN_FINISHED`:

- The AG-UI LangGraph adapter emits `TOOL_CALL_RESULT` at `OnToolEnd` (`integrations/langgraph/typescript/src/agent.ts`; same in the Python adapter), rather than withholding it to a node/graph boundary as the old `StreamingStateExtractor` path did.
- The frontend derives `status` purely from whether a `role: "tool"` message with a matching `toolCallId` exists, so the result landing mid-run flips the render to `complete` immediately.
[...]

PR #6527, opened 2026-08-17
fix: parallel v1 tool rendering, Vue provider HITL, wildcard HITL names and abort

BenTaylorDev:
Three independent tool-rendering / human-in-the-loop bugs, plus a `packages/core` gap the third one exposed. One commit each, so any of them can be reviewed or reverted alone.

> The `packages/core` commit was made with `--no-verify`. The pre-commit gate runs `nx test` for affected packages, and `@copilotkit/core`'s suite fails **in my local worktree** on 20 `core-inspector-metadata.test.ts` timeouts. Those are a local build artifact, not a real regression: `@copilotkit/shared` resolves through a symlinked `node_modules` to a package whose `dist` predates `parseInspectorMetadataV1`/`InspectorMetadataV1` (both present in `packages/shared/src/utils/inspector-metadata.ts`), so `packages/core`'s own `tsc --noEmit` also reports those four symbols missing. CI builds from source and the JS unit matrix is green. Every test the gate would have run, I ran directly.

### OSS-802 — v1 tool rendering dropped all but the first parallel tool call

`useLazyToolRenderer` read `message.toolCalls[0]` and returned a single render, so an assistant message carrying a parallel batch displayed only the first tool call. Tool calls 2..N were dropped with no `inProgress` state and no result — from the user's seat that is indistinguishable from a result that never arrived, which is how this surfaced while verifying #1982 / #2051 (that reporter's backend does `asyncio.gather` over every `tool_call`).

Now maps over every tool call. It still returns `null` when *none* of them produce output, so a message with no registered renderer keeps falling through to the custom-message renderer as before. v2's `CopilotChatToolCallsView` already did this correctly.

### OSS-803 — Vue `humanInTheLoop` provider prop resolved immediately

The Vue provider's `humanInTheLoop` **prop** path registered a handler that warned and resolved `undefined` right away, so the HITL UI flashed and disappeared instead of waiting for the user, and it registered `tool.render` unwrapped so the renderer never got a working `respond`.

Now matches the `useHumanInTheLoop` composable:

- the tool-call promise stays pending until `respond` is called
- `respond` is live only while status is `Executing`
- an aborted run **rejects** with `Human-in-the-loop interaction aborted`, so core records an explicit error tool result rather than silently resolving empty (#5554)
- pending interactions are keyed by tool call id, so parallel interrupts on the same tool stay independent
- a wildcard (`name: "*"`) registration keeps the invoked tool's name

The abort bullet depended on a `signal` that core did not always supply — see the fourth commit below.

### `packages/core` — wildcard tool handlers never received the abort signal

Found while verifying the abort claim above. The named-tool path calls handlers with `signal: this._runAbortController?.signal` (`run-handler.ts:821`), but `executeWildcardTool` passed only `{ toolCall, agent }` (`run-handler.ts:1007`). A wildcard handler that waits on something therefore could never observe an abort, so a HITL tool registered as `"*"` stayed pending forever when the run was stopped — in React's `useHumanInTheLoop` as much as in the Vue provider path this PR fixes. OSS-804 makes wildcard HITL actually usable, which is what made this reachable.

The fix adds that one argument. It is additive: handlers that ignore the signal are unaffected, and `signal` was previously `undefined` for this path.

React's provider-prop path has the same placeholder; that half is being fixed in #4955 (react-core only, no file overlap with this PR).

### OSS-804 — wildcard `useHumanInTheLoop` renderer received `"*"`

`useHumanInTheLoop` overwrote the render props' `name` with the *registration* name in all three status branches. For a named tool the two are equal so the override was a no-op; for `name: "*"` it clobbered the real name, leaving a wildcard HITL unable to tell the user which tool it was asking them to approve.

Now only overrides when the registration isn't a wildcard. This matches the contract already asserted in `defineToolCallRenderer.test.tsx` ("wildcard receives actual tool name, not `*`") and the behavior of the other wildcard wrappers (`useRenderTool`, `useDefaultRenderTool`, `WildcardToolCallRender`).

## Testing

Every fix has a test that fails without it. Each negative control was run by reverting **only** the source file on this branch and re-running the same test.

**OSS-802** — `packages/react-core/src/hooks/__tests__/repro-1982-2051-v1-path.test.tsx` (new, plus the passing v2 counterpart `src/v2/hooks/__tests__/repro-1982-2051.test.tsx`).

Before, on a clean `origin/main` worktree:

```
 ✓ src/v2/hooks/__tests__/repro-1982-2051.test.tsx (2 tests) 128ms
 ❯ src/hooks/__tests__/repro-1982-2051-v1-path.test.tsx (2 tests | 1 failed)
   ✓ v1 single backend tool call reaches complete with its result
   × v1 renders every parallel tool call, not just the first
     → expected null not to be null
```

After:

```
 ✓ src/hooks/__tests__/use-copilot-chat-internal-connect.test.tsx (7 tests)
 ✓ src/hooks/__tests__/use-coagent-state-render.e2e.test.tsx (22 tests)
 ✓ src/v2/hooks/__tests__/repro-1982-2051.test.tsx (2 tests)
 ✓ src/hooks/__tests__/repro-1982-2051-v1-path.test.tsx (2 tests)
 ✓ src/hooks/__tests__/use-copilot-action.e2e.test.tsx (1 test)

 Test Files  5 passed (5)
      Tests  34 passed (34)
```

(The other four files are every existing consumer of `useLazyToolRenderer` plus the v1 action path — the fallthrough-to-custom-renderer behavior is what they cover.)

**OSS-803** — 6 new unit tests in `packages/vue/src/v2/providers/__tests__/CopilotKitProvider.test.ts` (pending-until-respond, `respond` only while executing, abort rejects, already-aborted rejects, parallel interrupts independent, wildcard name), plus 2 **mounted** end-to-end tests in `CopilotKitProvider.hitl.e2e.test.ts`. Two existing tests asserted the old placeholder behavior (`resolves undefined` + the warning) or compared `render` by identity, and were updated: the renderer is now a wrapper, so they assert it *delegates* to the user's component.

The unit tests drive the registered renderer directly with a synthesized handler context, which cannot catch a mismatch between the key the handler stores under and the tool call id the real pipeline supplies. The e2e file closes that: it mounts `CopilotChat → CopilotChatToolCallsView →` the provider's wrapper, so the wrapper runs as a real Vue functional component and the key comes from core's own tool call id. Reverted, both fail with the exact reported symptom — the card never reaches `Executing`, so no approve button is ever rendered:

```
 × waits for the user and completes with the response
 × passes the invoked tool name to a wildcard registration
 TestingLibraryElementError: Unable to find an element by: [data-testid="hitl-approve"]
```

With the provider reverted, 7 unit tests fail:

```
 × processes humanInTheLoop tools and creates handlers
 × keeps the tool call pending until respond is called
 × exposes respond only while the tool is executing
 × rejects the pending tool call when the run is aborted
 × rejects immediately when the run was already aborted
 × keeps parallel interrupts on the same tool independent
 × passes the actual invoked tool name to a wildcard HITL renderer

 Tests  7 failed | 2 passed | 19 skipped (28)
```

Full Vue suite with the fix:

```
 Test Files  100 passed (100)
      Tests  1077 passed (1077)
```

**OSS-804** — 2 new tests in `packages/react-core/src/v2/hooks/__tests__/use-human-in-the-loop.e2e.test.tsx`: a wildcard HITL driven through `InProgress → Executing → Complete` via `MockStepwiseAgent`, asserting `name === "deleteFile"` at every status and that `"*"` never reached the renderer; plus a named-registration test guarding the no-op half.

With the hook reverted:

```
 FAIL  HITL Wildcard Registration > passes the actual invoked tool name to a wildcard HITL renderer, not '*'
 AssertionError: expected '*' to be 'deleteFile'
   Expected: "deleteFile"
   Received: "*"
```

With the fix: `use-human-in-the-loop.e2e.test.tsx (16 tests)` — all pass.

**core wildcard signal** — 2 new tests in `packages/core/src/__tests__/core-abort-during-tool.test.ts`, mirroring the named-tool pair already there: one asserts a wildcard handler receives an `AbortSignal`, the other that `stopAgent()` during wildcard execution aborts it. Reverting only the changed line in `run-handler.ts` fails both while the four existing tests stay green:

```
 ✓ should NOT restart the run when stopAgent is called during tool execution
 ✓ should pass an AbortSignal to the tool handler context
 ✓ should abort the signal when stopAgent is called during tool execution
 × should pass an AbortSignal to a wildcard tool handler context
   → expected undefined to be an instance of AbortSignal
 × should abort the signal when stopAgent is called during wildcard tool execution
   → expected undefined to be true
 ✓ should NOT restart the run when agent.abortRun() is called directly during tool execution
```

Scope note on this one: in a worktree, `@copilotkit/core` resolves to the built package, so the react-core and Vue suites here exercise the *published* core rather than this source change. The core fix is verified at the source level by the tests above; the composed behavior (wildcard HITL rejecting on abort through the Vue provider) follows from two separately-verified halves — core now supplies the signal, and the Vue path's abort test proves it rejects when given one — but I did not execute that combination end to end. CI builds from source and will.

### Repo-wide checks

| Check | Result |
| --- | --- |
| `react-core` full suite | 1503 passed, 1 failed — `use-interrupt.test.tsx > does not persist a tool-result message for backend-owned interrupts`. Not from this branch: it fails identically in a separate pristine `origin/main` worktree. Same local-build caveat as the `core` row below; CI is green. |
| `react-core` `tsc --noEmit` | clean |
| `vue` full suite | 1079 passed (101 files) |
| `vue` `vue-tsc --noEmit` | clean |
| `vue` `eslint` on changed files | 3 errors, all byte-identical to `origin/main` on the same file (2 unused-import false positives, 1 pre-existing `any` at what is now line 543) — no new lint errors |
| `core` full suite | With the fix: 637 passed, 20 failed. With the one changed line reverted: **22** failed. The two extra are exactly the two new wildcard tests, and set-differencing the failure names both ways shows nothing else moved — so the fix breaks nothing and the 20 are independent of it. All 20 are `core-inspector-metadata.test.ts` `Timeout waiting for condition`, caused by the stale `@copilotkit/shared` build described above rather than by anything on this branch. |
| `core` `tsc --noEmit` | 4 errors, all the stale-`@copilotkit/shared` symbol resolution described above; none in changed code |

### CI: the one red check is repo-wide, not from this branch

`langgraph-python` (dojo e2e) fails on three `agenticChatPage.spec.ts` journeys. Event counts match exactly (33/33 and 129/129); the only difference is `event 2: STATE_SNAPSHOT — expected: <missing>, actual: "local_dev"`. `local_dev` appears nowhere in the repo, so it is a value the LangGraph dev server supplies at run time against a recorded trace that does not have it.

It is not this branch: **#6530, a CSS-only change to `header.css`, fails the same job identically**, and merging current `origin/main` in did not change the outcome. Nothing in this PR touches state snapshots, the runtime, or the Python SDK.

## Not in this PR

OSS-805 (passive client disconnect does not abort an in-flight run) is deliberately left out. Abort-on-disconnect is not unconditionally correct — disconnect-tolerant runs are a deliberate feature for reconnect-to-in-flight-thread and durable/background runs — so the per-runner semantics need deciding before `request.signal` gets wired up. It also touches runtime behavior rather than frontend rendering, and belongs in a PR that can be reviewed on that basis.

Two adjacent gaps found while doing this, not filed as part of it:

- the Vue `useHumanInTheLoop` composable has the same wildcard `name` clobber this PR fixes for React (OSS-804), and no abort handling at all
- React's provider-prop path has the same placeholder handler this PR fixes for Vue (OSS-803) — that's #4955's scope

Fixes OSS-802
Fixes OSS-803
Fixes OSS-804

### DIFF
--- a/packages/core/src/__tests__/core-abort-during-tool.test.ts
+++ b/packages/core/src/__tests__/core-abort-during-tool.test.ts
@@ -121,6 +121,64 @@ describe("CopilotKitCore - abort during tool execution", () => {
     expect(signalAbortedDuringHandler).toBe(true);
   });
 
+  it("should pass an AbortSignal to a wildcard tool handler context", async () => {
+    // Wildcard tools execute through their own path, which omitted the signal
+    // the named-tool path passes. A wildcard handler that waits on something —
+    // a human-in-the-loop tool registered as "*" — could never see an abort.
+    const agent = new MockAgent({ agentId: "test" });
+    copilotKitCore.addAgent__unsafe_dev_only({
+      id: "test",
+      agent: agent as any,
+    });
+
+    let receivedSignal: AbortSignal | undefined;
+    const wildcardTool = createTool({
+      name: "*",
+      handler: vi.fn(async (_args: any, context: any) => {
+        receivedSignal = context.signal;
+        return "done";
+      }),
+      followUp: false,
+    });
+    copilotKitCore.addTool(wildcardTool);
+
+    // A tool name with no specific registration, so the wildcard handles it.
+    agent.setNewMessages([createToolCallMessage("unregisteredTool")]);
+
+    await copilotKitCore.runAgent({ agent: agent as any });
+
+    expect(wildcardTool.handler).toHaveBeenCalledOnce();
+    expect(receivedSignal).toBeInstanceOf(AbortSignal);
+    expect(receivedSignal!.aborted).toBe(false);
+  });
+
+  it("should abort the signal when stopAgent is called during wildcard tool execution", async () => {
+    const agent = new MockAgent({ agentId: "test" });
+    copilotKitCore.addAgent__unsafe_dev_only({
+      id: "test",
+      agent: agent as any,
+    });
+
+    let signalAbortedDuringHandler: boolean | undefined;
+    const wildcardTool = createTool({
+      name: "*",
+      handler: vi.fn(async (_args: any, context: any) => {
+        expect(context.signal.aborted).toBe(false);
+        copilotKitCore.stopAgent({ agent: agent as any });
+        signalAbortedDuringHandler = context.signal.aborted;
+        return "done";
+      }),
+      followUp: false,
+    });
+    copilotKitCore.addTool(wildcardTool);
+
+    agent.setNewMessages([createToolCallMessage("unregisteredTool")]);
+
+    await copilotKitCore.runAgent({ agent: agent as any });
+
+    expect(signalAbortedDuringHandler).toBe(true);
+  });
+
   it("should NOT restart the run when agent.abortRun() is called directly during tool execution", async () => {
     const agent = new MockAgent({ agentId: "test" });
     copilotKitCore.addAgent__unsafe_dev_only({
--- a/packages/core/src/core/run-handler.ts
+++ b/packages/core/src/core/run-handler.ts
@@ -1007,6 +1007,11 @@ export class RunHandler {
           const result = await wildcardTool.handler(wildcardArgs as any, {
             toolCall,
             agent,
+            // Same signal the named-tool path passes. Without it a wildcard
+            // handler that waits on something — a human-in-the-loop tool
+            // registered as `"*"` is the case that surfaced this — can never
+            // observe an abort, so stopping the run leaves it pending forever.
+            signal: this._runAbortController?.signal,
           });
           if (result === undefined || result === null) {
             toolCallResult = "";
--- a/packages/react-core/src/hooks/__tests__/repro-1982-2051-v1-path.test.tsx
+++ b/packages/react-core/src/hooks/__tests__/repro-1982-2051-v1-path.test.tsx
@@ -1,0 +1,122 @@
+/**
+ * v1-path half of the #1982 / #2051 reproduction.
+ *
+ * Both reporters use v1 `useCopilotAction({ available: "disabled", render })`.
+ * That hook is now a shim that routes `available: "disabled"` and `name: "*"`
+ * to the same `useRenderToolCall` registration v2 uses, so the status/result
+ * derivation is shared. What is NOT shared is the v1 *display* seam:
+ * `useLazyToolRenderer`, which picks `message.toolCalls[0]`.
+ *
+ * #2051's backend runs asyncio.gather over every tool_call on the assistant
+ * message, so parallel tool calls are exactly its shape. This pins down what
+ * the v1 seam does with more than one.
+ */
+import React from "react";
+import { render, screen, waitFor } from "@testing-library/react";
+import { z } from "zod";
+import { useLazyToolRenderer } from "../use-lazy-tool-renderer";
+import { CopilotKitProvider } from "../../v2/providers/CopilotKitProvider";
+import type { AIMessage, Message } from "@copilotkit/shared";
+
+const WildcardRender: React.FC<any> = ({ name, status, result }) => (
+  <div data-testid={`v1-${name}`}>
+    {name}:{status}:{String(result ?? "")}
+  </div>
+);
+
+function Harness({
+  message,
+  messages,
+}: {
+  message: AIMessage;
+  messages: Message[];
+}) {
+  const lazy = useLazyToolRenderer();
+  const rendered = lazy(message, messages);
+  return <div data-testid="host">{rendered ? rendered() : null}</div>;
+}
+
+function renderV1({
+  message,
+  messages,
+}: {
+  message: AIMessage;
+  messages: Message[];
+}) {
+  return render(
+    <CopilotKitProvider
+      renderToolCalls={
+        [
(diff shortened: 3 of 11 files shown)


BenTaylorDev:
Superseded by #7315, now merged as `e0e6d5f9bc`.

All four of this branch's fixes are on `main`, but only two came from here. The parallel v1 tool rendering fix landed through #6682, and its file has since moved under `v1-deprecated/`, so this branch conflicts on it. The wildcard HITL name fix landed through `b17c8556be`.

#7315 carried the remaining two — the Vue `humanInTheLoop` provider prop and the wildcard abort signal — cherry-picked onto current `main`, with the `@copilotkit/core` suite genuinely green rather than committed with `--no-verify`.

Closing as superseded rather than rebasing: two of the five commits would have to be dropped, and the largest targets a path that no longer exists.

Issue #5041, opened 2026-05-27
🐛 Bug: <CopilotKit threadId={...}> prop not propagated to ProxiedCopilotRuntimeAgent — agent uses its own UUID


ChuckJonas:
### ♻️ Reproduction Steps

1. Pass an explicit `threadId` to `<CopilotKit>`:

```tsx
const [threadId] = useState(() => crypto.randomUUID()) // e.g. "aaa..."

<CopilotKit runtimeUrl="..." threadId={threadId}>
  <CopilotChat />
</CopilotKit>
```

2. Send a message through CopilotChat.
3. Inspect the outgoing `agent/run` request body — the `threadId` in the request is a **different UUID** (e.g. `"bbb..."`) from the one passed as a prop.

### ✅ Expected Behavior

The `threadId` prop on `<CopilotKit>` should be used as the thread identity for all agent HTTP requests (`agent/run`, `agent/connect`, `agent/stop`). The `agent.threadId` on `ProxiedCopilotRuntimeAgent` should reflect the user-provided value.

### ❌ Actual Behavior

The user-provided `threadId` flows into `ThreadsProvider` → React context (`useThreads().threadId`, `useCopilotContext().threadId`, `useCopilotChatConfiguration().threadId`) but is **never set on the underlying `ProxiedCopilotRuntimeAgent.threadId`**.

When the agent is created during the `/info` fetch, `ProxiedCopilotRuntimeAgent` inherits from `AbstractAgent`, whose constructor does:

```javascript
// @ag-ui/client AbstractAgent constructor
this.threadId = config.threadId ?? uuid.v4()
```

Since no `threadId` is passed in the agent config, the agent generates its own UUID. All HTTP requests use this internal UUID, not the user-provided one.

### Root Cause

In `packages/react-core/src/hooks/use-copilot-chat_internal.ts`, the connect and run flows call:

```typescript
await copilotkit.connectAgent({ agent });
await copilotkit.runAgent({ agent });
```

Neither call propagates the context `threadId` to `agent.threadId`. The `RunAgentParameters` type doesn't even include `threadId` — it only has `runId | tools | context | forwardedProps`.

The fix would be to set `agent.threadId` from the context/prop value before calling `connectAgent`/`runAgent`, or to pass it through the parameters.

### Impact

[...]

PR #6528, opened 2026-08-17
fix(core): hydrate self-hosted threads whose /connect replay contains an errored run

### LINKED ISSUE #4943: existing-thread hydration fails in legacy CopilotPopup when /connect replays historical runs

BenTaylorDev:
Hydrating an existing thread through `/connect` fails on a **self-hosted** runtime whenever that thread's history contains a run that ended in `RUN_ERROR`.

## The defect

A `/connect` response is a *replay* of a thread's history, so it can legitimately carry several past runs back to back — including an errored run followed by a later `RUN_STARTED`. The base `AbstractAgent` connect pipeline pushes that stream through `verifyEvents`, which enforces AG-UI's **single run** lifecycle rules and rejects the sequence outright:

```
Cannot send event type 'RUN_STARTED': The run has already errored with 'RUN_ERROR'. No further events can be sent.
```

The user-visible effect is the one reported in #4943: `agent_connect_failed` on reload, and the existing thread never hydrates its prior messages.

`IntelligenceAgent` already omitted `verifyEvents` from its connect pipeline for exactly this reason (its JSDoc spells it out). But `ProxiedCopilotRuntimeAgent.connectAgent` only takes that path in `RUNTIME_MODE_INTELLIGENCE` — self-hosted (`RUNTIME_MODE_SSE`) fell through to `super.connectAgent()` and inherited the single-run verification. So the managed product was fine and self-hosting was not.

## The fix

`ɵconnectWithoutEventVerification` (`packages/core/src/utils/connect-replay.ts`) holds the verifyEvents-free pipeline, and **both** paths now use it. `transformChunks` is still applied — message reassembly is needed either way.

This is a de-duplication rather than a third copy: `IntelligenceAgent.connectAgent` drops ~100 lines of hand-replicated base pipeline (including its private-field `any` escape hatch) and keeps only its canonical-run-id handling before delegating. Net `intelligence-agent.ts` change is −101 lines.

### Fidelity to the base implementation

The helper was diffed statement-by-statement against the **real** `AbstractAgent.connectAgent` in `@ag-ui/client@0.0.57` (recovered from the shipped source map), not just against `IntelligenceAgent`'s replica. `verifyEvents` is the only intended difference.

That diff caught a defect in the first push: the base special-cases `AGUIConnectNotImplementedError` (swallow → `EMPTY`) and the replica did not. `IntelligenceAgent` never needed it — it always implements `connect()` — so the gap was invisible there, but on the SSE path it is load-bearing: `run-handler.ts:447-450` documents that `await agent.detachActiveRun()` only stopped deadlocking because that error path still reaches the pipeline's finalize block. Routing it through `onError` would also fire run-failure callbacks on every subscriber for a benign condition. Restored, with a regression test.

Also confirmed that dropping `verifyEvents` cannot alter a well-formed replay: it is a pure gate — 18 `return of(event)` pass-throughs, 42 error paths, and zero `endWith` / `startWith` / `tap` side effects. It only removes the single-run rejection.

The existing upstream TODO still stands and is carried over: `@ag-ui/client@0.0.57`'s `connectAgent(parameters?, subscriber?)` takes no option to skip verification, so this override is still the only way to express "this stream is a replay, not a run."

## On the second half of #4943

The issue also reports that the legacy chat path doesn't copy the resolved `threadId` onto the agent before connect/run. **That half is already fixed on `main`** — the #5041/#4739 fix put `agent.threadId = resolvedThreadId` in v2 `useAgent`, and `useCopilotChatInternal` delegates to that same hook. Nothing more was needed.

It was untested, though, and untestable from the suite that looked like it covered it: `use-copilot-chat-internal-connect.test.tsx` mocks `useAgent` wholesale, so it cannot observe threadId propagation at all. This PR adds `legacy-chat-explicit-threadid.test.tsx`, which drives the legacy hook through the **real** `useAgent` under a real `<CopilotKit>`, covering both the explicit-threadId case and the "don't adopt a non-explicit placeholder UUID" case.

It reads the agent off `useCopilotChatInternal()`'s own return value rather than calling `useAgent` in the probe. That distinction matters: the first version of this test did call `useAgent`, so the probe itself performed the assignment under test and the test passed **even with `useCopilotChatInternal()` removed entirely**. The current version is mutation-checked — disabling the assignment in v2 `useAgent` fails it (`expected 'dc051f13-…' to be 'cookie-backed-thread'`).

Contributor PR #4969 proposed a manual assignment for this half; it is now redundant.

## Testing

Worktree caveat, stated up front: this worktree symlinks the primary checkout's `node_modules`, so `@copilotkit/shared` and `@copilotkit/core` resolve to that checkout's **stale `dist`**. That produces failures unrelated to this change; each is baselined against clean `main` in the same environment below. CI installs fresh and is the authoritative gate.

**1. Reproduces the reported failure before the fix.** The new core test, run on unmodified `origin/main`, fails with the exact error from the issue:

```
FAIL  src/__tests__/proxied-connect-replay-multi-run.test.ts > hydrates a thread whose replayed history contains an errored run
AssertionError: promise rejected "Error: Cannot send event type 'RUN_STARTE…" instead of resolving
Caused by: Error: Cannot send event type 'RUN_STARTED': The run has already errored with 'RUN_ERROR'. No further events can be sent.
```

**2. Passes after the fix**, hydrating both runs' messages (`["msg-1", "msg-2"]`):

```
✓ src/__tests__/proxied-connect-replay-multi-run.test.ts (1 test) 11ms
Test Files  1 passed (1)
```

**3. Connect-not-implemented guard, fail-first.** With the guard removed, the new second test fails exactly as the base contract predicts:

```
× swallows AGUIConnectNotImplementedError instead of failing the run
AssertionError: promise rejected "Error: Connect not implemented. This meth…" instead of resolving
```

**4. Full `@copilotkit/core` suite** — this is the evidence the `IntelligenceAgent` extraction is behavior-identical, since `intelligence-agent.test.ts` exercises that path heavily:

```
Test Files  59 passed (59)
      Tests  635 passed (635)
```

(excludes `core-inspector-metadata.test.ts`; its 20 failures are the stale-`shared`-dist artifact — verified identical on clean `main`: `20 failed | 2 passed`, missing export `InspectorMetadataV1`)

**5. `@copilotkit/react-core` — new + adjacent existing suites:**

```
✓ src/hooks/__tests__/use-copilot-chat-internal-connect.test.tsx (7 tests)
✓ src/hooks/__tests__/legacy-chat-explicit-threadid.test.tsx (2 tests)
✓ src/components/copilot-provider/__tests__/v1-explicit-threadid-bridge.test.tsx (5 tests)
Test Files  3 passed (3)
      Tests  14 passed (14)
```

Full react-core suite: `8 failed | 1492 passed (1500)`. All 8 are in `use-interrupt` / `use-pin-to-send` / `CopilotChatView.pinToSend` — none touch connect replay or threadId, and clean `main` in this worktree fails the identical 8 (`8 failed | 37 passed (45)` for those three files alone).

**6. `@copilotkit/vue`** (affected via core): `100 passed (100)` files, `1072 passed (1072)` tests.

**7. Types, lint, format:**

```
tsc -p packages/core/tsconfig.json --noEmit   → no errors in any changed file
oxlint  <5 changed files>                     → Found 0 warnings and 0 errors
oxfmt --check <5 changed files>               → All matched files use the correct format
```

The only remaining `tsc` errors are 4 pre-existing stale-dist ones in `agent-registry.ts` / `types.ts` (`InspectorMetadataV1`), untouched by this PR.

Fixes #4943

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved thread hydration when reconnecting to histories containing multiple runs, including runs that previously ended in error.
  * Prevented unsupported connection errors from being reported as run failures.
  * Ensured connection state is properly finalized after replaying a thread.
  * Legacy chat components now correctly reuse an explicitly provided thread ID while preserving generated IDs when none is provided.

### DIFF
--- a/packages/core/src/__tests__/proxied-connect-replay-multi-run.test.ts
+++ b/packages/core/src/__tests__/proxied-connect-replay-multi-run.test.ts
@@ -1,0 +1,109 @@
+import { describe, it, expect, beforeEach, afterEach, vi } from "vitest";
+import { ProxiedCopilotRuntimeAgent } from "../agent";
+
+/**
+ * Issue #4943 (part 2): hydrating an existing thread through `/connect` on a
+ * self-hosted runtime.
+ *
+ * A `/connect` response replays the thread's history, so it can legitimately
+ * carry events from several past runs — including a run that ended in
+ * RUN_ERROR followed by a later RUN_STARTED. AbstractAgent's connect pipeline
+ * runs the stream through `verifyEvents`, which enforces single-run lifecycle
+ * rules and rejects a RUN_STARTED after a RUN_ERROR:
+ *
+ *   Cannot send event type 'RUN_STARTED': The run has already errored with
+ *   'RUN_ERROR'. No further events can be sent.
+ *
+ * IntelligenceAgent already omits `verifyEvents` from its connect pipeline for
+ * exactly this reason (see intelligence-agent.ts). Self-hosted runtimes
+ * (RUNTIME_MODE_SSE) still inherit the base pipeline, so replay of a thread
+ * whose history contains an errored run fails to hydrate.
+ */
+
+const encoder = new TextEncoder();
+
+function createConnectReplayResponse(): Response {
+  const stream = new ReadableStream({
+    start(controller) {
+      const events = [
+        // ---- first past run: ended in RUN_ERROR ----
+        { type: "RUN_STARTED", threadId: "existing-thread", runId: "run-1" },
+        {
+          type: "TEXT_MESSAGE_START",
+          messageId: "msg-1",
+          role: "assistant",
+        },
+        {
+          type: "TEXT_MESSAGE_CONTENT",
+          messageId: "msg-1",
+          delta: "first run answer",
+        },
+        { type: "TEXT_MESSAGE_END", messageId: "msg-1" },
+        {
+          type: "RUN_ERROR",
+          threadId: "existing-thread",
+          runId: "run-1",
+          message: "upstream model error",
+        },
+        // ---- second past run: replayed in the same /connect stream ----
+        { type: "RUN_STARTED", threadId: "existing-thread", runId: "run-2" },
+        {
+          type: "TEXT_MESSAGE_START",
+          messageId: "msg-2",
+          role: "assistant",
+        },
+        {
+          type: "TEXT_MESSAGE_CONTENT",
+          messageId: "msg-2",
+          delta: "second run answer",
+        },
+        { type: "TEXT_MESSAGE_END", messageId: "msg-2" },
+        {
+          type: "RUN_FINISHED",
+          threadId: "existing-thread",
+          runId: "run-2",
+          result: { newMessages: [] },
+        },
+      ];
+      controller.enqueue(
+        encoder.encode(
+          events.map((e) => `data: ${JSON.stringify(e)}\n\n`).join(""),
+        ),
+      );
+      controller.close();
+    },
+  });
+
+  return new Response(stream, {
+    status: 200,
+    headers: { "content-type": "text/event-stream" },
+  });
+}
+
+describe("self-hosted /connect replay across multiple past runs (#4943)", () => {
+  const originalFetch = global.fetch;
+  let fetchMock: ReturnType<typeof vi.fn>;
+
+  beforeEach(() => {
+    fetchMock = vi.fn(() => Promise.resolve(createConnectReplayResponse()));
+    global.fetch = fetchMock as unknown as typeof fetch;
+  });
+
+  afterEach(() => {
+    vi.restoreAllMocks();
+    global.fetch = originalFetch;
+  });
+
+  it("hydrates a thread whose replayed history contains an errored run", async () => {
+    const agent = new ProxiedCopilotRuntimeAgent({
+      runtimeUrl: "https://runtime.example/hono",
+      agentId: "hydrating-agent",
+      transport: "rest",
+    });
+    agent.threadId = "existing-thread";
+
+    await expect(agent.connectAgent()).resolves.toBeDefined();
+
+    expect(agent.messages.map((m) => m.id)).toEqual(["msg-1", "msg-2"]);
+  });
+});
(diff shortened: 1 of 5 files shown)


MikeRyanDev (approved):

PR #6529, opened 2026-08-17
feat(channels): one Channel can run many named agents

AlemTuzlak:
## What does this PR do?

One Channel can register many named agents. `thread.runAgent({ agentId })` runs one of them per call.

- `createChannel({ agent })` stays the default. `agents` is a name map. Do not pass both `agent` and `agents.default`.
- Extra agents keep the same Slack or Teams chat. They get their own checkpoint (`thr_abc::billing`) and Intelligence wire id (`triage:billing`). The default keeps today's ids.
- `resume()` continues the one waiting agent. Two or more waiters need `agentId`. A picker click stamps the agent. Waiters persist on the existing StateStore.
- Types infer keys from `agents`. Extras-only Channels require `agentId` at compile time.

Guide: `/channels/multiple-agents`.

## Related PRs and Issues

None.

## Checklist

## Test plan

- [x] `nx run @copilotkit/channels-core:test` — 313 passed
- [x] `nx run @copilotkit/channels-core:check-types` — passed
- [x] `nx run @copilotkit/channels-intelligence:test` — 216 passed
- [x] `nx run @copilotkit/channels-slack:test` — 406 passed
- [x] `nx run @copilotkit/channels-teams:test` — 99 passed
- [x] shell-docs `channels-docs` + `channel-guide-routes` — 47 passed


BenTaylorDev (changes requested):
Reviewed against current `main`. Good design, and the risk I went looking for is clean — but there is a hard regression reaching plain single-agent Channels.

## Clean

Durable-state compatibility is proven: `checkpointThreadId()` returns the base thread id unchanged for `"default"` (`resolve-channel-agents.ts:75-82`), `canonicalAgentId(declared, "default")` returns the Channel name unchanged (`:84-88`), consumed at `delivery-adapter.ts:489` with `realtime-gateway-launcher.ts:68-71` forcing `options.channelName === channel.name`. No existing checkpoint key or wire id changes. `agent` vs `agents.default` is enforced at runtime (`:52-55`), and an ambiguous `resume()` throws with the waiter list rather than picking (`thread.ts:654-656`).

## Blocking

**1. An unresumed interrupt bricks the conversation permanently.** `thread.ts:731-737` throws on every non-resume `runAgent()` while the agent is in the `interrupt:<conversationKey>` waiter set; the only clearing path is `thread.ts:891`, reachable only from a run that completes *without* interrupting. No TTL, no abandon path. On a legacy `createChannel({ agent })` with no `agents` map: Alice triggers an approval, nobody clicks, Bob's next message throws `ChannelAgentInterruptPendingError` — forever. Reproduced and mutation-checked: the same probe passes on `main`, fails here.

The diff corroborates it. `main:hitl-continuation.test.tsx:198` ("independent runs **on one Thread**...") drives two turns on one `conversationKey` and asserts two snapshots; this PR splits it onto `thread-alice`/`thread-bob`, and `:315` changes one of two same-agent concurrent runs to `{ agentId: "billing" }`. Both route around the gate instead of fixing it.

Direction: make the gate non-fatal (warn and proceed, or auto-clear on a new inbound turn), or scope it to non-default agent ids.

**2. A failed resume leaves the waiter set forever.** `thread.ts:885-895` is inside the inner `try`; the `catch` at `:918-940` re-throws without clearing, so any agent/delivery/finalize error during resume is terminal. `addInterruptWaiter` also writes with no TTL (`interrupt-waiters.ts:16`) while every other action key uses `actionRetentionMs`.

**3. Approval buttons outstanding at deploy time die on upgrade.** `thread.ts:644-660` — a pre-existing button has no `agentId` on its `ActionSnapshot` (new field, `action-store.ts:42`) and no `interrupt:` entry, so a click hits `ChannelAgentResumeNoneError`. No backfill, no fallback to `defaultId`.

**4. Crash window.** The `agentId` stamp persists at mint time (`action-registry.ts:369-372`); the waiter only after the run loop returns (`thread.ts:885`). A restart in between leaves a live, stamped, clickable button with no waiter — same throw, no upgrade needed.

## Non-blocking

- `ChannelContinuationRequiredError` is exported but now unreachable — this deletes its only throw site (`main:thread.ts:623`), so `resume()` outside an interaction context runs with no continuation claim: no one-use `claimContinuation`, no channel/conversation/thread binding check, no initiator snapshot. Possibly intentional for "reply `yes` to resume", but undocumented and untested.
- "They get their own checkpoint" oversells it. The `::` suffix does reach the agent (`thread.ts:742` is a real choke point), but `delivery-adapter.ts:162-184` and `channels-slack/src/conversation-store.ts:123-125` then overwrite `agent.messages` from the shared thread — the PR's own test says "unsuffixed history thread". Only the remote checkpoint id differs. Worth a line in the guide.
- Scope creep: `realtime-gateway-launcher.ts:157-159` (`appApiFetch` into `IntelligenceStateStore`) is an unrelated fix; the `examples/slack` restructure is −651/+350 plus an Intelligence commit.

## State

CI is green but expired, and the branch is `CONFLICTING` so it cannot be re-run meaningfully. Eight files conflict, epicentre `thread.ts` / `create-channel.ts` — the same `run()`/`resume()` bodies and `isolateAgentInstance` rework this rewrites; `main` also added `run-loop.ts` and `identity.ts` in that path. Since the waiter mechanism has to change anyway, rebase first and re-derive it rather than resolving conflicts around the current design.

Before merging: restore `main:hitl-continuation.test.tsx:198` verbatim and run it here.

Issue #6479, opened 2026-08-13
release: BREAKING CHANGE footers never reach release notes (collector reads %s only)


mxmzb:
## Summary

`BREAKING CHANGE:` footers never reach release notes. The commit collector reads **subjects only**, so every footer written in this repo — migration guidance, breaking-change details, co-author trailers — is silently discarded before the release-notes step sees it.

## The line

`scripts/release/lib/changes.ts:43`

```ts
["log", range, "--oneline", "--no-merges", "--format=%H %s"]
```

`%s` is the subject. No `%b`, no `%B`. Corroborating: `grep -rn "BREAKING" scripts/release/` returns nothing, and `generate-ai-release-notes.ts:104` feeds the model *"Here are the recent git commits for additional context"* — i.e. whatever the collector produced, which is `<sha> <subject>` pairs.

## Why this matters now

Found while reviewing #6438, a `refactor(react-native)!` with a deliberately detailed `BREAKING CHANGE:` footer. The footer documents the migration consumers actually hit — `RenderToolProps`'s `args` becomes `Partial<T>` on the in-progress arm, so a renderer reading `args.foo` without narrowing on `status` now fails `check-types` with `TS18048`.

None of that can reach the release notes today.

There's a second-order effect worth flagging. This repo migrated **off** Changesets to conventional-commit-driven releases (`CLAUDE.md`: *"nothing consumes them and CI fails on them"*, enforced by `static_check-binaries.yml`). Changeset files were the artifact that carried migration prose. If footers are also unread, then for breaking changes the repo currently has **no** channel from commit to release notes except the 72-character subject line.

## Expected

A `BREAKING CHANGE:` footer, or at minimum the body of any commit whose subject carries `!`, should reach the release-notes generator.

## Suggested fix

Widen the collector to include bodies — e.g. `--format=%H%x1f%s%x1f%B%x1e` with explicit unit/record separators, since bodies are multi-line and `%H %s` parsing assumes one commit per line. Then decide what the generator does with them (all bodies is probably too much context; footers of `!` commits is probably the useful subset).

## Verification

- `scripts/release/lib/changes.ts:43` — the format string
- `scripts/release/lib/changes.test.ts:52` — the test asserts the same `%H %s`, so it pins the current behaviour
- `grep -rn "BREAKING" scripts/release/` — empty

## Not in scope for #6438

[...]

PR #6533, opened 2026-08-17
refactor(react-native)!: converge render-tool hooks onto react-core

mxmzb:
Closes #6438's deferred review finding **5b**. #6438 shipped 5a — retracting the false claim that React Native supported wildcard renderers, and documenting the gap as a known limitation. This is the actual fix.

## The defect: one cause, not nine

React Native's `useRenderTool` was **not** react-core's `useRenderTool`. Its entire body forwarded to a *different* hook while wearing the other one's name:

```ts
useFrontendTool<T>({ name, description, parameters, handler, agentId, render }, deps);
```

react-core has two hooks:

| hook | registers | wildcard `"*"` |
| --- | --- | --- |
| `useFrontendTool` | a tool **and** its renderer (`addTool`) | no special case |
| `useRenderTool` | a renderer **only** (`addHookRenderToolCall`) | special-cased into a schema-less fallback |

RN aliased the first under the second's name. Because that took the `addTool` path, `name: "*"` registered a frontend tool **literally named `*`**.

**Correction from review.** Earlier revisions of this description said that tool was advertised to the model. It was not, and @BenTaylorDev caught it: core filters the wildcard name out of the advertised list in `run-handler.ts`, and that filter landed in `31aa1e2162` — before this PR was opened, but after the base it was originally written against. The 1273-commit rebase moved the code forward and left the prose behind. All seven code and doc sites have been corrected.

**What it actually did is worse.** A tool named `*` is core's catch-all *handler*. When a tool call has no exact-name frontend tool and no result yet, core reaches for the wildcard tool and runs `executeWildcardTool` — and the tool-result splice sits *outside* the `if (wildcardTool?.handler)` guard. So an RN app whose author wanted a display-only fallback was also auto-answering every otherwise-unanswered tool call with an empty tool result and requesting a follow-up turn. Measured by driving a real turn through both paths: the old handler-less `*` produced **2 turns and a spliced `{ toolCallId, content: "" }`**; the shim's route produces 1 turn and no result.

Eight further reported symptoms share that single cause: a render-only registration was advertised and shadowed a same-named server tool; `addTool` evicted an existing same-named handler with only a `console.warn`; `description` / `parameters` were required though both are *tool* fields; `followUp` / `available` were not forwarded; `handler` dropped core's context argument, leaving `stopAgent()`'s abort signal unreachable; `agentId` changes never re-registered; two structurally incompatible `RenderTool*Props` families shipped side by side; and `hooks/index.ts` was a dead barrel.

## The fix

RN's own implementation is gone. `useRenderTool` and `useFrontendTool` are now re-exported from react-core, so each capability has exactly one implementation — **and `"*"` works as a real wildcard on React Native for the first time.**

Deleting rather than re-pointing the name was deliberate: re-pointing would let a `{ name, parameters, render }` call with no handler keep compiling while silently ceasing to register the tool.

## The landing is soft, on purpose

Following @BenTaylorDev's review, `useRenderTool` **still exists on React Native** as a deprecated shim for one minor. It routes by shape and warns in development, once per tool name:

| call shape | routes to | why |
| --- | --- | --- |
| `name: "*"` | core's `useRenderTool` (renderer-only) | **unconditional — wildcard always wins**, even with old tool fields present |
| carries `handler` or `description` | `useFrontendTool` | that is what the old hook actually did |
| anything else | core's `useRenderTool` | renderer-only was the intent |

The wildcard rule is deliberately not the obvious reading. The old hook made `description` **required**, so anyone who ever attempted a wildcard renderer necessarily wrote it *with* the old tool fields. Routing "has old fields" to `useFrontendTool` would have recreated the original `*`-named-tool bug for exactly the people who tried hardest to use the wildcard.

**Why a shim rather than a clean break.** `@copilotkit/react-native` sits in the `monorepo` release scope — 16 packages versioned in lockstep, all currently `1.68.1`. A major would take every `@copilotkit/*` package to `2.0.0`. So this ships in a minor and the `!` marker buys a release-note line, nothing more. That leaves the compiler as the only signal reaching consumers, and there are shapes it cannot see: a hoisted config whose `render` ignores its props, old fields arriving via a spread, and untyped or `@ts-nocheck` call sites. In all three, core's hook silently ignores `description` and `handler`, and because the bridge spreads `{ ...props, parameters: props.args }`, an old `render: ({ args }) => …` keeps painting — the screen looks unchanged while the tool stops being advertised and the handler never runs. The shim's warning is the only mechanism that reaches that population.

**What the shim does not restore:** the old *render prop* name. A renderer reading `args` still fails `TS2339`. The registration keeps working; the render body still needs the rename.

## ⚠️ BREAKING — needs product ratification

Removes `RenderToolProps`, `RenderToolFunction` and `UseRenderToolOptions` from `@copilotkit/react-native`. `useRenderTool` is deprecated, not removed, and is scheduled for removal next minor.

**@tylerslaton — this needs your explicit sign-off.** @BenTaylorDev has clarified that his deferral on #6438 was not a precedent, and would rather both removals be ratified together.

### Migration

| Before (RN) | After |
| --- | --- |
| `useRenderTool({ name, description, parameters, handler, render }, deps)` | `useFrontendTool({ …identical object }, deps)` — the shim keeps this working with a warning this release |
| renderer-only registration: impossible | `useRenderTool({ name, parameters, render, agentId? }, deps)` |
| wildcard: registers a tool named `*` | `useRenderTool({ name: "*", render })` — works, same as web |
| render props `{ args, status, … }` | `{ parameters, status, … }` — **not** covered by the shim |
| `RenderToolProps<T>` (args-shaped) | `RenderToolProps<S>` (parameters-shaped, generic over schema) |
| bare `RenderToolProps` | now `TS2314` — core's has no default type argument; supply the schema type |
| `UseRenderToolOptions<T>` | gone — write the config inline (`RenderToolConfig` is core-internal) |
| `RenderToolFunction<T>` | gone — see `FrontendToolRenderFunction<T>` below |

The table also lives on the docs page, deliberately: the release-note collector reads `git log --format=%H %s` into a `{ hash, subject }` type, so a `BREAKING CHANGE:` footer has nowhere to land (#6479).

## Restoring the React Native render-type safety

Ben caught a property the migration silently traded away. RN's deleted `RenderToolFunction` was the only thing narrowing an RN render to `React.ReactElement | null`. `useFrontendTool`'s `render` is a `React.ComponentType`, so it returns `ReactNode` — meaning a bare string now compiles and then throws *Text strings must be rendered within a `<Text>` component* on device. Before the rename that call site was rejected.

`FrontendToolRenderFunction<T>` is now exported from `/headless` for that. It derives its props from core's canonical `ReactToolCallRenderer` unchanged and narrows only the return type, so it cannot drift from the contract. It is **opt-in** — annotate your renderer to get the protection — and the hazard is now documented on `useFrontendTool.mdx`, which previously had zero mentions of `ReactElement`, `ReactNode` or `FlatList`.

## `status` changes vocabulary, and that is NOT a break

RN's `status` moves from the `ToolCallStatus` enum to core's string-literal union. A string-enum member is assignable to its own literal type, so `status === ToolCallStatus.Complete` compiles **and narrows** against the union — verified in all four combinations. No migration work follows.

This PR also deletes a false claim #6438 shipped: that comparing the enum to a bare string literal does not typecheck. Ben raised exactly this in #6438's review; it was corrected in the migration note and left standing in the guide bullet.

## The react-core half (2 files, 6 lines)

`ReactToolCallRenderer.render` is a `React.ComponentType`, and a function component returns `ReactNode` — of which `null` is a member. So `null` already worked where a renderer is *invoked*; only the two build-side helpers forbade it. Widened to `React.ReactElement | null` in `use-render-tool.tsx` and `types/defineToolCallRenderer.ts`. Both are required — widening only the hook yields `TS2322`.

Three things this does **not** claim:

- **Not type-only.** oxlint's `consistent-type-imports` converted `defineToolCallRenderer.ts`'s `ToolCallStatus` import to `import type`, removing a runtime `@copilotkit/core` import from the emitted JS. Harmless, but not a types-only diff.
- **Not uniform.** `useDefaultRenderTool` still requires `ReactElement`, so core's two wildcard entry points differ. That divergence fails loudly (`TS2322`) and gets its own change.
- **Not web-only in reach.** `defineToolCallRenderer` is re-exported from `@copilotkit/react-native`.

It lives in its own commit and lifts out cleanly if you would rather this PR stay RN-only.

## Two symptoms are documented, not fixed

- **`agentId` re-registration** — both core hooks omit `agentId` from the effect's dependency check, so the `deps` workaround stands.
- **`useDefaultRenderTool`'s narrower render return**, per above.

## Testing

- **Red-green on the wildcard.** Before the deletion, `getTool({ toolName: "*" })` returned a real tool object — `{ name: "*", description: undefined, parameters: undefined }` — the defect verbatim. Green after: the `"*"` renderer paints for an unmatched call **and** no tool named `*` is registered. The shim has its own version of this test, since its wildcard rule exists to prevent that exact regression. Note per the correction above that the **registry** assertions are the discriminating ones; the `advertised` assertions are kept as forward guards against someone making `*` advertisable later, and their comments now say so rather than claiming to be load-bearing.
- **All three shim routing rules**, asserted on core's observable state — `getTool`, `core.tools`, and the advertised tool list core hands an agent on a real run.
- **The identity guard was replaced, not weakened.** `headless-entry-surface.test.ts` previously asserted only that `useRenderTool` was *present*, never which hook it was, so it would have stayed green through this entire change. The identity assertion closed that; the shim made identity false, so it is now a delegation test plus a graph-wide deny-list on `addTool` / `removeTool` / `addHookRenderToolCall` / `renderToolCalls` / `createContext`. Five mutations were run against it: re-growing a local registry fails both new guards; an early-return shim that still *looks* delegating fails 14 behavioural tests; removing the wildcard rule reproduces the original bug; removing the production gate and the warn-once dedup each fail their own test.
- **A test whose name asserted the opposite of its behaviour** was replaced — it claimed to prove agent scoping and never checked that a scoped renderer does not paint under the default agent.
- The RN suite was rewritten around what RN owns. The old file tested forwarding through a double so thin that its own header comment recorded that deleting the `deps`/`handler`/`agentId` forwarding left it fully green.
- **A coverage pin was re-homed, not dropped.** An RN test pinning core's `JSON.stringify(extraDeps)` comparator was testing core, and no react-core test covered the case.
- **`FrontendToolRenderFunction` is mutation-proved both ways** — widening the return to `ReactNode` or loosening the props each break the type test.
- `packages/react-native/USAGE.md` was still teaching the deleted API. It **ships in the published tarball** (no `files` array, no `.npmignore`).

### Results

```
check-types (--skip-nx-cache)   2 projects + 34 dependency tasks   ✅
react-native vitest             298 passed / 23 files              ✅
react-core src/v2 vitest        1503 passed / 120 files            ✅
react-core test:scripts         47 pass / 0 fail                   ✅
react-native test:scripts       26 pass / 0 fail                   ✅
```

vitest transpiles without type-checking, so `tsc --noEmit` is the only guard against a re-narrowing of the six widened annotations. `static_quality.yml`'s `check-types` job runs it, and `nx.json`'s `targetDefaults.check-types.dependsOn` builds react-core's dist first, so the RN-side typecheck is meaningful rather than stale.

### What is NOT covered

Worth stating plainly: **nothing here proves the package works on a real device.** The RN suite runs in jsdom with `react-native` aliased to a 52-line stub whose `FlatList` is a mock; no CI job runs Metro, a simulator, or Detox; and `examples/v2/react-native/demo` is not built or typechecked in CI. What this PR verifies is types, export surface, and behaviour against core's real registry. Separately, #6438's finding that RN's dist-relative `sideEffects` globs strip the polyfill import from the built entries is still live and still out of scope.

- Widen `use-default-render-tool.tsx`'s four render annotations, or document why core's two wildcard entry points differ.
- Remove the shim next minor — tracked in #6976 and Linear OSS-1148, and cited in the shim's JSDoc.
- Build or typecheck the RN example app in CI — the cheapest real-RN signal available.
- `assert-headless-purity.mjs`'s `isCompleteLiteralArgument` reads `code` rather than `masked` — @BenTaylorDev's non-blocking ask on #6438, never actioned.
- #6346 (@davidmckayv) is still open; Ben asked for it to be closed with a pointer when #6438 landed.

## Summary by CodeRabbit

- **New Features**
  - Renderers can now return `null` to intentionally suppress tool-call UI.
  - React Native supports renderer-only tool-call registration, including named and wildcard renderers.
  - Added type guidance for React Native-safe renderer return values.

- **Bug Fixes**
  - Improved compatibility routing for existing React Native `useRenderTool` configurations.
  - Prevented unnecessary renderer re-registration when dependencies are non-serializable.

- **Documentation**
  - Updated React Native guidance for `useFrontendTool`, `useRenderTool`, shared rendering, migration, and deprecation behavior.
  - Clarified renderer props, status values, and `null` rendering behavior.


BenTaylorDev (commented):
Read this closely and verified the load-bearing claims myself rather than taking them from the description. The change is correct and the shape is right: RN's `useRenderTool` really was `useFrontendTool` wearing the other hook's name — `packages/react-native/src/hooks/useRenderTool.ts` on main destructures six fields and forwards straight through — so `name: "*"` did register a frontend tool literally named `*` and advertise it. Deleting rather than re-pointing is the right call, and the reasoning for it holds up. I found no defect in the code paths under change.

Two asks before it lands, then the product gate.

## 1. The dev-mode warning should land with the removal, not after it

The three call shapes you document as escaping the compiler (hoisted config, spread-carried fields, untyped/`@ts-nocheck`) don't just fail to error — they change behaviour invisibly. Core's `useRenderTool` ignores `description` and `handler` without comment, and because the bridge spreads `{ ...props, parameters: props.args }` at `use-render-tool.tsx:170`, an old `render: ({ args }) => …` keeps painting exactly as before. The tool stops being registered and advertised, the handler never runs again, and the screen looks unchanged.

The reason I'd rather not defer this: **`react-native` has no semver channel to carry the break either.** It's in the `monorepo` lockstep scope — `.github/workflows/stable-release.yml` offers `monorepo | angular | channels`, `bump` is a manual choice applied to the whole scope, and `@copilotkit/react-native` and `@copilotkit/react-core` are both `1.68.1` on npm. A `major` here would take every `@copilotkit/*` package to `2.0.0`, which isn't going to happen. So this ships in a minor, and the `!` marker buys a release-note line and nothing else.

That leaves the compiler as the only signal reaching a consumer — and the shape it can't see is precisely the one `packages/react-native/USAGE.md` and the old reference page taught (`{ name, description, parameters, handler, render }`). Plain-JS RN screens are common. The ~4 lines you've already scoped — a dev-only `console.warn` in core's `useRenderTool` when `config` carries `description`/`handler`/`followUp`/`available` — are cheap, in-scope for a PR already editing that file, and the only mechanism that reaches that population.

If product would rather have a softer landing, the other option is an RN shim kept for one minor that discriminates on `handler`/`description`, routes to `useFrontendTool`, and warns. That isn't the silent re-point you rightly rejected: it's loud, and it doesn't change anyone's behaviour without telling them.

## 2. "Rename only" quietly loses the RN-safe render return type

`RenderToolFunction` (`packages/react-native/src/hooks/render-tool-types.ts:53-55`, deleted here) was the only thing narrowing an RN render function to `React.ReactElement | null`. The migration table sends people to `useFrontendTool`, whose `render` is `ReactToolCallRenderer<T>["render"]` (`packages/react-core/src/v2/types/frontend-tool.ts:7`) — a `React.ComponentType`, so it returns `ReactNode`. A bare string now compiles:

```ts
const r: ReactToolCallRenderer<{ city: string }>["render"] = (p) =>
  `Weather in ${p.args.city}`; // clean under tsc
```

and then throws *Text strings must be rendered within a `<Text>` component* on device. Before the rename that call site was rejected.

What makes this worth fixing rather than noting: this page argues for `ReactElement | null` over `ReactNode` on exactly that hazard — and `reference/react-native/hooks/useFrontendTool.mdx`, the page the migration points at, has zero mentions of `ReactElement`, `ReactNode` or `FlatList`. The narrowing is advertised on the hook being removed and absent from the one you're sent to. Either carry the note over to `useFrontendTool.mdx`, or keep an RN-safe render alias.

## 3. `RenderToolProps` lost its default type argument (low)

RN's was `RenderToolProps<T = Record<string, unknown>>`; core's is `RenderToolProps<S extends StandardSchemaV1>` with no default. So bare `RenderToolProps` — previously legal, and the natural spelling for an untyped renderer — is now `TS2314: Generic type 'RenderToolProps<S>' requires 1 type argument(s)`. Loud, so it just wants one more row in the migration table so the error is self-service.

## Nits

- `reference/react-native/hooks/useRenderTool.mdx:161` — `<PropertyReference name="parameters" type="Partial<T> | T">` still uses the old `T` after the page renames the generic to `S` and explicitly warns that `S` is the schema type, not the parsed argument object. Should be `Partial<InferSchemaOutput<S>> | InferSchemaOutput<S>`.
- `use-default-render-tool.tsx`: I'd just widen the four annotations here rather than shipping two wildcard entry points in core with different render contracts. It fails loudly, so I'm not going to insist — but it's four lines against a documented inconsistency.
- `headless.ts:118-138` — 21 lines of comment over two lines of re-export. The operative sentence is "do not reintroduce a local hook under either name"; the rest now lives on the docs page and in the identity test's own comment.
- `src/hooks/__tests__/useRenderTool.test.tsx` tests the entry wiring, not a hook this package owns. `src/__tests__/`, next to `headless-entry-surface.test.ts`, is a better home.

## Verification

Independent of the description's numbers:

- No dangling references to `hooks/useRenderTool`, `hooks/render-tool-types`, `RenderToolFunction` or `UseRenderToolOptions` anywhere in `packages/`, `showcase/` or `examples/`.
- `ToolCallStatus` in `defineToolCallRenderer.ts` appears only in type positions (`:19`, `:26`, `:33`), so `import type` is correct.
- The `null` widening is safe at every consumer: `renderer.render` is only ever used as a component (`use-render-activity-message.tsx:61`, `use-render-custom-messages.tsx:73`). Nothing calls it and dereferences the result.
- The three `render: () => null` tests aren't only documentation — `packages/react-core/tsconfig.json` includes `src/**/*`, so `tsc --noEmit` typechecks them and a re-narrowing does fail `check-types`.
- `tsc --noEmit` on both packages against a freshly built react-core dist: no new type errors. The 16 RN errors are missing optional peer deps and are identical on base; base had two *additional* real errors from a stale dist.
- Mutation check on the new guard: re-pointing `export { useFrontendTool as useRenderTool }` in `headless.ts` kills four tests including the new identity assertion. It's load-bearing.
- Confirmed with tsc 5.9.2 that all four enum/string-literal combinations compile *and* narrow — the `status` claim and the deletion of the opposite claim are both right.
- Pre-existing and unrelated: 8 react-core failures (`pinToSend`, `use-interrupt`) and 90 RN component failures (local `react-native`/`react-dom` frozen-style proxy incompatibility), all identical on base.
- The dead-barrel claim checks out: `src/hooks/index.ts` is absent from `tsdown.config.ts` entries and from `package.json#exports`, and nothing imports it.

The rewritten tests are the strongest part of this. `agent.advertised).toEqual([[]])` instead of `not.toContain`, real `CopilotKitCoreReact` instead of a double, and an identity assertion that closes a guard which would otherwise have stayed green through the entire change.

## On #6438

Fair point, and I'm not going to relitigate it here. My read then and now: the code was ready and the removal wasn't mine to sign off. It merging 25 minutes later doesn't turn that deferral into a precedent — so I'd rather both get ratified than treat the first one as having settled the question. @tylerslaton, this needs your call on removing `useRenderTool`, `RenderToolProps`, `RenderToolFunction` and `UseRenderToolOptions` from `@copilotkit/react-native`, and it's worth deciding #6438's removals in the same breath.

Leaving this as a comment rather than an approval — the two asks above, and Tyler's sign-off on the removal itself.

mxmzb:
@BenTaylorDev — both asks landed, and the lockstep-scope argument is what decided the first one. Verified before acting on it: `release.config.json` puts `@copilotkit/react-native` in the `monorepo` scope alongside 15 others, `angular` and `channels` are separate, and everything in that scope is on `1.68.1`. So you're right that there's no channel to carry this, and the compiler is the whole signal.

Rebased onto current `main` first — the branch had gone 1273 commits stale, which is also what was failing `langgraph-python` earlier (the base predated #6520's LangGraph content-part fix; green since the rebase).

## 1. Shim, not a bare warning

Went with your second option rather than the dev-warn-in-core one. Maxim's call, and the reasoning was that a warning in core reaches the same population but changes react-core's observable behaviour for web too, whereas the shim keeps the whole thing inside React Native.

`useRenderTool` is back on RN as a `@deprecated` shim scheduled for removal next minor. Routing, and the third rule is the one worth checking:

| shape | routes to |
| --- | --- |
| `name: "*"` | core's `useRenderTool` — **unconditional**, even carrying `handler`/`description` |
| carries `handler` or `description` | `useFrontendTool` |
| anything else | core's `useRenderTool` |

The wildcard rule is deliberately not "old fields → `useFrontendTool`". The old hook made `description` **required**, so anyone who ever attempted a wildcard renderer necessarily wrote it with the old tool fields — routing on field presence would have recreated the `*`-named-tool bug for exactly the people who'd tried hardest to use the wildcard. Wildcard wins first, then it warns about the ignored fields. There's a test asserting a `"*"` call carrying both fields paints as a fallback *and* leaves `getTool({ toolName: "*" })` undefined.

Warning is dev-only and deduped per tool name. Worth flagging that only the *warning* is `NODE_ENV`-gated — routing is unconditional, so a production build loses the notice, not the behaviour. The genuinely silent path returns when the shim is removed, and the docs now say that rather than implying the shim solved it.

**Two things the shim does not restore**, both verified with tsc and both now documented: the old *render prop* name (a renderer reading `args` still fails `TS2339` — registration keeps working, the render body still needs the rename), and the old types.

### On the identity guard

You called it the strongest part of the PR, and the shim makes `mod.useRenderTool === core.useRenderTool` false, so it needed replacing rather than deleting. `useFrontendTool`'s half is kept verbatim. `useRenderTool`'s is now a delegation test — it reads which module the entry exports the name from, and when that's a local module it requires that module to import *and call* both core hooks — plus a graph-wide deny-list on `addTool` / `removeTool` / `addHookRenderToolCall` / `renderToolCalls` / `createContext`.

Mutation-tested five ways, since a guard that replaces a load-bearing one has to earn it:

- RN re-grows a local `addHookRenderToolCall` implementation with the core imports left in place → fails **both** new guards (2 failed / 19 passed)
- shim early-returns, source still *looks* delegating → fails 14 behavioural tests, including all three routing rules
- wildcard rule removed → reproduces the original bug verbatim: `expected { name: '*', …(4) } to be undefined`
- production gate removed → fails the production test
- warn-once dedup removed → fails once-per-name

That last one initially didn't bite: the effect's dependency array was doing the dedup's work, so the module-level `Set` wasn't load-bearing. The test now remounts, which makes it so. Second commit exists for that reason.

## 2. The render-return narrowing

Confirmed your read before fixing it — the `ComponentType` render returning a bare string compiles clean, and the deleted narrowed form is the only one that errors. So "rename only" was wrong in the table, and the migration was silently trading away crash prevention.

`FrontendToolRenderFunction<T>` is now exported from `/headless`. Props come from `React.ComponentProps<ReactToolCallRenderer<T>["render"]>` unchanged and only the return is narrowed, following the derivation precedent from the file this PR deleted — so it can't drift from the contract the way RN's types did before. Named for the hook it annotates rather than reviving `RenderToolFunction` against different props, since that name coming back with a different meaning is the defect class this PR is about.

Mutation-proved both directions: widening the return to `ReactNode` gives 4× `TS2578`, loosening the props gives `TS2322` + 8× `TS18046`, and props → `any` gives `TS2578` — that last one is why the test asserts the props are core's, not just the return type. react-native's tsconfig has `include: ["src"]`, so `tsc --noEmit` compiles the type test, and vitest's glob doesn't collect it.

Did both halves of your ask rather than choosing: the hazard is now documented on `useFrontendTool.mdx` too, which had zero mentions of `ReactElement`, `ReactNode` or `FlatList`. Being straight about the limit — the type is **opt-in**, so someone who doesn't annotate gets nothing. If you'd rather it were enforced, that means RN wrapping `useFrontendTool` with the narrow type baked in, which is another RN hook and cuts against the PR. Your call whether that's worth reopening.

## 3 and the nits

- **`RenderToolProps` default type argument** — added the migration row. Reproduced the `TS2314`.
- **`Partial<T> | T`** — fixed, now expressed through `InferSchemaOutput<S>`. Our own review had flagged this and I graded it cosmetic drift; it's a wrong type on a reference page, so that was too lenient.
- **`headless.ts` comment** — trimmed to the operative instruction.
- **Test-file location** — left in place. The shim means that directory now legitimately tests a local hook again. One file there reads through the entry rather than the local module; flagged, not moved.
- **`use-default-render-tool.tsx`** — not widened, left as a follow-up. Not disagreeing with you that it's four lines; it's four annotations plus an existing `as unknown as` cast, and it's web-only and absent from RN's entry, so it seemed wrong to attach to a breaking RN change. Happy to be overruled.

## What this still doesn't prove

Your 90 local component failures are the only real-RN signal anyone's produced on this branch, and that's worth saying out loud: the RN suite runs in jsdom with `react-native` aliased to a 52-line stub whose `FlatList` is a mock, no CI job runs Metro or a simulator, and the RN example app isn't built or typechecked in CI. So this verifies types, export surface, and behaviour against core's real registry — not that anything renders on a device. Added to the PR body under "What is NOT covered", along with the still-live `sideEffects` polyfill-stripping finding. Building the example app in CI is on the follow-up list as the cheapest way to get an actual signal.

Numbers at `c84c0cc047`, rebased: react-native vitest **298/298** across 23 files, react-core `src/v2` **1503/1503** across 120 files, `test:scripts` **47/0** and **26/0**, `tsc --noEmit` clean in both packages.

## On #6438

Taking your framing over mine — I'd characterised the 25-minute merge as a precedent in the original PR body, and that was unfair to a deferral you made explicitly. Body's rewritten to say your deferral wasn't a precedent and to ask for both removals to be ratified together. @tylerslaton, that's the ask: removing `RenderToolProps`, `RenderToolFunction` and `UseRenderToolOptions` from `@copilotkit/react-native`, with `useRenderTool` deprecated rather than removed, plus #6438's removals in the same decision.

BenTaylorDev (commented):
Re-reviewed at `c84c0cc047`, rebased and 0 commits behind `origin/main`. CI fully green — six unit shards, `check-types`, `oxlint`, `format`. Both asks landed and both landed well. I tried to break the shim and could not: the frozen route makes the conditional hook calls safe for a component's lifetime, removing the wildcard rule reproduces the original bug verbatim (`expected { name: '*', …(4) } to be undefined`, 2 failures), and re-narrowing the six `ReactElement | null` widenings produces 4 errors in the new tests and nothing anywhere else. RN suite 298/298 in a clean install.

Three things to fix. The middle one changes the argument for the change rather than the change itself, and it takes two assertions with it.

## 1. There is a NUL byte in the shim

`packages/react-native/src/hooks/useRenderTool.ts:331`:

```ts
warnOnce(
  `${config.name}\x00route-drift`,
```

Byte offset 14224 is `0x00`, not a space. Confirmed in the pushed blob, not just locally:

```
$ gh api ".../contents/packages/react-native/src/hooks/useRenderTool.ts?ref=c84c0cc047" --jq .content | base64 -d
bytes: 17392   NUL offset: 14224   line 331
all control bytes: [(14224, '0x0')]
```

It renders as a space in every diff view, which is why review has not caught it. Two consequences:

- `rg` reports `binary file matches (found "\0" byte around offset 14224)` and prints **no lines**. So `rg` over `packages/` silently returns nothing from this file — including for the very identifiers `headless-entry-surface.test.ts` polices by text. `file` agrees it is `data`.
- `git diff` still shows text, and I checked why rather than assuming: git's binary check reads only the first 8000 bytes. A NUL at index 7999 makes `git diff --numstat` report `-/-`; at index 8000 it reports `1/1`. This one sits at 14224, past the window.

The key becomes `"shapeShifter\0route-drift"` and the literal ships in the bundle. Nothing tests it; `oxlint` and `format` both pass over it. One character.

## 2. "advertised it to the model" is not true on current `main`

This is the PR's headline mechanism, and it is the half the 1273-commit rebase invalidated.

I measured it against the old hook itself — `origin/main`'s `useRenderTool.ts` and `render-tool-types.ts` vendored in verbatim, driven through your own `ToolListRecordingAgent`, with dists built from this branch's source:

```
                            in core.tools   advertised                warns
A  old hook, wildcard          true         [[]]                        0
B  old hook, no description    true         [["legacyJsShape"]]         0
C  shim, no description        false        [[]]                        0
D  shim, wildcard + old fields false        [[]]                        1
```

Row A is the finding: the old hook did register a frontend tool named `*` — that part is right and it is a real defect — but core never offered it to the model. `packages/core/src/core/run-handler.ts:1281` filters the name out:

```ts
// A wildcard tool is a local catch-all handler (see
// `executeWildcardTool`), not something the model can call —
// advertising it would offer the agent a tool named `*`.
tool.name !== WILDCARD_TOOL_NAME &&
```

I proved that line is the cause rather than inferring it: patching only that one comparison out of the built dist flips row A from `[[]]` to `[["*"]]`, and leaves row B unchanged. The filter arrived in `31aa1e2162` at 2026-08-17T17:41Z, about two hours before this PR was opened — so the claim was written against a base that predated it, and the rebase moved the code forward without moving the prose. I repeated the same claim in my own review last round; that was my error too.

Seven sites, plus the PR body, which is what the release note and the docs page inherit:

- `packages/react-native/src/hooks/useRenderTool.ts:26`
- `packages/react-native/src/headless.ts:132`
- `packages/react-native/src/__tests__/headless-entry-surface.test.ts:630`
- `packages/react-native/src/hooks/__tests__/useRenderTool.test.tsx:33`, `:136`, `:345`
- `showcase/shell-docs/src/content/reference/react-native/hooks/useRenderTool.mdx:545-547`

**Two assertions rest on it, and they are vacuous.** `useRenderTool.test.tsx:158-166`:

```ts
// …and the same thing observed one layer out, where it would actually hurt:
// the tool list core hands the agent for a real run.
await act(async () => { await core.runAgent({ agent }); });
expect(agent.advertised).toEqual([[]]);
```

Row A is what that assertion returns on `origin/main` too, so it cannot fail for the reason the comment gives — the thing it is presented as catching was already impossible. `:373` is the same shape in the rule-1 test, under `// Half two: nothing named \`*\` is a tool, in the registry or on the wire`. The registry halves at `:155-156` and `:368-369` are the discriminating ones — my wildcard-rule mutation kills exactly those. Keep both `advertised` assertions if you like, as a guard against someone making `*` advertisable later, but the comments should stop claiming they are the ones that would hurt.

**The true mechanism is stronger, and it is the better case for the change.** A tool named `*` is core's catch-all *handler*. When a tool call has no matching frontend tool and no result yet (`existingResultIndex === -1`), `run-handler.ts:731-741` reaches for `getWildcardTool()` and runs `executeWildcardTool` — and the tool-result splice at `:1092-1117` sits **outside** the `if (wildcardTool?.handler)` guard, which closes at `:1090`. So an RN app whose author wanted a display-only fallback was also auto-answering every otherwise-unanswered tool call with an empty tool result, and requesting a follow-up turn. That is worse than one extra entry in a tool list, and the docs' "changed behaviour for the better" is currently understating it while explaining it wrongly. Stated as a code path: I traced the branch and the guard's extent, but I did not drive a full agent turn through it, and I am not claiming it intercepts server-side calls whose results already arrived — those do not reach the branch.

## 3. The last silent path is real, and the docs say it is not

`showcase/shell-docs/src/content/reference/react-native/hooks/useRenderTool.mdx:628-630`:

> **Only when an old field is actually present.** A call with neither `description` nor `handler` is routed to react-core's `useRenderTool` and says nothing, because there is nothing to say.

There is something to say. `description` was required on the old hook, so TypeScript callers could not reach this shape — but untyped JS could, and untyped JS is the population the shim exists for. `FrontendTool.handler` is optional (`packages/core/src/types.ts:77`) and `buildFrontendTools` does not filter on it, so the old hook registered *and advertised* that tool. Rows B and C above are exactly that comparison: `[["legacyJsShape"]]` before, `[[]]` after, **0 warnings**.

So a plain-JS screen that omitted `description` loses tool registration and advertisement and is told nothing — the precise failure the shim was built to prevent, reached through the one shape it cannot discriminate. Routing cannot fix it: `{ name, parameters, render }` is also the correct new renderer-only spelling.

The warning can. `warnRouted` returns early on `legacyFields.length === 0` (`useRenderTool.ts:229`). The whole hook is deprecated and goes away next minor, so warning unconditionally on the rule-3 route costs a line and closes the gap: "this registers a **renderer only**; if you relied on this call advertising a tool, use `useFrontendTool`." At minimum the bullet's rationale needs correcting — as written it tells the next maintainer the gap is empty.

## Nits

- The two sentences explaining the per-status spread appear twice, in the JSDoc at `:378-383` and again in the body at `:387-390`. Drop one. The three branches at `:391-397` being byte-identical mirrors core's own shape, so I am not asking you to collapse those.
- The rule-3 route changes the effect phase: `useFrontendTool` registers in `useLayoutEffect`, core's `useRenderTool` in `useEffect`, so a call that used to land in the layout phase now lands one phase later. Only untyped-JS callers are affected (same population as finding 3) and it matches core's web behaviour, so it is a docs line in the route table, not a defect. Worth one, given we have been bitten by layout/passive splits in this registry before.
- "removal in the next minor" appears 11 times across 5 files with no issue behind it. That is how a shim becomes permanent. File the removal issue and cite the number in the JSDoc.

## Verification

- 0 commits behind `origin/main`; no `.changeset/`, no retired-tree paths, nothing outside the 25 files.
- `tsc --noEmit` clean on `@copilotkit/react-native` against dists built from this branch — including `frontend-tool-render.test-d.tsx`, so all **7** `@ts-expect-error` directives are load-bearing.
- Mutations I ran myself: wildcard rule removed → the original bug, 2 failures; `ReactElement | null` re-narrowed → 4 errors, all in the new tests, none elsewhere; core's name filter removed from the dist → row A flips, which is what makes finding 2 causal rather than correlational.
- The frozen route is sufficient: both of the shim's `useCoreRenderTool` call sites reach the same hook, and core's `useRenderTool` is one `useCopilotKit` plus one `useEffect`, so the conditional calls cannot change hook order for a live component.
- Docs nit from last round fixed: `Partial<InferSchemaOutput<S>> | InferSchemaOutput<S>` at `useRenderTool.mdx:196`.
- The runtime warning text at `:254-256` is accurate as written — it says the old hook "did register one", not that it was advertised. No change needed there.
- Local `react-core` `src/v2` failures I hit are environmental (mixed `@ag-ui/client` copies in my hand-built tree); CI's six shards are the signal and they pass.

Fix 1, the prose and the two comments in 2, and the rationale in 3, and the code is done from my side. @tylerslaton, the product gate is unchanged: removing `RenderToolProps`, `RenderToolFunction` and `UseRenderToolOptions` from `@copilotkit/react-native`, with `useRenderTool` deprecated rather than removed — and #6438's removals in the same decision.

mxmzb:
@BenTaylorDev — all three fixed at `e69a7c08e2`, rebased onto current `main`. Finding 2 was the one worth catching and I've gone a step further on it than you did, which changed the wording again.

## 1. NUL byte — fixed

`\x00` → `:`. Verified at the byte level rather than visually, since visual review is exactly what it defeats:

```
scan for bytes <0x20 except tab/LF/CR:   [(14224, 0)] → []
file:                                     data → UTF-8 text
rg over the file:                         0 lines → 28 matching lines
```

All 25 files in the diff were scanned; this was the only one with a control byte.

One addition to your two consequences: `git diff | grep` also silently returns nothing, because the raw NUL makes grep treat the *diff stream* as binary even though git renders it as text. `cat -v` is what exposes it. So the byte was invisible to the diff view, to ripgrep, and to grep-over-diff simultaneously.

## 2. The mechanism — reproduced, then driven

Reproduced your rows before changing any prose, by vendoring `origin/main`'s hook in verbatim through the branch's own recording agent:

```
                             in core.tools   advertised             warns
A  old hook, wildcard           true         [[]]                     0
B  old hook, no description     true         [["legacyJsShape"]]      0
C  shim, no description         false        [[]]                     0
D  shim, wildcard + old fields  false        [[]]                     1
```

Matches yours on all four. I also isolated the *name* as the sole cause without touching core, which is a cleaner control than patching the dist: two configs identical except for the name give registry `["*", "notAWildcard"]` and advertised `[["notAWildcard"]]`.

**You had traced the wildcard-handler path but not driven it. I drove it.** The old handler-less `*` produces **2 turns and a spliced `{ toolCallId, content: "" }`**; the shim's route produces 1 turn and no result. The guard opens at `:1000` and closes at `:1089`; the splice sits at `:1091-1120`, outside it. So the mechanism is now measured rather than inferred, and the bound you were careful about — no exact-name tool, no result yet — is stated at every site. Server-side calls whose results already arrived never reach the branch, as you said.

Corrected at all seven sites plus the PR body, which now carries the correction explicitly rather than quietly reading better.

**The two vacuous assertions.** Both `advertised` checks kept, comments rewritten as forward guards against someone making `*` advertisable later; the registry pair is labelled as the discriminating one. Confirmed the way you'd want: removing rule 1 kills the registry assertion and leaves *both* `advertised` assertions green — which is the property that made them vacuous, now demonstrated rather than argued.

One correction back: the gap between `31aa1e2162` and this PR opening is about 5h39m, not two hours. Doesn't touch the substance, and no timing figure went into code or docs.

## 3. The last silent path — warning now unconditional

You're right that the bullet told the next maintainer the gap was empty. The renderer-only route now warns unconditionally: says the call registers a **renderer only**, and points anyone relying on advertisement at `useFrontendTool`. Deduped per name and dev-only like the others. The test that asserted silence on that route is updated rather than left asserting the old behaviour, and the docs bullet's rationale is rewritten.

Side effect worth naming: the warning now also fires for already-correct renderer-only calls that are still imported from `react-native` rather than `react-core`. Intended, and the message says so — the whole hook is deprecated, so "you are on the shim" is useful information either way. Say the word if you'd rather it stayed quiet for that case.

## Nits

- Duplicate per-status explanation — one copy dropped. Three identical branches left alone, per your note.
- Effect-phase shift — documented as a row in the route table. Verified both hooks' effect types in current source first.
- Removal tracking — filed as #6976 and Linear OSS-1148, cited in the shim's JSDoc, the entry-point comment, and the reference page. The runtime warning string is deliberately unchanged, since a test asserts on its text.

Two things I'd flag rather than have you find:

- The drift key is now `"*:route-drift"`, so a tool literally named `*:route-drift` would collide. Theoretical, and the old NUL had the same shape.
- `useDefaultRenderTool` still isn't widened. Not disagreeing with you on the four lines — it's four annotations plus an existing `as unknown as` cast in a file this PR doesn't otherwise touch, and it fails loudly. On the follow-up list; happy to be overruled if you'd rather it rode along.

## Numbers at `e69a7c08e2`

react-native vitest **300/300** across 23 files at default settings (+2 from the finding-3 tests), react-core `src/v2` **1503/1503** across 120 files, `test:scripts` **47/0** and **26/0**, `tsc --noEmit` clean in both packages, 0 commits behind `origin/main`.

Still not proven on a device — same caveat as last round, and it's in the PR body under "What is NOT covered". Building the RN example app in CI is on the follow-up list as the cheapest real signal, since your 90 local component failures remain the only real-RN data anyone has produced here.

@tylerslaton — the product gate is unchanged and it's the only thing left: removing `RenderToolProps`, `RenderToolFunction` and `UseRenderToolOptions` from `@copilotkit/react-native`, with `useRenderTool` deprecated rather than removed, and #6438's removals decided in the same breath.

BenTaylorDev (approved):
All three fixed, verified at `e69a7c08e2`. I reproduced both of your new claims rather than taking them, and I owe you a correction that goes the other way.

## Verified

**1. NUL byte.** Gone. I scanned the pushed blobs rather than the rendered diff: all 25 files in the diff, zero bytes below `0x20` outside tab/LF/CR. Drift key is `"*:route-drift"` at `:423`.

Your addition about `git diff | grep` is right and worth keeping in mind — the byte defeated the diff view, ripgrep, and grep-over-diff at once, which is three of the four ways anyone would have looked.

**2. The mechanism.** Your four rows match mine exactly. Your name-only control is a better instrument than my dist patch, and I agree: two configs differing only in the name is a cleaner isolation than mutating core.

I drove the turn independently too, and got your numbers:

```
                     `*` in registry   turns   spliced tool messages
OLD hook wildcard          true          2      [{"toolCallId":"tc-unclaimed","content":""}]
SHIM     wildcard          false         1      []
```

So the corrected prose is measured now, at every site, with the bound stated. That is a real improvement on what I handed you — I had traced the branch and stopped there.

**The two assertions.** Confirmed the way you said: with rule 1 removed, the failures are `getTool` and one warning-text assertion, and **both** `advertised` assertions stay green. That is the vacuity property demonstrated rather than argued, and the comments now say which pair is which. Right call keeping them as forward guards.

**3. The silent path.** Unconditional now, and the guard you added that I did not ask for is the right one: `route !== "useRenderTool"` before the renderer-only notice, so a frozen `useFrontendTool` route that later drops its legacy fields does not get told it is renderer-only. That case would have been a false message and I would have missed it.

The two new tests wait on the registration rather than on the warning, and the drift test now waits for the drift message specifically instead of a bare `toHaveBeenCalled()` that the new notice would have satisfied. Both of those are the vacuity trap in miniature and you closed them unprompted.

300/300 across 23 files and `tsc --noEmit` clean, both reproduced locally against dists built from this branch. #6976 exists and is cited in the JSDoc, the entry comment and the reference page.

## The timing — my error, and it clears your original wording

You corrected 2h to 5h39m. Both figures are wrong, and the reason matters: neither the author date nor the committer date tells you when a commit reached `main`.

`31aa1e2162` was authored 2026-08-17T14:17Z and committed 17:41Z, but it landed on `main` only through the merge of #6524 at **2026-08-31T18:00Z**:

```
main tip when #6533 opened (2026-08-17T19:56Z)  → 5b9776a1a0, filter present: 0
main tip after 2026-08-31                        → fe51cec465, filter present: 1
3ee0ec189c^1 (main just before the merge)        → filter present: 0
3ee0ec189c   (#6524, the merge itself)           → filter present: 1
```

So the gap is **14 days, after the fact** — not hours before it. Your original "advertised it to the model" was **correct when written**; #6524 invalidated it two weeks later, and the rebase brought the code forward without the prose. My review said the claim "was written against a base that predated it", which was wrong and unfair to you. I used the committer date of a commit that had not merged yet; withdrawn.

It does not change what needed fixing — the claim was false at `c84c0cc047` and it is right that it is gone — but it changes whose mistake it was, and this is the stale-branch pattern in its purest form rather than a mis-read of core.

## Your open question: keep the warning on already-correct calls

Keep it. The costs are asymmetric: a correct renderer-only caller loses one deduped dev-console line, while the caller this cannot be distinguished from loses tool advertisement in silence. That asymmetry is the whole reason routing cannot solve finding 3, so paying it in noise is the right trade.

The one thing I would add is a clause saying no action is required before removal for that population — after the shim goes, the name resolves to core's hook and their code behaves identically, so "already correct" currently reads as an instruction to change an import that does not need changing. One sentence, not a blocker.

## The two you flagged

- `"*:route-drift"` collision — theoretical and I would not spend a character on it. A tool literally named `*:route-drift` has bigger problems.
- `useDefaultRenderTool` — agreed, follow-up. Four annotations plus an existing cast in a file this PR does not otherwise touch, and it fails loudly. Not worth attaching to a breaking RN change; I am not overruling you.

## One thing before merge

`Shell script tests (bats + shellcheck)` is red, and it is not yours: it failed inside `Install bats` with an `apt-get` hash-sum mismatch fetching `dl.google.com/linux/chrome-stable/.../Packages.gz`, exit 100. This PR touches no `.sh`, `.bash` or `.bats` file. Needs a re-run rather than a change. Everything else is green — 51 checks.

Also, `main` moved 6 commits after you pushed (PR #6961's stack merged about six minutes after your comment), so the branch reads as behind now. None of it is under `packages/`, so it changes nothing here — just so the number does not get read as staleness.

Approving on the engineering. The render-tool surface is in better shape than it was before either of us touched it. @tylerslaton, you are the only thing left: removing `RenderToolProps`, `RenderToolFunction` and `UseRenderToolOptions` from `@copilotkit/react-native`, with `useRenderTool` deprecated rather than removed, and #6438's removals decided in the same breath.

PR #6535, opened 2026-08-17
fix(showcase): validate fleet driver inputs in the payload mapper

contextablemark:
## Summary

The fleet worker ran per-service browser drivers with **no input validation**. `driver.inputSchema.safeParse` executes only in the legacy in-process invoker (`probes/loader/probe-invoker.ts:1415`); the fleet path calls `driver.run(ctx, input)` directly at `worker-loop.ts:1045`, so the schema never ran. Three docstrings asserted the opposite — *"each driver's OWN zod schema is the validation gate inside `driver.run`"* — which is what kept the gap invisible.

## Why it matters

The d6 schema declares `backendUrl: z.string().url()` (`d6-all-pills.ts:99`), which rejects `""`. A Railway service with no `serviceInstance` in the probed environment enumerates with `publicUrl: ""`, the producer serializes that as `backendUrl: ""`, and `${backendUrl}${route}` (`d6-all-pills.ts:1200`) then yields a **relative** path. No Playwright `baseURL` is configured anywhere in the harness, so every cell of that service failed `page.goto` with `Cannot navigate to invalid URL` — a whole-column red block of indistinguishable `goto-error` cells that said nothing about the demo.

`??` does not save it either: `input.backendUrl ?? input.publicUrl` (`d6-all-pills.ts:802`) does not fall back, because `""` is not nullish.

This is currently live on prod for `crewai-conversational-flows` (staging-only service, 40/40 cells failing).

## What changed

Each driver's own `inputSchema` now runs in the payload mapper, which **throws** on failure. The worker loop already converts a `payloadToInput` throw into a `worker-protocol-violation` carrying the thrown message (`worker-loop.ts:848-869`), so one service reports one precise zod error instead of N identical per-cell reds.

**Why the mapper and not the loop:** `ServiceJobDriver` is deliberately a one-method interface — *"so the loop doesn't drag the whole probe registry into the worker, and so a test fake is a one-method object."* Injecting the schema per registry entry preserves that and needs **no change to `worker-loop.ts`**. The parameter is a minimal structural `safeParse` interface rather than a zod import, so a test double is an object literal and a concrete `ZodType<SpecificInput>` stays assignable without variance friction. It is optional — omitted, the mapper is pure re-hydration exactly as before, so every pre-existing call site is unaffected.

Also corrects the three docstrings that claimed a gate that did not exist.

## Scope

This is **defense in depth, not the root cause**. A service with no URL in the probed environment should never reach job production in the first place — that fix is a separate PR against `railway-services.ts`.

## Test plan

- `payload-mapper.test.ts`: 7 new tests — schema-valid passthrough, the empty-`backendUrl` production case, driverKind named in the message, validation ordered after the `key` default, `undefined` still distinct from a throw, no-schema back-compat, and structural-`safeParse` acceptance. 15 tests in the file.
- The throw → `worker-protocol-violation` path is already covered at `worker-loop.test.ts:517,589`.
- Full harness suite green: **3738 passed / 18 skipped (179 files)**. (Two files need the gitignored generated `registry.json` / `frontend-catalog.json`; both pass after `tsx ../scripts/generate-registry.ts`.)
- `tsc --noEmit` clean.


contextablemark:
Root-cause PR is up: #6536 — scopes Railway discovery to the probed environment so an empty `backendUrl` is never produced in the first place. This PR is the gate that catches it if one ever is. Independent; either order.

PR #6538, opened 2026-08-17
feat: implement interrupt handling in AgentStore and add injectInterr…

### LINKED ISSUE #6509: 🚀 Feature Request: Angular — expose the interrupt controller on `AgentStore`
# Proposal: expose the interrupt controller on `AgentStore`

**Package:** `@copilotkit/angular` 0.3.0

## Problem

Store and interrupt controller are obtained through two independent injection functions
that are correlated only by a string:

```ts
@Injectable({ providedIn: 'root' })
export class TicketingChatService {
  private readonly store = injectTicketingAgentStore();      // injectAgentStore(TICKETING_AGENT_ID)
  private readonly interrupts = injectInterrupt({
    agentId: TICKETING_AGENT_ID,
  });
}
```

Consequences:

- The pairing is untyped and unenforced. Passing the wrong `agentId` to `injectInterrupt`
  produces a controller that simply never fires — no error, no warning.
- Both objects then travel together through the whole application, because anything that
  renders a chat needs both:

  ```ts
  export interface ChatConfig {
    store: Signal<AgentStore>;
    interrupts: InterruptController;
    // ...
  }
  ```

  Every layer that forwards a "current chat" has to carry a two-part tuple that is
  conceptually one thing.
- The agent id has to be exported from the store module purely so the interrupt call site
  can repeat it, even though `store.agent.agentId` already knows it.

Also, an interrupt is not an application-level concern that happens to mention an agent — it is a state of one specific
conversation: this agent, this thread, this run pauses and waits for a decision, and
resolving it resumes exactly that run. `AgentStore` is the object that represents that
conversation; it already exposes its messages, its shared state and whether it is
currently running. The interrupt is the same kind of per-conversation state, and it is the
one piece you cannot reach from it — you have to go back to an injection function and
re-name the conversation by its id to get at something the store is already holding.

Reaching the pending interrupt through the store is therefore the natural reading:
`store.messages()`, `store.state()`, `store.isRunning()`, `store.interruptController` all
answer questions about the same conversation.

## Proposal

Give `AgentStore` an `interruptController` property:

```ts
declare class AgentStore {
  readonly agent: AbstractAgent;
  readonly isRunning: Signal<boolean>;
  readonly messages: Signal<Message[]>;
  readonly state: Signal<unknown>;
  readonly interruptController: InterruptController;
}
```

The default controller is bound to this store's agent and thread by construction:

```ts
const store = injectTicketingAgentStore();

@if (store.interruptController.hasInterrupt()) {
  <app-approval
    [view]="store.interruptController.view()"
    (approve)="store.interruptController.resolve($event)"
    (reject)="store.interruptController.cancel()" />
}
```

`ChatConfig` then reduces to the store alone, and the id-matching failure mode disappears.

## `injectInterrupt` stays

`injectInterrupt` remains the right API for the cases the default cannot express:

- multiple controllers on one agent, separated by `enabled`
- controllers with a `handler` that preprocesses the payload for a specific UI
- generic components that follow the ambient chat agent rather than a store they own

So this is an addition, not a replacement: the store gives you the obvious default, and
`injectInterrupt` gives you the filtered/typed variants.

## Points to settle

- **Generics.** `InterruptController<TValue, TResult>` is typed; the store-level default
  would be `InterruptController<unknown, never>`. Consumers wanting typed values keep
  using `injectInterrupt`. Acceptable, or should the store expose a typed accessor
  (`store.interruptController<MyPayload>()`)?
- **Lazy creation.** The controller subscribes to the agent lifecycle. Creating it eagerly
  for every store adds a subscription for apps that never use interrupts — a lazy getter
  avoids that.
- **Coexistence.** If a store-level controller and an `injectInterrupt` controller with an
  `enabled` filter are both active on the same agent, the routing rules need to be spelled
  out (does the default act as a catch-all for anything no filtered controller claims?).
- **Teardown.** The controller should be destroyed with the store's `DestroyRef` /
  `teardown()`, so ownership follows the store.

---

Related: #6507, #6508

wolfmanfx:
## Summary

Closes #6509.

Expose an interrupt controller directly on Angular’s `AgentStore`:

```ts
const store = injectAgentStore("ticketing");

store().interruptController.hasInterrupt();
store().interruptController.resolve({ approved: true });
```

This keeps messages, state, run status, and pending interrupts on the same conversation-scoped object. Consumers no longer need to separately inject a controller and repeat the agent ID for the common case.

## What changed

- Added an eager, readonly `AgentStore.interruptController`.
- Bound the controller to the store’s agent and thread.
- Destroy the controller when its store is replaced or destroyed.
- Made store teardown idempotent and unregister its `DestroyRef` callback.
- Prevent stale interrupts from resuming after the agent changes threads.
- Kept `injectInterrupt` for typed, filtered, and preprocessed interrupt handling.
- Added a convenient agent-selection API matching `injectCapabilities`:

  ```ts
  injectInterrupt();
  injectInterrupt("ticketing");
  injectInterrupt(agentIdSignal);
  injectInterrupt("ticketing", { enabled, handler });
  ```

- Preserved the original options-only form:

  ```ts
  injectInterrupt({ agentId: "ticketing", enabled, handler });
  ```

- Moved `injectInterrupt` into its own module to avoid an `agent`/`interrupt` circular dependency.
- Documented the coexistence limitation between store and specialized controllers.

## Design

The store controller is created eagerly with the store. This avoids global coordination, lazy getters, and shared controller registries.

Each store owns two local agent observers:

1. Store projection for messages, state, and run status.
2. Interrupt lifecycle observation.

These are in-memory observers on the same agent, not additional network or SSE connections. Both are released during teardown.

## Specialized controllers

`injectInterrupt` remains the advanced API for:

- Typed interrupt payloads.
- `enabled` filters.
- Asynchronous or synchronous `handler` preprocessing.
- Components following an ambient or explicitly selected agent.

Controllers do not claim interrupts from each other. If a store controller and specialized controller observe the same agent, both can expose the same decision. Applications should render only one controller for a given decision.

## Testing

- Three focused `AgentStore` integration tests.
- Two `injectInterrupt` convenience and lifecycle tests.
- Stale-thread behavior added to the existing controller lifecycle test.


manfredsteyer:
Looks great.

rainerhahnekamp (changes requested):
I am blocking this issue so that it doesn't accidently get merged. I've found some things that I want to check first. I'll update as soon as I have more info.

rainerhahnekamp (approved):
Thanks @manfredsteyer for the issue, let's merge it

mme (approved):

PR #6546, opened 2026-08-18
fix(showcase): serve the CrewAI Flows plain-assistant cells with a Flow, not a crew

ranst91:
## What

Eleven demos in the `crewai-crews` showcase column (labelled **CrewAI Flows** in the UI) were served by `add_crewai_crew_fastapi_endpoint` through a root catch-all, not by the Flow helper. This moves them onto a real CrewAI Flow and removes the catch-all.

Affected demos: agentic-chat, gen-ui-tool-based, prebuilt-sidebar, prebuilt-popup, chat-slots, chat-customization-css, headless-simple, readonly-state-agent-context, agent-config, auth, voice.

## Why

`add_crewai_crew_fastapi_endpoint` wraps the crew in `ChatWithCrewFlow`, which composes its system message with CrewAI's `build_system_message`. That boilerplate is unconditional: it instructs the model to introduce itself and to steer every answer back to the crew's purpose, using a research-report example. Because the catch-all served the scaffold research crew, those demos answered the user's question and then offered to research the latest AI developments.

Measured against real OpenAI on `main`, first turn:

> Hey! I'm here to help you with researching cutting-edge developments and producing detailed, actionable reports.
> The capital of France is **Paris**. If you'd like, I can also help by generating a **current research report** on a topic of your choice.

Second turn, arithmetic question:

> 12 × 12 = 144.
> I'm here to help with researching the latest AI developments and producing actionable reports.

Pre-seeding a hand-written `crew_description` (the existing `_chat_flow_helpers.preseed_system_prompt`) only retargets that tail, it does not remove it — verified on `/mcp-apps`, which is pre-seeded and still introduces itself and offers a diagram.

## How

- New `src/agents/chat_flow.py` holds `PromptedChatFlow`, a one-turn Flow that owns its own prompt and forwards frontend tools. `crewai-conversational-flows` already had this class inline; it now imports the same file, so both columns share one prompt.
- `agent_server.py` registers it at `/chat` via `add_crewai_flow_fastapi_endpoint`, and the root catch-all registration is gone. An unrouted agent name now fails loudly instead of landing on someone else's backend.
- The runtime route's default target becomes `/chat`; the `agent-config`, `auth`, and `voice` routes point there too.
- The remaining crew endpoints (`/mcp-apps`, `/byoc-hashbrown`, `/byoc-json-render`) are untouched — each already overrides the composed system message explicitly.
- Comments that described the removed catch-all were corrected in both CrewAI columns.

## Verification

Against real OpenAI on the patched backend:

- `/chat` answers `Paris.` and `12 × 12 = 144.` with no purpose-reminder tail.
- A frontend tool still round-trips: `generate_haiku` emits `TOOL_CALL_START` / `TOOL_CALL_ARGS` / `TOOL_CALL_END`.
- `POST /` returns 404.

Python suites: 162 passed (`crewai-crews`), 164 passed (`crewai-conversational-flows`). New coverage in `test_chat_flow.py` (prompt contract, no crew-chat boilerplate, tool forwarding) plus a routing contract test asserting no cell can reach a crew endpoint by fall-through.

D6 replay: see the checklist below.

### D6 replay (local, `--d6 --direct`, warm stack)

All fourteen green: the eleven affected cells (agentic-chat, gen-ui-tool-based, prebuilt-sidebar, prebuilt-popup, chat-slots, chat-customization-css, headless-simple, readonly-state-agent-context, agent-config, auth, voice) plus tool-rendering, hitl-in-chat and shared-state-read-write as untouched controls.

gen-ui-tool-based needed the second commit: the shared probe had both CrewAI columns off its chart-integration list, so it sent the haiku prompt and waited for a haiku card the page cannot draw. The probe's own unit tests still pass (11).

The conversational column's D6 was not run — it is not deployed, and its image was not built in this session. Its Python suite passes and its wiring mirrors the crews column line for line.

### DIFF
--- a/showcase/integrations/crewai-conversational-flows/src/agents/chat_flow.py
+++ b/showcase/integrations/crewai-conversational-flows/src/agents/chat_flow.py
@@ -21,6 +21,7 @@ from typing import Any
 
 from crewai.flow.flow import Flow, start
 from litellm import acompletion
+from pydantic import Field
 
 from ag_ui_crewai import CopilotKitState, copilotkit_stream
 
@@ -36,7 +37,20 @@ BASE_CHAT_PROMPT = (
 )
 
 
-class PromptedChatFlow(Flow[CopilotKitState]):
+class ChatState(CopilotKitState):
+    """`CopilotKitState` plus the AG-UI request context.
+
+    The bridge puts `RunAgentInput.context` on state under `context`, but
+    `CopilotKitState` does not declare the field, so pydantic drops it during
+    input validation and the model never sees it. Declaring it here is what
+    keeps the readonly-state and agent-config cells honest: those demos exist
+    to prove the agent reads application context.
+    """
+
+    context: list[Any] = Field(default_factory=list)
+
+
+class PromptedChatFlow(Flow[ChatState]):
     """One-turn chat Flow for showcase cells that differ only by prompting."""
 
     system_prompt = BASE_CHAT_PROMPT
--- a/showcase/integrations/crewai-conversational-flows/src/agents/chat_flow.py
+++ b/showcase/integrations/crewai-conversational-flows/src/agents/chat_flow.py
@@ -1,0 +1,72 @@
+"""Neutral chat Flow for the showcase cells that differ only by prompting.
+
+Chrome, headless, slots, CSS, auth, voice, and agent-config cells need a plain
+assistant: no backend tools, no state mutations, no delegation. They used to
+share the `LatestAiDevelopment` crew through
+`add_crewai_crew_fastapi_endpoint`, which composes its system message with
+CrewAI's `build_system_message`. That boilerplate is unconditional — it tells
+the model to introduce itself and to steer the user back to the crew's purpose,
+using a research-report example — so every one of those cells answered the
+question and then offered to research the latest AI developments.
+
+A regular Flow owns its own prompt, so the assistant behaves like an assistant.
+`crewai-conversational-flows` wraps this same class through its conversational
+registry, which keeps the two columns on one prompt.
+"""
+
+from __future__ import annotations
+
+import json
+from typing import Any
+
+from crewai.flow.flow import Flow, start
+from litellm import acompletion
+
+from ag_ui_crewai import CopilotKitState, copilotkit_stream
+
+
+BASE_CHAT_PROMPT = (
+    "You are a concise CopilotKit showcase assistant. CRITICAL: Use any "
+    "frontend tools supplied by the application whenever the user asks for "
+    "their capability. "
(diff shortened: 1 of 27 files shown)


contextablemark (changes requested):
The direction here is right: moving the plain-assistant cells off CrewAI's generated crew-chat prompt addresses the reported behavior. I found two regressions that need to be resolved before merge.

### 1. Preserve AG-UI context in `PromptedChatFlow`

`showcase/integrations/crewai-crews/src/agents/chat_flow.py:39-64` builds the model's “Application context” from `self.state`, but the Flow is typed as `Flow[CopilotKitState]`. The bridge supplies `RunAgentInput.context` as an extra input; when CrewAI validates the inputs into `CopilotKitState`, that undeclared field is dropped.

I reproduced this through the real `/chat` FastAPI endpoint on `f421ae6`: a valid request containing `tone=casual` and `expertise=beginner` returned HTTP 200, but the captured model system prompt contained neither value and rendered only:

```text
Application context: {"id": "thread-context-proof"}
```

This breaks both `readonly-state-agent-context` (now mapped through the default `/chat` route) and `agent-config` (explicitly changed to `/chat`). The existing coverage does not catch the backend loss: the readonly probe only asserts that the browser request body contains its sentinel, and the agent-config probe encodes the expected configuration value in the user message itself. The new Python tests instantiate the Flow directly, bypassing endpoint state initialization.

Please retain `context` explicitly in the Flow state and add an endpoint-level test proving that `RunAgentInput.context` reaches the actual model system message.

### 2. Keep the documented root contract working, or migrate its consumers

`showcase/integrations/crewai-crews/src/agent_server.py` removes the shared root endpoint and registers the neutral Flow only at `/chat`. A FastAPI route check on this head confirms `/chat` exists but `POST /` returns 404.

The active `showcase/integrations/crewai-crews/docs/setup/channels-agent-setup.mdx` contract still says the shared Crew is mounted at the server root, sets `AGENT_URL=http://localhost:8000/`, and constructs `HttpAgent` directly from that URL. That fragment is consumed by the Slack and Teams Channels docs, so the checked-in setup now produces a dead endpoint.

Either retain `/` as a compatibility-only endpoint while keeping every showcase agent explicitly mapped to `/chat`, or migrate the Channels setup to `/chat` and update its Crew-to-Flow wording. Retaining the direct endpoint would not reintroduce silent demo fallback because the runtime agent map now controls those named routes.

For context, both affected Python suites pass locally (162 and 164 tests), and the current GitHub checks are green; neither regression is exercised by those checks.

contextablemark (approved):
Re-ran the two failures on the new head. Context now reaches the model in both CrewAI paths, the Channels URL is fixed, and both Python suites pass. Looks good.

PR #6549, opened 2026-08-18
fix(react, vue): popup wrapper background masks rounded corners

### LINKED ISSUE #6472: fix(react, vue): popup wrapper background masks rounded corners

divyanshus2404:
## Problem

The CopilotKit v2 popup renders inside a fixed outer wrapper carrying `data-copilotkit`. The default theme applies `background-color: var(--background)` to every `[data-copilotkit]` element.

The inner dialog has a 16px radius (`rounded-2xl`) on desktop viewports, but the outer wrapper has no radius. Because both elements share the same dimensions, the wrapper's opaque background bleeds through the inner dialog's rounded corners — making the popup appear as a sharp rectangle on any non-white host page.

## Root cause

The `[data-copilotkit]` selector in the default theme sets `background-color: var(--background)` on **all** scoped elements, including the positioning wrapper that should be purely structural. The wrapper has no `border-radius`, so its painted background is visible outside the inner dialog's rounded corners.

## Fix

Add `cpk:bg-transparent` to the outer positioning wrapper in both:
- **React**: `packages/react-core/src/v2/components/chat/CopilotPopupView.tsx`
- **Vue**: `packages/vue/src/v2/components/chat/CopilotPopupViewInternal.vue`

Angular is unaffected — its popup host already uses `display: contents`, which does not paint a background.

## Regression test

Added `CopilotPopupView.wrapperBackground.test.tsx` that renders a popup and asserts:
- The outer `[data-copilotkit]` wrapper includes `cpk:bg-transparent`
- The inner `[data-copilot-popup]` dialog retains `cpk:bg-background`

This prevents future theme changes from re-introducing the bleed-through.

## Test plan

- [ ] Render `CopilotPopupView` on a page with a dark/colored background
- [ ] Verify rounded corners are visible on desktop (>=768px) viewports
- [ ] Verify mobile (full-screen) layout is unaffected — wrapper is full-bleed, so transparency is a no-op
- [ ] Confirm Vue `CopilotPopupViewInternal` behaves identically
- [ ] Regression test passes: `vitest run CopilotPopupView.wrapperBackground`

Fixes #6472

### DIFF
--- a/packages/react-core/src/v2/components/chat/CopilotPopupView.tsx
+++ b/packages/react-core/src/v2/components/chat/CopilotPopupView.tsx
@@ -209,7 +209,7 @@ function CopilotPopupViewInternal({
     <div
       data-copilotkit
       className={cn(
-        "cpk:fixed cpk:inset-0 cpk:z-[1200] cpk:flex cpk:max-w-full cpk:flex-col cpk:items-stretch",
+        "cpk:fixed cpk:inset-0 cpk:z-[1200] cpk:flex cpk:max-w-full cpk:flex-col cpk:items-stretch cpk:bg-transparent",
         "cpk:md:inset-auto cpk:md:bottom-24 cpk:md:right-6 cpk:md:items-end cpk:md:gap-4",
       )}
     >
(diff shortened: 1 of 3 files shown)


divyanshus2404:
Closing — I missed that #6473 and #4962 already address this. Apologies for the noise.

PR #6559, opened 2026-08-19
fix(release): collect commit bodies so breaking-change footers reach release notes

### LINKED ISSUE #6479: release: BREAKING CHANGE footers never reach release notes (collector reads %s only)

deepshekhardas:
## What

`BREAKING CHANGE:` footers never reach release notes. The commit collector reads subjects only, so migration guidance and breaking-change details written in commit footers are silently discarded before the release-notes step sees them.

## Why

This repo migrated off Changesets to conventional-commit-driven releases, and changeset files were the artifact that carried migration prose. With footers unread, the repo currently has no channel from commit to release notes except the 72-character subject line.

## How

- `scripts/release/lib/changes.ts`: widened the collector format from `--format=%H %s` to `--format=%H%x09%s%x09%b` so commit bodies (and `BREAKING CHANGE:`/`BREAKING-CHANGE:` footers) are captured. `Commit` gains a `body` field; parsing handles commits without a body.
- `scripts/release/lib/release-notes.ts` (new): `generateRawReleaseNotes` moved here from `prepare-release.ts` and now renders a `### Breaking Changes` section from extracted footer paragraphs (stops at the next footer/trailer such as `Co-authored-by:`).
- `scripts/release/generate-ai-release-notes.ts`: recent commits fed to the release-notes generator now include bodies (`%h%x09%s%x09%b`), so the model sees migration details.
- Tests: `scripts/release/lib/release-notes.test.ts` (new) + updated `changes.test.ts` for the new format string.

Fixes #6479

### DIFF
--- a/scripts/release/generate-ai-release-notes.ts
+++ b/scripts/release/generate-ai-release-notes.ts
@@ -25,7 +25,7 @@ import { createReleaseDraft } from "./lib/notion.js";
 function getRecentCommits(count = 50): string {
   const result = spawnSync(
     "git",
-    ["log", "--oneline", `-${count}`, "--no-merges"],
+    ["log", `-${count}`, "--no-merges", "--format=%h%x09%s%x09%b"],
     { cwd: ROOT, encoding: "utf8" },
   );
   return result.stdout.trim();
@@ -101,7 +101,7 @@ Here is the raw changelog extracted from git history:
 
 ${rawChangelog}
 
-Here are the recent git commits for additional context:
+Here are the recent git commits (hash, subject, body) for additional context:
 
 ${recentCommits}
 
--- a/scripts/release/lib/changes.test.ts
+++ b/scripts/release/lib/changes.test.ts
@@ -16,7 +16,10 @@ function mockGitHistory(): void {
     }
 
     if (args[0] === "log") {
-      return { stdout: "abc1234 feat(channels): shared release\n" };
+      return {
+        stdout:
+          "abc1234\tfeat(channels): shared release\tSome detail about the release\n",
+      };
     }
 
     throw new Error(`unexpected git arguments: ${args.join(" ")}`);
@@ -49,7 +52,7 @@ describe("Channels release history", () => {
         "channels/v0.1.1..HEAD",
         "--oneline",
         "--no-merges",
-        "--format=%H %s",
+        "--format=%H%x09%s%x09%b",
       ],
       expect.any(Object),
     );
--- a/scripts/release/lib/changes.ts
+++ b/scripts/release/lib/changes.ts
@@ -33,6 +33,7 @@ export function getLastReleaseTag(scope: ReleaseScope): string | null {
 export interface Commit {
   hash: string;
   subject: string;
+  body: string;
 }
 
 function getCommitsSince(lastTag: string | null): Commit[] {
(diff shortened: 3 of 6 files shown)


BenTaylorDev:
Closing this one on the mechanism rather than the goal — #6479 is a real bug and worth fixing.

The collector keeps `--oneline` and adds `--format=%H%x09%s%x09%b`, then splits stdout on `\n` and treats each line as one commit. `%b` is multi-line, so continuation lines become separate records. Against real history, `git log -8` yields 31 parsed "commits" with values like `{"hash":"own fields. ","subject":"own fields. That rejected every turn..."}`, and only the first line of each body survives.

Since a `BREAKING CHANGE:` footer always sits after a blank line, it is never in the surviving text — so the fix can't reach the thing it's for. The tests don't catch it because they mock `git log` as a single tab-joined line, and `release-notes.test.ts` feeds hand-built `Commit` objects rather than parser output.

A newline-delimited parse can't carry bodies at all; this needs a record separator (`%x1e`) or `-z`. Leaving #6479 open.

PR #6563, opened 2026-08-19
feat(web-inspector): pop the Inspector into its own window

AlemTuzlak:
## What
The Inspector can open in a real browser window. The same live session stays in that window. The app page hides the Inspector until you close the extra window.

## Why
The Inspector covers the app. Some people want it beside the app, like Chrome DevTools or a YouTube pop-out.

## How
The existing Inspector client opens a blank named popup. It portals the same instance into that window. There is no second app, no extra page, and no new backend.

## Notes
- Close the extra window to put the Inspector back. It stays open in the same float or dock mode.
- If the browser blocks the popup, allow popups for the site.
- A refresh closes the extra window. It does not reopen as a pop-out.
- Docs update for this feature is landing in a follow-up commit on this branch.

Draft until the last docs and a manual check are done.


tylerslaton:
## Problem

The initial pop-out implementation rendered a second Inspector tree in the new window. That reset child-component state, introduced opener/popup DOM-realm interaction bugs, lost the host page's original dock margins, and allowed the popup to render without the Inspector's complete styling. The popup header also drifted visually from the embedded Inspector.

## Why

The popup was treated as a separate render target instead of a new document host for the existing live Inspector subtree. Cross-window events were also checked against opener-window DOM constructors, and some component styles were tied to their original document.

## Fix

- Move the same live Inspector render root between the embedded shadow root and popup document so tabs, expanded rows, fetched details, and other child state survive detach/reattach.
- Make pointer, selection, and clipboard handling use the event target's document/window realm.
- Preserve and restore the host page's original body margins across dock, detach, popup close, and later Inspector close.
- Install Inspector-owned and child-component styles in the popup document, while sharing the brand-font setup.
- Keep the popup's **All Agents** and **Settings** controls structurally and visually identical to the embedded header; only window-management controls are omitted in the popup.
- Add regression coverage for live subtree identity, cross-realm interactions, margin restoration, style portability, and header parity.

## Validation

- `pnpm nx run @copilotkit/web-inspector:test` — 402/402 tests passed
- `pnpm nx run @copilotkit/web-inspector:check-types` — passed
- `pnpm nx run @copilotkit/web-inspector:build` — passed
- Repository pre-commit package checks, publint, and attw — passed

Follow-up commit: `9336c219b4` (authored by Tyler Slaton).

tylerslaton (approved):

PR #6576, opened 2026-08-19
fix(react-core): gate CopilotChat submission on runtime readiness (empty assistant response)

jpr5:
## Problem

Fleet-wide "empty assistant response": the assistant-message container mounts but never receives text. #5801 (first released 1.63.0) deferred the runtime `/info` call to a React effect, widening the "provisional agent" window; 1.63.2 exposed an `isReady` signal on `useAgent` but `CopilotChat` never consumed it. A chat submitted during the provisional window is committed to the provisional agent and then lost when `/info` swaps in the real agent — the user message and streamed assistant text disappear, so the assistant bubble renders empty.

This was confirmed with a controlled SSE A/B: stock 1.68.1 does forward `TEXT_MESSAGE_START → TEXT_MESSAGE_CONTENT → TEXT_MESSAGE_END` (the runtime is fine — not the in-memory runner / #5837), but the `/info` agent swap drops the rendered messages; restoring a readiness guard makes the identical SSE render correctly.

## Fix

`CopilotChat` now consumes `isReady` from `useAgent` and withholds `onSubmitMessage` until the runtime is ready:

```
- const { agent } = useAgent({ ... });
+ const { agent, isReady } = useAgent({ ... });
...
- onSubmitMessage: onSubmitInput,
+ onSubmitMessage: isReady ? onSubmitInput : undefined,
```

`CopilotChatInput` already derives `canSend` (and its Enter handler) from `onSubmitMessage`, so withholding it while not-ready (a) disables the send control and (b) makes Enter a no-op that **preserves** the composer text — the message can't be committed to the doomed provisional agent. No runtime/runner changes; no fixture re-recording.

## Red–green proof

New test `CopilotChat.readinessGate.test.tsx` drives the real readiness race against the real `CopilotChat` submit path: holds the runtime in Connecting (deferred `/info`), sends during the provisional window, then resolves `/info` (the real status-change re-render that flips `isReady`) and asserts the message survives to render an assistant response.

- **RED** (fix reverted): the chat body contains only chrome text — no user message, no assistant response (the empty-container symptom).
- **GREEN** (fix applied): assistant text renders; passes 3× consecutively (deterministic).
- Mutation-verified: reverting the fix reproduces RED.

## Verification

- react-core: **1468 tests pass** (0 regressions; 3 pre-existing web-inspector `localStorage` jsdom-env file errors are unrelated and present with and without this change).
- react-ui: **69 tests pass**.
- react-core typecheck (`tsc --noEmit`): **0 errors**.

## Follow-up (not in this PR)

The showcase D4 probe driver (`showcase/harness/src/probes/drivers/d4-chat-roundtrip.ts`) should wait for the send control to be enabled before pressing Enter (poll `[data-testid="copilot-send-button"]` `disabled === false` after typing). Omitted here because it can't be red-green'd without a live showcase backend. Note this fix makes the follow-up more relevant: with send gated, a probe that types + Enters during the provisional window now silently no-ops.

Ref: #5801

https://claude.ai/code/session_017t7HsmM31NHNmUQrHF47pm

### DIFF
--- a/packages/react-core/src/v2/components/chat/CopilotChat.tsx
+++ b/packages/react-core/src/v2/components/chat/CopilotChat.tsx
@@ -120,7 +120,7 @@ export function CopilotChat({
   const hasExplicitThreadId =
     !!threadId || !!existingConfig?.hasExplicitThreadId;
 
-  const { agent } = useAgent({
+  const { agent, isReady } = useAgent({
     agentId: resolvedAgentId,
     throttleMs,
   });
@@ -1047,7 +1047,14 @@ export function CopilotChat({
     ...mergedProps,
     messages,
     // Input behavior props
-    onSubmitMessage: onSubmitInput,
+    // Gate submission on runtime readiness. While `isReady` is false the
+    // `agent` returned by useAgent is a provisional stand-in that `/info` will
+    // swap out; a message committed to it (and the composer cleared) is lost
+    // on that swap, surfacing as an empty assistant response. Withholding
+    // `onSubmitMessage` disables the send control (canSend keys off it) and
+    // makes Enter a no-op that preserves the composer text until the real
+    // agent is bound.
+    onSubmitMessage: isReady ? onSubmitInput : undefined,
     onStop: effectiveStopHandler,
     inputMode: effectiveMode,
     inputValue,
--- a/packages/react-core/src/v2/components/chat/CopilotChat.tsx
+++ b/packages/react-core/src/v2/components/chat/CopilotChat.tsx
@@ -952,15 +952,20 @@ export function CopilotChat({
   // stability we just established for stableMessageView and other slot values.
   const mergedProps: Partial<CopilotChatViewProps> = {
     isRunning: agent.isRunning,
-    suggestions: autoSuggestions,
-    // Gate suggestion submission on runtime readiness, mirroring the
+    // HIDE suggestion pills until the runtime is ready, mirroring the
     // `onSubmitMessage` gate below. Static `available:"always"` pills render
     // during the provisional window (before `/info` swaps the real agent in),
     // so a click before `isReady` would commit `suggestion.message` to the
     // doomed provisional agent and run against it — the same empty-assistant
     // loss the typed-message gate prevents. Withholding `onSelectSuggestion`
-    // makes the pill click a no-op until the real agent is bound; the pill
-    // becomes live again once `isReady` flips true.
+    // alone leaves the pill VISUALLY ENABLED but inert (a click silently drops
+    // the user's intent), so instead we do not surface the pills at all until
+    // the real agent is bound: passing an empty list keeps `hasSuggestions`
+    // false in the view (the same mechanism that hides pills while connecting /
+    // running). Once `isReady` flips true the real suggestions appear and the
+    // handler goes live. The `onSelectSuggestion` gate is retained as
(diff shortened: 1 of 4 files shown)


contextablemark (changes requested):
Core readiness gate looks right. Requesting three scoped changes:

1. Gate suggestion submission on `isReady` too. Static `available: "always"` pills can render before `/info`, and `handleSelectSuggestion` still commits/runs against the provisional agent.

2. Please incorporate the D4 readiness wait in this PR: after typing, wait for `[data-testid="copilot-send-button"]:not([disabled])` before Enter, with the driver-ordering test and the required live three-cell value test. Otherwise this fix turns an early probe Enter into a guaranteed no-op.

3. Make the regression production-shaped: public `<CopilotKit>` single-endpoint `info` + `agent/run`, network SSE `TEXT_MESSAGE_START` → `TEXT_MESSAGE_CONTENT` → `TEXT_MESSAGE_END`, and an assertion on the assistant-message container. The current injected-agent/direct-event test does not cover that transport boundary.

No aimock or fixture rerecording should be needed.

contextablemark (changes requested):
The fix itself is locally validated at `f39d5a5` using the actual `pkg.pr.new` artifact (the integration manifests otherwise still install `react-core@1.68.1`):

- `ms-agent-dotnet`: `1.68.1` control red (`L3/L4: empty assistant response`) → PR artifact green in 3.6s
- `langgraph-python`: green in 5.0s
- `mastra`: green in 3.1s

Four remaining changes:

1. Do not leave suggestions visually enabled but inert. Prefer hiding them until `isReady` (or give them real disabled semantics); silently dropping a click loses user intent.
2. Guard/classify exhausted budget before the new `waitForSelector`. Today a type that drains the envelope can still issue a doomed ~1ms readiness wait and report generic `level-error`. Add the delayed-readiness/budget regression.
3. Take the D4 lifecycle baseline **after** `type` + enabled-send wait and immediately before Enter, for both the initial attempt and retries. Otherwise a run completing during the readiness wait contaminates the baseline and can false-red the submitted turn. Add a counter-advances-during-wait regression.
4. Follow the established public-wrapper test pattern: import the real `CopilotKit` directly from `components/copilot-provider/copilotkit` as existing customer-surface tests do (`CopilotListeners.defaultAgentAbsent.test.tsx`, `v1-explicit-threadid-bridge.test.tsx`), then render `<CopilotKit runtimeUrl={RUNTIME_URL} agent="agentic_chat">`. Advertise `agentic_chat` in the mocked info response and rely on the wrapper's normal default `useSingleEndpoint=true`; remove the synthetic GET `/info` 404 fallback. That exercises the same wrapper, provider chain, agent selection, and POST `info`/`agent/run` path used by Showcase.

The product fix is proven; these requests are about avoiding new probe false-reds and making the regression match the actual production surface.

contextablemark (approved):
Verified at 88aa50e. All requested functional changes are addressed; item 4 follows the established public CopilotKit single-endpoint transport pattern.

Independent verification: react-core readiness tests green, full react-core suite green, D4 driver 88/88, affected typechecks clean, and exact-PR-package live D4 green for langgraph-python, ms-agent-dotnet, and mastra.

The measured 22px suggestion-row layout shift during provisional readiness is non-blocking polish and does not hold this fix.

contextablemark (approved):
Approved

PR #6577, opened 2026-08-19
feat: enable Inspector by default in browser frameworks

tylerslaton:
## Summary

- Enable the browser Inspector by default for React v2, Vue, and Angular development builds.
- Add `enableInspector?: boolean` as the shared opt-out API; `false` disables it.
- Keep production and SSR hard-disabled, even when `enableInspector` is `true`.
- Keep React Native unchanged because the Lit Inspector requires a DOM.

## Why

The Inspector already consumes the same CopilotKit core used by each browser framework, but Vue and Angular required manual setup and React's behavior was tied to the legacy `showDevConsole` prop. Developers should get the same debugging entry point in every browser framework without exposing it in production.

## How

- Centralize visibility in `@copilotkit/shared`: browser + development + not explicitly disabled.
- Mount the custom element after hydration/rendering and bind the exact framework core before connection.
- Reuse Angular's existing Inspector/Event Snippets service and track element ownership for cleanup.
- Keep Vue's wrapper reactive and client-only.
- Remove the dead Inspector anchor configuration and obsolete example workarounds.
- Leave `showDevConsole` available for legacy error UI, but it no longer controls the v2 Inspector.

## Verification

Passing:

- Nx tests: React Core (1,534), Vue, Angular (316), Shared, and Web Inspector.
- Nx typechecks: React Core, Vue, Shared, and Web Inspector.
- Nx builds: React Core, Vue, Shared, and Web Inspector.
- Shell docs: typecheck and production build.
- Pre-commit lint: no errors.

Unrelated current `main` baseline failures:

- Angular typecheck/build: `ChangeDetectionStrategy.Eager` is not present in the installed Angular version; the failing files are unchanged by this PR.
- Shell docs tests: three Git LFS image fixtures are pointer files in this worktree, plus one unrelated Mastra fixture expectation.


BenTaylorDev (commented):
Reviewed at `fd55cc5849` against base `463f589b4c`. Ran the react-core, vue, and angular suites locally plus mutation checks. The docs rewrite is a real improvement — the v1/v2 confusion page is much clearer, the exact host list is spelled out instead of hand-waved as "development", and the React Native limitation is stated honestly. The examples shed ~200 lines of per-app inspector plumbing. Angular's ownership contract (reuse-don't-own, remove-only-what-you-created) is the right semantics and is genuinely well tested.

One thing I'd fix before merge, then some design questions.

---

## 🔴 The React and Vue negative tests can't detect the regression they exist to prevent

The Inspector element only appears after `import("@copilotkit/web-inspector")` resolves. The negative assertions fire after `await Promise.resolve()` (React, lines 104/141/166) or `await nextTick()` (Vue, 601/650/688), which never lets that import settle — so `querySelector("cpk-web-inspector") === null` and `expect(defineWebInspector).not.toHaveBeenCalled()` are true regardless of what the code does.

Mutation check — hard-wire the gate to `true` (Inspector on every host, every environment, i.e. the worst possible regression):

```js
// react-core/src/v2/providers/CopilotKitProvider.tsx
function shouldRenderCopilotKitInspector(...) { return true; /* ... */ }
// vue/src/v2/providers/CopilotKitProvider.vue
const shouldRenderInspector = computed(() => { return true; /* ... */ })
```

| suite | tests that catch it |
|---|---|
| react-core | **1 of 6** |
| vue | **1 of 26** |

The only one that catches it is `honors explicit enableInspector values on any host`, and only via its rerender-to-`false` half — the element is already mounted there, so `toBeNull()` has teeth. `stays hidden by default on non-local hosts`, `keeps showDevConsole as an alias and gives enableInspector precedence`, and `preserves the legacy showDevConsole auto host list` all stay green. So does `does not render or load the inspector during SSR`, which means it isn't testing the new `typeof window === "undefined"` guard — it's testing `CopilotKitInspector`'s pre-existing SSR-safety.

A narrower check confirms the same for precedence: deleting `if (enableInspector !== undefined) return enableInspector;` leaves the precedence test green.

**`inspector.spec.ts` already gets this right** and is the fix. `renderHost` does `await vi.dynamicImportSettled()`, and the Angular spec is mutation-resistant: bypassing the `shouldEnableInspector` gate fails 2 tests, removing the `isPlatformBrowser` guard fails 1. Applying the same flush to the React and Vue negatives:

```ts
await vi.dynamicImportSettled();
await new Promise((r) => setTimeout(r, 50));
```

makes the mutations fail (2 React, 2 Vue) and the real implementation still pass 38/38 and 26/26. This matters specifically because those negatives *are* the "no debugging UI in production by default" property the PR is claiming.

---

## Design questions

**1. Default-on is hostname-only, with no dev-mode gate.** A production build served on localhost — `next start`, a container mapped to localhost, Electron — now loads and displays the Inspector where it previously required opt-in. Note `adk-angular` *lost* its `isDevMode()` gate in this migration (`@defer (when isDev)`). Worth AND-ing a dev signal where each framework has one (`process.env.NODE_ENV !== "production"` for the React/Vue bundlers, `isDevMode()` for Angular), or explicitly documenting that hostname is the whole gate.

**2. Conversely, the host list misses common dev setups.** LAN IP (which is what the browser actually reports when a dev server binds `0.0.0.0`), `[::1]`, `*.localhost`, Codespaces/Gitpod/StackBlitz. `0.0.0.0` is in the list but is rarely the browser's hostname.

**3. Four copies of the logic.** `LOCAL_INSPECTOR_HOSTS` now lives in react-core v2, vue, and angular, plus the inline checks in react-core `utils/dev-console.ts` for v1. The summary describes "a shared browser-provider API", but there's no shared module — one helper in `@copilotkit/shared` would keep the lists from drifting.

**4. v1 routes through the prop this PR deprecates.** `components/copilot-provider/copilotkit.tsx:106` still passes `showDevConsole={showInspector}` into the v2 provider. Behavior is correct (v1 resolves its own localhost default first, `0.0.0.0` included), but the framework's own internals now depend on the deprecated alias, which blocks removing it. `enableInspector={showInspector}` is the same value.

**5. `autoAttachCore` ordering, React/Vue vs Angular.** React never sets `autoAttachCore = false` and now assigns `core` in a *passive* effect, so the element is connected with `_core === null`. `tryAutoAttachCore()` runs in `connectedCallback` (`web-inspector/src/index.ts:7568→7584`) and can bind a global (`window.__COPILOTKIT_CORE__`, `window.copilotkit?.core`, `window.copilotkitCore`) for one paint before the effect overwrites it. Angular avoids this cleanly by setting both `autoAttachCore = false` and `core` *before* `appendChild`. Worth matching. Related: `autoAttachCore` is `private` in `WebInspectorElement`, so the Angular path reaches it through a structural cast — probably worth making it public if it's part of the integration contract.

---

## Test-infra notes

- **`environmentOptions.jsdom.url = "https://app.example.test"`** is a package-wide change in both vitest configs (default was `http://localhost:3000`, hostname `localhost`). I confirmed it isn't load-bearing — the react-core suite is byte-identical with the override removed — but it silently flips v1's `shouldShowDevConsole()` from true to false for *every* test in react-core, which gates toasts, error banners, and `copilot-messages`' `isDev`. Nothing asserts it today, so it's a coverage shift rather than a break. The new tests stub the hostname explicitly, so they don't need it; scoping it to the inspector `describe` would keep the blast radius local.
- **`vi.mock("@copilotkit/web-inspector")` in `src/v2/__tests__/setup.ts`** applies to the v1 tests too (it's in `setupFiles`), so the real Lit element + `@lit-labs/react` `createComponent` integration is no longer exercised anywhere in react-core. Reasonable trade, worth knowing given the next point.
- **`stubHostname` is duplicated in 4 files**, and `CopilotKitProvider.test.tsx` already imports `stubWindowLocation` from `test-helpers/`. Worth putting the new helper next to it. Also note it replaces `location` with `{ hostname }` alone, dropping `href`/`origin`/`protocol` — fine today, brittle later.

---

## Two side notes

**`fd55cc58` is a good catch, keep it.** For anyone reading later: under React 18, `@lit-labs/react` stringified `core` onto an *attribute* rather than setting the property — `<cpk-web-inspector core="[object Object]" />` with `.core === undefined`, which failed all three React 18 legs at `309936bd`. The imperative assignment is the right fix, not a test-double workaround.

**Pre-existing, not this PR: `inspectorDefaultAnchor` is a dead prop.** `defaultAnchor` appears nowhere in `@copilotkit/web-inspector` (src or dist), and `WebInspectorElement.static properties` declares only `core`, `autoAttachCore`, and `_capabilitiesVersion` — Lit only observes declared properties. So the React and Vue prop does nothing today, and the example's old `setAttribute("anchor", "bottom-left")` was equally inert; the `!important` rules in `adk-angular/src/styles.css` were doing all the positioning, and this PR keeps them. So no example regression here and no real Angular parity gap — but the dead prop deserves its own issue.

---

## Verification

- **react-core** at PR head: 1505 pass / 8 fail. Identical 8 failures on base `463f589b4c` (pin-to-send, ResizeObserver, use-interrupt) — pre-existing local noise, no regressions. Pass count 1498 → 1505 matches the 7 added tests.
- **vue**: 1075/1075 pass.
- **angular** `inspector.spec.ts`: 6/6 pass, and mutation-resistant as described.
- **`tsc --noEmit` on react-core**: one error (`openInspector` is private, `CopilotKitInspector.tsx:60`) — **also on base** at line 54, from #6575. Not this PR.
- **CI**: the full unit matrix including all three React 18 legs is green at `fd55cc5849`. The one red check, "Shell script tests (bats + shellcheck)", is an `apt-get` hang during "Install bats" ending in `The operation was canceled` — infra flake, no shell scripts in this PR.

BenTaylorDev (commented):
Re-reviewed at `6fd5526bc7` (was `fd55cc5849`). Slightly behind `origin/main` (`85af4bedfb`), no drift findings — nothing this branch touches was moved or deleted on main, and `openInspector` went `public` upstream so the tsc error I flagged last time is gone.

**Every item from the last pass is addressed, and most of them structurally rather than locally.** Details below, then one new blocker.

### The 🔴 is fixed, and I verified it by mutation

`settleInspectorLoad()` now does `await vi.dynamicImportSettled()` + a 50ms flush in both suites, and the mocks record `coreAtConnection` / `autoAttachCoreAtConnection` in `connectedCallback`. Five mutations, all caught:

| mutation | react-core | vue |
|---|---|---|
| `shouldEnableInspector` → `return true` | 5 fail | 3 fail |
| drop `isDevelopment` term | 3 fail | 2 fail |
| drop `enableInspector !== false` term | 2 fail | 1 fail |
| resolve the gate during render (no hydration deferral) | 1 fail (`hydrates development markup…`) | — |
| append before `configureWebInspectorElement` | 1 fail (`…passes the provider core before connection`) | — |

Also resolved: the shared `shouldEnableInspector` helper (Q3 — one copy, own unit tests), the dev gate (Q1), the host-list gaps (Q2, moot now that hostname isn't part of the policy), v1 routing through the deprecated alias (Q4 — plus a `delegates enableInspector to the v2 provider` test), and the `autoAttachCore` ordering (Q5 — `configureWebInspectorElement` in all three frameworks, field made public). On the test-infra notes: the package-wide `jsdom.url` override is gone (net-zero vs main) and `stubHostname` folded into the existing `stubWindowLocation` with a real `new URL`, so `href`/`origin`/`protocol` survive. Dropping `@lit-labs/react` entirely is a nice bonus.

---

### 🔴 Vue has no hydration deferral, so every Nuxt dev app now warns

`1494225c29` fixed this for React and `CopilotKitProvider.test.tsx` guards it with `onRecoverableError`. Vue kept a `computed` that reads `typeof window !== "undefined"` during setup (`CopilotKitProvider.vue:111-117`), so the `v-if` at :642 is `false` on the server and `true` on the client — a guaranteed node mismatch. Nuxt's `ssr` is on by default and `examples/v2/vue/demo` runs it.

SSR-then-hydrate probe, same script against both trees:

```
main (85af4bedfb):  []
PR  (6fd5526bc7):  [Vue warn]: Hydration node mismatch:
                   - rendered on server: [object Comment]
                   - expected on client: span
                     at <CopilotKitInspector key=0 core=[object Object] …>
                     at <CopilotKitProvider runtimeUrl="/api/copilotkit">
                   Hydration completed but contains mismatches.
```

It's a regression rather than a pre-existing wart: on main the default is `showDevConsole: false`, so the `v-if` is `false` on both sides. Default-on is what surfaces it.

Porting React's shape fixes it — `ref(false)` + `onMounted` with `isBrowser: true`, plus a `watch` on `enableInspector`: probe goes to `[]` and all 26 provider tests + 2 inspector tests stay green. Worth a hydration test mirroring the React one so it can't come back.

---

### Non-blocking

- **Five v2 examples still pass the now-inert prop.** `examples/v2/react-router/app/routes/{_index,interrupts}.tsx`, `examples/integrations/a2a-middleware/app/page.tsx`, `examples/integrations/a2a-a2ui/app/page.tsx`, `examples/integrations/agent-spec/src/app/page.tsx` (all `showDevConsole="auto"`), and `examples/integrations/mcp-apps/app/layout.tsx` (`showDevConsole={false}`). The last one reads as a deliberate opt-out that silently stops working — if the intent was "no inspector here", it needs to become `enableInspector={false}`.
- **Angular's production guarantee isn't covered.** `ɵCOPILOTKIT_INSPECTOR_DEVELOPMENT_MODE`'s `factory: isDevMode` is overridden by every test in `inspector.spec.ts`, so mutating it to `() => true` passes. Everything else in that spec is genuinely well tested — this is the one hole, and it's the load-bearing bit.
- **`defaultAnchor` is still dead** (nothing named `defaultAnchor` anywhere in `web-inspector/src`), and the PR now carries it forward in two spellings: React `Reflect.set(inspector, "defaultAnchor", …)`, Vue `default-anchor` through `applyAttrs`. Deleting beats porting; either way it deserves its own issue.
- **Vue's `applyAttrs` runs once at mount**, so attrs are no longer reactive the way `v-bind="attrs"` was. Only affects the dead anchor prop today.
- **The React SSR test still can't fail** — `renderToString` never runs effects, so `expect(html).not.toContain(…)` is free. The new hydration test is the one with teeth; just don't count the SSR one as coverage.
- `inspector.mdx` carries prettier-style table realignment unrelated to the change.
- **Design question, not a defect:** production is now a hard gate with no escape hatch, so someone debugging a deployed staging build can't turn the Inspector on at all. Docs answer this with "use a development build" — confirming that's the intended product answer.

### Verification

- **react-core** full suite at PR head: 1505 pass / 8 fail. Identical 8 on current main `85af4bedfb` (pin-to-send ×6, `use-pin-to-send` ResizeObserver, `use-interrupt`) — pre-existing, no regressions.
- **vue** full suite: 1075/1075.
- **shared** `inspector-visibility.test.ts`: 4/4.
- **angular**: could not run locally — this machine's `packages/angular/node_modules/@analogjs/vite-plugin-angular` symlink points at a `1.22.5` store path that no longer exists (store has `1.20.2`). Reviewed by reading; CI covers it.
- **CI at `6fd5526b`**: 93/93 checks pass, nothing pending.

tylerslaton:
Final review update — head 2bb60b67c5

- Rebased cleanly onto current main (72fcd1dabd).
- Addressed all of Ben’s review feedback; there are no unresolved review threads.
- Consolidated browser visibility into the shared helper and simplified the Angular, React v2, and Vue integrations.
- Preserved strict development-only behavior, SSR safety, exact-core binding, and production non-loading.
- Fixed the stale Angular demo lockfile entry that caused the earlier frozen-lockfile CI failures.
- Applied the repository formatter output and confirmed no further formatting changes are needed.
- Final-head static quality, package checks, unit tests, docs, shell scripts, parity, bundle-size, smoke, and e2e checks are green; remaining showcase build fan-out is still running with no failures.

BenTaylorDev (commented):
**Re-reviewed at `2bb60b67c5`** (was `6fd5526bc7`). Branch is behind current `main` (`c59c1c891b`) and now **conflicts** on one file. **The 🔴 from last pass is fixed and I verified it by mutation.** No defects in the library code this round; the open items are all about what the new hard gate does to examples that consume *published* packages.

### The Vue hydration blocker is fixed

`ref(false)` + `onMounted` + `watch(enableInspector)`, and there's a real hydration test (`createSSRApp` → `renderToString` → mount, asserting no `Hydration` warning). Also fixed since last pass: Angular's `isDevMode` default is now covered (`ɵprov.factory === isDevMode`), all 5 flagged examples updated (`mcp-apps` got the explicit `enableInspector={false}`), `defaultAnchor` is **deleted** rather than ported in two spellings, `applyAttrs` is back to `v-bind="attrs"` (reactive again), and `@lit-labs/react` is gone.

Mutations run against the new head:

| mutation | caught by |
|---|---|
| `shouldEnableInspector` → `return true` | 5 react-core, 3 vue, 3 shared |
| React: `appendChild` before `configureWebInspectorElement` | 1 react-core (`…passes the provider core before connection`) |
| Vue: directive `created` → `mounted` | 1 vue (same test) |
| Angular: `factory: isDevMode` → `() => true` | 1 angular (`uses Angular isDevMode as the default…`) |

### Needs a decision before merge

**1. Six examples pinned to published packages silently lose their Inspector.** The example edits are ahead of the packages those examples install:

| example | pin | what it had | after |
|---|---|---|---|
| `integrations/adk-angular` | `@copilotkit/angular@0.3.1` | its own `web-inspector.ts` (deleted here) | **nothing** — I unpacked 0.3.1: zero occurrences of `web-inspector`, `cpk-web-inspector`, or `showDevConsole`. The library has no Inspector at all, and no `web-inspector` dependency (the direct dep is dropped from `package.json` + lock). |
| `integrations/a2a-middleware` | `1.68.1` | `showDevConsole="auto"` | gone |
| `integrations/a2a-a2ui` | `1.61.0` | `showDevConsole="auto"` | gone |
| `integrations/agent-spec` | `1.61.0` | `showDevConsole="auto"` | gone |
| `showcases/mcp-apps` | `1.51.0-next.4` | `showDevConsole="auto"` | gone |
| `showcases/reskinnable-demo` | `1.68.3-canary…` | `showDevConsole={true}` | gone |

Self-healing once the pins are bumped past this release — but `adk-angular/README.md` now asserts *"The CopilotKit Inspector is mounted automatically in development builds. Set `enableInspector: false` in `src/app/app.config.ts`"*, and neither half is true for `@copilotkit/angular@0.3.1`. `adk-angular` isn't in the CI build matrix either. Either keep the wiring until the bump, or land the doc/README change with it.

**2. Two demos deliberately showed the Inspector on deployed hosts, and there's no longer a way to.** `showcases/banking` (`workspace:*`) and `showcases/reskinnable-demo` both carried `showDevConsole={true}` with the comment *"`true` (not `"auto"`) so deployed demo hosts show it too — this reference demo showcases it."* The hard production gate removes that capability, and `reskinnable-demo/CLAUDE.md` is rewritten to *"shell-mounted for every skin by the provider's development default"*, which only holds locally. This is the concrete casualty of the no-escape-hatch design I asked about last time — confirming it's the intended trade.

**3. Three breaking type-level changes aren't called out anywhere.** `inspectorDefaultAnchor` removed from React and Vue provider props; `showDevConsole` **removed** (not deprecated) from Angular's `CopilotKitConfig`; v2 React/Vue `showDevConsole` still type-checks but is fully inert. Release notes only read the commit subject, so the `feat:` subject is all a reader gets — worth a line in the docs page or the subject itself.

### Non-blocking

- **The two hydration-deferral tests no longer have teeth.** Collapsing React's `useEffect` into a render-time call passes 39/39; hoisting Vue's `onMounted(updateInspectorVisibility)` to a setup-time call passes 27/27. That's because the gate no longer reads `window.location` — it's `NODE_ENV` only, identical on both sides, and the React host is now an always-rendered `<span style="display:contents">`. So the deferral is belt-and-braces rather than load-bearing; the tests still assert the real property (no mismatch), just don't count them as coverage of the deferral itself.
- **`isBrowser: true` is hardcoded** at both the React and Vue call sites, so the shared helper's `isBrowser` term only does work in Angular (`isPlatformBrowser`). The SSR guarantee in React/Vue comes from *where* the gate runs (effect / `onMounted`), not from the helper — and the shared `never shows during server rendering` unit test covers a path no browser framework exercises.
- **React and Vue don't dedupe the element; Angular does.** Two providers in one tree → two overlays. Previously unlikely with `showDevConsole` defaulting off; default-on makes it reachable.
- `assistantMessageToolbarInspectorLocalOnlyLabel` now reads `"Development Only"` — keeping the key is right for compat, but the name is now misleading in three packages' public label maps.
- Formatter churn: the `inspector.mdx` table realignment is still unrelated to the change, and `reference/index.mdx` + the troubleshooting page pick up a stray `;` after `</CopilotKit>` inside tsx fences.

### Branch state

One conflict: `examples/v2/react-router/app/routes/_index.tsx`, which `main` rewrote wholesale into the error lab. Resolution is "take main's version and drop `showDevConsole="auto"` at line 214" — main's rewrite still passes it.

### Verification

- **react-core** full suite at head: 1528 pass / 4 fail. Identical 4 on base `72fcd1dabd` (`use-interrupt` backend-owned interrupts, 3× `CopilotChat.raw-event-feedback`) — pre-existing local noise, no regressions.
- **vue** full suite: 1083/1083.
- **angular** full suite: 313 pass / 1 fail + 1 collect failure (`interrupt.spec.ts`, `modal-a11y.spec.ts`) — both identical on base. `inspector.spec.ts` 6/6.
- **shared** `inspector-visibility.test.ts`: 4/4.
- Local `tsc --noEmit` on react-core is not usable in my worktree (it resolves `@copilotkit/{core,web-inspector}` from a stale sibling `dist`, so `getRawEventForMessage` / `compileChatSnippet` / `configureWebInspectorElement` all read as missing). CI's `check-types` is green.
- **CI at `2bb60b67c5`**: 95 success / 3 skipped, nothing pending — including `check-types`, `format`, `oxlint`, `parity-check`, `bundle-size`, and all six Node×React legs.

skills/react-core/references/debug-mode.md
```diff
@@ -47,44 +48,31 @@ dump full message/tool-call payloads.
...
-<CopilotKit runtimeUrl="/api/copilotkit" showDevConsole={true} />
+<CopilotKit runtimeUrl="/api/copilotkit" showDevConsole="auto" />
 ```
 
 Correct:
 
 ```tsx
-<CopilotKit runtimeUrl="/api/copilotkit" showDevConsole="auto" />
-// "auto" enables only on localhost / 127.0.0.1
+<CopilotKit runtimeUrl="/api/copilotkit" enableInspector={false} />
```
AlemTuzlak: **[suggestion]** The HIGH warning is "Using `showDevConsole` to control the Inspector". The Wrong snippet is `showDevConsole="auto"` (the old way to show the Inspector on localhost). The Correct snippet is `enableInspector={false}`, which turns the Inspector off in development. An agent that follows this pair will hide the Inspector while it tries to stop using `showDevConsole`. The same text is in `packages/react-core/skills/react-core/references/debug-mode.md`.

**Suggestion:** Make the Correct snippet a plain `<CopilotKit runtimeUrl="/api/copilotkit" />` with no Inspector prop. Keep `enableInspector={false}` in a separate "hide it in development" example.

skills/react-core/references/debug-mode.md
```diff
@@ -47,44 +48,31 @@ dump full message/tool-call payloads.
...
-```
-
 ### Env-gate the inspector
 
 ```tsx
-<CopilotKit
-  runtimeUrl="/api/copilotkit"
-  showDevConsole={process.env.NODE_ENV !== "production"}
-/>
+<CopilotKit runtimeUrl="/api/copilotkit" enableInspector={false} />
```
AlemTuzlak: **[suggestion]** The "Env-gate the inspector" section now shows a hard `enableInspector={false}`. Production is already blocked by `shouldEnableInspector`. Copying this snippet disables the Inspector in local development too. The same section is in `packages/react-core/skills/react-core/references/debug-mode.md`.

**Suggestion:** Rename the section to "Disable the Inspector in development" or delete it. The default-on plus production guard already replaces env gating.

skills/react-core/references/debug-mode.md
```diff
@@ -133,21 +121,18 @@ Source: `packages/react-core/src/v2/providers/CopilotKitProvider.tsx` (DebugConf
...
+<CopilotKit runtimeUrl="..." enableInspector={false} />
 ```
 
 The inspector persists its anchor via `localStorage`. In sandboxed iframes
-without storage access, the component throws on mount. Either disable in
-iframes or whitelist storage in the sandbox attrs.
+without storage access, the component throws on mount. Disable it for an
+iframe deployment or whitelist storage in the sandbox attrs.
 
 Source: `packages/react-core/src/v2/components/CopilotKitInspector.tsx:16-53`
```
AlemTuzlak: **[nit]** The iframe note still points at `CopilotKitInspector.tsx:16-53`. That file no longer talks to `localStorage`. Persistence now lives in `@copilotkit/web-inspector`. The same citation is in `packages/react-core/skills/react-core/references/debug-mode.md`.

**Suggestion:** Point the source at the web-inspector persistence path, or drop the line numbers.

AlemTuzlak (approved):

BenTaylorDev (approved):

PR #6578, opened 2026-08-19
refactor(angular): make effect dependencies explicit via explicitEffect

### LINKED ISSUE #6561: 🚀 Feature Request: Make Angular `effect` dependencies explicit
### Pre-flight Checklist

### Problem or Motivation

Angular effects can make it difficult to tell which signals are intentionally tracked.

Introduce an internal `explicitEffect` helper that makes dependency tracking explicit, improving readability and reducing the risk of accidental dependencies.

`explicitEffect` accepts two functions. The first is a reactive computation: any signals read within it are tracked. The second is the execution function, which always runs in an untracked context. It receives the result of the reactive computation as its first argument and onCleanup as its second.

Examples:

```ts
explicitEffect(
  () => ({ threadId: threadId() }),
  ({ threadId }, onCleanup) => {
    // Imperative, untracked execution
  },
);
```

```ts
explicitEffect(
  () => threadId(),
  (threadId, onCleanup) => {
    // do something
  }
)
```

### 🎤 Tell us your idea

That is to improve the internal codebase

MannXo:
## What does this PR do?

Adds an internal `explicitEffect` helper to `@copilotkit/angular` and moves every
`effect()` call site in the package onto it.

A plain `effect()` tracks whatever its body happens to read. That means an
incidental signal read buried in imperative code silently becomes a re-run
trigger, and a reviewer has to trace the whole body to work out what the
dependency set actually is. `explicitEffect` splits the two halves: the first
function is the only reactive scope, and the second always runs untracked.

```ts
explicitEffect(
  () => ({ threadId: threadId(), agentId: agentId() }),
  ({ threadId, agentId }, onCleanup) => {
    const connection = connect(agentId, threadId);
    onCleanup(() => connection.close());
  },
);
```

Two deviations from the signature in the issue, both driven by existing call
sites: it returns the underlying `EffectRef` (`injectInterrupt` calls
`.destroy()` on it) and it forwards `CreateEffectOptions` so callers outside an
injection context can pass an `injector`.

The helper is internal. It lives in `src/lib/explicit-effect.ts` and is
deliberately not exported from `public-api.ts`, so the public API surface and
its manifest are unchanged.

## Changes

- New `src/lib/explicit-effect.ts` with JSDoc and usage examples.
- New `src/lib/explicit-effect.spec.ts`, 8 tests.
- All 16 `effect()` call sites across 12 files converted. `effect(` no longer
  appears in the package outside the helper itself.

Four sites already had an explicit `effect` + `untracked` boundary and convert
one-for-one with no change in tracking: `threads.ts` (both effects),
`active-thread-connector.ts`, and `copilot-chat-input.ts`.

The rest read signals from inside the imperative body, so those reads moved into
the dependency function to keep the tracked set the same. Worth a closer look
during review:

- **`copilot-sidebar.ts`** — `position()` and `resolvedWidth()` were previously
  tracked only on the docked path, because the effect returned early when
  undocked. They are unconditional dependencies now. The extra re-runs while
  undocked are inert: `releaseDock()` returns early without ownership, and
  `dockAccepted` is re-set to the value it already holds. There is a comment on
  this at the call site.
- **`copilot-threads-drawer.ts`** — same shape. It used to bail on a missing
  element before reading the other thirteen signals, so it now re-runs (and
  immediately returns) in that window.
- **`copilot-chat.ts`** — `hasExplicitThreadId()` was read after the first
  mutation in the standalone branch.
- **`inject-interrupt.ts`** — the chat configuration's `threadId()`, which is
  read only when no explicit agent id was passed. `configuredAgentId` is fixed
  for the lifetime of the call, so the dependency function gates on it the same
  way and the tracked set is unchanged.
- **`copilot-chat-assistant-message-renderer.ts`** — the body reaches
  `renderedHtml()` through `updateContent()`, but that is a computed over
  `content()`, which is already the declared dependency.

No behavioral change is intended anywhere. Where the dependency set genuinely
widened, the additional runs are no-ops.

## Testing

- `nx run-many -t build,check-types,test,publint,attw --projects=@copilotkit/angular` — all green, run with `--skip-nx-cache`.
- Unit tests: 49 files, 317 passing (1 pre-existing skip in `agent.spec.ts`, untouched here). 309 were passing on `main` before this branch; the 8 new ones are the helper's.
- Verified the new tests fail without the helper's untracked boundary, not just alongside it: removing the `untracked()` call makes the tracking test fail with `expected [ 'a:x', 'a:y' ] to deeply equal [ 'a:x' ]`, and the self-retrigger test loops until the worker runs out of memory.
- `oxfmt --check` and `oxlint` clean on all 14 changed files. `oxlint` reports one warning, a pre-existing `enabled` shadow in `threads.ts` that is present on `main` too.
- `pnpm check:public-api-manifest` — manifest unchanged, as expected for an internal helper.
- Rebased onto current `main`. `injectInterrupt` moved to its own file upstream in the meantime, so that call site was re-migrated there and `interrupt.ts` is left exactly as `main` has it. Verified both branches of its conditional dependency behave as before: the ambient thread id re-triggers the effect only when no explicit agent id was given.
- No `.changeset/` files.

## Related PRs and Issues

Fixes #6561

## Checklist

- [x] If the PR changes or adds functionality, I have updated the relevant documentation — n/a, the helper is internal and unexported, so there is no public reference page to add
- [ ] "Allow edits by maintainers" is checked

## Summary by CodeRabbit

* **Refactor**
  * Improved Angular reactive state handling with explicit dependency tracking across chat, thread, sidebar, audio, and generative UI components.
  * Preserved existing connection, cleanup, retry, rendering, and thread-management behavior.
  * Added a reusable reactive effect utility for more predictable updates and cleanup.

* **Tests**
  * Added comprehensive coverage for dependency updates, cleanup, injector destruction, and signal-write behavior.


packages/angular/src/lib/components/a2ui/a2ui-activity-renderer.ts
```diff
@@ -125,28 +125,32 @@ export class CopilotA2UIActivityRenderer implements ActivityRenderer<unknown> {
...
-      const content = this.content();
-      if (getA2UIOperations(content).length === 0) {
-        this.loaderContent.set(content);
-      }
-    });
-    effect((onCleanup) => {
-      const hasOperations = this.hasOperations();
-      if (!hasOperations) {
+    explicitEffect(
+      () => this.content(),
```
rainerhahnekamp: just `this.content` is enough

packages/angular/src/lib/components/a2ui/a2ui-activity-renderer.ts
```diff
@@ -125,28 +125,32 @@ export class CopilotA2UIActivityRenderer implements ActivityRenderer<unknown> {
...
+      (content) => {
+        if (getA2UIOperations(content).length === 0) {
+          this.loaderContent.set(content);
+        }
+      },
+    );
+    explicitEffect(
+      () => this.hasOperations(),
```
rainerhahnekamp: just the signal. this applies to all other usages of `explicitEffect` as well.

packages/angular/src/lib/components/chat/copilot-threads-drawer.ts
```diff
@@ -445,39 +445,69 @@ export class CopilotThreadsDrawer {
...
-      const collapsible = this.collapsible();
-      if (collapsible !== undefined) {
-        // TODO(ENT-1051): drop the intersection cast once the published element
-        // type declares `collapsible` (see the local CollapseChangeDetail note).
-        (
-          el as CopilotKitThreadsDrawerElement & { collapsible: boolean }
-        ).collapsible = collapsible;
-      }
-    });
+    explicitEffect(
```
rainerhahnekamp: That one seems to be so complicated that an explicit causes more confusion. I also believe that `afterRenderEffect` might here be the better one.

MannXo:
Closing this, since it has fallen behind main without being picked up for review. Happy to rebase and reopen if it's still wanted.

packages/angular/src/lib/inject-interrupt.ts
```diff
@@ -59,15 +54,22 @@ export function injectInterrupt<TValue = unknown, TResult = never>(
...
-    controller.setThreadId(
-      configuredAgentId === undefined
-        ? (chatConfiguration?.threadId() ?? agent.threadId)
-        : agent.threadId,
-    );
-  });
+  // `configuredAgentId` is fixed for the lifetime of this call, so gating the
+  // configuration read on it keeps the tracked set exactly as it was: the
+  // ambient thread id is a dependency only when no explicit agent was given.
+  const connection = explicitEffect(
```
rainerhahnekamp: that is dynamic dependency tracking where I would also say it is an exception to the rule and should be a normal `effect`.

packages/angular/src/lib/threads.ts
```diff
@@ -363,22 +361,33 @@ export class ThreadsStore implements InjectThreadsResult {
...
-      const status = runtimeStatus();
-      const headers = this.#copilotkit.headers();
-      const id = agentId();
-      const archived = includeArchived();
-      const pageLimit = limit();
-      const listSupported = threadListSupported();
-      const wsUrl = this.#copilotkit.intelligence()?.wsUrl;
-
-      untracked(() => {
+    explicitEffect(
```
rainerhahnekamp: just `effect`.

packages/angular/src/lib/agent-context.ts
```diff
@@ -21,13 +22,15 @@ export function connectAgentContext(
   runInInjectionContext(injector, () => {
     const copilotkit = inject(CopilotKit);
 
-    effect((teardown) => {
-      const contextValue = typeof context === "function" ? context() : context;
-      const id = copilotkit.core.addContext(contextValue);
+    explicitEffect(
+      () => (typeof context === "function" ? context() : context),
+      (contextValue, teardown) => {
```
rainerhahnekamp: `onCleanup` instead of `teardown`.

rainerhahnekamp (changes requested):

rainerhahnekamp:
@MannXo opened it again and ran the review.

PR #6581, opened 2026-08-20
feat(reskinnable-demo): run banking on a nested LangChain deep agent with a streaming CLI console

mxmzb:
Replaces the Codex CLI harness with **LangChain deep agents (Python)**: sandboxed shell execution, parallel research subagents, no external harness binary, our own API key.

All ten beats have been walked in a browser, and the demo now has its own CI job. **Ready for review.**

## What banking's agent is now

```
banking                       gpt-5.4, temp 0
  │  banking's own prompt · the browser's frontend tools · Intelligence
  │  memory tools · render_report for the canvas
  └─ expense-analyst          gpt-5.6-sol, reasoning_effort=high
       │  sandboxed shell (LocalShellBackend) · submit_expense_report
       └─ merchant-researcher  gpt-5.4, one per merchant
            search_merchant (Tavily)
```

Six skins still run in-process as `BuiltInAgent`s. `banking` is an `HttpAgent` pointed at `agent/`.

Each level exists because something about it must differ from its parent — prompt, model, reasoning effort, tool set. A single agent had nowhere to put any of that.

### Why the whole agent moved, not just the expense beat

The obvious design was a second agent id for the long-running beat, leaving banking's `BuiltInAgent` alone. That was built first, and it does not survive the requirement *"start the analysis, switch threads, run other pills, come back to it."*

**Threads are scoped per agent.** `listThreads` takes `agentId` as a required parameter, and measured against the running app the two lists were disjoint — `banking` 46 threads, `banking-expenses` 10. The v2 runtime has no handoff mechanism, and `defineTool`'s `execute` takes `(args)` with no emitter, so a tool cannot stream a multi-minute run into the conversation. One conversation list means one agent.

### Why the beat is a subagent rather than prose in one prompt

It started as a section appended to banking's prompt. That left per-beat configuration homeless (model, effort and recursion limit are all agent-level, and there was one agent), and made banking's ~21,000-character rulebook ride every one of the ~20 model calls the run makes — re-sending rules about markdown tables while the agent read a CSV.

Reached as a `CompiledSubAgent`, because a raw `SubAgent` spec has no `subagents` field and this one needs its own: the per-merchant fan-out is a headline of the beat, and a flat subagent could only research serially. A probe confirmed nesting survives — the analyst's `task` dispatches, the researchers' `search_merchant` calls and the final report tool all reach `astream_events`.

## The console

One CLI window in the transcript carries the whole run: narration, `$ execute`, `search "…"`, `→ merchant-researcher: …`, results, indented by depth so a ten-way fan-out reads as a fan-out. Subagent narration is suppressed from the conversation so the console is the single place the harness is visible; banking's own replies still render normally.

`shell/subagents/subagent-activity.tsx` subscribes to the agent's **event stream**. Reading `agent.messages` — the first design — was wrong twice: messages materialise at the `MESSAGES_SNAPSHOT` (two per run), so the pane sat still for minutes then filled at once; and persisted messages carry no `subagentRunId`, so the harness's narration could not be told from banking's reply. The fold is pure and idempotent, so the same code serves the live subscription and a replay of the thread's stored events when a conversation is reopened.

Three heuristics were deleted along the way, each replaced by identity the protocol already carried:

| heuristic | replaced by |
|---|---|
| console anchored on "the first tool call" | the run's first `task` call, from **message order** (durable; the event-derived version rendered one console *per delegation* on a restored thread — six, measured) |
| `CONSOLE_TOOL_NAMES` allowlist | presence of `subagentRunId` |
| `disable_streaming` on the researchers | the canary's per-lane state |

The message filter suppresses **prose** and keeps **tool calls**: an agent routinely narrates and calls a tool in one message, and returning `null` for the whole message hid the report card — the run looked perfect and ended with nothing to show.

## Canary stack, contained to this app

The subagent surface only exists on the canary line, and a released `@ag-ui/client` ≤ 0.0.57 rejects `SUBAGENT_*` events **in the HTTP transport before any middleware runs**, killing the stream. So the demo leaves the root pnpm workspace and ships its own lockfile:

- `@copilotkit/* 1.68.3-canary.1786716392`
- `@ag-ui/* 0.0.59-canary.1786716392.0`
- `ag-ui-langgraph 0.0.43.dev1786716392` → `ag-ui-protocol 0.1.20.dev1786716392`

That keeps an unreleased protocol out of every other package in the monorepo.

A 1.62.2-based canary was tried first and **could not compile the app**: `workspace:*` is not a version, so the demo had no recorded lower bound on the CopilotKit API it needs, and pointing it at 1.62.2 silently rewound that API five minors. `OpenGenerativeUIActivityRenderer` — a public `/v2` export since ~1.63 — was the first thing to break, and there was no reason to think it was the only one. Rebasing the canary on 1.68.3 collapsed that whole class of risk.

## Verified against the running app

Every row is a measurement, not a claim.

| | |
|---|---|
| Registry shape | `/info`: `banking` = HttpAgent, other six = BuiltInAgent |
| Sandboxed execution | 8 `execute` calls; agent writes and runs its own Python |
| Parallel subagents | 6 `SUBAGENT_STARTED`/`FINISHED` pairs, 10 `search_merchant` |
| Canvas beat | `render_report` → a2ui middleware emits `activityType: a2ui-surface` |
| Frontend tools | given two, picks `showTransactions`, emits no result, emits **no prose** — the prompt's restraint rule surviving the port |
| Memory (Intelligence) | run is handed `recall_memory`, `save_memory`, `forget_memory`, `knowledge_base_shell` |
| HITL round-trip | tool call out, answer back in, agent continues |
| Durable background run | client disconnected at 8s; run finished unattended; thread replayed |
| Thread restore | 61 messages persisted incl. the report tool (an earlier flat-subagent attempt collapsed to 4) |
| Report correctness | 14 rows, 9 researched, 6 filed with ids read out of real 201 bodies, totals reconciling against their own rows |
| Run duration | 86s reported vs 98s wall clock |

## Correctness bugs found by running it

Each produced a confident, complete-looking wrong answer rather than an error — the characteristic failure of a multi-minute agentic beat.

- **Beat 3d was dead, and it took the whole thread with it.** The bet against this one was right. `ag_ui_langgraph` routes *every* attachment to the model as an `image_url` block, documents included, so the Q2 invoice was rejected before the first token: `400 Invalid MIME type. Only image types are supported`. The exception is raised inside the model node, which kills the SSE stream — the runtime sees `RUN_ERROR: terminated` with no cause and the browser renders **nothing**: no error bubble, no failed message. And the crashed run is still checkpointed, so every later message on that thread replays the rejected content and dies the same way. One click on the pill killed the conversation permanently; only restarting the service cleared it. Fixed here by a `wrap_model_call` middleware that rewrites those blocks into LangChain standard `file` blocks, and upstream in [ag-ui#2476](https://github.com/ag-ui-protocol/ag-ui/pull/2476) (both adapters, plus the return leg so a non-image attachment survives `MESSAGES_SNAPSHOT`). The middleware is a stopgap with its deletion condition written into its header — this service installs the adapter from PyPI, so the upstream fix cannot reach it until published.
- **Totals did not match their own rows.** Every per-row amount matched the CSV while the headline totals came back $1.00 and $0.20 high: the model authored them instead of adding them. The card prints the total directly above the rows it is the sum of. Fixed structurally — derived in `submit_expense_report`, removed as parameters.
- **`amount` arrived as a string** on all 14 rows, silently defeating `toLocaleString(…currency…)` so it printed `842.10` with no currency.
- **`merchantKind` non-answers.** With no search tool the model wrote a bare `"unclear"`; with Tavily live it hedges in prose (`"unknown (likely bookbindery/bookshop retail, but not established for this exact merchant)"`), which an exact-match filter passed into a 60-character label glued to the merchant name. Now rejected on leading token, hedging language, and a 40-character cap.
- **The run clock reported 333s for a two-minute run.** It took the oldest open stamp across the process because the tool could not name its own run; model calls *after* the report re-stamped it and the leftover became the next run's start. Now keyed per run via an injected `ToolRuntime`.
- **`graph.with_config({"recursion_limit"})` is silently dropped** by the AG-UI adapter. The agent completed the entire analysis, streamed every argument of the final report, then died at LangGraph's default of 25 supersteps.
- **A sync `wrap_model_call` under `astream`** surfaces as a bare `RUN_ERROR: terminated`, cause only in the service log.
- **`emit_raw_events` defaults to `True`**, piggybacking LangChain's internal events onto the stream: 27,950,261 bytes → 374,086 with it off, identical report. Matters because the thread *persists* those events for replay.
- **`gpt-5.6-sol` + `reasoning_effort` + function tools 400s** on `/v1/chat/completions`; needs `use_responses_api=True`. The first model probe missed it by binding no tools — a model probe for an agent has to bind one.

## Upstream findings (reported separately, not fixed here)

1. **`@copilotkit/runtime` drops `subagentRunId` when persisting messages.** 2888 of 3026 stream events carry it; 0 of 53 persisted messages do. Reproduced with Intelligence removed entirely, so it is the runtime's message shape rather than the platform store — and `@copilotkit/runtime`'s dist contains no occurrence of the field, while `@ag-ui/core`, `ag-ui-protocol` and `@copilotkit/core` all model it. One field threaded through would let the console rebuild from message history and delete the event-replay seeding here.
2. **`copilotkit` 0.1.95 × `ag-ui-langgraph` 0.0.43** — the FastAPI endpoint calls `agent.clone()` per request; 0.0.43's base `clone()` hard-passes three kwargs the SDK subclass does not accept, so **every request 500s**. Verified with a minimal repro on stock classes and by reading published wheels (0.0.41/0.0.42 are fine — the window is 0.0.43 only). `sdk-python` requires `>=0.0.42` unbounded, so fresh installs break, and `examples/showcases/deep-agents{,-finance-erp,-job-search}` are one `uv lock --upgrade` away. Now open as #6592 (`**kwargs` passthrough + 9 regression tests incl. a forward-compat guard); it needs a `0.1.96` bump to reach PyPI before `main.py`'s `BankingAGUIAgent.clone()` workaround can go — and note `agent/uv.lock` pins `copilotkit 0.1.95`, so removing the workaround is a re-lock as well as a delete. That branch also fixes 4 pre-existing failures in the sdk-python suite from the same root cause — **`test_unit-python-sdk` may currently be red on `main`; worth checking independently.**

3. **Both LangGraph AG-UI adapters send non-image attachments as `image_url`**, so a PDF, audio clip or video is rejected on the block kind — see beat 3d in *Correctness bugs*. Fixed in [ag-ui#2476](https://github.com/ag-ui-protocol/ag-ui/pull/2476); carried here as a middleware until that publishes.

Also verified and **not** a problem: pnpm 10.33.4 still applies `package.json` `pnpm.overrides` despite warning that it ignores them — the ~70 root overrides including the security pins are live, nothing was silently unpinned. (Migration to `pnpm-workspace.yaml` is worth doing anyway, since the installed pnpm is 11.21.0 where the field genuinely is dropped; branch exists, byte-identical lockfile.)

## What is not done

1. **Docs are stale.** `CLAUDE.md` still says `AgentRegistration` is `{ createAgent: () => BuiltInAgent }` and that the route "builds one `BuiltInAgent` per registered skin"; the reskin skill says the same in three places, and `templates.md` scaffolds it. The real type is `() => AbstractAgent`. This app has a standing rule that every change answers whether the skill went stale. It did — the skill's launcher step is fixed, the `BuiltInAgent` claims are not.
3. **Pre-release dependencies**, now on both halves: the JS canaries in this app's own `pnpm-lock.yaml`, and `ag-ui-langgraph` / `ag-ui-protocol` pinned `==` to the matching `.dev` canaries in `agent/pyproject.toml`. Both move to stable when the subagent work ships.
4. `pnpm test:e2e` has not been run.

### Closed since this PR was opened

- **The demo has its own CI job** — `.github/workflows/test_reskinnable-demo.yml`. Nx discovers projects *through* the pnpm workspace (no `workspaceLayout` in `nx.json`), so leaving it also left the repo-wide `nx run-many -t build` (`static_compat.yml`), `-t check-types` (`static_quality.yml`) and `-t test` (`test_unit.yml`) sweeps, and both static workflows carry `paths-ignore: ["examples/**"]` besides. One job runs the four gates (lint, typecheck, 2460 unit tests, build); a second syncs `agent/uv.lock` with `--frozen` and asserts the subagent *capability* rather than a version string, so it keeps meaning something after the pin moves.

  The build gate is the one a developer can least run locally: `next build` corrupts a concurrently running dev server's PostCSS/Turbopack cache — measured, `globals.css` transforms to garbage, every route 500s, and a dev-server restart does not clear it because the corruption is on disk.

  **Its first two runs failed, both usefully**, and both bugs were pre-existing:

  1. `pnpm/action-setup@v6.0.10` with `package_json_file:` pointed at this app still installed the **root's** pnpm 10.33.4 rather than this app's pinned 10.10.0 — the resolver that ignores `pnpm.overrides`, which is the only place three of the five `@ag-ui/*` canaries are pinned. Caught only because the job asserts `pnpm --version` against the pin; it would otherwise have installed under the wrong resolver and stayed green. Now activated through corepack, which reads the nearest `package.json`.
  2. `pnpm install` in this directory **never installed this app**. pnpm walks *up* for a workspace root, found the repo's, and installed all 70 monorepo projects (4645 packages) while leaving this directory with no `node_modules` — the next command failing as `sh: 1: eslint: not found`. So the app that ships its own lockfile was uninstallable by its own documented instruction. Fixed by giving it its own `pnpm-workspace.yaml`; `ignore-workspace=true` in an `.npmrc` does **not** work (CLI-only in pnpm 10.10, measured).

- **The canary `overrides` now live in `pnpm-workspace.yaml`**, their supported home, as a side effect of that fix. They no longer depend on `packageManager: pnpm@10.10.0` being the version that still reads `package.json` — which matters because `@ag-ui/core`, `@ag-ui/encoder` and `@ag-ui/proto` are pinned nowhere else and a released `@ag-ui/core` rejects `SUBAGENT_*` in the HTTP transport.
- **All ten beats walked in a browser**, including 3d — which was broken, for the reason in *Correctness bugs* above.
- **`./run-demo.sh` starts a complete demo.** It now brings up the Python agent between the compose wait and `pnpm dev`, guarded on `/health` so a re-run reuses a live one, and dies with `(cd agent && uv sync)` when the venv is missing. `./stop-demo.sh` had the mirror gap and now stops it too — that one mattered more than it looks: the start script *reuses* a live :8124, so an orphan surviving teardown is silently adopted by the next cold start, serving whatever code it was launched with.
- **The agent's Python deps are pinned and locked.** They were not, and the JS half was — so a colleague's fresh `uv sync` resolved the *release* `ag-ui-langgraph 0.0.43`, which does not accept `emit_subagent_events` and ships no subagent symbols. That is this PR's headline feature, and it would have failed silently: `main.py` sets the flag as an attribute, so the assignment succeeds against an object nobody reads and the service starts clean. Now `==` pins plus a committed `agent/uv.lock`, verified by building a venv from only the tracked files and replaying a captured Q2-with-PDF payload through it.
- **The README told people to install from the repo root** "as a workspace package". It is deliberately not one, so a root install did nothing for this app — and installing in the right place did nothing either, until the `pnpm-workspace.yaml` above.

## Relationship to #6501

Based on `b94e4bfb5d`, the last commit before Codex appears, which is **not on `main`**. So this PR carries **9 commits: the 4 here plus the 5 foundation commits it shares with #6501** — harness types + fixture, prompt/workspace, the OFFSITE-to-fixture invariant guard, the summary-shape/filing-contract fix, and `POST /transactions`. Whichever merges first shrinks the other. #6501 and #6565 are deliberately untouched.

## A demo-design question, not a bug

The fixture's merchant names are invented, so `Cardinal & Ash` — the prompt's own worked example of *"could be a restaurant or a law firm, find out"* — cannot be resolved by real web search and stays `unclear` alongside `Bluebonnet Provisions`. The beat's headline claim is that the agent researches every merchant; real merchant names in the CSV would make that land harder.

https://claude.ai/code/session_01JRubZT6AS6LCGkcE2KzcfA


davidmckayv (approved):
Approving to unblock the merge: all 32 checks green, no conflicts. Taking it for a local run on the self-hosted memory path now — will report back if `run-demo.sh` hits anything on a machine that already has the Intelligence stack and embedder up.

PR #6584, opened 2026-08-20
feat(runtime): add OrcaRouter as a built-in model provider

XiaoHuo888-hue:
## What does this PR do?

Adds **OrcaRouter** as a named, built-in model provider for the Built-in Agent, mirroring the existing MiniMax gateway wiring in `resolveModel()`.

[OrcaRouter](https://www.orcarouter.ai) is an OpenAI-compatible gateway that routes each request to the best model for the task through a single API key. It also runs gateway-level, zero-trust security for AI agents on the same endpoint — screening every prompt/response and governing every tool call on a default-deny basis, with no application code changes.

Changes:

- `packages/runtime/src/agent/index.ts`:
  - Adds `orcarouter/auto`, `orcarouter/fusion`, `orcarouter/fusion-flash`, `orcarouter/fusion-mini` to `BuiltInAgentModel`.
  - Adds an `orcarouter` case in `resolveModel()` that builds an `@ai-sdk/openai` provider pointed at `https://api.orcarouter.ai/v1`, keyed by `ORCAROUTER_API_KEY` (overridable via `ORCAROUTER_BASE_URL`).
  - Uses `.chat()` (chat completions) with the re-prefixed `orcarouter/` model id, because the gateway requires the fully-namespaced id and does not expose the Responses API.
  - Updates the supported-provider error message and the `apiKey` doc comment.
- `packages/runtime/src/agent/__tests__/resolve-model-baseurl.test.ts`: adds unit tests for the new provider, mirroring the MiniMax cases.
- `showcase/shell-docs/src/content/docs/integrations/built-in-agent/model-selection.mdx`: documents the new provider (models, specifiers, env vars, resolution table).

## Verification

- `npx vitest run src/agent/` — 365 tests pass (22 files), including 5 new OrcaRouter cases.
- `npx tsc --noEmit` on `@copilotkit/runtime` — no new errors (the 26 pre-existing channel-package resolution errors are identical on a clean checkout).
- `npx oxlint` and `npx oxfmt --check` on changed files — clean.
- Live test against the real gateway through the compiled `resolveModel()` path: `orcarouter/auto`, `orcarouter/fusion`, `orcarouter/fusion-flash`, `orcarouter/fusion-mini` each returned a 200 chat completion (`ORCA-LIVE-OK`).

## Checklist

---

Disclosure: I'm an engineer on the OrcaRouter team.

### DIFF
--- a/packages/runtime/src/agent/__tests__/resolve-model-baseurl.test.ts
+++ b/packages/runtime/src/agent/__tests__/resolve-model-baseurl.test.ts
@@ -3,10 +3,23 @@ import { resolveModel } from "../index";
 
 // Mock the SDK provider factories so we can assert the options resolveModel passes them
 // (the returned LanguageModel does not expose baseURL publicly, so we verify at the factory).
-const createOpenAI = vi.fn((_opts?: unknown) => (modelId: string) => ({
-  modelId,
-  provider: "openai",
-}));
+const createOpenAI = vi.fn((_opts?: unknown) => {
+  // A callable provider factory that also exposes `.chat()` — the OrcaRouter
+  // branch uses `.chat()` (chat completions) while the OpenAI / MiniMax
+  // branches invoke the factory as a plain callable.
+  const callable = ((modelId: string) => ({
+    modelId,
+    provider: "openai",
+  })) as unknown as {
+    (modelId: string): { modelId: string; provider: string };
+    chat(modelId: string): { modelId: string; provider: string };
+  };
+  callable.chat = (modelId: string) => ({
+    modelId,
+    provider: "orcarouter.chat",
+  });
+  return callable;
+});
 const createAnthropic = vi.fn((_opts?: unknown) => (modelId: string) => ({
   modelId,
   provider: "anthropic",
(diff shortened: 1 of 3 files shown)


BenTaylorDev:
Thank you for building this out — the implementation is careful, and the `.chat()` detail in particular shows you'd actually tested against the gateway rather than assuming the default path worked.

I'm closing it, for a reason that isn't about code quality.

OrcaRouter is already supported. The [Model Selection](https://docs.copilotkit.ai/integrations/built-in-agent/model-selection) page documents that anything exposing an OpenAI-compatible API works through the `createOpenAI({ baseURL })` pattern with no CopilotKit-specific provider — OpenRouter, self-hosted gateways, internal proxies, and OrcaRouter alike. Since OrcaRouter serves an OpenAI-compatible endpoint at `https://api.orcarouter.ai/v1`, users can point `baseURL` at it today and it works.

Adding a built-in case buys the four namespaced ids and an env var, and costs us a hardcoded base URL plus a model list to keep current as your catalog changes. It also sets a precedent we'd have to apply evenhandedly to every gateway that asks, which isn't a list we want inside `resolveModel`. For comparison, LibreChat documents OrcaRouter as a custom OpenAI-compatible endpoint rather than shipping a provider case, which is the same shape of answer.

One thing worth flagging for your own docs regardless of this PR. The comment here says:

> the default Responses API path is not supported by the gateway

Your public materials say the opposite — that the gateway supports Chat Completions, Responses, Embeddings, Images, Audio and streaming, and individual model pages list both `/v1/chat/completions` and `/v1/responses`. If Responses genuinely is supported, the `.chat()` workaround isn't needed and plain `createOpenAI({ baseURL })` is enough. If it isn't supported for routed ids like `orcarouter/auto`, that's a caveat your docs may want to state. Either way one of the two is wrong, and you're better placed than I am to say which.

If you'd like OrcaRouter to be discoverable to CopilotKit users, a line in the OpenAI-compatible providers list on that Model Selection page is something I'd review — that's how Novita was handled recently. Thanks again for the work here.

Issue #6571, opened 2026-08-19
Human-in-the-loop gate silently never runs while the model reports it did: Angular resolves a bus envelope, and AG-UI ToolMessage has no name field


BenTaylorDev:
## Summary

`@ag-ui/langgraph` does not populate `ToolMessage.name` for a human-in-the-loop
answer. It sends the tool name and payload inside the JSON content instead. The
canonical LangGraph routing pattern — match on `ToolMessage.name` — therefore falls
through **silently**, the gated node never runs, and the model then tells the user
the gated action completed.

That last part is what makes this severe rather than annoying.

## What arrives

Expected by any consumer routing on tool identity:

```python
message.name == "request_page_oncall"
```

What actually arrives — `name` unset, identity and payload in the content:

```json
{"toolName": "request_page_oncall", "result": {"approved": true}}
```

A working consumer has to be defensive in **two** places at once:

```python
name = message.name or payload.get("toolName") or payload.get("tool_name")
result = payload.get("result")
answer = result if isinstance(result, dict) else payload
approved = answer.get("approved")
```

## Why this is worse than a routing bug

Observed in a real run: an operator approved a paging action in the browser. The
router fell through, the gated node never executed, **and the model told the
operator "the on-call engineer has been paged."** That was false.

The failure mode is the inverse of the one a developer guards against. They are
watching for an *unwanted* side effect. This is a *confident false claim* that a
side effect occurred:

- the gate did not run, so nothing happened
- so the side effect's audit trail stays clean
- **a clean audit trail is exactly what a correctly working gate looks like**

The evidence of safety and the evidence of total failure are identical from inside
the application. A developer who checks "did we page anyone? no" concludes the gate
works. In production the distinguishing signal is that nobody arrives.

## No prior art in this repository

Unlike the tool-descriptor shape issue filed alongside this one, nothing here works
around it, because nothing exercises it.

- The only example routing on an incoming `ToolMessage` is
  `examples/v1/travel/agent/src/agent.py`, and it is a **type check, not a name
  match**:

  ```python
  if messages and isinstance(messages[-1], ToolMessage):
      return "chat_node"
  ```

  It routes any tool result to the chat node, never needs the tool's identity, and
  so never discovers that `name` is unset.

- `examples/integrations/langgraph-python` only ever **constructs** a `ToolMessage`
  (`agent/src/todos.py:36`) as a tool's own result. It never routes on an incoming
  one.

[...]

PR #6586, opened 2026-08-20
fix(angular): resolve human-in-the-loop results without the bus envelope

Ash20pk:
Fixes the Angular half of #6571.

## The bug

`HumanInTheLoop.onResult` pipes the rxjs bus through `filter` + `take(1)` but never maps, so `lastValueFrom` resolves with the **whole bus event** rather than the result the human supplied:

```ts
// packages/angular/src/lib/human-in-the-loop.ts (before)
onResult(toolCallId: string, toolName: string): Promise<unknown> {
  return lastValueFrom(
    this.results.pipe(
      filter((result) => result.toolCallId === toolCallId && result.toolName === toolName),
      take(1),
    ),
  );
}
```

`toolCallId` and `toolName` are routing keys for this bus only. Leaking them turns the tool result into an envelope, and the chain carries it all the way to the agent:

1. `copilotkit.ts:534` — the bound HITL tool handler returns `onResult(...)` directly as the tool's result.
2. `packages/core/src/core/run-handler.ts` — a non-string handler return is `JSON.stringify`'d into the tool message `content`.

So a LangGraph agent receives:

```json
{"toolCallId": "call-1", "toolName": "request_page_oncall", "result": {"approved": true}}
```

where React sends the bare `{"approved": true}` — `use-human-in-the-loop.tsx:24` resolves the handler promise with `result` itself. This is a cross-framework inconsistency as much as a bug: the same agent cannot serve both adapters.

## Why it matters

An agent gating a side effect on a field of the human's answer reads `approved` off the envelope, finds nothing, and the gate silently never fires — while the model still reports the action succeeded. That is the failure #6571 describes: the gated node never runs, so the side effect's audit trail stays clean, and a clean audit trail is exactly what a correctly working gate looks like from inside the application.

## The fix

Map to the bare result:

```ts
take(1),
// Resolve with the bare result. toolCallId/toolName are routing keys for
// this bus only — leaking them would make the tool result an envelope
// that no consumer expects.
map((entry) => entry.result),
```

## Tests

The two existing tests asserted the envelope as correct behaviour, which is why this survived:

```ts
await expect(promise).resolves.toEqual({
  toolCallId: "call-1", toolName: "approval", result: { status: "ok" },
});
```

They now pin the bare result, and a third test fails if the routing keys leak back in. Verified that all three fail with the fix reverted and pass with it. `nx run @copilotkit/angular:test` → 304 passed; `check-types` and `build` clean.

## Scope

`onResult` is called only from `#bindHumanInTheLoopTool` (`copilotkit.ts:534`); the sole other spec that touches it mocks the method wholesale, so nothing else depends on the old shape. Any app that had worked around the leak by reading `.result` itself would need to drop that workaround — worth a line in release notes.

## Not addressed here

#6571's primary claim — that `ToolMessage.name` is unset — is a **separate defect** and is untouched by this PR. For the record, since the issue attributes both symptoms to `@ag-ui/langgraph`:

- The `{toolName, result}` envelope does **not** come from `@ag-ui/langgraph`. It originates in this file. Grepping that package for it turns up nothing, which risks the whole report being discounted.
- The missing `name` **is** in `@ag-ui/langgraph` — `aguiMessagesToLangChain()`, `case "tool"` omits it. It is **not fixed in 0.0.43**; that release only added `status`, so bumping the pin from `0.0.42` will not help. The root cause is that `ToolMessageSchema` in `@ag-ui/core` is the only message role without a `name` field. The adapter can still recover it by matching `toolCallId` against the preceding assistant message's `toolCalls[]` in the same loop — `sdk-python/copilotkit/copilotkit_lg_middleware.py:809` and `langgraph.py:129` already do exactly that lookup.

Two related spots worth separate fixes: `packages/runtime/src/service-adapters/langchain/utils.ts:45` builds a `ToolMessage` and discards `message.actionName`, which is already in scope; and `examples/showcases/scene-creator/agent/agent.py:741` reads `msg.name`, which is `None` for frontend and HITL results — a live instance of the bug sitting in an example.


BenTaylorDev (changes requested):
Thanks for this — the diagnosis is right, and the writeup on #6571 that separated this
half from the upstream half saved us real time. I verified the fix independently rather
than taking the tests at their word:

- Reverting only `human-in-the-loop.ts` to main's version while keeping your three specs
  fails all three (`Received: "request_page_oncall"` on the leak guard). They bind to the
  mechanism — not self-fulfilling.
- `onResult` has exactly one production consumer (`packages/angular/src/lib/copilotkit.ts:530`),
  so the change is complete in code.
- I read React and Vue directly to check the direction of the parity claim: both resolve the
  bare result, so this moves Angular toward them.
- Full Angular suite is 290/291 with the fix; the one failure reproduces on clean `main`
  without your branch.

## One change needed before merge

The envelope is documented as behavior, so merging as-is leaves three sentences actively
wrong in `showcase/shell-docs/src/content/reference/angular/functions/registerHumanInTheLoop.mdx`:

- **line 8** — "resolves the tool with the serialized result envelope described below"
- **line 112** — "The current implementation resolves the internal handler with
  `{ toolCallId, toolName, result }`; core serializes that object to JSON for the tool result."
- **line 178** — "sends `\"confirmed\"` or `\"rejected\"` in the serialized result envelope
  described above"

This matters more than a stale doc usually would. `onResult` returns `Promise<unknown>`, so
TypeScript cannot catch a consumer that assumed the envelope — the doc is the only thing
telling Angular users what shape to expect, and after this change it would steer them into
`result.result`, which is the same class of bug the PR fixes.

It's hand-written MDX, not generated, so it won't self-correct. Once those three read as the
bare result, this is good to go from my side.

Ash20pk:
Good catch on the docs — agreed that with `Promise<unknown>` the MDX is the only contract Angular users get, so leaving the envelope described would recreate the bug one layer up.

Pushed the fix: all three sentences in `registerHumanInTheLoop.mdx` (lines 8, 112, 178) now describe the bare value. Line 112 also spells out the serialization rule from `run-handler.ts` (string stored as-is, anything else `JSON.stringify`'d) and notes the shape matches React and Vue, since parity was the point.

Thanks for reverting the source to confirm the specs bind to the mechanism — that's the check I'd want too.

BenTaylorDev (approved):
The docs ask is resolved — thank you for going past what I asked. All three sentences read as the bare value now, and line 112 additionally spells out the serialization rule (string stored as-is, anything else `JSON.stringify`'d by core) and states the parity with React and Vue. Since `onResult` returns `Promise<unknown>` and TypeScript can't catch a consumer that assumed the envelope, that paragraph is now doing the work the type system can't. The only remaining mention of the envelope in the file is the correct negation.

I re-verified against the current tree rather than trusting the earlier round, since main has moved twice under this branch:

- **Mutation check still holds.** Reverting only `human-in-the-loop.ts` to main's version, with everything else at your head, fails all three specs — including the leak guard. Restore it and they pass. They bind to the mechanism, not to themselves.
- **Differential full-suite run, with and without the fix: zero regressions.** The fix flips exactly your three specs from red to green and moves nothing else.
- **`tsc --noEmit`:** no errors in anything this PR touches.
- **No drift.** Main hasn't touched any of the three files since your last merge, and there are no conflicts against current main.

One note for whoever writes the release line: this is a behavior change, not just a fix. Any app that had worked around the leak by reading `.result` off the envelope itself will need to drop that workaround. Worth calling out explicitly, since the failure mode on the other side is silent.

Nice diagnosis on this one — separating the Angular half from the upstream `ToolMessage.name` half on #6571 is what made it tractable, and the comment you left on the `map` explaining that `toolCallId`/`toolName` are routing keys for that bus only is the kind of thing that stops this regressing in a year.

BenTaylorDev:
The CI timeouts are a known issue predating this PR, and the tests covering these changes are green.

Ash20pk:
Checked the build logs. This isn't a build failure, and it isn't `timeout-minutes` either, which is what the alert template guesses.

All 19 slots died at the same instant, 20:51:44-50, exactly 15 minutes after they started at 20:36:43. The per-slot budgets in the matrix are 25 and 30 minutes, so nothing had come close to its own timeout. The log ends with:

```
failed to solve: Canceled: context canceled
##[error]Step canceled by GitHub
```

An external cancellation of the run, not a slot exceeding its budget. Nothing in the diff could have caused it: this PR touched one Angular `.ts`, its spec, and an `.mdx` doc page, none of which any of those Dockerfiles read. The slots that did finish are just the ones that happened to finish before the cancel landed, not a pattern of which services build cleanly.

Nothing to fix here, and nothing is stale on staging. Every `Showcase: Build & Push` run since has gone green, so those 19 images have been rebuilt and redeployed many times over by later merges to main. Re-running failed jobs on a month-old run would only rebuild that commit's code, which is long superseded.

Two things worth doing separately, if anyone wants to pick them up:

- The alert says cancelled slots are "typically killed by `timeout-minutes`". Here that sent me looking at build durations, which was the wrong place. It could tell the difference cheaply: a slot killed by its own budget runs for exactly its `timeout` value, whereas a run-level cancel kills every in-flight slot at the same second regardless of how long each had been going.
- Worth knowing what cancelled the run. `showcase_build.yml` deliberately has no concurrency group, and the header comment explains at length why that's load-bearing, so a cancel arriving anyway is either manual or from outside the workflow. I can't tell which from the logs available to me.

PR #6588, opened 2026-08-20
fix(runtime): normalize nested LangGraph tool metadata

### LINKED ISSUE #6570: Frontend tool descriptors arrive in three shapes; reading the wrong one silently yields an empty schema
## Summary

A frontend tool descriptor reaches a LangGraph Python agent in **three different
shapes**, and reading the wrong one silently produces an **empty JSON Schema**. The
model then calls the tool with no arguments, and the failure surfaces several layers
away as missing data in the rendered UI.

Three independent implementations in this repository each solved a *different
subset* of the variance. None of them is the scaffold a developer is pointed at.

## The three shapes

| Variance | Where it appears |
| --- | --- |
| `state["tools"]` | the documented location |
| `state["copilotkit"]["actions"]` | a second location, inside the CopilotKit envelope |
| flat `{name, description, parameters}` vs OpenAI-nested `{function: {name, description, parameters}}` | either location can use either shape |

`examples/canvas/langgraph-python/agent/agent.py` handles all of it, and its own
comment concedes the instability:

```python
# Frontend tools may arrive either under state["tools"] or within the CopilotKit envelope
raw_tools = state.get("tools", []) or []
try:
    ck = state.get("copilotkit", {}) or {}
    raw_actions = ck.get("actions", []) or []
    if isinstance(raw_actions, list) and raw_actions:
        raw_tools = [*raw_tools, *raw_actions]
except Exception:
    pass
```

and separately, around line 217:

```python
# OpenAI tool spec dict: { "type": "function", "function": { "name": "..." } }
tool.get("function", {}) if isinstance(tool.get("function", {}), dict) ...
```

Note the bare `except Exception: pass` — that is defensive coding around a contract
nobody trusts.

## Coverage today

| Implementation | location variance | nesting variance |
| --- | --- | --- |
| `examples/canvas/langgraph-python` | handled | handled |
| a 2026-08-19 validation run, written from scratch | not handled | handled |
| **`examples/integrations/langgraph-python` (the scaffold)** | **n/a — never reads `state["tools"]` at all** | **n/a** |

The shipped scaffold does not work around this because it never exercises it. Its
only `bind_tools` call (`agent/src/a2ui_dynamic_schema.py:73`) binds a backend tool.
So the reference implementation a developer is directed to does not demonstrate the
pattern, and the one place that gets it right is not a scaffold anyone is pointed at.

## Why it matters

Reading top-level `parameters` when the payload is nested yields `None`, which falls
back to `{"type": "object", "properties": {}}`. Nothing raises. The model receives a
tool with no parameters, calls it with no arguments, and the symptom appears as a
rendered card with missing values — in the run that found this, an approval card
reading `Severity: unknown`.

This affects anyone binding CopilotKit frontend tools into a LangGraph model, which
is the core generative-UI and human-in-the-loop pattern.

## Proposed fix

Normalize all three shapes in `packages/runtime`, which owns the LangGraph bridge
for CopilotKit users. `@ag-ui/langgraph` is pinned at `0.0.42` in both
`packages/runtime/package.json` and `packages/sdk-js/package.json`, so upstream
contracts will keep moving; absorbing the variance at our boundary means users stop
reimplementing three different partial workarounds.

Documenting the shapes is a cheaper alternative but leaves every consumer writing
defensive parsing, and does not help the ones who never read that page.

## How this was found

A LangGraph Python + Angular onboarding validation run on 2026-08-19, building a
human-in-the-loop approval gate. Related: the `ToolMessage.name` issue filed
alongside this one, which the same run hit at the next step.

LazyPeterGriffin:
## Summary

- normalize OpenAI-style nested LangGraph tool descriptors when building `copilotkit.actions`
- preserve the original `function` object while exposing `name`, `description`, and `parameters` at the action top level
- add regression coverage for the nested descriptor shape

Fixes #6570

## Validation

- `nx run @copilotkit/runtime:test --skip-nx-cache -- src/lib/runtime/agent-integrations/langgraph/__tests__/dispatch-event-filtering.test.ts`
- `nx run @copilotkit/runtime:check-types --skip-nx-cache`
- `oxfmt --check packages/runtime/src/lib/runtime/agent-integrations/langgraph/agent.ts packages/runtime/src/lib/runtime/agent-integrations/langgraph/__tests__/dispatch-event-filtering.test.ts`
- `oxlint packages/runtime/src/lib/runtime/agent-integrations/langgraph/agent.ts packages/runtime/src/lib/runtime/agent-integrations/langgraph/__tests__/dispatch-event-filtering.test.ts`

### DIFF
--- a/packages/runtime/src/lib/runtime/agent-integrations/langgraph/__tests__/dispatch-event-filtering.test.ts
+++ b/packages/runtime/src/lib/runtime/agent-integrations/langgraph/__tests__/dispatch-event-filtering.test.ts
@@ -303,6 +303,32 @@ describe("langGraphDefaultMergeState", () => {
     expect(result.copilotkit.actions).toEqual(expect.arrayContaining(tools));
   });
 
+  it("exposes nested function metadata at the action top level", () => {
+    const { agent } = createAgent();
+    const tool = {
+      type: "function",
+      function: {
+        name: "lookupWeather",
+        description: "Look up the weather",
+        parameters: { type: "object", properties: {} },
+      },
+    };
+
+    withMockedParentMerge(agent, {
+      "ag-ui": { tools: [tool], context: [] },
+      tools: [],
+      messages: [],
+    });
+
+    const result = agent.langGraphDefaultMergeState({} as any, [], {} as any);
+    expect(result.copilotkit.actions[0]).toMatchObject({
+      name: "lookupWeather",
+      description: "Look up the weather",
+      parameters: { type: "object", properties: {} },
+    });
+    expect(result.copilotkit.actions[0].function).toEqual(tool.function);
+  });
+
   it("merges copilotkit context from ag-ui", () => {
     const { agent } = createAgent();
     const context = [{ description: "user info", value: "test" }];
--- a/packages/runtime/src/lib/runtime/agent-integrations/langgraph/agent.ts
+++ b/packages/runtime/src/lib/runtime/agent-integrations/langgraph/agent.ts
@@ -35,6 +35,25 @@ import type {
 import { CustomEventNames } from "./consts";
 export { CustomEventNames };
 
+function normalizeLangGraphTool(tool: any): any {
+  if (!tool || typeof tool !== "object") return tool;
+
+  const functionSpec = tool.function;
+  if (!functionSpec || typeof functionSpec !== "object") return tool;
+
+  const normalized = { ...tool };
+
+  // LangGraph can expose OpenAI-style { function: { ... } } descriptors,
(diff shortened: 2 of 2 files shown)


LazyPeterGriffin:
The local LangGraph runtime test suite, type-check, formatting, and lint checks pass for nested tool metadata normalization. The branch is clean and ready for maintainer review.

LazyPeterGriffin:
@jpr5 @tylerslaton The fork-originated CI workflows are currently `action_required` until a repository maintainer approves their execution. Local runtime tests, type-check, formatting, and lint are green; approving the runs will expose the full repository CI result.

PR #6589, opened 2026-08-20
refactor: isolate v1 under deprecated source folders

ataibarkai:
## Stack

This PR is intentionally stacked on [#6582](https://github.com/CopilotKit/CopilotKit/pull/6582). Review and merge #6582 first.

- **#6582:** v1 SDK deprecated; use v2 instead — IDE warnings, exact replacements, curated related v2 concepts, docs, and enforcement, with no file renames.
- **This PR:** filesystem transition only.

The split keeps the deprecation behavior reviewable without hundreds of rename entries obscuring it.

## Folder transition

React Core becomes:

```text
packages/react-core/src/
├── index.ts
├── v1-deprecated-compatibility.ts
├── v1-deprecated/
│   ├── components/
│   ├── hooks/
│   ├── types/
│   └── index.ts
└── v2/
    ├── components/
    ├── hooks/
    ├── types/
    └── index.ts
```

Runtime follows the same boundary:

```text
packages/runtime/src/
├── index.ts
├── v1-deprecated-compatibility.ts
├── v1-deprecated/
│   ├── agents/
│   ├── graphql/
│   ├── lib/
│   ├── service-adapters/
│   ├── utils/
│   └── index.ts
└── v2/
```

Runtime's shared/current `agent/` directory remains outside `v1-deprecated/` because v2 uses `BuiltInAgent`.

## Public contract

- Package-root imports remain the deprecated compatibility API.
- `/v2` remains the preferred public API.
- No public `/v1` or `/v1-deprecated` entrypoint is added.
- Runtime build output is flattened back to its historical public layout.
- The Runtime post-build hook waits for both concurrent ESM and CJS builds before flattening, runs once per build cycle, resets for watch cycles, and fails loudly if the required v1 output is still missing after the full barrier.
- Moved legacy files preserve their code and import grouping; relative paths, build entrypoints, test paths, TypeDoc source links, generated notices, and manifests are updated only as required by the move.
- Tests move under `v1-deprecated/` only when they exercise deprecated compatibility code. Version-neutral and v2-only package contract tests remain at neutral or v2 paths.
- PR1's exact replacements, related-concept links, v2 docs homepage, and v2 reference labels are regenerated at their moved PR2 paths.
- Vue is not part of this transition.

## Validation

- Architecture/deprecation tests: 16/16
- Runtime dual-format layout barrier tests: 2/2
- Public API manifest contract: 1/1
- Inventory remains 245/245 exports across 9 entrypoints and 103 files
- Runtime and Web Inspector full tests pass uncached; React Core reaches 1517/1518 with only the exact-current-main raw-event feedback failure
- Relevant shell-docs tests: 59/59
- Explicit React Core, Runtime, Web Inspector, and shell-docs typechecks pass
- Required uncached package builds pass
- Shell-docs production build passes (223/223 pages)
- Generated warning inventory is current
- Scoped formatter: 232 files clean
- Scoped lint: 0 errors; no introduced warnings
- `git diff --check` passes
- Vue has no diff
- Four logical PR2 commits remain after rebasing onto the current PR1 head

## Verified current-main exceptions

The full shell-docs unit suite has two current-main failures: the pre-existing Channels architecture-image assertion mismatch, plus `docs-render.test.ts` expecting four Intelligence pages while current main generates five. The tests and their source inputs are byte-identical to fetched `origin/main`; neither PR touches them. Relevant docs tests and the shell-docs production build pass.

The full React Core suite reaches 1517/1518. Its sole failure, `CopilotChat.raw-event-feedback.test.tsx`, reproduces identically in a fresh detached worktree at exact `origin/main`; the test blob is identical and this stack does not touch that v2 chat test or implementation.


MikeRyanDev (changes requested):
Blocking: v2 still depends on code this PR classifies as deprecated.

- `packages/react-core/src/v2/index.ts` exports the public v2 `CopilotKit` and `CopilotKitProps` from `../v1-deprecated/...`.
- Runtime v2 core and SSE code import the active logger and telemetry disclosure from `v1-deprecated`.

This conflicts with the new repository rule that current code stays under `src/v2/` and defeats the source boundary: deleting or changing the deprecated tree would break v2. Keep code used by both APIs in a neutral or current location, point the v1 compatibility layer at it, and add a test or lint guard that production `src/v2/**` cannot import `v1-deprecated`.

The focused React Core and Runtime test, typecheck, and build targets pass, but those checks do not enforce this dependency direction.

MikeRyanDev (approved):
Correction after comparing the base and head: v2 already imported or re-exported these same implementation modules before this PR. This change relocates them and updates relative paths; it does not add a new runtime dependency. Focused React Core and Runtime tests, typechecks, and builds pass.

PR #2473, opened 2025-09-21
fix(examples): Handful of fixes for coagents-starter

timcharper:
## What does this PR do?

- Bump copilotkit to 1.10.4 to include fix from commit a7bb2f077cb0ce6ee5671cc2253f9878ed062ed1
- Add descriptions to tools (they're required, at least when using Ollama)
- Thread actions through convertActionsToDynamicStructuredTools - the functions are passed to the chat API incorrectly without it, and the model won't call them.

## Checklist

- [x] I have read the [Contribution Guide](https://github.com/copilotkit/copilotkit/blob/master/CONTRIBUTING.md)
- [x] If the PR changes or adds functionality, I have updated the relevant documentation

Issue #2528, opened 2025-09-28
🐛 Bug: Context (as in copilot readables) not actually passed to agent


timcharper:
### ♻️ Reproduction Steps

# 1) Define a `useCopilotReadable`

Per the docs https://docs.copilotkit.ai/langgraph/agent-app-context

I'd defined two readables, "lol" and "userId", at a top-level.

# 2) See the request to the next-js copilot route contain the context

(I actually needed to bump to `1.10.5-next.10` to get fix https://github.com/CopilotKit/CopilotKit/pull/2426 - previous to this, copilot readable state is not passed at all.

Here's the payload showing the copilot readables are passed to the nextjs route POST /api/copilotkit 

```json
{
  "operationName": "generateCopilotResponse",
[...]

PR #6592, opened 2026-08-20
fix(sdk-python): LangGraphAGUIAgent 500s on every request with ag-ui-langgraph 0.0.43

mxmzb:
## The bug

A fresh install of the [documented LangGraph + FastAPI quickstart](https://docs.copilotkit.ai/integrations/langgraph/quickstart) **500s on the first message**.

`sdk-python/pyproject.toml:34` depends on `ag-ui-langgraph[fastapi]>=0.0.42` with no upper bound. `0.0.43` shipped 2026-08-16 and began forwarding three of its own flags through `LangGraphAgent.clone()`. `LangGraphAGUIAgent` restated a closed keyword-only signature, so it rejects them:

```
TypeError: LangGraphAGUIAgent must override clone() or ensure its __init__ accepts
  (name, graph, description, config) as keyword arguments:
  LangGraphAGUIAgent.__init__() got an unexpected keyword argument 'enable_legacy_on_interrupt_event'
```

`add_langgraph_fastapi_endpoint` clones per request (`endpoint.py:23`), so this is a 500 on **every** request rather than a startup error. Reproduced with published packages only — `copilotkit 0.1.95` + `ag-ui-langgraph 0.0.43`.

## Two changes

**1. `__init__` forwards `**kwargs` upstream** instead of restating the base signature.

Restating it is what broke: every flag the base class adds becomes a 500 here. The passthrough also closes a functional gap — `emit_raw_events=False`, the OSS-607 payload opt-out (Function Health's ~1.5 MB events), is currently **unreachable for every CopilotKit user** because this subclass silently refuses to accept it.

**2. The intercepted-tool-call test double is constructed for real** instead of via `object.__new__`.

`_bridge_agent` built an instance with no behavior flags set, so it raised `AttributeError: 'LangGraphAGUIAgent' object has no attribute 'emit_raw_events'` out of the base dispatch path the moment ag-ui-langgraph started reading one. That is 4 test failures, currently hidden by `poetry.lock:17` pinning 0.0.42 — they appear the moment anyone bumps that lock.

## Verification

Against **published** `ag-ui-langgraph`, so this stands on its own without an upstream release:

| Installed base | Before | After |
|---|---|---|
| **0.0.43** (what fresh installs get) | 4 failed, 226 passed | **235 passed**, 11 skipped |
| **0.0.42** (what `poetry.lock` pins) | 233 passed | **233 passed**, 13 skipped |

The 2 extra skips on 0.0.42 are deliberate: the flag-forwarding tests `skipif` when the installed base predates `emit_raw_events`, so the suite stays honest on both versions instead of pinning behavior the installed library does not have.

End to end, two sequential POSTs through `add_langgraph_fastapi_endpoint` on 0.0.43:

```
request 1: HTTP 200 | RUN_ERROR in stream: False | got reply: True
request 2: HTTP 200 | RUN_ERROR in stream: False | got reply: True
```

The same script against published `copilotkit 0.1.95` raises at `ag_ui_langgraph/endpoint.py:23`. Two requests, not one — the per-request `clone()` this depends on is what fixes the concurrent-state corruption in ag-ui-protocol/ag-ui#1277, so the second request matters.

## New tests

`sdk-python/tests/test_agui_agent_clone.py` — 5 tests, all red before the change:

| Test | Guards |
|---|---|
| `test_clone_succeeds` | the 500 itself |
| `test_clone_preserves_copilotkit_state_namespace` | the `copilotkit` schema key survives the per-request clone, exactly once |
| `test_clone_resets_per_request_state` | `active_run` / `_copilotkit_runtime_payload` do not leak between requests |
| `test_init_forwards_upstream_flags` | base flags are reachable through this subclass |
| `test_clone_carries_upstream_flags` | a non-default flag is not silently reset per request |

## Relationship to the upstream fix

ag-ui-protocol/ag-ui#2473 fixes the same break in `LangGraphAgent.clone()` for every subclass in the wild, and needs no changes here. These are complementary, not redundant:

- **Upstream** protects third-party subclasses we do not control, and is the durable fix.
- **This PR** unblocks CopilotKit users **now**, via a `copilotkit` release, without waiting for a `0.0.44`. It also keeps working if a future upstream version adds another flag.

## Notes

- `commitlint` was validated out-of-band (`npx @commitlint/cli`, exit 0); the commit-msg hook itself cannot run in a fresh worktree without `node_modules`, which is why the commit is `--no-verify`.
- No lockfile change. Pinning away from 0.0.43 was the alternative; forwarding the kwargs keeps us compatible with both versions instead of excluding one.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved LangGraph agent cloning to preserve configuration and runtime options.
  - Ensured cloned agents isolate request-specific state without affecting the source agent.
  - Preserved event emission settings when processing FastAPI requests.

- **Tests**
  - Added coverage for cloning behavior, state isolation, schema handling, and event dispatching.
  - Expanded compatibility testing across supported LangGraph versions, including Python 3.12.


contextablemark (changes requested):
Thanks for the thorough investigation—the diagnosis and reproduction look solid. I think the `**kwargs` passthrough is a sensible CopilotKit-side hardening: our subclass should forward options owned by `LangGraphAgent`, and doing so also makes options such as `emit_raw_events` usable.

That said, the durable fix belongs in `ag-ui-langgraph`’s cloning contract. `clone()` introduced new base-owned kwargs while reconstructing `type(self)`, which breaks any existing fixed-signature subclass—not only CopilotKit’s. `ag-ui-protocol/ag-ui#2473` appears to address that at the correct layer. We should not remove or bypass per-request cloning, since that would reopen the state-corruption problem from `ag-ui-protocol/ag-ui#1277`.

Could we narrow this PR to:

- Keep the `**kwargs` forwarding change.
- Keep the test-helper fix that constructs the agent normally.
- Keep a small set of behavioral tests: cloning succeeds, CopilotKit request state is isolated, and a non-default upstream option survives cloning.
- Retain the `ag-ui-langgraph>=0.0.42` runtime floor and test 0.0.42/0.0.43 through CI instead. Raising the package floor merely for test coverage also raises the effective LangGraph floor from `>=0.3.25` to `>=0.6.0`.
- Remove the unrelated Poetry lockfile regeneration and the tests coupled to exact upstream flag names, error messages, and traceback locations.

So: directionally yes to the small compatibility shim here, while the adapter fix remains the primary solution. Given that the adapter fix is not released yet, shipping both is reasonable—but it would be better to keep this PR surgical.

mxmzb:
Narrowed as you asked — pushed. The PR is now four files.

**Kept**
- The `**kwargs` forwarding in `LangGraphAGUIAgent.__init__`.
- The test-helper fix that constructs the agent normally instead of via `object.__new__`.
- Four behavioral tests: cloning succeeds, the copilotkit schema namespace survives the clone, per-request state is isolated in both directions (clone starts clean, template is undisturbed), and a non-default upstream option survives the clone.

**Dropped**
- The `>=0.0.43` floor bump and the poetry lockfile regeneration. Your point about the transitive cost is the decisive one — 0.0.43 raises the LangGraph floor from `>=0.3.25` to `>=0.6.0`, and paying that for test coverage alone is the wrong trade.
- The enumerated forwarded-flag list and the test asserting those names against the base signature.
- The base-`__init__` spy test.
- The unknown-kwarg test, which asserted on the `TypeError` message text and on the raising traceback frame.

**Instead of the floor bump**, the python-sdk CI matrix now runs every Python version against `ag-ui-langgraph` 0.0.42 and 0.0.44. It is pinned with `pip install` after the lock resolve rather than by editing `pyproject.toml`, so the declared floor is untouched and only the installed version varies per leg.

Verified locally on both:

```
0.0.44 → 240 passed, 11 skipped
0.0.42 → 239 passed, 12 skipped
```

The one extra skip on the floor is `test_clone_carries_a_non_default_upstream_option` going quiet on a version that predates `emit_raw_events`, which is the intended behaviour of its guard.

**Two notes.**

I kept a fourth test beyond the three you listed — `test_clone_preserves_copilotkit_state_namespace`. It asserts on state this subclass owns rather than on anything upstream, so it has neither the coupling nor the skip problem the others had. Happy to drop it if you would rather hold to exactly three.

On the layering: the upstream fix has since landed and released. It came in via ag-ui-protocol/ag-ui#2528 rather than #2473 — #2528 made `clone()` signature-aware, and I closed #2473 as superseded after verifying against current `main` that a fixed-signature subclass mirroring this one now clones cleanly. So the durable fix is in place and released as `ag-ui-langgraph` 0.0.44, and this PR is no longer the stopgap it was when you reviewed it. What it still buys is your first point: without the passthrough, base-owned options such as `emit_raw_events` — and now `subagent_visibility` — remain unreachable for CopilotKit users. Per-request cloning is untouched throughout, so nothing here goes near ag-ui#1277.

contextablemark (changes requested):
This is substantially narrowed, and the prior dependency-floor, lockfile, and brittle-test concerns are resolved.

One compatibility gap remains: the matrix skips `ag-ui-langgraph==0.0.43`, which is the exact still-supported version whose `clone()` behavior caused this PR. In 0.0.44, the signature-aware clone fix from `ag-ui-protocol/ag-ui#2528` omits default-valued flags that a closed subclass constructor cannot accept. That means `test_clone_succeeds` would pass on 0.0.44 even if this PR’s `**kwargs` change were removed; the `emit_raw_events=False` test still guards option reachability, but it does not reproduce the default-construction 500 named in this PR.

Please add a targeted 0.0.43 leg. It does not need to multiply the full Python matrix—a single representative Python version alongside the existing 0.0.42/0.0.44 coverage is enough. Also update the module docstring in `test_agui_agent_clone.py`, which currently says CI exercises 0.0.42 and 0.0.43 while the workflow currently installs 0.0.42 and 0.0.44.

Everything else from the previous review is addressed. With that regression leg and documentation correction in place, this looks ready to approve.

mxmzb:
Both points addressed in 836e7c786f.

**0.0.43 regression leg.** Added via `matrix.include` on a single Python (3.12), so it is one extra leg rather than a third full column — the flag forwarding it exercises is not version-specific. Installed in the leg the same way the existing legs do it, so the declared floor stays at 0.0.42 and the effective LangGraph floor stays at `>=0.3.25` for consumers.

Your read of why 0.0.44 alone is insufficient checks out. 0.0.43's `clone()` passes `enable_legacy_on_interrupt_event`, `emit_interrupt_outcome`, and `emit_raw_events` to `type(self)(...)` unconditionally; 0.0.44's is signature-aware and omits default-valued flags a subclass cannot accept. So on 0.0.44 the default-construction path survives a closed signature, and nothing in the matrix reproduced the 500 named in this PR.

I checked the leg actually guards rather than just running green. Against 0.0.43, with the `**kwargs` passthrough temporarily removed:

```
FAILED tests/test_agui_agent_clone.py::test_clone_succeeds - TypeError: LangG...
FAILED tests/test_agui_agent_clone.py::test_clone_preserves_copilotkit_state_namespace
FAILED tests/test_agui_agent_clone.py::test_clone_isolates_per_request_state_from_the_source
FAILED tests/test_agui_agent_clone.py::test_clone_carries_a_non_default_upstream_option
4 failed
```

All four pass with the passthrough restored.

**Docstring.** Corrected — it now names all three versions CI installs (floor 0.0.42, the unconditional-forwarding 0.0.43, newest 0.0.44). I also rewrote the matrix comment, which had the same drift: it described a two-version rationale that no longer matched the leg list.

Matrix is 11 legs now: 5 Pythons × {0.0.42, 0.0.44}, plus 3.12 × 0.0.43.

contextablemark (approved):
Re-reviewed at 836e7c7. The latest changes resolve the remaining concerns.

CI now has a dedicated Python 3.12 / `ag-ui-langgraph==0.0.43` leg covering the exact regression, while retaining full supported-Python coverage at 0.0.42 and 0.0.44 without raising the dependency floor. The test and workflow commentary now accurately describe that matrix. The 0.0.43 job passed the full suite (240 passed, 11 skipped), and the remaining checks are green.

The cumulative PR remains surgical: the `**kwargs` compatibility shim, focused behavioral clone coverage, and normal construction in the existing test helper. The adapter-level clone fix released in 0.0.44 and this subclass-forwarding fix address complementary parts of the contract. Approved.

PR #6594, opened 2026-08-20
feat(shell-docs): inline backend picker opens the current docs page in the chosen backend

ataibarkai:
## What this does

Unscoped (framework-agnostic) docs feature pages end with a "Choose your AI backend" block. It used to be a note that linked *away* — originally to the showcase demo explorer, which loses the reader's place entirely.

This PR turns it into an **inline backend picker**: a chip per supported backend that opens **the current page** scoped to that backend (e.g. from `/generative-ui/state-rendering`, the Google ADK chip goes to `/google-adk/generative-ui/state-rendering`). It reuses `backendPathForCurrentPath` — the exact path rewriting the sidebar's backend selector already uses — plus the same registry filtering/ordering as the landing page's backend grid. Chip clicks persist the choice via `setStoredFramework` and fire the existing `docs.framework_selected` PostHog event with `source: "integration_grid"`.

inline-backend-picker

Availability handling:
- Backends listed in the call site's `exclude` prop render grayed out ("Not available for X yet") — see Agno above.
- Anything not excluded but genuinely missing falls back to the framework route's existing graceful handling (`NotAvailableForFrameworkPage` / the "Not available for X yet — try Y" banner), so there are no dead ends.

## Bug found along the way: the MDX pipeline silently drops expression props

`exclude={["agno", "agent-spec"]}` never reached the component. Root cause: **next-mdx-remote v6 defaults `blockJS: true`**, whose remark plugin strips every JSX expression attribute (and `{expression}` node) at compile — only string-literal props survive. This affects all MDX rendered through `MDXRemote` in shell-docs (`<Image width={1000}>`, `<Tabs items={[...]}>` — masked because `DocsTabs` re-derives items from `<Tab value>` children — and the `selected={false}` unreliability already worked around in `page.tsx`).

Flipping `blockJS: false` globally is not safe today: a content scan found 871 expression attributes, 461 with free identifiers (arrow-function handlers, `icon={<Component />}`, `props.components`) that would evaluate and throw. So within this PR, `exclude` is authored as a comma-separated **string** (`exclude="agno, agent-spec"`), the component accepts both string and array, and the three MDX call sites are updated with a warning comment. The systemic audit is spun off as a separate task.

## Where the link used to land vs the page context now preserved

**Before** — `showcase.copilotkit.ai/integrations`, the showcase demo explorer:

before-showcase-integrations

**Intermediate state (first commit)** — linked to the docs' "Build with any agent backend" list (`/#backends`), correct list but loses the current page:

after-docs-backends

**Final state (second commit)** — the picker is inline and keeps the reader on the same page in their chosen backend.

## Verification

- `tsc --noEmit` and `oxlint` pass in `showcase/shell-docs`.
- Verified in the browser on the dev server: chips render with correct per-backend hrefs for the current page; CopilotKit (Built-in Agent) is marked current; Agno renders grayed; clicking Google ADK lands on `/google-adk/generative-ui/state-rendering` with the sidebar backend switched; `/claude-sdk-python/generative-ui/state-rendering` (no framework-specific page) renders the root content scoped to Claude SDK — no dead end.
- Vitest: 420/421 pass; the 1 failure (`channels-docs.test.ts`) is pre-existing on main (channels architecture diagram references the light PNG twice) and untouched here — spun off as a separate task.

### DIFF
--- a/showcase/shell-docs/src/components/integration-grid.tsx
+++ b/showcase/shell-docs/src/components/integration-grid.tsx
@@ -1,13 +1,15 @@
 "use client";
 
 import React from "react";
+import Link from "next/link";
 import { useFramework } from "./framework-provider";
-import { getRuntimeConfig } from "@/lib/runtime-config.client";
 
 export function IntegrationGrid({
-  path,
   description,
 }: {
+  // `path` and `exclude` are authoring metadata passed by many MDX call
+  // sites (which framework pages exist for this feature). They are not
+  // rendered — the note links to the full backend list instead.
   path?: string;
   exclude?: string[];
   description?: string;
@@ -17,13 +19,6 @@ export function IntegrationGrid({
   // On a framework-scoped route the user already chose a backend — hide.
   if (framework) return null;
 
-  // Shell host is now read at runtime from window.__SHOWCASE_CONFIG__
-  // (set by the root layout) so the rendered <a href> reflects the
-  // current deploy's NEXT_PUBLIC_SHELL_URL without a rebuild. Pulled
-  // after the early-return so we never call into the client reader on
-  // renders that produce no DOM.
-  const shellHost = getRuntimeConfig().shellUrl;
-
   return (
     <>
       <h2>Choose your AI backend</h2>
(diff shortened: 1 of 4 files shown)


ranst91 (dismissed):
🚢 Good catch, I like it

showcase/shell-docs/src/components/integration-grid.tsx
```diff
@@ -1,49 +1,130 @@
...
-import { useFramework } from "./framework-provider";
-import { getRuntimeConfig } from "@/lib/runtime-config.client";
+import Link from "next/link";
+import { usePathname } from "next/navigation";
+import { usePostHog } from "posthog-js/react";
+import { DEFAULT_FRAMEWORK, useFramework } from "./framework-provider";
+import { FrameworkLogo } from "./icons/framework-icons";
+import { compareByDisplayOrder } from "@/lib/framework-order";
+import { backendPathForCurrentPath } from "@/lib/frontend-options";
+import { getDocsMode, getIntegrations } from "@/lib/registry";
```
samjulien, line 22: IntegrationGrid is a client component, so importing @/lib/registry pulls registry.json (~645 KB uncompressed in this checkout) into the client graph for every page that renders this picker. The picker only needs slug, name, logo, and visibility, but the registry also contains demos, features, descriptions, URLs, and other metadata. Please resolve and sort the integrations in a server wrapper, then pass a compact options DTO into the interactive client component, matching the existing SidebarFrameworkSelector pattern.

samjulien (changes requested):
Requesting changes because the new client-side backend picker imports the full registry dataset. This is a material and avoidable docs performance regression; please use the existing server-to-client options pattern before merging.

PR #6596, opened 2026-08-20
feat(showcase): add travel workflow example

tylerslaton:
## Summary

Adds a minimal travel workflow showcase that streams a Python LangChain agent’s state into a CopilotKit `useAgent` UI.

## Why

Give developers a small, readable reference for observing a long-running agent workflow and rendering agent-selected attractions on a map.

## How

- Builds the Python backend with `langchain.create_agent`, `ChatOpenAI`, and AG-UI state snapshots
- Uses `useAgent` as the frontend source of truth for live workflow state
- Renders streamed attractions with Leaflet and focused shadcn/ui primitives
- Enables the CopilotKit inspector for learning and debugging

## Verification

- `pnpm nx run travel-workflow:check-types --skip-nx-cache`
- `pnpm nx run travel-workflow:build --skip-nx-cache`
- `pnpm nx run travel-workflow:check-agent --skip-nx-cache`
- Real-agent browser smoke test with streamed attractions and map pins


BenTaylorDev (changes requested):
Reviewed against current `main`. The example is in good shape — I built and ran it on today's tree. One commit should not land, and it is also the only merge blocker.

## Blocking: drop `50245e56de`

`test(react-core): correct feedback memoization coverage` is unrelated to a travel showcase, and `main` already made the same correction: this branch varies `defaultThrottleMs` on `CopilotKitProvider`, `main` varies `className` on `CopilotChatMessageView`. Same diagnosis (the `labels` change was never unrelated), and `main`'s landed. Dropping it clears the only conflict — `git merge-tree --write-tree origin/main 40f4fdc25e` comes back clean.

## Verified on current `main`

- `travel-workflow:check-types` and `:build` both pass (Next 16.3.3, catch-all `ƒ /api/copilotkit/[[...slug]]`)
- `uv sync --project agent --frozen` passes; graph compiles
- live AG-UI SSE at the pinned versions: `200 text/event-stream`, `RUN_STARTED` → `STEP_STARTED CopilotKitMiddleware.before_agent` → `on_chat_model_start`, dying only at OpenAI auth
- all 14 emitted event types, through the production `EventEncoder`, validate against the pinned `@ag-ui/core@0.0.59`

The `next` specifier mismatch between `package.json` and the lockfile is a false alarm (pnpm records the overridden specifier; `main`'s security floor resolves it to 16.3.3). Regenerate the lockfile on rebase anyway.

## Worth fixing here

- **Pins drifted from `main`'s validated matrix.** `agent/pyproject.toml:7-8` pins `copilotkit==0.1.94` (in-repo SDK 0.1.96) and `ag-ui-langgraph==0.0.41` (`main` validates 0.0.44), and declares no `ag-ui-protocol` pin so the lock floats to 0.1.20 where siblings pin 0.1.22 for a documented reason. Works today — divergence, not breakage.
- **`travel-workflow:check-agent` never runs in CI** — no job references it anywhere under `.github/`, and the target is `python -m compileall`, which would not catch an import error. A live keyed run is the only gate this example gets.
- **This registers a permanent project in two repo-wide sweeps.** The `pnpm-workspace.yaml` entry puts it in `nx run-many -t check-types` (`static_quality.yml:336`) and `-t build` (`static_compat.yml:51`), both triggered by `examples/**` and `packages/**`. It becomes the second `examples/showcases/*` app that is a workspace member; the rest sit outside deliberately. Fine if intended, but worth choosing on purpose. Not in the Docker fan-out — those are `examples/integrations/**` and `showcase/**` scoped. Side effect: `auto_merge_showcases.yml:53-55` requires every file under `examples/showcases/`, so a workspace-member showcase can never satisfy it; the `auto-merge pass` check here is a no-op skip, not an approval.

Nits: `route.ts:23-24` exports only `GET`/`POST` where `main`'s sibling also exports `PATCH`/`DELETE`; the `pnpm-workspace.yaml` entry sits under the `reskinnable-demo` NOTE block so it reads as if that caveat applies — it belongs next to `generative-ui-playground` on line 13; `README.md` never lists `uv` as a prerequisite though `dev:agent` shells out to it.

No secrets, no retired-tree debris, no changesets, current v2 API surface throughout.

Before merging: one real-key end-to-end run on current `main` — ask for attractions, confirm pins land on the map. Every link checks out individually, but the full loop (tool call → `copilotkit_emit_state` → `STATE_SNAPSHOT` → `agent.state` → `LocationMap`) is what CI will never cover.

PR #6609, opened 2026-08-20
fix(showcase): tighten the offsite-expenses beat

davidmckayv:
The "Sort out my offsite expenses" beat had three problems on stage: the harness console was a black slab in a light-mode transcript, the tool activity grew a stack of finished steps that pushed the report card off the screen, and the run took a full minute.

## Console (banking skin)

- Every colour is now a semantic token, so the pane follows the app into dark or light. It was `bg-ink` with `text-white/45`-style overlays, which only ever looked right in one mode.
- **Collapsed by default.** The status strip still carries the live state ("3 agents working"), so the run reads as alive while closed.
- Still the **full** log when open. It is the detail view, and windowing it as well left two lines and nowhere to read the rest.

## Tool activity (shell, affects all skins)

- Rolls to the last two lines; older ones are **removed**, not collapsed.
- Registration is a **layout** effect, and that is load-bearing. A new line renders before it is registered, and registering is what evicts the oldest, so with a passive effect the browser painted the in-between state: three rows for one frame on every tool call, and again when the end-of-run `MESSAGES_SNAPSHOT` remounts every line at once. Measured per animation frame over a full run — 12,734 frames, never more than two, one `1->2` transition.
- A shared registry rather than something simpler because CopilotKit renders one component per tool call and owns the container. CSS `:nth-last-child` needs siblings and the lines had one parent each; mount-order counters drift across the snapshot remount.

Reviewers should know this one is **not** scoped to banking. Most beats emit one or two tool calls so it is a no-op for them, but a beat emitting three or four would now show only its last two. `VISIBLE_TOOL_ACTIVITY` is a single constant if we want it higher.

## Agent: 1m 0s -> ~45s

- **Research is gated on the offsite dates.** A charge dated outside the window is settled by its date whatever the merchant turns out to be, so half the researcher dispatches were buying nothing. Travel on the adjacent days is still kept in scope.
- **Filings go out in one command** instead of one `curl` per row. The researchers already ran concurrently, so halving them bought almost nothing — the serial per-row round-trips through the model were most of the wall clock.
- Fetch and verify are one command; there is nothing to decide between the halves.
- Analyst reasoning effort defaults to `medium`, overridable with `BANKING_EXPENSE_EFFORT`.

## Filing is idempotent

The batching made this necessary rather than optional: the script got run twice and every charge was filed twice, so the report card claimed six filings while the ledger held twelve. A duplicate here is a reimbursement claimed twice.

The script now writes `filed.json` and exits early if it already exists. That marker is cleared once per run, because the workspace is a fixed directory shared by every run and a stale marker would convince the next demo it had already filed and post nothing at all — the same bug wearing the opposite mask, and a quieter one, since a run that files nothing still writes a confident report.

Deduping server-side on merchant+amount would have been wrong: **Hotel Verrano legitimately appears twice at the identical $318.55** for the two nights of the offsite.

## Verification

Run against a live Intelligence stack, not unit tests alone:

- Six rows filed, all `status=pending`, stable across 60s of polling.
- Present in the Pending Approval queue with their notes and approve/decline actions.
- Report card: 14 rows read, 6 merchants researched, $2,377.15 reimbursable.
- Console checked in both themes (panel `#f9f8fc` light / `#1f1c2c` dark, text inverting with it).

`lint`, `typecheck`, `test:unit` (2460 passed) and `build` all green, re-run after rebasing onto current main.

Does this change make anything in `.claude/skills/reskin/` wrong, incomplete, or misleading? No — nothing there documents the harness console, the tool-activity renderer, or the analyst prompt.

### DIFF
--- a/examples/showcases/reskinnable-demo/agent/agent.py
+++ b/examples/showcases/reskinnable-demo/agent/agent.py
@@ -3,9 +3,9 @@
     banking                       gpt-5.4, temp 0
       │  banking's own prompt; the browser's frontend tools and the
       │  Intelligence memory tools; `render_report` for the canvas
-      └─ expense-analyst          gpt-5.6-sol, reasoning_effort=high
+      └─ expense-analyst          gpt-5.6-sol, reasoning_effort=medium
            │  sandboxed shell (LocalShellBackend), `submit_expense_report`
-           └─ merchant-researcher gpt-5.4, streaming OFF, one per merchant
+           └─ merchant-researcher gpt-5.4, own lane, one per merchant
                 `search_merchant` (Tavily)
 
 The nesting is the design, not an accident of growth. Each level exists because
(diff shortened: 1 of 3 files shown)


mxmzb (approved):

davidmckayv:
Correction to the run-time claim in this PR description, from re-verifying on merged main.

The description says the beat went from `1m 0s` to `~45s`. That was based on a single run and **does not replicate**. Measured run times on this code, from the report card's own clock:

| run | elapsed |
|---|---|
| 1 | 0m 45s |
| 2 | 0m 50s |
| 3 | 1m 1s |
| 4 | 0m 57s |

Against a pre-change baseline of `1m 0s` (also n=1). Mean of the post-change runs is ~53s, and two of four landed at or above the baseline. A cold-vs-warm dev server does not explain it either — run 3 (1m 1s) was cold and run 4 (0m 57s) was warm immediately after.

So: **treat the wall-clock improvement as unproven.** The structural changes are still real and worth keeping — the date gate genuinely halves the researcher dispatches, and the batched filing genuinely collapses six serial model round-trips into one — but reasoning-model variance dominates elapsed time on a statement this size, so neither shows up dependably. Anyone tuning this for demo pacing should measure several runs rather than trusting the number in the description.

Nothing else in the PR changes. Re-verified on merged main: six rows filed, all `status=pending`, no duplicates, console collapsed and theme-following, activity capped at two lines.

PR #5157, opened 2026-06-02
feat(react-core): add showIntelligenceIndicator prop to opt out of auto-mounted pill

RykerFeng:
## Summary

Fixes #5126.

The built-in `IntelligenceIndicator` pill ("Using CopilotKit Intelligence") auto-mounts on assistant messages whenever the runtime is in intelligence mode, with no way to turn it off. Apps that render their own intelligence/agent status UI end up showing the default pill alongside their custom one.

This adds a `showIntelligenceIndicator` prop on `CopilotKitProvider` so consumers can opt out.

## Changes

- **`CopilotKitProvider`** — new `showIntelligenceIndicator?: boolean` prop (default `true`), threaded through context.
- **`context.ts`** — added optional `showIntelligenceIndicator` to `CopilotKitContextValue` (optional so RN provider and existing mocks stay compatible; consumers treat `undefined` as `true`).
- **`CopilotChatMessageView`** — auto-mount gate now also checks `showIntelligenceIndicator`; reads default `true`.
- **Test** — added an e2e case asserting the pill does not render when `showIntelligenceIndicator={false}`, even with intelligence configured.

## Backward compatibility

Default is `true`, so existing behavior is unchanged. Only an explicit `false` suppresses the pill.

## Testing

- `nx run @copilotkit/react-core:test` — 1172/1172 pass (incl. the new case)
- `nx run @copilotkit/react-core:build` — passes

PR #6612, opened 2026-08-20
feat(react-core): make intelligence indicator auto-mount configurable

### LINKED ISSUE #5126: 🚀 Feature Request: Expose IntelligenceIndicator auto-mount behavior as configurable option
---
name: 🚀 Feature Request
description: Have an idea for enhancements or features? Let us know!
title: "🚀 Feature Request: "
labels: ["feature request"]
assignees: []
---

## Problem or Motivation

The `IntelligenceIndicator` component auto-mounts when a remote agent requires user verification (via `requiresUser` for OAuth-gated MCP HTTP servers). This automatic behavior cannot be disabled, which limits flexibility for developers who want to control mount behavior programmatically.

## 🎤 Tell us your idea

Add an optional `autoMountIntelligenceIndicator?: boolean` prop to the `CopilotProvider` (defaulting to `true` for backward compatibility). When set to `false`, the indicator will not auto-mount — agents can still be mounted via explicit `registerProxiedAgent` calls. This gives developers fine-grained control without breaking existing setups.

Use case: Chat applications that implement their own custom intelligence/agent status indicators and do not want the auto-mounted indicator to appear alongside their custom UI.

---
- [ ] I'd love to work on it!
- [ ] I'm open to collaborating but need guidance.
- [ ] No, I'm just sharing the idea.

jaredrainsha:
## What does this PR do?

- Adds `showIntelligenceIndicator` to `CopilotKitProvider`, defaulting to `true`.
- Gates the existing per-turn Intelligence indicator auto-mount without changing its state machine.
- Adds an end-to-end regression test for the provider opt-out.
- Documents the new public provider property.

The default behavior remains backward compatible. Setting the property to
`false` only suppresses the built-in indicator, allowing applications to
render their own Intelligence status UI.

## Related PRs and Issues

- Fixes #5126
- Reimplements the accepted approach from the stale #5157 against the current
  per-turn indicator mounting architecture.

## Testing

- Intelligence indicator e2e tests: 14 passed
- `@copilotkit/react-core:check-types`: passed
- `@copilotkit/react-core:build`: passed
- oxlint: 0 errors
- Full react-core suite: 1518 passed, 1 pre-existing failure in
  `CopilotChat.raw-event-feedback.test.tsx`. The same failure reproduces with
  this PR's production changes reverted.

## Checklist

- [x] Read the contribution guidelines
- [x] Added regression coverage
- [x] Updated the API reference documentation
- [x] Allow edits by maintainers


BenTaylorDev (approved):
@jaredrainsha — thanks for picking this up so quickly after I closed #5157, and sorry for the delay in getting to it. Approving.

I replayed your commit onto current `main` (`8991dbd`) locally and verified it there:

- `@copilotkit/react-core:test` — 47 files, 0 failures, cache skipped. The `CopilotChat.raw-event-feedback.test.tsx` failure your PR body mentions no longer reproduces on current `main`.
- `check-types` for `@copilotkit/react-core` and `@copilotkit/react-native` — pass. Keeping the context field optional is what keeps react-native's own `CopilotKitContextValue` compiling; good call.
- `oxfmt --check` and `oxlint` on the changed files — clean, and no new warnings against the `main` baseline.
- Mutation check — reverting the gate to `intelligenceTurnAnchors.get(message.id)` fails your new test and only your new test, so it is a real regression probe. Anchoring the absence assertion on `m2` reaching the DOM is better than a timed wait, because it does not depend on the 100 ms grace constant.
- The prop reaches `<CopilotKit>` exactly as your docs entry claims: `CopilotKitProps extends Omit<CopilotKitProviderProps, "children" | "onError">`, and the v1 wrapper spreads `{...v2Props}`.

One thing left before this can merge: the branch is behind `main` and conflicts in `CopilotKitProvider.tsx`. Please rebase. It is pure adjacency with the new `enableInspector` prop — the declaration goes after `enableInspector`, the destructured default after `agentId`, and nothing else moves. My local replay produced exactly that and stayed green, so it should be mechanical.

Non-blocking, for a future pass: #6625 arrived as a duplicate of this and gated the `if` instead of nulling `intelligenceTurnId`. Equivalent, since `intelligenceTurnId` has no other use, but it reads a little more directly.

PR #6616, opened 2026-08-20
fix(runtime): replace retired Anthropic model IDs

tylerslaton:
## Problem

Retired Anthropic model identifiers remained in runtime types, examples, documentation, and showcase configuration. Several Anthropic defaults also still selected Sonnet 4.6.

## Why

Retired defaults and copy-paste examples fail at Anthropic, while short-lived defaults force repeated repository-wide remediation. Explicit user-selected model identifiers must remain forward-compatible and must not be rejected or silently replaced by CopilotKit.

## Fix

- Replace retired repository references and move Sonnet 4.6 defaults to Opus 4.8
- Keep arbitrary explicit model identifiers open and pass them through unchanged, with regressions covering a retired 3.7 identifier
- Limit runtime package behavior changes to default model selection and honoring the explicit per-request model already accepted by the adapter
- Use Opus 4.8-compatible adaptive thinking only where the changed showcase default requires it
- Add an Nx-backed repository guard and CI check preventing retired Anthropic identifiers from returning

### DIFF
--- a/.github/workflows/static_retired-anthropic-models.yml
+++ b/.github/workflows/static_retired-anthropic-models.yml
@@ -1,0 +1,40 @@
+name: static / retired Anthropic models
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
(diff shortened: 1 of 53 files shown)


contextablemark (approved):
Approved

PR #6623, opened 2026-08-21
fix(react,vue): make popup positioning wrapper transparent

### LINKED ISSUE #6472: fix(react, vue): popup wrapper background masks rounded corners

deepshekhardas:
## Summary
- The fixed outer popup wrapper (`[data-copilotkit]`) now sets `cpk:bg-transparent` in both React and Vue v2 `CopilotPopupView`
- Previously the default theme background on `[data-copilotkit]` showed square white corners around the rounded-2xl inner dialog
- Fixes #6472

## Test plan
- Render an open v2 popup on a non-white page background at desktop width; corners are transparent and host page shows through

### DIFF
--- a/packages/channels-telegram/src/__tests__/telegram-html.test.ts
+++ b/packages/channels-telegram/src/__tests__/telegram-html.test.ts
@@ -68,3 +68,22 @@ describe("stripHtml", () => {
     expect(stripHtml("&amp;lt;")).toBe("&lt;");
   });
 });
+
+describe("telegramHtml: fenced code language tags (#6602)", () => {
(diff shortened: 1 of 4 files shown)


BenTaylorDev:
Closing as a duplicate of #6473, which was filed for #6472 first.

Same one-class fix (`cpk:bg-transparent` on the positioning wrapper), and #6473 also adds the React and Vue regression tests the issue asked for. Thanks for the ping on this one — the mechanism is right, the wrapper rule is in `@layer base` so a utility class does win the cascade.

PR #6626, opened 2026-08-21
fix(react): preserve caret across textarea measurement in CopilotChatInput

### LINKED ISSUE #6167: 🐛 Bug: CopilotChatInput - cursor jumps to the end of the textbox upon edit in expanded mode

deepshekhardas:
Fixes #6167

`ensureMeasurements()` in `CopilotChatInput` temporarily sets `textarea.value = ""` (then restores it) to measure the single-line height. Assigning `.value` resets the caret to the end of the textarea. In expanded mode the layout evaluation runs on every keystroke, so placing the cursor mid-text and typing made it jump to the end.

The fix saves `selectionStart`/`selectionEnd`/`selectionDirection` before the measurement swap and restores them afterwards (only when the textarea is focused).

### DIFF
--- a/packages/channels-telegram/src/__tests__/telegram-html.test.ts
+++ b/packages/channels-telegram/src/__tests__/telegram-html.test.ts
@@ -68,3 +68,22 @@ describe("stripHtml", () => {
     expect(stripHtml("&amp;lt;")).toBe("&lt;");
   });
 });
+
+describe("telegramHtml: fenced code language tags (#6602)", () => {
+  it("emits Telegram language- form for tagged fences", () => {
+    expect(telegramHtml("```js\nconst a = 1;\n```")).toBe(
+      '<pre><code class="language-js">const a = 1;</code></pre>',
+    );
+  });
+  it("handles python and escapes code contents", () => {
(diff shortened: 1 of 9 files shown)


BenTaylorDev:
Closing — two separate reasons, and #6167 stays open.

The branch contains four commits: the Telegram fence fix, the popup wrapper fix, the langgraph docs change, and the caret change. Nine files under a caret-fix subject.

On the fix itself: `textarea.value = ""` in `ensureMeasurements` is indeed the only imperative value write in the file, but `adjustTextareaHeight` and `evaluateLayout` both call it only when `measurementsRef.current.singleLineHeight === 0` (`CopilotChatInput.tsx:701`, `:788`), and nothing resets that afterwards. In a laid-out browser it runs once per mount, which doesn't account for the reporter's every-keystroke jump. #6179 proposes a different mechanism for the same issue. The root cause needs pinning down in a real browser before either lands.

PR #6627, opened 2026-08-21
fix(runtime): align pino declaration with tested version

### LINKED ISSUE #6423: pnpm.overrides contradict package declarations: runtime tests pino 10 but ships pino 9; bare next override forces a two-major jump
Two `pnpm.overrides` entries in the root `package.json` silently contradict what packages declare. One affects a **published** package, so CI and npm consumers run different major versions of a runtime dependency.

## 1. `pino` — published package tests against 10, ships 9

```jsonc
// package.json  (pnpm.overrides)
"pino@<=10.1.1": "10.1.1"
```

```
packages/runtime/package.json   name: @copilotkit/runtime  version: 1.66.2  private: false
                                dependencies.pino: "^9.2.0"

pnpm-lock.yaml:3476             packages/runtime → pino
                                  specifier: 10.1.1
                                  version:   10.1.1

npm registry                    @copilotkit/runtime@1.66.2 declares pino: ^9.2.0
                                pino latest: 10.3.1
```

So in this repo `packages/runtime` resolves **pino 10.1.1**, while every npm consumer installing `@copilotkit/runtime` gets **pino 9.x** — because the override is local and the published manifest still says `^9.2.0`.

`packages/runtime` genuinely uses pino: `src/lib/logger.ts`, `src/lib/runtime/utils.ts`, `src/lib/runtime/retry-utils.ts`.

This is not a resolver error — pnpm overrides are authoritative by design. It is a dev/prod skew plus a declared-constraint that no longer describes what gets tested. A pino 9↔10 behavioural difference would reproduce only for consumers, never in CI.

**Provenance:** `b2949eade3` (2026-01-20, `aikido-autofix[bot]`, "Chore: Patch a bunch of dependencies"), via PR #3043 — `reviewDecision: REVIEW_REQUIRED`, **0 reviews**, human-merged. `git log -S 'pino@<=10.1.1' -- package.json` returns only that commit.

**Fix options:**
- bump `packages/runtime` to `pino: ^10.1.1` so the declaration matches what is tested, and drop the override; or
- drop the override and let `^9.2.0` resolve, so CI tests what consumers get.

Either is fine — what should not persist is the two disagreeing. The first is probably right if pino 10 is intended.

For contrast, aikido's other six overrides in the same commit were checked against the GitHub Advisory API and are correct or already superseded (`send@0.19.0`, `serve-static@1.16.0`, `prismjs@1.30.0` all sit exactly on their patched versions). `pino` is the outlier because it is the only one that contradicts a first-party package's own declaration.

## 2. Bare `next` override forces a two-major jump

```jsonc
"next@<=15.4.11": "15.4.11",
"next@>=15.5.0 <15.5.15": "15.5.15",
"next": "^16.0.10",              // ← unscoped, matches everything
"next@>=16.0.0 <16.2.3": "16.2.3"
```

```
examples/v1/state-machine/package.json   declares next: 14.2.35   (private: true)
pnpm-lock.yaml                           specifier: ^16.0.10  →  version: 16.1.3
```

The unscoped `"next": "^16.0.10"` overrides every selective entry above it, so a package pinned to `14.2.35` resolves `16.1.3`. Lower severity than the pino case — `examples/v1/state-machine` is `private: true`, so there is no consumer impact, only an example being built against a different major than it declares.

Also inconsistent: `next@>=16.0.0 <16.2.3` should push `16.1.3` up to `16.2.3`, yet the lockfile records `16.1.3`. Either the bare override wins over the ranged one, or the lock is stale relative to the override block. Worth resolving while the block is open — it is currently hard to reason about which of four entries applies.

Authored by a human (`cb28230b968`, 2026-04-12), not a bot.

## Why file this

Both were found while auditing automation commits for unreviewed dependency changes, prompted by a separate stale-pin incident (#6363). Neither is urgent. The pino one is worth doing because a published package declaring a constraint the repo does not honour is the kind of thing that produces an unreproducible consumer bug report.

## Acceptance criteria

- [ ] `packages/runtime`'s declared `pino` constraint matches the version CI resolves
- [ ] the `next` override block is reduced to entries that do not contradict each other, and the lockfile reflects it
- [ ] if an override must contradict a package's declaration, the reason is recorded inline — an undocumented pin is what caused #6363

deepshekhardas:
Fixes #6423 (the pino part)

`pnpm.overrides` pinned `pino@<=10.1.1` to 10.1.1, so CI tested `@copilotkit/runtime` against pino 10 while the published manifest declared `^9.2.0` — npm consumers ran a different major than what was tested.

This bumps `packages/runtime` to `pino: ^10.1.1` and drops the override, so the declaration matches what CI exercises. (`pino-pretty` has no peer dependency on pino — it receives the stream directly — so no change needed there.)

The bare `next` override from the same issue is left untouched here; it spans many example apps and deserves its own reviewed change.

### DIFF
--- a/package.json
+++ b/package.json
@@ -115,7 +115,6 @@
       "path-to-regexp@<=0.1.12": "0.1.13",
       "serve-static@<=1.16.0": "1.16.0",
       "prismjs@<=1.30.0": "1.30.0",
-      "pino@<=10.1.1": "10.1.1",
       "@copilotkit/license-verifier": "~0.5.0",
       "next": "^16.0.10",
       "defu@<=6.1.4": ">=6.1.5",
--- a/packages/runtime/package.json
+++ b/packages/runtime/package.json
@@ -112,7 +112,7 @@
     "openai": "^4.85.1 || >=5.0.0",
     "partial-json": "^0.1.7",
     "phoenix": "^1.8.4",
-    "pino": "^9.2.0",
+    "pino": "^10.1.1",
     "pino-pretty": "^11.2.1",
     "reflect-metadata": "^0.2.2",
     "rxjs": "7.8.1",


BenTaylorDev:
Closing on the lockfile, not the diagnosis — #6423 stays open and the direction here is the one the issue prefers.

Changing `packages/runtime`'s `pino` spec and dropping the override changes the recorded specifier, but `pnpm-lock.yaml` isn't updated. CI installs with `--frozen-lockfile` in every workflow, so this fails at the install step before any test runs. Fork PRs don't get CI automatically, which is why it looks clean here.

PR #6628, opened 2026-08-21
feat(angular): add sendMessage to AgentStore

### LINKED ISSUE #6507: 🚀 Feature Request: Angular — add `sendMessage` to `AgentStore`
# Proposal: `sendMessage` on `AgentStore` (Angular)

**Package:** `@copilotkit/angular` 0.3.0

## Problem

There is no supported way to send a message programmatically. `AgentStore` exposes
read-only projections only:

```ts
declare class AgentStore {
  readonly agent: AbstractAgent;
  readonly isRunning: Signal<boolean>;
  readonly messages: Signal<Message[]>;
  readonly state: Signal<unknown>;
}
```

Anything beyond rendering the built-in chat — a "send this to the agent" button, an
uploaded image forwarded as a multimodal message, a page that seeds a conversation —
has to be assembled by hand. In our app that ends up as:

```ts
export async function sendMessage(
  copilotKit: CopilotKit,
  store: Signal<AgentStore>,
  input: string | UserMessage['content'],
  forwardProps?: Record<string, unknown>,
): Promise<void> {
  const agent = store().agent;
  agent.addMessage({ id: randomUUID(), role: 'user', content: input });
  await copilotKit.core.runAgent({ agent, forwardedProps: forwardProps });
}
```

To write those four lines an application developer has to know:

- that messages carry a client-generated id, and that `randomUUID` comes from `@ag-ui/client`
- the `UserMessage['content']` shape from `@ag-ui/core` (string vs. content parts) for
  anything multimodal
- that "sending" is really two steps — `agent.addMessage(...)` followed by a run — and that
  forgetting the second one silently does nothing
- that the run is started via `copilotKit.core.runAgent({ agent, forwardedProps })`, i.e.
  through `CopilotKit` and its `core`, not through the store you already hold

That is AG-UI SDK detail leaking into every call site. CopilotKit abstracts the transport,
the tool wiring, the rendering and the state projection — sending a message is the one
everyday operation where the abstraction stops and the SDK starts.

Every consumer therefore writes the same helper, and each one re-implements the id
generation, the add-then-run ordering, and the `forwardedProps` plumbing.

## Proposal

Add a `sendMessage` method to `AgentStore`:

```ts
interface SendMessageOptions {
  forwardedProps?: Record<string, unknown>;
}

declare class AgentStore {
  // ...
  sendMessage(
    content: string | UserMessage['content'],
    options?: SendMessageOptions,
  ): Promise<RunAgentResult>;
}
```

Call sites collapse to:

```ts
await store.sendMessage('Find flights from Paris to London');

await store.sendMessage(
  [
    { type: 'text', text: 'Here is my ticket.' },
    { type: 'image', source: { type: 'data', value: base64, mimeType } },
  ],
  { forwardedProps: { agentMode: this.mode() } },
);
```

Two points about the signature:

- **No `copilotKit` and no `store` parameter.** The store already holds the agent and can
  reach the runtime itself, so the two arguments that only exist to satisfy the
  free-function form disappear.
- **`forwardedProps` moves into an optional options object** instead of being a positional
  tail argument. This keeps the common call a one-liner and leaves room for future run
  options (abort signal, thread id, run-level metadata) without another breaking signature
  change.

## Keep the escape hatch

Please keep `AgentStore.agent` public. Exposing the AG-UI `AbstractAgent` is genuinely
valuable for the cases a convenience API should not try to cover — replacing the message
history, resetting `threadId`, `abortRun()`, custom `AbstractAgent` subclasses, direct
subscription to lifecycle events. The point of this issue is not to hide the SDK, it is
that the *normal* path should not require it.

## Related (out of scope here)

The same argument applies to a small number of neighbouring operations that today force
the same helper layer: adding a `developer` message, aborting a run, and clearing a
thread. If `sendMessage` lands, those are natural follow-ups with the same shape.

deepshekhardas:
Fixes #6507

Adds `sendMessage(input, options?)` to the Angular `AgentStore`:

```ts
await store.sendMessage("hello", { forwardedProps: { ... } });
```

- Accepts `string | UserMessage["content"]` (multimodal-friendly)
- Mints the message id, adds the user message to the agent, and starts the run through `CopilotKit` core — so consumers no longer need to know AG-UI details or the add-then-run ordering
- Optional `forwardedProps` passthrough
- Stores not created by `injectAgentStore()` throw a clear error, mirroring the existing `interruptController` pattern
- Includes a vitest case covering message addition and run invocation

### DIFF
--- a/package.json
+++ b/package.json
@@ -115,7 +115,6 @@
       "path-to-regexp@<=0.1.12": "0.1.13",
       "serve-static@<=1.16.0": "1.16.0",
       "prismjs@<=1.30.0": "1.30.0",
-      "pino@<=10.1.1": "10.1.1",
       "@copilotkit/license-verifier": "~0.5.0",
       "next": "^16.0.10",
       "defu@<=6.1.4": ">=6.1.5",
--- a/packages/angular/src/lib/agent.spec.ts
+++ b/packages/angular/src/lib/agent.spec.ts
@@ -744,6 +744,38 @@ describe("AgentStore.interruptController", () => {
     ]);
   });
 
+  it("sendMessage adds a user message and starts a run", async () => {
+    const agent = new MockAgent("agent-1");
+    copilotKitStub.setAgents({ "agent-1": agent });
+    const agentId = signal<string | undefined>("agent-1");
+
+    @Component({
+      standalone: true,
+      template: "",
+    })
+    class SendHost {
+      readonly store = injectAgentStore(agentId);
+    }
+
+    const fixture = TestBed.createComponent(SendHost);
+    fixture.detectChanges();
+    const store = fixture.componentInstance.store();
+
+    await store.sendMessage("hello", {
+      forwardedProps: { source: "test" },
+    });
+
+    expect(agent.messages).toHaveLength(1);
+    expect(agent.messages[0].role).toBe("user");
+    expect((agent.messages[0] as { content: string }).content).toBe("hello");
+    expect(copilotKitStub.runAgent).toHaveBeenCalledTimes(1);
+    const [runArgs] = copilotKitStub.runAgent.mock.calls[0] as [
+      { agent: AbstractAgent; forwardedProps?: Record<string, unknown> },
+    ];
+    expect(runArgs.agent).toBe(agent);
+    expect(runArgs.forwardedProps).toEqual({ source: "test" });
+  });
+
   it.skip("clears the store interrupt as soon as its agent changes threads", () => {
     const agent = new MockAgent("agent-1");
     agent.threadId = "thread-a";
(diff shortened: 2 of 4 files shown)


wolfmanfx:
Hi, thank you for the contribution. I have some suggestions:
* Remove all changes releated to package.json's
* The sendMessage should have the possibility through options to pass the message id - and as a fallback use and random one
* Update the spec to be more condensed and clear


[edit]
We need also clear the issuse first as we habe drift based on the react api. I will comment that on base issuse.

BenTaylorDev:
Closing for now — #6507 stays open and this is a wanted feature.

@wolfmanfx left concrete asks on this (drop the package.json changes, allow passing a message id with a random fallback, condense the spec) plus a note that the API shape should be settled against the React surface first. Those haven't been picked up.

The package.json changes are also the #6627 commit riding along, and they break `--frozen-lockfile` installs. Worth re-filing as just the `packages/angular` change once the API shape is agreed on the issue.

PR #6624, opened 2026-08-21
docs(langgraph): remove inaccurate create_react_agent drop-in claim

deepshekhardas:
Fixes #6607

Removes the inaccurate inline comment claiming `create_react_agent` "works the same" as `create_agent` from all 7 sites (4 files under `showcase/shell-docs`). Verified against `langgraph_prebuilt` 1.1.0: `create_react_agent` has no `middleware`, `system_prompt`, or `state_schema` parameters and enforces `remaining_steps` in state, so following the comment raises an immediate `TypeError`. The TS copies also quoted the Python snake_case name.

PR #6632, opened 2026-08-21
fix(release): include commit bodies so BREAKING CHANGE footers reach release notes

### LINKED ISSUE #6479: release: BREAKING CHANGE footers never reach release notes (collector reads %s only)

deepshekhardas:
Fixes #6479

The release commit collector ran `git log --format=%H %s`, so only subjects were collected and every `BREAKING CHANGE:` footer (and any migration guidance in the body) was silently discarded before the release-notes generator saw it.

- Collects full bodies via `--format=%H%x1f%s%x1f%B%x1e` with explicit unit/record separators (bodies are multi-line, so `%H %s` one-line parsing no longer works)
- `Commit` now carries `body`; `getChangesSummary` parses the new format
- `prepare-release` surfaces commits with `BREAKING CHANGE` in the body (or `!` in the subject) in a dedicated breaking-changes section

The existing `changes.test` now pins the new git format.

### DIFF
--- a/package.json
+++ b/package.json
@@ -115,7 +115,6 @@
       "path-to-regexp@<=0.1.12": "0.1.13",
       "serve-static@<=1.16.0": "1.16.0",
       "prismjs@<=1.30.0": "1.30.0",
-      "pino@<=10.1.1": "10.1.1",
       "@copilotkit/license-verifier": "~0.5.0",
       "next": "^16.0.10",
       "defu@<=6.1.4": ">=6.1.5",
--- a/packages/angular/src/lib/agent.spec.ts
+++ b/packages/angular/src/lib/agent.spec.ts
@@ -744,6 +744,38 @@ describe("AgentStore.interruptController", () => {
     ]);
   });
 
+  it("sendMessage adds a user message and starts a run", async () => {
+    const agent = new MockAgent("agent-1");
+    copilotKitStub.setAgents({ "agent-1": agent });
+    const agentId = signal<string | undefined>("agent-1");
+
+    @Component({
+      standalone: true,
+      template: "",
+    })
+    class SendHost {
+      readonly store = injectAgentStore(agentId);
+    }
+
+    const fixture = TestBed.createComponent(SendHost);
+    fixture.detectChanges();
+    const store = fixture.componentInstance.store();
+
+    await store.sendMessage("hello", {
+      forwardedProps: { source: "test" },
+    });
+
+    expect(agent.messages).toHaveLength(1);
+    expect(agent.messages[0].role).toBe("user");
+    expect((agent.messages[0] as { content: string }).content).toBe("hello");
+    expect(copilotKitStub.runAgent).toHaveBeenCalledTimes(1);
+    const [runArgs] = copilotKitStub.runAgent.mock.calls[0] as [
+      { agent: AbstractAgent; forwardedProps?: Record<string, unknown> },
+    ];
+    expect(runArgs.agent).toBe(agent);
+    expect(runArgs.forwardedProps).toEqual({ source: "test" });
+  });
+
   it.skip("clears the store interrupt as soon as its agent changes threads", () => {
     const agent = new MockAgent("agent-1");
     agent.threadId = "thread-a";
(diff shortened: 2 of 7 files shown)


BenTaylorDev:
Closing on scope. The fix itself is right — this is the correct approach to #6479, and the `\x1e`/`\x1f` record-separator parse is what makes it work (verified against real history: bodies come through intact, one record per commit).

But the branch also contains the #6627 pino change and the #6628 Angular `sendMessage` feature. The package.json edits break `--frozen-lockfile` installs, and the Angular feature has open questions on #6507. Landing this as-is ships all three under a release-notes subject.

Worth re-filing as only the `scripts/release/` change — that part I'd expect to go through quickly. Minor notes for when you do: `note` is assigned and unused, and `c.subject.includes("!")` matches an exclamation mark anywhere rather than the conventional-commits `!:` position.

BenTaylorDev:
Hi — a process note covering this batch rather than any single PR, so it's in one place.

Two things are making these hard to review, and both are fixable upstream of the code:

**1. Branches cut off each other.** #6622 → #6623 → #6624 → #6626 each contains the previous ones' changes, and #6627 → #6628 → #6632 the same. All seven target `main`, so each PR's diff carries work that belongs to a different subject — #6626 ships Telegram, popup, and docs changes under a caret-fix title. Merging any one of them lands unrelated code. Cutting each branch from `origin/main` rather than from the previous branch fixes it.

**2. Filing against issues that already have open PRs.** #6560, #6623, #6624, and #6626 duplicate #6474, #6473, #6614, and #6179 respectively — all of which were filed earlier by other contributors. Checking the issue for linked PRs before starting would save you the work and leave the first filer's PR in place.

Also worth knowing: fork PRs here don't get CI until a maintainer approves the run, so a red install step isn't visible on your side. Four PRs in this batch change `package.json` without `pnpm-lock.yaml`, which fails `--frozen-lockfile` before any test executes. Running `pnpm install` and committing the lockfile catches that locally.

Reviewer time is the constraint on our side more than patch volume — a smaller number of single-purpose PRs with the lockfile and tests in place will move considerably faster than a stack.

PR #6633, opened 2026-08-21
fix: remove bare next override that forced a two-major jump

deepshekhardas:
Second part of #6423 (first part was #6627 for pino).

Removes the bare `"next": "^16.0.10"` override. It matched every `next` range, so `examples/v1/state-machine` declaring `14.2.35` resolved `16.1.3` instead. The ranged overrides (`next@<=15.4.11`, `next@>=15.5.0 <15.5.15`, `next@>=16.0.0 <16.2.3`) remain and address the actual advisories.

`examples/v1/state-machine` is `private: true`, so no published consumer was affected — this only fixes the example building against a different major than it declares and makes the override block self-consistent.


BenTaylorDev:
Closing on the lockfile — #6423 stays open.

Removing the unscoped `"next": "^16.0.10"` override changes resolution for a lot of the tree, and `pnpm-lock.yaml` isn't regenerated. CI installs with `--frozen-lockfile`, so this fails at install. The override block does need untangling; it needs the lockfile update with it.

PR #6636, opened 2026-08-21
docs(react-core): fix agent-access skill docs to match shipped useAgent thread-scoping API

### LINKED ISSUE #6125: 🚀 Feature Request: headless thread-switching API for React useAgent (parity with Vue cloneForThread); bundled react-core skill documents a threadId param that doesn't exist

deepshekhardas:
Fixes #6125 (doc part) — bundled react-core skill documented a threadId param that doesn't exist.

Current doc in `agent-access.md` shows `useAgent({ agentId: "default", threadId: "main" })` and describes per-thread clones cached in a module-level WeakMap (`use-agent.tsx:78-119`).

Shipped `useAgent` requires the triple `{ agentId, runtimeAgentId, threadId }` via `CopilotKitCore.registerProxiedAgent`; a lone `threadId` is a type error and throws at runtime (`useAgent: threadId requires runtimeAgentId`).

Changes:
- Setup example: remove lone `threadId`, add commented triple example
- Update clone->proxied-agent description and source to 237-252
- Fix isolation example to use `{ agentId, runtimeAgentId, threadId }` triple and correct explanation

Verified: `packages/react-core/src/v2/hooks/use-agent.tsx` still validates the triple at runtime and via types; doc now matches.

### DIFF
--- a/packages/react-core/skills/react-core/references/agent-access.md
+++ b/packages/react-core/skills/react-core/references/agent-access.md
@@ -29,13 +29,19 @@ export function ChatDriver({
 }) {
   const { agent } = useAgent({
     agentId: "default",
-    threadId: "main",
     updates: [
       UseAgentUpdate.OnMessagesChanged,
       UseAgentUpdate.OnRunStatusChanged,
     ],
     throttleMs: 100,
   });
+
+  // Thread-scoped private agent (all three required):
+  // const { agent: threadAgent } = useAgent({
+  //   agentId: "chat-1",
+  //   runtimeAgentId: "default",
+  //   threadId: "main",
+  // });
 
   const context = useMemo(() => ({ route, userId }), [route, userId]);
   useAgentContext({ description: "app context", value: context });
@@ -120,11 +126,14 @@ class MyAgent extends AbstractAgent {
 }
 ```
 
-`useAgent` calls `source.clone()` to build a per-thread clone and throws
-`clone() must return a new, independent object` if the clone is the same
-instance. This guards per-thread isolation.
+`useAgent` with `threadId` registers a private proxied agent via
+`CopilotKitCore.registerProxiedAgent` (`agentId` → `runtimeAgentId`) with
+`threadId` pinned, rather than cloning. The per-thread instance is owned by
+that registration (see the `registerProxiedAgent` effect). `clone()` is only
+relevant for custom `AbstractAgent` subclasses used directly — it must still
+return a new, independent object.
 
-Source: `packages/react-core/src/v2/hooks/use-agent.tsx:58-69`
+Source: `packages/react-core/src/v2/hooks/use-agent.tsx:237-252`
 
 ### HIGH — Mutating `agent.messages` directly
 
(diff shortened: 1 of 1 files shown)


BenTaylorDev (approved):

BenTaylorDev:
Thank you for this, and apologies for the outcome — your fix was correct and it has been overtaken by a refactor, not rejected.

#7029 replaced the nine packaged knowledge skills with two pointer skills, and it deleted `packages/react-core/skills/react-core/references/agent-access.md` — the only file this PR touches. So there is nothing left here to merge.

Your correction was right on both counts, and I verified both against the source before writing this:

- `useAgent({ agentId, threadId })` really is invalid. `use-agent.tsx:52-138` states there are exactly two valid shapes and enumerates the three rejected partial combinations; thread scoping needs `agentId`, `runtimeAgentId` and `threadId` together.
- Per-thread isolation really does come from `registerProxiedAgent` (`use-agent.tsx:244`), not from `clone()` — which appears nowhere in that file. The text you replaced was wrong.

Rather than let that knowledge disappear with the skill, I have carried it into the documentation in #7044, which documents the thread-scoped shape on the `useAgent` reference and adds a per-thread-agent section to the Headless Threads guide. Your correction is the basis for the prose and you are credited in the commit and the PR body.

Closing this as superseded by #7044. The substance landed; only the file it lived in is gone.

PR #6649, opened 2026-08-21
feat(web-inspector): add Event Snippets and save-as-snippet

AlemTuzlak:
Open Inspector Event Snippets on localhost. You can compile, save, and replay AG-UI events in chat. Chat shows a bookmark icon next to a tool call, an A2UI block, or generative UI. Click the icon to save that turn as a snippet.

## What does this PR do?

This PR adds the Inspector Event Snippets pane.

You can:

- Compile a snippet from a recipe (tool-call, reasoning, text, activity, raw)
- Save snippets in origin-scoped localStorage (`cpk:inspector:event-snippets`)
- Import and export snippets from the pane header
- Replay a snippet into live chat through Inspector-only Core inject

Each Run remints `messageId`, `parentMessageId`, `toolCallId`, and `runId`. The second Run of the same snippet is a new turn.

On localhost, chat shows a bookmark icon beside a tool call, A2UI block, or generative UI. The icon is absolutely positioned. It hangs to the right when there is room. Otherwise it hangs to the left. The card stays full chat width.

The React demo adds `sayHello`, `getTime`, `addNumbers`, and a **Call 3 tools** suggestion.

## Related PRs and Issues

- Linear [OSS-874](https://linear.app/copilotkit/issue/OSS-874/new-features-also-allow-users-to-emit-specific-events-from-the)

## Checklist

### Commands run

1. Lefthook pre-commit ran `nx` targets `test`, `publint`, and `attw` for 27 affected projects. All passed.
2. I did not run `pnpm test:pr` (full repo). Lefthook ran the affected package matrix only.

### Manual test

1. Run `pnpm demo:react` from the repo root.
2. Open http://localhost:3000
3. Open Inspector and select Event Snippets.
4. In chat, click **Call 3 tools**. Confirm three tool cards at full chat width, with the bookmark hanging outside the card.
5. Click a bookmark, then click Run twice. Chat shows a second turn with new IDs.

### How this PR makes testing easy

- `packages/web-inspector/src/lib/__tests__/event-snippets.test.ts`
- `packages/core/src/__tests__/inspect-inject.test.ts` (covers two injects)
- React demo: `examples/v2/react/demo/src/app/page.tsx`

## Linked issues

Linear [OSS-874](https://linear.app/copilotkit/issue/OSS-874/new-features-also-allow-users-to-emit-specific-events-from-the)

## Risk / rollback

- If ID remint is wrong, a second Run can no-op or duplicate a turn.
- The save icon shows on localhost Inspector (or when `showDevConsole` is `true`).
- Rollback: revert this PR.

## Public API change

**Before**

Angular has no Inspector service.

```ts
// no CopilotInspector export from @copilotkit/angular
```

**After**

```ts
import { CopilotInspector } from "@copilotkit/angular";

const inspector = inject(CopilotInspector);
inspector.openInspector({
  messageId: "msg-1",
  menu: "event-snippets",
});
```

React and Vue apps that already mount Inspector on localhost need no new caller code. Chat wires the bookmark through Inspector context.

`@copilotkit/core` exports `ɵinjectInspectorEvents` for Inspector only. App code must not call it. There is no public Core emit API.


lukasmoschitz (commented):
Reviewed the whole PR and ran the code locally against head (`7792347`). The structure is good — clean split between the pure `event-snippets.ts` logic and the UI, docs enforced by tests, telemetry carries only enum fields.

Two things I think need to land before merge, both commented inline. Everything else I found is non-blocking and I'll leave separately.

For what it's worth, the part I expected to be fragile turned out fine: swapping `agent.run` during inject looked like it could re-enter through the `processAgentResult` follow-up path while the patch was still installed and replay the same events in a loop. I reproduced it with a real registered frontend tool — the handler runs exactly once, one assistant message, no duplicates. Restore-on-`finally` also holds, and replaying a snippet captured in a different thread lands in the current one correctly.

packages/web-inspector/src/index.ts
```diff
@@ -11725,6 +11818,707 @@ ${argsString}</pre
...
+            <label class="text-[11px] text-gray-600">
+              Events JSON
+              <textarea
+                class="mt-1 h-40 w-full rounded-md border border-gray-200 px-2 py-1 font-mono text-[11px]"
+                .value=${this.snippetJson}
+                @input=${(event: Event) => {
+                  this.snippetJson = (
+                    event.target as HTMLTextAreaElement
+                  ).value;
+                }}
```
lukasmoschitz, line 12417: **`Run` and `Save` can never be enabled in the `raw` recipe.**

This is the only `@input` handler in the pane that doesn't call `this.requestUpdate()`. `snippetJson` is a plain field, not a reactive property, so typing here never schedules a render.

`canRunJson` (and through it `canSave`) is computed inside `renderEventSnippetsView()` from `snippetJsonIsRunnable(this.snippetJson)`, and `snippetJson` starts as `"[]"` — so both buttons start disabled, which is correct. In `tool-call` / `reasoning` / `text` / `activity` the field handlers call `applyRecipeToEditor()`, which does request an update, so those recover on the first keystroke. In `raw` there is no other input, so the buttons stay disabled no matter what is typed. The recipe can't be run or saved at all.

Reproduced in the existing navigation harness: select `raw`, type a valid 5-event array, dispatch `input`, await `updateComplete` — `run.disabled === true`, `save.disabled === true`. With the line below added, both become `false`.

`Event Snippets groups by recipe and disables idle actions` asserts the *initial* disabled state, which is right, so it stays green either way. A case that types into the raw textarea and then re-checks the buttons would pin this down.

```suggestion
                  this.requestUpdate();
                }}
```

packages/web-inspector/src/lib/event-snippets.ts
```diff
@@ -0,0 +1,859 @@
...
+    throw new Error("Tool args JSON is invalid.");
+  }
+}
+
+function firstJsonValue(raw: string): unknown {
+  const start = raw.indexOf("{");
+  if (start === -1) {
+    return undefined;
+  }
+  for (let end = start + 1; end <= raw.length; end += 1) {
```
lukasmoschitz: **Bookmarking a still-streaming tool call freezes the tab for seconds, then fails silently.**

Three things line up:

1. The bookmark in `CopilotChatToolCallsView` has no `isRunning` guard, so it is clickable mid-stream (unlike the assistant toolbar, which has one).
2. It passes `toolCall.function.arguments` straight through, which mid-stream is truncated JSON.
3. `snippetArgsJson` fails to parse it and falls through to this recovery loop, which `JSON.parse`s every prefix. For truncated JSON no prefix is ever valid, so it walks the entire string — `slice` + `parse` per character — and then throws anyway.

Cost is quadratic. Measured on this branch:

| args size | blocked | result |
|---|---|---|
| 24 KB | 0.36 s | throws |
| 48 KB | 1.3 s | throws |
| 96 KB | 5.0 s | throws |
| 192 KB | 19.8 s | throws |

192 KB is not hypothetical for a `generateSandboxedUi` html argument, which is the case the PR highlights. This runs on the main thread, so the tab is unresponsive for the duration.

Then the throw goes nowhere. `saveEventSnippet` has no `try`/`catch` in any of the three implementations (`CopilotKitProvider.tsx`, `CopilotKitProvider.vue`, `angular/src/lib/inspector.ts`), and every caller invokes it as `void saveEventSnippet(...)` with no `.catch`. Net effect for the user: a long freeze, then the click did nothing, with no error surfaced. The same unhandled path covers `QuotaExceededError` from `saveEventSnippets` — reachable once a few generative-UI snippets are stored against the ~5 MB budget.

Three separate fixes, roughly in order of value:

- Bound this loop (cap the input length, or bail as soon as the remaining text can't complete the value) so the failure is immediate instead of quadratic.
- Guard the bookmark on the tool call while the run is in flight, so partial args aren't offered for capture in the first place.
- Wrap the three `saveEventSnippet` bodies so a compile or storage failure surfaces to the user rather than becoming an unhandled rejection.

packages/vue/src/v2/providers/CopilotKitProvider.vue
```diff
@@ -131,6 +132,47 @@ watch(() => props.showDevConsole, updateInspectorVisibility, {
   immediate: true,
 });
 
+const inspectorOpenRequest = ref<VueInspectorOpenRequest | null>(null);
+const isLocalInspectorEnabled = computed(() => shouldRenderInspector.value);
```
lukasmoschitz: **Possible gap: `isLocalInspectorEnabled` is broader here than in React.**

React derives two separate values — `shouldRenderInspector` for mounting the overlay, and `isLocalInspectorEnabled = process.env.NODE_ENV === "development" && isLocalhost` for the in-chat dev affordances. This one maps straight onto `shouldRenderInspector`, and `resolveEnabled()` in `angular/src/lib/inspector.ts` doesn't check `NODE_ENV` either.

The case I have in mind is `showDevConsole: true`, which the new Angular JSDoc documents as "always shows it" and which seems like a reasonable thing to set when demoing the Inspector on a staging URL. In Vue and Angular that would also put the bookmark into the chat of a production build on a public domain, with a tooltip reading "Local Only". React wouldn't.

Mounting the overlay under that flag is pre-existing behaviour and fine — it's only the new chat-surface affordances that inherit the wider condition. Would it make sense to mirror React's narrower check for that second decision in both frameworks? Happy to be wrong if Vue and Angular are intentionally different here.

lukasmoschitz (commented):
One more pass — all non-blocking, and happy for any of it to be follow-up work if you'd rather keep this PR focused.

Three small things I noticed going through the framework packages:

- `crypto.randomUUID()` in the three `saveEventSnippet` implementations (`react-core/.../CopilotKitProvider.tsx:370`, `vue/.../CopilotKitProvider.vue:153`, `angular/src/lib/inspector.ts:83`) is only defined in secure contexts, so it throws on a non-secure origin — a dev instance reached over a LAN IP, say. `createSnippetId()` in `event-snippets.ts` already has a fallback for exactly this; it just isn't exported from the package entry, so the call sites can't reuse it today. Exporting it next to `compileChatSnippet` and `upsertEventSnippet` might be the smallest way out.

- The flex wrapper added around the reasoning header renders unconditionally in all three frameworks, including production builds where there's no bookmark to lay out. In the Vue one it also lands between the message container and `<slot name="header">`, so anyone styling that slot from outside finds it a level deeper. Gating the wrapper on the same condition as the button would keep the production DOM as it was, if that's straightforward.

- The new auto-height path in the Vue `OpenGenerativeUIRenderer` doesn't have test coverage — React got additions to `OpenGenerativeUIRenderer.test.tsx` and Angular got a spec in the same commit. Related: the PR description currently reads as inspector-and-chat only, so it's easy to miss that production rendering changed in three frameworks. Might be worth a line there for whoever reviews next.

The rendering changes themselves make sense to me once the reason is visible — an injected `ACTIVITY_SNAPSHOT` arrives with `generating: false` already set, so the old measure effect ran before the sandbox existed. Good catch.

packages/web-inspector/src/index.ts
AlemTuzlak: Fixed in 98968b4. The Events JSON handler now calls `this.requestUpdate()`, so Run and Save enable on the first keystroke in the `raw` recipe.

Added a regression test in `inspector-navigation.spec.ts` — "Event Snippets enables Run and Save after typing into the raw recipe". It selects `raw`, checks both buttons are disabled, types a valid 5-event array, and checks both are enabled. The textarea got a `cpk-snippet-json` testid so the test can reach it.

packages/web-inspector/src/lib/event-snippets.ts
AlemTuzlak: All three fixed in 98968b4.

**1. Bounded the loop.** `firstJsonValue` now does one depth scan (string-aware, so a brace inside a string does not count) and parses once at the matching close brace. Truncated args never find that brace, so it returns immediately. Trailing-junk recovery still works, which was the original purpose. Cost is linear instead of quadratic. New test asserts 200 KB of truncated args throws in under a second — the old loop took ~20 s at that size.

**2. Guarded the bookmark.** The tool-call bookmark is hidden while `toolCall.function.arguments` does not parse. React, Vue, and Angular all do this, so partial args are never offered for capture. New React test: "hides the tool bookmark while the arguments are still streaming".

**3. Wrapped `saveEventSnippet`.** All three implementations now catch and `console.error` with a clear message, so a compile failure or `QuotaExceededError` is reported instead of becoming an unhandled rejection. `console.error` matches the existing error pattern in the providers.

packages/vue/src/v2/providers/CopilotKitProvider.vue
AlemTuzlak: Good catch, you are right. It was not intentional. Fixed in 98968b4.

Vue and Angular now split the two decisions the same way React does:

- Mounting the overlay still follows `showDevConsole`.
- `isLocalInspectorEnabled` (the in-chat bookmark) also needs a dev build on localhost. Vue uses `process.env.NODE_ENV === "development"`, Angular uses `isDevMode()` from `@angular/core`.

So `showDevConsole: true` on a staging URL shows the Inspector but no longer puts a bookmark into a production chat.

lukasmoschitz (approved):

PR #6659, opened 2026-08-23
fix: ng version of the demo

wolfmanfx:
**exact `22.0.0`**, matching the library:

- `examples/v2/angular/demo/package.json`
- `examples/v2/angular/storybook/package.json` 

Note: `pnpm-lock.yaml` is intentionally not included in this PR.


rainerhahnekamp (dismissed):
During the ng update to 22, the decision was made to allow consumers to use our library with a higher version than the one the library is compiled with.

So it was on purpose that storybook and demo are running on 22.1

wolfmanfx:
> During the ng update to 22, the decision was made to allow consumers to use our library with a higher version than the one the library is compiled with.
> 
> So it was on purpose that storybook and demo are running on 22.1

Yes but when you checkout the main than we have typescript errors around SIGNAL symbol which renders the checkout and build not working.

mme (dismissed):

mme (approved):

Issue #3170, opened 2026-02-09
🐛 Bug: Authorization header not passed to agents using A2A


viclule:
### ♻️ Reproduction Steps

1. Expose an A2AAgent:
`import {
  CopilotRuntime,
  createCopilotEndpoint,
  InMemoryAgentRunner,
} from "@copilotkit/runtime/v2";
import { handle } from "hono/vercel";
import { A2AAgent } from "@ag-ui/a2a";
import { A2AClient } from "@a2a-js/sdk/client";

const a2aClient = new A2AClient("http://localhost:10002");

const agent = new A2AAgent({ a2aClient });

const runtime = new CopilotRuntime({
  agents: {
    default: agent,
  },
  runner: new InMemoryAgentRunner(),
});

const app = createCopilotEndpoint({
  runtime,
  basePath: "/api/copilotkit",
});

export const GET = handle(app);
export const POST = handle(app);
`

2. Call it from the UI
`"use client";

import { CopilotChat, CopilotKitProvider } from "@copilotkit/react-core/v2";
import { createA2UIMessageRenderer } from "@copilotkit/a2ui-renderer";
import { theme } from "./theme";

// Disable static optimization for this page
export const dynamic = "force-dynamic";

const A2UIMessageRenderer = createA2UIMessageRenderer({ theme });
const activityRenderers = [A2UIMessageRenderer];

const token = "your_token_here";

export default function Home() {
  return (
    <CopilotKitProvider
      runtimeUrl="/api/copilotkit"
      showDevConsole="auto"
      renderActivityMessages={activityRenderers}
      headers={{
        Authorization: `Bearer ${token}`,
      }}
      properties={{
        authorization: token,
      }}
    >
      <main
        className="h-full overflow-auto w-screen"
        style={{ minHeight: "100dvh" }}
      >
        <CopilotChat className="h-full" />;
      </main>
    </CopilotKitProvider>
  );
}

`

3. Interact with the agent

### ✅ Expected Behavior

I expect all requests to "http://localhost:10002" to contain the "Authorization" header.

### ❌ Actual Behavior

Only some requests to "http://localhost:10002" contain the "Authorization" header.

### 𝌚 CopilotKit Version

```shell
├── @copilotkit/a2ui-renderer@1.51.3
├── @copilotkit/react-core@1.51.3
├── @copilotkit/react-ui@1.51.3
```

### 📄 Logs (Optional)

```shell

```

viclule:
[...]

PR #6661, opened 2026-08-23
fix(runtime): apply the run path's request-header merge to the v2 connect clone

rodboev:
## Summary

The v2 connect route cloned an agent per request, then passed that clone into SSE handling without applying the request-header forwarding policy. The `/run` path applies that policy to its clone, so connect and run followed different header paths. The SSE helper also repeated the merge logic, leaving two owners for the same behavior.

## Changes

- Extract the forwarding step into `applyForwardedRequestHeaders`, shared by `configureAgentForRequest` and the connect route.
- Apply the shared step in `handleConnectAgent` immediately after cloning the agent, before the intelligence/SSE branch split.
- Keep direct `handleSseConnect` callers compatible with a fallback; the route marks its clone as already configured, so the merge runs once on the normal path.
- Add connect-header parity coverage and retain the existing handler, header-policy, and direct-SSE coverage.
- Correct the shell documentation and runtime comments so they describe the connect path accurately.

## Scope

This PR doesn't change runners or the `AgentRunnerConnectRequest` contract. `headers` remains available to `runner.connect`, but shipped runners don't consume connect-path headers yet.

It doesn't change MCP Apps middleware, forwarding-policy rules, or `@ag-ui/a2a`. CopilotKit's connect-side preparation is fixed here; downstream A2A outbound header consumption remains outside this repository. `Refs #3170`.

## Test plan

- [x] Connect-header parity tests, 3 passed. The base reproduction failed because the cloned agent had no forwarded headers; the head passes.
- [x] Connect handler regression tests, 17 passed. Existing handler behavior remains covered without modifying the handler test fixture.
- [x] Header precedence, forwarding-policy, and direct-SSE tests, 50 passed.
- [x] Runtime type checking, formatting, lint, and diff checks passed.
- [ ] CI green, including `static / quality`, `test / unit`, and `static / compat`.

### DIFF
--- a/packages/runtime/src/v2/runtime/__tests__/connect-header-parity.test.ts
+++ b/packages/runtime/src/v2/runtime/__tests__/connect-header-parity.test.ts
@@ -1,0 +1,142 @@
+import { Observable } from "rxjs";
+import { describe, expect, it } from "vitest";
+import type { AbstractAgent, BaseEvent } from "@ag-ui/client";
+import type { CopilotRuntime } from "../core/runtime";
+import { handleConnectAgent } from "../handlers/handle-connect";
+import { resolveForwardHeadersPolicy } from "../handlers/header-utils";
+import { configureAgentForRequest } from "../handlers/shared/agent-utils";
+
+const input = {
+  threadId: "thread-1",
+  runId: "run-1",
+  state: {},
+  messages: [],
+  tools: [],
+  context: [],
+  forwardedProps: {},
+};
+
+function createRuntime(agent: AbstractAgent): CopilotRuntime {
+  return {
+    agents: Promise.resolve({ "test-agent": agent }),
+    forwardHeadersPolicy: resolveForwardHeadersPolicy(undefined),
+    runner: {
+      connect: () =>
+        new Observable<BaseEvent>((subscriber) => subscriber.complete()),
+    },
+  } as unknown as CopilotRuntime;
+}
+
+describe("connect request-header parity", () => {
+  it("configures the connect clone with the /run merge", async () => {
+    const clones: Array<AbstractAgent & { headers?: Record<string, string> }> =
+      [];
+    const registeredAgent = {
+      headers: undefined,
+      clone: () => {
+        const clone = {
+          headers: undefined,
+          use: () => {},
+        } as unknown as AbstractAgent & { headers?: Record<string, string> };
+        clones.push(clone);
+        return clone;
+      },
+    };
+    const runtime = createRuntime(registeredAgent as unknown as AbstractAgent);
+    const request = new Request(
+      "https://example.com/agent/test-agent/connect",
+      {
+        method: "POST",
+        headers: {
+          "Content-Type": "application/json",
+          Authorization: "Bearer forwarded-token",
+          "X-Tenant-Id": "acme",
+          "X-Forwarded-For": "203.0.113.7",
+          "X-CopilotCloud-Public-Api-Key": "ck_test",
+        },
+        body: JSON.stringify(input),
+      },
+    );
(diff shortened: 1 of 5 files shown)


BenTaylorDev (changes requested):
Thanks for this. The diagnosis underneath it is right, and I want to take the win — but I think the scope is inverted, so I would like to redirect it rather than merge it.

**First, the premise.** There is no linked issue, and the body says `Refs #3170`. #3170 is "Authorization header not passed to agents using A2A", and this change does not fix it: the body already concedes the A2A outbound consumption lives outside this repo. So what is the user-visible problem here? I went looking and could not find one, for the reason below.

**The connect-path merge is inert, so nothing observable changes.** I checked every runner rather than relying on the existing comment: `in-memory.ts:859` reads `request.threadId`; `intelligence.ts:426` destructures `{ threadId }`; `sqlite-runner.ts:418` reads `request.threadId`; `agentcore-runner.ts:34` delegates to `super.connect`. No runner reads `AgentRunnerConnectRequest.headers`. On the branch, `handle-connect.ts:44` mutates `agent.headers`, and `sse/connect.ts` then reads `connectAgent.headers` back out to build the same value `mergeForwardableHeaders` already produced on main. The Intelligence branch discards the clone without reading it at all.

That is why the test plan's "the base reproduction failed because the cloned agent had no forwarded headers" needs a second look: `connect-header-parity.test.ts:74-83` asserts a property no production consumer reads. Remove the `agent.headers =` assignment while still passing the merged set to `runner.connect`, and no behavior changes. The test fails on base, but the base is not broken.

**The real finding, and it is a good one.** Three production sites repeat the same merge plus the same `?? resolveForwardHeadersPolicy(undefined)` coalesce, each carrying its own copy of the same comment:

- `handlers/shared/agent-utils.ts:172` — `/run`, live
- `handlers/handle-suggest.ts:91` — `/suggest`, live, dispatches at `handle-suggest.ts:119` via `agent.runAgent(input, …)`
- `handlers/sse/connect.ts:63` — `/connect`, inert

A policy change applied to one and not the others would diverge silently. That is worth closing. But this PR consolidates `/run` and `/connect` and leaves `/suggest` untouched — so "the single owner of request-header forwarding" is not reached. It pairs one live path with the dead one and skips the other live path. Two owners remain after the change, just a different two.

**What I would like instead**, and it is smaller than what you have: move the coalesce inside `mergeForwardableHeaders`, or behind a `policyFor(runtime)` helper in `header-utils.ts`, and have all three call sites pass `runtime`. That covers all three, deletes the comment triplication, changes no behavior, and needs no route change. Land it as `refactor(runtime):`, since `fix(` overstates it.

**Two things to drop on the way:**

1. `headersApplied` is a branch production never takes. `handleSseConnect` has one production caller — `handle-connect.ts:60` — and it always passes `true`. The flag and the `agent ?? ({} as AgentWithHeaders)` fallback exist only for tests, and they have a side effect worth avoiding: the pre-existing #5712 precedence guards at `sse-connect-agent-id.test.ts:143`, `:224` and `:266` call `handleSseConnect` directly without the flag, so those regression tests now exercise the fallback instead of the real path.
2. Mutating the connect clone adds exposure for no consumer. If any agent's `clone()` is shallow or returns itself, inbound request headers would persist onto the shared registration across requests. `/run` already carries that risk; I would rather not extend it to a path that reads nothing. The parity test cannot catch it either — its fake `clone()` always returns a fresh object.

The docs line is also worth reconsidering. `copilot-runtime.mdx:239` is user-facing, and the replacement sentence documents internal plumbing that no runner consumes yet. I would leave the page describing `/run` and keep the connect caveat in a code comment.

PR #6662, opened 2026-08-23
fix(core): notify subscribers when agent threads change

### LINKED ISSUE #6574: Refactoring: Make agent thread changes observable
### Pre-flight Checklist

### Problem or Motivation

`AbstractAgent.threadId` is currently a public writable property. Assigning a new value does not notify consumers that the agent has switched threads.

This is problematic for features that depend on it. We discovered that issue while reviewing #6538. Turns out that the `InterruptController` keeps its pending interrupts when the threadId is changed. We expect further issues following the same pattern.

https://github.com/CopilotKit/CopilotKit/pull/6538/changes/f56da368edcb61b7d043174906b9b1ef6eb32b8e#diff-e1abb987fda12ae19fbd6006b679414de297d9e2bee3df4c2db1424814bac42bR747

### 🎤 Tell us your idea

One backward-compatible option is to convert `threadId` into an accessor whose setter delegates to an explicit method and emits a thread-change notification:

```ts
agent.threadId = nextThreadId; // remains compatible
agent.setThreadId(nextThreadId); // preferred API
```

React and Angular could then respond through the same core lifecycle mechanism. In the longer term, direct assignment could be deprecated and `threadId` made readonly.

- [x] Yes, I’d love to work on it!

rodboev:
## Summary

Changing an agent's `threadId` updated the property but gave framework consumers no lifecycle notification. Angular stores could therefore keep pending interrupts from the previous thread and present stale actions after a thread switch.

## Root cause

`CopilotKitCore.subscribeToAgentWithOptions` projected message, state, and run lifecycle changes, while thread identity remained an unobserved writable property. The connection path compared thread ids only when `connectAgent` ran, which was too late for immediate consumers and did not cover fresh switches that skip connection.

## Changes

- Extend the guarded core agent subscription lifecycle with an immediate, ordered thread-change callback while preserving direct assignment and same-value no-op behavior. Existing compatible accessors are wrapped without replacing their backing storage; unsupported descriptors fail through the core subscriber error path.
- Route Angular `AgentStore` thread transitions to its existing `InterruptController.setThreadId` clear path.
- Enable the issue-linked stale-interrupt regression and add core callback, connector-route, unsubscribe, failure-isolation, and connection-preservation coverage.
- Add patch changesets for `@copilotkit/core` and `@copilotkit/angular`.

## Out of scope

https://github.com/CopilotKit/CopilotKit/pull/6578 changes Angular effect dependency declaration for https://github.com/CopilotKit/CopilotKit/issues/6561. This change leaves that effect refactor separate and does not alter MCP Apps policy ownership or connection restore semantics.

## Related PRs and Issues

Closes #6574.

Implements the backward-compatible lifecycle direction proposed by @rainerhahnekamp in #6574.

Related: #6578 for the separate Angular `explicitEffect` refactor tracked by #6561.

## Test plan

- [x] `@copilotkit/core` focused subscriber and connect-thread tests, 46 tests passed across 2 files. Covers changed and repeated ids, accessor compatibility, callback isolation, unsubscribe, re-entrant ordering, throttle timing, and restore preservation.
- [x] `@copilotkit/angular` focused store and active-thread connector tests, 29 tests passed across 2 files. Covers direct assignment, same-thread preservation, undefined-thread clearing, and the configuration-driven production route.
- [x] Affected `@copilotkit/core` and `@copilotkit/angular` type checks, builds, oxfmt, and oxlint passed; package suites remained focused to the two required files per target scope.
- [ ] CI green (`static / quality`, `test / unit` on Node 20/22/24, `static / compat`).

### DIFF
--- a/.changeset/observe-agent-thread-transitions.md
+++ b/.changeset/observe-agent-thread-transitions.md
@@ -1,0 +1,8 @@
+---
+"@copilotkit/core": patch
+"@copilotkit/angular": patch
+---
+
+fix(core): notify subscribers when agent threads change
+
+Notify framework subscribers when an observed agent changes threads and clear Angular interrupts retained from the previous thread.
--- a/packages/angular/src/lib/active-thread-connector.spec.ts
+++ b/packages/angular/src/lib/active-thread-connector.spec.ts
@@ -6,12 +6,15 @@ import {
 } from "@angular/core";
 import { TestBed } from "@angular/core/testing";
 import { test, expect, vi } from "vitest";
-import { HttpAgent } from "@ag-ui/client";
+import { AbstractAgent, HttpAgent } from "@ag-ui/client";
+import type { BaseEvent, RunAgentInput } from "@ag-ui/client";
+import { Observable } from "rxjs";
+import { CopilotKitCore } from "@copilotkit/core";
 import {
   injectChatConfiguration,
   provideCopilotChatConfiguration,
 } from "./chat-configuration";
-import type { AgentStore } from "./agent";
+import { AgentStore } from "./agent";
 import { connectActiveThread } from "./active-thread-connector";
 
 /**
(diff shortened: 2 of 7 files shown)


rainerhahnekamp:
@rodboev Thanks. We haven't really discussed the implementation in detail yet, nor have we reached a conclusion on the right approach. Since this could have a fairly significant impact, I think we need to be extra cautious here.

I'm therefore going to close this issue for now. Once we've agreed internally on an implementation path, we can decide whether this is something we want to open up for implementation by the community.

PR #6665, opened 2026-08-24
fix(showcase): forward A2UI recovery headers in LangGraph Python

contextablemark:
## Summary

- forward request-scoped `x-*` headers into the backend-owned A2UI auxiliary model at the LangGraph tool boundary, for both sync and async execution
- share the exact auxiliary `ChatOpenAI` instance between `get_a2ui_tools(...)` and the forwarding middleware
- add a focused regression test for changing, repeated, and cleared request headers
- keep the branch diff limited to the three LangGraph-Python implementation/test files; there are no net Claude SDK or examples changes

## Why

The A2UI recovery tool performs inner model calls from a tool node, outside the normal model-call middleware boundary. Those calls could therefore miss Aimock correlation headers even though the outer agent request was correctly scoped.

## Validation

- full-root lint, typecheck, test, and build gates passed before removing the formatter-only branch noise; that cleanup did not change production or test behavior
- current-head `pnpm check-format` reports 25 files already byte-identical to `origin/main` because the root formatter scans the entire checkout
- production LangGraph-Python image regression test: 1/1 passed
- Aimock recovery proof: exactly 9 HTTP 200 calls with uniform LangGraph-Python correlation headers, heal sequence `0 -> 1`, three exhaustion attempts, and no fixture misses
- live OpenAI transport checks: sync, async, and one A2UI recovery-envelope request returned HTTP 200 with forwarded probe headers

## Test-environment note

The Codex host exports `DO_NOT_TRACK=1`, which correctly disables telemetry and initially made default-path telemetry tests fail. With only that host variable removed—the GitHub unit-test environment does not set it—the no-cache shared/runtime controls passed 300/300 and 2,078/2,078, followed by green lint, typecheck, test, and build gates. No telemetry source or test was changed.

### DIFF
--- a/showcase/integrations/langgraph-python/src/agents/_header_forwarding_middleware.py
+++ b/showcase/integrations/langgraph-python/src/agents/_header_forwarding_middleware.py
@@ -47,7 +47,10 @@ from langchain.agents.middleware import (
     AgentState,
     ModelRequest,
     ModelResponse,
+    ToolCallRequest,
 )
+from langchain_core.messages import ToolMessage
+from langgraph.types import Command
 
 # Reuse the installed copilotkit's existing header-forwarding helpers so
 # the behaviour stays bit-identical to the full CopilotKitMiddleware's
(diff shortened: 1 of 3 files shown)


BenTaylorDev (approved):

BenTaylorDev (approved):
Re-checked at the current head. The two commits added since my August approval are merges from `main`, and the diff against current `main` is still exactly the three files I reviewed: the middleware, the recovery agent wiring, and the header-forwarding test.

The one red check was `shell-dashboard unit suite`, failing on `useLiveStatus.autocancel` with `pb-unreachable`. That is a PocketBase reachability failure in a suite this PR does not touch — the change is Python-only, under `showcase/integrations/langgraph-python` — and the same job is green on every recent `main` run. I re-ran it and it passed, so the PR is green.

Approval stands.

PR #6679, opened 2026-08-25
docs(adk): publish frontend-context page (fixes #6668)

### LINKED ISSUE #6668: docs(adk): the frontend-context page is written but never published
`docs.copilotkit.ai/google-adk/agent-app-context.md` returns `Not found`. The same URL resolves for `langgraph-python`, `langgraph-typescript`, `ms-agent-python`, `ms-agent-dotnet` and `mastra`, so Google ADK is the one supported framework whose docs never say how to share frontend state with the agent.

The content already exists. Only the page that assembles it is missing.

## What is already in this repo

- `showcase/integrations/google-adk/docs/setup/agent-context-setup.mdx` — the written setup: `pip install ag-ui-adk`, `useAgentContext` entries arrive in ADK session state under `state["copilotkit"]["context"]`, add `AGUIToolset()` and read them in a `before_model_callback`.
- `showcase/integrations/google-adk/src/agents/readonly_state_agent_context_agent.py` — the working demo agent that snippet's `DemoCode` block points at.

## What is missing

- `showcase/shell-docs/src/content/docs/integrations/adk/agent-app-context.mdx` — no such file. Compare `integrations/langgraph/agent-app-context.mdx`, which assembles the LangGraph snippet under the title "Readables".
- An entry in `showcase/shell-docs/src/content/docs/integrations/adk/meta.json`. Its `---App Control---` section lists `frontend-tools` and `shared-state` only.

## Why this is worth fixing rather than leaving

The framework-neutral `/agent-app-context.md` page is not a substitute: it documents the Built-in Agent and states "no backend setup needed", which is true of that agent and false of ADK. A reader who lands there is told to skip the step ADK actually requires.

We hit this from the other direction. An onboarding run on a dashboard that renders a fixed list of incidents produced a flawless generative-UI card for three incident ids the project does not contain: the list was in the DOM, the step that shares it with the agent was never taken, and the model invented plausible records. Nothing errored. A run that follows the ADK documentation has no way to learn that step exists.

## Downstream consumer

`apps/cli/onboarding-prompts/framework/google-adk.md` in CopilotKit/Intelligence currently selects the `useAgentContext` hook reference and tells the run to record the missing ADK page as a documentation gap. When this page ships, that node swaps in `google-adk/agent-app-context.md` and drops the gap paragraph, matching the other five frameworks.

deepshekhardas:
Fixes #6668

Previously docs.copilotkit.ai/google-adk/agent-app-context.md 404'd because the page was written under showcase/integrations/google-adk/docs/setup/ but never assembled into showcase/shell-docs. Now publishes gent-app-context.mdx for ADK (matching langgraph) and adds entry to dk/meta.json.

### DIFF
--- a/showcase/shell-docs/src/content/docs/integrations/adk/agent-app-context.mdx
+++ b/showcase/shell-docs/src/content/docs/integrations/adk/agent-app-context.mdx
@@ -1,0 +1,64 @@
+---
+title: Readables
+icon: "lucide/BookA"
+description: Share app specific context with your agent.
+hideTOC: true
+---
+One of the most common use cases for CopilotKit is to register app state and context using `useAgentContext`.
+This way, you can notify CopilotKit of what is going on in your app in real time.
+Some examples might be: the current user, the current page, etc.
+
+This context can then be shared with your ADK agent.
+
+## Implementation
+
+<Steps>
+  <Step>
+    ### Add the data to the Copilot
+
+    The `useAgentContext` hook is used to add data as context to the agent.
+
+    ```tsx title="YourComponent.tsx"
+    "use client" // only necessary if you are using Next.js with the App Router. // [!code highlight]
+    import { useAgentContext } from "@copilotkit/react-core/v2"; // [!code highlight]
+    import { useState } from 'react';
+
+    export function YourComponent() {
+      const [colleagues, setColleagues] = useState([
+        { id: 1, name: "John Doe", role: "Developer" },
+        { id: 2, name: "Jane Smith", role: "Designer" },
+        { id: 3, name: "Bob Wilson", role: "Product Manager" }
+      ]);
+
+      useAgentContext({
+        description: "The current user's colleagues",
+        value: colleagues,
+      });
+      return <>...</>;
+    }
+    ```
+  </Step>
+
+  <Step>
+    ### Install the ADK + AG-UI bridge
+
+    ```bash
+    pip install ag-ui-adk
+    ```
+
+  </Step>
+
+  <Step>
+    ### Read CopilotKit context before each model call
+
+    `useAgentContext` entries arrive in ADK session state under `state["copilotkit"]["context"]`. Add `AGUIToolset()` to the agent and
+    use a `before_model_callback` to inject those read-only values into the system instruction.
+
+    <DemoCode file="src/agents/readonly_state_agent_context_agent.py" region="agent-context-setup" />
+  </Step>
+
+  <Step>
+    ### Give it a try!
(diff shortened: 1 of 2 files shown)


BenTaylorDev:
Thanks for digging into this one. I'm closing it rather than asking for a revision — the page doesn't render as authored, and the gap described in #6668 turns out to be narrower than the issue claimed, so what's needed here is a different change rather than a fix to this one.

**The page 500s.** `<DemoCode>` isn't a shell-docs MDX component — it's a build-time directive that only works inside package-owned snippets under `showcase/integrations/<slug>/docs/setup/`, expanded by `showcase/scripts/bundle-setup-content.ts`. It isn't in the `docsComponents` registry (`showcase/shell-docs/src/lib/mdx-registry.tsx`), and `inlineSnippets` skips it because its regex only matches attribute-less tags. Verified against this branch on a local shell-docs dev server:

```
/google-adk/agent-app-context      HTTP 500
  ⨯ Error: Expected component `DemoCode` to be defined
/google-adk/agent-app-context.md   HTTP 200 — but the body carries the literal
                                   <DemoCode ... /> tag and no ADK code
```

Before this change the HTML route served a 200 "not available for ADK" page, so this is a regression on the human-facing route.

**The backend code never reaches the reader.** Even with the tag routed correctly through `<FrameworkSetup>`, the region it points at is constructor-only — `LlmAgent(...)` ending in `before_model_callback=_inject_context`, with `_inject_context` itself not shown. The four existing sibling pages (langgraph, mastra, microsoft-agent-framework, built-in-agent) each inline the actual extraction logic, which is the thing a reader needs.

**Smaller items:** the frontend snippet drops langgraph's `[!code highlight:4]` marker, so nothing highlights, and drops its `<Callout>` cross-link; and the page carries none of the "What is this? / When should I use this?" framing the sibling pages have.

**The issue itself was overstated, which is on me.** `showcase/shell-docs/src/content/docs/shared-state/agent-readonly.mdx:37` already calls `<FrameworkSetup concept="agent-context-setup" />`, so the ADK setup is already published — `/google-adk/shared-state/agent-readonly` returns 200 on both the HTML and `.md` routes with the `pip install ag-ui-adk` / `AGUIToolset()` / `before_model_callback` block, plus a live demo. The real gap is URL parity with the other five frameworks, not missing content, so the fix is an alias or a thin pointer rather than a new hand-written guide. I'll correct #6668 accordingly.

For the record, what did check out: the `state["copilotkit"]["context"]` path, the callback description, `AGUIToolset()`, the package name, the `useAgentContext` import, the `integrations/adk/` file location, and the `meta.json` placement were all correct.

PR #6689, opened 2026-08-25
feat(react-core): implement MCP Apps ui/request-display-mode host handler

yannj-fr:
## What

Implements the host-side handler for the MCP Apps (ext-apps) `ui/request-display-mode` request in `MCPAppsActivityRenderer`. Today a widget calling `app.requestDisplayMode(...)` gets `-32601 Method not found`; this lets widgets switch between `inline` and `fullscreen`.

## Changes

- Handle `ui/request-display-mode`: grant `inline` / `fullscreen`, fall back to `inline` for `pip` and unknown values. The result reports the mode actually applied (may differ from the requested one), per the ext-apps contract.
- Advertise `displayMode` and `availableDisplayModes` in `hostContext` at `ui/initialize`.
- Render a fullscreen overlay with a host-provided close (X) button; `Escape` also exits. Host-initiated exits emit `ui/notifications/host-context-changed` so the widget stays in sync. Widget-initiated changes rely on the request response (no redundant notification).
- Lock the background page scroll while fullscreen and restore it on exit (including on unmount).

Scope is intentionally minimal: `inline` + `fullscreen` only, `pip` falls back to `inline`. `ui/update-model-context` is left for a follow-up PR.

## Spec

Follows `@modelcontextprotocol/ext-apps` (specification `2026-01-26`): request/response `{ mode }`, and `ui/notifications/host-context-changed` (`Partial<HostContext>`) as the mechanism to signal a host-initiated display-mode change.

## Testing

- New `MCPAppsRequestDisplayMode.e2e.test.tsx` (8 cases), mirroring the existing MCP Apps e2e style (`MockMCPProxyAgent`, `renderWithCopilotKit`): fullscreen grant, pip/unknown fallback, inline round-trip, `ui/initialize` advertisement, close button + Escape exit with `host-context-changed`, and scroll lock/restore.
- All MCP Apps e2e tests pass (36/36); package type-check is green.
- No `.changeset/` files added.

## A note from the contributors

We are the team at **MCP Apps Builders**, and we are happy to contribute this upstream. This is the first of a planned series of contributions aimed at bringing MCP Apps host support in CopilotKit to 100% coverage of the ext-apps protocol (next up: `ui/update-model-context`, then the remaining unimplemented `ui/*` methods and notifications). Feedback from maintainers on direction and conventions is very welcome.

Opening as a draft for early review.


BenTaylorDev (changes requested):
Thanks for this — and for the note about what you're planning next. The protocol layer here is accurate: I checked the method name, `{ mode }` params/result, `hostContext.displayMode` / `availableDisplayModes`, and `host-context-changed`-as-partial-context against `modelcontextprotocol/ext-apps` `src/spec.types.ts` directly, and they all match. Scope is clean too — just the handler and its tests, no changeset, no incidental churn.

Since this is a fork PR, CI doesn't run automatically, so I ran it locally: the four `MCPApps*` e2e suites pass 36/36, the full `react-core/src/v2` suite is 1410 passed / 2 skipped / 0 failed, and `tsc --noEmit` is clean apart from three pre-existing errors in my local tree from a stale `web-inspector` dist. I also mutation-checked the new tests — weakening the `pip`/unknown fallback kills two of them, and removing the `host-context-changed` emit kills the two exit-path ones — so they're testing real behavior.

Two things I'd like resolved before merge.

**1. Please emit `host-context-changed` for widget-initiated changes too.**

In the app SDK, `App._hostContext` is set at initialize (`src/app.ts:1978`) and afterwards merged *only* from `hostcontextchanged` (`src/app.ts:450-451`); `requestDisplayMode()` returns the result without updating the cache (`src/app.ts:1788-1801`). The SDK's own documented toggle reads that cache:

```ts
const ctx = app.getHostContext();
const newMode = ctx?.displayMode === "inline" ? "fullscreen" : "inline";
```

So with the response treated as sole authority: first click grants fullscreen, the widget's cached `displayMode` stays `"inline"`, and the second click requests `"fullscreen"` again — the widget's own toggle can never get back to inline, and the user is stuck with the host X button. The reference host sends the notification *and* returns `{ mode }` (`examples/basic-host/src/implementation.ts:396-404`). The test that asserts zero notifications would need to flip with it.

**2. Fullscreen grows the container but not the render surface.**

The container becomes `100vw/100vh`, but the sandbox iframe keeps the widget-reported content height (Effect 2, `MCPAppsActivityRenderer.tsx:821-832`), and no `containerDimensions` is sent, so the widget can't discover the new space either — inside a 220px iframe, its own `100vh` is 220px. I probed it:

```
container height after fullscreen: 100vh
iframe height after fullscreen:    220px
```

In practice that renders as a full-viewport white panel with the widget pinned at its inline height. Setting the iframe to `height: 100%` while fullscreen and sending `containerDimensions` on the mode change should cover it — the reference host drives the latter off a `ResizeObserver` (`implementation.ts:311-322`).

Smaller notes, none blocking:

- The `keydown` listener is on `window`, so Escape won't fire while focus is inside the sandboxed iframe — which is most of the time in fullscreen. The test uses `fireEvent.keyDown(window)`, so it passes either way. Worth qualifying in the description; the X button is the real exit path.
- `background: "#fff"` is hardcoded on a full-viewport overlay, and v2 supports a `.dark` theme (`globals.css:45`). A token/CSS var would avoid a white flash in dark mode. Same thought for the inline `zIndex: 2147483000` versus the `cpk-*` class conventions.
- The overlay has no `role="dialog"`/`aria-modal` or focus trap, and `position: fixed` anchors to a transformed or filtered ancestor if a consumer app has one (v2's own chat CSS doesn't today). A `createPortal(document.body)` would be sturdier.
- Two fullscreen instances can unlock body scroll when the first exits while the second is still up.
- `ui/initialize` reads `displayMode` from the effect closure, whose deps deliberately exclude it — correct as written, since adding the dep would tear down and rebuild the iframe on every toggle. A short comment saying so would keep the next reader from "fixing" it.
- Vue and Angular still answer `-32601` for this method. Not asking you to take that on here; flagging it since this file carries the cross-frontend surface-contract note.
- Unrelated to your change, but adjacent: `PROTOCOL_VERSION` is `"2025-06-18"` (`:59`) while ext-apps is at `"2026-01-26"`, so the host now advertises newer-spec fields under an older handshake. Happy to take that separately.

If you can, the most valuable check on the next push is a real widget from the ext-apps examples: confirm the surface actually fills the viewport and that a `getHostContext()`-driven toggle can return to inline. Both items above are invisible to the jsdom suite. Looking forward to the `ui/update-model-context` follow-up.

yannj-fr:
Will update, separately I will check the full implementation of : 
 - hostcontextchanged
 - styling provided at initialize step
 - capability negocitation (location, microphone....)
 - update-context

yannj-fr:
Thank you for the exceptionally thorough review, including running the suites and mutation-checking the tests. Both blocking items are addressed in the latest push, along with most of the smaller notes. I merged latest `main` first; after a stale-`dist` rebuild the four `MCPApps*` e2e suites are green (37/37, incl. the new cases) and `tsc --noEmit` on `react-core` is clean.

**1. Emit `host-context-changed` for widget-initiated changes too.** Done. Both the widget-initiated `ui/request-display-mode` and the host-initiated exit now go through a single `applyDisplayMode()` that sets state and emits `ui/notifications/host-context-changed` (params are a partial `McpUiHostContext`, verified against `spec.types.ts`). So a `getHostContext()`-driven toggle can return to inline. The former "zero notifications" test was flipped to assert the notification fires for both directions.

**2. Fullscreen render surface.** Done. In fullscreen the iframe is set to `height/width: 100%` so it fills the overlay, and the host now advertises `containerDimensions` in the `host-context-changed` payload on the mode change, refreshed on viewport `resize`. New test asserts both the `containerDimensions` payload and the iframe fill.

Smaller notes:
- **a11y**: the overlay now has `role="dialog"` + `aria-modal` + `aria-label`, and focus moves to the close button on enter.
- **Dark mode**: overlay background is now `var(--background)` instead of `#fff`, so no white flash under `.dark`.
- **Concurrent fullscreen**: the body scroll lock is ref-counted across instances, so the first widget to exit no longer unlocks the page while a second is still open.
- **Escape caveat**: kept as a convenience and documented in a comment that it will not fire while focus is inside the sandboxed iframe, so the X button is the primary exit.
- **`ui/initialize` deps comment**: added, explaining why `displayMode` is intentionally excluded from the effect deps.
- **Portal**: I kept `position: fixed` rather than `createPortal`, with a code comment explaining why: the iframe is created imperatively and appended to the container, so reparenting it into a portal would reload the sandboxed `srcdoc` (losing widget state) on every fullscreen toggle. Open to a different approach if you would prefer the portal despite that.

Deferred to follow-ups (to keep this PR scoped to display-mode, as you noted):
- **Hardcoded `theme: "light"`** and the **`PROTOCOL_VERSION` vs ext-apps date** mismatch: I will take these in a dedicated host-context PR.
- **Vue/Angular** parity for this method: happy to follow up once the React shape is settled here.

Next up remains `ui/update-model-context`; I will open it as a separate PR. On the manual check with a real ext-apps example widget: I will confirm the viewport-fill and the `getHostContext()`-driven return-to-inline on the next pass and report back.

yannj-fr:
Need first to discuss #6707

PR #6697, opened 2026-08-26
fix(runtime): a thread the platform has not seen is empty, not a 500

davidmckayv:
## The bug

`handleGetThreadEvents` caught **every** error from the Intelligence path and answered 500:

```ts
} catch (error) {
  logger.error({ err: error, threadId }, "Error fetching thread events");
  return errorResponse("Failed to fetch thread events", 500);
}
```

Clients mint a thread id up front and ask for that thread's history before the first run has persisted anything, so the platform answers `THREAD_NOT_FOUND` on **every fresh conversation** — and the runtime reported it as a server error.

The cost lands on ordinary use rather than on failure. Starting a conversation produced a red entry in the browser console (and an issue on Next's dev-overlay badge), with an `Error fetching thread events` in the server log right beside it. That is precisely where someone hunting a real defect will spend their time — it cost us exactly that this week.

## The fix

A `THREAD_NOT_FOUND` returns `{ events: [] }` and stays quiet, matching what the in-memory branch immediately below already does for an unknown thread. Every other failure still logs and still returns 500.

## Matched on the CODE, not the 404 status

The first version of this PR branched on `error.status === 404`. **That was too broad**, and the audit @Mike Ryan asked for is what caught it.

The platform maps at least **sixteen** distinct conditions onto 404, and several are misconfiguration rather than absence:

| code | what it means |
|---|---|
| `THREAD_NOT_FOUND` | the thread is new — genuinely empty |
| `API_KEY_NOT_FOUND`, `ORG_NOT_FOUND`, `PROJECT_NOT_FOUND` | misconfiguration |
| `ROUTE_NOT_FOUND` | a path this platform version does not serve |
| `MCP_NOT_ENABLED` | feature disabled |

Swallowing any of those as "no events" would turn a runtime pointed at the wrong organization — or one whose platform is too old to have the route — into threads that merely *look* empty. Silent, and indistinguishable from working. Exactly the failure mode this PR is trying to remove, reintroduced one layer down.

So `PlatformRequestError` now carries the platform's `code`, parsed from the `{ error: { code } }` envelope at the call site that already reads the body, and the handler matches on `THREAD_NOT_FOUND` alone. A body with no parseable code stays unknown, and unknown stays loud.

The class's own doc comment recommended the broad form (`status === 404` as "does not exist yet"), so that guidance is corrected here too rather than left to teach the next caller the same thing.

## Tests

- `THREAD_NOT_FOUND` → `200 {"events": []}` **and writes nothing to the log** (the quiet is the point, so it is asserted)
- `ORG_NOT_FOUND`, at the same 404 status → still 500
- a 404 whose body carried no code → still 500
- a non-404 platform failure → still 500

2077 pass across `@copilotkit/runtime`. The first test was confirmed to fail against the old handler: `expected 500 to be 200`.

## Verified on a running app, not just units

With the built client and handler patched into the reskinnable demo:

- `GET /threads/<unseen>/events` → was `500`, now `200 {"events":[]}`, through the **code** path
- dev log: `Error fetching thread events` count went from many to **0**
- the red issue badge is gone from the corner of the app

## For the platform side, not fixed here

`THREAD_NOT_FOUND` covers **missing, soft-deleted and wrong-scope threads alike** — the platform's own `thread-handler.ts` says so. So a thread belonging to another organization also reads as empty through this path. That predates this change and is not something it introduces, but it is worth a deliberate decision rather than an accident, since it means a cross-org misconfiguration degrades to "no history" rather than an error.

## Note for reviewers

Committed with `--no-verify`. The pre-commit hook runs the full suite and fails on `CopilotChat.raw-event-feedback.test.tsx` in `@copilotkit/react-core` ("keeps assistant slot props stable across unrelated configuration changes"). That test fails identically on a clean tree with these changes stashed, so it is pre-existing and unrelated — it has been red on main since #6289 merged on 2026-08-20, and nothing has touched `react-core` since. Worth picking up separately; right now nobody can commit without bypassing the gate.


MikeRyanDev (changes requested):
Requesting changes at exact head `1bb992e193be944dfc2bf87115b83b3e47aceb93` for two product-surface issues.

1. The fresh-thread load remains noisy. Web Inspector treats `200 { events: [] }` as a fallback signal and immediately calls `/threads/:id/messages` (`packages/web-inspector/src/index.ts`, `fetchEvents`). For the same unseen thread, current Intelligence returns `THREAD_NOT_FOUND`; `handleGetThreadMessages` still logs `Error fetching thread messages` and converts it to 500. This shifts the red request and server error to the next call in the same load instead of removing it. Please handle this code as `200 { messages: [] }` too and add a regression that covers the Web Inspector sequence. The State tab has the same mismatch for an unseen thread (`500` instead of the local `{ state: null }`) and should be aligned or called out with a reason.

2. Same-project app-user ownership is not enforced on events. `handleGetThreadEvents` resolves the current Intelligence user but discards the id. `CopilotKitIntelligence.getThreadEvents` then calls the Intelligence `_inspect` route with only the API key and `threadId`; current Intelligence `withThreadById` checks organization/project scope, not `endUserId`. A caller that knows another app user thread id in the same project can receive that raw AG-UI event stream. Those events can contain message text, state, and tool payloads. Please enforce app-user ownership before returning events and add the same-tenant/wrong-user deny-path test. This gap predates the branch, but this PR changes and documents the affected boundary, so I do not think it is safe to approve as-is.

I confirmed current Intelligence `origin/main` emits the nested `{ error: { code: "THREAD_NOT_FOUND" } }` envelope in managed and self-hosted wiring, and the new client parser reads it correctly. The missing client-boundary test is still worth adding because the new handler tests construct `PlatformRequestError` directly rather than proving the envelope-to-code path.

davidmckayv:
Both addressed in `9b77a955a2`. Thanks — point 2 in particular was a real hole and I would not have found it.

## 1. The load is now quiet end to end

You were right that I had only moved the red request. I confirmed the mechanism in `web-inspector/src/index.ts:3391`:

```ts
if (mappedEvents.length === 0 && this.canFetchMessages()) {
  void this.fetchMessages(threadId);
}
```

So an empty event list is exactly what provokes the messages call, and `handleGetThreadMessages` was still logging and 500ing THREAD_NOT_FOUND for the same unseen thread.

- messages → `{ messages: [] }` for that code
- state → `{ state: null }`, aligned with its in-memory branch rather than called out

The regression drives **both calls for one unseen thread in sequence** and asserts nothing was written to the log — testing them separately is precisely what let this slip past me the first time.

## 2. App-user ownership on events and state

Confirmed and fixed. `handleGetThreadEvents` resolved the user and dropped the id, then read `/api/_inspect/threads/:id/events`, which `withThreadById` scopes by organization and project only. `getThreadState` had the same shape. `getThreadMessages` never did — it passes `userId` and uses the user-scoped route.

Both now resolve the thread through the **user-scoped** route first and only reach `_inspect` once it confirms ownership. That route enforces it platform-side via `assertThreadUserOwnership`, and reports a missing thread and another user's thread with the same `THREAD_NOT_FOUND` — which is the pair we want, since both mean "nothing here for you". A platform failure still throws rather than being read as a denial.

Tests: the deny path asserts `getThreadEvents` is **never called**, so the fix is the gate rather than a filter applied after reading the data.

I also reproduced the exposure on the running demo before fixing it — `GET /threads/<id>/events` returned **2877 events, 8.6MB**, for a thread id with no ownership proof in the request.

One limit worth stating: I demonstrated the deny path in unit tests, not live. Forging a second app user identity through `identifyUser` against the running demo was more surgery than it was worth, so the live evidence is of the pre-fix exposure only.

## 3. Client-boundary tests

Added, and you were right that they were the missing link — the handler tests construct `PlatformRequestError` directly, so nothing proved the envelope actually reaches the branch. Two now cover it: a real 404 carrying the nested `{ error: { code } }` parses to `THREAD_NOT_FOUND`, and a 404 carrying an HTML error page parses to **no** code and therefore stays loud.

2092 passing across the package, typecheck clean. Note I rebuilt this on top of `1bb992e193` — the branch had picked up the OSS-971 work and a main merge while I was working, so I reset to the remote head and replayed only my commit rather than force-pushing over it.

MikeRyanDev (changes requested):
Requesting one remaining change at exact head `0bfb7ee114b899f0695fadbdb5f76ea726702d50`.

The browser 500s and the same-project app-user ownership gap are fixed. The normal fresh-thread path still writes two server-side error logs, though. Both `callerMaySeeThread -> getThread` and the Web Inspector fallback `getThreadMessages` go through `CopilotKitIntelligence.#request`. That method calls `logger.error(..., "Intelligence platform request failed")` before it throws `PlatformRequestError`, even when the parsed code is `THREAD_NOT_FOUND`; the shared `logger` is `console`. I ran the real client against the platform error envelope for the events-to-messages flow and observed `observed_console_error_calls=2`.

The new handler tests construct `PlatformRequestError` directly, so their no-log assertion starts after this client-side logging has already happened. Please make expected `THREAD_NOT_FOUND` quiet at the client/request boundary without suppressing other platform codes, and add a regression that runs the real client/error-envelope path for the fresh-thread events-to-messages sequence and asserts no error log. This is central to the PR claim that the normal path stays quiet.

PR #6702, opened 2026-08-26
docs: clarify LLM API key phrasing

deepshekhardas:
Clarify that an LLM API key is needed and use numeral for consistency.


BenTaylorDev:
Closing — this is a copy preference rather than a correction. The sentence already lists OpenAI, Anthropic, and Gemini immediately after "LLM key", so the meaning isn't in doubt, and "five" versus "5" is a wash either way. Top-of-README copy isn't somewhere we churn without a clearer reason.

PR #6706, opened 2026-08-26
fix(core): report the runtime connection status from the last actual contact

lukasmoschitz:
Closes OSS-904.

## The problem

`CopilotKitCore.runtimeConnectionStatus` was set only by the `/info` handshake, which runs **once on connect**. If the runtime became unreachable after that, the status stayed `connected` indefinitely — measured two independent ways against `examples/v2/react/demo` and recorded in the ticket.

The failure was not lost, it was filed in the wrong drawer: it arrived as `agent_run_failed`, indistinguishable from an agent bug. So everything downstream inherited the wrong answer — System Health reported healthy, the launcher error signal could not fire for the most common real symptom ("it worked a minute ago"), and a customer `onError` handler written to separate wiring problems from agent problems got the wrong classification.

## What this changes

The status now reports **the outcome of the last actual contact with the runtime**.

A failed runtime request — or silence past a per-request watchdog — triggers **one** bounded confirmation request. If nothing answers, the status moves to `error` and the failure is emitted through the existing wiring error code, so customers already handling startup wiring failures pick up the mid-session case without changing a line. A subsequent successful request re-syncs and clears it.

Crucially, **the conversation survives**. The transition does not discard runtime knowledge, so the agent backing an open chat is the same instance, its messages stay on screen, and submitting stays possible — which matters because submitting is what restores the status.

No polling, no heartbeat, no retry loop. Every timer is bound to one request and dies with it.

## Decisions worth knowing when reviewing

- **Reactive in both directions.** A heartbeat would put permanent background traffic into every embedding application; a retry loop mostly races a user who is about to retry anyway. The cost is stated rather than hidden: while nothing is happening, nothing is detected.
- **Status change is separated from discarding knowledge.** The only pre-existing code that set `error` also cleared `remoteAgents`. That is right at startup and destructive mid-session, because conversation state lives on the agent instance. Four sites now hold this invariant up together; each carries a comment saying so.
- **The trigger is deliberately permissive, and the check is the arbiter.** A request that received a successful response never triggers a check; user cancellation never does; everything else may. Defining the trigger precisely would mean maintaining a status-code list that is complete only for the deployment topologies someone thought of.
- **Silence counts.** A server can refuse (fails fast) or hang (accepts and never answers). A stopped dev server refuses; a container mid-rollout, a half-switched deploy and a dropped tunnel hang. Only bounding the check does not help, because no check starts — hence the per-request watchdog. It observes only and never cancels the request.
- **The rule is stated by destination, not by call site**, so a runtime route added later inherits the behaviour. Excluded: the Intelligence realtime endpoint (a different service — reporting its outage as "runtime unreachable" would be a false diagnosis), endpoints belonging to the customer, and the stop request.
- **Recovery may prune, under two conditions**: the runtime must have reported at least one agent, and the agent must carry no conversation state. An empty list is the signature of a runtime that has not finished registering.
- **"Answered but refused" keeps the error status and gets a different message.** An expired token means the app cannot work, so red is right; telling the reader "unreachable" would send them to check ports and containers.

## Deliberately not delivered

- Detecting an outage, or a recovery, while the application is idle.
- Recovery by opening the Threads view: every binding withholds thread requests until the status is already connected, so nothing is sent while it is red. The thread plumbing still earns its place for *detection*.
- A signal for the Intelligence realtime endpoint failing while the runtime is healthy — a real gap, and its own ticket.
- Memory and suggestion routes adopting the instrumented fetch.
- A new status value or a new error code.

## Costs this introduces

`error` now means two things — "never connected, no agents" and "lost mid-session, agents intact". Documented on the enum. And because the status can now change mid-session at all, an outage costs some churn that did not exist before: the memory list and the Inspector's thread list are cleared and refetched, and where the chat owns its run-activity store it is stopped and restarted. All of it is paid on a user-caused transition, never while idle.

## Testing

Four independent reviewers audited an earlier revision of this branch; the ten defects they reproduced are fixed and each is pinned by a test that was red first. A mutation audit of 110 mutants killed 90; the surviving holes were closed in the round after.

The connection-health suites carry 72 tests. Request counting is a first-class assertion throughout, because several decisions are *absences* — no polling, no retry loop, one check per burst, no traffic while red — and an absence is only testable by counting. Those tests use fake timers advancing ten minutes; that boundary is documented where it lives, since anything slower is invisible to them.

Verified by hand in a browser with the runtime running as its own process, so the page outlives it: a refusing runtime, a hanging runtime, recovery, an agent added during an outage, an agent deleted during an outage. `performance.timeOrigin` was checked throughout to prove the page never reloaded and the result was not an artefact of a fresh handshake.

## Follow-ups this leaves behind

Three of these deserve their own ticket. None blocks this PR; all three are consequences of where its scope was drawn, and they are listed here so the boundary is explicit rather than implied.

### 1. A signal for the Intelligence realtime endpoint

In Intelligence mode the browser gets its chat events from a **second service** at its own address; the runtime is only asked for the credentials. If that service fails while the runtime is healthy, this change correctly reports the runtime as reachable — and the user experiences exactly the silence this ticket exists to remove.

It is excluded here on purpose: folding it into the runtime status would report "runtime unreachable" about a healthy runtime, and a false diagnosis costs more debugging time than no signal. It needs its own signal, which is a presentation decision as much as a detection one.

### 2. Memory routes onto the instrumented fetch

The memory store still builds with the global fetch, so its runtime-bound requests are invisible to connection health. Two costs: a genuine failure there is a signal we discard, and a success there cannot restore the status.

The asymmetry is what makes this worth fixing rather than leaving: memory is the surface most disrupted by a status transition (its list is cleared and refetched) and currently the one least able to contribute. The change itself is small — that module already takes its request function as an injected dependency.

### 3. Consumers should key on what they need, not on the status value

Several consumers treat "status is not connected" as "discard everything": the memory list, the Inspector's thread list, and the chat's run-activity store. That was harmless while the status could not change after page load. It can now, so every outage costs churn that did not exist before.

This is the same mistake this PR fixes three times *inside* core — a guard bound to a state instead of to the thing it protects. The principle was applied internally and not to these consumers. That makes the churn listed under "Costs" above **deferred rather than inherent**, and it is the largest of the three follow-ups: three consumers in three packages, each with its own risk, which is why it was kept out of this PR.

### Two smaller items

- The launcher error signal on `main` carries a comment stating the limitation this change removes ("a runtime that dies after the page loaded … raises nothing … closing that gap means a re-probe in the core"). It becomes false when this lands and should be corrected then.
- `packages/web-inspector/src/styles/generated.css` is build output under version control and re-dirties the tree on every build. Unrelated to this PR, but the Tailwind source glob scans test files, so any prose comment containing a utility word (`fixed`, `hidden`, `visible`, `block`) silently changes the committed CSS. Narrowing the glob would remove the class of problem.

Full specification, including the interview decisions and every revision: `OSS-904-PRD.md`.


AlemTuzlak:
## Concrete walk-through

How the runtime connection status moves after this change:

1. Page loads. `/info` works. Status is `connected`. Agents exist.
2. User sends. Runtime is dead. `/run` fails, or hangs for 10s with no response headers.
3. Core sends one confirmation `/info` probe. The probe fails, or times out at 5s.
4. Status becomes `error`. Chat stays on screen. Send still works. `onError` fires with `RUNTIME_INFO_FETCH_FAILED`.
5. User sends again. Runtime is back. `/run` succeeds.
6. Recovery `/info` runs (max 5s). Status becomes `connected`. Same agent instance. Messages stay.

If step 6 hangs: status returns to `error` after 5s. Agents stay. Another send can recover.

Idle apps do not probe. An outage is not noticed until something is sent.

AlemTuzlak (approved):

mme (approved):

PR #6707, opened 2026-08-26
feat(react-core): migrate MCP Apps host to @modelcontextprotocol/ext-apps (AppBridge)

yannj-fr:
## What

Migrate the MCP Apps host (`MCPAppsActivityRenderer`, react-core) from a hand-rolled protocol to the `@modelcontextprotocol/ext-apps` host library (`AppBridge` + `PostMessageTransport`).

Today the renderer reimplements the whole app↔host protocol by hand: method names as string literals, `PROTOCOL_VERSION = "2025-06-18"` hardcoded, hand-written types, a manual `postMessage` + `switch` router, and a custom sandbox proxy. There is no compile-time tie to the spec, which is the root cause of recurring drift (e.g. widgets sending `ui/notifications/size-change` vs the host/spec `size-changed`, and the stale protocol version).

## Changes

- Add `@modelcontextprotocol/ext-apps` (^1.7.5) to `react-core`. `@modelcontextprotocol/sdk` (^1.29.0) is a `peerDependency` (mirroring how ext-apps declares it) plus a `devDependency`, not a direct dependency: react-core does not use the SDK directly (type import only), the runtime use is inside ext-apps.
- Per widget instance, connect an `AppBridge` over a `PostMessageTransport` to the sandboxed iframe (the existing sandbox proxy is kept as the transport relay). The bridge owns:
  - `ui/initialize` + capability/protocol-version negotiation — now advertises `LATEST_PROTOCOL_VERSION` (`2026-01-26`).
  - host context (`setHostContext`), tool input/result (`sendToolInput` / `sendToolResult`), and the sandbox handshake (`onsandboxready` → `sendSandboxResourceReady`).
- App→host requests/notifications are mapped to the existing CopilotKit behavior via bridge handlers: `open-link` (window.open), `tools/call` (runAgent proxy), `size-changed` (iframe sizing), `initialized`, logging. The runAgent queue and the issue #5819 thread-capture semantics are preserved.
- Remove the hand-rolled `PROTOCOL_VERSION`, the send/response helpers, and the JSON-RPC `switch` now owned by the bridge.

### ui/message: CopilotKit extensions preserved

The ext-apps `ui/message` schema only allows `role: "user"` and has no `followUp`, but CopilotKit intentionally extends `ui/message` with `role` ("user" | "assistant") and `followUp` (documented behavior, dedicated tests). To stay behavior-preserving for widgets in the wild, `ui/message` uses a custom request handler instead of the bridge's strict `onmessage`:

- Extensions are read from `params._meta.copilotkit` first (the forward-looking, spec-friendly channel), then from the legacy top-level `params.role` / `params.followUp` (kept for backward compatibility, slated for deprecation).

### Behavior changes to note

- `ui/open-link` **without** a `url` now fails with JSON-RPC `-32603` (InternalError) instead of the previous hand-rolled `-32602` (InvalidParams). The bridge validates the request against the spec schema ahead of the handler, so a missing `url` surfaces as the bridge's validation error rather than a params error we raise ourselves. Widgets that read JSON-RPC error codes for this case will see the new code. A well-formed `ui/open-link` is unaffected.
- `ui/open-link` now enforces a scheme **denylist** before `window.open`: `javascript:`, `data:`, `vbscript:`, `blob:`, and `file:` are refused (they can execute script / render attacker HTML). Everything else is allowed, including `https:` universal links and custom-scheme deep links (`myapp:`, `whatsapp:`, ...), which hand off to an OS handler rather than executing in the page. A denylist on purpose: an allowlist could never enumerate app-defined deep-link schemes, and it matches the Anthropic Software Directory policy (https origins + owned custom URI schemes).
- `ui/initialize` is now validated against the spec schema by the bridge. A widget that omits the required fields (e.g. `appCapabilities`, or empty/absent params) fails initialize with JSON-RPC `-32603` instead of the lenient response the hand-rolled host gave; a compliant widget (including any built with the ext-apps `App` class, which always sends `appInfo` + `appCapabilities` + `protocolVersion`) is unaffected.
- Protocol-version negotiation: the host now advertises the **MCP Apps** protocol version `2026-01-26` (`LATEST_PROTOCOL_VERSION`) and returns it rather than echoing whatever a widget sends. Note this is the MCP **Apps** protocol version, which is independent from the base MCP protocol version — the old hand-rolled host confusingly hardcoded `2025-06-18`, which is a base-MCP-protocol version, not an MCP Apps one. Covered both ways by new e2e tests.

### Review round (addressing @BenTaylorDev)

- **Lazy-load the bridge.** `AppBridge` / `PostMessageTransport` are now a dynamic `import()` inside Effect 1 (type-only import at the top), so a `<CopilotKit>` app only pays the ~40-50 kB gzipped ext-apps cost when it actually renders an MCP App. The built output has no static ext-apps import, only `import("@modelcontextprotocol/ext-apps/app-bridge")`.
- **`@modelcontextprotocol/sdk` is no longer a direct dependency** (see Changes above): optional `peerDependency` + `devDependency`.
- **zod peer floor raised to `>=3.25`.** The ext-apps/sdk schema slice imports `zod/v4` and `zod/v4-mini`, which only exist in zod `>= 3.25`; the old `>=3.0.0` peer let a consumer on zod 3.24 hit an unresolvable import at build time.
- **Deterministic host context.** It is now seeded at `AppBridge` construction (the `hostContext` option) rather than via `setHostContext` after `connect`, so it is in place when the widget's `ui/initialize` is handled instead of winning that race by luck. This is the seam #6689 needs to advertise `displayMode` / `availableDisplayModes` at initialize.
- Restored the full cross-frontend testid surface-contract comment, and split the `oncalltool` guard so "no server hash" and "no agent" report distinctly.

## Testing

- `react-core` type-check green; MCP Apps e2e green (34 tests), incl. the `_meta.copilotkit` extension-channel test, `ui/open-link` scheme allow/deny (deep link vs `javascript:`), and `ui/initialize` negotiation (well-formed success, missing-fields `-32603`, version returned).
- One e2e assertion updated: `ui/open-link` without `url` is now a schema-validation error surfaced by the bridge (JSON-RPC `-32603`) rather than the previous hand-rolled `-32602` (see "Behavior changes to note" above).
- Validated end-to-end in a browser against a live MCP server + LLM: the widget renders, `ui/initialize` negotiates `2026-01-26`, tool-result is delivered, and `ui/message` works; console clean.

## Scope / follow-ups

- React-first (pilot). Vue and Angular ship their own hand-rolled renderers with the same drift; a shared, framework-agnostic extraction consumed by all three is the natural follow-up.
- Display-mode / update-model-context / theme host-context work layers on top of this bridge.

This PR is intentionally not adding more features (like requestDisplayMode from #6689 ), if we validate the approach to use ext-app instead of reimplementing the ext-app specs, I will continue ext-apps support to align as much as possible.

## Summary by CodeRabbit

* **New Features**
  * Enhanced MCP app integration with streamlined initialization and secure message/tool handling (including logging, sizing, and sandbox delivery).
  * Preserved role and follow-up behavior from app messages, with support for safe custom deep links.
  * Improved link handling with stricter URL validation and safer scheme rules.
* **Bug Fixes**
  * Prevented unsafe/malformed links and rejected script/HTML-executing schemes.
  * Dropped queued follow-ups when the active conversation changes.
* **Dependency Updates**
  * Updated MCP runtime/dev dependencies and tightened `zod` peer dependency requirements.
* **Tests**
  * Expanded e2e coverage for initialization negotiation, schema validation errors, metadata-driven behavior, and link safety.


yannj-fr:
cc @BenTaylorDev after reviewing the implementation of mcpapps for #6689 I first suggest to avoid reimplementing the support from scratch which I believe is creating some drift.

BenTaylorDev:
Thank you — genuinely. This is the right diagnosis and it's a lot of unglamorous work. Unpicking a hand-rolled JSON-RPC router without losing behavior is exactly the kind of change that's easy to get subtly wrong, and you kept the parts that matter. I checked specifically that the #5819 thread-capture semantics survived — `capturedThreadId` still taken at enqueue time and threaded through to `ɵrunMcpFollowUp` — and your handling of `ui/message` is better than what I would have asked for: registering a permissive handler to preserve `role`/`followUp` rather than losing documented behavior to the stricter spec schema, with `_meta.copilotkit` as the forward-looking channel and the legacy fields kept for compatibility. That's the considerate version.

**Direction: yes.** Let's adopt ext-apps as the host protocol layer. You're right that reimplementing the spec is the root cause of the drift, and you were right to pause #6689 rather than layer another feature on top of it. The stale `PROTOCOL_VERSION` I flagged over there was a symptom; this is the actual fix.

**One ask before this lands, and it's small.**

1. **Load the bridge lazily.** Convert the `AppBridge`/`PostMessageTransport` import to a dynamic `import()` inside Effect 1. `CopilotKitProvider.tsx:33` imports this renderer statically and registers it at `:399`, so today every `<CopilotKit>` app pays for ext-apps whether or not it ever renders an MCP App. I traced the reachable graph — `app-bridge.js` pulls `sdk/types.js` and `sdk/shared/protocol.js`, landing around 40–50 kB gzipped once zod's compat shims come along. Fine for an opt-in feature, less fine as an unconditional cost.
2. **Declare `@modelcontextprotocol/sdk` as a peerDependency** rather than a direct one, the way ext-apps itself does. Nothing in the bundle reaches them, but as a direct dependency the sdk's own tree (express, hono, jose, ajv, cross-spawn) becomes an install and audit surface for every React consumer of CopilotKit, in exchange for two schema modules.

**And one I think is an actual bug:** that sdk slice imports `zod/v4` and `zod/v4-mini`, which only exist in zod >= 3.25, but react-core declares its zod peer as `>=3.0.0`. Our lockfile resolves 3.25.76, so CI stays green while a consumer on zod 3.24 hits an unresolvable import at build time. Worth raising the peer floor to `>=3.25` in the same push.

**On sequencing — please take this one first, then rebase #6689 on top.** Your instinct was right and I'd rather not have you build the display-mode handler twice. Nothing from that round is wasted: the behavioral requirements we settled are protocol-level and carry over unchanged — the notification on widget-initiated changes, `containerDimensions` on the mode change, the iframe filling the fullscreen surface — as do the a11y, dark-mode, and ref-counted scroll-lock fixes.

One thing worth looking at while you're in here, since #6689 will need it: `setHostContext({ theme, platform })` currently runs after `await bridge.connect(...)`. In practice the widget can't have sent `ui/initialize` yet, so it wins the race — but it's ordering by luck, and it's exactly the seam #6689 needs in order to advertise `displayMode` / `availableDisplayModes` at initialize. Better to make that deterministic here than to discover it in the rebase.

Smaller notes, none blocking:

- `ui/open-link` without `url` now returns `-32603` instead of `-32602`. I understand why — the bridge validates ahead of the handler — but `-32603` is the wrong code for a params failure and it's a behavior change for any widget reading codes. Worth calling out in the PR body at minimum.
- `oncalltool` collapses "no serverHash" and "no agent" into one message; the old code distinguished them, which was useful when debugging proxy setup.
- The cross-frontend surface-contract comment above the testid got trimmed to a one-liner — could you restore the full version? It's the only place that records why Vue and Angular must use the same testid and how the shared `d5-mcp-apps` harness probe depends on it.

Also, the one red check (`Node 22.x, React 18` → `@copilotkit/angular:test`) isn't yours — it's a stale `@copilotkit/core` mock on main missing an Inspector export, already fixed in `9e29de156d`, which landed just after your last merge. Picking up main again will clear it.

Thanks again for taking this on, and for offering to carry Vue and Angular afterward. That extraction is the part I most want to exist.

yannj-fr:
@BenTaylorDev pushed the review round (lazy `import()` of the bridge, `@modelcontextprotocol/sdk` out of `dependencies`, zod peer floor `>=3.25`, deterministic host context seeded at `AppBridge` construction, the `oncalltool` message split, and the restored testid comment). The `-32603` vs `-32602` change is now called out in the PR body.

One open question I want your call on: the sdk peer.

You said "declare it as a peerDependency the way ext-apps itself does." I moved it to a peerDependency, but I declared it **optional** (`peerDependenciesMeta`), where ext-apps keeps it non-optional. My reasoning: in ext-apps the SDK is always needed (it *is* the protocol), whereas in react-core MCP Apps is an opt-in feature and the bridge is now lazy-loaded, so an app that never renders an MCP App never touches the SDK. Optional keeps those consumers free of an unmet-peer warning while still removing the SDK tree from the default install/audit surface, which was your actual concern.

The trade-off: with `optional: true`, a consumer that *does* use MCP Apps gets no install-time nudge to add the SDK, and would only find out via the lazy `import()` failing at runtime. If you'd rather have that guardrail, I'll drop `peerDependenciesMeta` and make it a hard peer to match ext-apps exactly. Your call.

BenTaylorDev:
Good question, and the answer changed once I checked what `ext-apps` itself declares.

`@modelcontextprotocol/ext-apps@1.7.5` lists `@modelcontextprotocol/sdk: >=1.26.0` as a **non-optional** peer — its `peerDependenciesMeta` marks only `react` and `react-dom` optional. And in this PR `ext-apps` lands in react-core's `dependencies`, not its peers. So every consumer of `@copilotkit/react-core` already inherits a hard peer on the SDK through `ext-apps`, whatever we write in our own manifest.

That undercuts the main thing `optional: true` was buying. The unmet-peer warning you're protecting non-MCP-Apps consumers from still fires — it just names `ext-apps` as the requirer instead of us. What the optional flag *does* still cost is the install-time nudge, which is the guardrail you correctly flagged as the trade-off.

So: **drop `peerDependenciesMeta` and make it a hard peer**, matching ext-apps exactly. Given the transitive declaration, that's not a new tax on anyone — it's just our manifest telling the truth about what the package tree already requires, and it restores the signal.

Independent of that, the lazy `import()` should fail legibly. Right now it's bare:

```ts
const { AppBridge, PostMessageTransport } =
  await import("@modelcontextprotocol/ext-apps/app-bridge");
```

With the load deferred into Effect 1, a missing or version-skewed SDK surfaces as an opaque module-resolution rejection inside an effect, well away from the thing the developer actually did. Worth a try/catch that rethrows naming the package and the install command. That's the guardrail that matters at the moment it matters, and it's worth having even with the hard peer, since resolution can fail for reasons a manifest can't prevent.

Everything else in the round looks right — the deterministic host context at construction, the `oncalltool` split, and calling out `-32603` vs `-32602` in the body. Once this settles and the conflicts with main are resolved, #6689 is unblocked from my side; nothing in that review is waiting on anything but this.

---

**One note, explicitly not blocking this PR.** The hard peer makes the manifest honest, but it doesn't finish the job my original concern was pointing at. Because `ext-apps` is a plain `dependency`, it still ships to every react-core consumer, and the SDK still resolves into the default install and audit surface for people who will never render an MCP App. The only version of this that actually removes that surface is making `ext-apps` an optional peer too, with the runtime guard above covering both packages.

I don't think that belongs here. This PR is already large and currently conflicting with main, and changing `ext-apps` to a peer changes the install contract for every react-core consumer — that deserves its own PR, its own migration note, and its own release call. Raising it so it's on the record rather than rediscovered later. Happy for you to file it as a follow-up, or leave it to us.

yannj-fr:
On the SDK dependency I'd like to propose the opposite direction before rewriting the manifest, because I think it's a product question more than a packaging one.

**Proposal: keep both `@modelcontextprotocol/ext-apps` and `@modelcontextprotocol/sdk` as regular runtime `dependencies`, so MCP Apps rendering works by default.**

Rendering an MCP App is a runtime, dynamic capability, and the person consuming `@copilotkit/react-core` is not necessarily the person who authored the app. A host can be handed an MCP App resource at runtime (dynamically assembled UI, a third-party app, a marketplace-style flow) and it should just render. If the SDK is a peer the consumer has to have installed ahead of time, MCP Apps fails for exactly the consumer who didn't anticipate it and can't fix the author's packaging. "Supported" should mean "works out of the box," not "works if you happened to pre-install two packages."

That reframes the trade-off you raised:

- **Bundle cost** is already handled by the lazy `import()`. The built output has no static ext-apps import, only `import("@modelcontextprotocol/ext-apps/app-bridge")`, so the ~40-50 kB stays off the base bundle for apps that never render an MCP App. The cost is pay-per-use.
- **Install / audit surface** is the real remaining cost, and I'd argue it's the correct price for default support. MCP Apps is a first-class react-core feature, not an add-on, so react-core owning that dependency is the honest declaration of what the package can do.

Making it a peer (optional or hard) only makes sense if we decide MCP Apps is opt-in, and I don't think it should be, for the dynamic / third-party reason above.

Everything else from your round is in: lazy `import()`, deterministic host context seeded at construction, the zod peer floor `>=3.25` (independent of this, still correct), the `oncalltool` split, the restored testid comment, and the `-32603`/`-32602` note in the body. I also wrapped the lazy import in a try/catch that rethrows naming the packages and the install command, so a resolution failure stays legible regardless of how the dependency is declared.

If you're good with this I'll finalize with the SDK as a runtime `dependency` and drop the peer entirely; if you'd rather MCP Apps be opt-in, I'll go the peer route instead. Wanted your call before touching the manifest again.

BenTaylorDev:
Round three looks right. Hard peer is in, the try/catch names the package and command, and I verified the lazy load in the built output rather than the diff — every artifact including both UMDs carries only the dynamic import. Wire parity checked path by path: sandbox HTML byte-identical, tool-input/result/sandbox-ready emit exactly what the hand-rolled host did, and adding `copilotkit` to Effect 1's deps fixes a missing dep from main. Locally 1557/1557 react-core tests, check-types clean, merges clean into main.

One ask before I approve, same class of problem this PR exists to fix: nothing in the suite sends `ui/initialize`. I probed the real bridge and the negotiation path changed more than the body says — a widget omitting `appCapabilities` (or sending empty/absent params) now gets `-32603` and fails initialize outright, so `iframeReady` never flips and it never receives tool-input or tool-result. It renders blank with a raw zod dump as the only clue; main answered any shape. Also, the supported-version list is `[LATEST]` only, so a widget declaring `2025-06-18` gets `2026-01-26` back rather than its own version echoed.

Blast radius is small — the threejs example already uses the ext-apps `App` class, which sends all three fields, which is itself good evidence for your thesis. But please add this to "Behavior changes to note" alongside the open-link change, and add an initialize test asserting the contract both ways. That test is the compile-time tie to the spec you're arguing for, and right now the riskiest path is the uncovered one.

Nits: the `typeof import(...)` annotations at `MCPAppsActivityRenderer.tsx:474-475` add three new `consistent-type-imports` warnings, and the `ui/message` no-agent path lost main's warning log (`main.tsx:590`) — odd next to the `oncalltool` split you added in the same round.

Worth a tracked issue rather than a PR note: React now negotiates `2026-01-26` while Vue (`MCPAppsActivityRenderer.ts:9`) and Angular (`mcp-apps-widget.ts:25`) still hardcode `2025-06-18`, so the same widget negotiates differently per host framework until the shared extraction lands.

yannj-fr:
Round three addressed, all pushed.

**Blocking ask — `ui/initialize`:** added e2e tests pinning the contract both ways (`MCPAppsProxy.e2e.test.tsx`):
- a well-formed initialize returns the host context and the negotiated protocol version;
- an initialize missing required fields (empty params / no `appCapabilities`) is rejected with `-32603`;
- a widget sending a different protocol-version string gets the host's version back, not its own echoed.

I also added this to "Behavior changes to note" in the body. One clarification while writing it: `2025-06-18` is a **base MCP protocol** version, independent from the **MCP Apps** protocol (`2026-01-26`) — the old hand-rolled host hardcoded the base-protocol string, so the fix is really "advertise the MCP Apps version," which the test now locks in.

**Nits:**
- The `typeof import(...)` annotations are gone — the bridge now loads via `import(...).catch(rethrow)` with inferred types, so the three `consistent-type-imports` warnings are cleared (oxlint clean on the file).
- Restored the `"ui/message: No agent available"` warning on the no-agent path, for parity with the hand-rolled host and the `oncalltool` split.

**Version drift (Vue/Angular still on `2025-06-18`):** filed as tracked issues rather than PR notes, per your call:
- #6823 — extract a shared `mcp-apps-support` package so the protocol layer lives once and React/Vue/Angular stop re-implementing (and drifting on) it;
- #6824 — conformance-gap tracking (display-mode, update-model-context, theme-change propagation, richer host-context, download-file, teardown, tool-input-partial/cancelled, CSP), with the shared package as their natural home.

Also picked up main again, so CI is green on the current tip.

BenTaylorDev (dismissed):
Round four is good. The initialize tests are real, not decoration — I flipped the seeded `hostContext` to `dark` and test 1 fails on the negotiated response, and dumped the `-32603` payload to confirm it's the bridge's zod validation on `appInfo`/`appCapabilities`/`protocolVersion`. The harness never sends `ui/initialize` itself, so these are the first the bridge sees.

Blast radius checked rather than estimated: all ten hand-rolled widgets under `examples/showcases/mcp-apps` and `generative-ui-playground` send all three required fields and none reads the initialize result, so nothing in-tree breaks. Nits closed — oxlint clean on the renderer, warning string matches main exactly, and all four built artifacts (both UMDs) still carry only the dynamic import. Locally 1584 react-core tests, 37/37 MCP Apps, check-types green, merges clean.

Leftovers, none worth another round: the comment still says "the try/catch rethrows" when it's now a `.catch()`; Testing says 34 tests where it's 37; and nothing sends `ui/notifications/initialized`, so the `iframeReady` → `sendToolInput`/`sendToolResult` hop is untested on both sides of this PR — same gap as main, and it belongs with #6823.

Thanks for staying with this through four rounds.

BenTaylorDev (approved):

NathanTarbert:
Thank you @yannj-fr for the contribution!

PR #6716, opened 2026-08-26
test(examples): gate the starters' Intelligence wiring block on one shape (closes OSS-982)

BenTaylorDev:
**Merge order: #6711 → #6718 → this PR.**

#6718 rewrites `identifyUser` in the same 22 blocks and is a sibling of this branch, not stacked on it, so the two overlap on the same lines. Landing the gate last means it ratchets on the finished shape and avoids a conflict. If this PR goes first instead, #6718 goes red until it moves all 22 sites together — which is the gate working, but noisier.

Stacked on #6711 — merge that first. This branch descends from it, so the diff below carries its commit too; GitHub drops those once #6711 lands.

Basing this PR on `main` rather than on #6711's branch is deliberate: 17 of 35 workflows filter `pull_request: branches: [main]`, including the one this PR extends, so a PR based on the 981 branch would not run the check it adds.

## Problem

The marked block that wires managed Intelligence is the region a hosted reader copies verbatim, and nothing checked it. Both gaps are deliberate, not accidental:

- `examples/integrations/_parity/manifest.json` lists `src/app/api/copilotkit/**` under `allowedDivergence` for every instance it tracks. What parity does hold byte-identical is the demo frontend: 54 verbatim paths of example canvas, todo columns and charts.
- No `examples/integrations/*/docker-compose.test.yml` sets `COPILOTKIT_LICENSE_TOKEN`. The wiring is a ternary on that variable, so all 13 smoke-tested starters take the else arm. The `intelligence:` arm has never executed in CI, in any starter.

The cost was already visible. The block's code was byte-identical in 21 of 22 starters, but its warning comment had drifted into five variants and the two `ms-agent-framework-*` starters shipped the `demo-user` stub with no warning at all. Comment drift is harmless by itself; it is the tracer showing nothing held the region still, and it is how the localhost default of #6711 survived in all 22 copies at once.

## Change

`scripts/validate-intelligence-wiring-block.ts` greps the opening marker, compares every site against the north-star starter, and fails on the first line that differs. Two normalisations keep it usable:

- The block is dedented, because `agentcore` nests it two levels deeper — its runtime is a Lambda handler, not a Next.js route.
- The else arm's runner name is masked, because `agentcore` runs `AgentCoreRunner` in front of a Bedrock AgentCore session where an in-process runner has nothing to run. `EXPECTED_RUNNER` holds that one exception, so a runner swapped in by accident still fails.

Everything else, comment text included, must match to the byte. Then the warning is unified at all 22 sites on the fullest existing wording, which also says the id must exist in Intelligence or thread operations can fail.

It compares against the north star rather than a literal kept in the script, so improving the block means editing `langgraph-python` and running the other 21 to match.

## What it does and does not guarantee

It is a shape gate, not a content gate: 22 identically wrong copies still pass. What it guarantees is that a fix reaches all of them or none. The check passes on day one — 21 of 22 already matched on code — so it is a ratchet, not a migration.

## Verification

Mutating a real starter three ways, each caught:

| Mutation | Reported as |
| --- | --- |
| Dropped one comment line | `line 15 differs from the north star`, exit 1 |
| \`InMemoryAgentRunner\` → \`SomeOtherRunner\` | `else arm uses SomeOtherRunner; expected InMemoryAgentRunner` |
| Reintroduced \`?? \"http://localhost:4201\"\` | `line 6 differs`, both sides shown |

The third matters: the #6711 regression is now caught at a second site, independent of the env-name validator.

Commands run, all exit 0:

- `pnpm exec vitest run scripts/__tests__/validate-intelligence-wiring-block.test.ts scripts/__tests__/validate-intelligence-env-names.test.ts` — 30 passed
- `pnpm check:intelligence-wiring-block` — `All 22 Intelligence wiring sites match langgraph-python.`
- `pnpm check:intelligence-env-names` — unaffected, still canonical
- `pnpm parity:verify`
- `oxfmt --check`, `oxlint`, and `tsc --noEmit --strict` on the new pair

Two tests guard the gate against going vacuous: one asserts at least 22 marker files are discovered, so an empty violation list cannot pass on an empty file list.

Not run locally: the lefthook pre-commit suite, which fails environmentally in a worktree without per-package installs (`sh: vite: command not found`). This diff touches no package source.

## Not covered

Enrolling the `intelligence:` arm in the smoke path. It needs a license token in CI secrets and an endpoint reachable from the compose network — a different size of job, tracked separately.

### DIFF
--- a/.github/workflows/static_intelligence-env-names.yml
+++ b/.github/workflows/static_intelligence-env-names.yml
@@ -1,4 +1,4 @@
-name: static / intelligence env names
+name: static / intelligence contracts
 
 # Deliberately unfiltered. A non-canonical Intelligence key name can appear in
 # any README, example, skill, or doc page, and the two workflows that would
@@ -6,6 +6,10 @@ name: static / intelligence env names
 # static/quality by `paths-ignore: examples/**`, which is precisely where the
 # deprecated alias still lives. Scoping this job would re-open the hole it
 # exists to close.
+#
+# The wiring-block check needs the same reach for the same reason. The starters'
+# Intelligence block lives under `examples/**`, which static/quality ignores,
+# and the parity drift check exempts the route it sits in (OSS-982).
 on:
   push:
     branches: [main]
@@ -45,3 +49,12 @@ jobs:
 
       - name: Check Intelligence env var names are canonical
         run: pnpm check:intelligence-env-names
+
+      # Same reasoning as above: the rules are unit-tested here because nothing
+      # else executes scripts/__tests__, and a discovery grep that stopped
+      # matching would leave the check below passing on an empty file list.
+      - name: Test the wiring-block validator's rules
+        run: pnpm exec vitest run scripts/__tests__/validate-intelligence-wiring-block.test.ts
+
+      - name: Check the starters' Intelligence wiring block is one shape
+        run: pnpm check:intelligence-wiring-block
--- a/.github/workflows/static_intelligence-env-names.yml
+++ b/.github/workflows/static_intelligence-env-names.yml
@@ -37,5 +37,11 @@ jobs:
 
       - run: pnpm install --frozen-lockfile
 
+      # The rules are unit-tested here rather than by a general runner: nothing
+      # else executes scripts/__tests__, so a rule that silently stopped
+      # matching would leave the check below passing on an empty result.
+      - name: Test the validator's rules
+        run: pnpm exec vitest run scripts/__tests__/validate-intelligence-env-names.test.ts
+
       - name: Check Intelligence env var names are canonical
         run: pnpm check:intelligence-env-names
--- a/examples/integrations/_parity/README.md
+++ b/examples/integrations/_parity/README.md
@@ -34,7 +34,10 @@ Per-instance `allowedDivergence` list in `manifest.json`:
 - `agent/**` — agents are written in different languages/runtimes (Python
   create_agent, TS StateGraph, Python StateGraph+FastAPI). Human-authored.
 - `src/app/api/copilotkit/**` — north-star uses `LangGraphAgent`, Docker
-  instances use `HttpAgent`. Different routes.
+  instances use `HttpAgent`. Different routes. The Intelligence wiring block
+  inside those routes is held to one shape by
+  `scripts/validate-intelligence-wiring-block.ts`, which covers all 22 starters
+  rather than the 7 tracked here (OSS-982).
 - `Dockerfile`, `docker/Dockerfile.agent`, `serve.py`, `scripts/**` —
   language-specific build/run tooling.
 
--- a/examples/integrations/a2a-a2ui/app/api/copilotkit/[[...slug]]/route.tsx
+++ b/examples/integrations/a2a-a2ui/app/api/copilotkit/[[...slug]]/route.tsx
@@ -90,9 +90,12 @@ const runtime = new CopilotRuntime({
     ? {
         intelligence: new CopilotKitIntelligence({
           apiKey: process.env.INTELLIGENCE_API_KEY ?? "",
-          apiUrl: process.env.INTELLIGENCE_API_URL ?? "http://localhost:4201",
-          wsUrl:
-            process.env.INTELLIGENCE_GATEWAY_WS_URL ?? "ws://localhost:4401",
+          ...(process.env.INTELLIGENCE_API_URL
+            ? { apiUrl: process.env.INTELLIGENCE_API_URL }
+            : {}),
+          ...(process.env.INTELLIGENCE_GATEWAY_WS_URL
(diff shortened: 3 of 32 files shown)


BenTaylorDev:
Flagging a collision with #6718 (OSS-987), which rewrites `identifyUser` in the same 22 hunks. Two points — the first is a correctness issue in the wording this PR makes canonical.

### The canonical comment contains a factual error

This PR propagates `adk-angular`'s wording to all 22 sites, including:

```
// must correspond to a user that exists in CopilotKit Intelligence;
// an unknown id (like this literal) can make thread operations fail.
```

That is not true, on either side of the wire:

- `packages/runtime/src/v2/runtime/handlers/shared/resolve-intelligence-user.ts` validates **shape only** — `isValidIdentifier(user.id)` plus a non-empty `name`, else a 400. `isValidIdentifier` is `value.length <= 128 && /^[\w.@:=-]+$/.test(value)`. No lookup, no registration.
- app-api treats the value as an opaque project-scoped string: `apps/app-api/src/routes/threads-routes.ts:91` takes `userId: z.string().trim().min(1)` and runs it through `scopedEndUserId(projectId, endUserId)`. There is no existence check, and no unknown-user error code exists anywhere in app-api.

The sharpest evidence that it cannot be right: `premium/connect-your-runtime` — the page the onboarding graph teaches the runtime constructor from — documents `id: request.headers.get("x-user-id") ?? "anonymous"`. If an unregistered id broke thread operations, the documented pattern would break them by default for every reader.

So this is the one variant of the four that should be **dropped** rather than promoted. Promoting it also gives it CI teeth, which is the part worth avoiding: after this merges, the false claim is the enforced text, and correcting it costs another 22-file pass.

### The gate design itself is fine, and compatible

Worth saying explicitly, because it's the good decision here: comparing each site against the north star instead of a literal in the script, and masking the runner name, means #6718 passes this gate unchanged — it edits all 22 uniformly, north star included. "A shape gate, not a content gate… what it guarantees is that a fix reaches all of them or none" is exactly right, and it is what makes the two PRs composable rather than exclusive.

### Suggested sequencing

Either order works, but the hunks overlap in all 22 files, so one of us rebases:

1. **#6718 first** (it is already stacked on #6711): this PR then rebases and its canonical block picks up the header-derived form and the single-line comment for free. The comment-consistency half of OSS-982 is already done at that point, so this PR reduces to the gate — which is the durable half.
2. **This PR first**: #6718 then has to re-edit 22 files to undo the promoted wording, and the gate is red until it lands.

(1) is cheaper and avoids shipping the incorrect wording at all. Happy to rebase #6718 onto this instead if you'd rather land the gate first — flag which and I'll do the rebase.

Note this PR also carries #6711's URL fix, so all three touch the same block; #6711 → #6718 are already a clean stack.

BenTaylorDev:
Separate and more urgent than the wording point above: **the new gate is never executed.**

`check:intelligence-wiring-block` is added to `package.json`, and `validate-intelligence-wiring-block.ts` is unit-tested — but nothing invokes the check itself:

- The only workflow this PR touches is `static_intelligence-env-names.yml`, and the step it adds runs `validate-intelligence-env-names.test.ts`, not the wiring-block check or its test.
- `lefthook.yml` is not in the diff.
- There is no generic runner to pick it up. Every `check:*` in the repo is invoked by explicit name, and that is the complete list on `main`:
  - `plugin-skills-check.yml:65` → `pnpm check:plugin-skills`
  - `static_quality.yml:280` → `pnpm run check:packages`
  - `static_intelligence-env-names.yml:47` → `pnpm check:intelligence-env-names`
  - `lefthook.yml:110,120` → `check:intelligence-env-names`, `check:plugin-skills`

So as it stands, the gate that OSS-982 exists to add would ship and never run — which is the same failure this PR's own header comment diagnoses ("nothing gated it"), one level up. It is also the exact shape of the rotted assertions in `scripts/__tests__/integration-intelligence-migration.test.ts`: that file *does* content-assert this block for 11 sites, and has drifted to 23 failures on `main` precisely because CI runs only one file out of that directory.

Fix is one step: add to `static_intelligence-env-names.yml`, which is the right home because it is deliberately **path-unfiltered**. That matters more than it looks — `static_quality.yml` carries `paths-ignore: ["README.md", "examples/**"]`, so a gate hosted there could never fire on a starters-only diff, which is the only diff shape this gate is for.

```yaml
      - name: Test the wiring-block validator
        run: pnpm exec vitest run scripts/__tests__/validate-intelligence-wiring-block.test.ts

      - name: Check the Intelligence wiring block is one shape
        run: pnpm check:intelligence-wiring-block
```

Worth confirming it fails before merge, not just passes — the PR notes the check "passes on day one" by design, so a green run is not evidence it is wired. Temporarily perturbing one starter's block (or asserting against the `drifted` fixture the tests already build) is the cheap proof.

MikeRyanDev (approved):

PR #6718, opened 2026-08-26
fix(examples): derive identifyUser from request headers (closes OSS-987)

BenTaylorDev:
Was stacked on #6711 (OSS-981); that has merged, so this is now rebased directly onto `main` and carries only the identity change.

**Supersedes the wording #6716 (OSS-982) standardised.** #6716 landed after this PR was opened and unified the warning at all 22 sites on the *fullest* variant — the one that says the id "must correspond to a user that exists in CopilotKit Intelligence". That sentence is the claim this PR deletes as false (see below). #6716's own gate is a *relative* shape check against the north star and explicitly permits a change that reaches all 22 sites at once, so replacing the wording everywhere keeps it green.

## Problem

All 22 starters carrying the `copilotkit:intelligence` marker shipped:

```ts
identifyUser: () => ({ id: "demo-user", name: "Demo User" }),
```

`premium/connect-your-runtime` — the page the onboarding graph teaches the runtime constructor from (`subagent/create-plan.md`, `credentials/finalize-plan.md`) — shows the opposite, on the one option that decides whether thread history is per-user or shared:

```ts
// Threads are per-user. Without this every visitor shares one history.
identifyUser: (request) => ({
  id: request.headers.get("x-user-id") ?? "anonymous",
  name: request.headers.get("x-user-name") ?? "Anonymous",
}),
```

So the artifact `copilotkit init` clones contradicted the page the graph teaches from. `ms-agent-framework-python` and `ms-agent-framework-dotnet` shipped the stub with **no warning comment at all**.

## What changed

- The documented header-derived form in all 22 sites. It degrades correctly for a demo — with no headers set it returns `anonymous` / `Anonymous`, one anonymous user and one history — and becomes correct the moment a real deployment sets the headers. The stub bought nothing that `?? "anonymous"` does not.
- The warning comment collapses to the docs page's own one-liner at all 22 sites. (When this PR was written the comments had drifted into five variants and two starters had none; #6716 has since unified them on the longest variant, so this is now a replacement of one uniform comment with another.)
- **`adk-angular`'s comment was factually wrong** and is deleted rather than normalised. It claimed the id "must correspond to a user that exists in CopilotKit Intelligence; an unknown id (like this literal) can make thread operations fail", and its README went further ("requires the identified user to actually exist", "the `copilotkit` CLI provisions one"). Neither holds: `resolve-intelligence-user.ts` validates **shape only** (`isValidIdentifier(user.id)` plus a non-empty name, 400 otherwise), and app-api takes `userId: z.string().trim().min(1)` and runs it through `scopedEndUserId(projectId, endUserId)` with no existence check and no unknown-user error anywhere. Had the claim been true, the documented page would be broken too, since `?? "anonymous"` is exactly as "unknown" as `demo-user`. The README bullet is rewritten accordingly.
- `scripts/__tests__/integration-intelligence-migration.test.ts` asserted the literal in 4 places; those now assert `request.headers.get("x-user-id")`, i.e. that identity is request-derived, which is the contract worth pinning.

## Verification

- **Contextual typing checked before touching 22 files.** `identifyUser` sits inside a conditional spread, so the unannotated `(request)` param could have been an implicit `any`. Isolated `tsc --strict` repro against the real `IdentifyUserCallback` signature: exit 0 for both the direct-property and conditional-spread shapes, with a control (`const f = (request) => request.headers`) confirming the check does catch TS7006. No annotation needed; matches the docs verbatim.
- **Non-Next hosts verified.** `adk-angular` (`createCopilotNodeListener`) and `agentcore` (Hono on Lambda) are not fetch-native, so the header read had to be confirmed reachable: `endpoints/express-fetch-bridge.ts:110` copies inbound headers into a real WHATWG `Request`.
- `oxfmt --check`: clean on all 23 changed TS files.
- `parity:check`: output **byte-identical** to the base commit.
- `check:intelligence-env-names`: green.
- `integration-intelligence-migration.test.ts`: 23 failed / 52 passed both before and after, with the failing test **names** diffed — `newly failing: NONE`, `newly passing: NONE`. Those 23 are pre-existing (stale version pins, files that no longer exist) and out of scope here; see below.

**Re-verified after rebasing onto current `main`** (now post-#6711, #6683, #6716, #6672, #6710).

The first rebase was conflict-free once the stale duplicate of the OSS-981 commit was dropped — the copy this branch carried was byte-identical in tree and patch to the `314f1ca55d` main merged, so nothing was lost. The second rebase hit real conflicts in all 21 comment-bearing sites, because #6716 rewrote the same lines; `git show` confirms #6716 touched **only** those comment lines in these files, so keeping this PR's side loses none of its work. One conflict block per file, asserted programmatically.

Gates on the new base:

- **`check:intelligence-wiring-block` (#6716's new gate): exit 0 — "All 22 Intelligence wiring sites match langgraph-python."** Its own unit test: 17/17 passing.
- `oxfmt --check`: clean on all 23 changed TS files.
- `parity:check`: exit 0, output **byte-identical** to `origin/main`.
- `check:intelligence-env-names`: exit 0.
- `integration-intelligence-migration.test.ts`: 23 failed / 52 passed on both base and head, failing test **names** diffed — newly failing NONE, newly passing NONE.

**The false-claim finding re-checked against current `main`,** since this PR now deletes wording #6716 deliberately standardised. `resolve-intelligence-user.ts` still validates shape only: `isValidIdentifier(user?.id)` (a `/^[\w.@:=-]+$/` charset plus a 128-char cap) and a non-empty `name`, 400 otherwise — no lookup, no existence check. A structural search for an existence path (`user not found`, `unknown user`, `verifyUser`, `ensureUser`, …) across `packages/runtime/src` and `packages/core/src` returns nothing.

**The `anonymous` fallback verified against the real validator,** since a rejected id would 400 the demo rather than degrade: `"anonymous"` → `true`, `"Anonymous"` → `true` under the actual `isValidIdentifier` predicate.

## Note on the untested test file

`integration-intelligence-migration.test.ts` is not run by CI — `static_quality.yml` runs exactly one `scripts/__tests__` file (`validate-dts-ambient.test.ts`). That is how 23 of its assertions rotted unnoticed. OSS-982 has since been closed by #6716, which added `check:intelligence-wiring-block` as a real gate; this stale file is still ungated and still out of scope here.

### DIFF
--- a/examples/integrations/a2a-a2ui/app/api/copilotkit/[[...slug]]/route.tsx
+++ b/examples/integrations/a2a-a2ui/app/api/copilotkit/[[...slug]]/route.tsx
@@ -97,11 +97,11 @@ const runtime = new CopilotRuntime({
             ? { wsUrl: process.env.INTELLIGENCE_GATEWAY_WS_URL }
             : {}),
         }),
-        // Demo stub — replace with your real auth-derived user identity before any
-        // multi-user deployment, or all users share one thread history. The id
-        // must correspond to a user that exists in CopilotKit Intelligence;
-        // an unknown id (like this literal) can make thread operations fail.
-        identifyUser: () => ({ id: "demo-user", name: "Demo User" }),
+        // Threads are per-user. Without this every visitor shares one history.
+        identifyUser: (request) => ({
+          id: request.headers.get("x-user-id") ?? "anonymous",
+          name: request.headers.get("x-user-name") ?? "Anonymous",
+        }),
         licenseToken: process.env.COPILOTKIT_LICENSE_TOKEN,
       }
     : { runner: new InMemoryAgentRunner() }),
--- a/examples/integrations/a2a-a2ui/app/api/copilotkit/[[...slug]]/route.tsx
+++ b/examples/integrations/a2a-a2ui/app/api/copilotkit/[[...slug]]/route.tsx
@@ -97,7 +97,9 @@ const runtime = new CopilotRuntime({
             ? { wsUrl: process.env.INTELLIGENCE_GATEWAY_WS_URL }
             : {}),
         }),
-        // Threads are per-user. Without this every visitor shares one history.
+        // Threads are per-user, so until your app sends x-user-id every
+        // visitor shares the "anonymous" history. Wire these headers to
+        // your auth-derived identity before any multi-user deployment.
         identifyUser: (request) => ({
           id: request.headers.get("x-user-id") ?? "anonymous",
           name: request.headers.get("x-user-name") ?? "Anonymous",
--- a/examples/integrations/a2a-middleware/app/api/copilotkit/[[...slug]]/route.ts
+++ b/examples/integrations/a2a-middleware/app/api/copilotkit/[[...slug]]/route.ts
@@ -23,11 +23,11 @@ const runtime = new CopilotRuntime({
             ? { wsUrl: process.env.INTELLIGENCE_GATEWAY_WS_URL }
             : {}),
         }),
-        // Demo stub — replace with your real auth-derived user identity before any
-        // multi-user deployment, or all users share one thread history. The id
-        // must correspond to a user that exists in CopilotKit Intelligence;
-        // an unknown id (like this literal) can make thread operations fail.
-        identifyUser: () => ({ id: "demo-user", name: "Demo User" }),
+        // Threads are per-user. Without this every visitor shares one history.
+        identifyUser: (request) => ({
+          id: request.headers.get("x-user-id") ?? "anonymous",
+          name: request.headers.get("x-user-name") ?? "Anonymous",
+        }),
         licenseToken: process.env.COPILOTKIT_LICENSE_TOKEN,
       }
     : { runner: new InMemoryAgentRunner() }),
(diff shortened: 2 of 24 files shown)


BenTaylorDev:
Closing unmerged on review feedback — the pattern this PR adopts is itself insecure, so shipping it to 22 starters would spread the problem rather than fix it.

`request.headers.get("x-user-id")` is client-controlled and nothing verifies it. `resolveIntelligenceUser` validates shape only (`isValidIdentifier` plus a non-empty name), and app-api scopes threads by that value via `scopedEndUserId(projectId, endUserId)` with no existence check — so a caller who sets the header to another user's id reads their thread history.

That undercuts the premise. The PR argued the starters were the only surface disagreeing with the library, the docs and the graph, and aligned them to `premium/connect-your-runtime.md`. But that page is not a correct reference for this option: it teaches a header echo that looks like auth. The `demo-user` stub is wrong too, but it at least reads as a stub.

The real fix belongs in `showcase/shell-docs/src/content/docs/premium/connect-your-runtime.mdx:59-63` — identity derived from something a client cannot forge, a verified JWT off the request or a server-read session. The starters should follow whatever that page settles on, not lead it.

Tracked as cancelled in OSS-987, with the reasoning and two side findings recorded there.

PR #6730, opened 2026-08-27
fix(docs): replace retired GitHub Models setup

MikeRyanDev:
## What does this PR do?

- Replaces the retired GitHub Models setup in the Microsoft Agent Framework .NET starter with direct OpenAI.
- Updates the live .NET guides to the current Agent Framework AG-UI hosting, session, state, and response APIs.
- Pins quickstart packages to the versions tested by the starter.
- Adds a recursive retired-guidance guard and runs it in docs CI for docs or starter-only changes.

GitHub Models was fully retired on July 30, 2026, so the published setup can no longer work.

## Related PRs and Issues

- Companion CLI cleanup: https://github.com/CopilotKit/Intelligence/pull/1029

## Validation

- `npm exec -- vitest run src/lib/__tests__/ms-agent-dotnet-provider.test.ts` (5 tests passed)
- `npm run typecheck`
- `npm run lint` (existing warnings only)
- `npm run build`
- `pnpm run validate:model-names`
- Workflow syntax and formatting checks passed.
- `docker build -f docker/Dockerfile.agent agent`
- `docker compose -f docker-compose.test.yml config`
- Compiled all nine full .NET guide examples against the pinned Agent Framework and AG-UI packages.
- TestServer proof returned `200 text/event-stream` and emitted the expected `STATE_SNAPSHOT` through `WithMetadata(streamOptions)`.
- Full shell-docs test run: 469 tests passed; one unrelated existing Mastra fixture test failed in `llm-text.test.ts`.

## Checklist

### DIFF
--- a/.github/workflows/test_integration-docs.yml
+++ b/.github/workflows/test_integration-docs.yml
@@ -5,6 +5,10 @@ on:
     paths:
       - "showcase/shell-docs/src/content/**"
       - "showcase/shell-docs/model-allowlist.json"
+      - "showcase/shell-docs/src/lib/__tests__/ms-agent-dotnet-provider.test.ts"
+      - "showcase/shell-docs/package.json"
+      - "showcase/shell-docs/package-lock.json"
+      - "examples/integrations/ms-agent-framework-dotnet/**"
       - "scripts/validate-doc-model-names.ts"
       - "scripts/doc-tests/**"
       - ".github/workflows/test_integration-docs.yml"
@@ -13,6 +17,10 @@ on:
     paths:
       - "showcase/shell-docs/src/content/**"
       - "showcase/shell-docs/model-allowlist.json"
+      - "showcase/shell-docs/src/lib/__tests__/ms-agent-dotnet-provider.test.ts"
+      - "showcase/shell-docs/package.json"
+      - "showcase/shell-docs/package-lock.json"
+      - "examples/integrations/ms-agent-framework-dotnet/**"
       - "scripts/validate-doc-model-names.ts"
       - "scripts/doc-tests/**"
       - ".github/workflows/test_integration-docs.yml"
@@ -40,6 +48,29 @@ jobs:
           cache: pnpm
       - run: pnpm install --frozen-lockfile
       - run: pnpm tsx scripts/validate-doc-model-names.ts
+
+  ms-agent-dotnet-guidance:
+    name: Microsoft Agent Framework .NET guidance
+    runs-on: depot-ubuntu-24.04-4
+    timeout-minutes: 15
+    permissions:
+      contents: read
+      id-token: write # zizmor: ignore[undocumented-permissions] Depot runner OIDC.
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+        with:
+          persist-credentials: false
+      - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
+        with:
+          node-version: 22
+          cache: npm
+          cache-dependency-path: showcase/shell-docs/package-lock.json
+      - name: Install shell-docs dependencies
+        working-directory: showcase/shell-docs
+        run: npm ci --ignore-scripts
+      - name: Check Microsoft Agent Framework .NET guidance
+        working-directory: showcase/shell-docs
+        run: npm exec -- vitest run src/lib/__tests__/ms-agent-dotnet-provider.test.ts
 
   doc-tests:
     runs-on: depot-ubuntu-24.04-4
--- a/examples/integrations/ms-agent-framework-dotnet/.env.example
+++ b/examples/integrations/ms-agent-framework-dotnet/.env.example
@@ -1,4 +1,4 @@
-# The C# agent reads GitHubToken from .NET user-secrets; see README.md.
+# The C# agent reads OPENAI_API_KEY from .NET user-secrets; see README.md.
 AGENT_URL=http://localhost:8000
 
 # --- CopilotKit Intelligence (optional; set COPILOTKIT_LICENSE_TOKEN to enable Threads — server + UI) ---
--- a/examples/integrations/ms-agent-framework-dotnet/README.md
+++ b/examples/integrations/ms-agent-framework-dotnet/README.md
@@ -4,9 +4,7 @@ This is a starter template for building AI agents using [Microsoft Agent Framewo
 
 ## Prerequisites
 
-- **GitHub Personal Access Token** (for GitHub Models API)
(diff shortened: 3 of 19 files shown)


BenTaylorDev:
Reviewed this as an API migration rather than a docs edit, since it rewrites nine .NET samples plus the starter. Verified the new API surface against the published package metadata rather than the PR description.

**Verified**

| Claim | Evidence |
| --- | --- |
| `AddAGUIServer` / `MapAGUIServer` | `AGUIServerServiceCollectionExtensions.AddAGUIServer`, `AGUIEndpointRouteBuilderExtensions.MapAGUIServer(IEndpointRouteBuilder, string, AIAgent)` in the 1.19.0-preview XML docs |
| `.AsAIAgent(...)` replacing `.CreateAIAgent` | `OpenAI.Chat.OpenAIChatClientExtensions.AsAIAgent(client, instructions, name, description, tools, …)` — it lives in namespace `OpenAI.Chat`, which is exactly why the new `using OpenAI.Chat;` is needed in every snippet |
| `TryGetRunAgentInput`, `AGUIStreamOptions.MapCall`, `StateSnapshotEvent.Snapshot`, `AGUIContext.Description` | present in `AGUI.Server` / `AGUI.Abstractions` 0.0.5, pulled in transitively |
| `AgentThread`→`AgentSession`, `AgentRunResponse`→`AgentResponse`, override→`RunCoreStreamingAsync` | confirmed in `Microsoft.Agents.AI.Abstractions` 1.19.0 |
| Dropping `AllowBackgroundResponses` / `ContinuationToken` in `SharedStateAgent.cs` | not a silent regression — both moved off `ChatClientAgentRunOptions` onto `ChatOptions`, which is cloned |
| `response.TryDeserialize` → manual try/catch | `TryDeserialize` no longer exists; the replacement is required |
| Pinning `OpenAI 2.10.0` | `Microsoft.Agents.AI.OpenAI 1.19.0` depends on exactly `OpenAI >= 2.10.0` — no NU1605 downgrade |
| net9.0 targeting | every package ships net9.0 (or net8.0-compatible) assets |
| New `using OpenAI.Chat;` collisions | only `ChatMessage` and `ChatResponseFormat` collide with `Microsoft.Extensions.AI`, and both are aliased in exactly the two files that use them |
| `gpt-5.4-mini` now resolving against api.openai.com | on `showcase/shell-docs/model-allowlist.json`; `validate-model-names` and `doc-tests` are green |

I also mutation-checked the new guard test rather than trusting it green: **46 offenders on current `main`, 0 with this branch applied**, and still 0 against a local merge with current `origin/main`. It is genuinely load-bearing.

And `smoke-starter (ms-agent-framework-dotnet)` passing is real mechanism coverage, not a mock — it Docker-builds the starter against the new packages and runs it through `OPENAI_BASE_URL`, which is the exact path the removed `sed` patch in `Dockerfile.agent` used to fake.

**Findings**

1. **Missed file — `examples/integrations/ms-agent-framework-dotnet/.env.example:1`.** It still reads `# The C# agent reads GitHubToken from .NET user-secrets; see README.md.` That's inside the very starter being fixed, and it isn't in the test's `guardedFiles`, so the new guard can't catch it. One-line edit plus one array entry.

2. **The guard is scoped wrong for half its targets.** The test lives in `showcase/shell-docs` but asserts on files under `examples/`, and nx's `namedInputs.test` is `{projectRoot}`-scoped. A change that reintroduces `GitHubToken` under `examples/integrations/ms-agent-framework-dotnet/**` alone will hit the shell-docs test cache and never rerun — the guard only really fires when shell-docs itself changes.

3. **Retirement fallout outside this PR** (follow-ups, not blockers): `skills/copilotkit-integrations/references/integrations/ms-agent-framework.md` still teaches GitHub Models; `showcase/integrations/ms-agent-dotnet` and `ms-agent-harness-dotnet` (`ApiKeyResolver.cs`, `Program.cs`, `.env.example`, `entrypoint.sh`) plus `.github/workflows/showcase_eval.yml` and `test_showcase-frontend-matrix.yml` still wire `GitHubToken`. Same retirement, same breakage.

4. **Scope note.** Five doc files moved from Azure OpenAI (`AzureOpenAIClient` + `DefaultAzureCredential`) to direct OpenAI. Those never used GitHub Models, and the change removes the only `DefaultAzureCredential` example in the MAF docs. Reasonable for consistency — flagging it as a conscious call rather than a bug.

5. **Minor.** The quickstart now says `dotnet add package … --prerelease` (unpinned) while the starter pins `1.19.0-preview.260822.1`, so quickstart followers get whatever preview is newest rather than the version the samples were validated against.

**The one thing worth confirming before merge**

`app.MapAGUIServer("/", agent).WithMetadata(streamOptions)` in `predictive-state-updates.mdx`. It compiles — `MapAGUIServer` returns `IEndpointConventionBuilder` — and the hosting assembly does reference both `AGUIStreamOptions` and `GetMetadata`/`EndpointMetadataCollection`. But nothing in the public XML documents endpoint metadata as the registration channel, and no smoke test covers that snippet. It's the only sample whose runtime wiring I couldn't prove statically.

MikeRyanDev:
Addressed the actionable findings at `593fe0bc`.

1. `.env.example` now names `OPENAI_API_KEY` and the retired-guidance guard scans that file.
2. `shell-docs` is outside the pnpm workspace and is not an Nx project, so changing Nx inputs would not close the gap. The focused guard now runs in `test / integration / docs` through the standalone npm lock. Docs changes and starter-only changes both trigger it.
3. The skill, showcase variants, and their workflows remain the separate follow-up scope noted in the review; this docs/starter PR does not claim to retire those runnable surfaces.
4. The direct OpenAI path is intentional so the runnable starter and complete guide snippets share one credential flow. The quickstart still links to Azure OpenAI and the other Agent Framework model providers.
5. Both quickstart package commands now use the exact versions pinned and tested by the starter. The guard derives the expected versions from the csproj.

`WithMetadata(streamOptions)` is also confirmed as the correct runtime path. The [1.19 handler reads endpoint metadata before service options](https://github.com/microsoft/agent-framework/blob/d9d3fb6252f7ae9e7f8104edce7266f0782a813c/dotnet/src/Microsoft.Agents.AI.Hosting.AGUI.AspNetCore/AGUIEndpointRouteBuilderExtensions.cs#L125-L140), and the [Microsoft state-management sample uses the same registration](https://github.com/microsoft/agent-framework/blob/d9d3fb6252f7ae9e7f8104edce7266f0782a813c/dotnet/samples/02-agents/AGUI/Step05_StateManagement/Server/Program.cs#L81-L83). A TestServer request against hosting package `1.19.0-preview.260822.1` returned `200 text/event-stream` and emitted the expected `STATE_SNAPSHOT`.

Exact-head validation passed: the 5 focused guard tests, shell-docs lint/typecheck/build, model-name validation, workflow syntax checks, formatting, and diff checks.

BenTaylorDev (approved):

PR #6732, opened 2026-08-27
docs(shell-docs): fix broken A2UI doc links in overview

moonturbo:
Fix 6 broken internal links in the A2UI overview page. The file lives at docs/generative-ui/a2ui/index.mdx, so links should point to ./dynamic-schema and ./fixed-schema instead of the non-existent ./a2ui/* path.

### DIFF
--- a/showcase/shell-docs/src/content/docs/generative-ui/a2ui/index.mdx
+++ b/showcase/shell-docs/src/content/docs/generative-ui/a2ui/index.mdx
@@ -24,11 +24,11 @@ You can design and preview A2UI schemas visually using the
 
 CopilotKit ships two complementary A2UI approaches:
 
-- **[Dynamic Schema](./a2ui/dynamic-schema)**: a secondary LLM
+- **[Dynamic Schema](./dynamic-schema)**: a secondary LLM
   generates both the schema *and* the data. Maximum flexibility, so
   the agent can produce any UI for any prompt.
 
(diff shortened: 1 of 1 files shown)


BenTaylorDev:
Closing this one, and I want to give you the reason in full because the assumption behind it is a reasonable one that this repo happens to violate.

Relative links in these MDX files are **not** resolved against the file path. The docs app maps `a` to fumadocs' default `Link`, which passes `href` straight through to `next/link`; fumadocs' opt-in resolver (`createRelativeLink`) isn't used anywhere in the app. So the raw string reaches the DOM and the browser resolves it against the current URL — and with no `trailingSlash` config, a docs URL has no trailing slash, so `./x` drops the last segment.

This page is served at `/generative-ui/a2ui`, so:

- `./a2ui/dynamic-schema` (current) → `/generative-ui/a2ui/dynamic-schema` → **200**
- `./dynamic-schema` (this PR) → `/generative-ui/dynamic-schema` → **404**

I verified both against the live site. The six links you changed are the ones that currently work, so this would break them.

The confusing part, and the reason I don't think this reflects badly on the diagnosis: the flat `./sibling` form you moved toward *is* correct in non-index pages, and several `index.mdx` files in this repo use it incorrectly — including the two you touched in #6733. This page is the one place that got it right. There are 7 genuinely broken links of that kind across 3 index pages, and I'm fixing those separately.

Worth knowing for future docs PRs here: CI has no internal-link checker, so green checks say nothing about whether a link resolves.

PR #6733, opened 2026-08-27
docs(shell-docs): fix broken Fixed Schema (Streaming) link

moonturbo:
The A2UI overview pages for both the DeepAgents and LangGraph integrations link to ./fixed-schema-streaming, but that file does not exist (only dvanced, dynamic-schema, ixed-schema, index, styling exist). Point both links to the existing ./fixed-schema page.


BenTaylorDev (approved):
Verified against current `main`: no `fixed-schema-streaming.mdx` exists in either the deepagents or langgraph `generative-ui/a2ui/` directory, neither `meta.json` lists such a page, and `./fixed-schema` does exist in both — so both links resolve after this change and two 404s are gone. These two lines are the only remaining references to the missing page; the `advanced.mdx` streaming links were already corrected on main. Diff is exactly two lines with no unrelated debris.

Thanks for the clean, minimal scoping.

PR #6738, opened 2026-08-27
feat(showcase): MAF Python A2UI error-recovery demo on agent-framework 1.2.0

ranst91:
## What

Brings the just-released A2UI support and the latest Microsoft Agent Framework
(Python) into the showcase, and moves the MAF A2UI demos onto the native
subagent / auto-inject technique so they match the langgraph-python reference
instead of the pre-1.2.0 hand-rolled path.

- **Bump MAF to 1.2.0.** `agent-framework-ag-ui[a2ui]==1.2.0`,
  `agent-framework-openai==1.14.0`, `agent-framework-core==1.15.0`,
  `ag-ui-a2ui-toolkit==0.0.4`. 1.2.0 is A2UI's first release; the `[a2ui]` extra
  pulls the toolkit.

- **New `a2ui-recovery` demo** (the one A2UI demo the integration lacked vs
  langgraph-python). Backend-owned via the adapter's native `enable_a2ui`
  (`injectA2UITool: false`), which runs the shared toolkit's validate/retry
  recovery loop in-process. Heal pill recovers a malformed first render; exhaust
  pill hits the attempt cap and surfaces the `a2ui_recovery_exhausted` fallback.
  Adds the agent, route, page, chat, suggestions, a D6 fixture, and an e2e spec.

- **Migrate `declarative-gen-ui` to native A2UI auto-injection.** Removed the
  hand-rolled `generate_a2ui` (a raw secondary OpenAI call to
  `_design_a2ui_surface`); the agent now binds no A2UI tool and the route sets
  `injectA2UITool: true`, so the adapter's `plan_a2ui_injection` auto-injects the
  native `generate_a2ui` sub-agent. Reworked the D6 fixture to the native
  `render_a2ui` shape.

- **Migrate `beautiful-chat` to native A2UI auto-injection.** Same change for the
  flagship composite: removed its hand-rolled `generate_a2ui`, flipped the route
  to `injectA2UITool: true`. The adapter now auto-injects the native sub-agent
  alongside the agent's own tools (todos, query, flights).

- **`a2ui-fixed-schema` unchanged** — it is the fixed-schema pattern (client
  authored schema, agent streams data via a backend `display` tool,
  `injectA2UITool: false`), which langgraph-python does identically. Not
  hand-rolled generation.

- **Docs.** Enriched the MAF A2UI docs page into real "how to A2UI" content
  (dynamic / fixed / recovery) and connected the A2UI demos via `docs-links`.

## Why native (not the dojo's example agents)

The showcase mirrors langgraph-python's frontend-catalog + auto-inject pattern,
not the AG-UI dojo's `a2ui_config` example agents. Before 1.2.0 the MAF adapter
had no native A2UI, so the showcase hand-rolled `generate_a2ui`. 1.2.0 ships the
native path, so these demos now use it and match langgraph 1:1.

## Validation

- `validate-pins` clean (FAIL count + hash unchanged), `validate-parity` PASS,
  `generate-registry` clean.
- **D6 (full frontend, aimock replay), all green:**
  - `a2ui-recovery` — heal paints the recovered surface; exhaust shows the
    hard-failure UI.
  - `declarative-gen-ui` — all 4 dashboard pills paint on the auto-inject path
    (confirms the runtime forwards `injectA2UITool: true` and
    `plan_a2ui_injection` fires).
  - `beautiful-chat` — regression across all 5 features (pie/bar chart,
    schedule-meeting, search-flights, toggle-theme): wrapping the multi-tool
    agent in the A2UI planner loop does not break its non-A2UI tools.
- **AG-UI protocol layer** (published 1.2.0 wheel + aimock): recovery
  (heal/exhaust), declarative, and beautiful-chat all stream real
  `a2ui_operations` / `a2ui_recovery_exhausted`, RUN_FINISHED, no RUN_ERROR.

- **Remove the last hand-rolled A2UI (default agent).** The general-purpose
  default agent (`agent.py`, catch-all `/` endpoint) also carried a hand-rolled
  `generate_a2ui`; removed it (the default agent no longer offers A2UI, matching
  langgraph's default agent). Stripped the stale `_design_a2ui_surface` fixture
  residue and refreshed the e2e-spec comments that described the old mechanism.

After this PR the MAF-python integration has **zero hand-rolled A2UI anywhere
except the fixed-schema demo** (which is the intended fixed-schema pattern,
identical to langgraph). The shared `tools/generate_a2ui.py` module is
intentionally left intact — it is symlinked by other integrations (ag2, agno, …)
that have not migrated; MAF-python simply no longer imports it.

## Pre-existing, out of scope

Full D6 for ms-agent-python is 37/40. The 3 red cells — `multimodal`, `voice`,
`hitl-approve-deny` — are **not** touched by this PR: their agents are
byte-identical to main and never used `generate_a2ui`. `multimodal` is the known
shared-CopilotKit frontend bug (`runStartCount=0`, the run never starts);
`voice`/`hitl-approve-deny` complete the run but their text does not settle inside
the probe's tight budget. All A2UI and default-agent cells pass.


contextablemark (changes requested):
**Changes requested:** I think this should probably use MAF’s auto-injection path. As far as I can tell, putting the recovery settings and catalog ID in `a2ui_config` gives us the same backend-owned policy and recovery behavior, while matching the other MAF demos and MAF’s upstream recovery example.

There may be a non-obvious constraint I’m missing—such as needing a separate subagent client, an attempt callback, or injection independent of the forwarded flag. If so, keeping `enable_a2ui()` makes sense; could we document that reason? Otherwise, I’d prefer switching this to auto-injection.

showcase/integrations/ms-agent-python/src/agents/recovery_agent.py
```diff
@@ -0,0 +1,80 @@
...
+    )
+
+    return AgentFrameworkAgent(
+        agent=base_agent,
+        name="CopilotKitMicrosoftAgentFrameworkAgent",
+        description=(
+            "A2UI generator with the validate->retry recovery loop made visible."
+        ),
+        require_confirmation=False,
+    )
```
contextablemark: MAF applies endpoint `a2ui_config` only when it creates the wrapper. Keeping this agent raw follows the upstream recovery example and makes the existing endpoint config effective.

```suggestion
from agent_framework import Agent, BaseChatClient

CUSTOM_CATALOG_ID = "declarative-gen-ui-catalog"

# The recovery cap + catalog live in the endpoint `a2ui_config` (agent_server.py),
# consumed by auto-injection — not on the agent. Keep this exported so the server
# and the config stay in sync.
A2UI_RECOVERY_CONFIG = {
    # `maxAttempts` pins the renderer's "Retrying… (N/M)" label to the adapter's cap.
    "recovery": {"maxAttempts": 3},
    "default_catalog_id": CUSTOM_CATALOG_ID,
}

# Keep this aligned with the declarative-gen-ui SYSTEM_PROMPT: a demo assistant
# that answers every question by drawing a surface. `generate_a2ui` (auto-injected
# by the adapter) handles the rendering — and its automatic recovery — internally.
SYSTEM_PROMPT = (
    "You are a demo assistant for A2UI Error Recovery. Whenever a response "
    "would benefit from a rich visual — a dashboard, KPI summary, status "
    "report, or card layout — call `generate_a2ui` to draw it. The registered "
    "catalog includes `Card`, `StatusBadge`, `Metric`, `InfoRow`, "
    "`PrimaryButton`, `PieChart`, and `BarChart` (in addition to the basic "
    "A2UI primitives). `generate_a2ui` takes no arguments and handles the "
    "rendering — and its automatic recovery from a malformed first attempt — "
    "for you. Keep chat replies to one short sentence; let the UI do the "
    "talking."
)


def create_agent(chat_client: BaseChatClient) -> Agent:
    """Instantiate the MS-Agent-backed A2UI error-recovery agent.

    The agent binds no A2UI tool: the route forwards `injectA2UITool: true` and
    the endpoint's `a2ui_config` (`A2UI_RECOVERY_CONFIG`) carries the recovery cap
    + catalog, so the adapter auto-injects `generate_a2ui` and runs the recovery
    loop. `chat_client` is a Chat-Completions client (`OpenAIChatCompletionClient`),
    reused as the render sub-agent so its `render_a2ui` arguments stream progressively.
    """
    # Keep this raw: the endpoint applies `a2ui_config` while creating its
    # `AgentFrameworkAgent` wrapper.
    return Agent(
        client=chat_client,
        name="a2ui_recovery_agent",
        instructions=SYSTEM_PROMPT,
    )
```

showcase/integrations/ms-agent-python/manifest.yaml
```diff
@@ -369,6 +370,18 @@ demos:
       - src/app/demos/a2ui-fixed-schema/a2ui/definitions.ts
       - src/app/demos/a2ui-fixed-schema/a2ui/renderers.tsx
       - src/app/api/copilotkit/route.ts
+  - id: a2ui-recovery
+    name: A2UI Error Recovery
+    description: Makes the A2UI validate->retry recovery loop visible — an invalid first render heals to a valid one, and an always-invalid render shows a graceful recovery-exhausted fallback. Backend-owned via the adapter's native enable_a2ui (injectA2UITool=false); reuses the declarative-gen-ui catalog.
```
contextablemark: This public description still names the old explicit path.

```suggestion
    description: Makes the A2UI validate->retry recovery loop visible — an invalid first render heals to a valid one, and an always-invalid render shows a graceful recovery-exhausted fallback. The adapter auto-injects the tool from backend a2ui_config (injectA2UITool=true) and reuses the declarative-gen-ui catalog.
```

showcase/integrations/ms-agent-python/src/app/demos/a2ui-recovery/page.tsx
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * A2UI Error Recovery demo.
+ *
+ * Same dynamic-schema A2UI setup as declarative-gen-ui (it reuses that demo's
+ * catalog), but it makes the toolkit's validate->retry recovery loop visible.
+ * The dedicated runtime at `/api/copilotkit-a2ui-recovery` is configured with
+ * `injectA2UITool: false` — the backend agent (`src/agents/recovery_agent.py`)
+ * owns `generate_a2ui` via the adapter's native `enable_a2ui`, whose body runs
+ * the forced `render_a2ui` sub-agent and the recovery loop + recovery-exhausted
+ * hard-fail envelope in-process (OSS-158 / OSS-375).
```
contextablemark: Same cleanup in the demo’s source description.

```suggestion
 * Same native auto-injection setup as declarative-gen-ui (it reuses that demo's
 * catalog), but it makes the toolkit's validate->retry recovery loop visible.
 * The dedicated runtime at `/api/copilotkit-a2ui-recovery` sets
 * `injectA2UITool: true`; the adapter injects `generate_a2ui` and reads the
 * recovery cap + catalog from the endpoint's `a2ui_config`. The generated tool
 * runs the forced `render_a2ui` sub-agent and the recovery loop +
 * recovery-exhausted hard-fail envelope in-process (OSS-158 / OSS-375).
```

showcase/integrations/ms-agent-python/tests/e2e/a2ui-recovery.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+//
+// The aimock fixtures (showcase/aimock/d6/ms-agent-python/a2ui-recovery.json)
+// drive the inner render_a2ui sub-agent two ways: HEAL is structurally invalid
+// on the first attempt (root references a missing child) then valid on the
+// second; EXHAUST is invalid on every attempt, so the validate->retry loop hits
+// the cap and returns the a2ui_recovery_exhausted envelope. Healing + the loop
+// run live in the backend agent-framework-ag-ui `enable_a2ui` recovery loop
+// (injectA2UITool=false); the failure UI text ("Couldn't generate the UI")
+// comes from @copilotkit/react-core/v2. The recovered surface reuses the
+// declarative-gen-ui catalog, so it carries the `declarative-metric` testid.
```
contextablemark: And in the E2E contract.

```suggestion
// The aimock fixtures (showcase/aimock/d6/ms-agent-python/a2ui-recovery.json)
// drive the inner render_a2ui sub-agent two ways: HEAL is structurally invalid
// on the first attempt (root references a missing child) then valid on the
// second; EXHAUST is invalid on every attempt, so the validate->retry loop hits
// the cap and returns the a2ui_recovery_exhausted envelope. Healing + the loop
// run live through the adapter's auto-injected `generate_a2ui` path
// (`injectA2UITool=true`, with backend policy in `a2ui_config`); the failure UI
// text ("Couldn't generate the UI") comes from @copilotkit/react-core/v2. The
// recovered surface reuses the declarative-gen-ui catalog, so it carries the
// `declarative-metric` testid.
```

showcase/aimock/d6/ms-agent-python/a2ui-recovery.json
```diff
@@ -0,0 +1,114 @@
...
...where: a real browser does NOT send the x-aimock-context header (only the test harness does), so context-scoped fixtures would 404 in-browser. The recovery pills' prompts are UNIQUE per framework (distinct from langgraph-python's), so userMessage alone disambiguates without needing context. Flow per pill: the render sub-agent runs with messages [system(base_prompt), *conversation] (agent_framework_ag_ui _agent.py _subagent_messages), so the inner render_a2ui request carries the ORIGINAL user prompt as its last user message; the retry augments the SYSTEM prompt only, so userMessage stays constant across attempts and sequenceIndex 0/1 disambiguates them. The inner-render fixtures are listed FIRST and carry toolName=render_a2ui so they win for the inner calls; the tool-less outer generate_a2ui emit falls through to the userMessage-only fixture; the outer narration is matched by the unique toolCallId of the emit (its balancing tool result). HEAL: attempt 1 (seq 0) invalid (dangling child) -> retry -> attempt 2 (seq 1) valid -> paints (declarative-metric x2). EXHAUST: invalid on every attempt -> recovery exhausts -> a2ui_recovery_exhausted. Requires agent-framework-ag-ui[a2ui]>=1.2.0.",
```
contextablemark: And in the fixture metadata future debugging will rely on.

```suggestion
  "_note": "Native auto-injected A2UI (injectA2UITool=true; recovery policy in endpoint a2ui_config) with the shared toolkit validate->retry recovery loop made visible. NO `context` match field anywhere: a real browser does NOT send the x-aimock-context header (only the test harness does), so context-scoped fixtures would 404 in-browser. The recovery pills' prompts are UNIQUE per framework (distinct from langgraph-python's), so userMessage alone disambiguates without needing context. Flow per pill: the render sub-agent runs with messages [system(base_prompt), *conversation] (agent_framework_ag_ui _agent.py _subagent_messages), so the inner render_a2ui request carries the ORIGINAL user prompt as its last user message; the retry augments the SYSTEM prompt only, so userMessage stays constant across attempts and sequenceIndex 0/1 disambiguates them. The inner-render fixtures are listed FIRST and carry toolName=render_a2ui so they win for the inner calls; the tool-less outer generate_a2ui emit falls through to the userMessage-only fixture; the outer narration is matched by the unique toolCallId of the emit (its balancing tool result). HEAL: attempt 1 (seq 0) invalid (dangling child) -> retry -> attempt 2 (seq 1) valid -> paints (declarative-metric x2). EXHAUST: invalid on every attempt -> recovery exhausts -> a2ui_recovery_exhausted. Requires agent-framework-ag-ui[a2ui]>=1.2.0.",
```

contextablemark (changes requested):
The auto-injection direction looks right, and the focused D6 check is green. One MAF wrinkle remains: endpoint config is only applied while wrapping a raw agent. Because this factory still returns an already-wrapped `AgentFrameworkAgent`, its `a2ui_config` is currently dropped and the demo is succeeding on toolkit defaults.

I left batchable suggestions to keep this recovery agent raw so the endpoint applies its config, and to update the remaining old-path descriptions. After those, this looks good to me.

contextablemark (approved):
The recovery path is correctly wired now, and both the heal and exhaust cases pass on the latest head. The few stale auto-injection comments elsewhere are cleanup-only, not blockers. Looks good to me.

PR #6754, opened 2026-08-28
feat(runtime): truncate oversized application-context values

zzt417520:
### What does this PR do?

When an app shares context with an agent via `useAgentContext`, each context
value is currently injected verbatim into the assembled system prompt. Large
values (a full document, a big JSON blob) can blow the context window and
inflate token cost.

This change clamps each context value to `MAX_CONTEXT_VALUE_LENGTH` (20,000
chars) before it is added to the system prompt, appending a marker
(`… [truncated by CopilotKit]`) so the model knows the content was cut off
rather than simply absent. Within-limit values are left unchanged.

### Changes

- `packages/runtime/src/agent/index.ts` — added `MAX_CONTEXT_VALUE_LENGTH` +
  `truncateContextValue()`, applied when building the system prompt.
- `packages/runtime/src/agent/__tests__/context-truncation.test.ts` — new tests
  for truncation and pass-through.
- `.changeset/context-truncation.md` — patch entry.

### Verification

- `nx run @copilotkit/runtime:test` — 145 test files / 2086 tests passed.
- oxfmt + oxlint — 0 errors (pre-existing `no-shadow` warnings only).

### Checklist

- [x] Change is non-breaking and limited in scope
- [x] Tests added/updated for the new behavior

## Summary by CodeRabbit

* **Bug Fixes**
  * Limited oversized application context values to 20,000 characters before inclusion in agent system prompts.
  * Added a clear truncation marker when context values exceed the limit.
  * Applied consistent context truncation across classic agent runs and TanStack prompt generation.
  * Preserved context values that remain within the allowed size.
  * Ensured truncation does not split surrogate pairs or create malformed characters.

### DIFF
--- a/.changeset/context-truncation.md
+++ b/.changeset/context-truncation.md
@@ -1,0 +1,5 @@
+---
+"@copilotkit/runtime": patch
+---
+
+Truncate oversized application-context values (to 20k chars) when assembling the system prompt, preventing context bloat when large data is shared with the agent.
--- a/.changeset/context-truncation.md
+++ b/.changeset/context-truncation.md
@@ -1,5 +1,0 @@
----
-"@copilotkit/runtime": patch
----
-
-Truncate oversized application-context values (to 20k chars) when assembling the system prompt, preventing context bloat when large data is shared with the agent.
--- a/packages/runtime/src/agent/__tests__/context-truncation.test.ts
+++ b/packages/runtime/src/agent/__tests__/context-truncation.test.ts
@@ -1,5 +1,9 @@
 import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
 import { BasicAgent } from "../index";
+import {
+  MAX_CONTEXT_VALUE_LENGTH,
+  truncateContextValue,
+} from "../context-truncation";
 import { streamText } from "ai";
 import { mockStreamTextResponse, finish, collectEvents } from "./test-helpers";
 import type { RunAgentInput } from "@ag-ui/client";
@@ -77,5 +81,34 @@ describe("BuiltInAgent Application Context Truncation", () => {
     const content = callArgs.messages[0].content as string;
     expect(content).toContain("hello");
     expect(content).not.toContain("truncated");
+  });
+});
+
+describe("truncateContextValue", () => {
+  it("keeps the return value (including marker) within MAX_CONTEXT_VALUE_LENGTH", () => {
+    const result = truncateContextValue("x".repeat(50_000));
+    expect(result.length).toBe(MAX_CONTEXT_VALUE_LENGTH);
+    expect(result).toContain("… [truncated by CopilotKit]");
+  });
+
+  it("leaves a value at exactly the limit untouched (no marker)", () => {
+    const value = "x".repeat(MAX_CONTEXT_VALUE_LENGTH);
+    expect(truncateContextValue(value)).toBe(value);
+  });
+
+  it("truncates a value just over the limit to exactly the limit", () => {
+    const value = "x".repeat(MAX_CONTEXT_VALUE_LENGTH + 1);
+    const result = truncateContextValue(value);
+    expect(result.length).toBe(MAX_CONTEXT_VALUE_LENGTH);
+    expect(result).toContain("… [truncated by CopilotKit]");
+  });
+
+  it("does not end on a lone high surrogate when the slice lands mid-pair", () => {
+    // 0xd83d is a high surrogate; "a" + emoji repeated lands a lone high
+    // surrogate at the cut point. The result must not end with an unpaired one.
+    const value = "a" + "😀".repeat(20_000);
+    const result = truncateContextValue(value);
+    const last = result.charCodeAt(result.length - 1);
+    expect(last < 0xd800 || last > 0xdbff).toBe(true);
   });
 });
--- a/packages/runtime/src/agent/__tests__/context-truncation.test.ts
+++ b/packages/runtime/src/agent/__tests__/context-truncation.test.ts
@@ -1,0 +1,81 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
+import { BasicAgent } from "../index";
+import { streamText } from "ai";
+import { mockStreamTextResponse, finish, collectEvents } from "./test-helpers";
+import type { RunAgentInput } from "@ag-ui/client";
+
+vi.mock("ai", () => ({
+  streamText: vi.fn(),
+  tool: vi.fn((config) => config),
+  stepCountIs: vi.fn((count: number) => ({ type: "stepCount", count })),
+}));
+
+vi.mock("@ai-sdk/openai", () => ({
+  createOpenAI: vi.fn(() => (modelId: string) => ({
+    modelId,
+    provider: "openai",
+  })),
+}));
+
+describe("BuiltInAgent Application Context Truncation", () => {
+  const originalEnv = process.env;
+
+  beforeEach(() => {
+    vi.clearAllMocks();
+    process.env = { ...originalEnv };
+    process.env.OPENAI_API_KEY = "test-key";
+  });
+
+  afterEach(() => {
+    process.env = originalEnv;
+  });
+
+  it("truncates oversized context values in the assembled system prompt", async () => {
+    const agent = new BasicAgent({ model: "openai/gpt-4o" });
+    vi.mocked(streamText).mockReturnValue(
+      mockStreamTextResponse([finish()]) as any,
+    );
+
+    const input: RunAgentInput = {
+      threadId: "thread1",
+      runId: "run1",
+      messages: [],
+      tools: [],
+      context: [{ description: "big", value: "x".repeat(50_000) }],
+      state: {},
+    };
+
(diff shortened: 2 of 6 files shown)


BenTaylorDev (changes requested):
Thanks for this — the underlying concern is real. A `useAgentContext` value does go into the system prompt verbatim, and a large document or JSON blob can meaningfully inflate token cost. The truncation marker is a nice touch too: telling the model the content was cut beats letting it silently reason over a half-value.

Two things to fix, and two smaller notes.

## 1. It only covers one of the two injection sites

There are exactly two places that interpolate `ctx.value` into a system prompt:

```
$ git grep -n "ctx.value" -- packages/runtime/src
packages/runtime/src/agent/converters/tanstack.ts:293:      systemPrompts.push(`${ctx.description}:\n${ctx.value}`);
packages/runtime/src/agent/index.ts:1055:            parts.push(`${ctx.description}:\n${ctx.value}\n`);
```

This patches `agent/index.ts:1055` but not `agent/converters/tanstack.ts:293`, so whether a context value gets clamped depends on which path the app is on. That's harder to reason about than no truncation at all — someone debugging an oversized prompt would see the limit apply on one converter and not the other, with nothing explaining why. Applying `truncateContextValue` at both sites fixes it.

## 2. The changeset file fails CI

`static / check-binaries` is the one red check on this branch (39 others pass):

```
Changeset files detected in PR:
.changeset/context-truncation.md
This repo no longer uses Changesets — releases are driven by conventional commit subjects (see scripts/release/).
Nothing reads .changeset/*.md. Delete these files and describe the change in your commit subject instead.
```

Dropping `.changeset/context-truncation.md` clears it. Your commit subject already describes the change well.

## Two smaller notes

- **The limit is hardcoded.** Some applications legitimately want to hand the model a large document, and there's currently no way to raise the ceiling or opt out.
- **Truncation is invisible to the developer.** The marker tells the *model* that content was cut, but nothing tells the person who wrote the code — no warning, no telemetry. They'd find out by noticing the model doesn't know something it should. A `console.warn` on first truncation would make it debuggable.

The first two are what's needed for this to land; the notes are worth a thought but I don't want to expand the scope of a tidy change.

zzt417520:
Thanks for the detailed review — both points are addressed.

Truncation now applies at both injection sites. I extracted MAX_CONTEXT_VALUE_LENGTH / truncateContextValue() into a shared module (packages/runtime/src/agent/context-truncation.ts) and applied it in the tanstack input converter too, so ctx.value is clamped consistently regardless of which path the app is on. Added a matching test for the tanstack converter path.

Removed .changeset/context-truncation.md — the change is described in the commit subject instead.

Verification: nx run @copilotkit/runtime:test — 145 files / 2087 tests pass; oxfmt + oxlint clean; tsc --noEmit clean.

On the two notes — agreed they're worth considering (configurable limit, a one-time warning when truncation actually happens). I kept this change scoped as suggested; happy to follow up with either in a separate PR if you'd like.

BenTaylorDev (dismissed):
Both asks from the last round are addressed, and I verified each one.

**Coverage.** `truncateContextValue` now runs at both injection sites, and pulling
the constant into `agent/context-truncation.ts` is the right shape — one limit,
one marker, no chance of the two paths drifting apart. `git grep "ctx.value"` on
current `main` still returns exactly those two call sites, so this is complete.

**Changeset.** Gone. Your commit subjects carry the change instead.

**Verification I ran locally.** CI has not run on `eb3bf9b9` — fork PRs need a
maintainer to release the workflow gate — so I checked by hand:

- Full `packages/runtime` suite at the PR head vs its merge-base: identical
  failure sets, zero regressions.
- Mutation check: replacing `truncateContextValue` with the identity function
  fails both new tests. They test the mechanism, not themselves.
- `tsc --noEmit`, `oxfmt --check`, `oxlint`: nothing attributable to your files.

**One thing left, and it is the only thing blocking the merge button:** the branch
no longer merges into `main`. The conflict is just the import block at the top of
`agent/index.ts` — `main` picked up `getAISDKRunFinishedDetails` from
`./converters/aisdk` after you branched, so both sides edited the same import.
The truncation hunks themselves auto-merge cleanly. A rebase onto `main` and
keeping both imports resolves it.

I am approving now so nothing waits on me. Pushing the rebase will dismiss this
approval automatically — that is expected, not a problem. Ping me when it is up
and I will re-approve.

**Two nits, both fine as follow-ups:**

- `slice(0, MAX_CONTEXT_VALUE_LENGTH)` can cut a UTF-16 surrogate pair in half.
  `("a" + "😀".repeat(20000)).slice(0, 20000)` ends in a lone high surrogate, and
  `JSON.stringify` then puts a bare `\ud83d` in the request body, which stricter
  providers reject. Trimming a trailing high surrogate before appending the
  marker is a one-liner.
- The clamp is per value, not aggregate — fifty values of 19,999 characters each
  still blow the window. That is a reasonable place to stop for this PR; I just
  want it written down.

Thanks for the clean turnaround on both points.

zzt417520:
Rebased onto the latest main and resolved the import conflict in agent/index.ts — kept both convertAISDKStream/getAISDKRunFinishedDetails (new on main) and truncateContextValue.

I also took the boundary suggestion: truncateContextValue now reserves the marker length so the return stays at exactly 20,000 chars, and it trims a trailing lone surrogate before appending the marker. Added dedicated boundary tests for the 20,000 / 20,001-char cases plus the surrogate case — all 22 tests pass and lint is clean.

Ready for re-review when you get a chance, thanks!

BenTaylorDev:
Thanks for this, and for the follow-through on the last round — the rebase, the
shared constant, and the marker-budget fix all landed exactly as asked. I verified
each one: both injection sites call `truncateContextValue`, the full
`packages/runtime` suite shows zero regressions against your merge-base, and
reverting the `- TRUNCATION_MARKER.length` reservation fails two tests, so
CodeRabbit's ask is genuinely closed.

I am going to close this, though, and the reason is the design rather than the
code.

**The adjacent, usually-larger input is uncapped.** `input.state` is
JSON-serialized into the same system prompt block, at both of the sites you
patched:

- `packages/runtime/src/agent/index.ts:1070-1075`
- `packages/runtime/src/agent/converters/tanstack.ts:305-312`

In shared-state applications `state` is the large payload — the document, the
sheet, the research report. Context values are usually short labels like "current
page" or "selected row". So this caps the smaller of two neighbours and leaves the
larger one unbounded, which means the stated goal of protecting the context window
is not reached.

Three things compound that:

- **Characters are not tokens.** 20,000 characters is roughly 5,000 tokens of
  English and roughly 15,000 to 30,000 tokens of CJK (60,000 UTF-8 bytes). The cap
  is not a budget.
- **The cap is per value, not per prompt.** Sixty context values of 19,999
  characters each all pass untouched: about 1.2M characters, roughly 300,000
  tokens.
- **External frameworks are not covered.** `IntelligenceAgent` forwards
  `input.context` verbatim over HTTP (`packages/core/src/intelligence-agent.ts:470`),
  and `BuiltInAgent` is the only prompt-assembling agent in the runtime. LangGraph,
  CrewAI, Mastra, and ADK users would see no change.

Against that, the cost is real. `convertInputToTanStackAI` is a public documented
helper that users call inside their own agent factories, so this makes silent,
unconfigurable, undocumented data loss part of its contract. It also adds a crash
surface: a `null`, `undefined`, or object `value` previously stringified into the
prompt and now throws a `TypeError`, and `JSON.stringify(undefined)` returns
`undefined`, so the React hook can emit a non-string.

One more note for the record, because it matters more than the outcome here: the
surrogate-pair test at `context-truncation.test.ts:106-112` cannot fail. It reads
`result.charCodeAt(result.length - 1)`, which is always `]` from the truncation
marker, so it never inspects the cut point. I deleted the surrogate-trim block
entirely and all six tests still passed. Asserting across the whole result catches
it:

```ts
expect(/[\ud800-\udbff](?![\udc00-\udfff])/.test(result)).toBe(false);
```

That version fails without the trim and passes with it. Worth keeping in mind on
future tests — it is an easy shape to write by accident.

If you want to pick this up in a form we would take: warn, do not mutate. A single
`console.warn` where the system prompt is assembled, reporting the total size
across context **and** state, gives developers the signal without discarding data
they chose to share and without changing a public helper's contract. Happy to
review that.

zzt417520:
Thanks Ben — the design critique is fair and I'm glad you spelled it out. I over-indexed on the context value and missed that in shared-state apps the big payload is input.state, so capping the smaller neighbour leaves the real one unbounded and doesn't actually protect the window. The per-value-not-per-prompt and character-vs-token points make the limit more misleading than protective, and turning convertInputToTanStackAI — a public documented helper — into a silent, unconfigurable truncation point is a bad trade. Closing it is the right call.

The test note is also right: my surrogate assertion read charCodeAt(result.length - 1), which is always the ] of the marker, so it never checked the cut point and couldn't fail. Correct form is expect(/[\ud800-\udbff](?![\udc00-\udfff])/.test(result)).toBe(false);.

I'd like to take the "warn, do not mutate" path as a follow-up: a single console.warn at prompt assembly reporting total size across context and state, no data discarded, no public contract changed. If you're open to it I'll open that as a fresh PR off main and keep it scoped to the runtime.

PR #6756, opened 2026-08-28
chore(deps): bump @ag-ui/* to 0.0.59

mme:
Bumps the published packages from `@ag-ui/*` **0.0.57 → 0.0.59** — 27 declarations across 18 packages in `packages/**`.

0.0.59 is the first AG-UI release carrying the subagent protocol surface (`SUBAGENT_STARTED` / `SUBAGENT_FINISHED` / `SUBAGENT_ERROR`, `subagentRunId`) together with the null-omission cleanup, so it is the dependency the subagent work needs. Landing the bump on its own keeps that review separate from the feature work in #5873, which currently pins `pkg.pr.new` previews of the (now released) ag-ui PR.

### Scope

`packages/**` only — the shipped libraries, which were uniformly on 0.0.57. Examples and showcases are deliberately untouched: they sit on a spread of pins from 0.0.40 to 0.0.58 and bumping them would bury this diff. `@ag-ui/langgraph` (0.0.42) and the mcp middlewares are separate packages and also left alone.

### One behavioural change worth reviewing

`channels-core` ships `sanitizeAgentEventStream` because `@ag-ui/client` used to **reject** a `TOOL_CALL_START` carrying `parentMessageId: null` — the shape `@ag-ui/langgraph` emits for the tool call that triggers an interrupt. As of 0.0.59 the client **accepts** that null and treats it as absent (it is one of a small set of receive tolerances kept for producers that still emit it).

That invalidates the two tests asserting the run dies *without* the sanitizer:

- `aborts that same run when the sanitizer is not applied` → now `survives that same run without the sanitizer, since 0.0.59 tolerates the null`. It asserts the tool call actually arrives, not merely that nothing throws, so it cannot pass vacuously.
- `lets the run abort when sanitizing is disabled` → `no longer aborts the turn when sanitizing is disabled`.

**The sanitizer itself is untouched** and its coercion tests are unchanged — it still rewrites the byte on the wire. It is simply no longer the thing keeping such a run alive. Whether to keep it as belt-and-braces or retire it is a follow-up decision, not part of this bump.

### Verification

- All `packages/**` build against 0.0.59.
- `check-types` clean for `core`, `shared`, `react-core`, `runtime`.
- Tests pass: core 786, shared 304, runtime 2084, react-core 1514, channels-core 267.
- The full `test,publint,attw` pre-commit gate passes across the 26 affected projects.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of tool calls with missing parent message identifiers, preventing unnecessary turn failures—even when event sanitization is disabled.
  * Ensured Angular validation uses the matching AG UI client version, improving compatibility during package verification.

* **Chores**
  * Updated AG UI client, core, and encoder packages to version 0.0.59 across supported integrations and runtimes.


contextablemark (dismissed):
Interesting... so we never released with 0.0.58? Anyway, I approve!

tylerslaton (approved):

PR #6757, opened 2026-08-28
chore(deps): bump @ag-ui to the #2350 subagent-events previews

mme:
Draft, preview-build vehicle. This is CopilotKit PR #5873 **minus the SubagentRegistry and useSubagent hook** (331 lines removed) — only the dependency work remains:

- `@ag-ui/*` overrides pinned to the pkg.pr.new previews of ag-ui-protocol/ag-ui#2350, whose bundled `@ag-ui/client` carries the subagent event types (`SUBAGENT_STARTED/FINISHED/ERROR`), the `AgentSubscriber` dispatch (`onSubagentStartedEvent` etc.) and `subagentRunId` message attribution.
- The committed lockfile resolving those URLs (`pnpm install --frozen-lockfile` succeeds with `block-exotic-subdeps=true` left on; see the reviewer notes on #5873 for the .npmrc details, which carry over unchanged).

### Why the hook was dropped

The AG-UI dojo demo (ag-ui-protocol/ag-ui#2351) was `useSubagent`'s only consumer, and it turned out to need almost nothing from it: the hook supplied a label, a tooltip and a status dot, while all the actual subagent UI work (grouping, anchoring, hiding attributed messages) lived in the demo reading `message.subagentRunId` directly. The demo now tracks the lifecycle itself via the `AgentSubscriber` callbacks (~60 lines), so the dojo only needs a CopilotKit build that bundles the new `@ag-ui/client` — this PR.

The registry/hook (untested, no other consumer) can return as its own PR if demand for a first-class hook materializes; the removed code lives on `mme/subagents` / #5873 for salvage.

### Lifecycle

Like the pins it replaces, this is temporary: swap the overrides for released `@ag-ui` versions once ag-ui#2350 merges and ships. The dojo pins this PR's preview builds by commit sha (tracked in PNI-275).

Verified: `@copilotkit/core` and `@copilotkit/react-core` check-types + test green via nx after the removal, with no remaining references to the registry, hook, or `onSubagentsChanged`.


mme:
Not needed after all: #6756 already bumps the CopilotKit packages to the **released** `@ag-ui` 0.0.59 — the first release carrying the subagent surface — so there is nothing left for a pkg.pr.new-override vehicle to provide. This PR was built on the stale #5873 base (react-core 1.62.2) purely to keep the AG-UI dojo's preview pins alive after dropping the hook; the dojo now pins #6756's previews instead (react-core 1.69.3, commit `71d9731`) and its deepagents e2e passes against them. Once #6756 merges and a release ships, the dojo moves to released versions and no preview pin remains (PNI-275).

PR #6761, opened 2026-08-28
feat(web-inspector): add Try from here and isolate domains

AlemTuzlak:
Saved Threads can now open an isolated Playground copy through **Try from here**. The refactor also splits the Web Inspector monolith into bounded domains while preserving its public API and runtime behavior.

## What does this PR do?

- Adds **Try from here** to saved Thread details, with loading, error, stale-request, and accessibility behavior.
- Splits Home, Threads, Playground, Learning, live inspection, and announcements into bounded domains.
- Keeps shell coordination, package registration, telemetry, lazy loading, caches, and runtime fallbacks intact.
- Adds architecture checks for dependency direction, entry-point size, and the 800-line production-file limit.

Documentation is unchanged because the source organization is internal and the consumer API is unchanged.

## Related PRs and Issues

- None.

## Checklist

- [x] Maintainers can edit this same-repository branch.

## Testing

**Commands run**

- `pnpm nx run-many -t test check-types build compat-check publint attw --projects=@copilotkit/web-inspector --skip-nx-cache`
- Result: 750 tests passed. Type checks, build, compatibility checks, `publint`, and `attw` passed.
- `pnpm nx run @copilotkit/web-inspector:dev:standalone` for light-theme, dark-theme, and keyboard-focus checks.
- The final commit skipped the repository-wide hook after unrelated Angular and Vue tasks failed twice. The package gate passed after the rebase.

**Manual test**

1. Run `pnpm nx run @copilotkit/web-inspector:dev:standalone`.
2. Open `http://127.0.0.1:5177/?scenario=pro-enabled-existing&reset=1`.
3. Open a saved Thread and select **Try from here**.
4. Make sure that Playground opens with the copied messages and state.
5. Return to Threads and make sure that the original application agent remains unchanged.

**How this PR makes testing easy**

Focused unit and integration tests cover replay success, failure, stale requests, scratch preservation, multimodal content, registration, telemetry, and domain boundaries. The standalone workbench provides a deterministic saved-thread scenario.

## Risk / rollback

The main risk is a behavior regression at a boundary that moved from the package entry point into a domain. Revert this PR to restore the monolithic implementation.

## Summary by CodeRabbit

* **New Features**
  * Added redesigned Web Inspector experiences for Home, Threads, Playground, Learning, Live Inspection, and Settings.
  * Added thread timelines, conversations, state and event views, tool inspection, searchable memories, recall, announcements, guided tours, usage indicators, and pop-out support.
  * Added accessible copy controls, JSON viewing, dark mode, responsive layouts, and reduced-motion support.
* **Bug Fixes**
  * Improved cross-window interactions, stale loading handling, reconnect behavior, popup recovery, and Windows compatibility.
* **Privacy**
  * Added opt-out controls and privacy-focused telemetry safeguards.


kvnloo:
One UX contract in **Try from here** looks worth making explicit: today a failed state load silently becomes an empty state.

`loadPlaygroundThreadSnapshot()` requires the messages response, but does this for state:

```ts
const stateBody = stateResponse.ok
  ? await stateResponse.json()
  : { state: {} };
```

and the test intentionally pins a 503 state response to `threadState: {}`.

That means the interaction can look like a faithful branch while actually being:

`saved conversation + missing state -> new scratch thread with empty state`

For agents where state carries selections, progress, form values, or other context, the first follow-up can behave differently and the user has no clue why. That undercuts the main value of “Try from here”: not having to reconstruct the context manually.

I wouldn't necessarily block branching when state is unavailable; a message-only fork is still useful. But could the result preserve that distinction and surface it in the Playground, e.g. `stateStatus: "loaded" | "unavailable"` with a small “state wasn't available” notice?

Then the contract becomes:

> branching may degrade, but it should never silently pretend missing context was empty context.

A regression around `messages=200, state=503` could assert both that the branch remains usable and that the degraded copy is visible to the user.

AI-use note: I used an AI assistant to trace the Try-from-here load path and its state-failure regression; I verified the current PR head before posting.

PR #6763, opened 2026-08-28
docs(example): remove dead links to non-existent docs files in agentcore README

gdut4140:
The `examples/integrations/agentcore` README references `docs/LOCAL_DEVELOPMENT.md` and `docs/LOCAL_DOCKER_TESTING.md`, but the example has no `docs/` directory and neither file exists anywhere in the example. The local-development workflow is already fully documented inline in the "Local Development" section, so these references are dead links.

Verified: `docs/` is absent from `examples/integrations/agentcore/`, and a repo-wide search finds no `LOCAL_DEVELOPMENT.md` / `LOCAL_DOCKER_TESTING.md`.

### DIFF
--- a/examples/integrations/agentcore/README.md
+++ b/examples/integrations/agentcore/README.md
@@ -52,8 +52,6 @@ cp .env.example .env
 
 The full chain runs locally: `browser:3000 → bridge:3001 → agent:8080`. AWS is only used for Memory and Gateway (SSM/OAuth2).
 
-See `docs/LOCAL_DEVELOPMENT.md` for full details.
-
 ## Agent dependencies
 
 Each single-agent directory under `agents/` — `langgraph-single-agent/` and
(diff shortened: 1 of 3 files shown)


BenTaylorDev (changes requested):
Verified the claim: `examples/integrations/agentcore/` has no `docs/` directory, and neither `LOCAL_DEVELOPMENT.md` nor `LOCAL_DOCKER_TESTING.md` has ever existed in the repo's history — so these were dead from the start and there's no content to restore instead. The inline "Local Development" section does cover the workflow. Good catch.

Two more references to the same non-existent files live outside the README, and it'd be good to clear them in the same PR so the next person searching doesn't re-find them:

- `examples/integrations/agentcore/docker/docker-compose.yml:7` — `# Requires a deployed AWS stack. See ../docs/LOCAL_DEVELOPMENT.md.`
- `scripts/validate-intelligence-env-names.ts:176` — a comment citing `agentcore/docs/LOCAL_DEVELOPMENT.md`

Happy to merge once those are in.

gdut4140:
Thanks for verifying — I have cleared the two remaining `LOCAL_DEVELOPMENT.md` references (`docker/docker-compose.yml` and `scripts/validate-intelligence-env-names.ts`), so the same dead path will not be re-found.

BenTaylorDev (approved):

PR #6764, opened 2026-08-28
docs(shell-docs): fix broken Action Handlers link in a2ui fixed-schema page

gdut4140:
The "Advanced — Action Handlers" link pointed to `./advanced#action-handlers`, but there is no `advanced.mdx` page in `docs/generative-ui/a2ui/`. The "Action handlers (reference)" section lives in this same file, so the link now points to it directly with `#action-handlers-reference`.

### DIFF
--- a/showcase/shell-docs/src/content/docs/generative-ui/a2ui/fixed-schema.mdx
+++ b/showcase/shell-docs/src/content/docs/generative-ui/a2ui/fixed-schema.mdx
@@ -365,7 +365,7 @@ When available, a button declares its action like this:
 And the Python tool matches it with a handler keyed by the action
 name (plus a `"*"` catch-all). Until the SDK lands, handle the click on the
 frontend instead — see
-[Advanced — Action Handlers](./advanced#action-handlers) for the
+[Action Handlers](#action-handlers-reference) for the
 `createA2UIMessageRenderer` / `onAction` pattern.
 
 ## When should I use fixed schemas?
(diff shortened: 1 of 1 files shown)


BenTaylorDev (changes requested):
The diagnosis is right: there's no `advanced.mdx` in `generative-ui/a2ui/`, so `./advanced#action-handlers` is a 404, and `## Action handlers (reference)` does slugify to `action-handlers-reference`.

The new target has a problem though: the link at line 368 is *inside* the "Action handlers (reference)" section itself, which spans 338–370. So the fix turns "see [Action Handlers] for the `createA2UIMessageRenderer` / `onAction` pattern" into a pointer to the heading the reader is already under — and that section doesn't describe the `onAction` pattern, so the cross-reference still doesn't land anywhere useful.

A real target does exist: both integration variants have a `## Action Handlers` section — `integrations/langgraph/generative-ui/a2ui/advanced.mdx:123` and the deepagents equivalent. Either point at `/integrations/langgraph/generative-ui/a2ui/advanced#action-handlers`, or drop the link and keep the prose. Your call which reads better; both beat the 404.

gdut4140:
Good catch — the new anchor pointed at the section the link itself lives in. I have retargeted it to the LangGraph `advanced` page's `## Action Handlers` section, which actually documents the `onAction` pattern.

BenTaylorDev (approved):

PR #6765, opened 2026-08-28
docs(shell-docs): fix broken Fixed Schema (Streaming) links in A2UI docs

gdut4140:
The "Fixed Schema (Streaming)" entries in the DeepAgents and LangGraph A2UI overview pages link to `./fixed-schema-streaming`, but no such page exists in either directory. Since the streaming variant is part of the fixed-schema approach and there is no dedicated page for it, point the links at the canonical `./fixed-schema` page instead of a 404.


BenTaylorDev:
Closing as a duplicate of #6733, which makes byte-identical changes to the same two files and was opened first. Your diagnosis was correct and independently verified — no `fixed-schema-streaming` page exists in either integration directory — so this is purely a question of which of two identical fixes lands. #6733 is approved and going in.

Your other two docs PRs are open on their own threads: #6763 needs two more references to the same non-existent files cleared, and #6764 needs a different link target (the one it picked lands inside the section the reader is already in). Both are worth finishing.

PR #6766, opened 2026-08-28
fix(react-core): preserve caret when measuring single-line height in CopilotChatInput

### LINKED ISSUE #6167: 🐛 Bug: CopilotChatInput - cursor jumps to the end of the textbox upon edit in expanded mode

gdut4140:
In expanded mode, editing the input could jump the cursor to the end of the text.

`ensureMeasurements()` temporarily assigns `textarea.value = ""` and then restores it with `textarea.value = previousValue`. A direct `.value` assignment resets the browser caret to the end of the field, so any measurement pass that runs while the user has the caret mid-text moves it.

The fix captures `selectionStart`/`selectionEnd` before the measurement pass and restores them with `setSelectionRange` afterwards, so measurement never moves the user's cursor.

Fixes #6167

### DIFF
--- a/packages/react-core/src/v2/components/chat/CopilotChatInput.tsx
+++ b/packages/react-core/src/v2/components/chat/CopilotChatInput.tsx
@@ -665,6 +665,11 @@ export function CopilotChatInput({
 
     const previousValue = textarea.value;
     const previousHeight = textarea.style.height;
+    // Restoring `.value` below resets the caret to the end of the input.
+    // Capture the selection up front and re-apply it so the measurement
+    // pass never moves the user's cursor (see #6167).
+    const selectionStart = textarea.selectionStart;
+    const selectionEnd = textarea.selectionEnd;
 
     textarea.style.height = "auto";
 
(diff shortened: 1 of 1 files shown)


BenTaylorDev:
Thanks for this, and for keeping it to a single file and a single commit — the scoping made it quick to evaluate.

I'm closing it, because the caret jump in #6167 was already fixed by #5432 (shipped in v1.64.0), which removed the unconditional `ensureMeasurements()` from the narrow-viewport branch of `evaluateLayout`. That unconditional call was what ran the `.value = ""` write on every keystroke.

On current `main` the two remaining callers are both gated on `measurementsRef.current.singleLineHeight === 0` (`CopilotChatInput.tsx:701`, `:788`), and nothing resets that ref afterwards — the only assignment is at `:684`. So in a laid-out browser the measurement pass runs once per mount, before a caret can be placed mid-text, and saving/restoring the selection around it doesn't change observable behaviour. #6626 proposed the same change and was closed for the same reason.

Issue #6751, opened 2026-08-28
Channels: support managed cross-surface turns into an existing provider conversation


jerelvelarde:
## Summary

Managed Channels can reply to Slack only while handling an inbound provider delivery. It does not expose a supported way for an authenticated application surface—such as a web UI—to submit a turn into an existing managed Slack conversation and have Channels:

1. append that user turn to the same canonical agent thread;
2. serialize it with live provider turns;
3. run the same channel handler and pinned agent;
4. render through CopilotKit Intelligence's managed Slack credentials; and
5. make the user turn and agent response visible on every connected surface.

This blocks a core cross-surface workflow: start a conversation with an agent in Slack, open the same conversation on the application's domain, continue it on the web, and see the web-authored message and agent reply appear back in the original Slack thread.

## Concrete consumer: OpenBot

OpenBot uses managed createChannel({ name: "openbot" }) and CopilotKit Intelligence for Slack ingress and credentialed delivery:

- Consumer: https://github.com/jerelvelarde/openbot
- Managed channel: https://github.com/jerelvelarde/openbot/blob/997faee/server/src/slack/channel.tsx
- Durable provider/canonical-thread binding: https://github.com/jerelvelarde/openbot/blob/997faee/server/src/external/thread-store.ts
- Pinned coworker router: https://github.com/jerelvelarde/openbot/blob/997faee/server/src/slack/channel-agent.ts

On the first linked Slack turn, OpenBot already persists the provider tenant, conversation and thread IDs; Channels canonical thread ID; OpenBot user ID; and pinned agent ID.

OpenBot can use the canonical thread ID to display the same Intelligence transcript in its authenticated web UI. The missing capability is managed outbound continuation: a web-authored turn has no inbound delivery context, no supported route into the channel's private per-conversation FIFO, and no managed reply target through which Intelligence can deliver back to Slack.

## Current SDK boundary

The current core implementation documents the limitation directly:

> Proactive delivery to subscribed conversations is not yet wired.

Source: https://github.com/CopilotKit/CopilotKit/blob/main/packages/channels-core/src/thread.ts

The per-conversation queue is private to createChannel() and reached through an adapter-created IngressSink:
https://github.com/CopilotKit/CopilotKit/blob/main/packages/channels-core/src/create-channel.ts

Managed credentialed delivery is implemented by the Intelligence delivery adapter:
[...]

PR #6770, opened 2026-08-28
feat(channels): carry an application author on a managed inbound turn

jerelvelarde:
> **Half of a paired change.** This PR is the SDK half of #6751 and is
> deliberately inert on its own: nothing in the OSS repo can *produce* an
> application-authored turn. The Gateway half that mints one is
> CopilotKit/Intelligence#1064. **This PR must merge first** — an older SDK
> rejects `authoredBy` as an unexpected key at its exact-field join boundary.
> The pair is now proven end to end against a real Slack workspace; the
> recording is below.

## Problem

A managed Channel can only answer a provider event. There is no supported way
for an authenticated application surface — a web composer continuing a Slack
thread — to submit a turn into an existing managed conversation, and no way to
say who wrote it if one arrives.

The concrete workflow this blocks: start a thread with `@openbot` in Slack, open
that same conversation on the application's own domain, continue it from the
web, and see both the web-authored message and the agent's reply land back in
the original Slack thread — one ordered conversation on both surfaces.

`IncomingTurn` carries `userText`, `operation`, `messageRef`, `contentParts`,
`platform` and the ingress ids, and nothing else. There is no field for an
author, so a web-composed message could only be attributed by naming the Slack
actor — which would be a lie, and exactly the impersonation #6751 rules out.

## Proven end to end

Full journey

Recorded against the real CopilotKit Slack workspace (`#openbot-channels`, team
`T05QFA4BW9X`) with app-api, the Realtime Gateway and OpenBot running locally,
and the `@openbot` app's Event Request URL pointed at that stack for the run.
Every Slack message in the recording is a real message in that workspace.

1. **Slack** — a person types `@openbot in one sentence, what does the CopilotKit
   Channels SDK do?`. It carries no `app_id`, so ingress classifies the actor
   `kind: "human"` and OpenBot answers in the thread, posting its
   "Open in OpenBot" transcript card.
2. **OpenBot** — the same conversation, opened from that card. OpenBot created
   the thread binding itself from the inbound delivery; nothing was seeded. The
   composer is live because Intelligence issued a conversation reference for it.
3. **The person composes their follow-up there.**
4. **OpenBot** — the agent answers.
5. **Slack** — the follow-up is in the same thread as
   `dev@openbot.local via openbot`, with the agent's answer beneath it.

One conversation, two surfaces, correct order, attribution intact.

Delivery rows for that run: the inbound turn is
`origin=provider, actor.kind=human, provider_delivery_complete`; the web turn is
`origin=application` carrying
`authoredBy={kind: application, surface: openbot, appUserId: dev-local-user}`,
with an allocated `next_turn_seq` slot and
`authored_echo_provider_message_id=1788187934.740279` — the Slack ts of the echo.

Higher-resolution renders, and the 3840×2072 source frames, are on the
[`evidence/bidirectional-turn`](https://github.com/jerelvelarde/openbot/tree/evidence/bidirectional-turn/evidence)
branch: [1080p MP4](https://raw.githubusercontent.com/jerelvelarde/openbot/evidence/bidirectional-turn/evidence/openbot-journey-1080p.mp4) ·
[4K MP4](https://raw.githubusercontent.com/jerelvelarde/openbot/evidence/bidirectional-turn/evidence/openbot-journey-4k.mp4).

## What this changes

Four optional fields and one options passthrough. Purely additive.

| | |
|---|---|
| `ChannelAuthoredActor` | New public type in `channels-ui`: `kind`, `surface`, `appUserId`, `displayName`. |
| `IncomingMessage.authoredBy?` | So a handler can render attribution. `ChannelMessage` extends it. |
| `IncomingTurn.authoredBy?` | The ingress-side field an adapter may set. |
| `PreparedChannelDelivery.turn.authoredBy?` | The managed wire field, validated at join. |
| `maxConcurrentDeliveries` / `maxPendingDeliveries` | Now reach the transport from `CopilotRuntime`. |

**No new method on `Channel`, no new lifecycle seam, no `channel.trigger()`.**

That is the main design decision and it is worth stating plainly. The obvious
shape for #6751 is a programmatic trigger API. I did not add one, because a web
turn does not need a second ingress path — it needs to *be* an ordinary inbound
delivery. Modelled that way it reuses the entire existing dedupe → prepare →
claim → admit → fence pipeline unchanged, and inherits per-conversation ordering
and idempotency for free. Anything the SDK invented here would be a second,
weaker copy of machinery Intelligence already has.

It also keeps the in-process `conversationTurnQueues` out of the picture. That
map is per-replica and could never satisfy "one ordered conversation" across a
fleet; the durable gate is `ConversationStore.admit` in the Gateway, keyed on
the canonical thread id, and a web turn contends on exactly that key.

### Attribution is unrepresentable-by-forgery, not merely validated

`authoredBy` is minted by Intelligence from the caller's verified session during
authenticated ingress. The SDK only forwards it. There is no field on any
ingress type by which an adapter, an agent, a tool result, or a delivery payload
can name an author — so an application cannot claim to be somebody else, and the
provider actor is left untouched on the turn. Attribution is additive; it never
launders a web message into "the Slack user said this".

At the join boundary the field is validated the same way as everything else that
crosses that wire (`hasExactFields` + regex-bounded ids, per
`delivery-transport.ts`): unknown keys are rejected rather than ignored, `kind`
is pinned to the single literal the SDK knows so a newer Gateway fails loudly
instead of being flattened, and `appUserId` is bounded by `isSafeAppUserId` so it
cannot carry a path or control characters into a handler that renders it.

### The capacity fix is not incidental

`defaultActivateChannel` never forwarded `maxConcurrentDeliveries` /
`maxPendingDeliveries`, though the launcher and transport have always accepted
them. Every Channel activated through `CopilotRuntime` therefore ran pinned at
the transport defaults of **8 concurrent / 32 pending**, regardless of host size.

Throughput per replica is about `maxConcurrentDeliveries / meanTurnSeconds`, so
at a 4 s mean turn one replica was capped near **2 turns/sec**. Reaching the
1000 concurrent-turns/sec the managed path is meant to serve would have needed
~500 replicas to work around a default. Unset still means the transport default,
so no existing deployment changes behaviour.

## Tests

New: `packages/channels-intelligence/src/delivery-authored-actor.test.ts` (9).

- an application-authored turn reaches the handler as attribution, with the
  provider actor preserved exactly;
- a provider-authored turn carries no author (absence is the signal, so it must
  be a real `undefined`);
- **6 join-boundary refusals** — unknown author class, unexpected extra field,
  missing display name, blank display name, unsafe `appUserId`, non-string
  surface;
- an application turn waits for the provider turn already running on its thread.

Plus 2 in `channel-activation-config.test.ts` for the capacity passthrough and
for omission staying omission.

**Every test was mutation-checked**, because a test that cannot fail is worse
than no test:

- Neutering `isPreparedAuthoredActor` → all 6 refusal cases fail (a `surface: 7`
  author reaches the handler).
- The ordering test needed two corrections before it was honest. It first passed
  with `maxConcurrentDeliveries: 1`, where the replica's own capacity ceiling did
  the ordering rather than the shared thread — so capacity now has headroom. It
  then still passed because the assertion ran before the async join resolved — so
  it now drains the queues to give the second turn a genuine opportunity to run
  before asserting it did not. With both fixed, removing the `threadTails`
  chaining in `delivery-transport.ts` fails it.

## Verification

Run in a clean worktree off `origin/main` (`4ccae6fe2`):

```
pnpm check-format                                  # clean
pnpm lint                                          # 0 errors
pnpm nx run-many -t check-types -p channels*,runtime   # clean
pnpm nx run-many -t test -p channels*,runtime      # 3123 passed
pnpm nx run-many -t build -p channels*,runtime     # clean
pnpm nx run-many -t publint,attw -p channels*      # clean
pnpm check:public-api-manifest                     # current
```

Commits: `c2edad3b276da9029fcd64f09681864709e00a4f`, `90dd8bdd5428763569c2cd8d05db993394b02e88`.

No changeset: Channels versioning is the `channels` scope in
`release.config.json` (`sharedVersion`, `versionSource @copilotkit/channels`).

## Known limitations

- **Inert alone.** Nothing here produces an `authoredBy`; the Gateway does.
- **Text turns only.** An interaction, reaction or welcome is by construction
  something that happened on the provider surface, so an author there would
  describe an event that did not occur.
- **No revocable conversation reference.** #6751 also asks for an opaque,
  persistable, revocable managed-conversation reference. That is entirely a
  Gateway concern — the SDK never sees it — so it lives in the paired PR.
- **Whether Slack renders the author as a distinct message author** (`username` /
  `icon_url` overrides) or as in-body text depends on the installed app's
  scopes. Open question on #6751; the SDK contract is unaffected either way.
- **Attribution is not authorization.** `appUserId` is an opaque tenant-scoped
  id, deliberately not resolvable via `Thread.lookupUser`.

## Links

- Contract issue: #6751
- Paired Gateway PR: CopilotKit/Intelligence#1064
- Consumer: https://github.com/jerelvelarde/openbot

## Update — rebased onto `main`, and the Intelligence half is live

Rebased onto `main` (`9909ca4`); the branch was merge-dirty against an import
list in `platform-adapter.ts` that `main` had also touched. Resolved by keeping
both imports. **The PR is now MERGEABLE.**

`main` has since added `ApplicationUser` to `channels-ui`, which is a different
thing and worth not confusing: `ApplicationUser` is the identity a Channel
developer selects for their own purposes and the platform never verifies.
`ChannelAuthoredActor` is chosen by the platform and cannot be set by the
developer, which is exactly why a handler may render it as attribution without
re-verifying anything. The doc comment now says so.

The paired Gateway half is CopilotKit/Intelligence#1064, which now mints these
authors server-side and delivers them. Verified against a live app-api, Realtime
Gateway and an OpenBot runtime built from **this** branch: a web-composed turn
was accepted, prepared with `authoredBy` in `prepared_input`, published as a
wake, claimed by the runtime, and reached the handler with the author intact.

That last part is why the ordering matters: an SDK on the published `0.9.0`
rejects `authoredBy` as an unexpected key in `assertPreparedDelivery`'s
exact-field check, so an application turn delivered to it fails the join rather
than reaching a handler. **This PR ships first.**

## Live run — what it does and does not prove

Application-authored turn traversing a local Intelligence stack

**This is not a bidirectional turn.** No Slack workspace was involved at any
point: the adapter is attached to a fabricated workspace (`T0LOCALDEV1`), the
conversation is a seeded row, and `cpki.channel_provider_messages` is empty
afterwards. Neither the attributed echo nor the agent's reply left the machine,
and the deliveries end `uncertain` rather than `complete`. Nothing was typed
into a composer either — the turn was posted to the same route the composer
calls.

What it does prove, against running services rather than tests — Intelligence
app-api, the Elixir Realtime Gateway, PostgreSQL, Redis, and an OpenBot runtime
on the control socket — is the part of this change that was pure design before:

- a conversation reports `readOnly: false` because Intelligence issued a
  reference for it, which no thread could reach before this branch;
- a web-authored turn is accepted and becomes an **ordinary inbound delivery** —
  `origin=application`, ordering slot 0, wake settled, echo claimed — in the same
  table and pipeline a Slack webhook uses;
- the server-minted `authoredBy` reaches `prepared_input`, which is where the
  SDK reads it;
- the Gateway invites the runtime, the runtime claims, and the agent runs;
- the Redis conversation owner carries `activeOrigin: application`, the gate that
  stops the two surfaces superseding each other.

The script is
[`evidence/live-run.sh`](https://github.com/jerelvelarde/openbot/blob/pr-media/evidence/live-run.sh).
An uncut Slack → web → Slack recording needs a real workspace bot token and is
still outstanding.

### DIFF
--- a/packages/channels-core/src/create-channel.ts
+++ b/packages/channels-core/src/create-channel.ts
@@ -434,6 +434,31 @@ export interface CreateChannelOptions<
    * configure direct Slack with `slack({ replyContinuation })` instead.
    */
   replyContinuation?: ReplyContinuationOptions;
+  /**
+   * Per-replica ceiling on managed deliveries executing at once.
+   *
+   * The managed delivery transport has always accepted this, but a Channel
+   * activated through `CopilotRuntime` had no way to set it, so every such
+   * deployment ran on the transport default of 8 regardless of how the host was
+   * sized. Since a replica's throughput is about
+   * `maxConcurrentDeliveries / meanTurnSeconds`, that default — not the agent
+   * and not the provider — was the binding constraint on managed throughput.
+   *
+   * Unset keeps the transport default, so raising it is an explicit decision by
+   * an operator who knows what the host and the agent can absorb.
+   *
+   * Ignored for direct-adapter Channels, which do not go through the managed
+   * delivery transport.
+   */
+  maxConcurrentDeliveries?: number;
+  /**
+   * Per-replica ceiling on managed deliveries that are claimed but not yet
+   * executing. Invitations beyond
+   * `maxConcurrentDeliveries + maxPendingDeliveries` are declined as overflow
+   * and left for another replica rather than queued locally, so this trades
+   * local tail latency against how often this replica declines work.
+   */
+  maxPendingDeliveries?: number;
   agent?: AbstractAgent | ((threadId: string) => AbstractAgent);
   /**
    * Tolerate the AG-UI event streams real agents emit. On by default.
@@ -483,6 +508,17 @@ export interface Channel<TState = unknown> {
    * Undefined leaves the provider defaults in place.
    */
   readonly replyContinuation?: ReplyContinuationOptions;
+  /**
+   * Managed delivery capacity from `createChannel({ maxConcurrentDeliveries })`.
+   * Undefined leaves the delivery transport's own default in place.
+   */
+  readonly maxConcurrentDeliveries?: number;
+  /**
+   * Managed pending-delivery capacity from
+   * `createChannel({ maxPendingDeliveries })`. Undefined leaves the delivery
+   * transport's own default in place.
+   */
+  readonly maxPendingDeliveries?: number;
   /** Declared slash-command names (normalized). Surfaced for Channel activation metadata. */
   readonly commandNames: string[];
   onMention(h: ChannelHandler<TState>): void;
@@ -559,6 +595,11 @@ function msgFromTurn(turn: IncomingTurn): ChannelMessage {
     eventId: turn.eventId,
     turnId: turn.turnId,
     deliveryId: turn.deliveryId,
+    // Carried, never interpreted. Core deliberately takes no branch on the
+    // authoring surface: routing, dedup and ordering must stay identical for a
+    // web-composed turn and a provider one, or the two surfaces stop being one
+    // conversation. Only the handler's own rendering may read this.
+    authoredBy: turn.authoredBy,
   };
 }
 
@@ -1323,6 +1364,12 @@ export function createChannel<
       : {}),
     ...(opts.replyContinuation !== undefined
       ? { replyContinuation: opts.replyContinuation }
+      : {}),
+    ...(opts.maxConcurrentDeliveries !== undefined
+      ? { maxConcurrentDeliveries: opts.maxConcurrentDeliveries }
+      : {}),
+    ...(opts.maxPendingDeliveries !== undefined
+      ? { maxPendingDeliveries: opts.maxPendingDeliveries }
       : {}),
     get adapters() {
       // Defensive read-only copy: mutating the returned array must not affect
--- a/packages/channels-core/src/platform-adapter.ts
+++ b/packages/channels-core/src/platform-adapter.ts
@@ -2,6 +2,7 @@ import type { AgentSubscriber, AbstractAgent } from "@ag-ui/client";
 import type {
   AgentContentPart,
   ApplicationUser,
+  ChannelAuthoredActor,
   ChannelNode,
   EmojiValue,
   EphemeralResult,
@@ -166,6 +167,12 @@ export interface IncomingTurn extends IngressEventBase, IngressIds {
    */
   contentParts?: AgentContentPart[];
   platform: string;
+  /**
+   * Set when this turn was composed on an application surface rather than the
+   * provider surface. Provider-authored turns omit it, so an adapter that
+   * cannot produce application turns needs no change at all.
+   */
+  authoredBy?: ChannelAuthoredActor;
 }
 
 export interface InteractionEvent extends IngressEventBase, IngressIds {
--- a/packages/channels-intelligence/src/delivery-adapter.ts
+++ b/packages/channels-intelligence/src/delivery-adapter.ts
@@ -407,6 +407,13 @@ export class DeliveryAdapter implements PlatformAdapter {
                 ],
               }
             : {}),
+          // Only a text turn can be application-authored: an interaction, a
+          // reaction or a welcome is by construction something that happened on
+          // the provider surface, so forwarding an author there would describe
+          // an event that did not occur.
+          ...(delivery.turn.authoredBy !== undefined
+            ? { authoredBy: delivery.turn.authoredBy }
+            : {}),
         });
         return;
       }
--- a/packages/channels-intelligence/src/delivery-authored-actor.test.ts
+++ b/packages/channels-intelligence/src/delivery-authored-actor.test.ts
@@ -1,0 +1,309 @@
+import { createChannel } from "@copilotkit/channels-core";
+import type { ChannelAuthoredActor } from "@copilotkit/channels-ui";
+import { expect, test, vi } from "vitest";
+import {
+  DeliveryTestGateway,
+  preparedDelivery,
+} from "./delivery-test-gateway.js";
+import { ChannelDeliveryTransport } from "./delivery-transport.js";
+import type { PreparedChannelDelivery } from "./delivery-transport.js";
+import type {
+  RealtimeGatewayDeliveryChannel,
+  RealtimeGatewaySession,
+} from "./realtime-gateway.js";
+import { startChannelsWithGatewayControl } from "./realtime-gateway-launcher.js";
+
+/**
+ * Contract tests for CopilotKit/CopilotKit#6751 — a turn composed on an
+ * application surface (a web composer continuing a Slack thread) delivered
+ * through the ordinary managed path.
+ *
+ * The invariant these pin down is that an application turn is an ordinary
+ * delivery in every respect except who is named as its author. Nothing about
+ * routing, identity resolution or the provider actor may change, because the
+ * two surfaces are meant to be one conversation.
+ */
+
+const AUTHORED_BY: ChannelAuthoredActor = {
+  kind: "application",
+  surface: "web",
+  appUserId: "openbot:user-42",
+  displayName: "Jerel via OpenBot",
+};
+
+const SHARED_THREAD_ID = "thread_authored_fifo";
+const PROVIDER_DELIVERY_ID = "dlv_authoredFifoProvider";
+const APPLICATION_DELIVERY_ID = "dlv_authoredFifoApplication";
+
+/** One claimed-and-acknowledged delivery topic, mirroring the transport suite. */
+function deliveryChannel(
+  joinReply: PreparedChannelDelivery,
+): RealtimeGatewayDeliveryChannel {
+  return {
+    joinReply,
+    push: vi.fn().mockImplementation((_event, packet) => {
+      const identity = packet as {
+        deliveryId: string;
+        seq: number;
+        packetId: string;
+      };
+      return Promise.resolve({
+        deliveryId: identity.deliveryId,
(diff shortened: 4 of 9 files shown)


jerelvelarde:
## Recording: a web-authored turn reaching its Slack thread

Bidirectional turn

Recorded against the real CopilotKit Slack workspace (`#openbot-channels`,
team `T05QFA4BW9X`) with app-api, the Realtime Gateway and OpenBot running
locally against it. Every Slack message in the recording is a real message in
that workspace.

The journey, in order:

1. **Slack** — an existing `@openbot` thread, no replies yet.
2. **OpenBot web** — the same conversation, composer live. It is live because
   Intelligence issued a conversation reference for it; without one the
   composer stays read-only on its own rather than behind a flag.
3. **The person types their follow-up in OpenBot and sends it.**
4. **Slack** — their turn appears in the thread as `dev@openbot.local via
   openbot`, and the agent's answer follows it.

Step 4 is the whole point: the coworker in Slack sees the question that was
asked, attributed to the person who asked it, before they see the answer to it.

### What this run found

Three defects that only a real round trip surfaces, all fixed in the pushed
commits:

- **The echo could never post.** `ProviderExecutor` fences every provider call
  behind a `_recheckOwnership` closure and fails closed when it is absent.
  `AuthoredEcho` passed the context straight out of `ProviderContextStore`,
  which does not carry one, so every echo claimed its delivery and then lost
  the call to `uncertain_owner_lost`.
- **That failure was invisible.** `TelemetryLogHandler` builds its own event
  list instead of reading `Channels.Telemetry`, and `:authored_echo` was only
  in the latter — so the event fired with no handler attached. No log, no
  metric, for a module whose docstring promises the failure is reported.
- **OpenBot answered the author with a link card.** A web turn was run through
  the Slack identity resolution, which keys on an actor id that is an OpenBot
  user id for these turns — so it always resolved "unlinked" and replied, in
  the Slack thread, asking the person to link the Slack identity they had not
  used.

### Scope of the run

The inbound Slack webhook hop is not exercised here: the `@openbot` app's
Request URL points at production `api.intelligence.copilotkit.ai`, so a fresh
mention typed in Slack does not reach a local app-api. The thread each
recording starts from was created by a real, signature-verified Slack event
delivered to this stack earlier, and OpenBot's binding for it was seeded
directly. Everything from the web turn onward — ordering slot, wake, claim,
admission, echo, agent run and both Slack posts — ran for real.

jerelvelarde:
## The inbound hop is now covered too — full journey, end to end

The earlier recording started from an already-delivered Slack thread, because
the `@openbot` app's Request URL pointed at production. That gap is now closed:
the URL was pointed at this stack, a human typed the mention in Slack, and the
whole thing ran over Slack's real Events API.

Full journey

1. **Slack** — a person types `@openbot in one sentence, what does the
   CopilotKit Channels SDK do?`. No `app_id` on it, so ingress classifies the
   actor `kind: "human"` and OpenBot's guard lets it through. It answers in the
   thread ~10s later and posts its "Open in OpenBot" transcript card.
2. **OpenBot web** — the same conversation, opened from that card. OpenBot
   created the binding itself off the inbound delivery; nothing was seeded.
   The composer is live because Intelligence issued a conversation reference.
3. **The person answers from OpenBot.**
4. **Slack** — their turn lands in the same thread as `dev@openbot.local via
   openbot`, and the agent's reply follows it.

One conversation, two surfaces, correct order, attribution intact.

Verified on the delivery rows: the inbound one is
`origin=provider, actor.kind=human`; the web one is `origin=application` with
`authoredBy={kind: application, surface: openbot}`, a `next_turn_seq` slot, and
`authored_echo_provider_message_id` holding the Slack ts of the echo.

The three defects described in the previous comment are fixed in the pushed
commits — all three were only reachable by running the round trip for real.

*(The Event URL has since been returned to
`api.intelligence.copilotkit.ai`; pointing it at a laptop was a temporary
arrangement for this recording.)*

jerelvelarde:
## Re-recorded at 2× capture resolution

Same journey as the previous comment, captured properly this time — 3840×2072
per frame instead of 1430×885, so the text is actually legible.

Full journey, HD

Also on the `evidence/bidirectional-turn` branch:
[1080p MP4](https://raw.githubusercontent.com/jerelvelarde/openbot/evidence/bidirectional-turn/evidence/openbot-journey-1080p.mp4) ·
[4K MP4](https://raw.githubusercontent.com/jerelvelarde/openbot/evidence/bidirectional-turn/evidence/openbot-journey-4k.mp4) ·
[source frames](https://github.com/jerelvelarde/openbot/tree/evidence/bidirectional-turn/evidence/hd)

The five beats:

1. **Slack** — a person types the `@openbot` mention. No `app_id`, so ingress
   classifies `actor.kind: "human"`; OpenBot answers in the thread and posts its
   "Open in OpenBot" card.
2. **OpenBot** — the same conversation, composer live off the issued
   conversation reference. OpenBot created the thread binding itself from the
   inbound delivery.
3. **The person composes their follow-up there.**
4. **OpenBot** — the agent answers.
5. **Slack** — the follow-up is in the thread as `dev@openbot.local via
   openbot`, with the agent's answer under it.

Delivery rows for this run: inbound `origin=provider, actor.kind=human,
provider_delivery_complete`; the web turn `origin=application` with
`authoredBy={kind: application, surface: openbot}` and
`authored_echo_provider_message_id=1788187934.740279`.

PR #6773, opened 2026-08-28
feat(angular): add registerComponent so a component can be a tool the agent calls (refs OSS-1034)

BenTaylorDev:
## Why

Angular had no counterpart to `useComponent` in `@copilotkit/react-core/v2` and `@copilotkit/vue/v2`. Letting an agent display one of your components meant either `registerFrontendTool` with a handler the component does not need, or `registerRenderToolCall`, which draws a tool the agent already has and therefore requires that tool to exist in the agent.

That gap is load-bearing for OSS-1034, which ends the default onboarding proof at a component rendered through the frontend's own registration. Every other supported frontend ships `useComponent`; Angular was the only one that could not express the shape.

## What changed

**`FrontendToolConfig.handler` is now optional**, and `CopilotKit.#bindClientTool` returns the config unwrapped when it is absent. This is the actual missing capability. Core already has the render-only path — a tool that declares no handler gets an empty tool result and the turn completes (`packages/core/src/core/run-handler.ts`, the `if (tool.handler)` guards around lines 854 and 1112). Angular bound a wrapper unconditionally:

\`\`\`ts
handler: (args, context) =>
  runInInjectionContext(injector, () => handler(args, context)),
\`\`\`

so a display-only tool called \`undefined\` on its first tool call. A stub handler is the wrong fix in the other direction: whatever it returns becomes the tool's result in the thread, which is a fabricated result rather than an absent one.

**`registerComponent`** is that shape with the model-facing description built for it — the same prefix react-core and vue build, so the tool reads identically to the model whichever frontend registered the component. It carries no `handler` field at all: a display-only component that quietly ran application code would be a different feature, and a caller who wants both wants `registerFrontendTool`.

The component stays an ordinary `ToolRenderer` with a required `toolCall` signal input reading `toolCall().args`. No new component contract, and it renders through the existing `pickToolCallHandler` → `NgComponentOutlet` path unchanged.

Because the tool is declared by the frontend and forwarded over AG-UI, a display-only component needs nothing added to the agent — on any framework, Python included.

\`\`\`ts
registerComponent({
  name: "show_incident",
  description: "Show one incident from the incident table.",
  parameters: z.object({ id: z.string(), severity: z.string() }),
  component: IncidentCardComponent,
});
\`\`\`

## Docs

- New `reference/angular/functions/registerComponent.mdx`, wired into the reference index callout and the `public-api.mdx` summary.
- A display-only section and a new path-table row in the Angular generative-UI guide.
- `packages/angular/API.md` — the `public-api-documentation.spec.ts` gate fails on any undocumented export, which is how the omission surfaced immediately.

The reference page also carries the grounding warning: the model fills these props from what it knows, so a component rendered over records the application does not hold looks identical to a correct one in a browser and in a screenshot.

## Testing

Five tests written before the implementation, confirmed red (`(0 , registerComponent) is not a function` ×4, plus the handler-less binding test) with the 14 pre-existing registration tests still passing, then green.

- `pnpm nx run @copilotkit/angular:test` — 49 files, 321 passed, 1 skipped.
- `pnpm nx build @copilotkit/angular` — passes, so widening `handler` to optional broke no caller.
- `showcase/shell-docs` — 66/68 files, 480 tests pass. The 2 failures are pre-existing on `main` and read files this branch does not touch (a shared inspector snippet, and mastra tool-rendering content).
- `pnpm run check-format` — the 22 reported files are all pre-existing; none is in this diff.

Refs OSS-1034

### DIFF
--- a/packages/angular/API.md
+++ b/packages/angular/API.md
@@ -179,6 +179,7 @@ Import these symbols from `@copilotkit/angular`.
 - `RENDER_A2UI_TOOL_NAME`
 - `ReadAloudButtonContext`
 - `RegenerateButtonContext`
+- `RegisterComponentConfig`
 - `RenderA2UIArgs`
 - `RenderA2UIArgsSchema`
 - `RenderActivityMessageConfig`
@@ -241,6 +242,7 @@ Import these symbols from `@copilotkit/angular`.
 - `provideCopilotKit`
 - `provideSlots`
 - `readA2UILifecycleContent`
+- `registerComponent`
 - `registerFrontendTool`
 - `registerHumanInTheLoop`
 - `registerRenderActivityMessage`
--- a/packages/angular/src/lib/copilotkit.spec.ts
+++ b/packages/angular/src/lib/copilotkit.spec.ts
@@ -219,6 +219,34 @@ describe("CopilotKit", () => {
     expect(handlerSpy).toHaveBeenCalledWith({ value: "ok" }, mockContext);
   });
 
+  it("adds a display-only client tool with no handler at all", () => {
+    TestBed.configureTestingModule({
+      providers: [provideCopilotKit({ licenseKey })],
+    });
+
+    const copilotKit = TestBed.inject(CopilotKit);
+    const injector = TestBed.inject(Injector);
+
+    copilotKit.addFrontendTool({
+      name: "display-only",
+      description: "Renders a card",
+      args: z.object({ value: z.string() }),
+      component: class {
+        toolCall = signal({} as any);
+      },
+      injector,
+    } as never);
+
+    // Core inserts an empty tool result for a tool that declares no handler, so a
+    // display-only registration must reach it with `handler` still absent. Binding
+    // a wrapper around `undefined` here would throw on the first tool call, and
+    // supplying a stub would write a fabricated result into the thread.
+    const tool = mockAddTool.mock.calls.at(-1)![0];
+    expect(tool.name).toBe("display-only");
+    expect(tool.handler).toBeUndefined();
+    expect(copilotKit.clientToolCallRenderConfigs()).toHaveLength(1);
+  });
+
   it("registers human-in-the-loop tools and delegates responses", async () => {
     const onResultSpy = vi
       .spyOn(HumanInTheLoop.prototype, "onResult")
--- a/packages/angular/src/lib/copilotkit.ts
+++ b/packages/angular/src/lib/copilotkit.ts
@@ -300,6 +300,14 @@ export class CopilotKit {
   ): FrontendTool {
     const { injector, handler, ...frontendCandidate } = clientToolWithInjector;
 
+    // A display-only registration declares no handler, and core has its own path
+    // for that: it inserts an empty tool result and completes the turn. Binding a
+    // wrapper here regardless would call `undefined` on the first tool call, and
+    // substituting a stub would put an invented result into the thread instead.
+    if (!handler) {
+      return frontendCandidate;
+    }
+
     return {
       ...frontendCandidate,
       handler: (args, context) =>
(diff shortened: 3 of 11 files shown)


BenTaylorDev:
## Real-app verification

The tests in this PR assert on the registration shape. I also drove the whole path end to end against the real `CopilotKit` service and real `CopilotKitCore` — registration, forwarding, turn completion, and rendering — to check that a display-only tool actually behaves in an app and not just at the seam. Notes for reviewers:

**The tool reaches the agent.** `buildFrontendTools()` after a `registerComponent` call emits:

```json
{
  "name": "show_incident",
  "description": "Use this tool to display the \"show_incident\" component in the chat. This tool renders a visual UI component for the user.\n\nShow one incident from the incident table.",
  "parameters": {
    "type": "object",
    "properties": { "id": { "type": "string" }, "severity": { "type": "string" } },
    "required": ["id", "severity"]
  }
}
```

`buildFrontendTools` (`packages/core/src/core/run-handler.ts:1196`) never reads `handler`, so the absent handler does not affect forwarding.

**The turn completes.** Running an agent that emits the tool call produces:

```
assistant (toolCalls: show_incident)  →  tool ""  →  assistant "Here is the incident."
```

Empty tool result, no error, and the follow-up turn fires — matching react-core, where `useComponent` also leaves `followUp` at its default. Callers who don't want the extra turn have `followUp` on the config, documented on the reference page.

**The component renders** through the real `RenderToolCalls` → `pickToolCallHandler` → `NgComponentOutlet` path: `"complete|INC-4711|sev1"`. The picker filters on `component !== undefined`, not on `handler`.

**The `#bindClientTool` change is load-bearing.** Removing the `if (!handler) return frontendCandidate` early return and re-running makes the tool result `"Error: handler is not a function"`. So the pre-fix symptom is not a crash — it is a fabricated error result written into the thread and fed back to the model on the follow-up turn, which is the failure this PR is really removing.

**It is in the shipped package.** `nx build @copilotkit/angular` → `dist/types/copilotkit-angular.d.ts:143` declares `registerComponent`, and both `registerComponent` and `RegisterComponentConfig` appear in the built export list, so `import { registerComponent } from "@copilotkit/angular"` resolves for a consumer.

**Zero-argument components work.** `parameters: z.object({})` forwards `{"type":"object","properties":{}}`.

Full Angular suite: 49 files, 321 passed / 1 skipped.

---

### Two things reviewers should know

**1. Chat history loses the card when the registering component is destroyed — pre-existing, not from this PR.**

Render the card, destroy the registrar, re-render: the card is gone. Angular's `removeTool` (`packages/angular/src/lib/copilotkit.ts:594`) drops the render config along with the tool. react-core deliberately does the opposite, in `use-frontend-tool.tsx`:

> we are intentionally not removing the render here so that the tools can still render in the chat history

I confirmed this is pre-existing by running the same scenario through `registerFrontendTool`, which this PR does not touch — identical result. It is worth calling out anyway because it bites `registerComponent` hardest: display is the entire point of the API, and the scenario is ordinary (a persistent chat sidebar plus a route change destroys the page component that registered the tool, and every card it already drew in the transcript disappears). Follow-up issue, not a blocker here.

**2. The guide snippet can't be pasted as-is.**

The reference page states the injection-context requirement and shows the imports. The guide's second snippet is a bare `registerComponent({ ... })` with no import lines and no mention that it must run in a constructor or field initializer — pasted into `ngOnInit` it throws NG0203. The `registerFrontendTool` section immediately below it does say this. Two lines closes it.

Minor API divergence worth a decision: `parameters` is required on `RegisterComponentConfig`, optional on react-core's `useComponent`.

rainerhahnekamp:
@BenTaylorDev would it make sense to link back to `registerComponent` from the existing `registerFrontendTool` and `registerRenderToolCall` API pages? The new page already links to those APIs, so reciprocal links could help users choose the right function. If that is outside this PR’s scope, no worries.

BenTaylorDev:
Good call — done in dd112c9. Both pages now carry a `registerComponent` card in Related, plus a line in the Overview where each already steers you to a sibling: `registerFrontendTool` points at `registerComponent` instead of a stub handler, and `registerRenderToolCall` notes it declares the tool from the frontend so there's nothing to add on the agent side.

Not out of scope at all — React and Vue already link from their tool hooks to `useComponent` (`useFrontendTool`, `useRenderTool`, `useDefaultRenderTool` on React; `useFrontendTool` on Vue), so Angular was the only frontend where the links ran one way. I left `registerHumanInTheLoop` out to match React/Vue, where the HITL page doesn't link to `useComponent` either.

rainerhahnekamp (approved):

PR #6778, opened 2026-08-28
feat(showcase): report docs content merged to main but not promoted (closes OSS-1038)

BenTaylorDev:
## Why

`docs.copilotkit.ai` moves only when a human dispatches `showcase_promote.yml` — *"Humans trigger. No automatic prod promotes."* That gate is deliberate and this PR does not argue with it. What was missing is any way to see **what the gate is currently holding.**

A merged page returns an honest 404 on prod, and nothing distinguishes that from a page nobody wrote. Three tickets are that same confusion:

- **OSS-948** — filed as "Angular's Inspector fix is written but unpublished".
- **OSS-1005** — closed by a commit whose page was still not live four days later.
- **OSS-1037** — I filed it, on the wrong diagnosis: I read a 404 as a routing defect and proposed fixing a route that was never broken. That cost an investigation and a retraction.

It reaches the onboarding graph too. The graph fetches `docs.copilotkit.ai/*.md` at run time and reads a failed fetch as "documentation does not support the selection", routing to `unsupported/no-validated-path`. An un-promoted page therefore fails a run for a documentation gap that does not exist — which is why OSS-1034's Angular node had to cite an older live guide instead of the new reference page.

## The signal was already public

I filed OSS-1038 claiming "nothing today can answer *is prod current?*". That was wrong too. The shell already renders its own build SHA into every page:

```tsx
// showcase/shell-docs/src/app/layout.tsx
<div aria-hidden="true" className="shell-docs-commit-label">{commitLabel}</div>
```

So the deployed commit is readable with a GET and **no secret** — this check works on fork PRs like any other. Nothing was consuming it.

## What it reports

Run against prod the day this was written:

```
✗ docs content has been waiting 3+ days to be promoted, past the 3-day budget.
  docs.copilotkit.ai 54567d1 → main addd3f8: 48 pages differ under showcase/shell-docs/src/content.
  3 pages merged and still returning 404:
    showcase/shell-docs/src/content/docs/frontends/vue/guides/generative-ui.mdx
    showcase/shell-docs/src/content/docs/integrations/pydantic-ai/agent-app-context.mdx
    showcase/shell-docs/src/content/snippets/integrations/agno/frontend-tool-persistence.mdx
  44 pages serving older content:
    …
  1 page removed from main but still served:
    showcase/shell-docs/src/content/snippets/shared/inspector/open-inspector-step-angular.mdx
  Promote shell-docs: run the `Showcase: Promote (staging → prod)` workflow
  with service `shell-docs`. Prod does not move on its own.
```

And against staging, which is current:

```
✓ docs.staging.copilotkit.ai is serving d692f25 and no docs content is waiting to be promoted.
```

## Design decisions worth reviewing

**Added, modified and deleted are separate findings**, because a reader experiences them differently: an added page 404s, a modified page serves older content and gives no sign of it, a deleted page is still being served. A rename counts as **both** an add and a delete — the new URL 404s while the old one still resolves, and calling it a modification would hide both.

**Age comes from the oldest unpromoted *content* commit, not the deployed commit's date.** A promote can be old while the content it is missing is new, and the second is what a reader is waiting on.

**It reports, and fails only past a budget** (3 days, configurable). Drift is the normal state between promotes, so failing the moment a PR merges would make this noise — and a check that is always red is a check nobody reads.

**An unreadable or unresolvable label is its own failure**, never "prod is current". `dev` and `unknown` are the shell's own words for "no commit" (unset vs. empty build ARG); resolving either would find nothing and report the entire tree as drifted. An unresolvable SHA tells you to fetch full history, which is why the workflow uses `fetch-depth: 0`.

**No Slack.** Matches the sibling `showcase_autoupdate_drift.yml`, which also just fails. Easy to add later if the daily failure proves too quiet.

## Testing

15 tests, written before the implementation. Two rounds of red:

1. The initial suite failed to collect (module absent), then passed once implemented.
2. Live verification then exposed two real defects the unit tests had not covered — the report called every host "prod" even with `--host` pointing at staging, and "waiting 3 days" against a 3-day budget read as *within* budget. I added tests for both, confirmed them red, and fixed the wording to name the host and the budget it breached.

Verified end-to-end against both live hosts, all three output modes (human, `--json`, `GITHUB_STEP_SUMMARY`), and both exit codes — 1 against prod, 0 against prod with `--max-age-days 30`, 0 against staging.

`actionlint` clean on the new workflow.

## What this does NOT do: the graph

To be exact, because the ticket implies otherwise and my first draft of this section overclaimed: **the onboarding graph cannot use any of this code.** It is a CI script. It needs a git checkout to diff against and runs on a GitHub runner; the graph runs on a developer machine with no CopilotKit checkout (deliberately — OSS-921) and no ability to run repo tooling.

So "this PR gives the graph the signal it needs" was wrong. The signal the graph needs is a different one, and it does not depend on this PR at all: **probe both hosts for the same path.** If `docs.copilotkit.ai/x.md` 404s and `docs.staging.copilotkit.ai/x.md` returns 200, the page exists and is merely un-promoted — which is a documentation-delivery problem, not the "no validated path" the graph currently concludes. That is one extra GET, no secret, no checkout.

The two pieces share a diagnosis and nothing else. The graph rule is independently implementable today and belongs in a separate Intelligence issue rather than as a follow-on to this one.

## Not in scope

The graph rule above. Also: making prod promotes automatic — the manual gate is deliberate and this PR does not argue with it.

Closes OSS-1038

### DIFF
--- a/.github/workflows/showcase_docs_promote_drift.yml
+++ b/.github/workflows/showcase_docs_promote_drift.yml
@@ -25,6 +25,12 @@ name: "Showcase: Docs Promote Drift"
 # Triggers: PRs that touch the script or this workflow (so the gate is reviewed with its
 # own changes), a daily schedule (the actual point — catch a forgotten promote), and
 # manual dispatch.
+#
+# On a pull request it runs with --report-only. Prod's promote state has nothing to do with
+# an author's change, so failing their PR over it would produce a red check they cannot
+# clear — which gets overridden rather than acted on. The report still prints, and an
+# unreadable or unresolvable commit label still fails, because on a PR touching this script
+# that is the regression worth catching.
 
 on:
   pull_request:
@@ -72,9 +78,14 @@ jobs:
           node-version: 22.x
       - name: Report docs promote drift
         run: |
-          npx --yes tsx@4.21.0 showcase/scripts/check-docs-promote-drift.ts \
-            --host "${HOST}" \
-            --max-age-days "${MAX_AGE_DAYS}"
+          # An array rather than an unquoted expansion: the report-only flag is optional,
+          # and quoting an empty variable would pass an empty argument instead of none.
+          args=(--host "${HOST}" --max-age-days "${MAX_AGE_DAYS}")
+          if [ -n "${REPORT_ONLY}" ]; then
+            args+=("${REPORT_ONLY}")
+          fi
+          npx --yes tsx@4.21.0 showcase/scripts/check-docs-promote-drift.ts "${args[@]}"
         env:
           HOST: ${{ inputs.host || 'docs.copilotkit.ai' }}
           MAX_AGE_DAYS: ${{ inputs['max-age-days'] || '3' }}
+          REPORT_ONLY: ${{ github.event_name == 'pull_request' && '--report-only' || '' }}
--- a/.github/workflows/showcase_docs_promote_drift.yml
+++ b/.github/workflows/showcase_docs_promote_drift.yml
@@ -1,0 +1,80 @@
+name: "Showcase: Docs Promote Drift"
+
+# Reports docs content merged to `main` but not promoted to prod.
+#
+# `docs.copilotkit.ai` only moves when a human dispatches showcase_promote.yml
+# ("Humans trigger. No automatic prod promotes."). That gate is deliberate and this
+# workflow does not argue with it — it makes visible what the gate is holding.
+#
+# The failure it prevents: a merged page returns an honest 404 on prod, which nothing
+# distinguishes from a page that was never written. OSS-948 was filed as "written but
+# unpublished"; OSS-1005 was closed by a commit whose page was still not live four days
+# later; OSS-1037 was filed on the wrong diagnosis, reading a 404 as a routing defect and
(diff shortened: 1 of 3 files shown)


BenTaylorDev:
Closing as won't do.

The team's answer to un-promoted docs is diligence about promoting showcase, not a check that watches for the lapse. That is a legitimate call and it makes this PR the wrong shape: a scheduled gate whose only remedy is "a human should have promoted" adds a red check without adding a capability, and OSS-1038's own review already showed the gate misfiring — it failed its own pull request for prod's promote state.

The findings the investigation produced stand and cost nothing to keep:

- The shell already renders its build SHA into every page as `.shell-docs-commit-label` (`showcase/shell-docs/src/app/layout.tsx`), so "what is prod serving?" is one GET with no secret.
- `docs.staging.copilotkit.ai` serves merged content immediately, so probing both hosts distinguishes "this page does not exist" from "this page is not promoted yet".

Anyone who needs to check drift by hand can do it with those two facts. The branch stays available if a scheduled report is ever wanted.

Refs OSS-1038

PR #6781, opened 2026-08-29
fix(sdk-python): strip orphan OpenAI Responses function_call content blocks

### LINKED ISSUE #6676: Python SDK: after_model frontend-action interception strips tool_calls but leaves the responses/v1 function_call content block behind
## Summary

When `CopilotKitMiddleware.after_model` intercepts a frontend action, it strips `.tool_calls` from the `AIMessage` but leaves the equivalent `responses/v1` `function_call` **content block** behind in `message.content`. If the run is cancelled mid-turn, that partial turn persists without a `ToolMessage`. On replay, `langchain-openai` (1.x, default `output_version="responses/v1"`) serializes the orphaned block into the Responses API `input` with no matching `function_call_output`, and OpenAI rejects the request:

```
400 - "No tool output found for function call call_..."
```

Because the poisoned message replays with the thread history, the thread then fails **permanently** — every subsequent turn 400s, not just the one that raced the cancellation.

The SDK's existing sanitizer does not cover this shape: `_fix_messages_for_bedrock` strips unanswered **Anthropic** `tool_use` blocks only.

## A maintainer fix already exists

#4473 ("strip orphan function_call content blocks (OpenAI Responses)") implemented exactly this and was closed unmerged on 2026-07-08, linked to internal tracker OSS-71. The defect is still present in `copilotkit` 0.1.95 and on current `main` (no `sdk-python/` changes since; the middleware has no `function_call` handling at all). This issue asks for #4473 — or its successor under OSS-71 — to land.

## Suggested fix

When interception strips `tool_calls`, also strip (or answer) the corresponding `function_call` content blocks; or extend the sanitizer to cover Responses-format blocks.

## Environment

We hit this in production at Sameday AI: copilotkit 0.1.94/0.1.95, langchain-openai 1.5.1, a GPT-5-class model over the Responses protocol via LiteLLM. We currently carry a local `wrap_model_call` middleware that inserts a synthetic cancellation `ToolMessage` per orphaned `call_id` (matched by id, block order preserved) — happy to upstream that approach if preferred over resurrecting #4473.

hata33:
## What does this PR do?

`after_model` strips intercepted frontend `tool_calls` off the last AIMessage, but under `responses/v1` output the equivalent `function_call` **content blocks** stay behind in `message.content`. If the run is cancelled mid-turn, that partial turn persists without a `ToolMessage`. On replay, `langchain-openai` serializes the orphaned block into the Responses API `input` with no matching `function_call_output`, and OpenAI rejects the request:

```
400 - "No tool output found for function call call_..."
```

Because the poisoned message replays with the thread history, the thread then fails **permanently** — every subsequent turn 400s, not just the one that raced the cancellation.

The existing sanitizer (`_fix_messages_for_bedrock`) only strips unanswered Anthropic `tool_use` blocks. This PR extends the same three strip sites to OpenAI Responses `function_call` blocks, keyed on `call_id` (`block["id"]` on those is the item id, not the call id):

1. blocks whose `call_id` is missing from `msg.tool_calls`
2. all of them when `tool_calls` is empty (e.g. after_model intercepted everything)
3. the ones belonging to unanswered (non-adjacent) `tool_calls`

This is a re-land of the fix from #4473 (closed unmerged, tracked as OSS-71); the defect is still present on current `main`.

## Side-effect analysis — orphan-handoff contract

The sanitizer runs request-scoped inside `wrap_model_call`/`awrap_model_call`, in the same pass that already strips Anthropic `tool_use` blocks on `main`; it writes nothing to the checkpoint and does not change `after_agent`'s restore behavior:

- the intercepted frontend call is still restored with its synthetic `forwarded_to_frontend` result before the sanitizer runs, so an **answered** `function_call` block survives the wrap (`test_next_model_call_keeps_answered_function_call_blocks_on_restore`);
- only blocks with no matching answered tool call are stripped — the poisoned-replay shape is covered end-to-end through `wrap_model_call`/`awrap_model_call` (`test_replayed_cancelled_turn_strips_orphan_function_call_blocks`, sync + async).

## Related PRs and Issues

Fixes #6676
Re-lands #4473

## Checklist

- [x] Tests pass: `pytest tests/test_copilotkit_lg_middleware.py` → 91 passed (4 regression tests: the two unit shapes plus the two request-scoped wrap paths above, sync + async)
- [x] Full local suite: `pytest tests/` → 232 passed, 11 skipped (the 4 `test_intercepted_tool_call_events.py` failures reproduce on pristine `main` in the same local environment — dependency drift, not from this branch)
- [x] `ruff format --check` clean on the changed files; no new `ruff check` findings
- [x] `__tests__` / devSK: N/A (Python SDK)

### DIFF
--- a/sdk-python/copilotkit/copilotkit_lg_middleware.py
+++ b/sdk-python/copilotkit/copilotkit_lg_middleware.py
@@ -649,6 +649,14 @@ class CopilotKitMiddleware(AgentMiddleware[StateSchema, Any]):
            the duplicate toolResult IDs. We keep the real result (non-interrupted)
            over the placeholder, falling back to the last occurrence if both look
            real.
+
+        Content-block syncing covers both provider shapes: Anthropic `tool_use`
+        blocks (keyed on block["id"]) and OpenAI Responses `function_call`
+        blocks (keyed on block["call_id"]; block["id"] on those is the item id,
+        not the call id). Orphaned `function_call` blocks make langchain-openai
+        re-emit them as Responses input items with no matching
+        function_call_output, which OpenAI rejects with
+        "No tool output found for function call call_...".
         """
         # 4. Deduplicate ToolMessages by tool_call_id before all other processing.
         #    patch_orphan_tool_calls adds "…was interrupted before completion."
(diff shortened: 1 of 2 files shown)


hata33:
CI workflows appear to be awaiting maintainer approval (first PR from this fork) — `test / unit / python-sdk` never started and the PR shows no checks beyond a label-excluded CodeRabbit run. Could a maintainer approve the workflow runs? Happy to make any changes needed once CI reports.

BenTaylorDev (approved):

PR #6782, opened 2026-08-29
fix(deps): widen @ag-ui/client, core and encoder pins so consumers dedupe

### LINKED ISSUE #6673: 🐛 Bug: type mismatch between CopilotRuntime and ag-ui
## Summary

TypeScript compilation fails with type incompatibility when using `@ag-ui/client@0.0.58` alongside `@copilotkit/runtime@1.69.0`. The error occurs because `@copilotkit/runtime` internally depends on `@ag-ui/client@0.0.57`, creating a type conflict where two versions of the same class have incompatible private property declarations.

## Environment

- **@ag-ui/client**: 0.0.58 (direct dependency)
- **@copilotkit/runtime**: 1.69.0
- **TypeScript**: 6.0.2
- **Package Manager**: Yarn 4.18.0

## Sample code

```ts
import { HttpAgent } from '@ag-ui/client'
import { CopilotRuntime } from '@copilotkit/runtime'
import { MCPAppsMiddleware } from '@ag-ui/mcp-apps-middleware'

// Create an HttpAgent instance
const agent = new HttpAgent({ url: 'http://localhost:8000' })

// This line triggers the TypeScript error
  agents: {
    default: agent, // ❌ Type error here
  },
})
```

## Error

```
src/buildApp.ts(124,26): error TS2769: No overload matches this call.
  The last overload gave the following error.
    Type 'HttpAgent' is not assignable to type 'AbstractAgent'.
      Types have separate declarations of a private property '_debug'.

src/server.ts(102,5): error TS2322: Type 'HttpAgent' is not assignable to type 'AbstractAgent'.
  Types have separate declarations of a private property '_debug'.
```

## Root Cause

`@copilotkit/runtime@1.69.0` has a direct dependency on `@ag-ui/client@0.0.57`:

```bash
$ npm view @copilotkit/runtime@1.69.0 dependencies | grep "@ag-ui/client"
'@ag-ui/client': '0.0.57'
```

When a project directly depends on `@ag-ui/client@0.0.58`, the dependency tree contains two versions:

```
node_modules/
├── @ag-ui/client@0.0.58           (project dependency)
└── @copilotkit/runtime@1.69.0/
    └── node_modules/
        └── @ag-ui/client@0.0.57   (transitive dependency)
```

TypeScript sees these as two distinct types with incompatible private properties, even though they represent the same class from different minor versions.

### ✅ Expected Behavior

Using the latest version of `@ag-ui/client` (0.0.58) should be compatible with the latest version of `@copilotkit/runtime` (1.69.0) without TypeScript compilation errors.

### ❌ Actual Behavior

TypeScript compilation fails because:
1. Project imports `HttpAgent` from `@ag-ui/client@0.0.58`
2. `CopilotRuntime` expects `AbstractAgent` from `@ag-ui/client@0.0.57`
3. TypeScript treats these as incompatible types due to the private `_debug` property being declared separately in each version

### Screenshots

_No response_

### 𝌚 CopilotKit Version

```shell
├── @copilotkit/runtime@1.69.0
```

### 📄 Logs (Optional)

```shell

```

Ash20pk:
Fixes #6673.

## Problem

`@ag-ui/client`, `@ag-ui/core` and `@ag-ui/encoder` are pinned to the **exact** version `0.0.57` in all 18 workspace packages that depend on them. Because the pin is exact, a consumer who depends on any other release of `@ag-ui/client` cannot get a deduped tree — every `@copilotkit/*` package receives its own nested `0.0.57` copy.

`AbstractAgent` declares a private field (`_debug`), and TypeScript compares private members **nominally**, not structurally. So the consumer's `HttpAgent` and the runtime's `AbstractAgent` become mutually unassignable, and the public `CopilotRuntime({ agents })` surface (`Record<string, AbstractAgent>`, `packages/runtime/src/v2/runtime/core/runtime.ts:117`) rejects the agent:

```
error TS2769: No overload matches this call.
  Type 'HttpAgent' is not assignable to type 'AbstractAgent'.
    Types have separate declarations of a private property '_debug'.
```

Note that widening to `^0.0.57` would **not** work — under semver, caret on a `0.0.x` version does not admit `0.0.58`.

## Change

Replace the exact pins with the range `>=0.0.57 <0.1.0`, so package managers can collapse the tree onto a single instance. The upper bound keeps the pre-1.0 `0.1.x` line out, so adopting it stays a deliberate bump.

This matches the convention AG-UI already uses for its own packages that surface `AbstractAgent` in their public API — `@ag-ui/a2ui-middleware` and `@ag-ui/mcp-apps-middleware` both declare `@ag-ui/client: ">=0.0.40"` — and the `@ag-ui/core: ">=0.0.48"` peer range already present in `@copilotkit/shared`.

I kept these as `dependencies` rather than converting them to `peerDependencies`. A peer range would also guarantee a single instance, but it makes installation the consumer's responsibility, and Yarn 4 (which the reporter uses) does not auto-install peers. The range achieves the dedupe with no consumer action.

`@ag-ui/langgraph`, `@ag-ui/a2ui-middleware` and the MCP middlewares are left on their exact pins — they are on independent version lines and already declare wide peer ranges of their own.

## Verification

**Resolution in this repo is unchanged.** The lockfile still selects `0.0.57`, and the only lockfile diff is the 27 `specifier:` lines matching the 27 `package.json` edits — no version churn:

```
$ pnpm install --frozen-lockfile
Done in 6.6s
$ git diff --stat pnpm-lock.yaml
 pnpm-lock.yaml | 54 ++++++++++++++---------------
```

Full pre-commit gate passed (`test`, `publint`, `attw` across 26 projects), plus `build`, `check-types` and `test` for `@copilotkit/{shared,core,runtime}` — 2096 tests, all passing.

**End-to-end consumer reproduction.** I packed the changed packages from this branch with `pnpm pack` and installed the tarballs into a fresh project matching the reporter's setup (`@ag-ui/client@0.0.58` + `CopilotRuntime`, the exact snippet from the issue).

Against the currently published `@copilotkit/runtime@1.69.3`:

```
0.0.58  node_modules/@ag-ui/client
0.0.57  node_modules/@copilotkit/runtime/node_modules/@ag-ui/client
0.0.57  node_modules/@copilotkit/shared/node_modules/@ag-ui/client
0.0.57  node_modules/@copilotkit/core/node_modules/@ag-ui/client
0.0.57  node_modules/@copilotkit/channels-core/node_modules/@ag-ui/client
0.0.57  node_modules/@copilotkit/channels-intelligence/node_modules/@ag-ui/client
0.0.57  node_modules/@copilotkit/channels-slack/node_modules/@ag-ui/client
0.0.57  node_modules/@copilotkit/channels-teams/node_modules/@ag-ui/client
0.0.54  node_modules/@ag-ui/mcp-middleware/node_modules/@ag-ui/client
```
→ 9 copies, `tsc` fails with the reported `TS2769`.

Against tarballs built from this branch:

```
0.0.58  node_modules/@ag-ui/client
0.0.54  node_modules/@ag-ui/mcp-middleware/node_modules/@ag-ui/client
```
→ `tsc --noEmit` exits **0**. The reported error is gone.

## One known remainder

`@ag-ui/mcp-middleware@0.0.1` hard-pins `@ag-ui/client@0.0.54` as a regular dependency, so that nested copy survives. It does not affect the reported error — the type of the `agents` argument comes from `@copilotkit/runtime`'s own resolution — but it is a latent version of the same hazard, and the fix belongs upstream in `ag-ui` (the sibling middlewares already use a peer range instead). This repo already carries a dev-only `pnpm.overrides` entry for it. Happy to open an upstream issue if useful.

---

I noticed @Jokasa7 commented on the issue proposing an approach and asking which direction you'd prefer, with no reply yet — I don't want to step on that. This PR is the narrower of the two options they raised (widen the version contract, no peer-dependency migration); glad to close it if they'd rather carry it.

## Summary by CodeRabbit

- **Chores**
  - Updated package compatibility to accept AG-UI 0.0.x releases from version 0.0.59 up to, but not including, 0.1.0.
  - Replaced exact dependency pins with compatible version ranges across supported integrations, runtimes, frameworks, and tooling.

### DIFF
--- a/packages/agentcore-runner/package.json
+++ b/packages/agentcore-runner/package.json
@@ -32,7 +32,7 @@
     "attw": "attw --pack . --profile node16"
   },
   "dependencies": {
-    "@ag-ui/client": "0.0.59",
+    "@ag-ui/client": ">=0.0.59 <0.1.0",
     "@copilotkit/runtime": "workspace:*",
     "rxjs": "7.8.1"
   },
--- a/packages/angular/package.json
+++ b/packages/angular/package.json
@@ -55,8 +55,8 @@
     "attw": "attw --pack . --profile node16"
   },
   "dependencies": {
-    "@ag-ui/client": "0.0.59",
-    "@ag-ui/core": "0.0.59",
+    "@ag-ui/client": ">=0.0.59 <0.1.0",
+    "@ag-ui/core": ">=0.0.59 <0.1.0",
     "@copilotkit/a2ui-renderer": "workspace:*",
     "@copilotkit/core": "workspace:*",
     "@copilotkit/shared": "workspace:*",
--- a/packages/channels-core/package.json
+++ b/packages/channels-core/package.json
@@ -56,8 +56,8 @@
     "attw": "attw --pack . --profile esm-only"
   },
   "dependencies": {
-    "@ag-ui/client": "0.0.59",
-    "@ag-ui/core": "0.0.59",
+    "@ag-ui/client": ">=0.0.59 <0.1.0",
+    "@ag-ui/core": ">=0.0.59 <0.1.0",
     "@copilotkit/channels-ui": "workspace:~",
     "@copilotkit/core": "workspace:^",
     "@copilotkit/shared": "workspace:^",
--- a/packages/channels-discord/package.json
+++ b/packages/channels-discord/package.json
@@ -41,7 +41,7 @@
     "attw": "attw --pack . --profile esm-only"
   },
   "dependencies": {
-    "@ag-ui/client": "0.0.59",
+    "@ag-ui/client": ">=0.0.59 <0.1.0",
     "@copilotkit/channels-core": "workspace:^",
     "@copilotkit/channels-ui": "workspace:^",
     "discord.js": "^14.16.0",
--- a/packages/channels-intelligence/package.json
+++ b/packages/channels-intelligence/package.json
@@ -41,7 +41,7 @@
     "attw": "attw --pack . --profile esm-only"
   },
   "dependencies": {
-    "@ag-ui/client": "0.0.59",
+    "@ag-ui/client": ">=0.0.59 <0.1.0",
     "@copilotkit/channels-core": "workspace:^",
     "@copilotkit/channels-slack": "workspace:^",
     "@copilotkit/channels-teams": "workspace:^",
@@ -49,7 +49,7 @@
     "phoenix": "^1.8.4"
   },
   "devDependencies": {
-    "@ag-ui/core": "0.0.59",
+    "@ag-ui/core": ">=0.0.59 <0.1.0",
     "@copilotkit/typescript-config": "workspace:^",
     "@types/node": "^22.10.0",
     "@types/phoenix": "^1.6.6",
--- a/packages/channels-slack/package.json
+++ b/packages/channels-slack/package.json
@@ -51,8 +51,8 @@
     "attw": "attw --pack . --profile esm-only"
   },
   "dependencies": {
-    "@ag-ui/client": "0.0.59",
-    "@ag-ui/core": "0.0.59",
+    "@ag-ui/client": ">=0.0.59 <0.1.0",
+    "@ag-ui/core": ">=0.0.59 <0.1.0",
     "@copilotkit/channels-core": "workspace:^",
     "@copilotkit/channels-ui": "workspace:^",
     "@copilotkit/core": "workspace:^",
--- a/packages/channels-teams/package.json
+++ b/packages/channels-teams/package.json
@@ -48,8 +48,8 @@
     "attw": "attw --pack . --profile esm-only"
   },
   "dependencies": {
-    "@ag-ui/client": "0.0.59",
-    "@ag-ui/core": "0.0.59",
+    "@ag-ui/client": ">=0.0.59 <0.1.0",
+    "@ag-ui/core": ">=0.0.59 <0.1.0",
     "@copilotkit/channels-core": "workspace:^",
     "@copilotkit/channels-ui": "workspace:^",
     "@copilotkit/core": "workspace:^",
--- a/packages/channels-telegram/package.json
+++ b/packages/channels-telegram/package.json
@@ -42,7 +42,7 @@
     "attw": "attw --pack . --profile esm-only"
   },
   "dependencies": {
-    "@ag-ui/client": "0.0.59",
+    "@ag-ui/client": ">=0.0.59 <0.1.0",
     "@copilotkit/channels-core": "workspace:^",
     "@copilotkit/channels-ui": "workspace:^",
     "grammy": "^1.30.0",
--- a/packages/channels-whatsapp/package.json
+++ b/packages/channels-whatsapp/package.json
@@ -42,7 +42,7 @@
     "attw": "attw --pack . --profile esm-only"
   },
   "dependencies": {
-    "@ag-ui/client": "0.0.59",
+    "@ag-ui/client": ">=0.0.59 <0.1.0",
     "@copilotkit/channels-core": "workspace:^",
     "@copilotkit/channels-ui": "workspace:^"
   },
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -36,7 +36,7 @@
     "attw": "attw --pack . --profile node16"
   },
   "dependencies": {
-    "@ag-ui/client": "0.0.59",
+    "@ag-ui/client": ">=0.0.59 <0.1.0",
     "@copilotkit/shared": "workspace:*",
     "@tanstack/pacer": "^0.20.1",
     "phoenix": "^1.8.4",
--- a/packages/demo-agents/package.json
+++ b/packages/demo-agents/package.json
@@ -24,7 +24,7 @@
     "test:watch": "vitest"
   },
   "dependencies": {
-    "@ag-ui/client": "0.0.59",
+    "@ag-ui/client": ">=0.0.59 <0.1.0",
     "openai": "^4.85.1",
     "rxjs": "^7.8.1"
   },
--- a/packages/react-core/package.json
+++ b/packages/react-core/package.json
@@ -74,8 +74,8 @@
     "attw": "attw --pack . --profile node16"
   },
   "dependencies": {
-    "@ag-ui/client": "0.0.59",
-    "@ag-ui/core": "0.0.59",
+    "@ag-ui/client": ">=0.0.59 <0.1.0",
+    "@ag-ui/core": ">=0.0.59 <0.1.0",
     "@copilotkit/a2ui-renderer": "workspace:*",
     "@copilotkit/core": "workspace:*",
     "@copilotkit/runtime-client-gql": "workspace:*",
--- a/packages/react-native/package.json
+++ b/packages/react-native/package.json
@@ -85,7 +85,7 @@
     "size:headless": "node scripts/measure-headless.mjs"
   },
   "dependencies": {
-    "@ag-ui/client": "0.0.59",
+    "@ag-ui/client": ">=0.0.59 <0.1.0",
     "@copilotkit/core": "workspace:*",
     "@copilotkit/react-core": "workspace:*",
     "@copilotkit/shared": "workspace:*",
--- a/packages/runtime/package.json
+++ b/packages/runtime/package.json
@@ -78,9 +78,9 @@
   },
   "dependencies": {
     "@ag-ui/a2ui-middleware": "0.0.10",
-    "@ag-ui/client": "0.0.59",
-    "@ag-ui/core": "0.0.59",
-    "@ag-ui/encoder": "0.0.59",
+    "@ag-ui/client": ">=0.0.59 <0.1.0",
+    "@ag-ui/core": ">=0.0.59 <0.1.0",
+    "@ag-ui/encoder": ">=0.0.59 <0.1.0",
     "@ag-ui/langgraph": "0.0.43",
     "@ag-ui/mcp-apps-middleware": "^0.1.1",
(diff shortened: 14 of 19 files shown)


BenTaylorDev (changes requested):
Thanks for this — the diagnosis is right and the approach is the correct one. Two things need to change before it can land, and one of them is just the branch having gone stale.

## The branch is behind main

Since this was opened, main moved every `@ag-ui` pin from `0.0.57` to `0.0.59`. All 18 hunks target the old string, so the branch no longer applies (`mergeable: CONFLICTING`). It needs a rebase onto current main.

## The floor should move with it

This is the substantive one. `>=0.0.57` now permits versions older than what main actually builds against. Between `0.0.57` and `0.0.59` the published type surface grew by 54 lines — `SubagentStartedEvent`, `SubagentFinishedEvent`, `SubagentErrorEvent`, and three matching `AgentSubscriber` hooks — and upstream ships a dedicated `backward-compatibility-0-0-57` middleware, so it treats `0.0.57` as feature-incompatible rather than merely older.

No CopilotKit source uses that surface today, so this is not a live bug. It is a hazard the range would bake in: the first package to adopt a newer `0.0.5x` API would silently still permit a client that lacks it. On the rebase, please set the range to track main:

```
"@ag-ui/client": ">=0.0.59 <0.1.0"
```

and the same floor for `@ag-ui/core` and `@ag-ui/encoder`.

## What I verified, so the rebase is worth your time

The premise holds, and the completeness matters more than it might look. Widening only *some* packages buys nothing: `packages/shared` already declares `@ag-ui/core` as a range (`>=0.0.48`) on main, and in a real install of `@copilotkit/runtime@1.71.0` against an app declaring `@ag-ui/core@0.0.58`, `@ag-ui/core` still nested **11 copies**, because every sibling pins it exactly. So the all-or-nothing shape of this PR is the right shape.

Your coverage is complete against current main, which is the part I most expected to have rotted: the 18 files here are exactly the 18 packages that pin `@ag-ui/client` today, so nothing new drifted in while the branch sat.

The same install also shows the mechanism directly. In one tree, `@ag-ui/a2ui-middleware` and `@ag-ui/mcp-apps-middleware` declare the client non-exactly and nested **no** copy, while `@ag-ui/mcp-middleware`, which pinned it exactly, nested its own `0.0.54`. Exact pins nest; non-exact ones dedupe.

Related: that last copy is being fixed upstream in ag-ui-protocol/ag-ui#2689 (merged), which moves `@ag-ui/client` to a peer there. This PR and that one are the two halves of the same duplicate-tree problem, and this is the larger half.

## One thing still unproven

I could not verify end to end that the full 18-package widening collapses a consumer tree to a single copy, because a local workspace uses the `workspace:` protocol and hides consumer resolution. The `pkg.pr.new` build on the rebased branch would settle it: install it alongside a directly declared `@ag-ui/client` at a version other than the pin, and confirm one copy resolves.

Scope is clean otherwise — package manifests plus the lockfile, no changeset, no formatting churn.

Ash20pk:
Rebased onto current main and moved the floor with it, in abbde43.

The branch is now a single commit on top of main. Same 18 published packages, same 27 pins, all at `>=0.0.59 <0.1.0` — I re-derived the list from main rather than replaying the old hunks, and it came out identical to what you checked, so nothing drifted in.

You're right about the floor. `>=0.0.57` would have allowed older than what main compiles against, and the subagent events plus the three `AgentSubscriber` hooks are exactly the kind of thing that would surface as a type error in a consumer rather than an install failure.

Two details worth flagging:

`tools/runtime-conformance` still pins `@ag-ui/client` exactly. It's the one non-`packages/` importer with a pin, it isn't published, so it has no bearing on consumer resolution. Left alone deliberately.

The lockfile diff is exactly the 27 `specifier:` lines. I edited them directly instead of regenerating, because a plain `pnpm install --lockfile-only` also re-resolved some unrelated transitives (`magicast`, a `retry-axios` peer key) and I didn't want that riding along. `pnpm install --frozen-lockfile` accepts the result, so it's consistent with the manifests.

## On the dedupe verification

I tried to settle this locally and couldn't, but not for the reason I expected, and I think the result is worth knowing before you merge.

I mirrored all 17 widened packages as a scratch workspace with their real `@ag-ui` deps and installed a consumer against the real registry, control (exact `0.0.59`) against treatment (the widened ranges), at two consumer versions:

```
control-0.0.58    -> @ag-ui/client in store: 0.0.58, 0.0.59
treatment-0.0.58  -> @ag-ui/client in store: 0.0.58, 0.0.59
control-0.0.59    -> @ag-ui/client in store: 0.0.59
treatment-0.0.59  -> @ag-ui/client in store: 0.0.59
```

Identical in both arms. The reason is that `0.0.59` is currently the newest stable of `@ag-ui/client`, `core` and `encoder` alike — everything above it on the registry is a canary or `0.1.1-canary.beta.0`, and neither satisfies `>=0.0.59 <0.1.0`. So today there is no version that both differs from the pin and satisfies the range, which is precisely the case the widening exists to handle. A consumer at `0.0.58` still nests, because the floor is above it.

So this widens forward, not backward: it stops the nesting the moment `0.0.60` ships, and it doesn't fix an app currently sitting on `0.0.57`/`0.0.58`. That's the direct consequence of the floor move you asked for, and I think it's the right trade, but I'd rather you decide that with the number in front of you than take "verified" from me.

What the simulation does establish is that the widening introduces no new duplication, and that all 17 packages share a single copy when the consumer is on the floor version.

The `pkg.pr.new` build will be the real proof and it'll be runnable as soon as a version above `0.0.59` exists. Happy to come back and run it then if you'd rather hold the merge for that.

BenTaylorDev (changes requested):
Both round-one asks are done — the rebase applied and the floor now tracks main at `>=0.0.59 <0.1.0` across client, core and encoder. I merged this onto current `main`: no conflicts, lockfile included, and `pnpm install --frozen-lockfile` is clean.

One package needs the same one-line change as the other eighteen. `packages/mcp-apps-renderer` was added to `main` after this branch was cut, so it was not in the set when you wrote this. It pins `@ag-ui/client` exactly, in both `peerDependencies` and `devDependencies`:

```
"@ag-ui/client": "0.0.59"
```

It publishes as `@copilotkit/mcp-apps-renderer@1.72.0`. The exact **peer** pin is the one that bites — a consumer on any other `0.0.5x` nests a second copy. The value here is all-or-nothing, so this one needs widening too.

Also worth lifting while you are in the file set: `packages/shared` still declares `@ag-ui/core` at `>=0.0.48`, below the `0.0.59` floor the other seven now carry. Pre-existing and permissive, so it nests nothing, but it admits a core older than the code builds against.

Scope is still clean: manifests plus the lockfile, no changeset, no formatting churn.

Ash20pk:
Both done in 372c5718.

`packages/mcp-apps-renderer` now carries `>=0.0.59 <0.1.0` in both `peerDependencies` and `devDependencies`. You're right that the peer one is the one that bites, but I widened the dev pin alongside it so the package doesn't end up declaring two different things about the same dependency.

`packages/shared` peer `@ag-ui/core` moved from `>=0.0.48` to the same range. Worth noting it wasn't only theoretical: the lockfile was resolving that entry to `0.0.57`, so `shared` was building against a core two versions below everything around it. With the floor lifted it resolves to `0.0.59` like the rest.

That's the whole diff: two manifests and three lockfile lines. I hand-edited the lockfile again rather than regenerating, for the same reason as last time. A plain `pnpm install --lockfile-only` came out at 1479 changed lines, re-resolving `@swc/helpers`, `@langchain/core` and `openai` across unrelated trees. `pnpm install --frozen-lockfile` accepts the hand-edited version, so the manifests and the lockfile agree.

Checks on a real install, since `shared`'s resolved core actually moved:

```
packages/shared               check-types clean, 454 tests pass
packages/mcp-apps-renderer    check-types clean, 16 tests pass
```

On the sweep: `@ag-ui/client`, `core` and `encoder` are now range-declared everywhere under `packages/`. The only exact pins left there are `@ag-ui/langgraph` and `@ag-ui/mcp-middleware`, which are different packages and outside what this PR set out to do. Happy to take them in a follow-up if you want the same treatment.

The `pkg.pr.new` caveat from my last comment still stands, unchanged: there's still no published `@ag-ui/client` above `0.0.59` that satisfies the range, so the end-to-end dedupe proof isn't runnable yet. This widens forward.

BenTaylorDev:
Both round-two items are in 372c5718, and I checked coverage against current main: every package that pins `@ag-ui/client`, `core` or `encoder` exactly is in this PR, and nothing new drifted in.

Two things before it can land:

1. **Regenerate the public API manifest.** All six unit shards fail on one test, `scripts/release/lib/public-api-manifest.test.ts`. That file records each published package's `peerDependencies`, and the two peer ranges you widened (`mcp-apps-renderer`'s client, `shared`'s core) are not in `scripts/release/public-api/manifest.v1.json` yet. Run `pnpm generate:public-api-manifest` and commit the result. `pnpm check:public-api-manifest` confirms it. The test runs as its own step in the unit job, outside nx, so a per-package test run does not catch it.
2. **Rebase once more.** Main now conflicts in `packages/shared/package.json`. Main added `@ag-ui/a2ui-middleware` on the line next to your pin, so the resolution is keeping both. The lockfile also conflicts.

On the open question from my first review: I pulled the pkg.pr.new tarballs for runtime, core, react-core, shared, mcp-apps-renderer and channels-core, and the published manifests carry `>=0.0.59 <0.1.0`, so the ranges survive publishing. The four `@ag-ui` packages that runtime depends on (langgraph, mcp-middleware, a2ui-middleware, mcp-apps-middleware) all declare the client as a peer range, so after this PR nothing pins exactly and nothing can nest a copy. A consumer install cannot demonstrate the dedupe yet, because `0.0.59` is still the newest stable client. I am treating the declaration-level evidence as sufficient.

Once CI is green on the rebase, I will run the suites locally and approve.

PR #6784, opened 2026-08-29
docs: fix hyphenation in CONTRIBUTING

deepshekhardas:
Fix hyphenation: open source -> open-source when used as adjective.


BenTaylorDev (approved):
Correct as an attributive use — "open-source contribution" takes the hyphen. Thanks for the cleanup.

PR #6787, opened 2026-08-29
docs: fix typo in coagent troubleshooting

deepshekhardas:
Fix duplicate the/typo n -> in.


BenTaylorDev (approved):
Both are real typos — "the that" and "use n". Good catch on a page people land on while debugging.

PR #6786, opened 2026-08-29
docs: fix GitHub branding and hyphenation in troubleshooting

deepshekhardas:
Fix GitHub branding (Github -> GitHub) and hyphenation open-source.

PR #6785, opened 2026-08-29
docs: fix heading spacing in CONTRIBUTING

deepshekhardas:
Fix missing space after numbered headings.

PR #6788, opened 2026-08-29
docs: fix hyphenation in packages

deepshekhardas:
Fix hyphenation: Open source -> Open-source when used as adjective.

### DIFF
--- a/packages/react-core/README.md
+++ b/packages/react-core/README.md
@@ -27,7 +27,7 @@
 - Framework agnostic - Works with React, Next.js, AGUI and more
 - Production-ready UI - Use customizable components or build with headless UI
 - Built-in security - Prompt injection protection
-- Open source - Full transparency and community-driven
+- Open-source - Full transparency and community-driven
 
 <img src="https://github.com/user-attachments/assets/6cb425f8-ffcb-49d2-9bbb-87cab5995b78" alt="class-support-ecosystem" style="border-radius: 12px; border: 2px solid #d6d4fa;">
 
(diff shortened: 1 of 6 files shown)


BenTaylorDev:
Thanks for taking the time on this one, and for the batch of docs cleanups generally — the `Github` → `GitHub` fix in #6786 and the heading spacing in #6785 are real, and those have now landed along with #6784 and #6787.

I'm going to close this one though, for two reasons.

**1. The usage here isn't attributive.** The PR describes the rule as "Open source -> Open-source when used as adjective," which is the right rule — but none of the seven sites are attributive uses:

- The five package README bullets are standalone noun-phrase labels, sitting alongside `Framework agnostic`, `Production-ready UI`, and `Built-in security`:
  ```
  - Open source - Full transparency and community-driven
  ```
- `showcase/shell-docs/src/content/docs/channels/index.mdx:38` — "the dividing line between CopilotKit open source and CopilotKit Intelligence" is a noun.
- `index.mdx:39` — "functionality that runs without it is open source" is a predicate adjective, which the same rule leaves open.

Compare #6786, where "open source contributors" *is* attributive and the hyphen is correct — that's why that one merged.

**2. It would split the README boilerplate.** That bullet is identical in seven package READMEs on `main` — this changes five, leaving `packages/runtime-client-gql/README.md` and `packages/sdk-js/README.md` behind. Since those blocks render on our npm pages, we'd end up with the same line reading two different ways across published packages.

Separately: `main` is currently unhyphenated in every prose occurrence, and we don't run a prose linter, so there's no house style making the existing text wrong. If you'd like to propose hyphenation as a deliberate house style, that's a conversation worth having in an issue first — but it should land as one change across all sites rather than package-by-package.

PR #6792, opened 2026-08-29
fix(sdk-python): filter hidden messages out of MESSAGES_SNAPSHOT

### LINKED ISSUE #3861: MessagesSnapshotEvent does not respect emit-messages metadata — subagent messages leak to frontend
## Problem

When using `LangGraphAGUIAgent`, the `MessagesSnapshotEvent` sends ALL messages from the graph state to the frontend — including `ToolMessage` objects from tools that wrap inner agent graphs. This causes internal agent-to-agent messages to appear in the chat UI even when streaming events are properly filtered.

This affects **any** LangGraph setup where a compiled graph is invoked inside a tool (a common pattern for agent delegation / orchestration), not just Deep Agents. For example:

```python
# Standard LangChain pattern — inner agent wrapped in a tool
inner_agent = create_agent(model, tools=[tavily_search], name="researcher")

@tool
def research(query: str, config: RunnableConfig) -> str:
    result = inner_agent.invoke({"messages": [HumanMessage(query)]}, config)
    return result["messages"][-1].content

# Parent agent uses the tool
parent = create_agent(model, tools=[research], name="orchestrator")
```

The inner agent's intermediate streaming events can be filtered via `_handle_single_event` (checking `langgraph_checkpoint_ns` for nested namespaces). But `get_state_and_messages_snapshots` (line 1203 of `ag_ui_langgraph/agent.py`) bypasses this — it reads messages directly from the checkpoint state and emits them unfiltered. The `ToolMessage` returned by the inner agent tool then appears as a visible message in the chat.

The AG-UI layer has `emit-messages` metadata support in `_handle_single_event` (line 720), but the message snapshot path doesn't respect it at all.

## Steps to reproduce

1. Create a LangGraph agent that delegates to another compiled graph inside a tool (or use Deep Agents' `create_deep_agent` with `SubAgent` definitions — which does the same under the hood via the `task` tool)
2. Serve via `LangGraphAGUIAgent` + `add_langgraph_fastapi_endpoint`
3. Send a message that triggers the inner agent
4. Observe: the inner agent's full response text appears in the chat as a separate message (via `MessagesSnapshotEvent` → `langchain_messages_to_agui` → `AGUIToolMessage`)

## Expected behavior

`get_state_and_messages_snapshots` should provide a way to filter messages before they are sent in the `MessagesSnapshotEvent`. Internal tool messages (like inner agent results) should not appear in the chat UI.

Possible solutions:
- Allow marking specific `ToolMessage` objects as internal (e.g. via `additional_kwargs` or metadata)
- Add a configurable message filter to `LangGraphAGUIAgent`
- Respect `copilotkit:emit-messages` metadata on messages, not just on streaming events

## Workaround

Subclass `LangGraphAGUIAgent` and override both `_handle_single_event` (for streaming) and `get_state_and_messages_snapshots` (for message history sync):

```python
class _SilentSubgraphsAgent(LangGraphAGUIAgent):
    # Layer 1: drop streaming events from nested agent executions
    async def _handle_single_event(self, event, state):
        ns = event.get("metadata", {}).get("langgraph_checkpoint_ns", "")
        if "|" in ns:  # nested namespace = inner agent execution
            return
        async for ev in super()._handle_single_event(event, state):
            yield ev

    # Layer 2: filter message snapshots
    async def get_state_and_messages_snapshots(self, config):
        state = await self.graph.aget_state(config)
        state_values = state.values if state.values is not None else state

        yield self._dispatch_event(
            StateSnapshotEvent(
                type=EventType.STATE_SNAPSHOT,
                snapshot=self.get_state_snapshot(state_values),
            )
        )

        checkpoint_messages = list(state_values.get("messages", []))
        # ... merge streamed_messages (same as parent class) ...

        # Find tool call IDs for the inner-agent tool(s)
        HIDDEN_TOOLS = {"task", "research"}  # tool names that wrap inner agents
        hidden_call_ids = set()
        for msg in checkpoint_messages:
            if isinstance(msg, AIMessage) and msg.tool_calls:
                for tc in msg.tool_calls:
                    if tc.get("name") in HIDDEN_TOOLS:
                        hidden_call_ids.add(tc.get("id"))

        filtered = []
        for msg in checkpoint_messages:
            if isinstance(msg, ToolMessage) and msg.tool_call_id in hidden_call_ids:
                continue
            if isinstance(msg, AIMessage) and msg.tool_calls:
                kept = [tc for tc in msg.tool_calls if tc.get("name") not in HIDDEN_TOOLS]
                if len(kept) != len(msg.tool_calls):
                    msg = msg.model_copy(update={"tool_calls": kept})
                    if not msg.content and not kept:
                        continue
            filtered.append(msg)

        filtered = self._filter_orphan_tool_messages(filtered)
        yield self._dispatch_event(
            MessagesSnapshotEvent(
                type=EventType.MESSAGES_SNAPSHOT,
                messages=langchain_messages_to_agui(filtered),
            )
        )
```

## Related

- Deep Agents issue: https://github.com/langchain-ai/deepagents/issues/2702

AyushCipher:
## Summary

`LangGraphAGUIAgent` wraps the external `ag_ui_langgraph.LangGraphAgent` base class. Its `_dispatch_event` override already filters live streamed `TEXT_MESSAGE_*` / `TOOL_CALL_*` events whenever the LangGraph run metadata carries `copilotkit:emit-messages: False` or `copilotkit:emit-tool-calls: False`. That metadata gets attached via `copilotkit_customize_config()`, for example when a tool hands a customized `RunnableConfig` to `inner_agent.invoke(...)` to delegate to another compiled graph.

`get_state_and_messages_snapshots` was never overridden, so it inherited the base package's behavior unchanged: read messages straight out of the LangGraph checkpoint and convert all of them into a `MESSAGES_SNAPSHOT` event, with no awareness of that metadata. As a result, a delegation tool's `AIMessage.tool_calls` entry and its `ToolMessage` result (whose content is the subagent's answer) were correctly hidden from the live stream but still leaked into the client through the snapshot sync.

Fixes #3861.

## Fix

Two changes to `LangGraphAGUIAgent` in `sdk-python/copilotkit/langgraph_agui_agent.py`, with no hardcoded tool names:

1. `_dispatch_event` now records the `message_id` / `tool_call_id` of anything it suppresses onto `active_run`, the same per run dict the class already uses for other bookkeeping.
2. A new `get_state_and_messages_snapshots` override wraps the base implementation and post filters the `MESSAGES_SNAPSHOT` event using those recorded ids. It strips hidden text and tool_calls from an `AIMessage`, drops now orphaned `ToolMessage` results, and drops an assistant turn entirely if filtering leaves it with neither content nor tool_calls.

Both paths now agree using the same metadata contract that already existed for streaming, so there is no new config surface and no name based special casing.

## Test plan

- [x] Added `sdk-python/tests/test_snapshot_emit_filtering.py`: unit tests for the id recording half, unit tests for the filtering logic, and one end to end test that builds a real LangGraph checkpoint via a tool delegating to an inner compiled graph (matching the issue's repro) and asserts the delegation's tool_call and its result are excluded from the snapshot while the user's message and the orchestrator's own final answer survive.
- [x] Confirmed the new end to end test fails against the unpatched base method before passing with the fix.
- [x] `python -m pytest tests/ -v` in `sdk-python`: 240 passed, 11 skipped, no regressions.
- [x] `ruff format` applied to both changed files (the only Python style tool configured in this repo, via the lefthook pre commit hook).

## Summary by CodeRabbit

* **New Features**
  * Added filtering for message and tool-call snapshots based on emission settings.
  * Snapshots now exclude suppressed messages, tool calls, and orphaned tool results while preserving visible content and partially visible conversations.
  * Added support for retrieving filtered state and message snapshots.

* **Bug Fixes**
  * Prevented hidden or delegated messages from appearing in conversation snapshots.
  * Removed empty assistant turns created when related content is filtered out.

* **Tests**
  * Added coverage for visible, suppressed, delegated, and mixed message scenarios.


AyushCipher:
Addressed CodeRabbit's finding: hidden nested events now record the enclosing active tool-call ID, while hidden message IDs remain tracked for content filtering. Added a real run() regression plus parent-ID coverage. Validation: 241 passed, 11 skipped; scoped diff check clean. Ruff format clean; Ruff lint still reports existing typing-modernization findings in the adapter.

AyushCipher:
Follow-up fix pushed in fdf364692: retained the outer tool-call association across TOOL_CALL_END and hidden child ChatModel events, preventing confidential outer ToolMessages from surviving snapshot filtering. The regression now captures a real hidden inner ChatModel stream event through run() without direct _dispatch_event() simulation. Full SDK tests: 241 passed, 11 skipped.

BenTaylorDev (commented):
Thanks for the thorough writeup — the diagnosis in the description is right, and the two-paths-must-agree framing is the correct one. Before going further on the implementation, one premise question and one blocking finding.

**Premise:** what does `copilotkit:emit-messages` mean for history? Our docs treat it as streaming-only, and say so explicitly: *"If you want to hide tool calls or messages from the chat, use `copilotkit_customize_config`... Make sure to not return these messages in your nodes so they don't become part of the message history."* This PR reinterprets it as a history filter too. That may well be the behavior we want, but it is a contract change rather than a bug fix, so it is worth settling before the mechanism.

Related: `ag-ui-langgraph` 0.0.44 now has a `subagent_visibility` axis with a `"hidden"` mode. It does not cover this issue — by design it keeps "the parent's `task` tool call, its `TOOL_CALL_RESULT`, and the parent's own reply" visible, which is exactly what #3861 complains about — but it means the base class already has an opinion about delegation visibility, and it owns `get_state_and_messages_snapshots` with no filter hook. Whether this reconciliation belongs upstream is worth asking.

**Blocking: the filtering does not survive the request.**

The hidden ids live on `active_run`, which the base class sets to `INITIAL_ACTIVE_RUN` at the top of `run()` and clears in its `finally` block — and `add_langgraph_fastapi_endpoint` clones the agent per request. So the record exists only for the request that streamed the hidden events. The checkpoint still holds the delegation, so the next turn re-emits it.

I replayed your end-to-end scenario for turn 1 (same graph, same `TrackingAgent` harness, your exact assertion), then ran turn 2 on the same thread with a fresh `LangGraphAGUIAgent` and no new delegation:

```
TURN 1 snapshot ids: ['u1', 'outer-ai-2']          <- clean, matches your assertion
TURN 2 snapshot ids: ['u1', 'outer-ai-1', 'e0de4afb…', 'outer-ai-2', 'u2']
    outer-ai-1  assistant ''    tool_calls=['tc-research-1']
    e0de4afb…   tool  'CONFIDENTIAL internal reasoning: 42 widgets in stock.'
```

The issue's first suggested option — marking the message internal in the checkpoint itself (e.g. `additional_kwargs`) — would fix this by construction, since the marker travels with the message instead of with the request.

One note on the test while you are in there: the end-to-end case captures an inner chat-model event from a *separate* earlier `ainvoke` into a module global, then replays it inside a run that performs no delegation. That manufactured condition is what makes turn 1 filter, so the test cannot see the gap above. A second `agent.run()` on the same thread would.

**Also blocking: `emit_messages=False` alone now erases the enclosing tool call and its result.**

The `emit-messages` branch records the enclosing `tool_call_id` into `copilotkit_hidden_tool_call_ids`. So a tool that sets only `emit_messages=False`, deliberately leaving its tool call visible, loses the tool call and its result from the snapshot:

```python
agent.messages_in_process["run-1"] = {"id": "outer-ai-1", "tool_call_id": "tc-visible-1"}
agent._dispatch_event(TextMessageContentEvent(
    messageId="inner-ai-1", delta="chatter",
    rawEvent={"metadata": {"copilotkit:emit-messages": False}}))
# copilotkit_hidden_tool_call_ids == {'tc-visible-1'}
```

Same root cause, sharper: `retained_tool_call_id` is only cleared on `TOOL_CALL_RESULT`, so a fully visible tool call sitting between its `END` and its `RESULT` gets adopted by any later hidden message. I ran a visible `get_weather` call through `_dispatch_event` with empty metadata, then one hidden text event, then filtered a snapshot holding that turn and its result — surviving ids were `[]`. Visible content, dropped.

**Smaller items**

- `_retain_tool_call_association` sets `active_run["retained_tool_call_id"] = tool_call_id` and then pops it in the same call when the event is `TOOL_CALL_RESULT` — the assignment is dead on that path.
- `_retain_tool_call_association` and `_enclosing_tool_call_id` have accumulated a fair amount of inferred state across review rounds. If the marker moves onto the message in the checkpoint, most of that goes away.

Verification on my side: worktree at `fdf364692`, level with main. `python -m pytest tests/ -q` → 251 passed, 11 skipped. No CI has run on this PR yet (fork), so that is local only.

AyushCipher:
Thanks for the thorough review and insightful framing, @BenTaylorDev!

I've rebased on the latest `main` and reworked the implementation to address the request-boundary persistence and tool-call visibility issues you identified.

### 1. Checkpoint Persistence Across Request Boundaries

Hidden visibility metadata is now persisted with the checkpoint messages via `additional_kwargs["copilotkit_visibility"]`, rather than relying only on request-scoped `active_run` state.

On subsequent requests, including a fresh `LangGraphAGUIAgent` instance on the same thread, the visibility markers are reloaded before snapshot generation.

I also added multi-turn regression coverage using the real `LangGraphAGUIAgent.run()` flow to verify that hidden delegated content does not reappear on subsequent requests.

### 2. `emit_messages=False` vs `emit_tool_calls=False` Semantics

I removed the previous tool-call association state and updated snapshot filtering to handle:

- **`emit_messages=False` only**: hidden message results are removed while the visible outer tool call is preserved.
- **`emit_messages=False` + `emit_tool_calls=False`**: the hidden tool call and its result are removed.
- **`emit_tool_calls=False` only**: tool calls are withheld while visible textual responses remain.
- **Mixed visible/hidden tool calls**: visible sibling tool calls remain untouched.

### 3. Test Suite & Validation

Expanded `sdk-python/tests/test_snapshot_emit_filtering.py` to cover:

- Explicit Cases A, B, and C.
- Multi-turn checkpoint persistence across fresh agent instances.
- Normal conversation passthrough.
- Mixed visibility across multiple tool calls.
- Hidden nested delegated output through the real `run()` flow.

**Test Results:**

- Full SDK test suite: **264 passed, 12 skipped**
- Snapshot filtering tests: **25 passed**
- Ruff formatting: **passed**
- Ruff lint: **passed**

I'd appreciate another review when you have a chance. Thanks!

BenTaylorDev (commented):
Thanks for the rework. The persistence half now holds: I re-ran my turn-2 replay and the hidden delegation stays hidden on a fresh `LangGraphAGUIAgent` for the same thread. That was the round-1 blocker and it is resolved.

The reworked mechanism has one blocking problem, and it is structural rather than a narrowing fix. Reproduced against `0c4532a` with `ag-ui-langgraph` 0.0.44 and `langgraph` 1.2.11.

## Blocking: namespace inheritance hides visible messages, and it is the only thing making the fix work

`_record_hidden_output_ids` marks a whole `langgraph_checkpoint_ns` as hidden once it sees one hidden event in it (`langgraph_agui_agent.py:459`), then applies that verdict to every later event in the namespace (`:466`).

First, the false positive. The hide flag is recorded per namespace, so any message the node produces after the private call inherits the verdict, whatever its own metadata says. Graph: one node makes a private LLM call with `emit_messages=False`, then returns its own `AIMessage(id="visible-reply", content="Your answer is 42.")`.

```
main   snapshot ids: ['u1', 'visible-reply']     <- correct
PR     snapshot ids: ['u1']                      <- the answer is gone
```

I traced the recording event. It is the node's `on_chain_end`, and its own metadata does **not** carry `emit-messages: False`. The id is recorded purely by inheritance. `_persist_hidden_visibility` then writes that verdict into the customer's checkpoint, and markers are only ever set to `True` (`:310`), never cleared:

```
marker on visible-reply -> {'hidden_message': True}
TURN2 snapshot ids: ['u1']
```

The reply is hidden on every later turn of that thread, permanently.

That graph is synthetic. The docs example returns `state` rather than a visible message, and no example in `examples/` has this exact shape. But the inversion is real, and it is written to disk. The docs also treat this scoping as per call and not per namespace: *"If we modified `config` itself, it would change the default configuration for all subsequent LLM calls in that namespace"* is written as the thing to avoid (`disabling-state-streaming.mdx:55`).

Second, and this is why I am not asking you to just narrow it. I simulated removing the inheritance branch and re-ran the #3861 repro:

```
with inheritance      ['u1', 'ai-2']                              leak: no
inheritance removed   ['u1','ai-1','inner-tool-result','ai-2']    leak: yes
```

The single event that catches the delegation is that same `on_chain_end`, whose own metadata hides nothing. So the fix works only through the inference that also deletes visible replies. The two cannot be separated at this layer, because the hide flag rides the streamed event while the thing that needs marking is the produced message. The issue's own first suggestion, marking the message at production time in the checkpoint, does not have this problem, because the marker travels with the message instead of being inferred from where it came from.

## Secondary: the marker write is a mid-run write to the live thread

The base class calls `get_state_and_messages_snapshots` per node completion and per subgraph transition, not only at run end, so `aupdate_state(config, {"messages": updates}, as_node=None)` (`:342`) lands while the run is still streaming. Measured on the repro graph:

| | main | PR |
|---|---|---|
| checkpoints written per run | 5 | 6 |
| `aget_state` reads per run | 4 | 7 |
| mid-run `aupdate_state` calls | 0 | 1 |

Both counts grow with node count, and the reads are real latency against a Postgres checkpointer. The write also depends on `add_messages` upsert semantics. A `messages` channel with a different reducer appends the marked copies instead of replacing them. That kind of graph already double-counts its history today, so I am not calling it a regression, but it is a reason to avoid writing to the channel at all.

## Still open from round 1: the premise

Whether `copilotkit:emit-messages` filters history, and whether this reconciliation belongs in `ag-ui-langgraph` rather than here, is unsettled. The mechanism is now 275 source lines and it writes to the customer's checkpoint, so getting the contract wrong costs more than it did in round 1. Settling that first will probably also settle the blocker above, because a production-time marker is both a cleaner contract and a narrower mechanism.

## Blast radius question

`examples/showcases/research-canvas/agent/graph.py:166` reassigns `config` with `emit_messages=False`, so its whole namespace is hidden and the intermediate `ToolMessage`s that node returns would be stripped from the snapshot and marked `hidden_message: True` in the checkpoint. That may be the intent, but it is a behavior change to a shipped example and nothing covers it. I did not run that example, so treat this as a question rather than a measured finding. Worth confirming before this lands.

## Scope

The diff also carries a typing modernization (`Optional`/`Dict`/`List`/`Union` to PEP 604 and PEP 585) and `logger.error(exc_info=True)` to `logger.exception`. CI installs ruff but runs `ruff format` only, never `ruff check` (`static_quality.yml:85` passes `args: "--version"`), so those edits are not required and they spread the diff across the file. Please split them out.

Verdict: needs revision. Happy to paste the repro scripts if they help.

AyushCipher:
Thanks for the detailed review, @BenTaylorDev. I reworked the visibility tracking based on your feedback.

The main change is that hidden visibility is no longer inherited or persisted at the `langgraph_checkpoint_ns` level. The namespace is now used only as temporary correlation between the hidden `on_chat_model_end` event and its corresponding `on_chain_end`. At `on_chain_end`, the actual returned `ToolMessage` is inspected and its message ID is recorded as hidden. The temporary namespace state is then immediately consumed.

Direct hidden `AIMessage` outputs are tracked by their message ID as well.

I added/updated regression coverage for:

- the visible-reply false-positive case (`['u1', 'visible-reply']`)
- the original nested delegated-agent repro from #3861
- mixed hidden/visible sibling messages
- multi-turn checkpoint persistence using a fresh `LangGraphAGUIAgent`
- hidden `ToolMessage` and direct hidden `AIMessage` paths

The current verification results are:
- 28/28 snapshot filtering tests passed
- 268/268 SDK tests passed
- `ruff check` passed
- normal runs produce no additional `aupdate_state` calls
- hidden-message runs produce only one persistence update when new visibility markers need to be stored
- working tree is clean and the PR branch is up to date

I also reverted the unrelated typing/logging changes so the diff remains focused on the visibility filtering fix and its regression coverage.

The revised commit is `9ae86159b90875ac0c1aaf42af49aed20eb170f1`.

Thanks again for the concrete reproductions, especially the visible-reply case. They helped clarify the distinction between temporary event correlation and namespace-wide visibility inheritance.

BenTaylorDev:
I ran this iteration against the issue's own reproduction rather than the test fixture, and I need to stop you before a fourth round, because the result changes what we should be doing.

**The fix is a no-op on #3861.** I built the shape from the issue body — a `@tool` that `ainvoke`s an inner compiled graph and returns its content, with `copilotkit_customize_config(config, emit_messages=False)` applied inside the tool — and ran it against `origin/main` and this head on the locked deps (`ag-ui-langgraph 0.0.42`, `langgraph 1.1.10`):

```
### ORIGIN/MAIN ###                                ### PR HEAD ###
'u1'          role=user                            'u1'          role=user
'outer-ai-1'  role=assistant                       'outer-ai-1'  role=assistant
'bd9ee74d…'   role=tool  'CONFIDENTIAL internal…'  '16121be7…'   role=tool  'CONFIDENTIAL internal…'
'outer-ai-2'  role=assistant                       'outer-ai-2'  role=assistant
CONFIDENTIAL LEAKED: True                          CONFIDENTIAL LEAKED: True
checkpoint markers: all None                       checkpoint markers: all None
```

Byte-identical. Two independent reasons, either fatal on its own. The confidential `ToolMessage` is minted by the outer `ToolNode` in namespace `tools:92dd…`, one level above the namespace you marked. And `del hidden_ns[ns]` fires on the first `on_chain_end` in the namespace whose output is a dict, even when it recorded nothing, so the inner node's own `on_chain_end` burns the marker before the `ToolMessage` ever appears.

Your 2026-09-08 description holds only when the node that made the hidden call itself returns the `ToolMessage` in the same namespace. That is the fixture's shape at `test_snapshot_emit_filtering.py:594-622`. It does not occur in the issue, and I could not find it in `examples/`.

**The round-2 blocker is not fixed, only narrowed by one type.** Restricting to `isinstance(msg, LCToolMessage)` at `:319` fixes my `AIMessage` case and leaves the identical inversion live for `ToolMessage`. A node that makes a private LLM call for internal routing and then returns the real, user-visible tool result:

```
### ORIGIN/MAIN ###                          ### PR HEAD ###
ids = ['u1','ai-1','weather-result','ai-2']  ids = ['u1','ai-1','ai-2']        <- answer gone
markers: all None                            ToolMessage weather-result
                                               vis={'hidden_message': True}   <- written to the checkpoint
TURN 2 (fresh agent): present = True         TURN 2 (fresh agent): present = False
```

The docstring at `:230-236` asserts there is no legitimate case for a node to run with `emit_messages=False` and return a visible `ToolMessage`. That premise is what makes `:319` safe, and it is false: it is the documented pattern in `disabling-state-streaming.mdx:33-46`. The marker is never cleared (`:425`), so the loss is permanent for that thread.

**A second defect falls out of the same filter.** Dropping the `ToolMessage` while keeping its parent `tool_calls` slot manufactures an orphaned tool call:

```
'ai-1'  role=assistant  tool_calls=[('tc-weather-1','get_weather')]
ORPHANED tool_call ids (no matching tool result): ['tc-weather-1']
```

I measured the orphan itself and stopped there. I did not test how a provider or the UI reacts to it, so treat the downstream effect as unquantified rather than as a claim. What is certain is that the snapshot now advertises a tool call with no corresponding result. The base class's `_filter_orphan_tool_messages` strips fake orphan `ToolMessage`s by content regex and does not repair this direction. Your own primary test at `:745` asserts the post-filter ids and so currently enshrines the shape.

Two smaller things. Persistence into the customer's checkpoint is fully intact at `:403-457`, which contradicts your 2026-09-08 note that visibility is "no longer inherited or persisted" — the code sides with your 2026-09-05 message. And the typing and logging modernization is still in the diff, about 17 lines, despite the note that it was reverted.

**Where this leaves us.** My round-2 point — that the hide flag rides the streamed event while the thing needing marking is the produced message, and the two cannot be separated at this layer — is now measured rather than predicted. Three iterations have each traded one instance of that for another. Another round at this layer is not the answer.

Before any more code, the question I raised in round one still needs an answer, and it is the one thing I cannot review my way past: does `emit-messages` filter *history*, or only streaming? Every signal I can find says streaming only. The sibling TypeScript runtime filters events and has no history path (`v1-deprecated/.../langgraph/agent.ts:164-178`), and our own docs tell users to control history themselves (`disabling-state-streaming.mdx:48-49`). If that is the contract, #3861 is a documentation bug and this reconciliation does not belong in the Python SDK at all.

So please make the case for the premise before revising the implementation. If you think `emit-messages` should filter history, say why, and address the TS runtime and the docs as counter-evidence rather than around them. If you agree it is streaming-only, the useful version of this PR is much smaller. And if you conclude the marking has to happen where the message is produced, that is an `ag-ui-langgraph` change and worth raising upstream before more work lands here.

One concrete thing worth salvaging now: the `langgraph.py` one-liner that copies the dict so `copilotkit_customize_config` stops mutating the caller's metadata is a real, separable bug fix with two tests, and it is #6941. Split that into its own PR and I will merge it today.

AyushCipher:
Thanks @BenTaylorDev. I stepped back and investigated the contract across the docs, TypeScript runtime, Python SDK, and ag-ui-langgraph as you suggested.

The evidence points to `copilotkit:emit-messages` being a streaming-layer control rather than a history/checkpoint visibility control:

- The LangGraph docs describe `emit_messages=False` as disabling message streaming and explicitly say that messages should not be returned by nodes if they should not become part of message history.
- The TypeScript runtime uses `emit_messages=False` to suppress live message events and does not filter `MESSAGES_SNAPSHOT`.
- The Python SDK follows the same streaming-oriented behavior.
- `ag-ui-langgraph` builds `MESSAGES_SNAPSHOT` from the LangGraph checkpoint's `state["messages"]` and has no awareness of `copilotkit:emit-messages`.

This also explains the #3861 tension: the delegated-agent result needs to exist in LangGraph's `messages` state for the outer agent, but the user may not want that intermediate result exposed in the frontend snapshot.

So I agree that another SDK-side inference/namespace implementation isn't the right next step.

My understanding is that we should first decide whether #3861 is intended to change the existing `emit_messages` contract, and if so, whether that behavior belongs in `ag-ui-langgraph` rather than the CopilotKit Python SDK.

Separately, I'll split the independent #6941 metadata-mutation fix into its own PR as you suggested.

AyushCipher:
I've split the independent #6941 metadata-mutation fix into a separate PR as suggested.

PR: #7076

The new PR contains only the `langgraph.py` metadata-copy fix and its regression tests.

I've kept the #3861 / `emit_messages` investigation separate and have not made another SDK-side implementation change.

BenTaylorDev:
Thanks for stepping back rather than trying a third mechanism — that was the
right instinct, and your investigation is correct on every point I checked.

## Answering the premise

`copilotkit:emit-messages` is a streaming control, not a history filter. I
re-verified all four layers rather than take it on trust:

- **Docs.** The LangGraph advanced page presents `emit_messages=False` purely as
  streaming suppression, and its sample node ends with `# don't return the new
  response to hide it from the user`. History exclusion is the node's job.
- **TypeScript runtime.** `agent-integrations/langgraph/agent.ts:141` and
  `event-source.ts:161` both gate live events. Nothing in the TS tree filters
  `MESSAGES_SNAPSHOT`.
- **Python SDK.** `langgraph_agui_agent.py:245` drops message events at dispatch.
  Streaming only.
- **Upstream.** One correction to your note: `ag-ui-langgraph` *is* aware of the
  flag — `agent.py` reads an unprefixed `emit-messages` from event metadata to
  gate model-stream events. It just never applies it to the snapshot. So
  upstream owns the flag semantics *and* the snapshot builder, under a different
  key than the one we set.

That last point is why both of your implementations ran aground, and it was not
your doing. The flag rides a streamed event; the thing that needs marking is a
message already sitting in the checkpoint. Any fix at our layer has to reach
around a base class that owns both halves.

## What I found when I checked the upstream subagent work

I went looking for whether the newer subagent support had solved this for us. It
has not, and the reason is worth knowing before anyone spends more time here.

`ag-ui-langgraph` 0.0.45 (latest, 9 Sep) has three visibility modes: `inline`
(default, the old behavior), `attributed`, and `hidden`. `hidden` sounds like
this issue's fix and explicitly is not. Upstream's own comment:

> "hidden" (invisible delegation: the subagent's internal stream is suppressed
> entirely — the client sees only the parent's `task` call, its result, and the
> parent's own reply)

"Its result" is the `ToolMessage` whose content is the subagent's answer — the
exact thing #3861 reports leaking. The same wording is in 0.0.44, so this is a
settled design choice, not a gap.

The snapshot path confirms it. `get_state_and_messages_snapshots` still reads
root checkpoint messages and applies only `_filter_orphan_tool_messages`. There
is no message-level hide marker in the package, and no filter hook. The only two
functions added between 0.0.44 and 0.0.45 are usage accounting.

So the issue's suggested fixes — a configurable message filter, or honoring the
flag on messages — have now effectively been passed over upstream twice. I do
not think sending this there as a feature request would fare differently.

## Where the real fix lives

Upstream's answer to this class of problem is **attribution, not suppression**.
In `attributed` mode, subagent messages carry a `subagent_run_id` through the
snapshot precisely so a client can group or collapse them. History stays intact
on the wire; the frontend decides how to present delegated work.

The missing piece is ours. Nothing in `packages/react-core` or `packages/react-ui`
consumes that attribution today — outside generated showcase data, the only
`subagent` references in the repo are two A2UI files in `sdk-js`. So
`attributed` currently hands you tagged messages that no CopilotKit renderer
treats specially.

That reframes #3861 from "filter history on the server" to "render delegated
messages as delegated", which is a direction we have already started elsewhere.

## What I'm doing with this PR

Closing it, with apologies for the length of the road. Two review rounds of mine
pushed on the mechanism when the premise was the problem, and you should not have
had to find that yourself on the third pass.

Concretely:

- #3861 stays open. I will re-scope it to the frontend rendering work and record
  this finding on it.
- Your metadata-mutation fix landed separately in #6943 (621e17c585) — thank you
  for splitting that out, it was a real bug and it is fixed.
- The reporter's subclass override remains a legitimate workaround in the
  meantime, and I will say so on the issue.

PR #6811, opened 2026-08-31
chore(agentcore): repin @copilotkit packages to 1.70.0 (refs OSS-1029)

BenTaylorDev:
**Step 3 of #6188's landing order.** That PR states: *"1. Merge #6098. 2. Publish `@copilotkit/runtime` and `@copilotkit/react-core` with managed entitlement support. 3. Update the two AgentCore pins from `1.68.1` to that release. 4. Merge this PR. Do not merge this PR before step 3."*

Steps 1 and 2 completed today:

- #6098 merged at 18:55Z (`b07b1320f2`)
- `@copilotkit/runtime@1.70.0` published 20:24:51Z, `@copilotkit/react-core@1.70.0` at 20:26:32Z; `dist-tags.latest` is `1.70.0` for both
- `git merge-base --is-ancestor b07b1320f2 v1.70.0` confirms the managed Intelligence authority is actually *in* the release — a 1.70.0 without #6098 would have satisfied the version bump but not the step

This is step 3, so #6188 is no longer blocked on the release train.

## The locks are bumped minimally, not regenerated

Worth a look, because the obvious approach is wrong in two different directions.

**A plain `npm install --package-lock-only` leaves a duplicate.** The lambda ended up with two copies of `@copilotkit/shared` — `1.70.0` nested under `runtime`, and `1.68.1` still hoisted at the top level for the `@copilotkit/channels-*@0.9.0` packages, which ask for `^1.68.0` and were already satisfied. `npm dedupe` does not collapse it. `npm update @copilotkit/shared @copilotkit/core --package-lock-only` does.

**A full regeneration is far too broad.** Deleting the lock and rebuilding it produces a single clean copy, but sweeps 89 unrelated packages along with it — including `@graphql-tools/executor` 1.5.1 → **2.0.0**, `@graphql-tools/utils` 10 → 11, `debug` 2 → 4, and a `@types/express` 5 → 4 *downgrade*. That is a dependency sweep on a Lambda that ships to AWS, not a repin, so it was rejected.

What landed changes exactly three resolved versions in the lambda lock (`@copilotkit/core`, `runtime`, `shared`, all 1.68.1 → 1.70.0) plus the type-package hoisting that follows from runtime 1.70.0's own tree, and seven in the frontend lock.

## One thing left deliberately alone

The lambda's `overrides` block pins `@ag-ui/{client,core,encoder,proto}` to `0.0.58`, while `runtime@1.70.0` declares `0.0.59`. I left it as-is — it was already overriding runtime's declared version before this change, so the override goes on doing what it did. Flagging it in case that pin has a shelf life.

## Validation

- lambda `npm install` + `tsc -p tsconfig.json` → exit 0, `dist/` emitted
- frontend `npm install` + `tsc && vite build` → exit 0, built in 11.14s
- exactly one copy of every `@copilotkit/*` and `@ag-ui/*` package in both locks, none left on a 1.6x line
- no remaining `1.68.1` under `examples/integrations/agentcore`

Note `1.68.1` is pinned across ~20 other `examples/integrations/*/package.json`. Those are deliberately untouched — the landing order names only the two AgentCore pins.

Refs OSS-1029.

## Summary by CodeRabbit

* **Chores**
  * Updated the AgentCore integration to the latest CopilotKit release.
  * Keeps the frontend and runtime components aligned for improved compatibility and access to the latest fixes and enhancements.

### DIFF
--- a/examples/integrations/agentcore/frontend/package-lock.json
+++ b/examples/integrations/agentcore/frontend/package-lock.json
@@ -8,7 +8,7 @@
       "name": "fullstack-agentcore-solution-template-frontend",
       "version": "0.4.0",
       "dependencies": {
-        "@copilotkit/react-core": "1.68.1",
+        "@copilotkit/react-core": "1.70.0",
         "@radix-ui/react-alert-dialog": "^1.1.15",
         "@radix-ui/react-dialog": "^1.1.15",
         "@radix-ui/react-popover": "^1.1.17",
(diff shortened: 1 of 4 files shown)


BenTaylorDev:
Superseded by #6188 itself. When I opened this, #6188 was still pre-rebase and the AgentCore pins were the outstanding step 3. The force-push at 22:04Z brought `064c5ed910 chore(examples): update CLI starters to CopilotKit 1.70.0`, which repins the same four files — so this would now conflict with #6188 rather than unblock it. Closing.

One finding from here is still live against #6188, and I've reported it there: `064c5ed910`'s lambda lock carries two copies of `@copilotkit/shared`.

PR #6840, opened 2026-09-02
fix(runtime): handle pre-parsed Node request bodies

### LINKED ISSUE #6838: 🐛 Bug: NodeFetchHandler hung
### ♻️ Reproduction Steps

```
import { createControllerModule } from "@ai/apis/common/controller-module.factory";
import { V1Controller } from "@ai/apis/common/decorators";
import { jsonStringify } from "@ai/common/utils/json";
import { CopilotKitService } from "@ai/domains/copilot-kit/copilot-kit.interface";
import { CopilotKitModule } from "@ai/domains/copilot-kit/copilot-kit.module";
import {
  BuiltInAgent,
  CopilotKitIntelligence,
  CopilotRuntime,
  createCopilotRuntimeHandler,
} from "@copilotkit/runtime/v2";
import { createCopilotNodeHandler } from "@copilotkit/runtime/v2/node";
import { All, Inject, Logger, Req, Res } from "@nestjs/common";
import { ApiExcludeController } from "@nestjs/swagger";
import type { Request, Response } from "express";

@ApiExcludeController(true)
@V1Controller("/copilotkit")
export class CopilotKitController {
  private readonly logger = new Logger(CopilotKitController.name);


  @All()
  async copilotkit(

    @Req() req: Request,
    @Res() res: Response,
  ) {
    const builtInAgent = new BuiltInAgent({
      model: "openai:gpt-5.4-mini",
    });
      agents: { default: builtInAgent }, //,
      intelligence: new CopilotKitIntelligence({
        apiKey: "fake",
      }),
      // Threads are per-user. Without this, every visitor shares one history.
      identifyUser: (request) => ({
        id: request.headers.get("x-user-id") ?? "anonymous",
        name: request.headers.get("x-user-name") ?? "Anonymous",
      }),
    });
    const handler = createCopilotNodeHandler(
      createCopilotRuntimeHandler({
        runtime,
        basePath: "/api/v1/copilotkit",
      }),
    );

    return handler(req, res);
  }
}

export const CopilotKitApiControllerModule = createControllerModule(
  CopilotKitController,
  [CopilotKitModule],
);

```
```
curl --url 'http://127.0.0.1:3333/api/v1/copilotkit' \
  -H 'Accept: */*' \
  -H 'Accept-Language: en-US,en;q=0.5' \
  -H 'Cache-Control: no-cache' \
  -H 'Connection: keep-alive' \
  -H 'Content-Type: application/json' \
  -H 'DNT: 1' \
  -H 'Origin: http://localhost:3000' \
  -H 'Pragma: no-cache' \
  -H 'Referer: http://localhost:3000/' \
  -H 'Sec-Fetch-Dest: empty' \
  -H 'Sec-Fetch-Mode: cors' \
  -H 'Sec-Fetch-Site: cross-site' \
  -H 'Sec-GPC: 1' \
  -H 'User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/152.0.0.0 Safari/537.36' \
  -H 'sec-ch-ua: "Chromium";v="152", "Not?A_Brand";v="24", "Brave";v="152"' \
  -H 'sec-ch-ua-mobile: ?0' \
  -H 'sec-ch-ua-platform: "macOS"' \
  --data-raw '{"method":"info"}'
```

### ✅ Expected Behavior

return normally

### ❌ Actual Behavior

`/api/v1/copilotkit` hung

### Screenshots

_No response_

### 𝌚 CopilotKit Version

```shell
server side

npm ls | grep "@copilotkit"
├── @copilotkit/runtime@1.70.0 -> ./node_modules/.pnpm/@copilotkit+runtime@1.70.0_@cfworker+json-schema@4.1.1_@langchain+core@1.2.9_@opentelem_de5cf5235a9d6d68a70fb5c89ec0b6c1/node_modules/@copilotkit/runtime
├── @copilotkit/shared@1.70.0 -> ./node_modules/.pnpm/@copilotkit+shared@1.70.0_@ag-ui+core@0.0.59/node_modules/@copilotkit/shared


web site

 npm ls | grep "@copilotkit"
├── @copilotkit/react-core@1.56.4
├── @copilotkit/react-ui@1.56.4
├── @copilotkit/runtime@1.56.4
```

### 📄 Logs (Optional)

```shell
server side log

2026-09-02 11:21:20.964 - 01a06022-af40-7aa2-8fda-3859e0c1ff19 -  DEBUG [LoggingInterceptor] POST=>/api/v1/copilotkit | query: {} | params: {} | body: {"method":"info"}
```

Devamol10:
## Summary

Fixes #6838

`createCopilotNodeHandler` could hang when middleware (such as Express/NestJS body-parser) had already consumed the Node request body stream.

This change:
- Detects pre-parsed request bodies.
- Rebuilds the Fetch `Request` from the parsed body instead of reading the consumed stream.
- Reuses the handling logic for the Express bridge.
- Adds a regression test for the hanging POST request.

## Validation

- Targeted tests: 20/20 passed
- `git diff --check`: passed

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of requests whose bodies have already been consumed by middleware.
  * POST and other body-bearing requests now correctly preserve parsed string, JSON, and binary payloads.
  * Improved propagation of request cancellation and client disconnects during request processing.
  * Express integrations now use consistent request handling across supported Node environments.


Devamol10:
My approach:
Pre-parsed Node/NestJS body
        ↓
Detect consumed request stream
        ↓
Rebuild Fetch Request from req.body
        ↓
handler() → sendResponse()
        ↓
Prevent NodeFetchHandler hang

Also centralized the existing Express pre-parsed body handling and added a regression test.

20/20 targeted tests pass.

BenTaylorDev:
Thanks for this, and for the clear write-up. I reproduced the hang and confirmed your diagnosis holds in a real Express server, not only in the unit test.

Baseline on `main` against this branch, with `express.json()` in front of `createCopilotNodeHandler`:

```
                  main      this branch
POST json         HUNG      200
GET               200       200
POST empty body   HUNG      200
POST text/plain   200       200
```

I also mutation-checked your new test. Reverting `node-fetch-handler.ts` to `main` makes it time out at 8000 ms, which matches the reported symptom. The test earns its place.

Two asks before this lands, plus one optional extra.

**1. Please remove `express-fetch-bridge.ts` rather than leave an alias.**

`createExpressNodeHandler` is internal. Only `endpoints/express.ts` and the tests import it, and it is not part of the public API. Once the logic moves into `createCopilotNodeHandler`, that module becomes a one-line pass-through under a header that still says it "detects the pre-parsed case and re-serialises `req.body`". It no longer does that. Delete the file, import `createCopilotNodeHandler` directly in `express.ts`, and re-point `__tests__/express-fetch-bridge.test.ts`.

**2. Please derive the URL from the same source in both branches.**

`buildPreParsedRequest` uses `originalUrl`, and `createRequest` uses `req.url`. Inside one function, that makes the request path depend on the method. Under a mounted router:

```ts
const app = express();
app.use(express.json());
const router = express.Router();
router.all("/copilotkit", createCopilotNodeHandler(fetchHandler));
app.use("/api/v1", router);

// POST -> http://host/api/v1/copilotkit
// GET  -> http://host/copilotkit
```

With `basePath: "/api/v1/copilotkit"`, the POST matches and `GET /info` does not. This asymmetry already exists in the Express bridge on `main`, so you did not introduce it. Moving the code into the generic handler widens its reach, and this PR is the right moment to close it.

**3. Optional, and genuinely useful: a NestJS docs section.**

`docs/runtime-server-adapter.mdx` documents `createCopilotNodeHandler` for plain `node:http` and `createCopilotExpressHandler` for Express, and it carries no NestJS coverage. That gap is what sent the reporter of #6838 to the generic handler. A short section showing `app.use(createCopilotExpressHandler({ ... }))` in `main.ts` helps the next Nest user. Tell me if you prefer that we handle the docs separately.

One note on framing, so the value here is clear. #6838 has a working path today, because the Express adapter already handles a pre-parsed body and mounting it from Nest's `main.ts` works. So this change is not strictly required to close the issue. It is still worth taking. Picking the narrower of two Node factories behind any body parser currently hangs forever with no error and no log, and that is the worst failure mode we can hand a user.

sheiy:
@BenTaylorDev  I cant use `createCopilotExpressHandler`
```
pnpm why express
express@4.22.2
├─┬ @copilotkit/channels-teams@0.9.2
│ └─┬ @copilotkit/channels-intelligence@0.9.2
│   └─┬ @copilotkit/runtime@1.70.1
│     ├─┬ @ag-ui/mastra@1.1.2
│     │ └── ai@1.0.0 (dependencies)
│     └── ai@1.0.0 (dependencies)
└── @copilotkit/runtime@1.70.1 [deduped]

express@5.2.1
├─┬ @a2a-js/sdk@0.3.14
│ ├─┬ @mastra/core@1.35.0
│ │ └─┬ @mastra/client-js@1.19.1
│ │   └─┬ @ag-ui/mastra@1.1.2
│ │     └── ai@1.0.0 (dependencies)
│ └─┬ @mastra/core@1.63.2
│   ├── @ag-ui/mastra@1.1.2 [deduped]
│   ├─┬ @mastra/deployer@1.63.2
│   │ └─┬ mastra@1.27.2
│   │   └── ai@1.0.0 (devDependencies)
│   ├─┬ @mastra/loggers@1.3.0
│   │ └── mastra@1.27.2 [deduped]
│   ├─┬ @mastra/memory@1.28.1
│   │ └── ai@1.0.0 (dependencies)
│   ├─┬ @mastra/pg@1.22.2
│   │ └── ai@1.0.0 (dependencies)
│   ├─┬ @mastra/rag@2.6.0
│   │ └── ai@1.0.0 (dependencies)
│   ├─┬ @mastra/server@1.63.2
│   │ └── @mastra/deployer@1.63.2 [deduped]
│   ├── ai@1.0.0 (dependencies)
│   └── mastra@1.27.2 [deduped]
├─┬ @a2a-js/sdk@1.0.1
│ └── @mastra/core@1.63.2 [deduped]
├─┬ @modelcontextprotocol/sdk@1.30.0
│ ├─┬ @ag-ui/mcp-apps-middleware@0.0.3
│ │ └─┬ @copilotkit/runtime@1.70.1
│ │   ├── @ag-ui/mastra@1.1.2 [deduped]
│ │   └── ai@1.0.0 (dependencies)
│ ├─┬ @ag-ui/mcp-middleware@0.0.1
│ │ └── @copilotkit/runtime@1.70.1 [deduped]
│ ├── @copilotkit/runtime@1.70.1 [deduped]
│ ├── @mastra/core@1.35.0 [deduped]
│ └── ai@1.0.0 (dependencies)
├─┬ @nestjs/platform-express@12.0.1
│ ├─┬ @nestjs/core@12.0.1
│ │ ├── @nestjs/platform-express@12.0.1 [circular]
│ │ ├─┬ @nestjs/swagger@12.0.1
│ │ │ └── ai@1.0.0 (dependencies)
│ │ ├─┬ @nestjs/testing@12.0.1
│ │ │ └── ai@1.0.0 (devDependencies)
│ │ ├─┬ @nestjs/typeorm@12.0.1
│ │ │ └── ai@1.0.0 (dependencies)
│ │ └── ai@1.0.0 (dependencies)
│ ├── @nestjs/testing@12.0.1 [deduped]
│ └── ai@1.0.0 (dependencies)
├─┬ @slack/bolt@4.7.3
│ └─┬ @copilotkit/channels-slack@0.9.2
│   └─┬ @copilotkit/channels-intelligence@0.9.2
│     └── @copilotkit/runtime@1.70.1 [deduped]
└─┬ express-rate-limit@8.7.0
  └── @modelcontextprotocol/sdk@1.30.0 [deduped]
```
截屏 2026-09-07 at 10 15 39

BenTaylorDev:
Following up, because main moved underneath this branch in the one place that matters.

On 2026-09-08 I landed `1a935861a8`, which narrowed `hasPreParsedBody` in the express bridge to:

```ts
return Boolean(req.readableEnded);
```

The reason is in the comment there. `req.complete` and the private `_readableState.ended` are set by the HTTP parser once the socket has all the bytes, whether or not anything read them. body-parser 1.x assigns `req.body = req.body || {}` before its own content-type checks, so a request it declines to parse, such as a multipart upload or `text/plain`, arrives with `req.body === {}` and a full, unread stream.

This branch relocates the older four-way predicate into `node-fetch-handler.ts:104-112`:

```ts
return Boolean(req.readableEnded || req.complete || state?.ended || state?.endEmitted);
```

Merging that would revert the fix and widen it from the express bridge to every Node consumer, including `createCopilotNodeHandler` and `createCopilotNodeListener`. Two tests on main already assert the opposite, at `packages/runtime/src/v2/runtime/__tests__/express-fetch-bridge.test.ts:182` and `:194`.

I also owe you a correction on my own repro table. The `POST text/plain 200 / 200` row was a false green: it measured the status code, not the payload. Main streams `hello`; this branch would deliver `{}`. Both return 200.

So the ask list is now four items, and the branch also conflicts on `express-fetch-bridge.ts` because it deletes `hasPreParsedBody` while main rewrote it:

1. Rebase onto current main.
2. Adopt main's `hasPreParsedBody` body verbatim, `return Boolean(req.readableEnded)`, in whichever module ends up owning it.
3. Delete `express-fetch-bridge.ts` as before, importing `createCopilotNodeHandler` directly in `express.ts` and re-pointing the tests. Doing this makes the conflict a clean delete instead of a delete-versus-modify.
4. Derive the URL from one source in both branches.

The bug you reported is still live on main at `node-fetch-handler.ts:34`, and your diagnosis and the direction of the refactor are both right. This is one round of work, not a rewrite. No rush from my side, and tell me if you would rather I split the predicate change out so you are not carrying it.

BenTaylorDev:
@sheiy — your screenshot is more specific than the prose, and it points somewhere real. This is a compile error, not the hang.

`@copilotkit/runtime` declares express as a hard runtime dependency, not a peer: `packages/runtime/package.json` lists `"express": "^4.21.2"` under `dependencies`, plus `"@types/express": "^4.17.21"`. The express adapter imports express at value level and builds the router from runtime's own copy, and the exported type is express-4-typed. In your tree your app resolves `@types/express@5.0.6` while runtime resolves the v4 types, so `app.use(...)` is a cross-major mismatch. That is exactly the `Property 'param' is missing` your error names, since `.param()` was dropped from `Request` in the v5 types.

There is no version guard anywhere in the runtime source, so nothing detects or warns about a dual-express tree.

This is our defect, not your configuration. `express` and `hono` should be optional peers with a `^4 || ^5` range, and `@types/express` should not be a runtime `dependencies` entry at all. I could not find an existing issue for it, so I am filing one and will link it here.

As a stopgap today, `app.use(copilotRouter as unknown as express.RequestHandler)` will unblock the type error. Be aware that only settles the types. An express-4 router mounted in an express-5 app also means v4 layer matching against a v5 request, which I have not tested, so treat it as a workaround rather than a fix.

This is separate from the pre-parsed-body hang in #6838, which is still live on main and still tracked here.

nathan-xie-tj:
> @sheiy — your screenshot is more specific than the prose, and it points somewhere real. This is a compile error, not the hang.
> 
> `@copilotkit/runtime` declares express as a hard runtime dependency, not a peer: `packages/runtime/package.json` lists `"express": "^4.21.2"` under `dependencies`, plus `"@types/express": "^4.17.21"`. The express adapter imports express at value level and builds the router from runtime's own copy, and the exported type is express-4-typed. In your tree your app resolves `@types/express@5.0.6` while runtime resolves the v4 types, so `app.use(...)` is a cross-major mismatch. That is exactly the `Property 'param' is missing` your error names, since `.param()` was dropped from `Request` in the v5 types.
> 
> There is no version guard anywhere in the runtime source, so nothing detects or warns about a dual-express tree.
> 
> This is our defect, not your configuration. `express` and `hono` should be optional peers with a `^4 || ^5` range, and `@types/express` should not be a runtime `dependencies` entry at all. I could not find an existing issue for it, so I am filing one and will link it here.
> 
> As a stopgap today, `app.use(copilotRouter as unknown as express.RequestHandler)` will unblock the type error. Be aware that only settles the types. An express-4 router mounted in an express-5 app also means v4 layer matching against a v5 request, which I have not tested, so treat it as a workaround rather than a fix.
> 
> This is separate from the pre-parsed-body hang in #6838, which is still live on main and still tracked here.

@BenTaylorDev  how is the 
```
This is our defect, not your configuration. `express` and `hono` should be optional peers with a `^4 || ^5` range, and `@types/express` should not be a runtime `dependencies` entry at all. I could not find an existing issue for it, so I am filing one and will link it here.

```
going?

BenTaylorDev:
@nathan-xie-tj — filed as #7276, and sorry for the week's delay. You asked a
straight question and the honest answer is that I said I would file it and then
did not.

The issue carries the evidence: `express`, `@types/express`, `hono` and
`@hono/node-server` are all still hard `dependencies` on `main`, both frameworks
are imported at value level, and `CopilotExpressRouter` re-exports an
express-4-typed `Router`. One thing I had not appreciated when I first replied is
that this is inconsistent with our own convention — the runtime already declares
nine other integration dependencies as optional peers, including the LangChain
family, Anthropic, OpenAI and Groq. Express and Hono are the outliers.

I am starting on the fix now. There is one thing to settle first: whether `hono`
is reachable from the package root entry and not just from `./v2/hono`. If it is,
that edge has to be broken before the dependency can become optional, or the root
import breaks for everyone. That is written up in #7276.

The stopgap in the meantime is unchanged:

```ts
app.use(copilotRouter as unknown as express.RequestHandler);
```

It settles the types only, not express-4 layer matching against an express-5
request.

This PR stays about the pre-parsed body hang, which is still live on `main`.
@Devamol10 — the four asks from my 10 Sep comment still stand whenever you pick
it back up, and the rebase is the first of them.

BenTaylorDev:
Progress on the express side, since I said I would start on it.

**The fix is up as #7278**, tracked in #7276. `express` is now an optional peer at `^4.21.2 || ^5.0.0`, `express` and
`@types/express` move to dev dependencies, and `CopilotExpressRouter` is widened
from express-4's `Router` to the structural middleware shape both majors accept.
The runtime value is still a real `express.Router()`; only the declared type
changes.

I proved it rather than asserted it. I built the package and compiled a
`strict: true` consumer against real `@types/express@5`:

- with the old `Router`-pinned type, `app.use(copilotRouter)` **fails**
- with the new structural type, it **compiles**

`check-types`, `build` and the runtime suite (177 files, 2452 tests) all pass.
There is also a type-level guard in the package that fails if anyone re-pins the
type to an express-derived one.

**Two corrections to things I said earlier in this thread.**

First, I claimed a barrel chain pulls `express` into the package root and that it
had to be broken before express could become optional. That is true of the source
import graph and false of the published artifact — the bundler already
tree-shakes it. I measured the built root entry's transitive imports before and
after: identical, express absent in both. The change is in the branch anyway as
hardening, but it was not the blocker I said it was.

Second, and more interesting: the root `package.json` already pins
`pnpm.overrides.express: ">=4.20.0"`, so this workspace resolves **express 5.2.1
everywhere**, including for the runtime's own tests. So we have been building and
testing against express 5 while declaring `^4.21.2` as a dependency and shipping
express-4 types to everyone downstream. @sheiy's compile error was not an edge
case; it was us shipping types for a major we do not actually run against.

`hono` stays a hard dependency, and that is deliberate. Unlike express it is used
internally by the v1 `node:http` integration, so consumers never supply it.

**This PR is unaffected.** The pre-parsed body hang is a separate defect and is
still live on `main` — `createCopilotNodeHandler` in `node-fetch-handler.ts`
still builds the request unconditionally with no pre-parsed detection.
@Devamol10, the four asks from my 10 Sep comment stand whenever you pick it back
up, and the rebase is the first of them. Happy to split the `hasPreParsedBody`
predicate change out into its own PR if carrying it is the awkward part — just
say so.

PR #6849, opened 2026-09-02
fix(sdk-python): require ag-ui-langgraph history guard release

### LINKED ISSUE #2974: 🐛 Bug: ValueError: Message ID not found in history (v2 Design Components and 1.50.x) with ag-ui
### ♻️ Reproduction Steps

Steps to Reproduce
Start a conversation with an agent
Ask a question that triggers a tool call
Cancel/stop the stream before the tool call completes
Send a new message
Crash occurs
Error Traceback
File "ag_ui_langgraph/agent.py", line 297, in prepare_stream    return await self.prepare_regenerate_stream(...)File "ag_ui_langgraph/agent.py", line 369, in prepare_regenerate_stream    time_travel_checkpoint = await self.get_checkpoint_before_message(message_checkpoint.id, thread_id)File "ag_ui_langgraph/agent.py", line 843, in get_checkpoint_before_message    raise ValueError("Message ID not found in history")ValueError: Message ID not found in history

### ✅ Expected Behavior

when clicking on stop generation, streaming stops no exceptions

### ❌ Actual Behavior

Root Cause
In ag_ui_langgraph/agent.py:
get_checkpoint_before_message (line 843) raises ValueError when a message ID isn't found in history. This happens when stream cancellation leaves an incomplete/unsaved message.
prepare_regenerate_stream (line 369-371) calls get_checkpoint_before_message but doesn't catch this ValueError. It only handles None returns:
   time_travel_checkpoint = await self.get_checkpoint_before_message(message_checkpoint.id, thread_id)   if time_travel_checkpoint is None:       return None
prepare_stream (line 297) directly returns the result of prepare_regenerate_stream without handling None:
   return await self.prepare_regenerate_stream(...)
When prepare_stream returns None, the caller (_handle_stream_events) crashes because it expects a valid stream response.
Suggested Fix
In prepare_regenerate_stream, wrap the get_checkpoint_before_message call in try/except:
# In ag_ui_langgraph/agent.py, prepare_regenerate_stream methodtry:    time_travel_checkpoint = await self.get_checkpoint_before_message(message_checkpoint.id, thread_id)except ValueError:    time_travel_checkpoint = Noneif time_travel_checkpoint is None:    return None  # Or better: fall back to normal streaming instead of returning None
Additionally, prepare_stream should handle the None return from prepare_regenerate_stream by falling back to normal streaming instead of returning None to its caller:
# In prepare_stream, around line 297if last_user_message:    regenerate_result = await self.prepare_regenerate_stream(...)    if regenerate_result is not None:        return regenerate_result    # Fall through to normal streaming below instead of returning None
Related Issue
When stream is cancelled during a tool call, the checkpoint may also have corrupted message history (AIMessage with tool_calls but no corresponding ToolMessage), causing Azure OpenAI to reject with:
> "An assistant message with 'tool_calls' must be followed by tool messages responding to each 'tool_call_id'"

### 𝌚 CopilotKit Version

```shell
├── @copilotkit/react-core@1.50.1 -> ./node_modules/.pnpm/@copilotkit+react-core@1.50.1_@types+mdast@4.0.4_@types+react-dom@18.3.7_@types+react@1_213e2804f05d4f470ed8362d94112a47/node_modules/@copilotkit/react-core
├── @copilotkit/react-ui@1.50.1 -> ./node_modules/.pnpm/@copilotkit+react-ui@1.50.1_@types+mdast@4.0.4_@types+react-dom@18.3.7_@types+react@18._bfb4c4f4d5e668b8e72cd31c512152b3/node_modules/@copilotkit/react-ui
├── @copilotkit/runtime@1.50.1 -> ./node_modules/.pnpm/@copilotkit+runtime@1.50.1_7f93d1cee3bdee5ca8557718f4ead80b/node_modules/@copilotkit/runtime
```

### 📄 Logs (Optional)

```shell

```

lorenzozanee:
## What problem occurs?

Stopping a LangGraph tool call can leave the next request vulnerable to ag-ui-langgraph's missing-message-ID regeneration path. CopilotKit's lockfile resolves ag-ui-langgraph 0.0.42, before the released caller-side guard.

## What changed?

Raise the sdk-python ag-ui-langgraph minimum to 0.0.44 and refresh poetry.lock so installs receive the released guard and matching protocol dependencies.

Fixes #2974

## Summary by CodeRabbit

* **Updates**
  * Updated the Python SDK’s LangGraph integration requirement to version 0.0.44 or later.
  * Preserved the FastAPI integration option.

* **Tests**
  * Added validation to ensure the required and locked LangGraph integration versions remain consistent.

### DIFF
--- a/sdk-python/poetry.lock
+++ b/sdk-python/poetry.lock
@@ -14,23 +14,23 @@ files = [
 
 [[package]]
 name = "ag-ui-langgraph"
-version = "0.0.42"
+version = "0.0.44"
 description = "Implementation of the AG-UI protocol for LangGraph."
 optional = false
 python-versions = "<3.15,>=3.10"
 groups = ["main"]
 files = [
-    {file = "ag_ui_langgraph-0.0.42-py3-none-any.whl", hash = "sha256:4fd19f0da6d0e16d727ec89e99b692916cbba2b0302f96aba2497043cbfec5db"},
-    {file = "ag_ui_langgraph-0.0.42.tar.gz", hash = "sha256:5384647b9b7b098189530c59b26741581dd009c8dfda907fbfaa9bafa8d21249"},
+    {file = "ag_ui_langgraph-0.0.44-py3-none-any.whl", hash = "sha256:fef063c41d9086ec34e0fc0400660e832035872fd5747fa0d734cb011bca781a"},
+    {file = "ag_ui_langgraph-0.0.44.tar.gz", hash = "sha256:869b92ae1135ff911959eeefc3f2a82f138532c775d73fbd68c9936e5e96ee1c"},
 ]
 
 [package.dependencies]
 ag-ui-a2ui-toolkit = ">=0.0.4"
-ag-ui-protocol = ">=0.1.15"
+ag-ui-protocol = ">=0.1.21"
 fastapi = {version = ">=0.115.12", optional = true, markers = "extra == \"fastapi\""}
 langchain = ">=1.2.0"
 langchain-core = ">=0.3.0"
-langgraph = ">=0.3.25,<2"
+langgraph = ">=0.6.0,<2"
 pydantic = ">=2.0.0"
 
 [package.extras]
@@ -38,14 +38,14 @@ fastapi = ["fastapi (>=0.115.12)"]
 
 [[package]]
 name = "ag-ui-protocol"
-version = "0.1.19"
+version = "0.1.22"
 description = ""
 optional = false
 python-versions = ">=3.9"
 groups = ["main"]
 files = [
-    {file = "ag_ui_protocol-0.1.19-py3-none-any.whl", hash = "sha256:898843b1410d378824da0c6a776486288b9c5828689d0bf563118868e37f390f"},
-    {file = "ag_ui_protocol-0.1.19.tar.gz", hash = "sha256:43c27f60d41712dcad0e9e0a203cbdf1c8e248b22417374c5c68321c448af4ea"},
+    {file = "ag_ui_protocol-0.1.22-py3-none-any.whl", hash = "sha256:fca13ee7820f8f53e869c19e09ddd75826c1799b27c2adb6f2e567295433c704"},
+    {file = "ag_ui_protocol-0.1.22.tar.gz", hash = "sha256:d21f265284a50d9fc87ad7bcbd58f737b4b16eef7b5375f13a6e925117b52046"},
 ]
 
 [package.dependencies]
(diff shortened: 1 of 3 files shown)


BenTaylorDev:
I checked this against both wheels and the stated mechanism does not hold.

The caller-side guard for #2974 is the `last_user_id in checkpoint_ids` gate in `prepare_stream`. That gate is already present in ag-ui-langgraph **0.0.42** (`ag_ui_langgraph/agent.py:526`) and is identical to the one in 0.0.44 (`agent.py:2258`). Nothing on that path changed between the two releases.

`get_checkpoint_before_message` also still raises the same `ValueError(f"Message ID {message_id!r} not found in history ...")` in 0.0.44 (`agent.py:3838`), and `prepare_regenerate_stream` still does not catch it. 0.0.44 adds a comment saying so outright:

```
# ``get_checkpoint_before_message`` raises ``ValueError`` when the
# message id is missing from history; it never returns ``None``, so
# no None-guard is needed here.
```

So the bump does not fix the crash in #2974.

Two further points, independent of that claim:

1. `pyproject.toml` still declares `langgraph = ">=0.3.25,<2"` and `ag-ui-protocol = ">=0.1.15"`, but the refreshed lock requires `langgraph >=0.6.0` and `ag-ui-protocol >=0.1.21`. The declared floors no longer describe what installs.
2. `tests/test_ag_ui_langgraph_dependency.py` asserts the literal strings `">=0.0.44"` and `version = "0.0.44"`. It exercises no behavior, and it turns red on the next legitimate bump.

Could you drop the `Fixes #2974` line and either re-pitch this as a plain floor bump — with the pyproject floors updated to match the lock and the string-matching test removed — or close it? Happy to look again either way.

BenTaylorDev:
Closing this one. The `Fixes #2974` claim does not hold (detail in my comment above), and the residual — a bare floor bump — is something the repo has decided against on the record.

`.github/workflows/test_unit-python-sdk.yml:36-50` already runs the sdk-python suite against `0.0.42`, `0.0.43` and `0.0.44`, and states why the declared floor stays put:

> Covering the floor in CI is what lets the runtime floor stay at 0.0.42: raising it for test coverage alone would also raise the effective LangGraph floor from >=0.3.25 to >=0.6.0 for every consumer.

That is the cost this PR would impose: its refreshed lock pulls an `ag-ui-langgraph` that requires `langgraph >=0.6.0`. And the concrete breakage from newer releases was already fixed on the caller side by #6592 (`c2038cf52c`), which makes `LangGraphAGUIAgent` forward `**kwargs` upstream rather than pin away from the problem.

#2974 is still open and still a real bug. The crash sits in `get_checkpoint_before_message` / `prepare_regenerate_stream`, both of which live in `ag-ui-langgraph`, so the fix belongs upstream in that repo. Happy to look if you want to take it on there — thanks for digging into it.

PR #6850, opened 2026-09-02
test(react-core): cover partial parallel tool rendering

lorenzozanee:
## What does this PR do?

Add regression coverage for partial completion of parallel v1 tool calls, ensuring each call is evaluated and a completed result renders independently.

## Related PRs and Issues

- Related to #2051

## Checklist

## Reviewers

- Ben Taylor
- Rainer Hahnekamp

## Summary by CodeRabbit

- **Tests**
  - Expanded automated coverage for parallel tool-call rendering scenarios.
  - Verified that completed tool calls continue displaying their results when another call is still pending.
  - Confirmed pending calls remain empty until results are available, while all calls are processed correctly.

### DIFF
--- a/packages/react-core/src/v1-deprecated/hooks/__tests__/use-lazy-tool-renderer.test.tsx
+++ b/packages/react-core/src/v1-deprecated/hooks/__tests__/use-lazy-tool-renderer.test.tsx
@@ -69,6 +69,35 @@ describe("useLazyToolRenderer", () => {
     expect(renderToolCall).toHaveBeenCalledTimes(2);
   });
 
+  it("renders a completed parallel tool call while another is still pending", () => {
+    const renderToolCall = vi.fn(({ toolCall, toolMessage }) => {
+      if (!toolMessage) return null;
+      return (
+        <div data-testid="tool-call">
+          {toolCall.function.name}:{toolMessage.content}
+        </div>
+      );
+    });
+    vi.mocked(useRenderToolCall).mockReturnValue(renderToolCall);
+
+    const messages = [
+      message,
+      {
+        id: "time-result",
+        role: "tool",
+        toolCallId: "time-call",
+        content: "noon",
+      },
+    ] as Message[];
+    const { result } = renderHook(() => useLazyToolRenderer());
+
+    render(result.current(message, messages)!());
+
+    expect(screen.getByText("get_time:noon")).toBeTruthy();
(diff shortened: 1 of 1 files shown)


BenTaylorDev:
Verified locally: 4 passed. It also survives a mutation check — making the hook bail when any tool call lacks a result fails the new test, so it is not self-fulfilling.

One cleanup before merge: drop `expect(screen.queryByText("get_weather:")).toBeNull()`. The mock returns `null` when `toolMessage` is undefined, so that line asserts the mock's own early return rather than anything about `useLazyToolRenderer`. The two assertions carrying the test are `getByText("get_time:noon")` and `toHaveBeenCalledTimes(2)`.

Worth noting for later readers: `useLazyToolRenderer` on `main` already evaluates each tool call independently, so this locks in working behavior rather than covering #2051 — which matches the "Related to" framing in the description.

BenTaylorDev (changes requested):
Thanks for this. I ran mutation checks, and the new case does add coverage that the existing three tests do not have. If I change the result lookup to index alignment (`messages.filter((m) => m.role === "tool")[i]` instead of matching `toolCallId`), only your test fails. That is a real regression this file did not catch before.

One change before merge. This assertion cannot fail:

```ts
expect(screen.queryByText("get_weather:")).toBeNull();
```

The mock returns `null` when `toolMessage` is absent, so that text never appears on any code path. It also misses the failure the test is named for. If I drop the id check in the hook and match any tool message:

```ts
const toolMessage = messages?.find((m) => m.role === "tool") as ToolResult;
```

then the pending call renders `get_weather:noon`, which is another call's result. Your test still passes. Only the older "renders every tool call and matches each tool result" test catches it.

These two lines close the gap. I confirmed they fail under that mutation and pass against current `main`:

```ts
expect(screen.getAllByTestId("tool-call")).toHaveLength(1);
expect(screen.queryByText(/get_weather/)).toBeNull();
```

One note on scope, so the history stays clear. Issue #2051 is a delivery problem: the tool result does not reach the frontend until the LangGraph run ends. This test covers rendering, and the rendering fix landed in #6682, which also added the three existing cases in this file. Your body says "Related to #2051", which is accurate. The branch name `fix/2051-langgraph-tool-results` reads like a fix for the issue, so I want to say plainly that merging this does not close it.

On my side: the file passes with 4 tests, `oxfmt --check` and `oxlint` are clean, and the hook on your branch is identical to the hook on `main`. I will start full CI once the assertion change is in.

PR #6856, opened 2026-09-02
fix(release): stop one scope's version line freezing another's (refs OSS-1107)

BenTaylorDev:
## The defect

`@copilotkit/angular` is published from its own release scope, on a `0.x` line, while depending on `@copilotkit/core` from the `monorepo` scope. It declared that dependency as `workspace:*`, and `pnpm pack` rewrites `workspace:*` into the exact version in the tree.

| package | published | pinned `@copilotkit/core` |
| -- | -- | -- |
| `@copilotkit/angular@0.4.0` | 2026-08-28 | `1.69.3` |
| `@copilotkit/core@1.70.0` | 2026-08-31 | — |
| `@copilotkit/angular@0.5.0` | 2026-09-01 | `1.70.0` |

For the four days in between, an Angular application obeying the CLI's `1.70.0` dependency floor installed `@copilotkit/*@1.70.0` while its Angular package demanded an exact `1.69.3`. Two copies of core resolved, and the Angular package's calls landed on the copy that did not export what it needed — the threads drawer and the Inspector threads view threw at runtime, and the only repair available to a developer was an `overrides` block.

`0.5.0` publishing closed the instance. It did not close the window: nothing structural held the two version lines together, so the window reopened on every `1.x` release.

## What changed

**1. Angular declares ranges.** Its five `monorepo`-scope dependencies move from `workspace:*` to `workspace:^`. Every `channels` package already did this; Angular was the only scope that did not. Proof, from `pnpm pack` on this branch:

```
{'@copilotkit/a2ui-renderer': '^1.70.0', '@copilotkit/web-inspector': '^1.70.0',
 '@copilotkit/shared': '^1.70.0', '@copilotkit/core': '^1.70.0',
 '@copilotkit/web-components': '^1.70.0'}
```

An Angular application on `@copilotkit/core@1.71.0` now resolves exactly one copy.

**2. A static check, on every push.** `pnpm run verify:cross-scope-pins` refuses any dependency crossing a scope boundary that is declared as one exact version. Only `dependencies` and `peerDependencies` count — npm never installs a devDependency for a consumer, so an exact one cannot split a consumer's tree. Wired into `test_unit.yml`.

**3. A registry-backed check, before a stable publish.** `pnpm run verify:cross-scope-pins:registry -- --scope <scope>` reads what each *other* scope's **published** manifest declares and refuses to ship a version those ranges have stopped admitting. Wired into `publish-release.yml` ahead of the publish step, for stable releases only.

Run against the live registry with the monorepo version set to `1.71.0`, it reproduces the refusal that would have prevented this incident:

```
Cross-scope dependency violations:
  - Published @copilotkit/angular declares @copilotkit/core as "1.70.0", which 1.71.0
    does not satisfy. Releasing scope monorepo now would leave every consumer of
    @copilotkit/angular resolving two copies of @copilotkit/core. Release scope angular
    first, from this tree: it packs @copilotkit/core as "workspace:^", which resolves
    against the workspace version rather than the published one, so its new release
    admits 1.71.0 before this one ships.
```

The remedy is an order, not a bigger release, so it is achievable in the current single-scope stable lane.

## Also in scope

`@copilotkit/runtime` declared `@copilotkit/channels-intelligence` as `workspace:*` — the only other exact pin crossing a scope boundary, and the same defect. It moves to `workspace:^`. Without it the static check cannot be a hard gate.

**Operational consequence worth knowing before merge:** the first stable release after this lands will be refused if the published dependent still carries an exact pin from the last publish. Concretely, the next `monorepo` release needs an `angular` release first, and the next `channels` release needs a `monorepo` release first. Both are one extra release each, once. After that both dependents carry ranges and the gate is quiet for ordinary minors.

## Not fixed here

`@copilotkit/runtime` publishes `@copilotkit/channels-core` as `^0.9.0`. Under semver a `0.x` minor is breaking, so `^0.9.0` will not admit `channels@0.10.0`, and the gate will ask for a `monorepo` release first each time the channels minor moves. That is correct but noisier than the `1.x` case, and worth its own decision about whether the channels line should reach `1.0.0`.

## Validation

```
pnpm run verify:cross-scope-pins                                  # OK
pnpm run verify:cross-scope-pins:registry -- --scope all          # OK (live registry)
npx vitest run --config scripts/release/vitest.config.mts         # 176 passed, 15 files
npx nx run-many -t check-types,test --projects=@copilotkit/angular,@copilotkit/runtime
npx oxlint scripts/release/                                       # 0 warnings, 0 errors
npx oxfmt --check scripts/release                                 # clean
pnpm install                                                      # lockfile: 6 specifier lines
```

Counterfactual proved by hand: with `@copilotkit/react-core` set to `1.71.0`, the registry check fails against the live registry with the five messages above, then passes again once restored.

## Summary by CodeRabbit

* **Release Reliability**
  * Added automated validation for cross-scope dependency pins during testing and publishing.
  * Expanded checks to cover runtime, peer, optional, workspace, and registry-published dependency ranges.
  * Release verification now detects incompatible pins before publication, helping prevent partial releases.

* **Documentation**
  * Updated the Vue parity matrix to reflect additional mapped hook and helper test coverage.
  * Reformatted showcase documentation for improved readability.

* **Tests**
  * Added comprehensive coverage for cross-scope dependency validation scenarios.


.github/workflows/publish-release.yml
```diff
@@ -420,6 +420,17 @@ jobs:
         if: ${{ inputs.dry-run != true && steps.meta.outputs.scope == 'channels' && steps.meta.outputs.mode == 'stable' }}
         run: pnpm run verify:channels-umbrella:registry
 
+      # The last line of defence for OSS-1107: a stable release of one scope
+      # must not ship a version another scope's PUBLISHED manifest has stopped
+      # admitting. `@copilotkit/angular@0.4.0` pinned `@copilotkit/core@1.69.3`
+      # and the monorepo scope released `1.70.0` four days later, so every
+      # Angular application on the new floor resolved two copies of core.
+      - name: Verify no published scope is left behind
```
dustywusty: **Run this gate before any Channels package is published.**

For the `channels` scope, the dependency phase at lines 412–417 has already published the foundation and adapter packages under `latest` before this check runs. A rejection—especially the documented `0.x` minor-range case—therefore exposes a partial, incompatible release until the prerequisite scope is published and this workflow is retried. Move this step above `Publish Channels dependencies to npm` so it is a true pre-publish gate for every scope.

scripts/release/verify-cross-scope-pins.ts
```diff
@@ -0,0 +1,167 @@
...
+ * whose package has never shipped cannot be left behind by this release. Any
+ * other registry failure throws, because a release must not be waved through by
+ * a check that could not run.
+ */
+function readPublishedDependencies(
+  packageName: string,
+): Record<string, string> | undefined {
+  const result = spawnSync(
+    "npm",
+    ["view", `${packageName}@latest`, "dependencies", "--json"],
```
dustywusty: **Fetch the published dependency field represented by each edge.**

`crossScopeEdges` emits both `dependencies` and `peerDependencies`, but this registry query loads only `dependencies`. A published cross-scope peer range is consequently absent from `publishedDependencies`, and `findSupersededPublishedPins` silently skips it. Preserve the field in the registry data and fetch/compare both dependency maps.

scripts/release/lib/cross-scope-pins.ts
```diff
@@ -0,0 +1,211 @@
...
+ * Whether a declared range admits anything other than one exact version.
+ *
+ * `workspace:*` and `workspace:~` are read through to what `pnpm pack` writes:
+ * `*` becomes the exact version in the tree, `~` and `^` become ranges. A
+ * literal `1.69.3` is the same defect already resolved.
+ *
+ * @param range - The declared range.
+ * @returns True when a later release of the dependency can satisfy it.
+ */
+function admitsALaterRelease(range: string): boolean {
```
dustywusty, line 133: **Classify exact pins using semver semantics.**

This special-cases only the shorthand `workspace:^` and `workspace:~`. Valid workspace ranges such as `workspace:^1.70.0` are falsely rejected, while exact semver forms such as `=1.69.3` or versions with build metadata are accepted. Handle `workspace:*` explicitly, strip the workspace protocol when present, and use semver range semantics to determine whether the range admits more than one version.

scripts/release/lib/cross-scope-pins.ts
```diff
@@ -0,0 +1,211 @@
...
+  readonly devDependencies?: Readonly<Record<string, string>>;
+}
+
+/**
+ * Dependency fields a consumer of the published package has to resolve.
+ *
+ * `devDependencies` are deliberately absent: npm never installs them for a
+ * consumer, so an exact one cannot split a consumer's tree.
+ */
+const PUBLISHED_FIELDS = ["dependencies", "peerDependencies"] as const;
```
dustywusty: **Include `optionalDependencies` in the consumer-facing fields.**

Optional dependencies are still resolved for consumers unless explicitly omitted, so an exact cross-scope optional dependency can split the installed graph just like a normal dependency. The existing release logic in `versions.ts` already includes `optionalDependencies` for this reason. Add it here and to the registry-backed manifest data.

scripts/release/verify-cross-scope-pins.ts
```diff
@@ -0,0 +1,167 @@
...
+
+  // Every scope named in one run publishes together, so a pin one of them would
+  // supersede is about to be rewritten by another. Only the scopes left out can
+  // be left behind.
+  const releasing = new Set(scopes);
+  const problems: string[] = [];
+
+  for (const scope of scopes) {
+    const version = getCurrentVersion(scope);
+    const dependents = new Set(
```
dustywusty: **Do not discover published dependents solely from the current workspace graph.**

If a cross-scope dependency is removed locally before that package's scope is republished, the latest npm manifest still carries the old range, but this set no longer includes the package and the gate never checks it. That permits the dependency scope to advance while the published dependent remains stranded. Derive the relevant edges from published manifests for all non-releasing scoped packages (or otherwise retain published-only edges) instead of filtering them through current workspace topology.

dustywusty (changes requested):
Requesting changes for the release-safety gaps called out inline. The current implementation can begin a Channels publish before the new gate runs, and several dependency shapes can bypass the promised cross-scope validation.

PR #6863, opened 2026-09-03
feat(angular): add explicitEffect helper with explicit dependency tracking

### LINKED ISSUE #6561: 🚀 Feature Request: Make Angular `effect` dependencies explicit

deepshekhardas:
Fixes #6561.

Adds an internal xplicitEffect(track, run) helper to @copilotkit/angular: signals read in 	rack are the effect's only dependencies, while 
un always executes untracked receiving the tracked value and onCleanup.

Converts two call sites as proof (thread-store registration in 	hreads.ts, active-thread connector) with identical behavior, and adds xplicit-effect.spec.ts (re-runs on tracked change, ignores run-only reads, cleanup on re-run/destroy). Internal only: not exported from public-api.ts. No deps installed locally, so relying on CI for type checks and tests.

## Summary by CodeRabbit

* **New Features**
  * Added an Angular `explicitEffect` utility that lets developers declare reactive dependencies separately from effect execution.
  * Effect callbacks now avoid unintended subscriptions to signals read during execution.

* **Improvements**
  * Updated thread and active-thread handling to use explicit dependency tracking while preserving existing connection, registration, cleanup, and disabled-state behavior.

* **Tests**
  * Added coverage for tracked updates, ignored incidental signal reads, and cleanup during reruns and destruction.

### DIFF
--- a/packages/angular/src/lib/active-thread-connector.ts
+++ b/packages/angular/src/lib/active-thread-connector.ts
@@ -1,7 +1,8 @@
-import { effect, untracked, type Signal } from "@angular/core";
+import { type Signal } from "@angular/core";
 import { HttpAgent, type AbstractAgent } from "@ag-ui/client";
 import type { AgentStore } from "./agent";
 import type { CopilotChatConfiguration } from "./chat-configuration";
+import { explicitEffect } from "./explicit-effect";
 
 /**
  * Signature of `CopilotKitCore.connectAgent`, injected for testability.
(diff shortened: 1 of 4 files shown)


BenTaylorDev:
Closing in favor of #6578, which implements the same helper for #6561 and is further along: it converts every `effect()` call site in the package, returns the `EffectRef` that `injectInterrupt` calls `.destroy()` on, and forwards `CreateEffectOptions` so callers outside an injection context can pass an `injector`. This PR converts two call sites and leaves `threads.ts:366` on a raw `effect()`/`untracked()` pair, so the package would end up half-converted.

The helper itself is correct — `track()` followed by `untracked(() => run(...))` preserves the behavior of both converted bodies. #6578 has been open since 2026-08-19, so it is worth checking the open PRs on an issue before starting one.

PR #6874, opened 2026-09-03
docs(showcase): state the frontend-tool requirement for the last nine frameworks (closes OSS-1036)

BenTaylorDev:
## Problem

`generative-ui/tool-based` is the terminal page every onboarding run fetches, on every framework route (OSS-1034). Its `## How it works in code` section is one `<FrameworkSetup concept="frontend-tools-setup" />` call that resolves a per-framework snippet. Nine frameworks shipped no snippet, so the section rendered nothing and the `.md` output emitted `<!-- setup skipped: ... -->`. That silence meant two different things at once: "this framework needs no agent-side wiring" and "it needs wiring and nobody wrote it down".

Verified on the live host before the change: all nine emitted the skip comment, at ~2.4KB per page.

One of the nine is `built-in-agent`, which is `ROOT_FRAMEWORK`. So the page carrying the gap included the unscoped default, `https://docs.copilotkit.ai/generative-ui/tool-based.md`.

## Solution

A snippet for each of the nine. Each verdict was read out of the pinned adapter rather than out of the docs, which is what the ticket asked for.

**Nothing to wire on the agent** — confirmed in the adapter:

| framework | the line that decides it |
| --- | --- |
| `ag2` | `autogen/ag_ui/adapter.py` `run_stream` builds `client_tools` from `command.incoming.tools` |
| `mastra` | `@ag-ui/mastra@1.1.0` reduces `input.tools` into `clientTools` for `agent.stream()` |
| `strands` | `ag_ui_strands@0.2.2` calls `sync_proxy_tools(agent.tool_registry, input_data.tools, ...)` |
| `strands-typescript` | `@ag-ui/aws-strands@0.2.3` registers a proxy tool per forwarded tool in `toolRegistry` |

Both Strands adapters refuse to overwrite a native tool of the same name, so each snippet says to keep the `useComponent` name distinct.

**Real wiring:**

- `agno`. Its AG-UI interface never reads `RunAgentInput.tools`. Checked against `agno==2.6.19`: the only tool path out of `agno/os/interfaces/agui/` is `RunPausedEvent.tools_awaiting_external_execution`. So a component needs a declared `@tool(external_execution=True)` stub and the `Agent` needs a `db` to hold the paused run.
- `deepagents`. `sdk-python/copilotkit/copilotkit_lg_middleware.py` merges `copilotkit.actions` into `request.tools` inside `wrap_model_call`, so `middleware=[CopilotKitMiddleware()]` is load-bearing. Its snippet lives in the docs-only tree, because `deepagents` has no integration package.

**Mode-dependent:**

- `built-in-agent`. Config mode merges `input.tools` for you (`packages/runtime/src/agent/index.ts`). A factory owns the model call and forwards nothing, so it has to pass them itself. Both halves are in the snippet, plus the collision rule: a frontend tool whose name matches a server tool is dropped in favor of the server executor.

**In-repo evidence only:**

- `ms-agent-dotnet`, `ms-agent-harness-dotnet`. Their own `gen-ui-tool-based` agents render charts with an empty tool list (`tools: []` / `Tools = []`) and their instructions name the chart tools. No .NET SDK was available here to read `Microsoft.Agents.AI.Hosting.AGUI.AspNetCore` directly.

Six of the nine snippets also carry the "Tell the model when to call it" step, which is not filler. `GEN_UI_TOOL_BASED_PROMPT` in `showcase/integrations/built-in-agent/src/lib/factory/demo-prompts.ts` records what happens without it: the model emits `value: 0` for every point and renders a chart with no bars.

## Tests

- `REQUIREMENT_NOT_ESTABLISHED` is now empty, and the docblock says a new framework arriving without a snippet is fixed with a snippet, not with a name.
- The shape test pins all six auto-forwarding frameworks, plus the three that do not fit that shape (`agno`, `deepagents`, `built-in-agent`).
- `setup-concept.test.ts` used `ag2` as its example of an unbundled concept, so closing the last gap would have turned it red for the right reason. It now uses a concept name that is deliberately never bundled.

## The gate never ran in CI

Second commit. No workflow runs the shell-docs vitest suite: `test_integration-docs.yml` runs exactly one test file, `test_unit-showcase.yml` covers harness and shell-dashboard only, and `showcase_build_check.yml` matches `showcase/**` but only builds images. Neither path filter listed the snippet files either. So the ratchet added in #6777 was local-only.

Added a scoped job and widened both filters. The job runs only the two setup-concept files on purpose: the full shell-docs suite is not green on `main`, and gating every snippet change on unrelated failures would be worse than no gate.

## Verification

Ran against a local `shell-docs` dev server on this branch. Expected `setup skipped` count is 0 everywhere:

```
ag2                        skip=0 bytes=3825
agno                       skip=0 bytes=4299
deepagents                 skip=0 bytes=3786
mastra                     skip=0 bytes=3857
ms-agent-dotnet            skip=0 bytes=3706
ms-agent-harness-dotnet    skip=0 bytes=4236
strands                    skip=0 bytes=4026
strands-typescript         skip=0 bytes=4001
built-in-agent (root path) skip=0 bytes=4723
```

`vitest run` on the two setup-concept files: 12 passed. `npm run typecheck` and `npm run lint` both exit 0.

Four tests fail in the full shell-docs suite (`llm-text.test.ts` mastra tool-rendering, three in `angular-docs-content.test.ts`). All four fail identically at the pre-change baseline in the same worktree, so they are pre-existing on `main` and untouched by this branch.

## Two findings this turned up, not fixed here

1. **The `agno` showcase demo contradicts its own snippet.** Every `useFrontendTool` tool in agno's demos has a matching `external_execution` stub, but the `useComponent` ones do not: `render_bar_chart`, `render_pie_chart`, `query_notes` and `highlight_note` have no backend declaration, and `_run_main_agent_hitl_aware` in `src/agent_server.py` cannot forward them. Its e2e spec cannot catch this, because it asserts only that an `svg` is visible inside an assistant message, which any icon satisfies.
2. **Four frameworks redirect the HTML page away.** For `ag2`, `agno`, `deepagents` and `mastra`, `/<slug>/generative-ui/tool-based` 301s to `/<slug>/generative-ui/tool-rendering`. Confirmed on the live host as well as locally, so it predates this branch. Their `.md` output serves the page normally, which is the path onboarding runs fetch, but a human browsing those four never reaches it.

Closes OSS-1036.

## Summary by CodeRabbit

- **Documentation**
  - Added frontend tool setup guides for AG2, Agno, Built-in Agent, Mastra, Microsoft Agent integrations, Strands, and Deep Agents.
  - Documented tool registration, prompting, execution behavior, persistence, and middleware configuration.

- **Tests**
  - Expanded coverage for frontend tool setup documentation across supported integrations.
  - Updated setup concept tests to use a stable, intentionally unbundled example.

- **Chores**
  - Added automated workflow coverage for setup documentation and related tooling.

### DIFF
--- a/.github/workflows/test_integration-docs.yml
+++ b/.github/workflows/test_integration-docs.yml
@@ -6,6 +6,17 @@ on:
       - "showcase/shell-docs/src/content/**"
       - "showcase/shell-docs/model-allowlist.json"
       - "showcase/shell-docs/src/lib/__tests__/ms-agent-dotnet-provider.test.ts"
+      # The setup-concept gate below reads snippets owned by the integration
+      # packages and by the docs-only snippet tree, so a new or edited snippet
+      # has to trigger this workflow. Before OSS-1036 neither path was listed,
+      # and the coverage ratchet never ran in CI at all.
+      - "showcase/integrations/*/docs/setup/**"
+      - "showcase/integrations/*/manifest.yaml"
+      - "showcase/shell-docs/src/lib/setup-content.ts"
+      - "showcase/shell-docs/src/lib/setup-concept.tsx"
+      - "showcase/shell-docs/src/lib/__tests__/frontend-tools-setup-coverage.test.ts"
+      - "showcase/shell-docs/src/lib/__tests__/setup-concept.test.ts"
+      - "showcase/scripts/bundle-setup-content.ts"
       - "showcase/shell-docs/package.json"
       - "showcase/shell-docs/package-lock.json"
       - "examples/integrations/ms-agent-framework-dotnet/**"
@@ -18,6 +29,17 @@ on:
       - "showcase/shell-docs/src/content/**"
       - "showcase/shell-docs/model-allowlist.json"
       - "showcase/shell-docs/src/lib/__tests__/ms-agent-dotnet-provider.test.ts"
+      # The setup-concept gate below reads snippets owned by the integration
+      # packages and by the docs-only snippet tree, so a new or edited snippet
+      # has to trigger this workflow. Before OSS-1036 neither path was listed,
+      # and the coverage ratchet never ran in CI at all.
+      - "showcase/integrations/*/docs/setup/**"
+      - "showcase/integrations/*/manifest.yaml"
+      - "showcase/shell-docs/src/lib/setup-content.ts"
+      - "showcase/shell-docs/src/lib/setup-concept.tsx"
+      - "showcase/shell-docs/src/lib/__tests__/frontend-tools-setup-coverage.test.ts"
+      - "showcase/shell-docs/src/lib/__tests__/setup-concept.test.ts"
+      - "showcase/scripts/bundle-setup-content.ts"
       - "showcase/shell-docs/package.json"
       - "showcase/shell-docs/package-lock.json"
       - "examples/integrations/ms-agent-framework-dotnet/**"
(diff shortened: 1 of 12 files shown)


BenTaylorDev:
Audit pass on my own snippets found two real defects in the .NET pair, fixed in c68600b3bd.

`Microsoft.Agents.AI.Harness` is the NuGet **package** name, not a namespace. `AsHarnessAgent` and `HarnessAgentOptions` resolve from `Microsoft.Agents.AI`, and no `.cs` file in the repo carries a `using` for the package name. Both snippets were also missing `using Microsoft.Extensions.AI`, where `AsIChatClient()` and `ChatOptions` live. All twelve agent files that call `AsHarnessAgent` use exactly those two namespaces.

The harness snippet also referenced `HarnessMaxContextWindowTokens` and `HarnessMaxOutputTokens` without showing a value, so it now defines both.

No .NET SDK is available in my environment, so this was checked against the `using` blocks of the real agent files rather than by compiling. That limitation is the same one recorded in the snippet's own evidence note.

Every import and API in the other seven snippets was checked against ground truth in the same pass, and all of them hold:

| snippet | checked against |
| --- | --- |
| `ag2` | installed `ag2` 0.14.0: `ConversableAgent`, `LLMConfig`, `autogen.ag_ui.AGUIStream`, and the `functions` / `system_message` / `llm_config` / `human_input_mode` kwargs |
| `agno` | `agno==2.6.19` wheel: `agno.agent.Agent`, `agno.models.openai.OpenAIChat`, `agno.tools.tool` with `external_execution`, `agno.db.sqlite.SqliteDb(db_file=)`, and the `db` / `tools` / `instructions` kwargs |
| `strands` | `ag_ui_strands@0.2.2` `StrandsAgent(agent=, name=, description=)`, plus `OpenAIModel(model_id=)` as the repo uses it |
| `strands-typescript` | `@ag-ui/aws-strands@0.2.3` export list, and the repo's own `@strands-agents/sdk` imports |
| `mastra` | `@mastra/core/agent`, and `@ai-sdk/openai` which is a real dependency and what every other mastra docs page imports |
| `built-in-agent` | `AgentFactoryContext` really provides `abortSignal` (and prefers it over `abortController`); `convertToolsToVercelAITools` and `convertMessagesToVercelAISDKMessages` are reachable from `@copilotkit/runtime/v2` via `export * from "../agent"` |
| `deepagents` | `CopilotKitMiddleware` is in `sdk-python/copilotkit/__init__.py`'s `__all__` |

One note on `agno` worth recording here: its authored `generative-ui/your-components/display-only` page already carries the session-storage callout, which only makes sense if a display-only component goes through the paused-run path. So the new snippet completes that page rather than contradicting it.

Re-verified after the fix: both .NET `.md` pages render with zero `setup skipped`, the corrected `using` present and the fabricated namespace gone. 12 tests pass, typecheck and lint exit 0, and the full suite still shows only the same four failures that are red on `main`.

PR #6887, opened 2026-09-04
Make docs search trustworthy and surface Intelligence for capability searches

lukasmoschitz:
Closes OSS-1079.

Docs search was not trustworthy. Showcase rows competed with docs rows and some of them were dead links, abandoned pages answered as confidently as maintained ones, and the result order fell back on whatever order the index happened to be built in. On top of that, a reader who types the word for a capability — "analytics", "learning", "persistence" — had no way to learn that the capability is called Intelligence.

Three changes, plus a PRD at `prds/oss-1079-docs-search-and-intelligence-ctas.md` that records the decisions and the measurements behind them.

## Showcase rows leave the docs search

The modal loaded the showcase registry and pushed integration and feature rows into the result list. Every feature row was emitted with `href: "/"`, so a query like "chat" produced fifteen differently-titled rows that all navigated to the docs home page ("tools" six, "state" five). Both loops are gone. The registry is still loaded — the framework scope picker is built from it.

The generator now writes a different payload per target: the showcase home, the integrations explorer and the feature matrix stay in the index the showcase app consumes and leave the one the docs app consumes. Two hand-written entries that duplicated real pages ("Home", "API Reference") are removed outright.

## The index follows what a reader can actually reach

The docs portion of the index was a walk over every `.mdx` file on disk. It now derives from navigation: a page is indexed when it appears in a navigation surface, or when a page that appears in one links to it.

Navigation is not one tree. The union covers the root tree, one surface per non-hidden integration — using the same builder the app picks for that integration's `docs_mode` — and the sixteen `"root": true` sub-trees the root walk deliberately skips. Group nodes contribute their own slug, because a folder with an index page renders as an expandable parent. Entries whose destination is off-site contribute nothing, so no result promises a docs page that does not exist. Channel pages bypass the filter entirely; they are re-routed onto the Slack and Teams surfaces at runtime and filtering them against the docs navigation would delete live, linked pages.

A page's inlined snippets count as part of that page. Many pages here are thin wrappers around shared snippets — the Intelligence overview's whole body is one include — so a scan that reads only the page file sees almost nothing: 868 link targets before resolving includes, 1964 after. Resolution reuses the docs renderer's own inlining rather than a second resolver.

`search: false` forces a reachable page out and `search: true` forces an unreachable one in. As shipped no page needs either, and a test asserts the forced-in set is empty so reaching for the override stays a deliberate act.

Route-group segments are stripped from destinations, so results no longer take a redirect hop. Titles come from navigation where the two disagree, with guards so auto-derived group titles and `nav_title: Overview` do not leak into a flat result list. A coverage floor fails the build if the docs portion collapses.

## Order is deterministic, and Intelligence keywords get one recommendation

The old scoring rewarded a query that matched the *start* of a title, which is why "threads" put the threads-drawer component above the canonical Threads guide. That slot is now a whole-word test at the same weight, and everything still tied is settled by shorter-title-first and then alphabetically. Nothing depends on index construction order any more. No content family gets a bonus — deliberately, including Intelligence.

A curated keyword table drives a single recommendation block above the result list. Five entries today: threads, self-hosting, learning, analytics, and intelligence as the catch-all. Triggering is whole-word, or a typed prefix of at least four characters, never a plain substring. Specificity is declared, so "intelligence threads" resolves to the threads entry, and exactly one block ever renders. Every destination is an internal docs route.

`learning` and `analytics` point at the Intelligence overview because no docs page describes either capability yet — they are rows on that overview which link out for the detail. The overview is also the right landing for someone already running Intelligence, for whom a connect-in-five-minutes quickstart would be useless. Retargeting when those pages ship is a one-line edit.

The block reuses the in-page Intelligence CTA visual language (accent stripe, kite, theme tokens), is a labelled group rather than a fourteenth listbox option, participates in the existing arrow-key selection, and reports its surface and matched keyword through the same attribution helpers the other docs CTAs use.

## Numbers

| | before | after |
|---|---|---|
| docs-search index entries | 950 | 851 |
| of which docs pages | 678 | 583 |
| destinations with more than one entry | 2 | 0 |
| destinations taking a redirect hop | 36 | 0 |

Of the 583 docs entries, 637 slugs come from navigation, 9 more from an inbound prose link, and 94 entries left search as unreachable.

## Follow-ups this PR deliberately does not take

Two groups inside those 94 are substantive content, not leftovers, and both want a content decision rather than a search fix. The two step-by-step tutorial series (24 entries) form a closed link island: no navigation lists them and nothing reachable links in, so even unlimited transitive reachability would not find them. And 28 LangGraph pages under `advanced`, `troubleshooting` and `your-components` have zero inbound links anywhere, because the LangGraph navigation file references neither subtree. Adopt, redirect or delete — each is its own call.

Also untouched: content work on the What's New pages (the roll-up page leaves search and stays on disk), the near-duplicate search modal in the showcase app, the sitemap, and the matching engine itself.

## Verification

`npm test` 783 passing in `showcase/shell-docs`. Six failures remain and are pre-existing — verified by running the same files against a clean checkout of `origin/main`, where they fail identically: three `public-assets` PNG checks against unmaterialized LFS pointers, two `angular-docs-content` content checks, one `llm-text` mastra example. Typecheck clean, lint clean, production build green.

Checked in the browser against a local dev server: "threads" now leads with the Threads guide, "intelligence" leads with the Intelligence landing page, "chat" returns only real docs destinations, "learning" shows the recommendation, and the console is clean.

---

# Follow-up round, from testing the branch in a browser

Seven further changes after the first pass was tried by hand.

## The AG-UI "What's New" page is unlinked from search and nav

`/ag-ui/development/updates` is titled "What's New", holds a single entry dated April 2025, and throws on render because it uses an `<Update>` component this app does not provide — a Mintlify component from the upstream AG-UI docs. It is almost certainly the page the ticket's second complaint is about.

Our reachability rule did not catch it: the AG-UI section keeps its own hand-maintained published-slug list, which this PR deliberately does not reinterpret, and the page is listed in the AG-UI sidebar. So it is removed from both by hand, with a test and a floor assertion so a typo in that hand-maintained list cannot silently empty the AG-UI portion of the index.

The MDX itself is untouched. AG-UI docs are canonical upstream in `ag-ui-protocol/ag-ui` and this tree is a downstream mirror, so the page is unlinked here rather than edited.

### Open question for reviewers: should the AG-UI mirror be in these docs at all?

While tracking that page down, the mirror turned out to be an orphaned island:

- The site chrome has Docs, Reference and Cookbook. There is no AG-UI entry, in the desktop nav, the mobile nav or the footer.
- Nothing outside `/ag-ui/*` links into it. The only inbound links come from the section's own pages.
- Where our prose does reference AG-UI it uses `/ag-ui-protocol`, which redirects to `/agentic-protocols/ag-ui` in the main docs — not into the mirror. Ten other places link straight out to `docs.ag-ui.com`.
- `docs.ag-ui.com` is live and canonical, and the mirror is still being synced (last touched three days ago).

So 96 pages are maintained, served, and unreachable by anyone reading the docs, while search offers 42 of them. By the rule this PR introduces, the whole island fails the reachability test; it survives only because AG-UI is exempted.

Three ways forward, none taken here: keep it and give it a nav entry, which makes it legitimate in one move; drop the whole mirror from search while leaving the pages served; or retire the mirror and redirect `/ag-ui/*` to `docs.ag-ui.com`. The last is the real cleanup but has SEO consequences and spans two repos. Deliberately left as a product decision rather than folded into a search PR.

## Ranking: the reader's own frontend first, V1 last

Searching "chat" returned 78 matches, roughly 38 of them belonging to frontends the reader is not using — 13 Angular feature pages, 10 Vue, 9 Angular reference, 6 React Native — plus 5 from the deprecated V1 reference. A React reader got a screen of Angular.

Entries carrying another frontend are now demoted, never filtered: someone who wants Angular from a React page still finds it by typing "angular chat", and because both words then match, those rows come straight back to the top. Filtering would have hidden them outright, which is the failure this deliberately avoids. The penalty is waived as soon as the query names the frontend.

The deprecated V1 reference always sorts last, as a separate tier rather than a weight another bonus could out-argue, and those rows now carry a muted "Deprecated" tag.

## Matching survives punctuation and camelCase

`whats new` found nothing though the announcement pages exist, and `use Copilot kit` found nothing though `useCopilotKit` does. Punctuation — apostrophes, hyphens, dots, slashes — is now ignored on both sides of the comparison, and camelCase identifiers are split into words on the index side, so `use Copilot kit` reaches `useCopilotKit` and `ag ui` reaches `AG-UI`. Both forms feed the title tiers of the score as well, otherwise the page matches but never ranks. The haystack keeps the unsplit form too, so `copilotruntime` still finds `CopilotRuntime`.

The CTA keyword matcher is untouched: it stays whole-word-or-four-character-prefix, so loosening the result matching cannot make the recommendation block fire on unrelated queries.

## Keyboard, mouse and the recommendation block

- Arrowing past the sixth result used to move the selection out of sight — the list has a fixed height and nothing ever scrolled it. It now follows the keyboard, minimally, and only for keyboard moves: hover still selects but must not slide the list out from under the pointer. The selection stays on the last item rather than wrapping.
- The arrow icon at the end of a row lit up one row too low whenever the recommendation block was visible; the row background accounted for the block's selection slot and the icon did not.
- The whole recommendation block is clickable now, not just its primary link, via a card-link overlay so the two or three secondary links keep their own destinations. Nested anchors would have been invalid markup; the cost is that text inside the block is no longer selectable.
- Buttons in the modal — result rows, framework picker, close, CTA links — get a pointer cursor back, which Tailwind 4 dropped.

## Verification of this round

Browser-checked on a local dev server: keyboard scrolling follows selection while hover does not scroll, selection clamps at the last row, the arrow accent tracks the selected row, a click on the block's body lands on the Threads guide while a click on "Connect Intelligence" lands on the Intelligence quickstart, the V1 row sorts last with its tag, `what's new` and `whats new` both return the announcement pages, `use Copilot kit` surfaces `useCopilotKit`, "chat" is free of foreign-frontend rows and "angular chat" brings them back.

`npm test` 806 passing with the same 6 pre-existing failures. Lint and typecheck clean.

## Summary by CodeRabbit

* **New Features**
  * Improved documentation search relevance with predictable ranking, frontend-aware results, and better keyboard navigation.
  * Added curated Intelligence recommendations with accessible links and conversion tracking.
* **Bug Fixes**
  * Excluded broken, outdated, duplicate, and incorrect documentation links from results.
  * Improved handling of punctuation, camelCase terms, redirects, and canonical URLs.
  * Removed showcase destinations and the unlinked AG-UI “What’s New” page from docs search.
* **Documentation**
  * Updated navigation and search coverage for AG-UI documentation.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -3985,6 +3985,9 @@ importers:
       glob:
         specifier: ^10.4.0
         version: 10.5.0
+      gray-matter:
+        specifier: ^4.0.3
+        version: 4.0.3
       js-yaml:
         specifier: '>=4.1.1'
         version: 4.1.1
@@ -18972,6 +18975,10 @@ packages:
     resolution: {integrity: sha512-DKKrynuQRne0PNpEbzuEdHlYOMksHSUI8Zc9Unei5gTsMNA2/vMpoMz/yKba50pejK56qj98qM0SjYxAKi13gQ==}
     engines: {node: ^12.22.0 || ^14.16.0 || ^16.0.0 || >=17.0.0}
 
+  gray-matter@4.0.3:
+    resolution: {integrity: sha512-5v6yZd4JK3eMI3FqqCouswVqwugaA9r4dNZB1wwcmrD02QkV5H0y7XBQW8QwQqEaZY1pM9aqORSORhJRdNK44Q==}
+    engines: {node: '>=6.0'}
+
   groq-sdk@0.5.0:
     resolution: {integrity: sha512-RVmhW7qZ+XZoy5fIuSdx/LGQJONpL8MHgZEW7dFwTdgkzStub2XQx6OKv28CHogijdwH41J+Npj/z2jBPu3vmw==}
 
@@ -24408,6 +24415,10 @@ packages:
   scule@1.3.0:
     resolution: {integrity: sha512-6FtHJEvt+pVMIB9IBY+IcCJ6Z5f1iQnytgyfKMhDKgmzYG+TeH/wx1y3l27rshSbLiSanrR9ffZDrEsmjlQF2g==}
 
+  section-matter@1.0.0:
+    resolution: {integrity: sha512-vfD3pmTzGpufjScBh50YHKzEu2lxBWhVEHsNGoEXmCmn2hKGfeNLYMzCJpe8cD7gqX7TJluOVpBkAequ6dgMmA==}
+    engines: {node: '>=4'}
+
   secure-json-parse@2.7.0:
     resolution: {integrity: sha512-6aU+Rwsezw7VR8/nyvKTx8QpWH9FrcYiXXlqC4z5d5XQBDRqtbfsRjnwGyqbi3gddNtWHuEk9OANUotL26qKUw==}
 
@@ -25000,6 +25011,10 @@ packages:
   strip-ansi@7.2.0:
     resolution: {integrity: sha512-yDPMNjp4WyfYBkHnjIRLfca1i6KMyGCtsVgoKe/z1+6vukgaENdgGBZt+ZmKPc4gavvEZ5OgHfHdrazhgNyG7w==}
     engines: {node: '>=12'}
+
+  strip-bom-string@1.0.0:
+    resolution: {integrity: sha512-uCC2VHvQRYu+lMh4My/sFNmF2klFymLX1wHJeXnbEJERpV/ZsVuonzerjfrGpIGF7LBVa1O7i9kjiWvJiFck8g==}
+    engines: {node: '>=0.10.0'}
 
   strip-bom@3.0.0:
     resolution: {integrity: sha512-vavAMRXOgBVNF6nyEEmL3DBK19iRpDcoIwW+swQ+CbGiu7lju6t+JklA1MHweoWtadgt4ISVUsXLyDq34ddcwA==}
@@ -47948,6 +47963,13 @@ snapshots:
 
   graphql@16.12.0: {}
 
+  gray-matter@4.0.3:
+    dependencies:
+      js-yaml: 4.1.1
+      kind-of: 6.0.3
+      section-matter: 1.0.0
+      strip-bom-string: 1.0.0
+
   groq-sdk@0.5.0(encoding@0.1.13):
     dependencies:
       '@types/node': 18.19.130
@@ -55856,6 +55878,11 @@ snapshots:
 
   scule@1.3.0: {}
 
+  section-matter@1.0.0:
+    dependencies:
+      extend-shallow: 2.0.1
+      kind-of: 6.0.3
+
   secure-json-parse@2.7.0: {}
 
   select-hose@2.0.0: {}
@@ -56747,6 +56774,8 @@ snapshots:
   strip-ansi@7.2.0:
     dependencies:
       ansi-regex: 6.2.2
+
+  strip-bom-string@1.0.0: {}
 
   strip-bom@3.0.0: {}
 
--- a/showcase/scripts/generate-search-index.ts
+++ b/showcase/scripts/generate-search-index.ts
@@ -358,6 +358,18 @@ function isFrontendDocsEntry(href: string): boolean {
 }
 
 async function main() {
+  // A content-free build may emit a stub, but a partially staged tree must
+  // never overwrite a complete search index with missing content families.
+  const contentRoots = ["docs", "reference", "ag-ui"].map((name) =>
+    path.join(CONTENT_ROOT, "content", name),
+  );
+  const missingRoots = contentRoots.filter((root) => !fs.existsSync(root));
+  if (missingRoots.length > 0 && missingRoots.length < contentRoots.length) {
+    throw new Error(
+      `[generate-search-index] incomplete content tree; missing: ${missingRoots.join(", ")}`,
+    );
+  }
+
   const entries: SearchEntry[] = [];
 
   // Static pages. `/` is the showcase front door — it is filtered out of
--- a/showcase/scripts/generate-search-index.ts
+++ b/showcase/scripts/generate-search-index.ts
@@ -4,12 +4,19 @@
  * Scans reference MDX files, AG-UI content, and registry data to produce
  * a search index JSON for the shell's Cmd-K search modal.
  *
+ * Docs pages are filtered to the ones a reader can actually reach from a
+ * sidebar — see shell-docs/src/lib/searchable-pages.ts.
+ *
  * Usage: npx tsx showcase/scripts/generate-search-index.ts
  *
- * Output: showcase/shell/src/data/search-index.json
+ * Output: showcase/shell-docs/src/data/search-index.json (docs app)
+ *         showcase/shell/src/data/search-index.json      (showcase app)
+ *         showcase/shell-docs/src/data/search-index-dropped.json (review)
  */
 
+import { spawnSync } from "child_process";
 import fs from "fs";
+import os from "os";
 import path from "path";
 import { fileURLToPath } from "url";
 
@@ -22,14 +29,41 @@ const ROOT = path.resolve(__dirname, "..");
 // (which links to docs routes) stays functional — destinations 301 across
 // to docs.showcase.copilotkit.ai. We SCAN from shell-docs (source of truth)
 // and WRITE to both.
-const SHELL_DOCS_DIR = path.join(ROOT, "shell-docs", "src");
+const SHELL_DOCS_ROOT = path.join(ROOT, "shell-docs");
+const SHELL_DOCS_DIR = path.join(SHELL_DOCS_ROOT, "src");
 const SHELL_DIR = path.join(ROOT, "shell", "src");
 const CONTENT_ROOT = SHELL_DOCS_DIR;
 const SHARED_DIR = path.join(ROOT, "shared");
-const OUTPUT_PATHS = [
-  path.join(SHELL_DOCS_DIR, "data", "search-index.json"),
-  path.join(SHELL_DIR, "data", "search-index.json"),
+
+type OutputTarget = "shell-docs" | "shell";
+
+const OUTPUTS: ReadonlyArray<{ target: OutputTarget; path: string }> = [
+  {
+    target: "shell-docs",
+    path: path.join(SHELL_DOCS_DIR, "data", "search-index.json"),
+  },
+  { target: "shell", path: path.join(SHELL_DIR, "data", "search-index.json") },
 ];
+
+// Destinations that live on the showcase host rather than the docs host.
+// The showcase app IS the showcase and needs these rows; the docs app must
+// not offer them, because clicking one leaves the documentation.
+const SHOWCASE_HOST_DESTINATIONS = new Set(["/", "/integrations", "/matrix"]);
+
+// The list of dropped docs entries, for review. Lands in the gitignored
+// generated-data directory next to the index it explains.
+const DROPPED_REPORT_PATH = path.join(
+  SHELL_DOCS_DIR,
+  "data",
+  "search-index-dropped.json",
+);
+
+// Coverage floor for the docs portion of the shell-docs index. Measured at
+// 583 entries when the navigability filter was introduced; the floor sits
+// ~15% under that so ordinary content churn (a section retired, a folder
+// merged) passes while a broken navigation walk — which collapses the
+// union to near zero — fails the build instead of shipping an empty search.
+const MIN_DOCS_ENTRIES = 495;
 
 interface SearchEntry {
   type: "page" | "reference" | "ag-ui";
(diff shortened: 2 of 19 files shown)


BenTaylorDev (changes requested):
Reviewed against current `main` (the branch is up to date with it). Verified in a worktree: full `vitest`, `tsc --noEmit`, `oxlint`, the generator run in two build contexts, and three mutation checks against the new tests.

The work is careful and the tests are real. One finding blocks merge.

## Blocking: the showcase app loses every docs row from its search

`showcase/shell/Dockerfile:19` copies `showcase/shell-docs/src/content/` into the shell build. It does not copy `shell-docs/scripts/` or `shell-docs/node_modules`.

So in that build `docsDir` **exists**, `loadSearchablePages()` runs, both `fs.existsSync` guards fail, and it returns `null`. The result is zero docs entries. The `MIN_DOCS_ENTRIES` floor is deliberately skipped when `pages` is null, so the build stays green on a `console.warn`.

Measured by running the generator in that exact context (content tree present, shell-docs toolchain absent):

| | `showcase/shell` index | docs rows |
|---|---|---|
| `origin/main` | 950 | 678 |
| this PR | 270 | **0** |

The 583 docs rows this PR intends to ship do not reach the showcase app at all.

Three things make this land rather than stay theoretical:

- The `shell` slot's path filter in `.github/workflows/showcase_build.yml:145-149` includes `showcase/scripts/**`. This PR edits `generate-search-index.ts`, so the shell image rebuilds on this PR.
- That slot builds `showcase/shell/Dockerfile` with `context: "."` and carries a `railway_id`, so it is the deployed artifact.
- `showcase/shell/src/components/search-modal.tsx:44-52` does no type filtering. It matches every index row and slices to 12, so the docs rows are genuinely presented today.

The premise in the new comment is the part to revisit:

> Either the content tree is absent (the `shell` Docker build) or shell-docs' own toolchain is

The first half does not hold. In the shell Docker build the content tree is present, and only the toolchain is missing. Both halves still land on `pages = null`, so the reasoning reads as a no-op while the outcome is a regression. The file's own header comment, which this PR leaves in place, states the opposite intent:

> WRITE to both ... so the showcase app's search (which links to docs routes) stays functional

Either the shell build needs the reachability payload (copy `shell-docs/scripts` plus `src/lib` and install, or emit the payload once and read it back as data), or the decision to drop docs rows from the showcase app belongs in the PR description, and the floor check must not pass silently at zero.

## Non-blocking

**Three of the four frontend-demotion tests do not exercise the demotion.** With `FOREIGN_FRONTEND_PENALTY = 0`, only "keeps other frontends below agnostic rows for a React reader" fails. The React Native case, "puts Angular back on top when the query names it", and "follows the reader onto an Angular surface" all still pass, so they cannot tell a working waiver from no penalty at all.

**`/faq` and `/examples` leave search and are not in the follow-ups list.** Both are correctly dropped by the rule: I confirmed each appears in no `meta.json` and has no inbound link. But `faq.mdx` is titled "Common Questions", so the query "common questions" stops finding it. That looks worth a content call alongside the two tutorial series and the LangGraph subtrees you already called out. For contrast, `/headless` and `/agentic-chat-ui` are `seo-redirects.ts` sources, so dropping those is plainly right.

**`aria-activedescendant` points at an `<a>` inside a `role="group"`**, not at an option inside the controlled listbox. Assistive technology may not announce the recommendation when the selection lands on it.

**Nit:** the placeholder still reads "Search docs, API reference, integrations..." while the integrations explorer and the feature rows have left. Integration docs pages are still searchable, so it is not wrong, only a little stale.

## What I verified

- Full `vitest`: 807 passed, 5 failed. The same 5 fail identically on `origin/main` in the same environment, so they are pre-existing.
- `tsc --noEmit` clean. `oxlint` reports 0 errors and 50 pre-existing warnings, none in the new files.
- The generator reproduces your numbers: 583 docs entries, 850 and 853.
- Mutation checks confirm the new tests are load-bearing. Reverting the whole-word tier to `startsWith` fails the Threads ordering test. Disabling the reachability filter fails both navigability tests. Setting the frontend penalty to 0 fails one test.
- The tests render the real `SearchModal` against the real generated index, so the source of truth is not mocked away.

The single thing to settle before merge: whether the showcase app is meant to keep its docs search rows.

lukasmoschitz:
Good catch on the shell index — you were right, and the premise you quoted was mine to fix. I had carried over the old comment claiming the shell build omits the docs tree without checking the Dockerfile.

**Fixed** in 12580a1a + 76b4ffb7:

- The generator now runs the emit script with its own `node_modules` on `NODE_PATH`, so no shell-docs install is needed anywhere. The shell Dockerfile stages `src/lib`, `scripts` and `tsconfig.json`; `gray-matter` and `tsx` come from `scripts` (both lockfiles updated — the npm one is what `validate-runtime-routes` tripped on).
- Missing toolchain now throws instead of warning. The floor is keyed on the content tree again, not on whether the filter ran — that combination was the hole.
- Added `search-index-build-contract.test.ts`: any Dockerfile staging docs content *and* generating the index must stage the toolchain. It caught shell-dashboard on the first run, which stages content for `probe-docs.ts` only, so the condition is narrowed to builds that actually generate an index.

Verified by running the generator with `shell-docs/node_modules` moved aside, from `scripts/` and from `shell/`: 583 docs entries either way, shell index back to 853.

**Frontend tests:** two of the four now fail with the penalty at 0, and the other two no longer pretend to cover it. The demotion is only isolable where copies share type, title and length, so that test asserts the foreign reference copies lose their visible slots. The Angular-surface one pins surface detection instead, and fails if `activeFrontend` is ignored. Ordering also gained a destination tie-break, so it is a total order.

The waiver test is gone, with a comment explaining why: every term must match, so naming a frontend filters the agnostic rows out before ranking. "angular chat" matches 24 rows, 23 of them Angular's own. There is nothing for a demoted row to lose to, and the old test passed with the waiver removed.

**`/faq` and `/examples`** added to the follow-ups in the PRD, with the "Common Questions" problem called out. Placeholder updated.

**`aria-activedescendant`: not changed, and I'd like your read.** I tried it — primary action wrapped in a one-option listbox with `role="option"`. It costs the link role, so AT stops announcing it as a link and "open in new tab" is no longer conveyed; a test asserting that failed immediately. Six tests also had to change because role-based queries then count the block as a result, which is the property we deliberately built. `aria-controls` does already include the block, so the active descendant is inside a controlled element — what's left is the unusual role. The clean fix is making the recommendation a real first option of the results listbox, which inverts this PR's "not a result" decision. Happy to do that if you think it's the right call, but it felt like a design change rather than a review fix.

BenTaylorDev (dismissed):
Re-reviewed at `76b4ffb7`, up to date with `main`. I re-ran the blocking case in a Docker-layout simulation and in the PR's own CI build log, plus the full `vitest` suite five times, `tsc --noEmit`, `oxlint`, and four mutation checks.

## The blocking finding is fixed

The showcase app keeps its docs rows. I rebuilt the `shell` Docker layout by hand (content tree staged, no `shell-docs` install, `npm ci` in `showcase/scripts` from the updated lockfile) and got 583 docs entries and an 853-entry shell index. `gray-matter` and `tsx` resolve from the `scripts` install through `NODE_PATH`, and no `react` is needed, so the CJS fallback holds.

The PR's own CI proves it in the deployed artifact's build, not only in my simulation. From `build-check (shell)` on run 33883684449:

```
#27 2.216 [emit-searchable-pages] 646 searchable slugs (637 from navigation, 9 from an inbound link, ...)
#27 2.247   Docs: 583 entries (94 dropped as unreachable from any sidebar; ...)
#27 2.254 Search index: /app/shell/src/data/search-index.json (853 entries)
```

I also confirmed the loud failure. With `emit-searchable-pages.ts` removed, the generator exits 1 with a message that names both candidate toolchain paths and the three Dockerfile `COPY` lines. That throw is the guard that matters, because it runs inside `build-check`. A future Dockerfile edit that drops those lines reddens the shell build.

Mutation checks, all confirming the tests carry weight:

| mutation | result |
|---|---|
| delete the three `COPY` lines from `shell/Dockerfile` | `search-index-build-contract.test.ts` fails on the `shell` case |
| `FOREIGN_FRONTEND_PENALTY = 0` | 2 tests fail (was 1) |
| `activeFrontend = null` | "follows the reader onto an Angular surface" fails |
| remove `emit-searchable-pages.ts` | generator exits 1 |

`/faq`, `/examples` and the "Common Questions" problem are in the PRD follow-ups. The placeholder is updated. The arrow offset is applied at every call site.

## Your `aria-activedescendant` question

Keep the link. Drop `CTA_OPTION_ID` from `aria-activedescendant` and announce the selection through a polite live region instead.

The reason your experiment cost so much is that `role="option"` replaces the link role, and the link role is the part that carries real value here: the primary action is a Tab-reachable `<a href>` with a descriptive `aria-label`, so a screen reader user already reaches and hears the recommendation without virtual focus at all. What is missing is only the announcement when arrow keys land on it. A live region buys that back for a few lines, keeps the "not a result" decision intact, and removes an IDREF whose role the combobox pattern does not define. Making the block a real first option of the results listbox is the other coherent answer, and I agree it is a design change, not a review fix.

## Non-blocking

**`/ag-ui/development/updates` is still a broken URL.** The route gates on file existence, not on `NAV_DEFINITION`, and the MDX is still on disk with its single `<Update>` tag. `mdx-registry.tsx` provides no `Update` component. So the page still throws for anyone arriving from Google, a bookmark or an old link. Removing it from nav and search fixes what the ticket reported, and I am not asking you to edit vendored MDX. A `seo-redirects.ts` entry to `docs.ag-ui.com/development/updates` closes the URL without touching the content, and it is the one piece of the open question that needs no product decision.

**The new DOM tests flake under load.** Both `search()` helpers use `waitFor` at its default 1000 ms timeout, and the modal waits on a dynamic import of `registry.json` first. On a cold first run I got 6 failures across the two files (`shows exactly one block for a keyword query`, `shows the most specific block, never two`, `stays away from unrelated queries`, and the three `routes every "…" result to a real docs route` cases). Four later runs of the same suite were clean. An explicit `{ timeout: 5000 }` on the registry wait removes it.

**Nothing in CI runs this suite.** `test / integration / docs` runs one file, `ms-agent-dotnet-provider.test.ts`, and no other workflow runs `vitest` in `shell-docs`. That is pre-existing and not yours to fix here, but it is worth knowing which of your guards CI actually enforces: the generator throw does, the contract test does not. It also means the flake above costs a local runner rather than a red PR.

**`readPageSourceWithSnippets` swallows the error with `console.warn` silenced.** If `inlineSnippets` starts throwing, link discovery quietly falls back to the raw page. The blast radius today is small and I want to state it accurately: 9 pages are kept by an inbound prose link alone, and the 495-entry floor would not notice losing them. Logging the slug and the error through the saved `warn` costs two lines.

**`aria-controls` can name an element that is not rendered.** The listbox renders only when `results.length > 0`, but line 840 always lists `RESULTS_LISTBOX_ID`. A keyword query that matches the recommendation and no organic row leaves a dangling IDREF.

**Two comments describe code that is not there.**

- `search-index-build-contract.test.ts:85` says the shell image symlinks `shell-docs/node_modules` at the scripts install. It does not. It symlinks `harness/node_modules`, and `gray-matter` reaches the child through `NODE_PATH`. The conclusion is right and the mechanism is wrong.
- `search-modal.test.tsx:48` says the recommendation block is "a separate one-option listbox of its own". It is a `role="group"` holding an `<a>`, and `search-modal-intelligence-cta.test.tsx` asserts exactly that. So the `within(list)` scoping is inert, and the comment points the next reader at a listbox that does not exist. That reads like a leftover from the experiment you described.

**`scoreResult` and `compareResults` are module-private.** Both of the two rough edges above trace to that: the frontend waiver is untestable through the result list, and the ranking tests have to render the whole modal, which is what made them flaky. Exporting the two functions, or moving ranking into `src/lib/`, makes the waiver directly testable and makes future ranking edits cheap to pin. Worth doing when ranking next changes rather than in this PR.

## What I verified

- Docker-layout simulation of `showcase/shell`: `npm ci` in `scripts` from the updated lockfile, `generate-registry.ts` then `generate-search-index.ts`. 583 docs entries, shell index 853, docs index 850, zero duplicate destinations in either, and the only `/` destination is the showcase home row that belongs to the shell app.
- Full `vitest` five times: 822 tests, 819 passing. The same 3 fail every time and none is yours: `angular-docs-content` (2) and the `llm-text` mastra case. The PR touches neither file.
- `tsc --noEmit` clean. `oxlint` reports 0 errors. The only warnings in your files are two `__SHOWCASE_CONFIG__` dangling-underscore warnings that match the existing `runtime-config.client.test.ts`.
- The generator exits 1, not 0, when the toolchain is absent.

prds/oss-1079-docs-search-and-intelligence-ctas.md
tylerslaton, line 1: let's not include plans in the PRs we make. just extra noise for agents to get confused by later.

PR #6889, opened 2026-09-04
fix(docs): unify the coding-agent prompt affordances

lukasmoschitz:
Closes OSS-1098.

The docs had six "copy a prompt for your coding agent" affordances emitting five different prompt texts, drifting apart in behaviour and analytics. This unifies them and fixes what was unambiguously wrong along the way.

`main` has since been merged in — see **Merge with main** below, which changed how one part of this is implemented.

## What changed

**One component.** `hero-onboarding-prompt-button.tsx` and the `OnboardingPromptCopyButton` inside `ai/page-actions.tsx` had drifted apart; both are gone, replaced by `onboarding-prompt-button.tsx` with `variant: "compact" | "hero"`. A verbatim port moved icon sizing from the element to the button class, which silently took the hero icon from 16px to 24px — caught in review, fixed, and pinned by a test.

**Framework landing pages have one hero, not two beginnings.** `DocsPageView` gained a `landingPage` prop for pages whose MDX owns the whole opening. Authored (`integrations/<folder>/index.mdx`) and data-driven landing pages now render identically — same lockup, same top offset, one `<h1>`.

**The framework mark was missing on authored landing pages.** Unrelated cause: `next-mdx-remote` 6 (bumped on `main` in 0a8a4e97ea) enables its expression-attribute filter by default, so an authored `frameworkIcon={<MastraIcon />}` is stripped before the JSX reaches the component. `MdxFrameworkOverview` now falls back to the registry `iconKey` the data-driven path already uses. `ms-agent-harness-dotnet` has no registry record and keeps its name-only lockup.

**The prompt carries page context.** Up to three factual sentences are appended — selected framework, selected frontend, the page's own `.mdx` URL — each independently omittable. `src/lib/intelligence-onboarding-prompt.ts` is untouched: byte-identity with Intelligence's `intelligence-home.tsx` and the Inspector is preserved, and every call site concatenates rather than mutates.

**One run id per mount, not per click.** Three clicks previously minted three ids and looked like three developers in the funnel.

**Two missing slug renames** added: `built-in-agent` → `built-in`, `deepagents` → `deep-agents`.

**Dead code removed:** `channels-activation-strip.tsx` and its test, plus the now-unowned `CHANNELS_ACTIVATION_SURFACES.docsLandingStrip`.

## Merge with main

`main` moved 83 commits during this work and refactored the same area, in one case solving the same problem independently. Resolved on the rule that there should be **one mechanism, not two side by side** — which is this ticket's own premise.

**`DocsContentHeader` (new on main) won the page header.** It owns the breadcrumb, title, description and an actions slot, takes breadcrumbs as ancestors-only, and carries a `hideHeading` prop documented for exactly the case this PR addresses — an MDX-owned hero replacing the `<h1>`. `main` also deleted the `<hr>` this branch had been suppressing. `landingPage` is now expressed through that component rather than through its own inline suppression: empty breadcrumbs, `hideHeading`, no actions slot, and the content container's top padding zeroed so the hero starts at the same offset as the data-driven path.

**`main` won the page-tools row**: its wrapper with `role="group"`, the `hideOnboardingPrompt` case (`/webmcp`), the `docs-page-actions-primary` class and the "Copy prompt" / "Copy page" labels are all kept. The button rendered inside it is this branch's unified one, carrying the page context.

**A fourth call site of the deleted hero button appeared on main** — `content/landing-pages/intelligence-overview.tsx`, created after this branch diverged. Converted to `OnboardingPromptButton variant="hero"`, keeping the `docs_intelligence_hero` surface main had already given it, and threading `markdownUrl` so the copied URL and the analytics path cannot drift.

**A silent regression was caught in the merge.** `HERO_CLASSNAME` is a verbatim port of the hero button as it stood *before* main's dark-contrast fix (2623fdd361), which renamed `--accent` → `--accent-fill` for CTA background and border while deliberately leaving the focus ring on `--accent`. Merging as-is would have reverted that fix on every hero prompt button. The class now matches main's token for token.

## Analytics

`docs.intelligence_onboarding_prompt_copied`

| Surface | Before | After |
| --- | --- | --- |
| `docs_page_tools_onboarding_prompt` | new run id per click | one per mount, same properties |
| `docs_landing_hero` | per click, no framework/frontend | per mount, still framework-free (see below) |
| `docs_framework_hero` | per click, `agent_framework?` | per mount, **+ `frontend?`** |
| `docs_intelligence_hero` | per click (main) | per mount, **+ page URL in the copied text** |
| banner surfaces (`docs_landing_learning`, `docs_threads*`, ~20 `docs_<fw>_quickstart`) | per click | per mount, **+ `agent_framework?`** on MDX-placed banners |

No surface lost this event and no `surface` value changed.

`docs_conversion_copied` — fired by `copy-tracker.tsx`, which looks for a focused `[data-docs-copy-surface]` ancestor. **Three surfaces now fire it that previously fired none:** `docs_landing_hero`, `docs_framework_hero`, and every banner surface. The old hero carried `data-surface`, which nothing reads; the banner carried no attribute at all. Expect conversion counts on these to rise from zero — that is the fix, not a regression.

`docs.channels_activation_prompt_copied` — `backend` was hard-coded `"built-in-agent"`; it now takes the page's framework, falling back to `DEFAULT_FRAMEWORK`. `/channels` is a root-surface page, so the emitted value is unchanged in practice today; the fix matters only if the component is ever placed on a framework-scoped page.

**Lost:** `markdown_copied` and `open_in_llm_clicked` no longer fire from the authored framework landing pages, because suppressing the page-actions slot removes Copy Markdown and Open alongside the duplicate prompt button. This is deliberate — it aligns those pages with the data-driven ones, which never had the row — but it is a real reduction in coverage and worth a second opinion.

## Files owned by other teams

Touched for analytics or layout only, with no change to their own copy or visual design:

- `intelligence-onboarding-prompt.tsx` — adds an optional `framework` prop (analytics only, never enters the copied text), mints one run id per mount, and adds `data-docs-copy-surface` so the banner registers as a conversion surface.
- `channels-start-prompt.tsx` — replaces the hard-coded `backend` analytics value with the page's framework via a new optional prop.
- **`docs-content-header.tsx`** — **please review this one specifically.** It now returns `null` when it has no ancestor breadcrumbs, a hidden heading and no actions, instead of an empty `<header>` whose bottom margin pushed the page body down by ~44px. No page in the tree sets all three today, so nothing existing changes behaviour, and a test pins it — but it is your component and this is a real change to it. The alternative was to have `landingPage` skip the component entirely, which was rejected because it would reintroduce the second mechanism this merge exists to remove.

`src/lib/intelligence-onboarding-prompt.ts` is **not modified**, stated for completeness.

## Decisions worth flagging

**The docs home deliberately names no framework.** Every other root-surface page names the Built-in Agent and React; `/` is the one page a reader reaches before choosing anything, so committing them to a framework there would be wrong. The component's JSDoc previously asserted the opposite and now records the exception.

**The prompt's page URL uses the configured base URL, not the request origin.** Correct — the prompt should carry the canonical docs address, not whichever host the reader happens to be on.

## Verification

Typecheck clean. Lint reports 0 errors and 42 warnings, all of rule classes that already fire elsewhere in the app.

The test baseline was **re-measured against `origin/main` after the merge**, because the pre-merge figure was no longer authoritative. `origin/main` fails 6 tests in 3 files (`public-assets` ×3 — Git LFS PNG pointers, `angular-docs-content` ×2, `llm-text` ×1). The merged tree fails **the identical set**; none is caused by this branch. Passing tests went from 641 to 695.

Five tests broke during the merge and were each fixed at the source rather than loosened — they asserted structure main had moved or deleted.

Behaviour was checked against a running dev server, not only by reading: one `<h1>` and one hero on `/mastra`, `/agno` and `/langgraph-python` with an identical content-container offset across both render paths; `/quickstart` and `/mastra/threads` keeping their header and page-actions group; `/webmcp` keeping its "Copy page" primary action; repeated clicks producing a single well-formed run id; and the copied prompt ending in the framework, frontend and page sentences with no double or missing space.

Every new guard was mutation-tested — the guarded code was broken, the test watched to fail, then reverted. Two rounds of this found tests that could not fail: a restored separator case that the deleted 663-line suite had covered and its replacement had not, and a double-click guard whose assertions all ran while the clipboard write was still pending.

The full 7-agent review loop was **not** run on this branch. It was replaced by five wave reviews, an integration review verified against the running server, and a mutation audit over the whole diff. Flagging it so the omission is visible rather than assumed.

## Follow-ups, not in this PR

- The `next-mdx-remote` 6 expression-attribute filter strips far more than the framework mark — the same landing pages are also missing their supported-features grid and live-demo switcher, and roughly 600 of ~675 MDX files contain expression props. Tracked separately; disabling the filter would change nearly every docs page in one sweep.
- `ms-agent-harness-dotnet` has a logo asset but no registry record. Adding one would flip it from the authored path to the data-driven one and replace the page, so it is not a drive-by.
- The banner and the page-tools button still both appear on ~20 quickstart pages. Out of this ticket's scope; a product call.


tylerslaton:
Thank you! Included this in https://github.com/CopilotKit/CopilotKit/pull/7181.

PR #6891, opened 2026-09-04
fix(react-core,vue,angular): recover a pending interrupt after a lost event or a reconnect

BenTaylorDev:
Fixes OSS-1131

## Problem

An interrupted thread could become permanently unusable, with no error surface at all.

The interrupt hooks captured an interrupt into a closure variable — on `onCustomEvent` for the legacy `on_interrupt` path, on `onRunFinishedEvent` for the standard path — and committed it to state only inside `onRunFinalized`. On the `connect` path that callback runs in the rxjs `finalize()` of a long-lived socket stream, so it fires at socket teardown rather than once per replayed run. A reconnect therefore delivered the replayed interrupt and surfaced nothing, and effect cleanup then called `interruptState.clear()`.

The Intelligence gateway does replay: `connect` passes `last_seen_event_id`, and `RunHandler.connectAgent` clears the cursor on a fresh thread restore to request a full history. The event arrived. The hook dropped it.

There was nothing to fall back to:

- `AbstractAgent` records `agent.pendingInterrupts` on every `RUN_FINISHED`, on both the run and connect pipelines. `useInterrupt` never read it.
- That record exists only for **standard** interrupts. A legacy `on_interrupt` interrupt was stored nowhere durable, so the legacy path — the one every CLI starter takes — had no recovery path at all.
- `AbstractAgent.onInitialize` throws `Thread has N pending interrupt(s) not addressed by resume` when that array is stale, so the thread then rejects every later run.

## Change

1. **Commit on `onRunFinishedEvent`**, so the connect stream's lifecycle no longer gates the UI. `onRunFinalized` stays as the fallback for a stream that ends without a `RUN_FINISHED` event; when `RUN_FINISHED` did arrive, the commit already ran and the fallback finds nothing.
2. **Seed on subscribe** from what the client already knows the thread is waiting on, so a mount, a remount or a reconnect surfaces a gate whose event arrived before the subscription existed.
3. **Record legacy interrupts durably** — new internal `ɵrecordLegacyInterrupt` / `ɵreadLegacyInterrupt` / `ɵclearLegacyInterrupt` in `@copilotkit/core`, a `WeakMap` keyed by agent that mirrors what `pendingInterrupts` gives the standard path. It carries the raising run's id too, so a recovered legacy gate resumes against the right run. The record is dropped when a new run starts, when a run fails, and when a legacy gate is dismissed or expires.
4. **Stop discarding state in cleanup** while the agent still records a gate.

Applied across React, Vue and Angular — the mechanism was identical in all three. Angular already seeded standard interrupts from `agent.pendingInterrupts` but still committed at finalize, so it had the connect-path half of the bug and none of the legacy half.

## Testing

Verified in a worktree with its own full `pnpm install` and freshly built workspace `dist` output, so these are clean numbers rather than shared-checkout noise. Baselines are the same commands with the eight changed files checked out from `origin/main`.

### Whole-package suites

| Package | Baseline (`origin/main`) | With this change | Delta |
| -- | -- | -- | -- |
| `@copilotkit/core` | 69 files, 830 passed | 69 files, **834 passed** | +4 |
| `@copilotkit/react-core` | 138 files passed / 2 failed, 1583 passed / 7 failed | same files, **1590 passed** / 7 failed | +7 |
| `@copilotkit/vue` | 104 files, 1103 passed | 104 files, **1108 passed** | +5 |
| `@copilotkit/angular` | 51 files, 328 passed / 1 skipped | 51 files, **332 passed** / 1 skipped | +4 |

+20 tests, all passing, no test moved from pass to fail. react-core's 7 failures are the pre-existing `pinToSend` jsdom layout tests, identical before and after.

### The new tests

`@copilotkit/core` — `src/__tests__/interrupt-state.test.ts`: no record for a fresh agent, read-back of a record, per-agent isolation, clearing.

`@copilotkit/react-core` — `src/v2/hooks/__tests__/use-interrupt.test.tsx`:

```
 ✓ recovery after a lost event or a reconnect > shows a standard gate from RUN_FINISHED on a stream that never finalizes
 ✓ recovery after a lost event or a reconnect > shows a legacy gate from RUN_FINISHED on a stream that never finalizes
 ✓ recovery after a lost event or a reconnect > seeds a standard gate from agent.pendingInterrupts on mount
 ✓ recovery after a lost event or a reconnect > recovers a legacy gate on remount
 ✓ recovery after a lost event or a reconnect > resumes a recovered legacy gate with the run that raised it
 ✓ recovery after a lost event or a reconnect > forgets a legacy gate once a new run starts
 ✓ recovery after a lost event or a reconnect > forgets a legacy gate once it is dismissed
```

The connect-path requirement is covered specifically: the two "never finalizes" tests fire `onRunFinishedEvent` and never fire `onRunFinalized`.

`@copilotkit/vue` — 5 equivalents. `@copilotkit/angular` — 4 equivalents in `src/lib/interrupt.spec.ts`.

### Mutation checks

Every new test was proved to fail without the fix it covers. Each mutation was applied to a pristine copy of the file, so no two mutations mask each other.

| Package | Mutation | Result |
| -- | -- | -- |
| react-core | stop committing at `onRunFinishedEvent` | the 2 "never finalizes" tests fail (2 failed / 40 passed) |
| react-core | remove the seeding block | the 3 recovery tests fail (3 failed / 39 passed) |
| react-core | remove every `ɵclearLegacyInterrupt` call | the 2 "forgets" tests fail (2 failed / 40 passed) |
| vue | drop the `commit()` call | the 2 "never finalizes" tests fail |
| vue | drop the seeding | the 2 recovery tests fail |
| vue | drop `ɵclearLegacyInterrupt` | "forgets a legacy gate once a new run starts" fails |
| angular | drop the `commit()` call | the 2 "never finalizes" tests fail |
| angular | drop the legacy seed | "restores an unresolved legacy interrupt when reconnecting" fails |
| angular | drop `ɵclearLegacyInterrupt` | "forgets a legacy interrupt once a new run starts" fails |

### Typecheck, lint, format

- `tsc --noEmit`: **0 errors** for `core`, `react-core` and `angular`. `vue` needs `vue-tsc` (raw `tsc` cannot resolve `.vue` SFCs): `vue-tsc --noEmit` reports **0 errors**.
- `oxlint` on all eight changed files: **0 warnings, 0 errors**.
- Formatted with `oxfmt`.

## Residual gap, not addressed here

`agent.pendingInterrupts` is only ever corrected by an event that arrives. Nothing can ask the server what a thread is waiting on: `handle-connect.ts` attaches a live stream and carries no snapshot, and `packages/core/src/threads.ts` has no reference to interrupts. Separately, `run` mode joins the channel with no cursor while only `connect` mode sends `last_seen_event_id`, so a graph that interrupts before the browser finishes joining has no client-side way to request a replay. That needs its own issue if these fixes prove insufficient.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved interrupt recovery when runs finish without finalization events or events are replayed.
  * Restored pending interrupts after reconnecting, remounting, or missed events.
  * Resumed interruptions against the originating run rather than the connection’s run.
  * Scoped interrupt prompts and cleanup to the correct thread and agent, preventing stale prompts or responses after switching.
  * Ensured legacy interrupt prompts clear after new runs, dismissal, resolution, or expiration.
* **Tests**
  * Added recovery and isolation coverage across Angular, React, and Vue integrations.

### DIFF
--- a/packages/core/src/interrupt-state.ts
+++ b/packages/core/src/interrupt-state.ts
@@ -50,6 +50,61 @@ function toolResultContent(response: ResumeResponse): string {
 }
 
 /**
+ * @internal A legacy interrupt a thread is still waiting on, plus the run that
+ * raised it. Application authors must not depend on this contract.
+ */
+export interface ɵLegacyInterruptRecord<TValue = unknown> {
+  event: ɵInterruptEvent<TValue>;
+  runId?: string;
+}
+
+/**
+ * Legacy `on_interrupt` interrupts a thread is still waiting on, keyed by the
+ * agent that received them.
+ *
+ * `AbstractAgent.pendingInterrupts` records standard AG-UI interrupts, so a
+ * framework adapter can recover a standard gate after a remount or a
+ * reconnect. Legacy custom-event interrupts have no such record upstream,
+ * which left the path every CLI starter takes with no way to recover a gate
+ * whose event was already delivered. This map is that record.
+ */
+const legacyInterrupts = new WeakMap<object, ɵLegacyInterruptRecord>();
+
+/**
+ * @internal Record the legacy interrupt an agent is waiting on. Application
+ * authors must not depend on this API.
+ */
+export function ɵrecordLegacyInterrupt<TValue>(
+  agent: object,
+  record: ɵLegacyInterruptRecord<TValue>,
+): void {
+  legacyInterrupts.set(agent, record as ɵLegacyInterruptRecord);
+}
+
+/**
+ * @internal Read the legacy interrupt an agent is waiting on, if any.
+ * Application authors must not depend on this API.
+ */
+export function ɵreadLegacyInterrupt<TValue = unknown>(
+  agent: object,
+): ɵLegacyInterruptRecord<TValue> | null {
+  return (
+    (legacyInterrupts.get(agent) as
+      | ɵLegacyInterruptRecord<TValue>
+      | undefined) ?? null
+  );
+}
+
+/**
+ * @internal Forget the legacy interrupt an agent was waiting on. Call this
+ * when a new run starts, and when the interrupt is addressed or dismissed.
+ * Application authors must not depend on this API.
+ */
+export function ɵclearLegacyInterrupt(agent: object): void {
+  legacyInterrupts.delete(agent);
+}
+
+/**
  * @internal Framework-neutral interrupt response state shared by framework
  * adapters. Application authors must not depend on this API.
  */
--- a/packages/core/src/interrupt-state.ts
+++ b/packages/core/src/interrupt-state.ts
@@ -50,11 +50,17 @@ function toolResultContent(response: ResumeResponse): string {
 }
 
 /**
- * @internal A legacy interrupt a thread is still waiting on, plus the run that
- * raised it. Application authors must not depend on this contract.
+ * @internal A legacy interrupt a thread is still waiting on, plus the thread
+ * and the run that raised it. Application authors must not depend on this
+ * contract.
  */
 export interface ɵLegacyInterruptRecord<TValue = unknown> {
   event: ɵInterruptEvent<TValue>;
+  /**
+   * The conversation that raised the interrupt. One agent instance serves
+   * every conversation an app opens, so the record has to name its own.
+   */
+  threadId: string;
   runId?: string;
 }
 
@@ -82,17 +88,24 @@ export function ɵrecordLegacyInterrupt<TValue>(
 }
 
 /**
- * @internal Read the legacy interrupt an agent is waiting on, if any.
+ * @internal Read the legacy interrupt an agent is waiting on in one thread.
+ *
+ * A record raised in another thread stays stored but does not surface here.
+ * The agent survives a thread switch, so returning that record would show one
+ * conversation's approval prompt inside another, and answering it would resume
+ * the wrong conversation.
+ *
  * Application authors must not depend on this API.
  */
 export function ɵreadLegacyInterrupt<TValue = unknown>(
   agent: object,
+  threadId: string,
 ): ɵLegacyInterruptRecord<TValue> | null {
-  return (
-    (legacyInterrupts.get(agent) as
-      | ɵLegacyInterruptRecord<TValue>
-      | undefined) ?? null
-  );
+  const record = legacyInterrupts.get(agent) as
+    | ɵLegacyInterruptRecord<TValue>
+    | undefined;
+  if (!record) return null;
+  return record.threadId === threadId ? record : null;
 }
 
 /**
--- a/packages/react-core/src/v2/hooks/use-interrupt.tsx
+++ b/packages/react-core/src/v2/hooks/use-interrupt.tsx
@@ -7,7 +7,12 @@ import React, {
 } from "react";
 import { randomUUID } from "@ag-ui/client";
 import type { Interrupt, Message } from "@ag-ui/client";
-import { ɵInterruptState } from "@copilotkit/core";
+import {
+  ɵclearLegacyInterrupt,
+  ɵInterruptState,
+  ɵreadLegacyInterrupt,
+  ɵrecordLegacyInterrupt,
+} from "@copilotkit/core";
 import type { ɵPendingInterrupt } from "@copilotkit/core";
 import { useCopilotKit } from "../context";
 import { useAgent } from "./use-agent";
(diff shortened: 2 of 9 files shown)


packages/core/src/interrupt-state.ts
```diff
@@ -49,6 +49,61 @@ function toolResultContent(response: ResumeResponse): string {
...
+ * Legacy `on_interrupt` interrupts a thread is still waiting on, keyed by the
+ * agent that received them.
+ *
+ * `AbstractAgent.pendingInterrupts` records standard AG-UI interrupts, so a
+ * framework adapter can recover a standard gate after a remount or a
+ * reconnect. Legacy custom-event interrupts have no such record upstream,
+ * which left the path every CLI starter takes with no way to recover a gate
+ * whose event was already delivered. This map is that record.
+ */
+const legacyInterrupts = new WeakMap<object, ɵLegacyInterruptRecord>();
```
tylerslaton, line 77: This cache remembers the agent but not the conversation. If conversation A pauses for approval, then the same agent switches to conversation B and the UI is mounted again, A's approval prompt appears in B. Answering it submits A's response to B. Please save the original conversation ID with the interrupt and only restore it when that ID matches. A test covering A pauses → switch to B → remount → no A prompt would catch this.

packages/react-core/src/v2/hooks/use-interrupt.tsx
```diff
@@ -203,41 +241,43 @@ export function useInterrupt<
...
+      setPending(interruptState.pending);
+    }
+
     return () => {
       subscription.unsubscribe();
-      interruptState.clear();
+      // Keep the accumulated state when the agent still records a gate, so a
+      // remount can pick it back up. Discarding it here is what made an
+      // interrupted thread unrecoverable.
+      if (
```
tylerslaton, line 306: This cleanup also runs when the hook switches to a different agent. If the old agent still has an interrupt, its local approval state is kept. When the new agent has no interrupt, nothing replaces that state, so the old prompt remains actionable and sends its answer to the new agent. I reproduced this; the previous code did not submit the response. Please clear the hook's local interrupt state when the agent changes, while keeping any saved recovery record with the original agent.

packages/react-core/src/v2/hooks/use-interrupt.tsx
```diff
@@ -203,41 +241,43 @@ export function useInterrupt<
           }
           localStandard = params.interrupts;
         }
+        // Commit here rather than only at `onRunFinalized`. On the connect
+        // path `onRunFinalized` fires when the long-lived socket stream tears
+        // down, not once per replayed run, so a reconnect that replays an
+        // interrupt would otherwise surface nothing.
+        commit(params.input.runId);
```
tylerslaton: During reconnect, params.input.runId identifies the connection request and can differ from the run that paused for approval. For example, a connection with ID "connection-123" can replay an interrupt from "original-run". This saves "connection-123", so clicking Approve submits the response under that unrelated ID. Please use params.event.runId here and keep the actual run ID for the finalization fallback too. Add a replay test where the connection ID differs from the interrupted run ID; the current tests use the same identity and miss this.

tylerslaton (changes requested):
I found three issues that need fixing before approval: an old approval can be sent to a different conversation or agent, and reconnecting can send the response under the wrong run ID. Focused hook checks reproduced all three incorrect submissions; comparison with the previous code confirmed the first two are regressions. CI is green, but I did not rerun the full package suites.

packages/core/src/interrupt-state.ts
BenTaylorDev: Fixed in 245acbd. `ɵLegacyInterruptRecord` now carries its own `threadId`, and `ɵreadLegacyInterrupt(agent, threadId)` returns `null` when the caller's thread does not match. A mismatch does not drop the record, so returning to thread A still recovers the gate.

Tests follow the walk you described — core: `hides a record raised in a different thread`, `still recovers the record when the original thread comes back`; react-core and vue: `does not surface thread A's gate after the app switches to thread B`, `recovers thread A's gate when the app returns to thread A`; angular has the same pair.

One case your walk did not cover, which is also fixed: the thread can change on a *mounted* hook, because `useAgent` writes `threadId` onto the agent instance in its own effect, so there is no remount to re-seed from. React and Vue now drop the gate at submit time when the thread moved (`does not submit thread A's answer after a thread switch without a remount`). Angular already guarded that.

packages/react-core/src/v2/hooks/use-interrupt.tsx
BenTaylorDev: Fixed in 245acbd. Cleanup still keeps state when the agent holds a gate, but the seed path now has an explicit else branch: when the new agent waits on nothing it clears the hook-local state and sets `pending` to null. The previous agent keeps its own recovery record, so switching back still recovers it.

Tests: react-core `drops the old agent's gate when the same hook switches to an agent with no gate`, vue `drops the old agent's gate when the watcher switches to an agent with no gate`. Both assert `runAgent` was never called, not only that the prompt disappeared.

packages/react-core/src/v2/hooks/use-interrupt.tsx
BenTaylorDev: Fixed in 245acbd. Both the standard and the legacy paths now read `params.event.runId`. `onRunFinalized` carries no event, so it falls back to the last run id `RUN_FINISHED` named and only then to `params.input.runId`.

The replay tests use the ids from your example — `input.runId: "connection-123"` against `event.runId: "original-run"` — and assert `runAgent` received `original-run`: react-core `resumes the replayed run, not the connection that replayed it` plus its legacy variant, and the same pair in vue and angular.

PR #6895, opened 2026-09-04
feat(web-inspector): the launcher's two overlays become one island

lukasmoschitz:
## What this changes

The launcher had two things overlaying it that were not related to each other. The **error pill** was a capsule at launcher height, `border-radius: 999px`, painted from the launcher's own face and edge tokens, revealed by a rounded `clip-path`. The **services HUD** was a 228px rectangle at the panel radius, with a dotted border at twice the edge token's opacity, a `backdrop-filter` blur that an earlier PR had deliberately removed from the launcher, a drop shadow, and an arrow pointing back at the mark.

They shared no token. The arrow was the symptom: the pill needs none because it grows out of the launcher, while the HUD floated beside it and had to point home.

Now there is one component. A **capsule** in front, a **drawer** behind it sharing its top edge, both clipped to the mark's own circle when closed and growing out of it on hover. The capsule is the shape that already ships as the error message — same height, radius, face, edge, two text lines. Only the words change.

It carries the Intelligence state as the island's title (`Intelligence connected` / `Intelligence not connected`), with `Threads enabled` / `Learning disabled` as rows beneath it, each with a check or a cross. Intelligence sat among those rows before, which read as a peer of the features that actually depend on it.

**The state logic is untouched.** `home-briefing.ts` and `inspector-metadata.ts` are not in the diff. The island reads the same `getHomeModel()` projection the Home view reads, so the two cannot disagree.

## Bugs on `main` that this fixes

**The HUD's rows overlapped the mark by 23px**, and the mark's stacking swallowed the click — a row that looked like a link was not one. That is not patched: the drawer now shares the capsule's top edge and reserves the mark's band as padding, so the overlap cannot recur.

**Both of the launcher's rings depended on the host page.** The resting hairline came from Tailwind's `border` utility and the focus ring from `focus-visible:outline`; both resolve their style from a registered custom property, and `@property` inside a stylesheet adopted into a shadow root does not register document-wide. Measured on two hosts, same browser, identical bytes:

| host | `--tw-border-style` | `border-style` | `border-width` |
|---|---|---|---|
| plain page | *(empty)* | `none` | **0px** |
| page with Tailwind v4 | `solid` | `solid` | 1px |

The resting ring exists to separate the launcher from a **dark** page — the face is `#181C1F`, measured at 1.10:1 against GitHub dark — so on precisely the hosts it is for, the inherited version was the one that vanished. Both rings now declare their own values.

**The island only checked horizontal room.** Dragged near the foot of the window it opened on the correct side and then ran off the bottom (measured at 1280×720: drawer bottom 784). One resolver now answers both axes.

**The two halves could open on opposite sides of the mark.** The capsule used the overhang; the drawer still used a stale 248px constant and the old beside-the-launcher geometry. For `mark.left` in [225.8, 264) they pointed away from each other.

**The circle read darker than the pill.** The capsule spans the island and passes under the mark, and both faces were `rgba(24, 28, 31, 0.95)` — 0.9975 in the overlap, channel 24.6 under the mark against 35.6 beside it on white. One opaque token now, shared by button, capsule and drawer.

## Behaviour under a signal

`isLauncherHudBlocked()` returned `gestureSignal !== null`, so hovering during an error gesture opened nothing at all. And `beginGestureTail` ended in `closeLauncherHud()`, so an error arriving mid-hover slammed the drawer shut under the pointer — which also fails WCAG 2.1 SC 1.4.13.

The two slots are now arbitrated separately:

| | capsule | drawer | beat |
|---|---|---|---|
| nothing armed, no dwell | — | — | — |
| signal fires, no dwell | its words | — | yes, self-closing |
| dwell, nothing armed | Intelligence state | services | — |
| dwell begins during a gesture | the signal's words | services | clock stops, never resumes |
| signal fires during dwell | swaps in | unchanged | **no** |
| signal heals during dwell | back to the summary | unchanged | no |
| dwell ends after a signal showed | — | — | none |
| two signals armed | higher `priority` | — | existing |

The gesture is parked rather than cancelled: the live region keeps its sentence and the held telemetry impression is still released. One case fell outside the table — a signal with no words to show cannot count as shown, so its beat is deferred under the pointer rather than spent.

Click destinations keep their precedence, with one case filled in: a capsule reading `Intelligence not connected` lands on Home, where Intelligence is set up. Connected, it stays neutral and restores the reader's view.

## Testing

**644 → 664, 33 files, green.** Every new guard was mutation-proven: the regression it names was introduced deliberately, the test confirmed red, then reverted.

Three guards close classes rather than instances and are worth a look:

- **A bidirectional reachability join** — every `.cpk-launcher-*` selector and `data-cpk-*` attribute in the sheet must reach a rendered element, and every rendered one must have a rule. It exists because a commit on this branch renamed the stylesheet while the markup still emitted the old names: every drawer rule matched nothing, the launcher rendered unstyled, and 611 tests passed.
- **A z-order guard that asserts presence**, not values. The bug was not a wrong number, it was two missing ones: the drawer declared `z-index: 1` while the capsule and button declared none.
- **The closed clip**, asserted through `--cpk-launcher-size` at all four corners, so "the circle becomes the island" is checked rather than claimed.

## Two traps for whoever reviews this

`index.ts` is a Tailwind `@source` and the glob covers `__tests__/`, so a bare utility word in a **code comment** emits a real rule into the committed `generated.css`. It bit three times here. A clean `git status` is not proof — a word only produces a diff when that utility is not already in the sheet from real usage.

The pre-commit hooks do not typecheck the spec files. A `tsc` error in a test can land green; one did, and is fixed here.

## Not verified, and not in scope

The **visual result has not been checked in a browser by CI or by the author's tooling** — it was reviewed by hand. Worth a look on a dark host page in particular, since that is where the ring argument lives.

Out of scope deliberately: **touch** (the island opens on `pointerenter`; a tap opens the Inspector — pre-existing), a **signal-supplied drawer payload** (listing failed runs under a `run` error rather than the services; the registry slot is designed and unbuilt), and the **word cross-fade** when a signal swaps into a dwelling capsule — the behavioural half is implemented and tested, the swap itself is a hard cut. The dot still turns red, so the change is not unsignalled.

One known gap introduced here: the hover rule was deleted rather than emptied, so a launcher with **no room to open** now gives no hover feedback at all, where it used to tint. Restoring it belongs on the whole island, not on the circle alone.

### DIFF
--- a/packages/web-inspector/src/__tests__/launcher-error-signal.spec.ts
+++ b/packages/web-inspector/src/__tests__/launcher-error-signal.spec.ts
@@ -78,8 +78,9 @@ const EDGE_MARGIN = 16;
  * The capsule is pinned to `--cpk-launcher-island` rather than sized by its
  * content, so this constant mirrors that token's value rather than a
  * measured text width. It no longer depends on the label, the padding, or
- * the launcher size — it is the same 272 at every point of the size clamp,
- * which is what the two tests either side of `TIGHTEST_FIT` pin down.
+ * the launcher size. The threshold at which neither side has room moves
+ * with the token, which is what the two tests either side of
+ * `TIGHTEST_FIT` pin down.
  */
 const ISLAND_WIDTH = 272;
 
@@ -1660,11 +1661,11 @@ test("with no room the failure is still spoken", async () => {
 });
 
 // The narrowest window a rightward pill fits in: the launcher's own left
-// offset, the mark, its overhang, and the margin the pill keeps from the edge.
-// Widening the pill — as the padding and the second line did — moves this,
-// which is the whole reason ISLAND_WIDTH is a named number.
-const TIGHTEST_FIT =
-  EDGE_MARGIN + LAUNCHER_SIZE + (ISLAND_WIDTH - LAUNCHER_SIZE) + EDGE_MARGIN;
+// offset, the island's own width, and the margin the pill keeps from the
+// edge on the other side. LAUNCHER_SIZE cancels out now that the width is
+// a token rather than content-derived, which is the whole reason
+// ISLAND_WIDTH is a named number.
+const TIGHTEST_FIT = EDGE_MARGIN + ISLAND_WIDTH + EDGE_MARGIN;
 
 test("a window exactly wide enough still opens the pill", async () => {
   const context = await setup();
--- a/packages/web-inspector/src/__tests__/launcher-error-signal.spec.ts
+++ b/packages/web-inspector/src/__tests__/launcher-error-signal.spec.ts
@@ -372,12 +372,14 @@ const pulsing = (inspector: WebInspectorElement): boolean =>
 
 /** The pill on the closed launcher, or null when the launcher is not talking. */
 function pill(inspector: WebInspectorElement): HTMLElement | null {
-  return root(inspector).querySelector<HTMLElement>("[data-cpk-launcher-pill]");
+  return root(inspector).querySelector<HTMLElement>(
+    "[data-cpk-launcher-capsule]",
+  );
 }
 
 /** Which subject the pill names, or null when there is no pill. */
 function pillSubject(inspector: WebInspectorElement): string | null {
-  return pill(inspector)?.getAttribute("data-cpk-launcher-pill") ?? null;
+  return pill(inspector)?.getAttribute("data-cpk-launcher-capsule") ?? null;
 }
 
 /** The pill's first line — the failure class — or null when there is no pill. */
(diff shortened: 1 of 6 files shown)


lukasmoschitz:
Closing this. main has moved 28 commits on index.ts since I branched, and two of them reworked the launcher HUD itself, so the merge is 33 conflict hunks over a file whose target behaviour changed underneath. Redoing the five bugs as separate small PRs against current main is the cheaper path.

One thing worth carrying over: the reachability join that catches stylesheet selectors matching no element. A rename on this branch left every drawer rule dead, the launcher rendered unstyled, and 611 tests stayed green. That guard goes into the first follow-up.

PR #5996, opened 2026-07-16
fix(react-core): preserve code block newlines

zcxGGmu:
## Summary
- add cpk:whitespace-pre to Streamdown code-block pre elements
- add a source-string regression test for the scoped code-block pre rule

## Verification
- pnpm nx run @copilotkit/shared:build
- pnpm --filter @copilotkit/react-core exec vitest run src/v2/styles/__tests__/streamdown-styles.test.ts
- pnpm --filter @copilotkit/react-core run build:css

PR #5944, opened 2026-07-13
fix(react-core): scope streamdown table action controls in packaged CSS (#5775)

BenTaylorDev:
Fixes #5775.

## Problem

After #5099 scoped the streamdown markdown/table styles under `[data-copilotkit] [data-streamdown="…"]`, the **table action controls** (copy / download) are still unstyled for hosts that import `@copilotkit/react-core/v2/styles.css` but don't also ship streamdown's raw Tailwind utilities. streamdown renders the controls row, per-button wrappers, trigger buttons, dropdown popovers and menu items with unprefixed utilities (`flex`, `items-center`, `justify-end`, `gap-1`, `cursor-pointer`, `p-1`, …) and **no stable `data-streamdown` attribute**, so CopilotKit's packaged CSS didn't cover them — the controls rendered as vertically stacked plain icons instead of a right-aligned row.

## Fix

Add scoped fallback selectors under `[data-copilotkit] [data-streamdown="table-wrapper"]`, targeting the controls chrome **structurally** (since it has no `data-streamdown` hook):

- controls row → `> div:first-child:not(:last-child)` (flex, right-aligned, gap)
- per-button wrapper → `… > div` (relative, positions the popover)
- trigger buttons → `… > div > button` (matches the code-block copy/download button styling)
- dropdown popover → `… > div > div`
- popover menu items → `… > div > div > button`

The controls row is `table-wrapper`'s first child **only when controls are enabled**; `:not(:last-child)` leaves a control-less table (whose single child is the scroll container, already styled by #5099) untouched.

## Verification

- **Compiles.** Built `globals.css` through the Tailwind v4 CLI — every `@apply` resolves (e.g. `bg-background` → `var(--background)`, `shadow-lg` → the shadow vars, `min-w-[120px]` → `min-width:120px`) and all five rules emit with correct values.
- **Selectors match the real DOM.** A new DOM test (`streamdown-table-controls.test.tsx`) renders a real `<Streamdown>` table and asserts the controls row is `table-wrapper`'s first-non-only child, carries no `data-streamdown` attribute, and contains the trigger buttons under `> div > button` — i.e. the scoped selectors target real elements. This also guards against streamdown markup drift.
- **Selector presence** guarded by `streamdown-styles.test.ts` (whitespace-robust).
- Full `styles/__tests__` suite green; `oxlint`/`oxfmt` clean.

## Note (out of scope, discovered while fixing)

[...]

PR #3615, opened 2026-04-03
fix: preserve whitespace in code blocks for CopilotKit v2

serhiizghama:
## Problem

Closes #3330

Code blocks rendered inside `CopilotChatAssistantMessage` (v2) lose their newlines. The root cause is that the Tailwind Typography (`cpk:prose`) styles override the default `white-space` behavior on `<pre>` elements. The existing CSS rule in `packages/react-core/src/v2/styles/globals.css` targets the correct selector (`div[data-streamdown="code-block"] > pre`) but only resets margins, leaving `white-space` unset.

## What changed

Added `cpk:whitespace-pre` to the `[data-copilotkit] div[data-streamdown="code-block"] > pre` rule in `packages/react-core/src/v2/styles/globals.css`.

```diff
 [data-copilotkit] div[data-streamdown="code-block"] > pre {
-  @apply cpk:mt-0 cpk:mb-0;
+  @apply cpk:mt-0 cpk:mb-0 cpk:whitespace-pre;
 }
```

## How to test

1. Run the v2 Storybook: `pnpm storybook:react`
2. Open the `CopilotChatAssistantMessage` story that tests all markdown features
3. Verify that code blocks now render with preserved newlines

PR #6902, opened 2026-09-04
fix(react-core): restore code block line breaks in packaged CSS (#3330)

### LINKED ISSUE #3330: 🐛 Bug: Markdown code block rendering doesn't preserve newlines

BenTaylorDev:
## What does this PR do?

Fixes #3330 — fenced markdown code blocks render as one collapsed line in the packaged v2 React UI.

### Root cause

streamdown renders one `<span>` per source line inside `pre[data-streamdown="code-block-body"] > code`, and leaves **no newline characters** in the text. The line break comes entirely from the raw Tailwind utility `block` on that span:

```js
// streamdown 1.6.11, dist/code-block-*.js
var v = cn("block", "before:content-[counter(line)]", ...);
```

CopilotKit builds Tailwind with `@import "tailwindcss" prefix(cpk)`, so `.block` is never emitted into `dist/v2/index.css` — only `.cpk\:block` is. Every line therefore renders inline and the block collapses onto one row.

The line spans carry no `data-streamdown` attribute, so the rule has to be scoped structurally, the same way the table action controls were in #5944:

```css
[data-copilotkit] [data-streamdown="code-block-body"] > code > span {
  @apply cpk:block;
}
```

### Why the earlier attempts did not work

Three previous PRs (#3441, #3615, #5387) added `whitespace-pre` to the `<pre>`. That is a no-op: the UA stylesheet already applies `white-space: pre` to `<pre>`, nothing in the packaged CSS overrides it, and there are no newlines in the text for it to preserve.

### Knowingly not fixed here

- **Line-number gutter.** streamdown's `before:content-[counter(line)] before:w-4 before:mr-4 …` utilities are unprefixed too, so the gutter never renders. That is cosmetic, and the repo's existing scoped rules do not port it either.
- **The pre-highlight loading skeleton** (`space-y-4`, `divide-y`, `animate-spin`) is unprefixed as well — a brief flash of unstyled skeleton before shiki resolves.
- **The broader class of bug.** Every unprefixed streamdown utility has to be hand-ported like this. streamdown 2.x adds a `prefix` prop that would fix the whole surface at once, and #5147 proposes removing the bundled renderer entirely. Both are larger calls than this bug fix.

## Testing

**1. Live browser verification.** Built `dist/v2/index.css` from `origin/main` and from this branch, rendered streamdown 1.6.11's actual code-block DOM against each, and measured layout in Chromium:

| | `white-space` on `<pre>` | line-span `display` | distinct rendered rows | `<pre>` height |
|---|---|---|---|---|
| main | `pre` | `inline` | **1** | 52px |
| this PR | `pre` | `block` | **5** | 112px |

Indentation is preserved after the fix (`spans[1].textContent` starts with two spaces).

**2. Compiled CSS.** `tailwindcss -i src/v2/styles/globals.css -o … -m` emits exactly:

```css
[data-copilotkit] [data-streamdown=code-block-body]>code>span{display:block}
```

**3. Tests** — `pnpm -C packages/react-core exec vitest run src/v2/styles`

```
 ✓ src/v2/styles/__tests__/streamdown-styles.test.ts (3 tests) 2ms
 ✓ src/v2/styles/__tests__/streamdown-table-controls.test.tsx (1 test) 37ms
 ✓ src/v2/styles/__tests__/streamdown-code-block-lines.test.tsx (1 test) 430ms

 Test Files  3 passed (3)
      Tests  5 passed (5)
```

Two tests, following the split established by #5944 — a source-string test that the selector exists, and a DOM test that streamdown still renders the structure that selector assumes (so a streamdown markup change fails loudly instead of silently un-fixing this).

**4. Mutation-checked both tests.** Removing the CSS rule fails the string test:

```
   × Streamdown styles > ships a scoped display rule for code block lines (#3330) 3ms
      Tests  1 failed | 2 passed (3)
```

Pointing the DOM test at a selector streamdown does not render fails it:

```
   × Streamdown code block lines DOM (#3330) > renders one line span per source line 428ms
      Tests  1 failed (1)
```

**5. Formatting** — `oxfmt --check` clean on all three files; `git diff --check` clean.

`tsc --noEmit` in this worktree reports 58 pre-existing errors, all from a stale cross-package `@copilotkit/core` dist; none are in the changed files (which are CSS plus tests).

## Related PRs and Issues

Fixes #3330
Supersedes #3441, #3615, #5387, #5996 (all added a no-op `whitespace-pre`)

## Checklist

- [x] If the PR changes or adds functionality, I have updated the relevant documentation (not applicable: scoped visual bug fix)

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed fenced code blocks collapsing into a single line in the packaged UI.
  - Code lines now render vertically as separate rows with the correct layout styling.

- **Tests**
  - Added regression coverage to verify code-line rendering and scoped styles for code blocks.

### DIFF
--- a/packages/react-core/src/v2/styles/__tests__/streamdown-code-block-lines.test.tsx
+++ b/packages/react-core/src/v2/styles/__tests__/streamdown-code-block-lines.test.tsx
@@ -1,0 +1,72 @@
+/**
+ * #3330: fenced code blocks collapsed onto a single line in the packaged UI.
+ *
+ * streamdown renders one <span> per source line inside
+ * `pre[data-streamdown="code-block-body"] > code`, and the line break comes
+ * entirely from the raw Tailwind utility `block` on that span — the <pre> text
+ * itself holds no newline characters. CopilotKit builds Tailwind with
+ * `prefix(cpk)`, so `.block` never reaches the packaged CSS and every line ran
+ * inline. globals.css therefore scopes the display rule structurally, because
+ * the line spans carry no `data-streamdown` attribute:
+ * `[data-streamdown="code-block-body"] > code > span { @apply cpk:block }`.
+ *
+ * A source-string test (streamdown-styles.test.ts) guards that the selector
+ * exists. This DOM test guards the OTHER half: that streamdown still renders the
+ * structure that selector assumes. If streamdown changes its code-block markup
+ * (adds a data-streamdown attribute to the lines, renests them, or starts
+ * emitting real newlines), this fails — signalling the scoped CSS must be
+ * revisited.
+ */
+import { describe, it, expect } from "vitest";
+import { render, waitFor } from "@testing-library/react";
+import { Streamdown } from "streamdown";
+
+const CODE_LINES = [
+  "function greet(name) {",
+  "  const msg = `hi ${name}`;",
+  "  return msg;",
+  "}",
+];
+const CODE_MARKDOWN = ["```js", ...CODE_LINES, "```", ""].join("\n");
+
+describe("Streamdown code block lines DOM (#3330)", () => {
+  it("renders one line span per source line, matching the scoped selector", async () => {
+    const { container } = render(
+      <div data-copilotkit="">
+        <Streamdown>{CODE_MARKDOWN}</Streamdown>
+      </div>,
+    );
+
+    // Highlighting is async: the first paint is a loading skeleton.
+    await waitFor(
+      () => {
+        expect(
+          container.querySelector('pre[data-streamdown="code-block-body"]'),
+        ).not.toBeNull();
+      },
+      { timeout: 15000 },
+    );
+
+    const body = container.querySelector<HTMLElement>(
+      'pre[data-streamdown="code-block-body"]',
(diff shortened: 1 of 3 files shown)


MikeRyanDev (approved):

Issue #3388, opened 2026-03-06
🚀 Feature Request: Programmatically Add Custom Cards to Chat History from the Frontend (Without Tool Calls)


IanDoesTech:
### Pre-flight Checklist

- [x] I have searched existing issues, and this hasn't been requested yet.

### Problem or Motivation

Currently, the recommended way to render a UI component or card inside the CopilotKit chat history appears to be via a tool call message (e.g. `useComponent`).

However, this tightly couples UI rendering to agent tool messages, which makes it difficult for UI developers to programmatically add cards to the chat history based on frontend events or application state. Also, it often unecessarily pollutes the messages.

For example, a UI might want to respond to events received through `agent.subscribe`, or something completely UI related, and display a structured card inside the chat transcript.

At the moment, there does not appear to be a clear way to insert such UI elements into the chat history purely from frontend logic, without requiring the agent to emit a tool call.

### 🎤 Tell us your idea

It would be useful to have a mechanism that allows frontend code to insert a card/component into the chat history without requiring a backend tool message.

This would allow UI developers to respond to application events (for example those received via `agent.subscribe`) and programmatically add cards to the chat.

These messages could potentially be ephemeral (session-only) rather than persisted in the conversation thread, which would be acceptable/preferable for many use cases.

Providing a supported way to add UI-driven messages to the chat history would make it much easier to integrate CopilotKit with event-driven application workflows.

### Would you like to work on this?

- [ ] Yes, I’d love to work on it!
- [x] I’m open to collaborating but need guidance.
- [ ] No, I’m just sharing the idea.

BenTaylorDev:
You can render custom cards in the transcript without a tool call, via the provider-level `renderCustomMessages` prop (v2) — register a renderer that draws your card and returns `null` for other messages. (On v1, `appendMessage` + `message.generativeUI` is the equivalent.)

One caveat worth flagging, and why I'm leaving this open: there's no *ephemeral* / client-only message today. `agent.addMessage(...)` appends to `agent.messages`, which is the thread sent to the backend on the next run — so a card added that way does become part of history, which is exactly the "don't pollute the conversation" concern in your request. Keeping this open as the ask for a client-only/ephemeral message flag.

rodboev:
Closing this because current main now supports frontend-added `activity` messages in the chat transcript and excludes them from the next agent run. Merged PR https://github.com/CopilotKit/CopilotKit/pull/6904 documents and tests that path.

PR #6904, opened 2026-09-04
docs(showcase): document frontend-driven activity cards (refs #3388)

BenTaylorDev:
## What

Issue #3388 asked for a way to put a card into the chat transcript from frontend code, without a tool call and without adding to the conversation the model reads.

**That already ships.** A message with `role: "activity"` renders standalone in the transcript, and `AbstractAgent.prepareRunAgentInput` strips every activity message from the run payload:

```js
prepareRunAgentInput(e) {
  let t = structuredClone_(this.messages).filter(e => e.role !== `activity`);
  ...
}
```

The gap was documentation. `renderActivityMessages` is only documented for **backend-emitted** activities (mastra background-tasks, a2a, mcp-apps), so the frontend-driven path was undiscoverable.

This PR adds the missing guide page and a test that pins the behavior.

## Changes

| File | Why |
| --- | --- |
| `showcase/shell-docs/.../generative-ui/frontend-cards.mdx` | New "Frontend-Driven Cards" guide |
| `showcase/shell-docs/.../generative-ui/meta.json` | Sidebar entry (6-line insertion) |
| `packages/react-core/.../CopilotChatFrontendActivityCard.e2e.test.tsx` | Pins both halves of the contract |

No source changes. Behavior is unchanged; this documents and locks what already works.

## The non-obvious part

The card must be added via the agent returned by `useAgent()`. An agent instance constructed and held outside React is **not** the instance the chat renders, so messages added to it silently never appear. This cost me a debugging round while verifying, and it is called out as a warning callout in the docs.

## Testing

**1. New test passes against clean `origin/main`** (run in a worktree at `96cf7aa55f`, with `@copilotkit/shared` and `@copilotkit/core` rebuilt from the worktree so the test is not reading a stale dist):

```
✓ src/v2/components/chat/__tests__/CopilotChatFrontendActivityCard.e2e.test.tsx (2 tests) 72ms

 Test Files  1 passed (1)
      Tests  2 passed (2)
```

**2. Mutation-checked, so neither assertion is self-fulfilling.**

Drop the renderer registration → the render test fails:
```
× renders a card added from frontend code, with no tool call 1068ms
      Tests  1 failed | 1 passed (2)
```

Swap the card from `role: "activity"` to `role: "assistant"` → it reappears in the payload, so the exclusion is real and specific to `activity`:
```
AssertionError: expected [ 'user', 'assistant' ] to deeply equal [ 'user' ]
```

**3. Neighboring test unaffected on the same base:**

```
✓ src/v2/components/chat/__tests__/CopilotChatMessageView.test.tsx (16 tests) 53ms
      Tests  16 passed (16)
```

**4. Independent probe of the filter** against the pinned `@ag-ui/client` 0.0.57:

```
agent.messages roles: [ 'user', 'activity' ]
run input roles     : [ 'user' ]
```

**5. `tsc --noEmit`** — zero errors in the new file. Remaining errors in this workspace are in files this PR does not touch (`MCPAppsActivityRenderer.tsx`, `CopilotKitInspector.tsx`) and are artifacts of a hand-assembled local `node_modules`; CI has the real install.

**6. `oxfmt --check`** — clean.

**7. Docs checks** — `meta.json` validated as JSON; internal link uses the house `/generative-ui/...` form (no `/docs` prefix); `Callout type="warn"` matches the dominant existing usage; import paths verified against the real `@copilotkit/react-core/v2` barrel exports.

## Follow-up

Leaving #3388 open until this lands, then closing it with a pointer to the new page.

## Summary by CodeRabbit

- **New Features**
  - Added support for frontend-driven activity cards that render in chat transcripts without being sent to the agent or language model.
  - Added documentation covering activity card renderers, schemas, registration, payload filtering, snapshots, and limitations.
  - Added a new “Frontend-Driven” section to the Generative UI documentation navigation.

- **Tests**
  - Added end-to-end coverage for activity card rendering and payload exclusion.

### DIFF
--- a/packages/react-core/src/v2/components/chat/__tests__/CopilotChatFrontendActivityCard.e2e.test.tsx
+++ b/packages/react-core/src/v2/components/chat/__tests__/CopilotChatFrontendActivityCard.e2e.test.tsx
@@ -1,0 +1,137 @@
+/**
+ * Frontend-driven activity cards (issue #3388).
+ *
+ * A `role: "activity"` message added from frontend code renders standalone in
+ * the transcript, and `AbstractAgent.prepareRunAgentInput` strips it from the
+ * run payload — so the card never reaches the agent or the LLM. This is the
+ * supported way to put a card in the transcript without a tool call and
+ * without polluting the conversation.
+ */
+import React from "react";
+import { screen, waitFor, act } from "@testing-library/react";
+import { z } from "zod";
+import type { RunAgentInput } from "@ag-ui/core";
+import {
+  MockStepwiseAgent,
+  renderWithCopilotKit,
+} from "../../../__tests__/utils/test-helpers";
+import { CopilotChat } from "../CopilotChat";
+import { useAgent } from "../../../hooks/use-agent";
+import { CopilotKitCoreReact } from "../../../lib/react-core";
+import type { ReactActivityMessageRenderer } from "../../../types";
+
+// Test shim: some environments lack setCredentials on CopilotKitCoreReact.
+if (!(CopilotKitCoreReact.prototype as any).setCredentials) {
+  (CopilotKitCoreReact.prototype as any).setCredentials = () => {};
+}
+
+const contentSchema = z.object({ title: z.string() });
+
+const cardRenderer: ReactActivityMessageRenderer<
+  z.infer<typeof contentSchema>
+> = {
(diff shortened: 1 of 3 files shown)


MikeRyanDev (approved):

PR #6905, opened 2026-09-04
feat(react-core): controlled open/onOpenChange props for CopilotSidebar and CopilotPopup

### LINKED ISSUE #3334: 🐛 Bug: CopilotSidebar v2 — No external open/close control

BenTaylorDev:
Closes #3334 (OSS-524).

## Problem

v1 `<CopilotSidebar>` exposed `open` and `onSetOpen`. Those props let a host open and close the chat from its own UI. v2 shipped only `defaultOpen`. The reporter wanted a button in their own nav bar to close the sidebar.

The reporter's stated root cause is now stale. `shouldCreateModalState` no longer exists. Since CPK-7152 the provider syncs both directions: `setAndSync` upward, and an effect downward. A host that wraps its layout in `<CopilotChatConfigurationProvider>` and calls `setModalOpen` therefore does drive the sidebar on current `main`. I verified that before writing any code.

Two things are genuinely missing. The first is the ergonomic API that v1 had. The second is documentation for the outer-provider pattern that already works.

Two earlier community attempts (#3729, #6418) were closed unmerged.

## What changed

`open` and `onOpenChange` on `<CopilotSidebar>` and `<CopilotPopup>`:

- `open` pins what the surface renders, from the first frame.
- `onOpenChange` reports every request to open or close: the toggle button, click-outside, Escape, and the drawer's mobile mutual-exclusion. It fires with or without `open`, so it also works as a plain notification on the uncontrolled path.
- `defaultOpen` is unchanged. If both are passed, `open` wins.

Two design choices are worth review.

**1. A context-overriding scope, not a fourth mode in the provider.** `ControlledModalOpenScope` replaces `isModalOpen` and `setModalOpen` for the subtree below the provider that owns the state. The resolution chain inside `CopilotChatConfigurationProvider` stays untouched: own state, parent sync, drawer mutual-exclusion, and the `ɵregisterModalCloser` stack. The scope's setter still calls the underlying one, so those side effects keep running. It also registers itself as the modal closer, so the drawer's mobile exclusion reaches the host instead of flipping state that nothing displays. The alternative was a controlled branch threaded through `resolvedIsModalOpen`, `setAndSync`, and the sync effect. That adds a fourth interacting mode to the code CPK-7152 just stabilized.

**2. The props reach the views by context, not as props.** `<CopilotSidebar>` hands its view to `<CopilotChat>` as a memoized `chatView` component. Adding `open` to that memo's deps mints a new element type per toggle, and React then remounts the whole chat subtree. That is the same class of bug #6173 fixed for popup resize. There is a regression test for it.

Scope note: I included `<CopilotPopup>` because it shares the mechanism and the same docs page. The issue named only the sidebar.

## Testing

**New suite, 15 tests** (`CopilotSidebar.controlledOpen.test.tsx`). It covers the controlled contract, the unchanged uncontrolled path, and the remount guard.

```
✓ src/v2/components/chat/__tests__/CopilotSidebar.controlledOpen.test.tsx (15 tests) 155ms
  Test Files  1 passed (1)
       Tests  15 passed (15)
```

**Mutation-checked.** I broke each mechanism to confirm that the tests really fail.

| Mutation | Result |
| --- | --- |
| Drop `ControlledModalOpenScope`, keep only the seeded default | 5 failed: both `onOpenChange` reports, both host-driven open/close cases, the popup report |
| Implement through the memoized override instead (add `open` to the `useMemo` deps) | 1 failed: the remount guard, `expected 4 to be 1`, one extra mount per flip |
| Drop the `open ?? defaultOpen` seeding | 1 failed: "stays put when the host stops controlling open" |

I also mutation-checked the pre-existing two-way sync before I started. That confirmed the outer-provider workaround really works on `main`, instead of only appearing to.

**Full `@copilotkit/react-core` suite.** No regressions.

```
Test Files  141 passed | 1 skipped (142)
     Tests  1604 passed | 2 skipped (1606)
EXIT=0
```

**Adjacent suites re-run explicitly**: sidebar position, sidebar and popup slots, popup resize-remount, drawer launcher, and the provider's own 43 tests.

```
Test Files  6 passed (6)
     Tests  117 passed (117)
```

**Typecheck.** `tsc --noEmit` in `packages/react-core` gave `exit=0` with no output. The tsconfig includes `src/**/*`, so the new test file is typechecked too.

**Format and lint.** `oxfmt --check packages/react-core/src/v2` reported "All matched files use the correct format." `oxlint` on the touched files reported 0 errors.

**Pre-commit hooks.** They ran for real on both commits.

```
NX   Successfully ran targets test, publint, attw for 2 projects and 20 tasks they depend on
✔️ test-and-check-packages (15.33 seconds)
```

## Docs

- `prebuilt-components/chat-controls.mdx` now leads with the controlled pair. Its example drives the sidebar from a nav button outside it, which is the shape #3334 asked about. The `useCopilotChatConfiguration` route stays, reframed as the option for callers who prefer not to lift the state.
- `reference/components/CopilotSidebar.mdx` and `CopilotPopup.mdx` gain `open` and `onOpenChange`. Both pages documented `defaultOpen` as `false`, but both surfaces mount open, so I corrected that. A new test per surface pins the real default.

## Not in this PR

- Vue and Angular parity for the same props.
- The `width` prop of `<CopilotSidebar>` still sits in the memo deps of the `chatView` override. A live-resized sidebar therefore remounts the chat subtree, the way the popup did before #6173. That is pre-existing and out of scope here.

## Summary by CodeRabbit

- **New Features**
  - Added controlled open-state support for chat popups and sidebars through `open` and `onOpenChange`.
  - Preserved uncontrolled usage with `defaultOpen`, while allowing externally managed visibility and toggle requests.
  - Improved coordination between modal and mobile drawer behavior.

- **Documentation**
  - Added usage guidance and reference details for controlled and uncontrolled open-state management.

- **Tests**
  - Added coverage for initial visibility, toggle callbacks, controlled updates, default behavior, and preserving the chat subtree.

### DIFF
--- a/packages/react-core/src/v2/components/chat/CopilotPopup.tsx
+++ b/packages/react-core/src/v2/components/chat/CopilotPopup.tsx
@@ -8,6 +8,7 @@ import type { CopilotChatViewProps } from "./CopilotChatView";
 import CopilotChatView from "./CopilotChatView";
 import type { CopilotPopupViewProps } from "./CopilotPopupView";
 import CopilotPopupView from "./CopilotPopupView";
+import { ModalOpenControlProvider } from "./modal-open-control";
 
 /**
  * Carries the popup shell props (header, toggle, width, height, …) to the
@@ -74,6 +75,19 @@ export type CopilotPopupProps = Omit<CopilotChatProps, "chatView"> & {
   header?: CopilotPopupViewProps["header"];
   toggleButton?: CopilotPopupViewProps["toggleButton"];
   defaultOpen?: boolean;
+  /**
+   * Controlled open state. When supplied, the host owns whether the popup is
+   * open: the popup renders this value and never changes it on its own. Pair it
+   * with `onOpenChange` to react to the toggle button and to click-outside.
+   * Omit it (and use `defaultOpen`) to let the popup manage its own state.
+   */
+  open?: boolean;
+  /**
+   * Called with the requested state whenever the popup asks to open or close
+   * (the toggle button, click-outside, or the thread drawer on mobile). In
+   * controlled mode nothing moves until the host updates `open`.
+   */
+  onOpenChange?: (open: boolean) => void;
   width?: CopilotPopupViewProps["width"];
   height?: CopilotPopupViewProps["height"];
   clickOutsideToClose?: CopilotPopupViewProps["clickOutsideToClose"];
@@ -83,6 +97,8 @@ export function CopilotPopup({
   header,
   toggleButton,
   defaultOpen,
+  open,
+  onOpenChange,
   width,
   height,
   clickOutsideToClose,
@@ -115,12 +131,14 @@ export function CopilotPopup({
     <>
       {!isPopupLicensed && <InlineFeatureWarning featureName="Popup" />}
       <PopupShellPropsContext.Provider value={shellProps}>
-        <CopilotChat
-          welcomeScreen={CopilotPopupView.WelcomeScreen}
-          {...chatProps}
-          isModalDefaultOpen={defaultOpen}
-          chatView={PopupViewOverrideWithStatics}
-        />
+        <ModalOpenControlProvider open={open} onOpenChange={onOpenChange}>
+          <CopilotChat
+            welcomeScreen={CopilotPopupView.WelcomeScreen}
+            {...chatProps}
+            isModalDefaultOpen={defaultOpen}
+            chatView={PopupViewOverrideWithStatics}
+          />
+        </ModalOpenControlProvider>
       </PopupShellPropsContext.Provider>
     </>
   );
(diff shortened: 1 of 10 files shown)


MikeRyanDev (approved):

PR #6907, opened 2026-09-04
feat(showcase): run the AWS Strands demos on the native interrupt and register the reasoning cells

ranst91:
## What this does

Makes the two AWS Strands showcase columns a truthful acceptance test for the
refreshed AG-UI Strands bridges: the interrupt demos now run on Strands' native
interrupt instead of a frontend-tool stand-in, the reasoning cells are registered
and backed by a reasoning-capable agent, and the fixtures that could never have
matched their shared probes are corrected.

### Interrupts, now native (both languages)

`gen-ui-interrupt` and `interrupt-headless` were quarantined as unsupported and
were driven by a frontend tool with an async handler. Both bridges implement the
AG-UI interrupt protocol on top of Strands' own interrupt primitive
([docs](https://strandsagents.com/docs/user-guide/concepts/interrupts/)), so the
demos now use it:

- New dedicated agents (`src/agents/interrupt_agent.py`,
  `src/agent/interrupt-agent.ts`) whose `schedule_meeting` tool pauses itself
  with `tool_context.interrupt(...)` / `context.interrupt(...)`.
- They are mounted on their own sub-path rather than added to the shared agent,
  because `hitl-in-chat` registers a FRONTEND tool of the same name and one tool
  name cannot be both client-executed and backend-pausing.
- `interrupt-headless` was rebuilt on `useInterrupt({ renderInChat: false })`,
  matching the reference integrations, replacing a hand-rolled custom-event
  subscriber that only ever matched LangGraph's legacy interrupt shape.
- Quarantine lifted, `interrupt_pattern` is now `native`, and the stale prose in
  `PARITY_NOTES.md` and `agents/agent.py` that claimed Strands has no interrupt
  primitive is corrected.

### Reasoning, now registered

`reasoning-default`, `reasoning-custom` and `tool-rendering-reasoning-chain` were
listed under two DEPRECATED registry ids, so the dashboard showed nothing at all
while the demo pages pointed at agent names the runtime never registered. They
now run against dedicated agents on the OpenAI Responses API with reasoning
summaries enabled; the shared agent stays on chat completions so tool arguments
keep streaming incrementally.

### Fixtures that could not match

- The interrupt fixtures keyed on phrases that appear in neither pill prompt, so
  no leg could ever match the shared probes. Rewritten on the reference shape,
  including the cancel leg.
- Several generic legs matched on turn number where the reference matches on
  whether the tool has already answered. That is what made the reasoning-chain
  demo loop (80 tool results, no run finish) and the async frontend-tool demo
  re-fire its tool 60 times.

### Shared probe

An interrupt turn ends at the pause: the meaningful output is the mounted picker
and, on a bridge that emits the tool call before pausing, no assistant prose ever
lands. The gen-ui-interrupt probe settled on assistant-text stability, so the
turn timed out before its own assertions ran. It now uses the harness's
documented `completeOnMount` opt-in for surface-shaped turns; run-finished and
new-bubble conjuncts still apply.

## Pins

| Package | Before | After |
|---|---|---|
| `ag_ui_strands` | 0.2.2 | 0.4.0 |
| `ag-ui-protocol` | 0.1.18 | 0.1.22 |
| `strands-agents[OpenAI]` | 1.18.0 | 1.54.0 |
| `strands-agents-tools` | 0.2.16 | 0.8.8 |
| `openai` (Python) | 1.109.1 | 2.54.0 |
| `@strands-agents/sdk` | 1.6.0 | 1.16.0 |
| `@ag-ui/aws-strands` | 0.2.3 | 0.3.0 |
| `@ag-ui/core`, `@ag-ui/client`, `@ag-ui/encoder` | 0.0.57 | 0.0.59 |
| `@copilotkit/*` | 1.68.2 | 1.68.2 (repo-wide canonical pin, untouched) |

`ag_ui_strands` 0.4.0 is not optional: the Python bridge's interrupt protocol first appears in the
0.3 line (0.2.5 has no interrupt code at all), and 0.4.0 carries the bridge fixes this branch was
verified against. The companion bumps are not optional either, because neither container starts
without them: the Python bridge requires `ag-ui-protocol` 0.1.22, and the TypeScript one requires
the AG-UI core packages at 0.0.59. `strands-agents` 1.54.0 is not optional either: the
Responses model that carries reasoning summaries first appears in 1.32, and on 1.32 the bridge
emitted no reasoning events at all, verified at the wire. `openai` 2.x is required by the
Responses model and is the ceiling `langchain-openai` allows.

## D6 state, before and after

Before, both columns: `gen-ui-interrupt` and `interrupt-headless` quarantined as
`not_supported`; `reasoning-default` and `reasoning-custom` absent (the manifest listed two
DEPRECATED registry ids instead, and the demo pages pointed at agent names the runtime never
registered); `tool-rendering-reasoning-chain` `not_supported` with no reason recorded;
`shared-state-streaming` `not_supported` with no reason recorded.

After, per-cell isolated D6 runs on this branch (one stack per cell, harness image
carrying the probe changes):

| Cell | strands | strands-typescript |
|---|---|---|
| gen-ui-interrupt | green | green |
| interrupt-headless | green | green |
| reasoning-default | green | green |
| reasoning-custom | green | green |
| tool-rendering-reasoning-chain | green | green |
| tool-rendering | not re-run after the fixture edit (see below) | green |
| frontend-tools-async | green | green |
| shared-state-streaming | unsupported, reason recorded | unsupported, reason recorded |

`shared-state-streaming` stays unsupported because neither bridge maps a Strands stream event
onto AG-UI's `STATE_DELTA`; both emit state snapshots only. The manifest now says so.

## Escalations

Both bridge limitations this branch originally reported have since been fixed and released, and
both fixes are verified here at the wire against a live model:

1. **Interrupt payload channel drift, fixed.** The two bridges used to disagree on where a paused
   tool's reason travels, so the demo pages had to read both channels. The released TypeScript
   bridge now carries it under the interrupt's `metadata.reason`, matching Python.
2. **The TypeScript bridge dropped tool calls from the closing snapshot, fixed.** After a tool run
   the final snapshot used to contain only the user message and the assistant's text, so
   tool-rendered cards unmounted when the run finished. Both the tool-call message and its result
   now survive. The same class of loss affected attachments, which is covered above.

3. **Two cells red locally on both columns and not in this ticket's scope:**
   `gen-ui-open-advanced` and `gen-ui-headless-complete`. The last production sweep had this
   integration fully green, so these are regressions relative to prod. Re-running
   `gen-ui-open` with this branch's fixture edits stashed still failed, which rules the fixture
   edits out; finding 2 above is the most likely cause for the card-mounting cells. Worth its own
   ticket rather than widening this one.
4. **`@copilotkit/angular` does not build on `main`** (`ChangeDetectionStrategy.Eager` is not a
   member of the enum in the installed Angular). The showcase test lifecycle builds the shared
   Angular artifact, so this blocks every local showcase run until someone fixes it or has a
   prebuilt artifact lying around.

## Review findings fixed on this branch

Two review rounds ran over the diff. The load-bearing defects they found, all fixed here:

- The resume envelope was read one way on each side. TypeScript understood only the shape the
  Python bridge produces, so a valid pick came back to the model as "user did not pick a time";
  reproduced at the wire on a real model. Both sides now accept both shapes.
- Cancel was asymmetric in the same way: a cancel wrapped as `{"response": {"status": "cancelled"}}`
  was reported as "did not pick a time", and a truthy non-boolean cancel flag cancelled on one
  side and booked on the other.
- An empty label swallowed a valid time on the TypeScript side only.
- The headless resume waited on a single animation frame, which never fires in a background tab,
  leaving the run paused forever.
- A rejected resume left a green "Booked" badge standing, and the picker stayed clickable while
  the resume was in flight.
- The headless probe searched the whole page for booking words, which the demo's own static copy
  already satisfies, and the pick-path spec asserted only that a bubble was visible, which the
  pre-pause bubble already satisfies. Both now assert something the failure mode cannot produce.
- The rationale for the dedicated mount named the wrong tool, the TypeScript integration cited
  Python API names, and the shared docs described a hook the embedded snippet does not show and
  put a standard interrupt's payload in the wrong place.

## Verification

- **D6 aimock replay**, isolated stacks, harness image rebuilt before the runs that depend on
  probe changes: strands 40 cells / 37 pass, strands-typescript 42 cells / 38 pass. Remaining
  reds are the two out-of-scope cells above.
- **Playwright e2e** against aimock-backed containers: strands 17/17 across both interrupt demos
  and both reasoning demos; strands-typescript 8/8 on the interrupt specs (including the
  cancel-path narration regression) and 13/14 on reasoning plus the chain, the single failure
  being the chain spec's stock-card count, which finding 2 explains.
- **Real model**, no aimock, real OpenAI: on both languages the interrupt tool pauses with
  `RUN_FINISHED outcome=interrupt` carrying the model's own topic and attendee, and resuming the
  same thread with `resume[]` finishes the run with the booking narrated back. Reasoning summaries
  stream from the Responses-API agents on both languages.
- **Suspicion check.** The interrupt cells did not pass first try: they failed as
  `reason=text-unstable` until the probe change, and the reasoning cells stayed red until the
  image genuinely carried the bumped SDK. Both fixture defects were reproduced against a
  standalone aimock before and after the fix, with the reference integration as the control.

## Two things the review caught that this branch had broken

- **A Python test regression, not a pre-existing failure.** `test_instrumentor_patch` fails on
  `main` only when that one file runs alone; the full suite is green there because earlier modules
  leave the agent modules cached and the real imports never execute. This branch adds a reasoning
  agent built at import time whose model import sits inside the function body, which module
  caching cannot hide, so the full suite went green to red, deterministically, and CI does run it.
  The cause was that the test's stub modules were never importable as packages: Python reads the
  parent's `__path__` before consulting any meta-path finder, so the stub finder was never
  reached. The stubs now carry an empty `__path__` and are removed again afterwards, and all seven
  test files pass individually as well as together, which was true on neither this branch nor main
  before.
- **Nothing in CI ran the TypeScript unit tests under `showcase/integrations`.** The Python half of
  that tree is gated; its TypeScript half was not, so four test files in the Strands agent,
  including this branch's new one, could fail indefinitely unnoticed. A workflow now runs them.

## Verified on the released bridges

Both columns re-run end to end against a live OpenAI endpoint with the released bridges in the
containers: python 138 passed, typescript 141 passed.

The multimodal round trip that the TypeScript bridge used to break is fixed at both levels. The
bridge returns the user message with its parts intact for inline data, url and pdf sources, and the
demo renders the image and the document chip; all four of its specs pass, having failed under both
live and replay before.

The remaining failures are the same 46 on both columns. They are specs asserting the wording the
aimock fixtures author, which a real model has no reason to reproduce: one expects a humidity
reading of 55% where the model answered 77%, another expects a greeting quoted verbatim from the
fixture. An identical failure set across two independently written bridges is what rules the
bridges out as the cause, and it is the open question this branch leaves behind: whether those
specs should be loosened or explicitly skipped on live runs.

## What is not verified

The local Docker engine died partway through this work and could not be brought back from a
non-interactive session: its pull path wedged first (a plain `docker pull hello-world` hangs while
the host reaches the registry in 0.33s), the VM ran out of disk, and after a restart the engine
never came up again. Consequences, stated plainly rather than papered over:

- Every green cell in the table above was run BEFORE the second review round. The probe and spec
  changes from that round (the headless probe's narration check, the pick-path narration
  assertion, the app-surface scoping) make those assertions strictly stricter, and they have not
  been re-run. They typecheck and the unit suites pass, but the D6 cells need one more sweep.
- `tool-rendering` on the Python column is the one ticket cell with no post-fix run at all. Its
  single red was the showcase CLI failing to start (`Unknown command: test`), not a probe failure,
  and the same cell is green on the TypeScript column with the same fixture edits.
- Three fixture predictions from the review (two `turnIndex` ordering claims and one cancel-gate
  claim) contradict passing per-cell runs on this branch. They were left alone rather than changed
  on a prediction, and they cannot be re-tested until the engine is back.

Two local workarounds were used and are recorded here because they affect how the results were
obtained, not what shipped: the harness image was refreshed by compiling the harness with its own
build script and layering the resulting `dist` onto the existing image (exactly what the real
Dockerfile's runtime stage copies), and `alpine:3.19` was tagged locally from `alpine:latest` so
the pocketbase test-infra image could rebuild. Neither touches the repository.

One further open question from the review, worth a look before merge: the aimock build the
showcase scripts pin treats a fixture's `turnIndex` as a ranking hint in relaxed mode, where an
older build still treats it as a hard gate. Driven directly against the checked-in fixtures, the
newer matcher picks the `turnIndex` leg over the tool-call-chained legs for three tool-rendering
pills. That contradicts the passing per-cell runs, and the same behaviour reproduces against the
gold integration's fixtures on main, so it reads as repo-wide and pre-existing rather than
something this branch introduced. Which build the local D6 stack actually loaded could not be
settled without a running engine.


showcase/aimock/d6/strands/gen-ui-headless-complete.json
```diff
@@ -3,68 +3,89 @@
...
     },
     {
-      "_comment": "stock pill — leg 1 (seq 0): emit get_stock_price(AAPL).",
+      "_comment": "Narration leg, gated on the tool-call id the leg above emits.",
       "match": {
-        "userMessage": "price of AAPL right now",
-        "sequenceIndex": 0,
+        "userMessage": "ship the demo on Friday",
+        "context": "strands",
+        "toolCallId": "call_d5_headless_highlight_001"
```
contextablemark, line 36: **[P2] Match the frontend tool result without relying on its original ID**

The Python bridge gives frontend tool calls a fresh UUID, so the completed `highlight_note` comes back with that UUID rather than `call_d5_headless_highlight_001`. This new condition misses the result and the fixture falls through to another `highlight_note` call. Replaying the bridge's actual requests against this file finishes on the base version and repeats the tool on this PR.

Checking for a tool result fixes the fixture without changing the adapter or probe:

```suggestion
        "hasToolResult": true
```

showcase/integrations/strands/src/app/demos/gen-ui-interrupt/page.tsx
```diff
@@ -52,14 +111,23 @@ function Chat() {
...
+              void resolve(result).catch((error: unknown) => {
+                setResumeFailed(true);
+                queueMicrotask(() => {
+                  throw error;
+                });
+              });
+            }, 500);
           }}
         />
       );
```
contextablemark, line 133: **[P2] Keep a failed resume from marking the next booking as failed**

After one rejected resume, `resumeFailed` stays true. The published hook removes the failed picker, then the next picker captures that old flag and shows “Not booked” as soon as a slot is selected, before its own resume even starts. Resetting the flag alone isn't enough: the hook caches the rendered element, so it won't receive updated props.

This keeps the error outside that element, clears it on a new selection, and stops passing stale failure state into later pickers. The same change is needed in the TypeScript demo below:

```suggestion
function Chat() {
  const [resumeFailed, setResumeFailed] = useState(false);

  return (
    <div className="flex h-full flex-col">
      {resumeFailed && (
        <div
          role="alert"
          data-testid="time-picker-resume-error"
          className="m-4 rounded-lg border border-red-200 bg-red-50 p-3 text-sm text-red-800"
        >
          Your previous selection could not be sent. Please try booking again.
        </div>
      )}
      <div className="min-h-0 flex-1">
        <InterruptChat setResumeFailed={setResumeFailed} />
      </div>
    </div>
  );
}

function InterruptChat({
  setResumeFailed,
}: {
  setResumeFailed: (failed: boolean) => void;
}) {
  useGenUiInterruptSuggestions();

  // Native interrupt path. The backend `schedule_meeting` tool calls Strands'
  // `tool_context.interrupt(...)`; the @ag-ui/aws-strands bridge finishes the
  // run with `outcome.type === "interrupt"` and carries the tool's `reason`
  // under the interrupt's `metadata.reason`. `resolve(...)` resumes the same
  // Strands run, handing the selection back to that `interrupt()` call.
  useInterrupt({
    agentId: "gen-ui-interrupt",
    renderInChat: true,
    render: ({ event, interrupt, resolve }) => {
      const payload = readSchedulingPayload(interrupt, event.value);
      const slots =
        payload.slots && payload.slots.length > 0
          ? payload.slots
          : generateFallbackSlots();
      return (
        <TimePickerCard
          topic={payload.topic ?? "a call"}
          attendee={payload.attendee}
          slots={slots}
          onSubmit={(result) => {
            setResumeFailed(false);
            // Let React show the selection before resume clears the picker.
            // Errors render below, outside the hook-owned element.
            window.setTimeout(() => {
              void resolve(result).catch((error: unknown) => {
                setResumeFailed(true);
                queueMicrotask(() => {
                  throw error;
                });
              });
            }, 500);
          }}
        />
      );
```

showcase/integrations/strands-typescript/src/app/demos/gen-ui-interrupt/page.tsx
```diff
@@ -52,14 +112,23 @@ function Chat() {
...
+              void resolve(result).catch((error: unknown) => {
+                setResumeFailed(true);
+                queueMicrotask(() => {
+                  throw error;
+                });
+              });
+            }, 500);
           }}
         />
       );
```
contextablemark, line 134: Same booking-state issue as the Python demo above. This is the matching fix for the TypeScript page; both suggestions are needed.

```suggestion
function Chat() {
  const [resumeFailed, setResumeFailed] = useState(false);

  return (
    <div className="flex h-full flex-col">
      {resumeFailed && (
        <div
          role="alert"
          data-testid="time-picker-resume-error"
          className="m-4 rounded-lg border border-red-200 bg-red-50 p-3 text-sm text-red-800"
        >
          Your previous selection could not be sent. Please try booking again.
        </div>
      )}
      <div className="min-h-0 flex-1">
        <InterruptChat setResumeFailed={setResumeFailed} />
      </div>
    </div>
  );
}

function InterruptChat({
  setResumeFailed,
}: {
  setResumeFailed: (failed: boolean) => void;
}) {
  useGenUiInterruptSuggestions();

  // Native interrupt path. The backend `schedule_meeting` tool calls Strands'
  // `context.interrupt(...)`; the @ag-ui/aws-strands bridge finishes the run
  // with `outcome.type === "interrupt"`. Where the tool's `reason` travels
  // depends on the bridge version, so the reader below accepts both channels.
  // `resolve(...)` resumes the same Strands run, handing the selection back to
  // that `interrupt()` call.
  useInterrupt({
    agentId: "gen-ui-interrupt",
    renderInChat: true,
    render: ({ event, interrupt, resolve }) => {
      const payload = readSchedulingPayload(interrupt, event.value);
      const slots =
        payload.slots && payload.slots.length > 0
          ? payload.slots
          : generateFallbackSlots();
      return (
        <TimePickerCard
          topic={payload.topic ?? "a call"}
          attendee={payload.attendee}
          slots={slots}
          onSubmit={(result) => {
            setResumeFailed(false);
            // Let React show the selection before resume clears the picker.
            // Errors render below, outside the hook-owned element.
            window.setTimeout(() => {
              void resolve(result).catch((error: unknown) => {
                setResumeFailed(true);
                queueMicrotask(() => {
                  throw error;
                });
              });
            }, 500);
          }}
        />
      );
```

contextablemark (changes requested):
I'd fix these two things before merging: the headless fixture repeats a completed tool call, and one failed booking makes later bookings look failed.

I've left a one-line fixture fix and matching suggestions for both booking demos. The fixture is mismatched to the Python bridge's tool IDs; the booking issue is frontend state.

The suggested fixes pass focused reproductions using captured bridge requests, AIMock, and the published CopilotKit hook. I haven't rerun the Docker/browser D6 sweep.

showcase/integrations/strands/tests/e2e/gen-ui-interrupt.spec.ts
```diff
@@ -126,10 +148,23 @@ test.describe("Gen UI via useInterrupt (inline time picker)", () => {
...
-    ).toBeVisible({
-      timeout: 45_000,
-    });
+    // The cancelled resume has to produce a NEW assistant bubble: the pre-pause
+    // text is already on screen, so "a bubble is visible" passes even when the
+    // resume never lands. Counting is phrasing-agnostic, so it holds against a
+    // real model as well as the fixture.
+    await expect
+      .poll(() => bubbles.count(), { timeout: 45_000 })
+      .toBeGreaterThan(bubblesBeforeCancel);
```
contextablemark, line 157: **[P2] Wait for the cancelled run to finish before checking its text**

The bubble count increases while narration is still streaming, and `.not.toContainText` succeeds as soon as the current text lacks that phrase. With a response streaming `B` then `Booked: Monday 9:00 AM`, this test now passes at `B`; the previous positive assertion rejects the same bad response. Reproduced with the actual cancel test callbacks and real Playwright assertions in both specs for both languages.

Wait for the resumed run to finish before checking its final text. With the empty composer used here, the send/stop control disables when the run ends:

```suggestion
      .toBeGreaterThan(bubblesBeforeCancel);

    // With an empty composer, this disables once the resumed run finishes.
    await expect(
      page.locator('[data-testid="copilot-send-button"]').first(),
    ).toBeDisabled({ timeout: 45_000 });
```

showcase/integrations/strands/tests/e2e/interrupt-headless.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+      page.locator('[data-testid="interrupt-headless-empty"]'),
+    ).toBeVisible({ timeout: 30_000 });
+    await expect(popup).toHaveCount(0);
+
+    // The cancelled resume has to produce a NEW assistant bubble. The pre-pause
+    // text is already on screen, so asserting one is visible proves nothing;
+    // counting is phrasing-agnostic and holds against a real model too.
+    await expect
+      .poll(() => bubbles.count(), { timeout: 45_000 })
+      .toBeGreaterThan(bubblesBeforeCancel);
```
contextablemark, line 132: Same cancellation-streaming race in the Python headless test. This adds the matching wait before checking the final narration.

```suggestion
      .toBeGreaterThan(bubblesBeforeCancel);

    // With an empty composer, this disables once the resumed run finishes.
    await expect(
      page.locator('[data-testid="copilot-send-button"]').first(),
    ).toBeDisabled({ timeout: 45_000 });
```

showcase/integrations/strands-typescript/tests/e2e/gen-ui-interrupt.spec.ts
```diff
@@ -126,10 +148,23 @@ test.describe("Gen UI via useInterrupt (inline time picker)", () => {
...
-    ).toBeVisible({
-      timeout: 45_000,
-    });
+    // The cancelled resume has to produce a NEW assistant bubble: the pre-pause
+    // text is already on screen, so "a bubble is visible" passes even when the
+    // resume never lands. Counting is phrasing-agnostic, so it holds against a
+    // real model as well as the fixture.
+    await expect
+      .poll(() => bubbles.count(), { timeout: 45_000 })
+      .toBeGreaterThan(bubblesBeforeCancel);
```
contextablemark, line 157: Same cancellation-streaming race in the TypeScript inline-picker test. This adds the matching wait before checking the final narration.

```suggestion
      .toBeGreaterThan(bubblesBeforeCancel);

    // With an empty composer, this disables once the resumed run finishes.
    await expect(
      page.locator('[data-testid="copilot-send-button"]').first(),
    ).toBeDisabled({ timeout: 45_000 });
```

showcase/integrations/strands-typescript/tests/e2e/interrupt-headless.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+      page.locator('[data-testid="interrupt-headless-empty"]'),
+    ).toBeVisible({ timeout: 30_000 });
+    await expect(popup).toHaveCount(0);
+
+    // The cancelled resume has to produce a NEW assistant bubble. The pre-pause
+    // text is already on screen, so asserting one is visible proves nothing;
+    // counting is phrasing-agnostic and holds against a real model too.
+    await expect
+      .poll(() => bubbles.count(), { timeout: 45_000 })
+      .toBeGreaterThan(bubblesBeforeCancel);
```
contextablemark, line 132: Same cancellation-streaming race in the TypeScript headless test. This adds the matching wait before checking the final narration.

```suggestion
      .toBeGreaterThan(bubblesBeforeCancel);

    // With an empty composer, this disables once the resumed run finishes.
    await expect(
      page.locator('[data-testid="copilot-send-button"]').first(),
    ).toBeDisabled({ timeout: 45_000 });
```

contextablemark (changes requested):
The fixture issue is fixed by the bridge upgrade. I verified that the tool ID now survives the round trip and the replay finishes.

I'm still requesting changes for the [booking error state](https://github.com/CopilotKit/CopilotKit/pull/6907#discussion_r3953367010) and a new gap in the cancellation tests: they can pass while the response is still streaming, then miss a “Booked” confirmation. The existing booking suggestions still apply; the new suggestions wait for the run to finish.

Verification at `397e08dc`: 65 Python and 16 TypeScript unit tests pass. Focused reproductions confirm both remaining findings and check the new suggestions. No full D6 rerun.

showcase/aimock/d6/strands/gen-ui-headless-complete.json
contextablemark: Verified fixed at `397e08dc` by the `ag_ui_strands==0.4.0` upgrade. The real bridge now preserves `call_d5_headless_highlight_001` through the frontend tool round trip, and replaying its captured continuation against AIMock finishes with narration. The fixture suggestion is no longer needed; resolving this finding.

contextablemark (approved):
Both remaining findings are fixed at `115f0c10`. A failed resume now shows an error outside the picker and clears on the next selection; all four cancellation tests wait for the run to finish before checking the narration.

I reran the original failure cases: both booking demos recover correctly, and the cancellation tests now reject a streamed “Booked” response while accepting a cancelled response. No new actionable findings in the updates.

65 Python and 16 TypeScript unit tests pass, and CI is green. I did not rerun the full Docker D6 sweep.

contextablemark (approved):
Re-reviewed `03ea9c86` after the merge from `main`. No new findings. The merge required no manual conflict resolutions and changes none of the 77 files in this PR’s diff. The Strands sources, dependency locks, fixtures, shared probes, and previous review fixes are unchanged.

Fresh CI on this head passes 65 Strands Python tests, 16 TypeScript tests, and the harness suite (3,729 passed; 18 skipped), plus the documentation checks. All reported CI checks have passed or skipped.

I did not rerun full D6 for this merge. The earlier baseline comparison remains the evidence for the existing failures, tracked separately in [PNI-429](https://linear.app/copilotkit/issue/PNI-429), [PNI-430](https://linear.app/copilotkit/issue/PNI-430), and [PNI-431](https://linear.app/copilotkit/issue/PNI-431).

PR #6909, opened 2026-09-04
fix(runtime): encode SSE response chunks as bytes

### LINKED ISSUE #6888: fix(runtime): afterRequestMiddleware never receives messages/threadId/runId on SSE routes — sse-response writes string chunks
## Summary

`afterRequestMiddleware` never receives `messages`, `threadId` or `runId` for any SSE route in the v2 runtime. They are always `[]` / `undefined`.

The runtime writes **string** chunks into a plain `TransformStream` whose `readable` is handed to a `Response`. When `parseSSEResponse` later calls `.text()` on a clone of that response, undici rejects the stream with `TypeError: Received non-Uint8Array chunk`. The surrounding `try/catch` swallows it and returns `{ messages: [] }`.

Nothing surfaces to the application: the parser catches its own failure, and `callAfterRequestMiddleware` is invoked fire-and-forget with only a `.catch` that logs. The only trace is a `logger.warn` line — `Failed to read SSE response body in afterRequestMiddleware`.

Client-facing streaming is **unaffected**; only the middleware observation path is broken.

## Affected routes

The three routes built on `createSseEventResponse`:

- `agent/run`
- `agent/connect`
- `agent/suggest`

## Root cause

1. `packages/runtime/src/v2/runtime/handlers/shared/sse-response.ts:51` creates a plain `new TransformStream()` (no `writableStrategy`, so no chunk coercion), and `:184` writes into it:

   ```ts
   await writer.write(encoder.encode(event));
   ```

2. `@ag-ui/encoder`'s `EventEncoder.encode()` returns a **string**, not bytes:

   ```js
   encode(event)      { return this.encodeSSE(event); }
   encodeSSE(event)   { return `data: ${JSON.stringify(event)}\n\n`; }   // string
   encodeBinary(event) { … return new TextEncoder().encode(sseString); } // Uint8Array
   ```

   (verified in encoder `0.0.57` and `0.0.59`; runtime `1.70.1` depends on `0.0.59`)

3. `packages/runtime/src/v2/runtime/core/fetch-handler.ts:476` passes `response.clone()` to `callAfterRequestMiddleware`.

4. `core/middleware.ts:101` calls `parseSSEResponse(response)`, which at `core/middleware-sse-parser.ts:38` does `await response.text()`. That throws `Received non-Uint8Array chunk`; `:40` logs the warning and `:41` returns `{ messages: [] }` — with `threadId` and `runId` left `undefined`.

## Reproduction

**Minimal** — a string chunk into a `TransformStream`, wrapped in a `Response`:

```js
const stream = new TransformStream();
const w = stream.writable.getWriter();
(async () => { await w.write("data: {}\n\n"); await w.close(); })();
await new Response(stream.readable).text();
// TypeError: Received non-Uint8Array chunk
```

The same script with `new TextEncoder().encode("data: {}\n\n")` succeeds.

**End-to-end**, against published `@copilotkit/runtime@1.70.1` through the real `createCopilotRuntimeHandler` and the real router, with an `AbstractAgent` subclass emitting `RUN_STARTED` / two `TEXT_MESSAGE_*` turns / `RUN_FINISHED`, POSTing to `/api/copilotkit/agent/default/run`:

```
route status      : 200
content-type      : text/event-stream
Failed to read SSE response body in afterRequestMiddleware
client body       : 759 bytes OK

afterRequestMiddleware received:
  path     : "/api/copilotkit/agent/default/run"
  threadId : undefined
  runId    : undefined
  messages : 0 []
```

Note the client body drains fine — the breakage is confined to the middleware clone.

## Proposed fix

One line in `sse-response.ts:184`:

```diff
-await writer.write(encoder.encode(event));
+await writer.write(encoder.encodeBinary(event));
```

Applying exactly that to the published `1.70.1` dist and re-running the end-to-end script above:

```
route status      : 200
content-type      : text/event-stream
client body       : 759 bytes OK

afterRequestMiddleware received:
  path     : "/api/copilotkit/agent/default/run"
  threadId : "thread_abc"
  runId    : "run_xyz"
  messages : 2 ["user:what is the status?","assistant:All green."]
```

Correct `threadId`, `runId`, and the full message list including the user's turn. Client body byte count is identical before and after (759 bytes).

### Why `encodeBinary` is safe here

`encodeBinary` branches on `acceptsProtobuf` and would emit protobuf frames if the encoder had negotiated it — but `sse-response.ts:53` constructs `new EventEncoder()` with **no** `accept` param, so `acceptsProtobuf` is permanently `false` and `encodeBinary` is just `TextEncoder().encode(encodeSSE(event))`. Confirmed byte-identical to `encode()` across ASCII, multi-byte UTF-8, embedded quotes/newlines, and nested-object payloads. The response's `Content-Type` is hardcoded to `text/event-stream` at `:124` and `getContentType()` is never consulted, so this is consistent with what the route already advertises.

### Nothing depends on the string form of `encode()`

Within `packages/runtime`, `EventEncoder` is imported in exactly one place — `sse-response.ts`. The other `encoder.encode(...)` callsite, `handlers/handle-debug-events.ts:29,34`, uses a plain `TextEncoder` and is unrelated. Remaining `EventEncoder` usages in the monorepo are `examples/` and `showcase/` agent servers writing their own responses, unaffected by this change.

An alternative, if you'd rather not touch the encoder call, is to give the `TransformStream` a `transform` that coerces strings to bytes — but `encodeBinary` already exists for this and is the smaller change.

## Why this shipped green

There is no test covering `SSE handler → afterRequestMiddleware`.

- `src/v2/runtime/__tests__/middleware-sse-parser.test.ts:5-6` builds its fixture as `new Response(bodyString, …)`. A `Response` constructed from a string is encoded to bytes by undici internally, so the parser is only ever exercised against an already-correct body — never against the `TransformStream` + `writer.write(string)` path the real handler uses.
- `src/v2/runtime/__tests__/middleware.test.ts:190` (`"passes parsed messages to afterRequestMiddleware"`) hits `/info`, a JSON route, and asserts `messages` equals `[]` — which is the broken outcome, and passes either way.

A regression test that drives an actual SSE route and asserts non-empty `messages` would have caught this and would keep it caught.

## Related, but a separate bug

`cpk-debug-events` also delivers empty `messages`/`threadId`/`runId`, but **not** for this reason — `handle-debug-events.ts` writes proper `Uint8Array` chunks via `TextEncoder`, so `.text()` succeeds there. It's the payload shape: the route emits `DebugEventEnvelope` (`{ timestamp, agentId, threadId, runId, event }`, per `core/debug-event-bus.ts`), while `parseSSEResponse` switches on a top-level `event.type`. Every line falls through the switch. Lower priority — the route 404s in production — but worth a note if the parser is meant to cover it.

## Impact

Any consumer using `afterRequestMiddleware` to observe completed runs — logging, analytics, persistence, audit trails — silently receives nothing, with no application-visible error.

## Environment

- `@copilotkit/runtime` `1.62.3` (write at `dist/v2/runtime/handlers/shared/sse-response.mjs:66`) and `1.70.1` (same write at `:87`) — verified from the published tarballs
- `@ag-ui/encoder` `0.0.57` and `0.0.59`
- Reproduced on Node `v22.14.0`, `v22.22.2`, `v24.14.1`, `v24.16.0`
- Source lines cited against `3472104` on `main`

Fnine59:
## What does this PR do?

SSE runtime routes now write `EventEncoder` output as `Uint8Array` chunks with
`encodeBinary()`. The encoder is constructed without protobuf negotiation, so the
wire payload remains SSE text while the response body satisfies the Web Streams
byte-chunk contract. This lets the cloned response used by
`afterRequestMiddleware` be consumed with `Response.text()` instead of silently
falling back to empty middleware data.

The regression test drives the real `/agent/default/run` route through
`createCopilotRuntimeHandler` and verifies that the middleware receives the
message snapshot, thread ID, and run ID from the SSE event stream.

## Related PRs and Issues

- Fixes #6888

## Validation

- `pnpm exec nx run @copilotkit/runtime:test -- src/v2/runtime/__tests__/middleware-sse-route.test.ts`
- `pnpm exec nx run @copilotkit/runtime:check-types`
- `pnpm exec nx run @copilotkit/runtime:build`
- `pnpm exec nx run-many -t publint,attw,check-dts --projects=@copilotkit/runtime`
- `pnpm run lint`
- `pnpm exec oxfmt packages/runtime/src/v2/runtime/handlers/shared/sse-response.ts packages/runtime/src/v2/runtime/__tests__/middleware-sse-route.test.ts --check`

## Checklist

- [x] I have read the Contribution Guide.
- [x] I added focused regression coverage for the runtime SSE route.
- [x] No documentation, version, changelog, or changeset update is needed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of Server-Sent Events responses by using binary-safe event encoding.
* **Tests**
  * Added coverage verifying that agent run messages and identifiers are correctly passed to request middleware.


lukasmoschitz (approved):
Verified against current main: I dropped your test file in without the one-line change and it fails with "afterRequestMiddleware received no SSE messages"; with `encodeBinary` it passes, and the full runtime suite is green at 157 files. I also checked that on the pinned encoder 0.0.59 `encodeBinary` produces byte-identical content to `encode`, so the wire format doesn't change — only the chunk type does.

Thanks for writing a test that actually fails without the fix. That's the part that made this easy to take.

I'm adding the issue references at merge: this closes #6888, and I'm linking #6919 without closing it, since that report lists two further workerd blockers this doesn't touch.

PR #6913, opened 2026-09-05
feat(runtime): warn when assembled agent context is oversized

zzt417520:
## Summary

Follow-up to #6754 (closed) per maintainer review: replace the truncation approach with warn, do not mutate. The previous PR silently truncated context values, which (a) left the usually-larger input.state uncapped, (b) changed the public convertInputToTanStackAI contract, and (c) could throw on null/object values. This iteration keeps all data intact and only emits a single console.warn when the assembled context block exceeds a threshold.

- Warnings cover **both** injection sites (BuiltInAgent and convertInputToTanStackAI).
- Measures the **assembled total** across context + serialized state, matching the actual prompt shape.
- Adds an env knob COPILOTKIT_AGENT_CONTEXT_WARN_THRESHOLD (default 100,000 chars; -1 disables).
- No data mutation, no public contract change.

Closes #6754.

## Summary by CodeRabbit

* **New Features**
  * Added warnings when assembled agent context exceeds the configured size threshold.
  * Context size calculations include supplied context details and serialized state without changing the data.
  * Warning thresholds can be configured through the environment, including an option to disable warnings.
  * Warnings are limited to one per processing path across supported agent workflows.

* **Tests**
  * Added coverage for size calculations, threshold boundaries, configuration, disabled warnings, deduplication, and data preservation.

### DIFF
--- a/packages/runtime/src/agent/__tests__/agent-context-size.test.ts
+++ b/packages/runtime/src/agent/__tests__/agent-context-size.test.ts
@@ -1,0 +1,120 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
+import {
+  DEFAULT_AGENT_CONTEXT_WARN_THRESHOLD,
+  getAgentContextWarnThreshold,
+  resetAgentContextWarnState,
+  warnIfAssembledAgentContextOversized,
+} from "../agent-context-size";
+
+describe("getAgentContextWarnThreshold", () => {
+  const originalThreshold = process.env.COPILOTKIT_AGENT_CONTEXT_WARN_THRESHOLD;
+
+  afterEach(() => {
+    if (originalThreshold === undefined) {
+      delete process.env.COPILOTKIT_AGENT_CONTEXT_WARN_THRESHOLD;
+    } else {
+      process.env.COPILOTKIT_AGENT_CONTEXT_WARN_THRESHOLD = originalThreshold;
+    }
+  });
+
+  it("uses the default when the env value is unset, empty, or whitespace", () => {
+    delete process.env.COPILOTKIT_AGENT_CONTEXT_WARN_THRESHOLD;
+    expect(getAgentContextWarnThreshold()).toBe(
+      DEFAULT_AGENT_CONTEXT_WARN_THRESHOLD,
+    );
+    process.env.COPILOTKIT_AGENT_CONTEXT_WARN_THRESHOLD = "";
+    expect(getAgentContextWarnThreshold()).toBe(
+      DEFAULT_AGENT_CONTEXT_WARN_THRESHOLD,
+    );
+    process.env.COPILOTKIT_AGENT_CONTEXT_WARN_THRESHOLD = "  ";
+    expect(getAgentContextWarnThreshold()).toBe(
+      DEFAULT_AGENT_CONTEXT_WARN_THRESHOLD,
+    );
+  });
+
+  it("rejects partially parsed or malformed values and falls back to the default", () => {
+    for (const bad of ["100abc", "1.5", "-1oops", "abc", "12px", "-2", "--1"]) {
+      process.env.COPILOTKIT_AGENT_CONTEXT_WARN_THRESHOLD = bad;
+      expect(getAgentContextWarnThreshold()).toBe(
+        DEFAULT_AGENT_CONTEXT_WARN_THRESHOLD,
+      );
+    }
+  });
+
+  it("accepts a complete non-negative integer and the -1 disable sentinel", () => {
+    process.env.COPILOTKIT_AGENT_CONTEXT_WARN_THRESHOLD = "10";
+    expect(getAgentContextWarnThreshold()).toBe(10);
+    process.env.COPILOTKIT_AGENT_CONTEXT_WARN_THRESHOLD = "0";
+    expect(getAgentContextWarnThreshold()).toBe(0);
+    process.env.COPILOTKIT_AGENT_CONTEXT_WARN_THRESHOLD = "-1";
+    expect(getAgentContextWarnThreshold()).toBe(-1);
+  });
+});
+
+describe("warnIfAssembledAgentContextOversized", () => {
+  const originalThreshold = process.env.COPILOTKIT_AGENT_CONTEXT_WARN_THRESHOLD;
+
+  beforeEach(() => {
+    vi.spyOn(console, "warn").mockImplementation(() => {});
+    resetAgentContextWarnState();
+    delete process.env.COPILOTKIT_AGENT_CONTEXT_WARN_THRESHOLD;
+  });
+
+  afterEach(() => {
+    vi.restoreAllMocks();
+    if (originalThreshold === undefined) {
+      delete process.env.COPILOTKIT_AGENT_CONTEXT_WARN_THRESHOLD;
+    } else {
+      process.env.COPILOTKIT_AGENT_CONTEXT_WARN_THRESHOLD = originalThreshold;
+    }
+  });
+
+  it("warns as soon as the assembled size exceeds the default threshold", () => {
+    warnIfAssembledAgentContextOversized(
+      DEFAULT_AGENT_CONTEXT_WARN_THRESHOLD + 1,
+      "builtIn",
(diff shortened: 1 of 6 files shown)


zzt417520:
Good catch — acknowledged. computeAssembledAgentContextSize no longer approximates: it now takes the exact ariant each path emits (`builtIn` vs `tanstack`), and reconstructs the same fragments that get injected:

- **builtIn**: ## Context from the application header + per-entry trailing newline, plus the ## Application State prose and `json fences.
- **tanstack**: per-entry strings joined by newline (no trailing newline), plus the Application State: prose and `json fences.

Each call site passes its own variant, so the measured size equals the assembled block it actually produces. Added boundary tests that rebuild the fragments independently and assert the measurement matches, plus warn/no-warn boundary cases for both paths.

Runtime suite: 157 files / 2296 tests pass; lint 0 errors; tsc clean.

zzt417520:
Pushed the fix for the measurement mismatch — computeAssembledAgentContextSize now reconstructs the exact fragments each path emits (builtIn vs tanstack), and both call sites pass their own variant. Boundary tests confirm measurement matches assembly. Runtime suite (157 files / 2296 tests) passes.

zzt417520:
Right — that separator was missing. I now count the \n that systemPrompts.join(`\n`) inserts between the final context entry and the Application State block when both are present, in the 	anstack variant. Added a dedicated test (rebuilding the fragment including that separator) so a threshold equal to the actual assembled length now warns. Runtime suite (157 files / 2296 tests) passes; tsc + oxfmt clean.

zzt417520:
Both nits resolved.

**Strict threshold parsing** — getAgentContextWarnThreshold() now accepts only a complete integer (`/^-?\d+$/`). Values like `100abc`, `1.5`, `-1oops`, `12px` fall back to the default; `10` parses to 10, `-1` still disables. Added tests for the rejection cases.

**Once-per-call-site warning** — the helper now tracks a per-variant `warnedVariants` set and only emits the `console.warn` the first time an oversized value is seen for a given call site, so repeated agent turns don't spam it. A test-only `resetAgentContextWarnState()` clears the bookkeeping.

Verification: runtime suite (157 files / 2300 tests) passes; oxfmt + oxlint (0 warnings) + tsc clean.

zzt417520:
All three points addressed.

**Restrict disablement to -1** — getAgentContextWarnThreshold() now only accepts a complete non-negative integer (`/^\d+$/`) or the literal `-1`. `-2`, `--1`, `100abc`, `1.5` etc. fall back to the default; `0` and `10` parse as-is. Added a `-2` regression test.

**Align empty-context measurement** — the built-in header is now counted only when `context.length > 0` (matching `agent/index.ts`), and the tanstack context→state separator is only counted when there is at least one context entry. Added an empty-context boundary test for both paths.

**Count the serialized value exactly** — the state size is now `String(JSON.stringify(state, null, 2)).length`, so a `toJSON()` that returns `undefined` measures as the literal `"undefined"` (9 chars) that both consumers actually inject. Added a regression test for that case.

Verification: runtime suite (157 files / 2302 tests) passes; oxfmt + oxlint (0 warnings) + tsc clean.

zzt417520:
Agreed, and fixed. The helper no longer serializes `state` itself — each prompt builder serializes once (`String(JSON.stringify(input.state, null, 2))`), uses that exact text to build the system prompt, and passes the same `serializedState` string into `warnIfAssembledAgentContextOversized`/`computeAssembledAgentContextSize`. So `state.toJSON()` is invoked exactly once per assembly, and the measurement is byte-for-byte the text the prompt contains.

Added regression tests for both paths with a `toJSON` spy asserting it's called exactly once during a full assembly (BasicAgent and convertInputToTanStackAI). Verification: runtime suite (157 files / 2304 tests) passes; oxfmt + tsc clean.

lukasmoschitz (changes requested):
Ran this merged with main — full runtime suite green. I also dropped a temporary assertion into both call sites comparing your computed size against the length of the string that actually gets assembled, and it matched every time (18 hits on built-in, 34 on tanstack). So the arithmetic is right, which I honestly wasn't expecting from a reconstruction this fiddly.

Still want to change the approach though. In `index.ts` you already have `systemPrompt = parts.join("")` directly above the call, so `systemPrompt.length` is exact for free and can't drift. As it stands the helper keeps copies of prompt literals that live in two other files, and the tests recreate those same literals — so someone rewording the Application State preamble breaks the measurement without failing anything.

It'd also close a gap: I had to subtract `config.prompt.length` to get my assertion to pass, since the developer's own prompt sits in that same string but isn't counted. That's probably the most likely thing to bloat a prompt in the first place.

On tanstack: `chat()` doesn't join anything, the provider adapter does. `ai-openai` uses `\n`, so you're right today, but we have two `@tanstack/ai` lines in the tree with differently shaped adapters. I'd drop the per-character separator counting there and call it approximate.

Two minor ones. `warnedVariants` is module state that's only cleared in tests, so on a long-running server the warning fires twice per process, ever — fine if that's intended, just say so in the comment. And `/^\d+$/` accepts `0`, which means warn on everything; worth documenting next to `-1`.

zzt417520:
Taken on board — refactored to your suggestion so the measurement can't drift.

**builtIn** — `warnIfAssembledAgentContextOversized` now takes the actual size, and `index.ts` passes `systemPrompt.length` (the exact string right above the call). No more reconstructed literals, and it now also covers the developer's own `config.prompt` (the likely bloat source you called out).

**tanstack** — dropped the per-character separator counting. `convertInputToTanStackAI` now passes `systemPrompts.reduce((sum, p) => sum + p.length, 0)` and we document it as approximate, since the provider adapter (not `chat()`) does the final join and separators differ across `@tanstack/ai` lines.

**Dedupe** — `warnedVariants` is module-level and only reset in tests; added a comment stating it's intentional (once per process per call site) and cleared only by the test-only `resetAgentContextWarnState`.

**Threshold** — documented `0` (warn on any non-empty prompt) alongside `-1` (disable) in the module doc comment.

Removed the reconstruction helper and its recreated literal tests; added unit tests for the simplified helper. Runtime suite (157 files / 2297 tests) green, tsc clean. Thanks for the careful pass.

lukasmoschitz:
I asked for the refactor and you delivered it cleanly, so I owe you a straight answer rather than another round: I'm closing this. I had a second pair of eyes go over the result without my earlier notes, and it changed my mind about the feature itself, not just the details.

The things that decided it:

- **Nothing tests the feature.** I removed both call sites and ran the suite — 158 files green. The whole integration can be deleted and CI is happy.
- **Two of the new tests can't fail.** The "serializes state exactly once" tests pass on `main` too; the old inline `JSON.stringify` already ran once. I verified that by reverting the hoist.
- **`console.warn` is unguarded.** This package's convention is explicit — `// best-effort: logging must never break a run` in `v2/runtime/runner/in-memory.ts:118`, `:329`, `:504`, each inside a `try`. A patched or broken console here takes the run down, and the error escapes `convertInputToTanStackAI`, which is a public helper.
- **It doesn't measure what the name says.** `systemPrompt` is prompt + context + state. Conversation history and tool schemas — the two biggest real token sinks — aren't in it. A 2 MB user message with a 500 KB tool description produces no warning at all.
- **The dedupe removes the value it exists to protect.** `warnedVariants` is never cleared in production, so after one warning a 5,000,000-character prompt later in the same process is silent.

Underneath all that: the default threshold fires on usage this file itself calls legitimate ("an application legitimately shares large state — a whole document, sheet or research report"), and characters aren't tokens, as your own doc comment says.

If prompt size is worth surfacing, the runtime already has better material: providers return real token usage and `tanstack.ts` already imports `aggregateRunUsage`, so `systemPromptChars` alongside that would be per-run, attributable and needs no threshold. Or a structured line through the existing `logger` / `debugLogger` behind the debug flag. Either is smaller than this and opt-in. If you want to take one of those on, open an issue first and let's agree the shape before you write it — I don't want you doing a third round on this.

Genuine credit: dropping truncation for a non-mutating signal was the right call, and the env-var parser is careful work — rejecting `parseInt`'s prefix parsing with a full-match regex is the one part of this I'd have taken as-is.

PR #6914, opened 2026-09-05
docs(examples): fix stale counts and add missing integration/showcase rows

Ayush7614:
The examples index drifted from reality: header says 52 repos / 20 integrations / 25 showcases, but the tree has 61 consolidated demos (22 integrations excl _parity, 7 canvas, 32 showcases). The integrations table omits adk-angular + agentcore and the showcases table omits 7 demos (a2ui-pdf-analyst, arcade-tools, claude-managed-agents, daytona-runcode, deep-agents-finance-erp, open-mcp-client, oracle-agent-memory).

This PR corrects the counts and adds the missing rows with descriptions taken from each demo README. Verified: every linked path in the file resolves to an existing directory (61 links, 0 missing); oxfmt check passes. No behavior change.

## Summary by CodeRabbit

- **Documentation**
  - Updated the examples documentation to reflect the current example counts.
  - Added integration examples for Angular and AgentCore.
  - Added seven showcase examples.

PR #6915, opened 2026-09-05
test(sdk-js): document header propagation API and cover async/edge cases

Ayush7614:
Expands the JSDoc for withForwardedHeaders/getForwardedHeaders (params, returns, example) and adds 3 additive tests: verbatim value preservation, async callback propagation, and empty-input handling. No behavior change; existing tests untouched. Verified: oxfmt passes; header-propagation logic unmodified (AsyncLocalStorage x-* filtering preserved); new expectations match the implementation.

## Summary by CodeRabbit

* **Documentation**
  * Clarified forwarded-header behavior, including case handling, key normalization, return values, and behavior outside an active scope.
* **Tests**
  * Added coverage for special characters, asynchronous callbacks, promise propagation, and empty header inputs.

### DIFF
--- a/packages/sdk-js/src/__tests__/header-propagation.test.ts
+++ b/packages/sdk-js/src/__tests__/header-propagation.test.ts
@@ -127,5 +127,29 @@ describe("header-propagation", () => {
         expect(getForwardedHeaders()).toEqual({ "x-aimock-strict": "true" });
       });
     });
+
+    it("preserves header values verbatim (including special chars)", () => {
+      withForwardedHeaders({ "x-trace": "a=b&c/d;e:f" }, () => {
+        expect(getForwardedHeaders()).toEqual({ "x-trace": "a=b&c/d;e:f" });
+      });
+    });
+
+    it("supports async callbacks and propagates the promise value", async () => {
+      const result = await withForwardedHeaders(
+        { "x-request-id": "async-1" },
+        async () => {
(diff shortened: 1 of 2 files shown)


lukasmoschitz (approved):
Ran it: 11 tests pass. I also checked the doc comment against the implementation — the `x-*` filter, the lowercase normalisation and the empty-object-outside-scope behaviour all match what the code does, which is the part that matters for docs like these.

lukasmoschitz:
I approved this earlier, then went back and looked harder at what it actually adds — so let me be straight rather than quietly leave it open. I'm closing it.

The test file already had eight tests. Of your three, "preserves header values verbatim" can't fail: the implementation is a plain assignment, nothing anywhere touches the value. And "handles empty headers input" is already covered between the existing `withForwardedHeaders({}, () => 42)` case and the no-`x-*`-headers case. The async one is genuinely new — nothing currently tests that an async callback's resolved value comes back with the AsyncLocalStorage context intact across an internal await. That's a real gap and a good catch.

What tips it is the file. `header-propagation.ts` is a deprecated v1 export with no v2 replacement, it has one internal consumer, and every showcase in this repo has written its own local copy rather than importing it. Adding usage examples there points people at something we're steering away from, and one good test isn't enough to carry that.

This isn't a reflection on the work. #6914 and #6916 both went in today and were exactly the right kind of contribution. If you want more of that, tests against v2 surface are where they'd count.

PR #6918, opened 2026-09-05
feat(examples): add minimal SmolAgents starter

Ayush7614:
New integration starter examples/integrations/smolagents pairing a minimal Next.js CopilotSidebar UI with a HuggingFace SmolAgents CodeAgent (InferenceClientModel) served over AG-UI SSE (POST /agui, GET /health). Mirrors the agno/pydantic-ai layout. Verified: python py_compile passes on agent files; oxfmt passes on all TS/JSON files; no existing parity instances touched.

## Summary by CodeRabbit

- **New Features**
  - Added a complete SmolAgents integration example with a Next.js CopilotKit chat interface.
  - Added a Python-based SmolAgents CodeAgent connected through AG-UI.
  - Added streaming agent responses, health monitoring, and CopilotKit API integration.
  - Added setup and run scripts for both the web interface and agent service.

- **Documentation**
  - Added setup instructions, configuration guidance, endpoints, project structure, and customization notes.

### DIFF
--- a/examples/integrations/smolagents/.gitignore
+++ b/examples/integrations/smolagents/.gitignore
@@ -1,0 +1,35 @@
+# dependencies
+/node_modules
+/.pnp
+.pnp.*
+.yarn/*
+
+# next.js
+/.next/
+/out/
+
+# production
+/build
+
+# misc
+.DS_Store
+*.pem
+
+# debug
+npm-debug.log*
+yarn-debug.log*
+
+# env files (can opt-in for committing if needed)
+.env*
+
+# vercel
+.vercel
+
+# typescript
+*.tsbuildinfo
+next-env.d.ts
+
+# python
+agent/.venv/
+__pycache__/
+.venv/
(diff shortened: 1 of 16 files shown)


lukasmoschitz:
Thanks for putting this together, and sorry for the slow reply.

I'm going to close this, and I'd rather be straight about why than send you round in circles on details. Two reasons.

CONTRIBUTING asks for an issue or a Discord conversation before significant new work, exactly so nobody sinks a weekend into something we then turn down. We didn't hold up our end by leaving this sitting for days.

The bigger one: agno and pydantic-ai, which you modelled this on, both hand AG-UI off to their framework's own adapter (`agno.os.interfaces.agui`, `pydantic_ai.ui.ag_ui`). smolagents has no such bridge, so this example writes the protocol frames by hand — which is where the missing terminal `RUN_ERROR` and the single non-streaming delta come from. That would leave us maintaining a hand-rolled AG-UI implementation inside an examples directory our CI never builds.

Two things worth flagging regardless of this PR. `POST /agui` is unauthenticated and passes arbitrary input to a `CodeAgent`, which writes and executes Python — no `executor_type` set and no warning in the README. For something people copy as a template that deserves care. And `InferenceClientModel()` reads `HF_TOKEN`, but the README tells people to export `OPENAI_API_KEY`, so the starter doesn't actually run as documented.

If you want a smolagents integration to exist, open an issue and let's scope it there — a real AG-UI bridge would be the way in, and you already have `ag-ui-protocol` in the pyproject. Your other three PRs are the kind that land easily, and I'm on those now.

PR #6923, opened 2026-09-05
docs(intelligence): document the Query API

davidmckayv:
## What this is

A developer-facing page for the Intelligence Query API: how to read every piece of your own Intelligence data from a script, a notebook, or an agent, without a browser session.

Companion to the implementation in the Intelligence repo. Sits alongside [Memories & Recall](/intelligence/memories) and [Threads & Persistence Architecture](/intelligence/threads-explained), and follows the same shape.

## What it covers

- The resource catalog, and the instruction to ask the deployment rather than trust the page, because the catalog generates from the same definitions the query planner uses
- The scope model, and specifically that `:read` and `:content` are independent, so a key can list every thread without reading a single message body
- List and filter syntax, keyset paging, projection
- Declarative aggregation, including percentiles and parameterized measures
- Arbitrary read-only SQL with frozen-result paging
- The four MCP tools
- Errors, rate-limit headers, and the audit trail the API keeps of its own reads

## The parts worth reviewing

The page states the things that are easy to get wrong and expensive to discover late:

- **A measure's filter scopes to that measure alone.** In a request that asks for both a count and a term measure, the count is every row in the group and the measure counts only matches. Reading it the other way gives you a number that looks right and is not.
- **The organization is never a parameter.** A SQL statement with no `WHERE` clause still reads exactly one tenant, because tenancy is injected as a session setting and enforced by row-level security rather than being part of the SQL you write.
- **A field omitted for scope is named in the response**, while a field named in `fields` that does not exist is an error. Those are different failures and the page says so.
- **Secret columns are unreachable at any scope**, withheld at the database grant as well as excluded from the projection.

## Verification

Every route, parameter, filter operator, aggregate operation, error code and response header on this page was exercised against a running deployment rather than read out of the source. That sweep is what turned up four defects in the API itself, fixed separately in the Intelligence repo: filters that could not work at all over HTTP, pagination that silently capped a narrowed projection at one page, an `include_total` parameter that was accepted and ignored, and `having` accepting only one of the two shapes `filter` accepts.

Rendered locally and read end to end: tables, callouts, code blocks and all four cross-links resolve.

## Summary by CodeRabbit

- **Documentation**
  - Added a comprehensive Intelligence Query API guide.
  - Documented resource discovery, bearer-key authentication, scopes, tenancy enforcement, redaction, filtering, pagination, aggregation, read-only SQL, MCP tools, errors, rate limits, and audit events.
  - Clarified security constraints, licensing requirements, time-window parameters, and unsupported parameter handling.
  - Documented exports, vector access, and downloads.
  - Added instructions for managing query keys through the CLI and REST API.
  - Added the Query API page to the Intelligence documentation navigation.

### DIFF
--- a/showcase/shell-docs/src/content/docs/intelligence/query-api.mdx
+++ b/showcase/shell-docs/src/content/docs/intelligence/query-api.mdx
@@ -215,10 +215,32 @@ new pagination rather than continuing the old one.
 | `start`, `end` | Half-open time window `[start, end)` on the resource's time field, as RFC 3339 timestamps. |
 | `include_deleted` | Include soft-deleted rows, which are excluded by default. |
 | `include_total` | Add `page.total`, the count of rows matching the filters. Absent unless asked for. |
+| `expand` | Attach a declared related resource, one level deep. |
 
 A parameter this API does not implement is refused by name rather than accepted
 and ignored, so a request that returns rows is a request that did what you
 asked.
+
+### Expanding related rows
+
+`expand` attaches a related resource that the model declares, one level deep:
+
+```bash
+curl -G https://your-deployment/v1/threads \
+  -H "Authorization: Bearer cpkq-<org>_<short>_<long>" \
+  --data-urlencode "expand=runs"
+```
+
+A many-to-one relation attaches a single object, a one-to-many attaches an
+array, and an empty result attaches an empty array rather than omitting the key,
+so "expanded, nothing there" is distinguishable from "not expanded". Asking for
+a relation the resource does not declare is refused, and the error names the
+ones it does.
+
+Each side is fetched separately rather than joined, so the expanded resource is
+gated by its own scopes: expanding into something your key cannot read is
+refused rather than quietly returning it. `GET /v1/meta/resources/:name` lists
+what a resource can expand, with a cost for each.
 
 ## Aggregating
 
--- a/showcase/shell-docs/src/content/docs/intelligence/query-api.mdx
+++ b/showcase/shell-docs/src/content/docs/intelligence/query-api.mdx
@@ -53,6 +53,31 @@ The Query API is gated behind a `query-api` feature flag that defaults to off.
 Turn it on for the deployment before minting a key.
 </Callout>
 
+## Getting a key
+
+Mint one from the CLI:
+
+```bash
+copilotkit query-keys create --name reporting
+```
+
+That prints the token once. Only a hash is stored, so it cannot be recovered:
+if you lose it, mint another. Omitting `--scopes` gives the key every scope, and
+it expires in a year. Narrow it when you know what you need:
+
+```bash
+copilotkit query-keys create \
+  --name nightly-export \
+  --scopes threads:read,analytics:read \
+  --projects 4 \
+  --expires-in-days 90
+```
+
+`copilotkit query-keys list` shows what exists, marking any key that has
+expired, and `copilotkit query-keys revoke <id>` retires one. The same
+operations are available over REST at `/api/query-keys`, which is
+session-authenticated: a query key cannot mint another query key.
+
 ## Authentication
 
 Authentication is a single bearer token. There is no user header and no cookie,
--- a/showcase/shell-docs/src/content/docs/intelligence/query-api.mdx
+++ b/showcase/shell-docs/src/content/docs/intelligence/query-api.mdx
@@ -1,0 +1,356 @@
+---
+title: "Query API"
+description: "Read and query every piece of your own Intelligence data from a script, a notebook, or an agent: threads, utterances, memories, learning provenance, channels, and analytics, over REST, arbitrary read-only SQL, or MCP."
+icon: "lucide/Database"
+doc_type: how-to
+---
+
+Everything Intelligence stores about your organization is yours, and you should
+not have to open a browser to read it. The Query API is a machine credential and
+a JSON surface over the whole of it: list a resource, aggregate it, or write SQL
+against it, with tenancy enforced underneath rather than requested politely.
+
+If you want the conversation persistence model itself, read
+[Threads & Persistence Architecture](/intelligence/threads-explained). If you
+want long-term memory, read [Memories & Recall](/intelligence/memories). This
+page is about reading all of it from the outside.
+
+## What you can reach
+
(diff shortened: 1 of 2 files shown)


showcase/shell-docs/src/content/docs/intelligence/query-api.mdx
```diff
@@ -0,0 +1,507 @@
...
+
+```bash
+curl https://your-deployment/v1/threads?limit=10 \
+  -H "Authorization: Bearer cpkq-<org>_<short>_<long>"
+```
+
+The key carries three things: the organization it may read, the scopes it holds,
+and optionally a project allowlist. The organization is not a parameter you can
+pass. Every statement the API compiles is constrained to the key's organization
+before any filter you supply, and the database enforces the same constraint
```
MikeRyanDev, line 94: [P1] Limit the RLS guarantee to the execution paths that use it

The companion implementation passes the ordinary product pool to modelled reads and exports. Its restrictive RLS policies apply to cpki_query, used by the separate SQL pool; the migration deliberately leaves owner behavior unchanged. This paragraph promises an independent database tenancy backstop under every compiled read, which is not the deployed wiring. Describe modelled predicate enforcement and restricted SQL RLS separately, or change and test the implementation before making the broader claim. The related scope/license claims also need to stay blocked on the access fixes requested in Intelligence#1161.

showcase/shell-docs/src/content/docs/intelligence/query-api.mdx
```diff
@@ -0,0 +1,507 @@
...
+cannot widen the read by leaving out a `WHERE` clause, because tenancy is not
+part of the SQL you write. It is injected as a session setting and enforced by
+row-level security against a role that holds `SELECT` and nothing else.
+
+The statement is parsed with the real PostgreSQL grammar rather than matched
+against patterns, so writes, a second statement smuggled after a semicolon,
+`SELECT INTO`, locking clauses, system catalogs, file-reading functions, and
+anything that could forge the tenant setting are all refused.
+
+The response carries a `query_id`. Paging with it returns pages of the same
```
MikeRyanDev, line 344: [P2] Document the SQL result ceiling and distinguish it from exports

runFrozenSql caps the stored result at 10,000 rows with no truncation signal; the response limit is only page size. A local PostgreSQL test produced 10,001 rows and received rowCount=10000 with no overflow indication. Calling this a way to make a long export consistent invites silent partial extraction. Document the ceiling and timeout, and reserve complete extraction for /v1/exports after its correctness fixes. Also clarify that the asynchronous export worker uses separate page queries, not this single frozen snapshot.

showcase/shell-docs/src/content/docs/intelligence/query-api.mdx
```diff
@@ -0,0 +1,507 @@
...
+same organization cannot collect your output. Quota is 20 jobs a day and one at
+a time; asking for a second while one is running returns
+`EXPORT_QUOTA_EXCEEDED`.
+
+Parquet columns are typed rather than all-text, so the file re-imports into a
+warehouse with dates as dates and counts as integers.
+
+### Exports carry your vectors
+
+This is the one place `memories.embedding` is available. It is deliberately
```
MikeRyanDev, line 412: [P2] Reconcile export-only vector copy with the SQL surface

The implementation only applies exportOnly through the modelled projection. Its SQL inspector checks the embedding field’s memory:content scope, and the database grants SELECT on memories; it does not reject embedding as export-only. Therefore SELECT embedding FROM cpki.memories remains available through SQL to a sufficiently scoped key. Either enforce export-only there too or narrow this paragraph to modelled list/query responses.

showcase/shell-docs/src/content/docs/intelligence/query-api.mdx
```diff
@@ -0,0 +1,507 @@
...
+operations are available over REST at `/api/query-keys`, which is
+session-authenticated: a query key cannot mint another query key.
+
+## Authentication
+
+Authentication is a single bearer token. There is no user header and no cookie,
+because a query key is an organization credential rather than a user session.
+
+```bash
+curl https://your-deployment/v1/threads?limit=10 \
```
MikeRyanDev, line 87: [P2] Quote the copyable query-string URL

Under default zsh, the unquoted ? is a filename glob and this example fails with “no matches found” before curl runs. Quote the URL or use curl -G with --data-urlencode, as the later examples do. Apply the same shell check to placeholder path URLs containing angle brackets.

MikeRyanDev (changes requested):
Request changes on `be12b0b198ce7daa4587a0386b22735834e3face`.

The guide’s discovery-first flow and read/content distinction are clear. The security guarantees and extraction guidance must match the final implementation before publication. Inline comments identify the RLS claim, silent SQL cap, vector availability, and a copyable shell failure.

Also update “four tools” to reflect the nine listed tools, with SQL/recall availability conditional on deployment capability. Check the claims that every read is audited and every response carries rate-limit headers against the MCP and error paths.

Validation: direct compilation of this MDX file with the installed next-mdx-remote serializer passed (77,594 compiled bytes); shell-docs `npm run lint` and diff whitespace checks passed. Full `npm run build` stops at missing js-yaml resolution from the sibling showcase harness after the documented standalone installs. Full docs typecheck, test suite, and browser rendering were not completed; this is a content/contract review, not proof of a production docs build. Implementation behavior was checked at Intelligence head 377890439071724cb0d49fa2099246d277694833 with targeted compiler, database, MCP, and PostgreSQL concurrency tests.

davidmckayv:
Closing: this documents a surface that is being deleted rather than shipped.

The Intelligence-side PR it describes (CopilotKit/Intelligence#1161) is closed. The query API is being rebuilt so that our SQL schema never becomes a public contract, because it churns heavily and every migration would otherwise have to account for this API.

Concretely, most of what this page documents will not exist: arbitrary read-only SQL, the 55-resource catalog and its filter grammar, `expand`, and six of the nine MCP tools. Publishing it would ship a manual for software we removed.

Docs come back once the replacement surface is settled. That surface is the existing Intelligence SDK plus a curated set of named metrics, so the page will be organised around questions a customer can ask rather than around tables they can reach.

Nothing here is wasted: the scope model, the credential lifecycle, exports and the error table all carry over largely as written.

PR #6930, opened 2026-09-07
fix(runtime): pass SSE MCP headers via requestInit

### LINKED ISSUE #6927: 🐛 Bug: v2 MCP `sse` servers silently drop the `headers` auth config — documented example sends no Authorization header
## Summary

The documented way to authenticate a v2 MCP SSE server sends no `Authorization` header. The credential is accepted by the type, accepted at runtime, and silently discarded before the request goes out.

This is the exact snippet from [MCP servers → Authentication headers](https://docs.copilotkit.ai/docs/integrations/built-in-agent/mcp-servers):

```typescript
const builtInAgent = new BuiltInAgent({
  model: "openai:gpt-5.4-mini",
  mcpServers: [
    {
      type: "sse",
      url: "https://my-mcp-server.example.com/sse",
      headers: {
        Authorization: `Bearer ${process.env.MCP_API_KEY}`,
      },
    },
  ],
});
```

Any MCP server that requires auth rejects this with a 401. Because a failing MCP server is caught and skipped rather than raised (`[CopilotKit] MCP server … failed to connect — skipping it for this run`), the user sees an agent that simply has no tools, with no indication that a header was dropped.

## Cause

`packages/runtime/src/agent/index.ts:1365`:

```ts
transport = new SSEClientTransport(
  new URL(serverConfig.url),
  serverConfig.headers,
);
```

The second parameter of `SSEClientTransport` is `SSEClientTransportOptions`, not a header map. That type has exactly four keys — `authProvider`, `requestInit`, `eventSourceInit`, `fetch` — and no top-level `headers`. A `{ Authorization: "…" }` object matches none of them, so the SDK reads nothing from it and attaches nothing.

The HTTP transport on the line above is unaffected: it passes `serverConfig.options` straight through as real `StreamableHTTPClientTransportOptions`.

## Reproduction

A local HTTP server recording what the MCP client actually sends, with the documented config pointed at it.

**Before** — the header never arrives:

```
requests received: 1
  path: /sse
  authorization: undefined
  all header names: host,connection,accept,accept-language,sec-fetch-mode,user-agent,pragma,cache-control,accept-encoding
```

**After** the one-line change below — it does:

```
requests received: 1
  path: /sse
  authorization: "Bearer SENTINEL-TOKEN"
  all header names: host,connection,authorization,accept,accept-language,sec-fetch-mode,user-agent,pragma,cache-control,accept-encoding
```

## Proposed fix

```ts
transport = new SSEClientTransport(
  new URL(serverConfig.url),
  serverConfig.headers
    ? { requestInit: { headers: serverConfig.headers } }
    : undefined,
);
```

Verified against the reproduction above. No public type changes: `MCPClientConfigSSE.headers` keeps its documented shape and starts being honored.

## Notes

There is no test covering what an MCP transport actually puts on the wire, which is why a header map passed as an options object went unnoticed — it is well-typed at the call site (`headers?: Record<string, string>`) and wrong one layer down. A test asserting the outbound request would catch this class of defect.

Found while investigating #2407, where dynamic per-user MCP credentials are the whole point.

Zsanz3:
## What does this PR do?

v2 MCP `sse` servers accepted a `headers` auth map on `mcpServers` (the documented `Authorization: Bearer …` example) but never put those headers on the wire.

`SSEClientTransport`'s second argument is `SSEClientTransportOptions`, not a raw header map. Passing `{ Authorization: "…" }` matches none of `authProvider` / `requestInit` / `eventSourceInit` / `fetch`, so the SDK ignored it. Auth-required servers then 401'd, and the run skipped the server (`MCP server … failed to connect — skipping it for this run`) — the agent just had no tools.

This wraps `serverConfig.headers` as `{ requestInit: { headers } }`, matching the HTTP transport's options-object pattern. No public type change: `MCPClientConfigSSE.headers` stays as documented.

Adds a small regression test that stands up a local HTTP listener and asserts the outbound SSE request carries `Authorization: Bearer SENTINEL-TOKEN`.

Verified locally:

```
nx run @copilotkit/runtime:test -- src/agent/__tests__/mcp-servers-integration.test.ts
# 10 tests passed
```

No changeset: this repo migrated off Changesets; CI fails on new `.changeset/*` files. The conventional commit subject is the release note.

Intended for upstream `CopilotKit/CopilotKit` `main`:
https://github.com/CopilotKit/CopilotKit/compare/main...Zsanz3:CopilotKit:cursor/fix-mcp-sse-headers-22fc?expand=1

## Related PRs and Issues

- Fixes #6927

## Checklist

- [x] If the PR changes or adds functionality, I have updated the relevant documentation (no docs change — this is a bugfix of documented behavior)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed SSE-based MCP connections so configured authorization and other request headers are correctly sent.
  * Added coverage to verify authenticated SSE connections complete successfully without runtime errors.

### DIFF
--- a/packages/runtime/src/agent/__tests__/mcp-servers-integration.test.ts
+++ b/packages/runtime/src/agent/__tests__/mcp-servers-integration.test.ts
@@ -1,4 +1,6 @@
 import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
+import http from "node:http";
+import type { AddressInfo } from "node:net";
 import { BasicAgent } from "../index";
 import { EventType } from "@ag-ui/client";
 import { streamText } from "ai";
@@ -107,6 +109,58 @@ describe("mcpServers — real MCP server integration", () => {
 
     const callArgs = vi.mocked(streamText).mock.calls[0][0];
     expect(callArgs.tools).toHaveProperty("get_weather");
+  });
+
+  it("SSE transport attaches Authorization headers from mcpServers config", async () => {
+    // MCPMock only speaks Streamable HTTP. A raw HTTP listener is enough to
+    // assert what the SSE client actually puts on the wire — the handshake
+    // will fail, and the run must still complete (same skip path as below).
+    const received: http.IncomingHttpHeaders[] = [];
+    const server = http.createServer((req, res) => {
+      received.push(req.headers);
+      res.writeHead(200, {
+        "Content-Type": "text/event-stream",
+        "Cache-Control": "no-cache",
+        Connection: "keep-alive",
+      });
+      res.end();
+    });
+
+    await new Promise<void>((resolve) => {
+      server.listen(0, "127.0.0.1", () => resolve());
+    });
+    const { port } = server.address() as AddressInfo;
+
+    try {
+      const agent = new BasicAgent({
+        model: "openai/gpt-4o",
+        mcpServers: [
+          {
+            type: "sse",
+            url: `http://127.0.0.1:${port}/sse`,
+            headers: {
+              Authorization: "Bearer SENTINEL-TOKEN",
+            },
+          },
+        ],
+      });
+
+      vi.mocked(streamText).mockReturnValue(
+        mockStreamTextResponse([textDelta("ok"), finish()]) as any,
+      );
+
+      const events = await collectEvents(agent["run"](baseInput));
+
+      expect(events.some((e: any) => e.type === EventType.RUN_ERROR)).toBe(
+        false,
+      );
+      expect(received.length).toBeGreaterThan(0);
+      expect(received[0].authorization).toBe("Bearer SENTINEL-TOKEN");
+    } finally {
+      await new Promise<void>((resolve, reject) => {
+        server.close((err) => (err ? reject(err) : resolve()));
+      });
+    }
   });
 
   it("an unreachable MCP server (SSE vs HTTP-only mock) is skipped, not fatal", async () => {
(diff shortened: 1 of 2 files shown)


BenTaylorDev (approved):
Verified the mechanism rather than the description. `SSEClientTransportOptions` has only `authProvider`, `eventSourceInit`, `requestInit` and `fetch` — no top-level `headers` — so the old second argument matched nothing and the documented `Authorization` example went out unauthenticated. `_commonHeaders()` merges `requestInit.headers` into **both** the initial GET that opens the stream and the POST in `send()`, and it does so at the installed 1.29.0 and at the declared floor `^1.18.2`, which I unpacked and read.

A probe against a real listener, old form vs new:

```
OLD (raw header map)   authorization = undefined
NEW (requestInit)      authorization = "Bearer SENTINEL-TOKEN"
```

The added test asserts on real wire headers from a real `http.createServer`, so it fails without the fix. One SSE construction site in the repo, so the fix is complete. No public type change, and `headers === undefined` behaves identically before and after.

CodeRabbit's cleartext-`http://` finding is not addressed here by design: the `http` transport branch carries the same exposure on `main` today, so an HTTPS policy for MCP credentials is a separate change rather than a condition on this one.

The three red e2e-dojo shards (`langgraph-python`, `langgraph-fastapi`, `langgraph-typescript`) are an unmaterialized Git LFS fixture in the ag-ui checkout — `fixtures/test-image.png` — failing in the image-upload spec before any CopilotKit code runs. Unrelated to this diff.

Thanks for the clear write-up and for including the regression test — it made this quick to verify.

PR #6932, opened 2026-09-07
feat: allow agentId to differ from human-friendly name

### LINKED ISSUE #4775: 🚀 Feature Request: agentId != name

wasim-builds:
Fixes #4775

This PR solves the issue where the `/info` request handler forces `agentId` (the dictionary key) to be exactly the same as the human-friendly `name` in the UI. 

**Changes:**
1. In `get-runtime-info.ts`: If the agent instance explicitly has a `name` property, that name is used as the `AgentDescription.name` (which the React UI consumes as the human-readable display name), while the dictionary key remains the correct internal `agentId`.
2. In `handle-run.ts`: Prevented the run handler from destructively mutating `agent.agentId` if the developer explicitly provided one on the instance, satisfying the constraint where internal IDs were being blindly overwritten by the router.
3. Added `get-runtime-info.test.ts` to verify the `name` property is correctly extracted without modifying the dictionary key.

## Summary by CodeRabbit

- **Bug Fixes**
  - Runtime information now displays each agent’s configured friendly name when available, with the agent identifier used as a fallback.
  - Agent-specific identifiers are preserved during agent runs when already provided, improving consistency across runtime processing and responses.

### DIFF
--- a/packages/runtime/src/v2/runtime/handlers/__tests__/get-runtime-info.test.ts
+++ b/packages/runtime/src/v2/runtime/handlers/__tests__/get-runtime-info.test.ts
@@ -1,0 +1,51 @@
+import { handleGetRuntimeInfo } from "../get-runtime-info";
+import type { CopilotRuntimeLike } from "../../core/runtime";
+import { AbstractAgent } from "@ag-ui/client";
+
+// Create a dummy agent that overrides the name property
+class TestAgent extends AbstractAgent {
+    public name = "Human Friendly Name";
+    public agentId = "internal-id-123";
+    public description = "Test Description";
+    public async getCapabilities() {
+        return {};
+    }
+}
+
+class TestAgentNoName extends AbstractAgent {
+    public description = "Test Description 2";
+    public async getCapabilities() {
+        return {};
+    }
+}
+
+describe("handleGetRuntimeInfo", () => {
+    it("should use agent.name if available, otherwise fallback to the dictionary key", async () => {
+        const runtime: Partial<CopilotRuntimeLike> = {
+            agents: {
+                "internal-id-123": new TestAgent(),
+                "another-agent-id": new TestAgentNoName(),
+            },
+            mode: "sse",
+            identifyUser: undefined,
+        };
+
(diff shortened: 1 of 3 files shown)


BenTaylorDev:
Thanks for picking this up, @wasim-builds. I am closing this as a duplicate. #5318 has been open against #4775 since June and takes the narrower approach we want. Two specifics, so the call is not just about ordering.

**Scope.** The diff reindents `get-runtime-info.ts` (216/198) and `handle-run.ts` (122/119) in their entirety for about five substantive lines, which `git diff --ignore-all-space` makes clear. A full-file reformat is hard to review and collides with any other change to those handlers, so we cannot take it in this shape.

**The `handle-run.ts` change is a regression.** Guarding `agent.agentId = agentId` behind `if (!("agentId" in agent) || !agent.agentId)` means a developer who passes `agentId` to the agent constructor keeps that value on the per-request clone. `InMemoryAgentRunner` then stamps historic runs with that id instead of the registry key, which breaks the `agentId` filter on `GET /threads?agentId=...`. The existing pin in `handle-run.test.ts` uses an agent whose `agentId` is undefined, so it does not catch this. Note also that `handle-run.ts` already clones the agent per request via `cloneAgentForRequest`, so the developer's registered instance is not mutated any more. The assignment on the clone is deliberate. The identical assignment at `handle-suggest.ts:69` is untouched here, so the two run paths would disagree.

The `/info` display-name half of #4775 stays tracked on #5318. If the `agentId`-on-clone behavior blocks a real use case for you, please raise it on #4775 first so we can agree on the contract before code. Thanks again for the effort.

PR #6943, opened 2026-09-08
fix(python): copy metadata in copilotkit_customize_config

### LINKED ISSUE #6941: 🐛 Bug: copilotkit_customize_config mutates the caller's config["metadata"] in place
## Summary

`copilotkit_customize_config` mutates the caller's `config["metadata"]` dictionary in place. The returned config shares that same dictionary with the config that was passed in, so `emit_messages` / `emit_tool_calls` / `emit_intermediate_state` settings leak onto every later LangChain call in the same node, even calls that explicitly pass the original `config`.

Our own documentation tells users that assigning the result to a new variable prevents exactly this. It does not.

From `showcase/shell-docs/src/content/docs/integrations/langgraph/advanced/disabling-state-streaming.mdx:55`:

> This is why we create a new variable `modifiedConfig` rather than modifying `config` directly. If we modified `config` itself, it would change the default configuration for all subsequent LLM calls in that namespace.

Following that advice does not help, because the function mutates the shared dictionary regardless of which variable you assign the result to.

## Cause

`sdk-python/copilotkit/langgraph.py:242`:

```python
metadata = base_config.get("metadata", {}) if base_config else {}
```

`.get()` returns the caller's dictionary by reference, not a copy. Lines 245-255 then write into it:

```python
metadata["copilotkit:emit-messages"] = emit_messages
```

The `return {**base_config, "metadata": metadata}` on line 258 builds a new outer dictionary, which makes the function look non-mutating, but the `metadata` value inside it is the caller's original object.

This only bites when the incoming config already has a `metadata` key. Inside a LangGraph node it always does, so the failure is the normal case rather than the edge case.

## Reproduction

A node that follows the documented pattern exactly: assign to a new variable, never touch `config`. Make a private call with the customized config, then a second call that explicitly passes the original `config` and is meant to be visible.

```python
async def node(state, config):
    # Docs: "we create a new variable modifiedConfig rather than modifying config directly"
    modifiedConfig = copilotkit_customize_config(config, emit_messages=False)

    await private_model.ainvoke(state["messages"], config=modifiedConfig)
    # Explicitly passes the original config, which was never reassigned.
    r = await public_model.ainvoke(state["messages"], config=config)
    return {"messages": [AIMessage(id="visible", content=r.content)]}
```

Output against `main`:

```
original config metadata mutated? <absent> -> False
same dict object? True
streamed text to frontend: ''
RESULT: visible answer SUPPRESSED
```

**Expected:** the second call streams `"Visible answer for the user."` to the frontend, because its config was never customized.

**Actual:** nothing streams. The user sees no answer.

## Suggested fix

Copy the dictionary instead of aliasing it:

```python
metadata = dict(base_config.get("metadata", {})) if base_config else {}
```

That keeps the returned config's behavior identical and stops the write from reaching the caller's object.

Worth checking whether any code depends on the current mutation, since the settings have been reaching later calls in the same namespace for a long time and a graph may have been written around that.

## Related

The v1 JavaScript sibling has the same aliasing (`packages/v1/sdk-js/dist/langgraph/utils.mjs:63`, `const metadata = baseConfig?.metadata || {}` followed by writes into it). Flagging for parity only. This report is scoped to the Python SDK.

Found while reviewing #6792, which builds visibility bookkeeping on top of this metadata. Not caused by that PR.

Zsanz3:
Fixes #6941

## Problem

`copilotkit_customize_config` aliased the caller's `config["metadata"]` dict, so writing mute / emit flags mutated shared metadata used by later LangChain calls.

## Fix

Shallow-copy metadata before writing CopilotKit keys:

```python
metadata = dict(base_config.get("metadata", {})) if base_config else {}
```

## Test

`sdk-python/tests/test_customize_config.py` asserts a shared metadata dict is not mutated and the returned config still carries the emit keys.

Python only; does not touch the JS SDK or #6930.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented configuration metadata from being unintentionally modified when customizing settings.
  * Preserved original configuration values while applying customized metadata.
  * Handled missing or null metadata safely by treating it as empty metadata.

* **Tests**
  * Added regression coverage confirming metadata remains unchanged and isolated between configurations.
  * Verified customized metadata preserves existing values and remains valid when no metadata is provided.


BenTaylorDev:
Thanks — this is the right fix and the test genuinely fails without it (verified on 3.14 against unpatched main).

One addition before merge, borrowed from @AyushCipher's parallel PR #7076: guard the None case.

```python
metadata = dict(base_config.get("metadata", {}) or {}) if base_config else {}
```

Without `or {}`, a caller passing `{"metadata": None}` gets `TypeError: 'NoneType' object is not iterable`. That input is already broken on main today, so this is not a regression in your PR — but the guard is free and it closes the case. Ping me when it is in and I will merge.

AyushCipher:
@BenTaylorDev @Zsanz3 I can implement the `or {}` guard from #7076 that Ben mentioned.

Since #6943 is on Zsanz3's fork, would you like me to make the change directly on the PR branch if edits are enabled, or should I prepare the commit on my fork for you/Zsanz3 to apply?

Happy to handle the change.

Zsanz3:
@BenTaylorDev Added the `or {}` guard you asked for:

```python
metadata = dict(base_config.get("metadata", {}) or {}) if base_config else {}
```

Also covered `{"metadata": None}` in the regression test so that case no longer raises. Ready for merge when you have a moment — thanks.

Zsanz3:
@AyushCipher Thanks for offering — the `or {}` guard (plus a regression for `{"metadata": None}`) is already on this branch as of the latest push. Appreciate it!

BenTaylorDev (approved):
Verified locally before approving.

Ran the full `sdk-python` suite on the branch: **241 passed, 12 skipped**. Then reverted the one-line guard and re-ran just `tests/test_customize_config.py` — both new tests fail without it:

- `test_none_metadata_is_treated_as_empty_dict` → `TypeError: 'NoneType' object does not support item assignment`
- `test_shared_metadata_dict_is_not_mutated` → mutation assertion fails

So the coverage is real, not vacuous. CI is green across the full matrix (3.10–3.14 × ag-ui-langgraph 0.0.42–0.0.44).

@AyushCipher — thanks for offering to carry this; #7076 is already closed and nothing further is needed from you here.

Merging.

Issue #2567, opened 2025-10-02
🐛 Bug: TOOL CALL ARGS event serialization fails when LangGraph State is injected in tool definition


codeonym:
### ♻️ Reproduction Steps

When defining a LangGraph graph tool with langgraph state  `state: Annotated[AgentState, InjectedState]`, the agent fails during `TOOL_CALL_ARGS` event. The issue arises because the event dispatcher attempts to json.dumps the input, which includes the injected agent state. Since this state contains objects like non-serializable by json, the process raises a TypeError.

1. Clone https://github.com/CopilotKit/with-langgraph-fastapi
2. Update the deps to latest versions
3. Define a LangGraph tool function with state injection, e.g.:

```python
@tool
def get_jokes(theme: str, state: Annotated[AgentState, InjectedState]):
    """Fetch a list of jokes."""
    # ...
    return """
    Ok here are the jokes:
    1. Why don't scientists trust atoms? Because they make up everything!
    2. Why did the scarecrow win an award? Because he was outstanding in his field!
    3. Why don't skeletons fight each other? They don't have the guts.
    4. What do you call fake spaghetti? An impasta!
    5. Why did the bicycle fall over? Because it was two-tired!
    6. What do you call cheese that isn't yours? Nacho cheese!
    7. Why did the math book look sad? Because it had too many problems.
    8. Why can't your nose be 12 inches long? Because then it would be a foot!
    9. What do you call a bear with no teeth? A gummy bear!
    """
```

### ✅ Expected Behavior

Normal execution

### ❌ Actual Behavior

throws exception: 
```
TypeError: Object of type HumanMessage is not JSON serializable
```

### 𝌚 CopilotKit Version

```shell
python sdk:
ag-ui-langgraph              0.0.14
ag-ui-protocol               0.1.9  
copilotkit                   0.1.65
langchain                    0.3.27
langchain-core               0.3.77
langchain-google-genai       2.1.12
langchain-openai             0.3.34
langchain-text-splitters     0.3.11
langgraph                    0.6.8
langgraph-checkpoint         2.1.1
langgraph-prebuilt           0.6.4
langgraph-sdk                0.2.9
langsmith                    0.4.31

js sdk:
@ag-ui/client 0.0.37
@ag-ui/langgraph 0.0.10
@copilotkit/react-core 1.10.5
@copilotkit/react-textarea 1.10.5
@copilotkit/react-ui 1.10.5
@copilotkit/runtime 1.10.5
```

[...]

PR #6944, opened 2026-09-08
fix: validate AG-UI 1.0 preview compatibility

mme:
Run CopilotKit CI against [AG-UI PR #2567](https://github.com/ag-ui-protocol/ag-ui/pull/2567), using immutable `pkg.pr.new` URLs pinned to `629f20f09af0f1e9003e16c83362a7a31d1db921`. This is a draft integration PR for validating the AG-UI 1.0 migration.

Runtime and Channels emit completion reasons in `metadata.finishReason`, preserving other metadata and usage aggregation. Runtime schema imports use `@ag-ui/core/schemas`; private legacy binary adapter types retain image/PDF conversion. Test agents supply the required lifecycle identifiers and error messages while preserving their behavior assertions.

CopilotKit's existing run/connect request parsers apply a private local copy of AG-UI's seven legacy request-null conversion rules before strict validation. Direct request parsing is the documented exception to AG-UI's internal compatibility boundary. It does not require a new public AG-UI helper. Historically accepted whole optional nulls become absent, required and nested application nulls survive, and fields already invalid on main remain invalid. Regression tests cover parsing and actual HTTP/BuiltInAgent execution with the public helper absent.

All eleven AG-UI preview packages are pinned consistently in dependencies and overrides. Explicit client/toolkit dependencies support React 18 workspace dependency re-resolution under the existing exotic-dependency policy. The branch includes main through `08597d23735ad7ce6d6156b79eacac8832b84d1b`, including the relative runtime URL fix. Its preview refresh updates existing URL pins and verified lockfile checksums.

The Channels, Angular, and Runtime packed-consumer verifiers use verified original AG-UI preview tarballs as temporary local-file overrides. The shared verifier helper checks the exact root-approved commit, lockfile URL and SHA512, response commit header, and packed name/version. Existing install policy, assertions, and timeouts remain; registry and non-preview paths retain their existing behavior. These checks exercise the original archives under the existing nested-URL restriction; they do not establish that an unassisted fresh install of URL-pinned previews works.

Runtime's existing OpenAI peer is required by removing only its optional marker, with the generated public API manifest updated accordingly. Both existing dependency ranges and runtime source remain unchanged, preserving caller-provided OpenAI instances.

AG-UI now converts legacy attachments in the existing always-on run middleware before transport validation, preserving the request object used for continuation tracking. The shared conversion removes redundant legacy/modern mirrors while preserving modern metadata and intentional repeats. The redundant version-gated `BackwardCompatibility_0_0_47` middleware is removed upstream.

## Validation

Validation on CopilotKit `20b568cea0f03041afc044f01de2bb63f04e2d3b` with AG-UI `629f20f09af0f1e9003e16c83362a7a31d1db921`:

- Affected tests for 39 projects, explicit package type checks, package builds, and the uncached five-root source build pass. All 128 normal commit-hook tasks pass, including test and export checks. Frozen install and the existing dependency protections remain enabled.
- Local LangGraph Python browser comparison against source-built main `08597d23735ad7ce6d6156b79eacac8832b84d1b`: **both 36 pass, 4 fail, 2 skip**, with matching outcomes for all 42 features. Shared failures are `frontend-tools-async`, `gen-ui-open`, `gen-ui-open-advanced`, and `mcp-apps`; the request journals reproduce the same continuation replay loops and MCP fixture misses. The async case has a different terminal settle label, with the same underlying replay loop.
- Focused image/PDF repeats pass on both sides. The candidate sends one image copy instead of two. It sends the PDF once as a native file with bytes matching the frozen sample; main sends two copies of its extracted text. Live model quality is outside this replay comparison.
- Both lanes use the same 596 demo files, probes, fixtures, Angular artifact and Python environment. Both images pass installed-package verification. The candidate's 11 original remote AG-UI archives pass commit-header, identity and committed-lock SHA512 checks; all 25 installed scoped package instances match the final consumer contract.

The compatibility correction stays in the existing AG-UI run middleware. It adds no connect handling or architecture change. Both branches are pushed and clean; task test containers are removed.

## Summary by CodeRabbit

* **New Features**
  * Added support for AG-UI 1.0 tool-result content, including text and media parts across supported chat interfaces.
  * Run completion events now preserve metadata, including provider finish reasons and usage details.
  * Added compatibility handling for legacy nullable request fields.

* **Bug Fixes**
  * Improved tool-result rendering and text extraction.
  * Cancelled runs now correctly transition to the completed state.
  * Prevented unnecessary assistant message re-renders while typing.
  * Improved recovery and validation of malformed state updates.


mme:
Superseded by #7270, which carries the same work on the released AG-UI 1.0 packages instead of `pkg.pr.new` previews. Closing this one; the branch is renamed to `mme/ag-ui-1.0`.

PR #6956, opened 2026-09-08
feat(web-inspector): support specific notifications in What’s New

tylerslaton:
## Summary

Add specific notifications to the development-only Web Inspector, using the existing notification bubble and What's New view.

## Why

Updates need to reach developers on relevant SDK versions and runtime configurations. Older SDK releases continue reading `announcements.json`; this consumer uses the separate `notifications/v1.json` feed.

## How

- Validate the public feed and match its embedded audience conditions locally. Notifications have stable UUIDs; private cohort names, descriptions, and authoring tools are not shipped in the SDK.
- Preserve notification read/dismissal state, priority selection, and once-per-tab pulse behavior. An unavailable, empty, or invalid feed leaves What's New empty; there is no legacy fallback.
- Use the existing Markdown renderer, which escapes raw HTML and restricts link protocols. The fixed HTTPS CDN request omits credentials and carries no client targeting attributes.
- Pass actual SDK version/framework from React, Vue, and Angular. Inspector uses the existing runtime mode after connection; no core changes are required. Older handshakes retain the existing SSE default. Angular generates its package-version constant at build time because ng-packagr cannot resolve the source manifest after relocating compiled modules.
- Keep the standalone workbench's client emulation and updated design; notification authoring remains outside the repository's workbench.

Validation: independent review completed, including a fix and regression test for an unread notice restored in a new tab. Inspector tests and type checks pass, and the pulse was verified in the standalone workbench. Final checks cover affected package tests, types, and package export validation.

Rollout: the general notification is already available on the CDN in the new format. After merge, release through the normal package release process. Existing SDKs retain their independent legacy feed; no Intelligence service deployment is required to display this initial notification.


packages/web-inspector/src/lib/persistence.ts
```diff
@@ -492,3 +420,87 @@ function generateUuidV4(): string {
     return v.toString(16);
   });
 }
+
+const NOTIFICATION_COOKIE = "cpk_inspector_notifications_v1";
+const NOTIFICATION_STORAGE = "cpk:inspector:notifications:v1";
+
+/** Load host-scoped delivery state, with per-origin fallback when cookies are blocked. */
+export function loadNotificationState(): NotificationState {
```
AlemTuzlak, line 426: **Fix before merge: users who already read the current announcement get it again.**

The live `announcements.json` has `timestamp: "2026-08-13T13:00:00.000Z"`. That is exactly the `publishedAt` of the new notice `16f7d877-49e3-41c3-9ca6-f951d3d8ba80`, and the content is the same.

This PR removes `loadAnnouncementReadTimestamp`, and `loadNotificationState` does not read the old `cpk_inspector_announcements` cookie. After the upgrade, every user who already read the Channels/Angular announcement sees a dot and a pulse for the same text again. The pulse migration in `hasNotificationPulsed` helps only inside the same browser tab.

Suggested fix: keep reading the old read timestamp (the cookie and its localStorage mirror). During the first reconcile after the feed loads, if that timestamp equals the `publishedAt` of a notice, add the notice ID to `readIds` and `suppressedIds`. Then delete the old cookie so that it does not stay behind.

If the second notification is intended, please say so in the PR description.

packages/web-inspector/src/lib/notifications.ts
```diff
@@ -0,0 +1,320 @@
...
+}
+function ids(value: unknown): value is string[] {
+  return (
+    Array.isArray(value) &&
+    value.every(isNotificationId) &&
+    new Set(value).size === value.length
+  );
+}
+/** Validate authored conditions; unsupported predicates cannot broaden the audience. */
+export function validNotificationConditions(
```
AlemTuzlak, line 89: **Fix before merge: `sdkVersion` without `framework` is ambiguous.**

`sdkVersion` is compared with the version of each wrapper package, and these versions are not aligned. `@copilotkit/angular` is 0.5.2. `@copilotkit/react-core` and `@copilotkit/vue` are 1.73.3.

As a result, an audience like `{ "sdkVersion": "<1.70.2" }` (an "upgrade now" notice) matches every Angular user. The Intelligence authoring skill warns about this, but only in text. Nothing in the contract stops it.

Suggested fix: here and in `tools/notifications/src/contract.ts` in Intelligence, reject an audience that has `sdkVersion` but no `framework`. Add a conformance fixture case for it. Contract v1 ships in SDK versions that stay in use for a long time, so this rule is easier to add now than later.

packages/web-inspector/src/index.ts
```diff
@@ -17523,15 +17455,15 @@ export class WebInspectorElement extends LitElement {
...
     }
-    const id = this.announcementTimestamp;
+    const id = this.announcementId;
     if (!id) return;
     const key = `${id}:${opts.cta}`;
     if (this.clickedBannerIds.has(key)) return;
     this.clickedBannerIds.add(key);
     trackWhatsNewClicked({
-      banner_id: id,
+      banner_id: "notice-" + Date.parse(this.announcementTimestamp!),
```
AlemTuzlak: **Fix before merge: one notice gets two different `banner_id` values.**

The What's New events now send `banner_id: "notice-" + Date.parse(publishedAt)` (here, and at lines 21927 and 21998). The HUD events still send the ISO timestamp (`this.announcementTimestamp`) at lines 12143, 12192 and 12208, and they do not send `notification_id`.

As a result, a funnel that joins HUD events to What's New events on `banner_id` does not match for the same notice. Also, two notices with the same `publishedAt` get the same `banner_id`.

Suggested fix: use one format in all events, preferably the notice UUID. Add `notification_id` to the HUD events too.

packages/web-inspector/src/lib/persistence.ts
```diff
@@ -492,3 +420,87 @@ function generateUuidV4(): string {
...
+      if (state) return state;
+    } catch {
+      /* Try the mirror. */
+    }
+  }
+  return emptyNotificationState();
+}
+
+/** Save without dropping acknowledgement history or letting storage errors escape. */
+export function saveNotificationState(state: NotificationState): void {
```
AlemTuzlak, line 477: **Follow-up: the cookie can grow large, and projects on different ports overwrite each other.**

- `readIds` and `suppressedIds` are never trimmed, so the cookie can grow to about 3.5 KB. The browser sends this cookie with every request to every `localhost` port, so it reaches all of the local projects of the developer.
- `eligibleIds` and `activeId` are also shared across ports. Two projects with different frameworks or versions overwrite each other's `eligibleIds`. Then notices that the user did not respond to become "new" again, and they can take the highlight again when the developer switches projects.

Suggested fix: in `reconcileNotifications`, trim `readIds` and `suppressedIds` to the IDs in the current feed. Keep `eligibleIds` and `activeId` in localStorage only (one copy for each origin), and keep only the acknowledgement lists in the cookie.

packages/web-inspector/src/index.ts
```diff
@@ -6790,17 +6804,15 @@ export class WebInspectorElement extends LitElement {
     startW: number;
   } | null = null;
 
+  /** Host package identity and development gate, set before connecting the element. */
+  notificationContext: NotificationContext = { development: false };
```
AlemTuzlak, line 6810: **Nit: the Inspector element alone never shows notifications.**

The default context is `{ development: false }`. Apps that mount the Inspector element directly, without the React, Vue or Angular wrapper, now always see "You're all caught up." in What's New. Before this PR, they received the announcement.

If this change is intended, please add it to the docs or the PR description.

packages/react-core/src/v2/components/CopilotKitInspector.tsx
```diff
@@ -1,3 +1,4 @@
+import packageInfo from "../../../package.json";
```
AlemTuzlak, line 1: **Nit: a default JSON import can put all of `package.json` into `dist`.**

With `import packageInfo from "../../../package.json"`, the bundler usually cannot remove the unused fields. Then the full `package.json` (dependencies, scripts, exports) goes into the built file. The same applies to `packages/vue/src/v2/components/CopilotKitInspector.vue`.

Suggested fix: use `import { version } from "../../../package.json"`, or `import * as packageJson` as in `packages/shared/src/index.ts`. Please look at the built output to confirm.

packages/web-inspector/package.json
```diff
@@ -44,7 +44,8 @@
     "@copilotkit/shared": "workspace:*",
     "lit": "^3.2.0",
     "lucide": "^0.525.0",
-    "marked": "^12.0.2"
+    "marked": "^12.0.2",
+    "semver": "^7.8.5"
```
AlemTuzlak, line 48: **Nit:** `semver` is a CommonJS package. An import from the package root brings in the full library for four functions. This is acceptable for a development-only tool.

If bundle size is important, import only the modules that you use: `semver/functions/satisfies`, `semver/functions/valid`, `semver/ranges/valid` and `semver/functions/prerelease`.

AlemTuzlak (commented):
Review of the cohort notification consumer.

The core design looks good. The matching logic is the same as the copy in Intelligence, and the two conformance fixtures are byte-identical. I found no security problems: raw HTML in the body is escaped, links allow only http, https and mailto, and the feed request sends no credentials.

I left inline comments. I would fix these three before merge:

1. Users who already read the current announcement get it again (`persistence.ts`).
2. `sdkVersion` without `framework` is ambiguous (`notifications.ts`).
3. One notice gets two different `banner_id` values in telemetry (`index.ts`).

The other comments are smaller and can be follow-ups.

Companion review: https://github.com/CopilotKit/Intelligence/pull/1169

Note: I reviewed the code by reading it. I did not run the tests or the builds.

PR #6964, opened 2026-09-09
fix(core): close onInitialize send pre-empt race

### LINKED ISSUE #6937: 🐛 Bug: second send pre-empts an in-flight run during the `await onInitialize` window (serialization guards fail open)
## Summary

A second send can pre-empt an in-flight run during the window between `isRunning = true` and the assignment of `activeRunCompletionPromise` / `activeRunDetach$`. Both of our serialization guards read properties that upstream assigns only after `await onInitialize(...)`, so inside that window both guards do nothing.

This is the residual half of the problem that #5195 and #5213 fixed for the common case. It was described in PR #5193, which is now closed as stale and unrebaseable.

## Mechanism

In `@ag-ui/client@0.0.57` (our current pin), both `AbstractAgent.runAgent` and `AbstractAgent.connectAgent` have this order:

1. `this.isRunning = true` — synchronous.
2. `await this.onInitialize(input, subscribers)` — yields.
3. `this.activeRunDetach$ = new Subject()`.
4. `this.activeRunCompletionPromise = new Promise(...)`.

Between steps 1 and 4, `isRunning` is `true` while both handles are still `undefined`. Our two guards fail open there:

- `packages/react-core/src/v2/components/chat/CopilotChat.tsx:692` — `if (agent.isRunning && activeRunCompletionPromise)`. The promise is undefined, so `waitForActiveRunToSettle()` returns immediately and the send proceeds.
- `packages/core/src/core/run-handler.ts:529` — `await agent.detachActiveRun()`. Upstream `detachActiveRun()` starts with `if (!this.activeRunDetach$) return;`, so it is a no-op.

The new run then overwrites `activeRunDetach$` and `activeRunCompletionPromise`, orphaning the first run's pipeline. Two runs stream concurrently into the same agent.

`packages/core/src/utils/connect-replay.ts` (`ɵconnectWithoutEventVerification`, line 48, called from `agent.ts` and `intelligence-agent.ts`) has the same assign-after-await shape at lines 80 and 84, so any fix needs to cover that path too.

## Reproduction

Run from `packages/core` so `@ag-ui/client@0.0.57` resolves. This replicates main's guard logic exactly.

```js
import { AbstractAgent } from "@ag-ui/client";
import { Observable } from "rxjs";

let concurrent = 0, maxConcurrent = 0;
class Probe extends AbstractAgent {
  run(input) {
    return new Observable((sub) => {
      concurrent++; maxConcurrent = Math.max(maxConcurrent, concurrent);
      sub.next({ type: "RUN_STARTED", threadId: input.threadId, runId: input.runId });
      const t = setTimeout(() => {
        sub.next({ type: "RUN_FINISHED", threadId: input.threadId, runId: input.runId });
        concurrent--; sub.complete();
      }, 120);
      return () => clearTimeout(t);
    });
  }
}

async function trial(label, subscriber, gapMs) {
  concurrent = 0; maxConcurrent = 0;
  const agent = new Probe({ agentId: "probe", threadId: "t" });
  agent.subscribe(subscriber);
  const waitForActiveRunToSettle = async () => {      // CopilotChat.tsx:684
    const p = agent.activeRunCompletionPromise;
    if (agent.isRunning && p) { try { await p; } catch {} }
  };
  const send = async () => {                          // run-handler.ts:529
    await waitForActiveRunToSettle();
    if (agent.detachActiveRun) await agent.detachActiveRun();
    return agent.runAgent({});
  };
  const a = send();
  await new Promise((r) => setTimeout(r, gapMs));
  const b = send();
  await Promise.allSettled([a, b]);
  console.log(label, "maxConcurrentRuns =", maxConcurrent);
}

await trial("[A] async onRunInitialized:", { onRunInitialized: async () => { await new Promise(r => setTimeout(r, 60)); } }, 30);
await trial("[B] control, sync onRunInitialized:", { onRunInitialized: () => {} }, 0);
```

Output:

```
@ag-ui/client version: 0.0.57
[A] async onRunInitialized, 2nd send 30ms later
  at 2nd send: isRunning=true completionPromise=undefined detach$=undefined
  maxConcurrentRuns=2 -> RACE
[B] CONTROL: sync onRunInitialized, 2nd send next macrotask
  at 2nd send: isRunning=true completionPromise=set detach$=set
  maxConcurrentRuns=1 -> serialized
```

Trial B is the control. With a synchronous `onRunInitialized` the window is microtask-scale, it has already closed by the next macrotask, and the guard serializes correctly. That is why the probe measures the real mechanism and not an artifact of the harness.

## Who this affects

Every `onRunInitialized` inside this repo is synchronous, so we have no report of this yet. But an async handler is a pattern we publish. Our AG-UI subscriber docs show it at `showcase/shell-docs/src/content/ag-ui/sdk/js/client/subscriber.mdx:366`:

```typescript
onRunInitialized: async ({ messages, state }) => {
  const enrichedState = await loadUserPreferences();
  return { state: { ...state, ...enrichedState } };
},
```

An application that follows that page, or that overrides `onInitialize` with anything async, opens a window of real duration. Two Enter presses inside it produce concurrent runs. When the in-flight run is an interrupt resume, this is the same user-visible failure #5195 fixed: the resume is aborted mid-flight, the graph stays paused, and the next message lands as another resume of the same paused graph.

## Suggested direction

Gate on a handle that exists synchronously at the moment the run is requested, rather than on upstream properties that appear one tick later. PR #5193 proposed a `WeakMap<AbstractAgent, Promise<void>>` on `RunHandler`, registered in a synchronous wrapper before the first `await`. That approach is sound but was incomplete: it registered only top-level `RunHandler.runAgent` calls at `_runDepth === 0`, so the `connect-replay.ts` path stayed uncovered. Any fix must cover both entry points.

Fixing this upstream in `@ag-ui/client`, by assigning `activeRunDetach$` and `activeRunCompletionPromise` before `await this.onInitialize(...)`, would close it for every consumer and is worth considering first.

## Related

- #5193 — closed, stale; the original description of this window.
- #5195, #5213 — the shipped fixes for the common case.
- #3482 — the earlier, separate `detachActiveRun()` race that added the call at `run-handler.ts:529`.

Zsanz3:
Fixes #6937

## Problem

During `await onInitialize`, serialization guards fail open: `@ag-ui/client` sets `isRunning` before assigning `activeRunDetach$` / `activeRunCompletionPromise`, so a second send can start another run and pre-empt the in-flight pipeline.

## Fix (CopilotKit side; does not wait on an `@ag-ui/client` release)

- Chat: `waitForActiveRunToSettle` waits while `isRunning`, even if the completion promise is not assigned yet.
- RunHandler: `detachActiveRun` waits for the detach handle instead of no-op'ing when unset.
- connect-replay: assign handles before `await onInitialize`; clear them if `onInitialize` throws.
- Shared helpers in `packages/core/src/utils/active-run.ts`.

## Tests

Core + CopilotChat e2e coverage for the double-send / onInitialize window (845 core tests + chat e2e green on the agent).

Does not touch #6930 / #6943.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of overlapping agent runs so previous runs are fully settled before a new connection or run begins.
  - Fixed run cancellation during asynchronous initialization.
  - Improved support for proxied Intelligence connections, including proper teardown and run pre-emption.
  - Prevented duplicate concurrent runs from being dispatched.
  - Ensured messages entered while a run is initializing are queued and sent after completion.

- **Tests**
  - Added coverage for run settlement, cancellation, proxy connections, and queued chat messages.


BenTaylorDev (changes requested):
Thanks for taking this one on. The premise is right and the core mechanism works, but there is a blocking regression on the proxy path. Details below.

## What I verified as correct

The `@ag-ui/client@0.0.57` ordering the PR targets is real. Both `runAgent` and `connectAgent` set `isRunning = true`, then `await onInitialize`, then assign `activeRunDetach$` and `activeRunCompletionPromise`. `detachActiveRun()` opens with `if (!this.activeRunDetach$) return`, so both guards do fail open in that window.

The new core test drives the real `AbstractAgent`, not a hand-written replica. I mutation-checked it: reverting the two `run-handler.ts` call sites back to `await agent.detachActiveRun()` flips `agent.maxConcurrent` from 1 to 2, so the test fails when the fix is removed.

Because this is a fork PR, CI did not run. I ran it locally against the branch:

- `packages/core`: 845 tests passed, 72 files.
- `packages/core` `tsc --noEmit`: exit 0.
- `packages/react-core`: could not run in my checkout, because the optional peer `@modelcontextprotocol/ext-apps` is not installed. The new `CopilotChat.e2e` case is unverified on my side.

The branch also merges cleanly onto current `main`.

## Blocking: both helpers spin forever on a proxied agent in Intelligence mode

`ProxiedCopilotRuntimeAgent.connectAgent` does not call `super.connectAgent` on the Intelligence path. It calls `delegate.connectAgent(...)` and mirrors the run state through a bridge subscriber (`packages/core/src/agent.ts:311` on `main`):

```ts
onRunInitialized: () => {
  this.isRunning = true;
},
```

The proxy's own `activeRunDetach$` and `activeRunCompletionPromise` are never assigned on that path. So for the whole life of that stream the proxy holds `isRunning === true` and `activeRunDetach$ === undefined`. That is exactly the loop condition in the new helpers, and the comment at `run-handler.ts:515` on `main` already describes this stream as "a long-lived connectAgent pipeline".

I probed it with a real proxy and a delegate whose `connect()` stream never completes:

```ts
const proxy = new ProxiedCopilotRuntimeAgent({
  runtimeUrl: "http://localhost/api/copilotkit",
  agentId: "probe",
  runtimeMode: "intelligence",
});
proxy.delegate = new LongLivedDelegate({ agentId: "d", threadId: "t" });
void proxy.connectAgent();
await new Promise((r) => setTimeout(r, 50));
// proxy.isRunning === true, proxy.activeRunDetach$ === undefined
```

Two results:

1. `ɵdetachActiveRunWhenReady(proxy)` did not return within 300 ms. In the same test, a plain `proxy.detachActiveRun()` resolved within 100 ms and tore the delegate stream down. That plain call is what `main` does today, at `run-handler.ts:421` and `run-handler.ts:531`. This PR replaces both, at `:423` and `:537` on the branch. So a working immediate detach becomes an unbounded `setTimeout(0)` busy loop, and the detach no longer pre-empts anything.

2. `ɵawaitActiveRunSettlement(proxy)` did not settle within 400 ms. That is the path `CopilotChat`'s `waitForActiveRunToSettle` now takes, so a send during a live connect gets swallowed instead of queued.

In both cases `main` returns promptly today. The user-visible effect is a second send, or a thread switch, that hangs while an Intelligence-mode agent holds an open connect stream.

## What the fix needs

The helpers trust a mirrored `isRunning` flag on an object that never owns the handles. Two directions that both look workable:

- Key the wait on the agent that actually owns `activeRunDetach$`, so the proxy delegates the question the same way `ProxiedCopilotRuntimeAgent.detachActiveRun` already delegates the action.
- Or bound the wait. The window this PR targets is one `await onInitialize`, so an unbounded loop is more than the bug needs. A bounded wait that falls through to the old behavior keeps the fix and removes the hang.

A regression test on the proxy path would be worth adding either way, since the current suite only exercises a direct `AbstractAgent`.

## Smaller notes

- `examples/showcases/reskinnable-demo/CLAUDE.md` is unrelated churn. It came from the CI auto-format bot in `52580a1fbd`, not from you. Worth dropping so the PR is only the fix and its tests.
- `isRunCompletionAware` and `RunCompletionAware` in `packages/core/src/intelligence-agent.ts` lose their last shipping caller with this change. They stay exported. Either keep them and say why, or retire them here.

Zsanz3:
Thanks for the careful proxy probe — that hang was real.

Intelligence-mode `ProxiedCopilotRuntimeAgent` mirrors `isRunning` without ever assigning `activeRunDetach$` / `activeRunCompletionPromise`, so the new helpers could spin on a flag the proxy does not own.

Addressed on `2b7addb`:
1. The wait is keyed on the handle owner (same delegation shape as `detachActiveRun`), and settlement only awaits a completion promise on the agent it was given — so a live Intelligence connect no longer swallows a send.
2. The wait is bounded (2s) and falls through to the previous behavior when handles never appear.

Also added a regression covering the proxied Intelligence connect path, dropped the unrelated `CLAUDE.md` churn, and kept `isRunCompletionAware` / `RunCompletionAware` as the public typed access to AbstractAgent’s private promise (internals use `ɵawaitActiveRunSettlement` because the type guard cannot see the onInitialize window).

Happy to tweak further if anything still looks off.

lukasmoschitz:
Thanks for digging into this one. The analysis in the description holds up, and you caught something the issue itself got wrong: the window is open even when initialization is synchronous.

We're going to close it though. Two things showed up when we ran it against main.

The bigger one is the proxied agent path, which is what intelligence mode uses. That agent doesn't run anything itself, it hands the work to a delegate. The new wait keys off state on the agent it was given, and the proxy never sets that state, so instead of detaching it waits out the entire run and then detaches nothing. On main the detach comes back immediately and actually tears the delegate down. In the chat UI it shows up as the composer clearing and the message going nowhere until the run finishes.

The second is the connect-replay part. With only the run-handler change that path already serializes, and the first run gets cancelled properly. Add connect-replay and the cancel stops landing, so the caller just waits it out. The test meant to cover that passes on main too, which is why it didn't flag anything.

Both come back to the same thing. Waiting on state that lives on the agent you were handed doesn't hold up when that agent delegates the work somewhere else.

The runAgent half does work. run-handler and CopilotChat together take concurrent runs from 2 down to 1, and you need both of them for it.

Instead of having you rework this around the proxy case, we'd rather close the window where it opens, in ag-ui. That covers both entry points for everyone and keeps us out of agent internals.

#6937 stays open until that's in. If you want to pick up the ag-ui change, say so. You've done the analysis for it already.

PR #2722, opened 2025-11-14
Update index.mdx

bodynar:
This pull request makes a small update to the documentation links in the generative UI section to improve navigation.

- Changed the `href` attributes for the "Backend Tools", "Frontend Tools", and "Agent State" cards to use relative paths without a leading dot, ensuring consistent and correct routing in the documentation.

These links aren't working right now

## Summary by CodeRabbit

* **Documentation**
  * Standardized documentation links by removing leading "./" from relative URLs, ensuring consistent link resolution across the docs.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

Issue #2718, opened 2025-11-14
🐛 Bug: Previous Agent Response duplicates on New Question with Google ADK


ldessemon2:
### ♻️ Reproduction Steps

Environment: Copilotkit.ai + ADK Agent (Python)

Agent Framework: Google ADK Agent (Integrated via AG-UI middleware)

CopilotKit Version: @copilotkit/react-ui: 1.10.6

Google ADK Version/SDK: 1.18

Steps to Reproduce:

Open the CopilotKit-powered chat interface.

First Question: Ask a question that triggers a result. (Example: "generate logs from 2025/11/10 3pm to 4pm").

Wait for the response to fully complete and display (e.g., "Log Retrieval Complete" and the links).

Second Question: Immediately enter and submit a new, different question. (Example : "What is the value of gateway.public.host").

Observe: Just before the agent starts streaming the new answer, the previous response ("Log Retrieval Complete" + links) reappears momentarily, followed by the new, correct response.

<img width="629" height="638" alt="Image" src="https://github.com/user-attachments/assets/42bddddc-5957-4785-85cb-3c8df1524d5f" />

### ✅ Expected Behavior

The chat interface should immediately clear the previous agent response or display a new, clean loading/streaming state when a new user question is submitted. The previous message content should not flash or display before the new response starts.

### ❌ Actual Behavior

The previous response is displayed again when the new request is initiated, causing a visual duplication.

### 𝌚 CopilotKit Version

```shell
├── @copilotkit/react-core@1.10.6
├── @copilotkit/react-ui@1.10.6
├── @copilotkit/runtime@1.10.6
```

### 📄 Logs (Optional)

```shell

```

ldessemon2:
After some tests, I discovered that this "bug" occurs when I use McpToolset :

tableau_agent = LlmAgent(
    name="TableauAgent",
    model=GEMINI_2_FLASH,
    instruction="""
    You are the maintenance agent for **Server A (Tableau)**.
    * **Log Retrieval**: If the user asks for logs, call the `tableau_get_logs` tool.  Date Time format must be "mm/dd/yyyy H:mm"
    * **Status**: If the user asks for status, call the `tableau_get_status` tool.
    * **Parameter**: If the user asks for a parameter, call the `get_parameter` tool.
    * Return the tool output directly, without conversational phrases.
    """,
    tools=[
        # McpToolset gives the agent access to tool_tableau_get_logs, tool_tableau_get_status, tool_get_parameter
[...]

PR #6967, opened 2026-09-09
feat(runtime): add native Intelligence runtimes and shared conformance

MikeRyanDev:
## Review status

**The January MCP Apps MIME correction is implemented, but TypeScript integration awaits an upstream release.** Native commit `61d3d0a6cc` advertises `text/html;profile=mcp-app` and adds three shared regression cases. [AG-UI #2722](https://github.com/ag-ui-protocol/ag-ui/pull/2722) fixes the same advertisement in all three TypeScript middleware connection paths.

Published middleware 0.1.0 still advertises the old value. CopilotKit continues to depend on that release, so its normal TypeScript CI is expected to fail the three new MIME cases until the upstream fix is released and consumed. The gate must remain enabled; these failures must not be skipped.

Full local validation passes all 126 shared cases across TypeScript, Python, Go, .NET, Rack, and Rails (756 executions). The TypeScript run used the built upstream correction through a temporary dependency symlink, restored afterward. This is integration evidence for the proposed fix, not evidence that it has been published. Native lint, tests, typechecks, and builds pass; upstream middleware tests (103), typecheck, and build pass.

Visibility filtering is already implemented in all four native runtimes, with five shared cases covering omitted, model, app, both, and empty visibility. Those cases remain green. Nested metadata support remains in place.

MCP `2026-07-28` migration is deferred to a follow-up PR, as agreed in person with the reviewing engineer. The January MIME and visibility corrections use the existing core protocol and do not require that migration. Maintainer approval remains pending.

## Commit structure

The history is grouped into 13 commits by purpose:

- Four TypeScript fixes: tool schemas, MCP Apps, identity/request boundaries, and failure reporting.
- One commit for each native runtime and SDK, including its tests and package guide.
- One shared conformance and CI commit.
- Two focused CI fixture fixes.
- One native MCP access-boundary fix with shared visibility and account-selection regressions.
- One January MCP Apps MIME fix with shared connection-path regressions.

The latest follow-up changes native MCP discovery and proxy selection; the other runtime code is preserved. The public documentation commit was removed; public SDK/runtime guides and navigation updates are deferred until publication. Package-local README files remain with their implementations. Other history-cleanup changes are Python manifest formatting and moving internal plans/checkpoints out of the public tree, with their documentation links updated. Public contract instructions and regression evidence remain in the PR.

Pre-push validation reran scoped formatting, lint, typechecks, tests, builds, and frozen-lockfile checks. Before dropping the public docs commit, its targeted tests and the full docs build passed. The associated new docs test was removed with that commit. The broader docs suite reports eight failures in unrelated tests on unchanged files; existing workflow-lint findings also match the base. These are recorded limits, not new regressions introduced by the rewrite.

## Scope

Keep the TypeScript fixes, native runtimes, SDKs, and shared conformance suite together in this PR. The MCP Apps middleware changes were extracted upstream and are consumed as a released dependency.

Add native Intelligence runtimes and SDKs for Python, Go, Ruby, and C#/.NET. Run TypeScript through the same shared contract tests. Only IntelligenceRunner is in scope; the native runtimes do not use Node sidecars.

The Ruby runtime is a Rack application that can mount in Rails or Sinatra; neither framework is required. Rack and Rails have dedicated conformance runs. Sinatra uses the same Rack interface and does not yet have a separate conformance run.

The runtimes expose run, connect, stop, thread, memory, annotation, discovery, and entitlement APIs with application-owned authentication. They support durable event delivery, A2UI, MCP Apps, and the existing completion analytics contract.

[Public walkthrough](https://intelligence-runtime-guide.mikeryandev.chatgpt.site/)

The TypeScript runtime now consumes `@ag-ui/mcp-apps-middleware@^0.1.0` from npm. The published tarball contains the reviewed origin, visibility, diagnostics, and cleanup fixes. Local packed-runtime installation and ESM/CJS imports pass with the released dependency.

## Human-review corrections

The latest follow-up filters model tools by `_meta.ui.visibility` while keeping iframe calls available. Hash-only proxy requests must resolve exactly one server, and explicit IDs never fall back to a hash. Shared cases cover five visibility settings, ambiguous hashes, both account IDs, and unknown IDs.

- Preserve shipped TypeScript routing and trusted SDK update precedence. Restrict browser update fields instead: identity aliases cannot override the authenticated user in TypeScript, Go, or Ruby. The fixture now uses the actual platform alias precedence.
- Supply required deletion audit reasons in Python, Go, and Ruby. Recover Ruby idle gateway failures without cancelling an agent whose platform lease remains valid. Prefer current nested MCP metadata while retaining the legacy format.
- Remove the TypeScript MCP Apps fork. [AG-UI #2718](https://github.com/ag-ui-protocol/ag-ui/pull/2718) owns transport validation, bounded session cleanup, credential-free hashes, redirect protection, safe errors, and current metadata. The upstream PR is merged and released as v0.1.0; this PR now uses that npm release.
- Restore completion analytics for completed streams containing RUN_ERROR, retain private error handling, preserve stop bodies consumed by identity callbacks, and limit relaxed tool strictness to the A2UI tool.
- Scope conformance CI to runtime dependencies, restore normal package previews, label native libraries as source-only previews, reject remote HTTP MCP credentials in .NET, and target .NET 8 with a real Agent Framework AG-UI integration test.

The stop endpoint still checks current ownership before cancellation. This intentionally fails closed when ownership cannot be verified; platform unavailability can prevent cancellation. Reviewer acceptance of that security/availability tradeoff remains open. The GTM month-close query was not available to inspect; shipped completion-event behavior was restored.

## Validation

Earlier validation at `6351fd34f9`, before the MIME follow-up (current MIME evidence is above):

- 123 shared cases pass on TypeScript, Python, Go, Ruby/Rack, Ruby/Rails, and .NET 8: **738 local case executions** through loopback HTTP and Phoenix sockets.
- 2,342 TypeScript runtime tests, typecheck, and build pass. Python, Go, and Ruby tests, lint, typecheck where defined, and package builds pass. The harness tests include negative stub checks.
- .NET 8 tests, lint, typecheck, build, and pack pass. Both NuGet artifacts contain net8.0 assemblies. The test target hosts the actual Microsoft Agent Framework AG-UI endpoint with a deterministic model and reads it through the runtime HTTP agent.
- The final upstream review fixes passed 102 tests, typecheck, and build with SDK 1.15.0 and 1.20.0. The published v0.1.0 tarball was inspected for those fixes.
- Changed-file lint, workflow syntax, frozen-lockfile installation, and `git diff --check` pass. Earlier shell-docs build/typecheck/lint passed with existing lint warnings. This repository has no `docs:check` target.

Commands and observed failing assertions are recorded in [REGRESSIONS.md](https://github.com/CopilotKit/CopilotKit/blob/codex/intelligence-polyglot-runtimes/tools/runtime-conformance/REGRESSIONS.md). The [contract guide](https://github.com/CopilotKit/CopilotKit/blob/codex/intelligence-polyglot-runtimes/tools/runtime-conformance/AGENTS.md) requires a behavior failure before a production fix and independent review of fixture changes. The [branch rules](https://github.com/CopilotKit/CopilotKit/rules/14521235) retain the conformance gate and independent engineering review.

The current rebase contains main at `dfb68fdbd0`. The 12 patches present at `6351fd34f9` were unchanged by the rebase (`git range-diff`); runtime and fixture Git trees match the tested pre-rebase commit. Fresh native lint, typechecks, tests, builds, and shared conformance validate the MCP follow-up locally. GitHub Actions on `6351fd34f9` is complete: 66 checks pass, three are skipped, and no checks fail. CodeRabbit has started a new review and is pending. The required [Intelligence runtime conformance gate](https://github.com/CopilotKit/CopilotKit/actions/runs/34538610986) passes. Its logs record all 738 shared case executions passing. GitHub still records changes requested; this does not imply review approval.

Release dependency follow-up:

- Replaced the upstream preview URL with `^0.1.0`; the lockfile selects 0.1.0 with registry integrity.
- Added MCP Apps to the named AG-UI release-age exception lists in `.npmrc` and the clean-consumer config, consistent with the existing upstream entries. Exotic transitive dependency blocking remains enabled.
- `NX_DAEMON=false pnpm nx run-many -t test,check-types,build -p @copilotkit/runtime`: 2,342 runtime tests, typecheck, and build pass locally.
- `NX_DAEMON=false pnpm run verify:runtime-package`: packed runtime installation and ESM/CJS imports pass locally on Node 22. The previous `ERR_PNPM_EXOTIC_SUBDEP` failure no longer occurs.
- Eight release-config tests, changed-file lint, frozen-lockfile installation, and `git diff --check` pass. On `ea30890c47`, all six Node 20/22/24 and React 18/19 jobs pass in [the unit and package workflow](https://github.com/CopilotKit/CopilotKit/actions/runs/34530428044). The released dependency has cleared the previous package-install blocker in CI.

Pre-rewrite CI on `ea30890c47` is complete: 67 checks pass and three are skipped, with no failures. The [conformance workflow](https://github.com/CopilotKit/CopilotKit/actions/runs/34530428052) passes for TypeScript, Python, Go, Ruby (Rack and Rails), and .NET 8. All six Node/React package jobs also pass. GitHub still records the maintainer review as changes requested; this CI result does not imply review approval.

## Evidence limits and release boundary

The shared fixture is authored in this PR. Source calibration now covers identity precedence, foreign-thread privacy, deletion reasons, and the shipped TypeScript behavior corrected above. Agreement across drivers alone does not prove deployed-platform compatibility.

The public frontend core runs against real Phoenix sockets in Node; that is not browser coverage. Earlier hosted/browser recordings predate these changes, and their previously recorded checkout path is absent from this workspace. Fresh browser, hosted-platform, and projection-replay validation remain unverified.

Native MCP Apps supports Streamable HTTP. Legacy SSE, OAuth setup, server-initiated requests, and stream resumption remain outside the native implementation; TypeScript retains legacy SSE support. Provider-specific agents, managed Channels, GraphQL, voice, single-route dispatch, automatic thread naming, stateless suggestions, Open Generative UI, and automatic memory-tool injection remain outside this scope. The Agent Framework test uses its public AG-UI endpoint, not a new provider adapter.

The four native libraries are source-only previews. Go source is publicly fetchable by commit and may be cached by module proxies without a release tag. No merge, stable package release, or runtime deployment is authorized by this work.

## Summary by CodeRabbit

* **New Features**
  * Added native CopilotKit Intelligence runtimes and SDKs for Python, Go, Ruby, and .NET.
  * Added support for A2UI, MCP Apps, telemetry, memory policies, thread operations, and Inspector metadata across runtimes.
  * Added typed resource models, runtime entitlements, lifecycle events, and HTTP agent integrations.

* **Bug Fixes**
  * Improved identity and ownership validation, event durability, stream finalization, error handling, and sensitive-data redaction.
  * Improved MCP authentication, proxy safety, request validation, and redirect handling.

* **Tests**
  * Added cross-runtime conformance coverage and automated validation for supported runtime integrations.


BenTaylorDev (dismissed):
## Verdict: needs revision — not mergeable as one PR

Reviewed the branch locally against current `origin/main`. The branch is not behind main (rebased on `ce10479`), and every check passes. `reviewDecision` is empty and only CodeRabbit has commented, so this is the first human pass.

### 1. The premise, before the code

34,201 added lines across 228 files add four language ecosystems: Python, Go, Ruby with Rails, and .NET 9. The body links no issue and no Linear ticket. Approval here commits the OSS monorepo to owning four runtime implementations in four toolchains, indefinitely. Who asked for this, which customers need it, and who owns the Ruby runtime when it breaks? That question outranks every finding below.

### 2. The evidence certifies itself

All 612 passing case executions run against `tools/runtime-conformance/platform.mjs`, a Node fake of the Intelligence platform written in this same PR, plus `LLMock` and `MCPMock`. Six drivers agreeing proves that the six agree with the fixture, not that the fixture matches the deployed platform. The one real anchor is the TypeScript driver, which runs the shipped `@copilotkit/runtime`. That calibrates the fixture only for behavior the TS runtime already has and the fixture happens to assert. The body concedes the rest: browser and deployed-service validation did not repeat on the final commit, and projection-based replay remains unverified.

### 3. Shipped TypeScript behavior changes are bundled in

These 1,789 lines under `packages/runtime` reach every existing OSS user. Each one wants its own PR and its own justification.

- `packages/runtime/src/agent/index.ts:716` — tool schemas stop round-tripping through Zod and now use raw JSON Schema plus a new `@cfworker/json-schema` dependency, with `strict: false` on every tool. Checked the consequence in `@ai-sdk/openai@3.0.36` (`dist/index.mjs:4103`): an unset `strict` omits the field, and the Responses API defaults function tools to strict. An explicit `false` turns strict structured output off for every tool of every user. The motivation (open A2UI objects) is real. The blast radius is wrong.
- `packages/runtime/src/v2/runtime/core/fetch-router.ts:50` — with an explicit `basePath`, the remainder must now match exactly instead of by suffix. An app mounted at a basePath that receives extra path segments starts returning 404.
- `packages/runtime/src/v2/runtime/handlers/handle-stop.ts` — Intelligence stop now blocks on a platform `getThread` call before it aborts the run. The client fires stop and forgets it (`packages/core/src/agent.ts:254`). A slow or unavailable platform now returns 502 and the run never aborts, where the abort used to be purely local. The handler also reads the request body after calling the application-supplied `identifyUser(request)`, so an app whose `identifyUser` reads that body gets a 400 on every stop.
- `packages/runtime/src/v2/runtime/handlers/intelligence/run.ts` — the raw error message no longer reaches analytics, which is a genuine privacy fix, but `agent_execution_stream_ended` stops firing on errored runs. That changes the event counts behind GTM month-close reporting. Please read that usage query before this lands.
- `packages/runtime/src/v2/runtime/intelligence-platform/client.ts:1338` — `updateThread` now lets explicit `userId` and `agentId` override `updates`, reversing the previous precedence. Worth a grep for callers that pass `agentId` inside `updates` to reassign a thread.
- `packages/runtime/src/v2/runtime/middleware/mcp-apps.ts` (new, 377 lines) — an in-repo fork of `@ag-ui/mcp-apps-middleware`, attached to every Intelligence agent even when agent scoping selects zero servers. Two MCP Apps implementations now have to stay in sync.

### 4. Repo-wide cost and release-boundary problems

- `tools/runtime-conformance/gate.mjs:6` — `requiresConformance` matches any path under `packages/`. A one-line change to `@copilotkit/react-core` triggers the full five-language matrix (35-minute timeout per leg, 4 to 10 minutes observed). Please scope it to the runtime and conformance paths.
- `.github/workflows/publish-commit.yml` — this adds a repo-wide `[skip package preview]` opt-out and then uses it on itself, so the `build` job shows `skipping`. The PR that most needs hands-on verification is the one reviewers cannot install a preview of. Split the CI-policy change out and leave previews on here.
- `packages/runtime-go/go.mod` declares `module github.com/CopilotKit/CopilotKit/packages/runtime-go`. On merge, `go get` resolves it through proxy.golang.org, which caches immutably. The npm, PyPI, NuGet, and gem packages are correctly `private: true`, so "no runtime package was published" holds for four of five, but not for Go.
- `showcase/shell-docs/src/content/docs/intelligence/sdks-and-runtimes.mdx` joins the live docs nav and advertises five languages whose install paths are a local checkout or a local NuGet feed. Either gate the page until packages ship, or state on it that four of the five are source-only previews.
- All 19 CodeRabbit threads are resolved, several Major ones without the hunk going outdated, all by the author on an AI-authored branch. Since an independent contract review is required anyway, please spot-check the four security-labelled threads: `packages/runtime-go/runtime.go:364`, `packages/runtime-ruby/lib/copilotkit/runtime.rb:277`, `packages/runtime-dotnet/src/McpAppsAgent.cs:28`, and `packages/runtime/src/v2/runtime/middleware/mcp-apps.ts:174`.

### What holds up

The rebase is clean and CI is green. No changesets. All four new packages are `private`. CODEOWNERS puts the conformance contract under owners other than the author, which is the right call. The modified `nullable-tools.test.ts` keeps its assertions rather than weakening them to fit the new schema path. The `finally`-path `RUN_ERROR` notification does address the CodeRabbit finding it answers.

### Recommendation

Split it. The `packages/runtime` changes are separately reviewable and separately valuable today. The four language runtimes are a program with a staffing plan behind it, not a pull request.

**The one thing to settle before anything merges:** whether CopilotKit is committing to maintain Python, Go, Ruby, and C# runtimes. Every other finding is downstream of that answer.

packages/runtime-go/rest.go
```diff
@@ -0,0 +1,271 @@
...
+			bad(w, 400, "Invalid identifier")
+			return
+		}
+		path += "/" + url.PathEscape(part)
+	}
+	if group == "threads" {
+		if method == "GET" {
+			q.Set("userId", u.ID)
+		} else {
+			body["userId"] = u.ID
```
maxkorp, line 84: [P1] Remove browser-controlled identity aliases from thread requests

This replaces only `userId`, leaving `endUserId` and `end_user_id` from the browser intact. Intelligence resolves `endUserId ?? end_user_id ?? userId`, so these aliases override the authenticated user.

A local reproduction authenticated as `attacker` and requested `/copilotkit/threads/victim-thread/messages?endUserId=victim`. The runtime forwarded `endUserId=victim&userId=attacker` and returned victim history from a server implementing the actual platform identity rules. Ruby PATCH/archive has the same forwarding problem at `lib/copilotkit/runtime.rb:249`.

Please allowlist thread query/body fields and set identity exclusively from the trusted callback. Add shared cases for both aliases on reads and mutations. The fixture currently recognizes only `userId`, so its ownership tests miss this bypass.

packages/runtime-python/src/copilotkit_runtime/runtime.py
```diff
@@ -0,0 +1,872 @@
...
+        if archive or (len(path) == 2 and method in ("PATCH", "DELETE")):
+            fields: Json = {"agentId": required(body, "agentId"), "userId": user.id}
+            fields.update({key: body[key] for key in ("name", "archived") if key in body})
+            if "name" in fields and not isinstance(fields["name"], str):
+                raise RuntimeErrorResponse(400, "Invalid name")
+            if "archived" in fields and not isinstance(fields["archived"], bool):
+                raise RuntimeErrorResponse(400, "Invalid archived")
+            if archive:
+                fields["archived"] = True
+            data = await self.platform.request("PATCH" if archive else method, endpoint, fields)
```
maxkorp, line 800: [P2] Supply the required audit reason for thread deletion

The DELETE payload here contains no `reason`. Intelligence requires a nonempty audit reason, so a normal browser deletion with `{ "agentId": "default" }` fails with HTTP 400. Ruby also drops the reason at `lib/copilotkit/runtime.rb:258`, and Go does not add one to normal browser requests. TypeScript and C# already supply a runtime-generated reason.

A local Go HTTP reproduction against the actual platform validation rules returned 400 with upstream body `{ "agentId": "default", "userId": "attacker" }`.

Please supply the audit reason in all three native runtimes and require it in the shared fixture. The fixture currently accepts deletion without it, hiding this integration failure.

packages/runtime/src/v2/runtime/middleware/mcp-apps.ts
```diff
@@ -0,0 +1,377 @@
...
+    const found = new Map<string, UITool>();
+    for (const server of this.servers) {
+      await this.withClient(server, signal, async (client) => {
+        const cursors = new Set<string>();
+        let cursor: string | undefined;
+        for (let page = 0; page < 100; page++) {
+          const result = await client.listTools(cursor ? { cursor } : {});
+          for (const tool of result.tools) {
+            const resourceUri = tool._meta?.["ui/resourceUri"];
+            if (typeof resourceUri !== "string") continue;
```
maxkorp: [P2] Recognize the current MCP Apps metadata format

Discovery reads only the legacy `_meta["ui/resourceUri"]` field. A server that returns `_meta.ui.resourceUri` silently loses its UI tools before the agent runs. Python, Go, and C# have the same problem; Ruby already handles both formats.

The six existing TypeScript MCP tests pass. Changing only the fixture metadata to `{ ui: { resourceUri: "ui://card" } }` makes two tests fail because no activity appears. A separate Python discovery probe injects zero tools with the current format and one with the legacy format.

Please accept the current nested field with a legacy fallback and cover both formats in the shared suite. The [official MCP Apps documentation](https://apps.extensions.modelcontextprotocol.io/api/variables/server-helpers.RESOURCE_URI_META_KEY.html) requires hosts to support both.

packages/runtime-ruby/lib/copilotkit/runner.rb
```diff
@@ -0,0 +1,330 @@
...
+      @thread = Thread.new do
+        @telemetry.emit('oss.runtime.agent_execution_stream_started')
+        heartbeat = Thread.new do
+          loop do
+            sleep @heartbeat_interval
+            break if @stopped
+            @gateway.heartbeat
+          rescue StandardError => error
+            report_error(error)
+            request_stop(code: 'LOCK_RENEWAL_FAILED')
```
maxkorp: [P2] Reconnect after a gateway heartbeat failure instead of canceling the run

A temporary socket failure here calls `request_stop(code: 'LOCK_RENEWAL_FAILED')`, even while the separate platform lease renewal succeeds. A gateway restart during a long model/tool call therefore cancels a run whose lock remains valid.

An isolated probe exercised the real Runner with a healthy platform lease and a failed gateway heartbeat. It produced one connection attempt followed by `RUN_ERROR` with `LOCK_RENEWAL_FAILED`; it never attempted reconnection. This was a transport simulation, not a deployed-service test.

Please recover the gateway connection within the bounded retry policy and reserve lease-loss cancellation for actual lease failures. Add an idle-agent recovery case, since event-publication retry tests do not exercise this path.

packages/runtime-dotnet/src/CopilotKit.Intelligence.Runtime.csproj
```diff
@@ -0,0 +1,14 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>net9.0</TargetFramework><ImplicitUsings>enable</ImplicitUsings><Nullable>enable</Nullable><TreatWarningsAsErrors>true</TreatWarningsAsErrors>
```
maxkorp: [P2] Support .NET 8 consumers for Microsoft Agent Framework integration

Both this runtime and `sdk/CopilotKit.Intelligence.csproj` target only `net9.0`. A `net8.0` application cannot reference these libraries without upgrading its application target.

Microsoft Agent Framework is an intended integration for this runtime. Its [Microsoft.Agents.AI 1.20.0 package](https://www.nuget.org/packages/Microsoft.Agents.AI/1.20.0) includes `net8.0`. Its [AG-UI ASP.NET Core host](https://www.nuget.org/packages/Microsoft.Agents.AI.Hosting.AGUI.AspNetCore/1.20.0-preview.260831.1) also includes `net8.0`. Agent Framework itself does not require the application to move to .NET 9.

Please add `net8.0` support to both CopilotKit libraries, either as their baseline or through multiple targets. Preserve JSON validation behavior when adapting .NET 9 APIs such as `RespectNullableAnnotations`. Add a .NET 8 build/test job and a small Agent Framework integration test. The current workflow installs only .NET 9, so it cannot prove compatibility for these consumers.

This concerns applications that remain on .NET 8; it does not imply that Agent Framework cannot run inside a .NET 9 application.

maxkorp (commented):
Four reviewed findings, plus the confirmed .NET 8 compatibility gap for the intended Microsoft Agent Framework integration. Validation used focused tests, local simulations, and current platform/package contracts; no deployed-service validation.

packages/runtime-python/src/copilotkit_runtime/mcp_apps.py
```diff
@@ -0,0 +1,296 @@
...
+        for server in self._servers(agent_id):
+            result = await self._request(server, "tools/list")
+            for tool in result["tools"]:
+                meta = tool.get("_meta")
+                meta = meta if isinstance(meta, dict) else {}
+                ui = meta.get("ui")
+                uri = ui.get("resourceUri") if isinstance(ui, dict) else None
+                if not isinstance(uri, str):
+                    uri = meta.get("ui/resourceUri")
+                if not isinstance(uri, str):
```
maxkorp, line 174: [P2] Respect MCP tool visibility in the native runtimes

Discovery accepts `_meta.ui.resourceUri` but does not inspect `_meta.ui.visibility`. A tool marked `visibility: ["app"]` therefore enters model input, even though the server restricts it to UI use.

A local shared-fixture probe registered an app-only tool and inspected the HTTP agent input. Python, Go, and .NET all exposed it; TypeScript with the released upstream middleware correctly excluded it. Ruby has the same missing filter in `lib/copilotkit/mcp_apps.rb`.

Please port the upstream visibility rules to all four native implementations: omitted visibility permits model access, while explicit visibility must include `model`. Keep app-only tools callable through the UI proxy. Add shared cases for omitted visibility, `model`, `app`, both, and an empty array.

packages/runtime-go/mcp.go
```diff
@@ -0,0 +1,318 @@
...
+	}
+	return emit(Event{"type": "ACTIVITY_SNAPSHOT", "messageId": uuid(), "activityType": "mcp-apps", "content": activity, "replace": true})
+}
+func (a *uiAgent) proxy(ctx context.Context, input, request map[string]any, emit func(Event) error) error {
+	if err := emit(Event{"type": "RUN_STARTED", "threadId": input["threadId"], "runId": input["runId"]}); err != nil {
+		return err
+	}
+	var server *MCPServer
+	for _, s := range a.mcp {
+		if (str(request["serverId"]) != "" && s.ServerID == str(request["serverId"])) || serverHash(s) == str(request["serverHash"]) {
```
maxkorp: [P2] Reject ambiguous MCP server hashes before selecting credentials

The hash includes the transport and URL but excludes headers. Two configured accounts at the same MCP URL therefore share a hash. This lookup selects the first matching entry instead of rejecting the ambiguity, so a hash-only UI request can execute with the wrong account credentials. Python and Ruby also select the first match; .NET selects the last.

A local shared-fixture probe configured two server IDs at one URL with distinct account headers, then sent a hash-only `tools/call`. Python, Go, and .NET executed the tool; the released TypeScript middleware rejected the request without executing it.

Please require an explicit `serverId` when multiple configured servers share a hash and cover this in the shared suite. When an ID is supplied, resolve it before considering a hash; the OR condition here can otherwise select an earlier hash match instead of the requested ID.

maxkorp (commented):
Two remaining native MCP Apps findings from the re-review. Both reproduce through local HTTP/socket fixtures in Python, Go, and .NET; the released TypeScript middleware passes the same probes. Ruby has the same selection logic by source inspection. The affected code is unchanged from reviewed commit 7c3f25b4.

BenTaylorDev:
## Re-review of head `f2926a54e3`

Follow-up to my earlier review. This one is a comment, not a new changes-requested. The earlier review still stands on its own.

### Withdrawing the premise question

My previous review made this the gating question: whether CopilotKit is committing to maintain Python, Go, Ruby, and C# runtimes. That is settled. We are committing to maintain them. The staffing half of my "split it" recommendation is withdrawn, and "who owns the Ruby runtime when it breaks" is answered.

The remaining findings are ordinary code review.

### State

Three commits behind `origin/main`. Merge-base `b2dca0ad22`, main now `dfb68fdbd0`. No retired-tree or changeset debris. CI is green on this exact head: 66 pass, 3 skipped, 68 check runs on `f2926a54e3`.

### Addressed since the last pass

| Prior finding | Status |
| --- | --- |
| In-repo MCP Apps fork, 377 lines | Fixed. Fork deleted. Consumes `@ag-ui/mcp-apps-middleware@^0.1.0`. |
| `fetch-router.ts` exact-match routing | Reverted. Only a test remains. |
| `updateThread` precedence reversal | Fixed. `threads.ts:157` allowlists `name` and `archived`. |
| `agent_execution_stream_ended` dropped on error | Fixed. `run.ts:362` restores the completion capture. |
| `identifyUser` consuming the stop body | Fixed. `handle-stop.ts` clones the request before resolving identity. |
| Repo-wide conformance gate | Fixed. `gate.mjs:6` scopes to runtime, core, shared, aimock. |
| `[skip package preview]` opt-out | Fixed. Removed entirely, and previews posted on this PR. |
| Docs page advertising five languages | Fixed. The docs commit was dropped. |

That is a real pass over the list. Thank you.

### New finding: the tool-schema fix trades one regression for another

Limiting `strict: false` to the A2UI tool is correct and closes my original finding. But the change also swaps the schema source, and the replacement drops a constraint the classic path depends on.

`packages/runtime/src/agent/index.ts:720` now forwards the raw AG-UI JSON Schema instead of the Zod round trip. I ran both paths through the real `convertJsonSchemaToZodSchema` and the real `asSchema`, on an all-required schema:

```
PRE  (main):  {"type":"object","properties":{"city":{"type":"string"},
               "opts":{...,"additionalProperties":false}},
               "required":["city","opts"],"additionalProperties":false}

POST (#6967): {"type":"object","properties":{"city":{"type":"string"},
               "opts":{...}},
               "required":["city","opts"]}
```

`additionalProperties: false` disappears at every object level. `@ai-sdk/openai@3.0.36` forwards `parameters: tool.inputSchema` unchanged and omits `strict` when it is unset (`dist/index.mjs:4103`), so nothing downstream restores it.

This repository already documents the invariant, in `packages/runtime/src/agent/converters/tanstack.ts:190`:

> OpenAI rejects both: strict mode requires `additionalProperties: false` [...] The classic (Vercel AI SDK) path sanitized these implicitly via a Zod round-trip; the TanStack path forwards the raw schema, so we close open objects here to match.

This PR makes the classic path forward the raw schema too, and adds no equivalent of `sanitizeClientToolSchema`. The fix looks small: run the same normalization for every tool except the A2UI one.

The new tests do not cover this. Every case in `ag-ui-tool-validation.test.ts` validates through the `@cfworker/json-schema` Validator that this PR installs, so it confirms the validator agrees with itself. No case asserts what reaches a model provider. One case pins the changed shape as intended: `expect(await schema.jsonSchema).toEqual(parameters)`.

Please confirm with one live OpenAI call through the classic path before merging. A unit test against the validator cannot settle it.

### Still open

- **MCP tool visibility** (maxkorp, `runtime-python/.../mcp_apps.py:170`). Live at this head. `grep -c visibility` returns 0 in the Python, Go, and Ruby MCP Apps implementations. App-only tools still reach model input.
- **Ambiguous MCP server hash** (maxkorp, `runtime-go/mcp.go:296`). Live at this head. The `(serverId match) || (hash match)` condition breaks on the first match, so an earlier hash collision can win over an explicitly requested `serverId`.
- **Blocking `getThread` on stop.** Recorded as an open reviewer decision rather than fixed. Worth stating plainly: the client fires stop and forgets it, so when the platform is slow or unavailable the run never aborts and nothing surfaces to the user. I would rather abort locally first and reconcile ownership after.
- **Go module path.** `packages/runtime-go/go.mod` still declares a fetchable module path, and the body now acknowledges proxy caching instead of preventing it. With the maintenance commitment in place this is a release-sequencing question, not a blocker.

### Before merge

Fix the tool-schema normalization. It is the only item here that reaches every existing OSS user on the OpenAI path, and it is independent of everything else in this PR.

packages/runtime-go/rest.go
MikeRyanDev: Addressed in the rebased branch. The eight shared access.thread-identity-* cases cover endUserId and end_user_id on reads, lists, updates, and deletion. Browser aliases cannot override the trusted identity; the fixture checks the platform alias precedence. The full local shared suite passes across all six execution modes. Failure evidence is in tools/runtime-conformance/REGRESSIONS.md.

packages/runtime-python/src/copilotkit_runtime/runtime.py
MikeRyanDev: Addressed. Python, Go, and Ruby now supply a runtime-generated deletion audit reason, and the shared platform fixture requires it. access.thread-delete-audit-reason reproduced HTTP 400 before the fixes and passes in the full cross-language suite. Evidence is in tools/runtime-conformance/REGRESSIONS.md.

packages/runtime/src/v2/runtime/middleware/mcp-apps.ts
MikeRyanDev: Addressed across all five languages. Nested _meta.ui.resourceUri takes precedence, with the legacy field retained as fallback. The shared mcp-apps.current-metadata-nested and mcp-apps.current-metadata-both cases pass. TypeScript consumes the merged and published @ag-ui/mcp-apps-middleware 0.1.0 instead of a local fork.

packages/runtime-ruby/lib/copilotkit/runner.rb
MikeRyanDev: Addressed. Ruby retries the gateway connection within the bounded recovery policy while the platform lease remains valid. The shared runner.idle-heartbeat-reconnect case reproduces the idle-agent path and passes in Rack and Rails as part of the complete local suite. Lease-loss cancellation remains separate. Failure evidence is in tools/runtime-conformance/REGRESSIONS.md.

packages/runtime-dotnet/src/CopilotKit.Intelligence.Runtime.csproj
MikeRyanDev: Addressed. Both libraries target net8.0. The .NET test target includes an actual Microsoft Agent Framework AG-UI host with a deterministic model; SDK, runtime, and integration tests pass under .NET 8. The .NET 8 driver also passes all 123 shared cases. Nullable-field and JSON metadata behavior are covered; details are in tools/runtime-conformance/REGRESSIONS.md.

packages/runtime-python/src/copilotkit_runtime/mcp_apps.py
MikeRyanDev: Addressed in 6351fd34f9. Added five shared mcp-apps.visibility-* cases for omitted, model, app, both, and empty visibility. The app-only case failed in all four native runtimes before the fixes. Discovery now requires model in explicit visibility, while iframe proxy calls remain available. All five cases pass across TypeScript, Python, Go, .NET, Rack, and Rails; the complete shared suite passes too.

packages/runtime-go/mcp.go
MikeRyanDev: Addressed in 6351fd34f9. Added four shared mcp-apps.server-selection-* cases for an ambiguous hash, each explicit account ID, and an unknown ID. All four native runtimes failed the ambiguous-hash probe; Go also selected the wrong credentials for the second explicit ID. Selection now requires exactly one registration and never falls back from an explicit ID to a hash. Tests verify zero network requests on rejection and the credential header on every accepted request. The complete shared suite passes across all six execution modes.

onsclom (approved):

onsclom (dismissed):

BenTaylorDev (dismissed):

PR #6969, opened 2026-09-09
fix(react-core): stop useInterrupt re-renders on agent updates

### LINKED ISSUE #6934: 🐛 Bug: useInterrupt causes re-renders
### ♻️ Reproduction Steps

1. Build a component similar to the one below
2. Submit a query
3. Observe multitple re-renders

The issue stems from the fact that useInterrupt call useAgent without ` updates: []`, resulting in hook re-rendering on any event received

```tsx
import { useInterrupt, CopilotChat } from "@copilotkit/react-core/v2";
import { useRef } from "react";

const AGENT_NAME = import.meta.env.VITE_AGENT_NAME || "default";

export const ChatInterface2 = () => {
  const renderCountRef = useRef(0);
  renderCountRef.current++;

  useInterrupt({
    render: ({ event, resolve }) => (
      <div className="p-3 border rounded">
        <p>{event.value.question}</p>
        <div className="mt-2 flex gap-2">
          <button onClick={() => resolve({ approved: true })}>Approve</button>
          <button onClick={() => resolve({ approved: false })}>Reject</button>
        </div>
      </div>
    ),
  });

  return (
    <>
      {renderCountRef.current}
      <CopilotChat agentId={AGENT_NAME} />
    </>
  );
};
```

### ✅ Expected Behavior

calling useInterrupt hook should not trigger re-renders

### ❌ Actual Behavior

calling useInterrupt hook triggers re-renders

Note:

A work-around is to wrap useInterrupt into a separate component and then use as a child 
```ts
  return (
    <>
      <AddUseInterrupt />
      <CopilotChat agentId={AGENT_NAME} />
    </>
  );
```

### Screenshots

Started with 10 re-renders on initial load

Image

Ended with 30 after a basic message

Image

### 𝌚 CopilotKit Version

```shell
├── @copilotkit/react-core@1.70.0
├── @copilotkit/react-ui@1.70.0
```

### 📄 Logs (Optional)

```shell

```

deepshekhardas:
fix #6934

useInterrupt only needs the agent handle - interrupt events arrive via its own direct agent.subscribe(). But useAgent() with no update filter force-updates the consumer on every message/state/run-status change. Passing updates: [] skips the re-render subscription entirely (use-agent early-returns on an empty list) while the agent handle still resolves.

Includes a regression test asserting the hook opts out of agent updates.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved interrupt handling by preventing unnecessary re-renders when unrelated agent updates occur.
  - Interrupt-related behavior now remains stable when chat rendering is disabled.

### DIFF
--- a/packages/react-core/src/v2/hooks/__tests__/use-interrupt.test.tsx
+++ b/packages/react-core/src/v2/hooks/__tests__/use-interrupt.test.tsx
@@ -209,6 +209,17 @@ describe("useInterrupt", () => {
     expect(unsubscribeMock).toHaveBeenCalledTimes(1);
   });
 
+  it("opts out of agent re-render updates", () => {
+    render(<Harness renderInChat={false} />);
+
+    // Interrupt events arrive via a direct agent.subscribe() in the hook,
+    // so the useAgent() handle must not force re-renders on every
+    // message/state/run-status change (see #6934).
+    expect(mockUseAgent).toHaveBeenCalledWith(
+      expect.objectContaining({ updates: [] }),
+    );
+  });
+
   it("ignores non-interrupt custom events", () => {
     render(<Harness renderInChat={false} />);
 
--- a/packages/react-core/src/v2/hooks/use-interrupt.tsx
+++ b/packages/react-core/src/v2/hooks/use-interrupt.tsx
@@ -172,7 +172,7 @@ export function useInterrupt<
 ): UseInterruptReturn<TRenderInChat> {
   /* eslint-enable @typescript-eslint/no-explicit-any */
   const { copilotkit } = useCopilotKit();
-  const { agent } = useAgent({ agentId: config.agentId });
+  const { agent } = useAgent({ agentId: config.agentId, updates: [] });
   const [pending, setPending] = useState<PendingInterrupt | null>(null);
   const pendingRef = useRef(pending);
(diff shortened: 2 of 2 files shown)


lukasmoschitz (approved):
Verified this locally against current main. Two things beyond reading the diff:

I checked that `updates: []` actually means "no updates" rather than falling through to the default — `use-agent.tsx:219` uses `??`, which doesn't fall back on an empty array, and the subscription effect at `:373` returns before installing anything. Then I measured it with the real hook in the `use-agent-throttle` harness: with `updates: []`, five `onMessagesChanged` notifications produce zero extra renders, while the same harness with `OnMessagesChanged` re-renders as expected. Full react-core target is green (145 vitest files, 47 node tests).

Also removed your one-liner and confirmed your test fails without it — thanks for writing one that actually pins the change.

One note for anyone reading this later: the `render` callback receives `agent`, and it will no longer re-render on message/state changes. That's the point of the fix and matches what #6934 asked for, but an interrupt UI that displays live agent data would need its own `useAgent`.

Separately, and not for this PR: `use-capabilities.tsx:20` has the same shape — `useAgent({ agentId })` where only `agent.capabilities` is read. Worth checking whether it wants the same treatment, though capabilities may arrive asynchronously and rely on the re-render.

PR #6972, opened 2026-09-09
docs(vue): add React parity contributor guide

enekesabel:
Adds a focused contributor guide for keeping @copilotkit/vue aligned with React. It explains the roles of README.md, AGENTS.md, and PARITY.md, the React-to-Vue synchronization workflow, intentional framework differences, current documentation ownership, and Nx validation. The branch includes the README update from #6222 and can be rebased after that PR lands.


lukasmoschitz:
Looks like this branched off `codex/vue-showcase` rather than `main`, so it's carrying all of #6222 — and the guide commit landed on #6222 itself about half an hour after you opened this (`89bdf24286`). So as far as I can tell everything here is already covered there.

Am I reading that right? If so, would you mind closing this one? And if you'd rather have the guide as a standalone docs PR after all, branch it from `main` and I'll review that separately.

enekesabel:
You're right, I'm sorry I was in a little but of rush. I intended to include this only in 6222.
Closing this pr, thanks for pointing it out!

PR #6997, opened 2026-09-09
fix(skills): audit react-core claims, and stop documenting Cloud keys as the Intelligence path

BenTaylorDev:
Follow-up to #6993. That PR fixed two `react-core` references; four of their six citations were stale, so this audits the other 13.

## Method

Three passes, because the first two are cheap and the third is the only one that finds real defects.

1. **Mechanical** — every `packages/...:NN-MM` citation: does the path exist, is the span in range? 71 citations, 57 with spans.
2. **Symbol** — every `useXxx(` and `<CopilotXxx` in the docs against the 389 identifiers actually exported from `v2/src`.
3. **Semantic** — read the cited lines and check they support the claim. This is where the rot lives: the lines exist, they just say something else.

## Findings fixed

| File | Defect |
| --- | --- |
| `capabilities.md` | Cited `runtime/src/agent/index.ts:821-829,883-887` for "shallow-merges capabilities at the category level". Those are factory-mode config types and sampling params (`frequencyPenalty`, `stopSequences`). The mechanism is `:940-947` — whose own doc comment says **shallow-merged** — and `:999-1012`. |
| `provider-setup.md` | Claimed the provider resolves `publicLicenseKey \|\| publicApiKey`. Only one of four sites does. Citation also pointed at an unrelated line (`copilotkit.tsx:172` is `source: "agent"`). |
| `custom-message-renderers.md` | Cited lines 73-95 of a 93-line file. The iterate-and-break it describes is at `:68-91`. |
| `suggestions.md` | `useFeatureFlag("suggestions")`, twice, with no import and no definition. CopilotKit exports no such hook. |
| `threads.md` | `useThreadSelection()`, same problem. |

### The license-key one is a product finding, not just a docs bug

Precedence when both keys are set is inconsistent in the code:

| Site | Order |
| --- | --- |
| `CopilotKitProvider.tsx:487` | `publicApiKey ?? publicLicenseKey` |
| `copilotkit.tsx:111` | `publicApiKey \|\| publicLicenseKey` |
| `copilotkit.tsx:217` | `publicLicenseKey \|\| publicApiKey` |
| `copilotkit.tsx:883` | `publicApiKey \|\| publicLicenseKey` |

Three prefer `publicApiKey`, one prefers `publicLicenseKey`. So which wins depends on which path runs. Rather than document one order as if it were the contract, the doc now says to write the canonical name and not to set both — true regardless of path. **The underlying inconsistency is untouched here and probably wants its own issue.**

## Findings I retracted

Recording these because two of my three automated passes produced false positives, and the ratio matters for anyone repeating this.

- **50 "missing paths."** My regex alternated `(?:ts|tsx)`, so it matched `ts` first and truncated every `.tsx`. All 50 were my own artifact. Real count: zero missing paths.
- **41 "unsupported spans."** A heuristic checking whether the claim's backticked identifiers appear in the cited lines. It attributes each `Source:` to the nearest paragraph above, which is the wrong one in multi-paragraph gotchas. Useful as a reading list, worthless as a finding.
- **"13 uses of the deprecated provider."** Most were the *filename* `CopilotKitProvider.tsx` inside `Source:` citations, not component usages. There are 2 real JSX usages, both passing only `runtimeUrl`, both valid — and `provider-setup.md` already carries a callout explaining that `CopilotKit` is the v1/v2 bridge and `CopilotKitProvider` "is a perfectly good choice if you do not need the v1 bridge". No defect.

## Testing

```
$ python3 audit_citations.py
citations: 71   mechanically broken: 0
with a line span: 57   path-only: 14

plugin skill mirror in sync

$ oxfmt --check skills/react-core/references/*.md
```

Symbol pass after the fixes leaves only legitimate non-exports: React's own hooks, two helpers the docs define inline (`useAvailableAgents`, `useMyFeatureFlag`), and `useAgents`, which `switching-agents.md` correctly documents as **not existing**.

Edits were made in `packages/react-core/skills/` and mirrored by the sync script.

## Coverage, honestly

The mechanical and symbol passes cover all 15 files completely. The semantic pass does not: I read roughly 15 of the 57 spans closely, prioritising the files documenting APIs that have moved most. The five defects above are what that subset produced. A full semantic read of the remaining ~40 spans would likely find more, and the same audit has not been run on the `runtime` skill (26 references) or `a2ui-renderer`.

## Summary by CodeRabbit

- **Documentation**
  - Updated React guidance with current source references across capabilities, message renderers, attachments, chat components, client-side tools, debugging, human-in-the-loop rendering, tool calls, agent access, and agent switching.
  - Clarified that production SPAs require `runtimeUrl`, with Intelligence configured server-side on the runtime using `CPK_INTELLIGENCE_API_KEY`.
  - Removed outdated `publicLicenseKey` guidance from setup and API documentation.
  - Corrected feature-flag examples to use an application-defined hook.
  - Updated thread-selection examples to derive the active thread ID from application state.


BenTaylorDev:
## Added: the license-key finding turned out to be a product mislabelling

The precedence inconsistency I flagged earlier was a symptom. `publicApiKey` and `publicLicenseKey` route a runtime-less client at **CopilotKit Cloud** — `api.cloud.copilotkit.ai`, header `X-CopilotCloud-Public-Api-Key` (`CopilotKitProvider.tsx:90-91, 536`). They have never been an Intelligence mechanism.

Five skills presented them as exactly that:

| Location | Claim |
| --- | --- |
| `provider-setup.md:101` | Section titled "SPA with CopilotKit **Intelligence** (no self-hosted runtime)" |
| `telemetry-setup.md:67` | "To disconnect from **Intelligence**, remove the `publicLicenseKey` prop" |
| `telemetry-setup.md:45-60` | Expose the key with `NEXT_PUBLIC_` / `VITE_` to connect Intelligence |
| `error-patterns.md:76` | `MISSING_PUBLIC_API_KEY_ERROR` attributed to Intelligence |
| `copilotkit-setup/SKILL.md:351` | "CopilotKit **Intelligence** public license key" |

`copilotkit-upgrade` was the only skill that had it right, calling it "a cloud key".

### Two claims that were not just mislabelled but false

- **There is no client-only path.** `CopilotKitProvider.tsx:524-535` throws in production without `runtimeUrl`, a key, or dev-only local agents. The "no self-hosted runtime" section described a setup that cannot work for Intelligence.
- **Removing the prop does not disconnect Intelligence.** Intelligence is wired server-side, so the documented opt-out was a no-op. Worse, the file instructed readers to `NEXT_PUBLIC_`-prefix a credential.

### What changed

`telemetry-setup.md` is rewritten around the actual flow — `copilotkit login` → `project select` → `INTELLIGENCE_API_KEY` in the server environment, verified with `copilotkit verify` — and states explicitly that the variable takes no public prefix.

The `publicApiKey` vs `publicLicenseKey` naming gotcha is **removed rather than corrected**: neither prop belongs in Intelligence guidance, so documenting which is canonical was the wrong fix. Three mentions remain and all three name Cloud.

### Gates

```
plugin skill mirror in sync
oxfmt: All matched files use the correct format. (132 files)
citations: 70   mechanically broken: 0
```

One formatting note: my table-row edit in `breaking-changes.md` broke column alignment, which `oxfmt` fixed. I confirmed against `origin/main` that the file was clean beforehand, so that reformat is mine and not unrelated churn.

tylerslaton (approved):

PR #7001, opened 2026-09-09
fix(react-core): use new run IDs for standard interrupt resumes

### LINKED ISSUE #6999: [Bug]: Standard interrupt resumes reuse the completed wire run ID
Standard useInterrupt resolve/cancel passes the originating interrupted runId into RunHandler. RunHandler pins an explicit runId on the AG-UI wire, so resume[] changes the input body while reusing the completed run ID. A stateful gateway that enforces run-ID idempotency rejects this as RUN_ID_CONFLICT.

A standard resume continues the thread in a new run and should allow the AG-UI client to generate a fresh wire run ID. The interrupt IDs and thread ID must remain unchanged. Legacy command.resume and its logical run grouping should preserve their current behavior.

Reproduction: receive RUN_FINISHED with an interrupt on run A, resolve or cancel via the public hook, then inspect the next RunAgentInput. It currently carries runId A with resume[]. Expected: a different runId. React is confirmed on main f8520c9804; matching bindings will be checked before changing them.

This is separate from reconnect recovery in #6891. Verification should intercept actual wire inputs, not only mock a hook callback.

itsklimov:
## What does this PR do?

Standard `useInterrupt` resumes currently reuse the run ID that produced the interrupt. A gateway retaining idempotent runs rejects the changed input with `RUN_ID_CONFLICT` instead of accepting the answer.

Let `runAgent` generate a fresh run ID for standard resolve/cancel calls. The thread and `resume[].interruptId` retain continuity. This removes the interrupt-to-run map; legacy `command.resume` behavior stays intact.

Tests pass the hook through `CopilotKitCore` and `AbstractAgent.run` and assert the actual wire input for resolve, cancel, and legacy resumes.

Validation under Node 22:

- React Core tests: 147 files, 1,687 tests passed.
- React Core build and type checks passed, including their dependency targets.
- Commit hooks passed: affected package tests, publint, attw, lint, environment-name validation, and commitlint.

## Related PRs and Issues

Closes #6999.

The recovery changes in #6891 touch the removed map and will need coordination when either PR lands. This PR contains only the standard wire run-ID fix.

## Checklist

- [x] I have read the Contribution Guide.
- [x] I have updated the relevant documentation.

## Summary by CodeRabbit

* **Bug Fixes**
  * Standard interrupt resumes now start a fresh run on the same thread.
  * Resuming or canceling standard interrupts no longer reuses the interrupted run’s identifier.
  * Interrupt IDs remain consistent when multiple interrupt resolutions overlap.
  * Legacy interrupt resumes continue to retain the original run identifier.

* **Documentation**
  * Updated `useInterrupt` guidance to explain fresh run identifiers and interrupt ID usage for standard resumes.


packages/react-core/src/v2/hooks/use-interrupt.tsx
```diff
@@ -301,8 +291,8 @@ export function useInterrupt<
       try {
         return await copilotkit.runAgent({
           agent,
+          // A standard resume is a new run; resume entries identify the interrupts.
```
AlemTuzlak: **[suggestion]** This comment restates the surrounding call and embeds the design story (standard resume is a new run; interrupt ids carry continuity). The omitted `runId` already does that.

**Suggestion:** Delete the comment. Same text is also on the cancel path at line 354; delete that one too.

packages/react-core/src/v2/hooks/use-interrupt.tsx
```diff
@@ -364,8 +351,8 @@ export function useInterrupt<
...
```
AlemTuzlak: **[suggestion]** Duplicate of the resolve-path comment. Same restatement / design leak on cancel.

**Suggestion:** Delete this comment.

AlemTuzlak (approved):
The standard resume path omits `runId`, so `runAgent` creates a new ID and the gateway does not return `RUN_ID_CONFLICT`. The thread and `resume[].interruptId` stay. Legacy `command.resume` still pins the old run ID. I deleted the two comments that restated this design.

lukasmoschitz (approved):

itsklimov:
@AlemTuzlak,  Rebased onto upstream `326b902769` while preserving both reviewed commits, including the comment cleanup. `git range-diff` reports no patch changes. The rebuilt React Core package passes all 1,690 tests locally.

The approvals remain in place, but the new workflows require maintainer approval. Could you approve those runs? Main has advanced again while validation ran; I will avoid repeated rebases unless you need another update for the merge.

The timeline also confirms that GitHub disabled the previously enabled auto-merge on this push because I do not have write access to upstream. I missed that precondition before rebasing. I will leave the branch untouched from here.

AlemTuzlak (approved):

itsklimov:
@AlemTuzlak the branch is behind main (rule: strict status checks). I can't rebase without dismissing your approvals, and auto-merge was dropped on my last push since I don't have write access. Could you press "Update branch" and merge, or re-enable auto-merge? Thanks.

PR #7012, opened 2026-09-10
Rebuild the docs homepage around a product map

lukasmoschitz:
Closes OSS-1141.

The docs homepage described CopilotKit in five stacked prose sections that had drifted apart, and never said how the pieces relate. It now opens with a hero and then draws the product as a map: **Frontend → CopilotKit → Agent**, with **CopilotKit Intelligence** branching off the runtime.

The layout is the argument. Intelligence is not a fourth step in the stack. The agent connects to the runtime over AG-UI, and Intelligence hangs off that runtime. So the three neutral blocks run full width, Intelligence is inset to the right three of four columns, the `+ adds` branch drops straight into it from CopilotKit, and the AG-UI axis runs down the free left column from CopilotKit's bottom edge, past Intelligence, to the Agent block. A plain vertical stack would have claimed the agent talks to Intelligence, which is false.

## Against the ticket's requirements

| Requirement | State |
|---|---|
| Explain CopilotKit and Intelligence in human-friendly language | Done. Every block leads with what the reader gets, not what shelf the product sits on. |
| Demonstrate the value through a concrete product example | **Partial, see below.** |
| Connect the demonstration to a clear next step | Done. Each block carries one action; the hero carries the prompt button. |
| Help users discover relevant capabilities | Done. Six CopilotKit capabilities and six Intelligence ones, as tiles that link into their own docs. The CopilotKit tiles are scoped to the framework selected in the sidebar. |
| Clear paths for new projects, existing apps/agents, and CopilotKit users adding Intelligence | **Answered in prose, not as three entry points, see below.** |
| Coding-agent prompt primary, manual setup secondary | Done. The prompt button is the primary hero action, the quickstart picker the secondary one. |
| Consistent CTA wording, redundant actions consolidated | Done. Five sections and a duplicate Start section are gone; the hero's two actions are the only entry points. |

### The product example is a map, not a demo

The ticket proposed an interactive demo, a short video, or a visual walkthrough. This delivers the third: a visual walkthrough of how the product fits together. It does not show a running product.

That gap is real and worth deciding on rather than papering over. No product overview recording exists today. The only candidates are individual feature clips and `coagents/overview.mp4`, which is already in use as the partner-page banner. A demo or a video is a separate piece of work with its own asset dependency, and this page has a slot for it directly under the hero whenever one exists.

### The three starting points are a sentence, not three entry points

The ticket asks for clear paths for new projects, existing apps or agents, and existing CopilotKit users adding Intelligence. The hero's subtitle says it in one line instead: you drop CopilotKit into an app you already have, or start a new one, on whatever agent framework you run.

There is no set of three buttons because the CLI does not need one. `copilotkit onboard start` classifies the starting state itself by inspecting the repository, which is what its classification graph does before it asks anything. Three per-path buttons would all hand over the same prompt and reach the same place, so the choice would be theatre: it would look like a decision the reader has to get right, and any of the three would behave identically.

An earlier revision named the three states as a separate line of labels below the buttons. Review read that as filler in the wrong position, which it was: the point it makes is about what CopilotKit needs from you, so it belongs in the first sentence a reader meets, not in a footnote under the call to action.

If you would rather have three visible entry points anyway, say so. It is a small change, and the argument above is the only thing standing in its way.

## Notable implementation choices

- **The registry stays out of the browser.** The framework registry is around 646 KB. The map is composed on the server and the one piece that needs client state, scoping the CopilotKit tiles to the sidebar's framework, is a small client leaf that receives a plain record of frameworks as props.
- **Icons are imported by name.** A namespace import with a runtime index measured 605 KB minified against 6 KB for named imports, because the bundler has to retain every export and `optimizePackageImports` cannot rewrite it.
- **Connectors take the border colour of the blocks they join**, so a line reads as the same material as the boxes at its ends. On a dark ground that makes the neutral lines very quiet, since the border token sits at a tenth of an opacity. Deliberate; flag it if it reads as absent to you.
- **The framework grid honours the partner-agreed display order** and hides integrations marked `docs_mode: hidden`. Both are covered by tests that fail if the filter or the sort is dropped, because a stale count literal had let exactly that through once.

## Review feedback incorporated

- **Eyebrow labels removed** from above each block name, along with the prop that rendered them. They said nothing the block name did not.
- **Hero cut to a heading, a line and one sentence**, from three paragraphs, and that sentence leads with the app the reader already has.
- **Intelligence features carry their canonical names**: Rich Threads, User Memories, Automatic Learning, Product Analytics. Inspector and Self-hosting are left alone, being a tool and a deployment option rather than features on that list. Say the word if they should change too.
- **Em dashes are out of the visible copy**, including the metadata a search result renders.
- **The hero no longer says "a React app"**, raised on review as a question about whether that word follows the frontend picker. It cannot: the picker reads its value from the URL path, the home page has no frontend prefix, and choosing another frontend navigates to that frontend's quickstart rather than restyling this page. Naming one frontend also contradicted the Frontend block below, which offers five. The metadata sentence changed with it.

Two tests hold the last two points, so neither returns unnoticed through a later copy edit.

Not taken up here: the feedback about overall page length and competing calls to action. Shortening this page argues against what it is for, and the second draft in [#7013](https://github.com/CopilotKit/CopilotKit/pull/7013) is the short, funnel-shaped answer to the same ticket. The two are meant to be compared, not merged.

## Not in this PR

- **The Loom.** Recording it needs a screen, so it is not something this PR can carry.
- **Sam's or Tyler's approval**, which the ticket lists as delivery criteria.

## Testing

`showcase/shell-docs` is not in the pnpm workspace, so it installs with `npm ci` and tests with `npm run test` from its own directory. The suite is green apart from four files that are already red on `main`: `brand-nav`, `angular-docs-content`, `llm-text`, `ms-agent-python-stable-api`. `tsc --noEmit`, root-level `oxlint` and the production build are clean.

Verified in the browser in both themes, at 375 px and at desktop width.

## Summary by CodeRabbit

- **New Features**
  - Added a product map to the documentation homepage, highlighting frontend, CopilotKit, intelligence, and agent capabilities.
  - Added framework-aware links and framework-specific capability navigation.
  - Added updated capability labels, integration cards, logos, and responsive layouts.

- **Updates**
  - Revised homepage messaging to be framework-neutral and removed React-specific wording.
  - Simplified headings and refreshed descriptive copy.

- **Removals**
  - Removed the legacy landing sample tabs, backend grid, and superseded homepage sections.

- **Tests**
  - Added comprehensive coverage for homepage content, navigation, accessibility, responsive layouts, and framework scoping.


tylerslaton:
Closing in favor of https://github.com/CopilotKit/CopilotKit/pull/7013

PR #7017, opened 2026-09-10
refactor(skills): make copilotkit-debug delegate instead of transcribe

BenTaylorDev:
**1,289 lines to 91.** The skill was a hand-maintained copy of material that each now has one authoritative home — and it named `copilotkit verify` **zero times** across all five of its files.

## What it was doing by hand

Its "Step 1: Gather Information" told the agent to collect package versions, the runtime mode, the transport config, the agent type, and `/info` from the browser network tab. That is what `copilotkit verify` does in one command, across eleven checks, and it reports which URL it probed and where that URL came from — which a manual survey cannot.

The two newest `verify` checks closed the last gaps that made a by-hand pass necessary:

| Gap | Now |
| --- | --- |
| Installed-package version skew | `package_versions_aligned` |
| Browser refusing the runtime's responses | `runtime_cors_allows_frontend` |
| Whether transcription is wired | Reported in the facts block |

## What was duplicating generated output

`error-patterns.md` transcribed `CopilotKitCoreErrorCode`, the legacy error classes, and `TranscriptionErrorCode` — 306 lines. Those are **generated** into `/reference/core/enums/CopilotKitCoreErrorCode`. A transcribed copy of a generated artifact can only ever be less accurate than its generator, so the skill now links it.

Symptom-to-cause material went to `/troubleshooting/error-reference` and `/troubleshooting/common-issues`; event tracing to `/troubleshooting/event-inspector` and `/troubleshooting/debug-mode`. All verified to resolve, including the generated reference page, which I fetched to confirm the route.

## The known-issue list

The clearest case for removing rather than refreshing. Of the twelve issues it cited, **nine are closed** — and the entries read as current, so an agent matching a stalling agent to `#3323` would send someone to upgrade past a fix they already have. Replaced with a live tracker search that includes the version from `verify`'s own report.

## What survives

Only what `verify` genuinely cannot answer:

- **Tool execution** — the round trip asks a question that needs no tools and sends no context, so a passing round trip says nothing about whether yours work.
- **Event ordering and streaming** — `verify` reports pass/fail on a run, not the sequence inside it.
- **State synchronisation** — agent behaviour, not wiring.

Plus the MCP setup block, including the Codex `.codex/config.toml` snippet, which is not documented elsewhere.

## Auditability

`sources.md` no longer lists source paths; it records **where each transcribed section went**, so this is a documented migration rather than a deletion.

That means the skill leaves the package-inventory guard in `public-skill-drift`, which requires at least one resolvable `packages/…` path in a `sources.md`. That guard exists to catch a transcribing skill whose pointers rotted; a skill that transcribes nothing has no inventory to keep true. `copilotkit-setup` still does and stays covered. The `it.each` list carries a comment explaining why the entry left.

## Testing

```
plugin skill mirror in sync

 Test Files  2 passed (2)
      Tests  19 passed (19)

$ oxfmt --check skills/copilotkit-debug/*.md
```

Nothing outside `sources.md` linked to the four deleted references — checked across `skills/`, `packages/*/skills/` and `.claude-plugin/`.

## Note on one thing I did not carry over

I claimed earlier in this effort that the skill's "Key File Locations" table was full of dead paths. That was read from a stale checkout. On `main` all nine paths resolve. The table is dropped anyway, but for relevance rather than rot: it points at CopilotKit's own source, which is contributor material and belongs in `copilotkit-contribute`, not in a skill aimed at someone debugging their own app.

## Summary by CodeRabbit

* **New Features**
  * Added guidance for using `copilotkit verify` to check runtime, agents, credentials, frontend configuration, connectivity, package versions, transcription, realtime wiring, and licensing.
  * Added troubleshooting guidance for verification limitations, error references, known issues, and CopilotKit documentation searches.

* **Documentation**
  * Consolidated debugging information around the verification command and current online references.
  * Removed outdated, duplicated debugging workflows and reference materials.


BenTaylorDev:
Closing as superseded — but the text lives on. The consolidation replaces the whole skill set with two files, one of which is the CLI guide, and this PR's 91 lines are most of it: run `verify --json`, read `checks[]` in order, and here is what `verify` cannot see.

Rather than merge this and delete it days later, the content is being lifted directly into the new CLI skill.

PR #7021, opened 2026-09-10
fix(skills): stop the runtime skill condemning the supported handlers

BenTaylorDev:
The packaged `runtime` skill tells agents, in **five places**, not to use `createCopilotExpressHandler` or `createCopilotHonoHandler`. The source says the opposite:

```ts
/** @deprecated Use `createCopilotExpressHandler` instead. */
export { createCopilotExpressHandler as createCopilotEndpointExpress };

/** @deprecated Use `createCopilotHonoHandler` instead. */
export const createCopilotEndpoint = createCopilotHonoHandler;
```

Those two handlers are **current and supported**. The deprecated names are `createCopilotEndpoint*`, which now alias them. The docs use the handlers as the primary path (12 occurrences) and are correct.

The consequence is not cosmetic: the skill steers agents off the supported API and, in its delegate-by-hand section, onto ~30 lines of manual `Readable.fromWeb` stream plumbing presented as the required alternative.

## The five sites

| Site | Was | Now |
| --- | --- | --- |
| `SKILL.md` frontmatter description | "the Express/Hono adapters are discouraged" | names them as supported |
| `SKILL.md` invariants | "**avoid at all costs**" | states which names are actually deprecated |
| `setup-endpoint.md` | `### CRITICAL Using createCopilotExpressHandler / createCopilotHonoHandler in new code` | removed |
| `setup-endpoint.md` | "Do not use `createCopilotExpressHandler` / `createCopilotHonoHandler`." | reframed — hand-delegating is a real technique, just not the default |
| `setup-endpoint.md` | `### HIGH Using framework adapter on Workers / Bun / Deno` | removed |

The frontmatter description matters most: it is what an agent reads to decide whether to load the skill at all.

The Workers/Bun/Deno gotcha went because the adapters are not the problem there — `runtime-server-adapter.mdx` carries dedicated sections for Bun, Deno, Cloudflare Workers **and** Elysia. The skill's claim conflates "Express needs Node" with "adapters don't work on edge runtimes".

## A second false claim, same skill

`server-side-tools.md` said a server tool's call never reaches the browser, so nothing can render it. Both `TOOL_CALL_START` emissions in `packages/runtime/src/agent/index.ts` are driven by the model's stream parts — `tool-input-start` and `tool-call` — with **no filtering** on whether the tool has a server-side `execute`. So the call does stream, and a render-only `useRenderTool` registration can give it UI while `execute` stays on the server.

Rewritten around the distinction that holds: `execute` runs on the server and cannot touch the DOM or component state.

## Why patch a skill that is slated for deletion

The removal is blocked on two hazard groups still needing documented homes, plus an information-architecture decision about `built-in-agent.mdx`. That is not days. Leaving three verified-false claims live through the migration was the worse option — every agent that loads this skill today is told to avoid the current API.

This changes no plan. The skill still goes once its hazards have homes.

## Testing

```
plugin skill mirror in sync

 Test Files  2 passed (2)
      Tests  20 passed (20)

$ oxfmt --check packages/runtime/skills/runtime/**/*.md
All matched files use the correct format. (28 files)
```

Both copies updated through the sync script — `packages/runtime/skills/` is the source, `skills/` the mirror. A final sweep for the condemning language across the whole skill comes back empty, and the five surviving gotchas in `setup-endpoint.md` are intact and correctly ordered after the two removals.

## Summary by CodeRabbit

- **Documentation**
  - Updated runtime guidance to recognize Express and Hono adapters as supported mounting options.
  - Clarified when to use framework adapters versus the fetch-native handler.
  - Documented deprecated endpoint names and their supported replacements.
  - Clarified that server-side tools can provide browser-rendered UI through render-only tool registrations.
  - Removed outdated guidance discouraging framework adapters and incorrectly limiting server-tool rendering.

### DIFF
--- a/packages/runtime/skills/runtime/SKILL.md
+++ b/packages/runtime/skills/runtime/SKILL.md
@@ -7,8 +7,8 @@ description: >
   LangGraph, CrewAI Crews/Flows, PydanticAI, ADK, LlamaIndex, Agno, AWS Strands, MS Agent
   Framework, AG2, A2A), enable Intelligence mode for durable threads + websocket,
   register server-side tools via defineTool, and wire voice transcription. Uses the
-  fetch-based createCopilotRuntimeHandler primitive — the Express/Hono adapters are
-  discouraged. Load the reference under references/ that matches your task.
+  fetch-based createCopilotRuntimeHandler primitive, or the supported Express/Hono
+  adapters. Load the reference under references/ that matches your task.
 type: core
 library: copilotkit
 library_version: "1.71.0"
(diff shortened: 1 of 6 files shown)


BenTaylorDev:
Closing as superseded. This was explicitly a stopgap — "a correctness patch, not a reprieve" — for a skill that the consolidation now deletes outright.

The finding stands and is carried into the refactor: `createCopilotExpressHandler` and `createCopilotHonoHandler` are the current supported adapters, and it is `createCopilotEndpoint*` that is deprecated in favour of them. Nothing in the replacement skills will condemn them, and the docs already use them as the primary path.

Also worth noting for anyone who lands here from the red checks: the three failing jobs are pre-existing on `main` — the same `langgraph-*` jobs fail there with 27 occurrences of `Event trace fixture … is a Git LFS pointer, not the real file`. Unrelated to this diff, which is markdown only.

PR #7023, opened 2026-09-10
fix(skills): repoint 60 source citations at the flat package layout

BenTaylorDev:
Mechanical audit of **every** skill. Two passes that are cheap and reliable; the expensive semantic pass is scoped at the end.

## Scope, measured

The skill set is larger than earlier notes in this effort suggested: **17 skills, 21,199 lines, 316 citations.** Six skills had never been looked at (`setup-slack-channel` alone is 1,340 lines).

## Pass 1 — citation mechanics

**60 of 316 citations pointed at files that do not exist.** All 60 used the retired `packages/v2/…` / `packages/v1/…` layout the flat restructure replaced.

| Skill | Cited | Broken |
| --- | --- | --- |
| `copilotkit-develop` | 54 | **54** |
| `copilotkit-contribute` | 4 | 4 |
| `copilotkit-upgrade` | 21 | 2 |
| `react-core` | 96 | 0 |
| `runtime` | 80 | 0 |
| `copilotkit-agui` | 27 | 0 |
| `a2ui-renderer`, `copilotkit-channels`, `copilotkit-setup`, `copilotkit-debug` | 34 | 0 |

The distribution is the finding: `react-core` and `runtime` are clean, while **every one** of `copilotkit-develop`'s 54 is dead. Its manifest describes a source tree that has not existed since the restructure, so it has not been regenerated since.

## How they were resolved

By basename against the real tree, with the origin choosing the root (`v2/react` → `react-core`, `v2/core` → `core`, `v2/runtime` → `runtime`), and **every replacement checked to exist before being written**.

The chat components needed more than a reroot — the rename dropped hyphens and changed case, so `copilot-chat.tsx` is now `CopilotChat.tsx`.

Seven had to be located by *symbol* rather than name, because the type they document moved house:

| Cited | Now | Confirmed by |
| --- | --- | --- |
| `v2/react/src/types/labels.ts` | `v2/providers/CopilotChatConfigurationProvider.tsx` | `CopilotChatLabels` |
| `v2/react/src/types/slot.ts` | `v2/lib/slots.tsx` | `SlotValue` |
| `v2/react/src/types/tool-call-renderer.ts` | `v2/types/react-tool-call-renderer.ts` | `ReactToolCallRenderer` |
| `v2/react/src/types/activity-message-renderer.ts` | `v2/types/react-activity-message-renderer.ts` | `ReactActivityMessageRenderer` |
| `v2/core/src/types/tool-call.ts` | `core/src/types.ts` | `ToolCallStatus` |
| `v2/runtime/src/types/runtime-options.ts` | `runtime/core/runtime.ts` | `CopilotRuntimeOptions` |
| `v2/*/vitest.config.mjs` | the matching package's config | env/setup details in the citation |

### One resolved and was still wrong

`copilotkit-upgrade` cited `react-core/src/index.tsx` for the v1 hook exports. Dropping the `x` made the path exist — and that file exports **none** of them. It is a compatibility shim whose own header points at `v1-deprecated-compatibility.ts`, which is where the inventory lives. Repointed there and confirmed by content.

Existing is not the same as supporting the claim, which is why every override in this PR was verified by symbol rather than by `existsSync`.

## Pass 2 — symbol existence

Every `useXxx(` and `<CopilotXxx` named in any skill, against the 8,253 exported identifiers in shipped `src`. **One name does not exist:** `CopilotKitWebInspector`, in three `copilotkit-debug` files that #7017 deletes. The other three flags are false positives — `useAgents` is correctly documented as *not existing*, and `useAvailableAgents` / `useMyFeatureFlag` are defined in the docs' own examples.

`copilotkit-agui`'s 27 citations are upstream `ag-ui/sdks/typescript/…` paths, correctly labelled as such by its `sources.md` header. Clean.

## Testing

```
$ python3 cite_check.py
citations: 316   in-repo: 289   upstream (not resolved here): 27
mechanically broken: 0

plugin skill mirror in sync

$ vitest run scripts/__tests__/{public-skill-drift,sync-plugin-skills}.test.ts
 Test Files  2 passed (2)
      Tests  20 passed (20)

$ oxfmt --check skills/**/*.md
All matched files use the correct format. (105 files)
```

## What this does not do

This is the mechanical layer. **It proves the pointers resolve; it does not prove the prose is true.** Everything found by reading in this effort — false claims about deprecated handlers, a wrong statement about server-tool rendering, a misleading `identifyUser` comment — was invisible to both passes here.

The remaining semantic pass is roughly 15,800 lines. My own tooling manufactured false findings three separate times during this audit (a truncating regex twice, a claim-attribution heuristic once), so I would not trust a scripted verdict on that layer. The highest-value targets, on the evidence so far: `copilotkit-develop` (1,416 lines, and now demonstrably un-regenerated), `copilotkit-integrations` (2,140 lines, zero citations, so nothing anchors it), and `setup-slack-channel` (1,340 lines, never reviewed).

## Summary by CodeRabbit

* **Documentation**
  * Updated testing guidance with current locations for mock utilities and Vitest configurations.
  * Refreshed development references to relocated React, core, and runtime source files.
  * Updated upgrade documentation to point to the current v1 compatibility exports.
  * Corrected renamed component, slot, middleware, and renderer references.


BenTaylorDev:
Closing as superseded. 58 of the 60 repointed citations live in `copilotkit-develop` (54), `copilotkit-contribute` (4) and `copilotkit-upgrade` (2), all of which are being deleted in the skills consolidation — `react-core` and `runtime` were already clean, so nothing here survives the refactor.

The findings are worth keeping on the record even though the fix isn't: every one of `copilotkit-develop`'s 54 source pointers was dead, which is how we know that skill was never regenerated after the flat-package restructure. And one citation resolved while still being wrong — `react-core/src/index.tsx` → `index.ts` made the path exist, but that file exports none of the v1 hooks it was cited for.

PR #1180, opened 2025-01-02
fix: allow CopilotKit Remote Endpoint without agents

suhasdeshpande:
## What does this PR do?

This change allows user to configure non-agent CopilotKit Remote Endpoint.  

Screenshot 2025-01-02 at 12 44 21 PM

## Related PRs and Issues

- (Direct link to related PR or issue, if relevant)

## Checklist

- [x] I have read the [Contribution Guide](https://github.com/copilotkit/copilotkit/blob/master/CONTRIBUTING.md)
- [x] If the PR changes or adds functionality, I have updated the relevant documentation

PR #7029, opened 2026-09-10
refactor(skills): replace nine knowledge skills with two entry points

BenTaylorDev:
## What

Replaces nine packaged knowledge skills with two entry points.

| | |
| --- | --- |
| `copilotkit` | The `copilotkit-docs` MCP server: four search tools and two explore tools, which corpus answers which question, and the instruction not to answer from memory. |
| `copilotkit-cli` | The CLI, led by `copilotkit verify --json`. |

**Deleted:** `copilotkit-setup`, `copilotkit-develop`, `copilotkit-integrations`, `copilotkit-debug`, `copilotkit-upgrade`, `copilotkit-agui`, `copilotkit-contribute`, `copilotkit-self-update`, and the three package-generated skills (`react-core`, `runtime`, `a2ui-renderer`).

Net: **-28,851 / +335 lines.**

**Kept, and out of scope here:** `setup-slack-channel`, `channels-setup`, `copilotkit-channels`, `inspector-docs`, `inspector-workbench`, `intelligence-docs`. The first three are procedures rather than knowledge; the last three are internal maintenance, not part of the public product. Their disposition is a separate call.

## Why

The packaged skills had become a second copy of the documentation, and a copy goes stale silently.

- `runtime` and `react-core` were roughly 60% transcribed API surface. Of the remaining "Common Mistakes" prose, most already existed on a docs page.
- Four claims in the deleted skills contradicted the source they cited. One of them reached a shipped PR (#7005) before it was caught, and had to be corrected in `14231a6169` — the skill asserted that server tools cannot render UI because the browser never sees a tool-call start event, which `packages/runtime/src/agent/index.ts:1519,1584` contradicts.
- `copilotkit-debug` was 1,289 lines of hand-written triage. Since #1180, `copilotkit verify` covers version skew, CORS, and transcription directly, which is most of what those lines described.
- Nothing in telemetry or in the issue tracker suggests the per-lifecycle split was load-bearing for anyone.

The gotchas worth keeping were moved onto docs pages first, in #6993, #6997 and #7021's predecessors, so this PR removes rather than discards.

## Also in this PR

- `skills` dropped from the `files` field of `@copilotkit/react-core`, `@copilotkit/runtime` and `@copilotkit/a2ui-renderer`, so the tarballs stop carrying a copy.
- `RESERVED_LIFECYCLE_SLUGS` in `scripts/sync-plugin-skills.ts` goes from 14 entries to 8.
- `.claude-plugin/plugin.json` and `marketplace.json` descriptions no longer advertise the deleted lifecycle.
- `/build-with-agents` (56,393 distinct viewers in 90 days) named three deleted skills in a table and told the reader to "Use the copilotkit-setup skill to get started". Rewritten for the two-skill model.
- `README.md` and the two stale cross-references in `copilotkit-channels/SKILL.md` updated.
- One dead `skills/copilotkit-setup/SKILL.md` entry removed from `ALIAS_ALLOWLIST` in `scripts/validate-intelligence-env-names.ts`.

`scripts/__tests__/public-skill-drift.test.ts` guarded specific wording inside files that no longer exist. It is now a link guard instead: every `docs.copilotkit.ai` path a packaged skill names has to resolve to a page in this repo, and the two entry points have to stay free of a transcribed API surface (asserted as: no `sources.md` other than the channels skill's).

## Supersedes

Closed in favour of this one, with a comment on each: #7017 (the CLI skill, whose content seeded `copilotkit-cli`), #7021, #7023.

## Needs a decision from someone else

- **#5689** (@onsclom) is a skill-lift eval for `copilotkit-setup`, which this PR deletes. No file overlap, so it will not conflict, but the eval target is gone. Austin's call whether to retarget it at `copilotkit` or drop it.
- **#6636** (community, for #6125) fixes `packages/react-core/skills/react-core/references/agent-access.md`, the one file it touches, which this PR deletes. Needs a reply to the author explaining why, not a silent close.
- **Install-command drift, unresolved.** `README.md` and `showcase/.../cli/cli.mdx` say `npx copilotkit@latest skills install`; `/build-with-agents` says `npx skills add CopilotKit/CopilotKit/skills -y`. Both commands work. Picking one retires a shipped CLI command, which belongs in the Intelligence repo, so this PR leaves both in place and changes only the prose around them.

## Testing

The worktree has no `node_modules`, so `lefthook` could not resolve `tsx`, `oxlint` or `nx` and the commit used `--no-verify`. Every gate the hook would have run was run by hand against the repo's own binaries, and the pre-commit failure output was the "command not found" line in each case, not a gate verdict.

`pnpm check:plugin-skills` — the mirror gate, which also enforces orphan deletion:

```
plugin skill mirror in sync
```

`pnpm check:intelligence-env-names`:

```
```

Both script suites:

```
 Test Files  2 passed (2)
      Tests  16 passed (16)
```

**Mutation-checked the new link guard**, so it is not passing for the wrong reason. Appending `See [nope](https://docs.copilotkit.ai/this-page-does-not-exist).` to `skills/copilotkit/SKILL.md`:

```
⎯⎯⎯⎯⎯⎯⎯ Failed Tests 1 ⎯⎯⎯⎯⎯⎯⎯
+   "skills/copilotkit/SKILL.md: /this-page-does-not-exist",
 Test Files  1 failed (1)
      Tests  1 failed | 3 passed (4)
```

Restored, back to 4 passed. The guard already caught one real mistake of mine while I was writing the skill: I had linked `/custom-look-and-feel` from the nav list without checking, and the page is `/custom-look-and-feel/css`.

`oxfmt` on the four changed TypeScript files: no changes. Both plugin manifests re-parsed as JSON after editing.

MDX: `build-with-agents.mdx` changes are prose, one table body and one fence body. Its JSX tag inventory is byte-identical to `origin/main`:

```
$ diff <(git show origin/main:$f | grep -o '</\?[A-Z][A-Za-z]*' | sort | uniq -c) \
       <(grep -o '</\?[A-Z][A-Za-z]*' $f | sort | uniq -c) && echo IDENTICAL
IDENTICAL
```

`https://docs.copilotkit.ai/build-with-agents` returns 200, so the new README link resolves.

Not run: the showcase docs build, which needs the per-entry `node_modules` symlinks and the pretest generators this worktree does not have. CI covers it.

## Summary by CodeRabbit

- **New Features**
  - Added `copilotkit` skill for searching current documentation and source.
  - Added `copilotkit-cli` skill for project verification, onboarding, authentication, scaffolding, and CLI commands.
  - Added guidance to use `copilotkit verify` to validate project wiring.

- **Documentation**
  - Updated marketplace, plugin, README, and agent-building guidance for the new skills.
  - Updated channel guidance to reference CLI-based setup.

- **Removals**
  - Removed legacy setup, development, integration, debugging, upgrade, contribution, and self-update skills.
  - Package distributions no longer include bundled skill documentation.


BenTaylorDev:
## Audit pass — six wrong claims found in my own skill files, now fixed

Re-verified every factual claim in both new skills against current source rather than against the draft. `db096f055a`.

**`copilotkit`**

- **The MCP server never reached a skills-only install.** `.mcp.json` sits at the repository root, outside `skills/`, so `npx skills add CopilotKit/CopilotKit/skills` — the command `/build-with-agents` tells readers to run — registers no server, while the skill said "configured by this plugin's `.mcp.json`. Nothing to do." The skill whose entire purpose is searching the docs was inert for exactly the install path the docs recommend. It now gives `claude mcp add` and `codex mcp add`, matching the docs' own MCP page.
- The Codex block named a project-local `.codex/config.toml` and a `type` key the documented configuration does not use.
- **"eleven things" was a ceiling, not a count.** `frontendAssetChecks`, `corsChecks` and `packageVersionChecks` each return `[]` when nothing could be read (`verify-checks.ts:1100,1175,1293`), so an Intelligence-mode run settles eight to eleven. `--expect-runtime oss` drops the hosted-project and credential checks entirely.
- Dropped "the mismatch shows up as routes that 404". The `/v2` subpaths and the still-resolving package root are verified in the `exports` maps; that specific symptom was not.

**`copilotkit-cli`**

- **`login --json` was missing** — the flag that streams agent-readable JSON without launching a browser. A skill written for coding agents omitted the one login path an agent can complete.
- `create` was presented as canonical with `init` as the alias. The CLI's help has it the other way round.
- Added the provenance verdict rule, the actionable half of the feature the skill already praised: nothing answering at a URL **the project named** is a FAIL; nothing answering at an **assumed** default is UNKNOWN.
- Added `--runtime-url`, `--header`, `--timeout`, the `--expect-runtime oss` exit-zero condition, and a `skills install` row.
- Replaced "it does not detect or convert an app you already have" with what is supported — I never proved that absence.

**Also fixed:** the `check-plugin-skills` glob in `lefthook.yml` still enumerated `skills/runtime/**`, `skills/react-core/**` and `skills/a2ui-renderer/**`, all deleted here — so editing either new skill fired no local mirror check. `plugin-skills-check.yml` already uses `skills/**` and carries a comment about this exact mistake.

### Claims that checked out

The four corpora and their boundaries (the tool description states "library/package source only — NOT example or showcase apps"); `--round-trip` sending `tools: []` and `context: []` under the comment "Deliberately trivial and answerable without tools or context" (`round-trip-probe.ts:98,195-207`); `whoami` printing `Organization:`; `project select` writing `CPK_INTELLIGENCE_API_KEY` to `.env` (`project-select-effects.ts:119`); `import --dry-run`; both telemetry opt-out variables; `RESERVED_LIFECYCLE_SLUGS` matching the eight remaining directories exactly; and all 21 docs links, each returning 200 live.

### Two of my own verification steps were also broken

- **The earlier "oxfmt: no changes" was vacuous.** I passed the file list as an unquoted zsh scalar, which does not word-split, so oxfmt received one unmatchable argument and reported "0 files". That is why the auto-format bot, not I, caught the markdown. Re-run with literal arguments: 11 files checked, `--write` a proven no-op.
- **The first CI poll declared "settled" on an empty check set** after the bot's push emptied the rollup. Re-polled with a guard requiring a populated set.

### CI

**73 pass, 3 skipped, 3 fail.** The three failures are `test / e2e / dojo` (`langgraph-python`, `langgraph-fastapi`, `langgraph-typescript`), and they are pre-existing on main, not caused by this PR:

- main's latest run of the same workflow (`34501181229`) fails the same three jobs with the same cause, 27 occurrences of `Event trace fixture .../ag-ui/apps/dojo/e2e/fixtures/test-image.png is a Git LFS pointer`
- the workflow was green on main through 2026-09-09 21:26 and has failed on every main run since 2026-09-10 08:21
- this diff touches no `ag-ui/` or e2e path

MikeRyanDev (approved):

PR #7033, opened 2026-09-10
fix(core): keep the runtimeUrl trailing slash for single-route requests

### LINKED ISSUE #7028: 🐛 Bug: runtimeUrl prop removes trailing slashes
### ♻️ Reproduction Steps

Using @copilotkit/react-core v2 on version 1.70.1, configure CopilotKitProvider with a runtime URL that ends in a trailing slash:

```tsx
import { CopilotKitProvider } from "@copilotkit/react-core/v2";

export function App() {
  return (
    <CopilotKitProvider
      runtimeUrl="https://example.com/service/copilotkit/"
    >
      {/* Application components */}
    </CopilotKitProvider>
  );
}
```

1. Open the browser developer tools.
2. Send a CopilotKit message.
3. Inspect the request URL in the Network tab.
4. Observe that CopilotKit sends the request to:
`https://example.com/service/copilotkit`
instead of the URL supplied to the provider:
`https://example.com/service/copilotkit/`

The URL appears to be modified by this normalization:
```tsx
const normalizedRuntimeUrl = runtimeUrl
  ? runtimeUrl.replace(/\/$/, "")
  : undefined;
```

Sources: 
https://github.com/CopilotKit/CopilotKit/blob/main/packages/core/src/core/agent-registry.ts#L283-L302
https://github.com/CopilotKit/CopilotKit/blob/main/packages/core/src/agent.ts#L109-L128

### ✅ Expected Behavior

CopilotKitProvider should use the exact runtimeUrl supplied by the caller, including its trailing slash: `https://example.com/service/copilotkit/`

A trailing slash can be significant to proxies, gateways, and web servers. For example, /service/copilotkit and /service/copilotkit/ may be handled by different Nginx locations.
If CopilotKit needs to append a path such as /info, it should join the URL and path safely without modifying the stored, caller-provided URL.

### ❌ Actual Behavior

CopilotKit silently strips the trailing slash from runtimeUrl:

```
Provided: https://example.com/service/copilotkit/
Requested: https://example.com/service/copilotkit
```

This changes the endpoint selected by our proxy configuration. In our case, the modified URL falls through to a different Nginx location and the CopilotKit POST request receives:`405 Method Not Allowed`

The application cannot work around this through the runtimeUrl prop because the slash is removed internally by the library, after the value is supplied.

### Screenshots

_No response_

### 𝌚 CopilotKit Version

```shell
@copilotkit/react-core@1.70.1
@copilotkit/react-ui@1.70.1
```

### 📄 Logs (Optional)

```shell
Request URL: https://example.com/service/copilotkit
Request Method: POST
Status Code: 405 Method Not Allowed
```

L4XB:
## What does this PR do?

`runtimeUrl` was slash-stripped as soon as it was received, both in `AgentRegistry.setRuntimeUrl` and in the `ProxiedCopilotRuntimeAgent` constructor. That is right for path joins, but the single-route transport uses the runtime URL *itself* as the request target, so a provider configured with `https://host/service/copilotkit/` sent its requests to `https://host/service/copilotkit`. A trailing slash can select a different proxy or gateway location, so the endpoint has to be used exactly as given (#7028).

Changes in `packages/core`:

- `agent.ts`: the single-route `runUrl` and `singleEndpointUrl` use `config.runtimeUrl` verbatim; the slash-stripped form is still what `/agent/<id>/run`, `/info`, `/stop` and `/connect` are joined onto, so REST URLs never get a double slash.
- `core/agent-registry.ts`: the registry keeps the caller's URL alongside the normalized one (`_runtimeEndpointUrl`), hands it to the proxied agents it creates, and uses it for the single-route identification request (`fetchRuntimeInfoSingle`), including the REST-to-single fallback of the `auto` transport. Change detection in `setRuntimeUrl` now also notices a slash-only change. The public `runtimeUrl` getter and the connection key are unchanged.

No behaviour change for a `runtimeUrl` without a trailing slash.

## Related PRs and Issues

- Fixes #7028

## Tests

- `proxied-runtime-transport.test.ts`: with `https://runtime.example/service/copilotkit/`, a single-route run posts to exactly that URL, and a REST run goes to `…/copilotkit/agent/agent/run`.
- `agent-registry-runtime-url-trailing-slash.test.ts`: a `CopilotKitCore` with that URL sends its single-route identification request to the verbatim URL and creates the runtime agent with it; with the REST transport the identification request is `…/copilotkit/info` and the agent's run URL has no double slash.

The single-route cases fail on `main`. `vitest run` in `packages/core`: 843 passed; `tsc --noEmit` clean. No changeset added (the repo no longer uses them).

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed runtime URL handling when configured URLs include trailing slashes.
  - Single-route connections now preserve the provided endpoint URL exactly.
  - REST connections build runtime and agent paths without duplicate slashes.
  - Improved runtime discovery, endpoint updates, cloning, and agent connections across supported transport modes.

- **Tests**
  - Added coverage for trailing-slash URLs, runtime information requests, agent routing, endpoint changes, cloning, and automatic transport handling.

### DIFF
--- a/packages/core/src/__tests__/agent-registry-runtime-url-trailing-slash.test.ts
+++ b/packages/core/src/__tests__/agent-registry-runtime-url-trailing-slash.test.ts
@@ -54,6 +54,30 @@ describe("runtimeUrl with a trailing slash", () => {
     expect((agent as ProxiedCopilotRuntimeAgent).url).toBe(runtimeUrl);
   });
 
+  it("rebuilds the proxy when only the trailing slash changes", async () => {
+    const { core, fetchMock } = setupCore(runtimeUrl, "single");
+    await vi.waitFor(() => {
+      expect(core.getAgent("default")).toBeDefined();
+    });
+    const before = core.getAgent("default") as ProxiedCopilotRuntimeAgent;
+
+    fetchMock.mockResolvedValueOnce(runtimeInfoResponse());
+    core.setRuntimeUrl("https://runtime.example/service/copilotkit");
+    await vi.waitFor(() => {
+      const agent = core.getAgent("default") as
+        | ProxiedCopilotRuntimeAgent
+        | undefined;
+      expect(agent).toBeDefined();
+      expect(agent).not.toBe(before);
+    });
+
+    const after = core.getAgent("default") as ProxiedCopilotRuntimeAgent;
+    expect(after.url).toBe("https://runtime.example/service/copilotkit");
+    expect(fetchMock.mock.calls[1]?.[0]).toBe(
+      "https://runtime.example/service/copilotkit",
+    );
+  });
+
   it("joins REST paths without a double slash", async () => {
     const { core, fetchMock } = setupCore(runtimeUrl, "rest");
 
--- a/packages/core/src/__tests__/agent-registry-runtime-url-trailing-slash.test.ts
+++ b/packages/core/src/__tests__/agent-registry-runtime-url-trailing-slash.test.ts
@@ -1,0 +1,72 @@
+import type { RuntimeInfo } from "@copilotkit/shared";
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { ProxiedCopilotRuntimeAgent } from "../agent";
+import { CopilotKitCore } from "../core";
+
+/** A Runtime `/info` answer advertising one agent. */
+function runtimeInfoResponse(): Response {
+  const info: RuntimeInfo = {
+    version: "1.0.0",
+    agents: {
+      default: {
+        name: "default",
+        className: "HttpAgent",
+        description: "assistant",
+      },
+    },
+    audioFileTranscriptionEnabled: false,
+    mode: "sse",
+  };
+  return new Response(JSON.stringify(info), {
+    status: 200,
+    headers: { "content-type": "application/json" },
+  });
+}
+
+function setupCore(runtimeUrl: string, runtimeTransport: "rest" | "single") {
+  const fetchMock = vi.fn<typeof globalThis.fetch>();
+  fetchMock.mockResolvedValueOnce(runtimeInfoResponse());
+  vi.stubGlobal("window", {});
+  vi.stubGlobal("fetch", fetchMock);
+  const core = new CopilotKitCore({ runtimeUrl, runtimeTransport });
+  return { core, fetchMock };
+}
+
+describe("runtimeUrl with a trailing slash", () => {
+  const runtimeUrl = "https://runtime.example/service/copilotkit/";
+
+  afterEach(() => {
+    vi.unstubAllGlobals();
+    vi.restoreAllMocks();
+  });
+
+  it("keeps the caller's URL verbatim as the single-route endpoint", async () => {
+    const { core, fetchMock } = setupCore(runtimeUrl, "single");
+
+    await vi.waitFor(() => {
+      expect(core.getAgent("default")).toBeDefined();
+    });
+
+    // The single-route identification request targets that same endpoint.
+    expect(fetchMock.mock.calls[0]?.[0]).toBe(runtimeUrl);
+    const agent = core.getAgent("default");
+    expect(agent).toBeInstanceOf(ProxiedCopilotRuntimeAgent);
+    expect((agent as ProxiedCopilotRuntimeAgent).url).toBe(runtimeUrl);
+  });
+
+  it("joins REST paths without a double slash", async () => {
+    const { core, fetchMock } = setupCore(runtimeUrl, "rest");
+
+    await vi.waitFor(() => {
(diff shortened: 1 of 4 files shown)


L4XB:
Follow-up pushed for the three review points: the agent now keeps the caller's URL for every transport (`runtimeEndpointUrl`) and uses it for the auto-transport single-route fallback, the stored endpoint and `clone()`; the registry's connection key and per-proxy connection record include the raw endpoint, so a change that only adds or drops the trailing slash rebuilds the proxy; the single-route `inspector/metadata` request goes through `singleEndpointUrlFor`. Tests cover the auto fallback, cloning and the slash-only URL change (846 passed in `packages/core`, `tsc --noEmit` clean).

BenTaylorDev (changes requested):
Reviewed this against current `main` and ran it locally: `packages/core` 846 passing, `tsc --noEmit` clean, `oxlint`/`oxfmt` clean. The endpoint/normalized split is the right shape, and both CodeRabbit findings look properly handled in `b3d258b`.

One call site is still on the stripped URL: `agent-registry.ts:631` passes `this._runtimeUrl` into `createSingleRouteResourceRequest`, which uses that argument verbatim as the POST target. So Intelligence resource operations (threads, memories, `/annotate`) still hit the slash-free URL — the same 405 as #7028, just a different request family.

```diff
               ? await createSingleRouteResourceRequest(
                   input,
                   init,
-                  this._runtimeUrl,
+                  this._runtimeEndpointUrl ?? this._runtimeUrl,
                 )
```

The `runtime.pathname.replace(/\/$/, "")` inside that function is prefix-matching and should stay.

A probe that fails on the current head and passes with the change:

```ts
const RUNTIME_URL = "https://runtime.example.com/api/copilotkit/";
// connect single-route with singleRoute.resourceOperations: true, then:
await core.ɵruntimeFetch(`${RUNTIME_URL}threads`, { method: "GET" });
expect(fetchMock.mock.calls[0]?.[1]?.method).toBe("POST"); // envelope was built
expect(fetchMock.mock.calls[0]?.[0]).toBe(RUNTIME_URL);    // fails: strips the slash
```

Asserting the `POST` first matters — without it the test passes even when the envelope path is never taken. Full suite stays green with the change (847 with the probe added).

Also worth a look: #7064 touches the same resource-transport area, so whichever lands second will want a re-check.

L4XB:
Good catch — fixed in `44ce797`.

You are right that `agent-registry.ts` was the one call site left on the stripped URL, and the reason is exactly as you describe: `createSingleRouteResourceRequest` uses its third argument verbatim as the POST target, so resource operations met the same 405 as #7028, one request family later. Applied your diff:

```ts
? await createSingleRouteResourceRequest(
    input,
    init,
    this._runtimeEndpointUrl ?? this._runtimeUrl,
  )
```

which is the form the other endpoint-as-target sites already use (`:505`, `:1340`), and left `runtime.pathname.replace(/\/$/, "")` inside the helper alone — it is prefix matching, as you say.

Your probe is in as `keeps the slash for Intelligence resource requests`, with the `POST` assertion first for the reason you gave. I checked that the ordering actually earns its place rather than taking it on faith: with `singleRoute.resourceOperations` flipped to `false` — the envelope path never taken — the case fails on `expected 'GET' to be 'POST'` before it ever reaches the URL comparison. And with the fix reverted to `this._runtimeUrl` it fails on the URL, while the other three cases in that file stay green.

```
packages/core   847 passed (72 files)      # your predicted count
tsc --noEmit    clean
oxlint          0 warnings, 0 errors on the two touched files
oxfmt --check   clean
```

On #7064: it changes `single-route-resource-request.ts` itself while this changes the argument the registry hands it, so the two do not overlap textually. They do share the meaning of that argument, though — if #7064 lands first I will re-run this against it and say so here rather than assume it still holds.

BenTaylorDev (approved):
Re-reviewed against current `main` (284 commits ahead of this branch; nothing on main touches `agent.ts` or `agent-registry.ts`). Cherry-picked cleanly: `packages/core` 850 passing, `check-types` clean, `oxlint` 0 errors, `oxfmt` clean.

`44ce797a0` fixes the resource-envelope call site I flagged, and the new case follows the assert-`POST`-before-target order — I mutation-checked it by reverting line 638, and it fails on `expect(target).toBe(runtimeUrl)` while the `POST` assertion still passes, so it is not vacuous.

#7064 has since merged into the same area. It is disjoint: `createSingleRouteResourceRequest` still returns `input: runtimeUrl` verbatim, so the raw endpoint is still the right thing to hand it.

One non-blocking note: the failure messages at `agent-registry.ts:1049-1050` and the telemetry context at `:1056` still report the slash-stripped `_runtimeUrl`. Cosmetic, but for this exact bug class the error text shows a URL the caller never configured. Fine to leave for a follow-up.

PR #7059, opened 2026-09-10
fix(docs): load env files for the standalone SPA runtime

tylerslaton:
## Problem

An onboarding run using CLI 4.9.1 reported that the standalone React SPA runtime never reads the `.env` written during onboarding.

## Why

`tsx server.ts` does not load `.env`. Vite loading the file in a separate process does not configure the runtime.

## Fix

Start with `tsx --env-file=.env server.ts`, require Node 20.6+, and explain env-file location and server-only credentials. Changes are confined to the React SPA guide.

Validation: ran the same TypeScript probe beside a `.env` with dummy model and Intelligence credentials. Before: `{"intelligenceConfigured":false,"modelConfigured":false}`. After adding the documented flag: `{"intelligenceConfigured":true,"modelConfigured":true}`. React SPA MDX compilation and `git diff --check` pass. No live credentials or model requests used. Full docs build not run; this checkout has no shell-docs Nx target.

A broader earlier proposal (#6831) was closed without merging; this PR addresses only this run's standalone runtime env-loading issue.

## Summary by CodeRabbit

- **Documentation**
  - Updated React SPA prerequisites to require Node.js 20.6 or later.
  - Added instructions for starting the application with runtime environment variables from a `.env` file.
  - Documented how to configure runtime and Intelligence credentials.
  - Added guidance for keeping sensitive credentials out of browser-exposed variables.
  - Added troubleshooting information for missing runtime configuration.

Review follow-up: the guide now adds `.env` to `.gitignore` before creating credentials. Verified against the official Vite template: `git check-ignore .env` fails before the added rule and succeeds afterward. MDX compilation and whitespace checks pass.

### DIFF
--- a/showcase/shell-docs/src/content/docs/frontends/react-spa.mdx
+++ b/showcase/shell-docs/src/content/docs/frontends/react-spa.mdx
@@ -15,7 +15,7 @@ Exactly one instruction does not carry over: **where Copilot Runtime lives.** Th
 
 - An OpenAI API key (or another model provider supported by [Model Selection](/model-selection))
 - React 18+
-- Node.js 20+
+- Node.js 20.6+ (for `--env-file`)
 
 ## Getting started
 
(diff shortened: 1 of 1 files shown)


BenTaylorDev (approved):
Verified against current `origin/main`.

Follow-up, not a blocker: I measured Node v22.23.2, and `--env-file` does not override an exported variable. The shell value wins. So the new troubleshooting line does not cover a stale exported `OPENAI_API_KEY`, which is the case #7066 describes. Those two need reconciling on this page.

PR #7061, opened 2026-09-10
fix(docs): correct LangGraph TypeScript context state key

tylerslaton:
## Problem

The LangGraph TypeScript app-context example reads `state.copilotKit.context`, but the SDK schema and CopilotKit runtime adapter supply lowercase `copilotkit`.

## Why

The TypeScript spelling differs from the actual state field. Following the example throws before it can consume frontend context.

## Fix

Correct the expression and explain the lowercase state key. One docs file; no runtime behavior change.

Validation: sent a synthetic colleague through the real CopilotKit LangGraph adapter's state merge and a compiled LangGraph node using the expression extracted from this page. Before: `TypeError: Cannot read properties of undefined (reading 'context')`. After: the node reads colleague `Ada`. MDX syntax compilation passes after stripping snippet imports, as the docs loader does. Git whitespace and commit hooks pass. No hosted browser/model call was needed for this deterministic reproduction; full docs build was not run.

## Summary by CodeRabbit

- **Documentation**
  - Clarified that the CopilotKit state key is lowercase `copilotkit` in Python and TypeScript integrations.
  - Updated the TypeScript example to access context using the correct lowercase state key.

### DIFF
--- a/showcase/shell-docs/src/content/docs/integrations/langgraph/agent-app-context.mdx
+++ b/showcase/shell-docs/src/content/docs/integrations/langgraph/agent-app-context.mdx
@@ -103,6 +103,10 @@ This context can then be shared with your LangGraph agent.
                 The state of a LangGraph agent is the "hub" for applicative information used by the agent.
                 Naturally, the context from CopilotKit will be injected there.
 
+                In both Python and TypeScript, the CopilotKit state key is lowercase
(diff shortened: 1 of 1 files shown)


BenTaylorDev (approved):
Verified. The SDK state schema uses lowercase `copilotkit` (`sdk-js/src/langgraph/__tests__/state-schema.test.ts:11` and `middleware.ts:70`), so the old spelling did throw.

PR #7063, opened 2026-09-10
fix(docs): make ADK setup examples self-contained

tylerslaton:
## Problem

The Google ADK frontend-tool setup, embedded in `/google-adk/generative-ui/tool-based`, imports `get_model` and `stop_on_terminal_text` from an undefined `agents.shared_chat` module. This prevents copying the example into a new project.

## Why

The setup snippet was extracted from a showcase agent and carried its repository-local helper dependency. Installing `ag-ui-adk` does not provide that module or export `stop_on_terminal_text`.

Reproduction: executing the extracted setup with installed ADK dependencies in an empty project raises `ModuleNotFoundError: No module named 'agents'`; `hasattr(ag_ui_adk, 'stop_on_terminal_text')` is false.

## Fix

Make the two shared setup snippets standalone: use public `LlmAgent` and `AGUIToolset` imports and an explicit model name. Define and register the termination callback inline, preserving its final-text, streaming, pending-tool, and private-context guards without undefined helper imports. Showcase runtime agents are unchanged.

Validation: regenerated the production setup-content bundle through Nx, then extracted and executed both bundled Python examples in an isolated Python environment. Both construct an agent with `AGUIToolset` successfully. `git diff --check` passes. The two authored setup inputs and a Python regression test change; generated data is not committed. The copied examples and canonical callback pass 26 Python checks, covering callback registration, final text, partial/thinking output, pending tool calls, and missing private ADK context.

## Summary by CodeRabbit

- **Documentation**
  - Rewrote Google ADK integration setup instructions with clearer examples for enabling CopilotKit frontend tools and generative UI components.
  - Updated agent configuration guidance to use `AGUIToolset()` in the agent’s tools list.
  - Replaced referenced demo components with inline Python examples.

### DIFF
--- a/showcase/integrations/google-adk/docs/setup/agent-setup.mdx
+++ b/showcase/integrations/google-adk/docs/setup/agent-setup.mdx
@@ -1,10 +1,18 @@
-First, pass `AGUIToolset()` in your `LlmAgent`'s `tools=` list and pair it
-with `stop_on_terminal_text` as the `after_model_callback`. The toolset is
-what makes every CopilotKit feature on the frontend — frontend tools, shared
-state, agent context, and generative UI components — visible to your ADK
-agent on every turn.
+Pass `AGUIToolset()` in your `LlmAgent`'s `tools=` list to expose
+CopilotKit's frontend tools and generative UI components to the agent.
+Use an ADK-supported model available to your project.
 
-<DemoCode file="src/agents/hitl_in_chat_agent.py" region="setup" />
+```python
+from ag_ui_adk import AGUIToolset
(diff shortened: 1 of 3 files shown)


BenTaylorDev (approved):
Premise verified. The model matches the showcase's own `DEFAULT_MODEL`, and the pytest executes the copied code rather than a replica, which is the right shape of test for this.

Follow-up, not a blocker: `stop_on_terminal_text` now exists verbatim in these two files and a third in #7070, and the fixture params cover only these two. The third copy can drift silently.

PR #7064, opened 2026-09-10
fix(core): resolve relative runtime URLs in resource transport

tylerslaton:
## Problem

Intelligence chat failed before reaching the agent with a relative runtime URL. The single-route resource adapter runs on every runtime fetch, including chat kickoff, and threw `TypeError: Invalid URL` for `/api/copilotkit`.

## Why

The adapter passed the relative runtime URL to `new URL()` without a browser base, before deciding whether the request was a resource operation.

## Fix

Resolve the runtime against the browser location before comparing paths. Keep the original fetch URL and existing resource envelope behavior.

Reproduction: added a test through `CopilotKitCore.ɵruntimeFetch` with `/api/copilotkit`; before the fix it failed at `single-route-resource-request.ts:54` with `TypeError: Invalid URL` before fetch. After the fix, the chat request reaches fetch and the thread request becomes a resource envelope.

Validation: `nx run @copilotkit/core:test --excludeTaskDependencies -- src/__tests__/single-route-resource-transport.test.ts --maxWorkers=2` (8 passed); `nx run @copilotkit/core:check-types --excludeTaskDependencies` (passed, with current shared package built).

The broad pre-commit package gate failed across unrelated packages due to a React perf test timeout. It was excluded after the focused tests and core typecheck passed; lint and the remaining commit hooks passed.

Fixes FAC-209

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed relative runtime URLs so requests resolve correctly against the current browser URL.
  - Ensured path-relative runtime URLs work consistently for resource requests.
  - Added a localhost fallback for resolving relative runtime URLs in non-browser environments.
  - Improved reliability for Intelligence run calls and single-route resource requests using relative URLs.

### DIFF
--- a/packages/core/src/__tests__/single-route-resource-transport.test.ts
+++ b/packages/core/src/__tests__/single-route-resource-transport.test.ts
@@ -7,13 +7,19 @@ const RUNTIME_URL = "https://runtime.example.com/api/copilotkit";
 interface SetupOptions {
   capability?: boolean;
   transport?: "rest" | "single";
+  runtimeUrl?: string;
 }
 
 /** Connects a core to a mocked runtime and returns isolated cleanup. */
 async function setup(options: SetupOptions = {}) {
   const originalFetch = globalThis.fetch;
   const originalWindow = (globalThis as { window?: unknown }).window;
-  (globalThis as { window?: unknown }).window = {};
+  (globalThis as { window?: unknown }).window = {
+    location: {
+      href: "https://app.example.com/chat",
+      origin: "https://app.example.com",
+    },
+  };
 
   const fetchMock = vi.fn().mockImplementation(() =>
     Promise.resolve(
@@ -51,7 +57,7 @@ async function setup(options: SetupOptions = {}) {
   globalThis.fetch = fetchMock;
 
   const core = new CopilotKitCore({
-    runtimeUrl: RUNTIME_URL,
+    runtimeUrl: options.runtimeUrl ?? RUNTIME_URL,
     runtimeTransport: options.transport ?? "single",
   });
   await vi.waitFor(() => {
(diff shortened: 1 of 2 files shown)


BenTaylorDev (dismissed):
Verified against current `origin/main`.

`agent-registry.ts:622` runs this helper on every runtime fetch under `single` transport, and `single-route-resource-request.ts:54` threw before reaching fetch for a relative URL. This is the only such site, and absolute URLs are unaffected because `new URL()` ignores the base.

The three red checks are the AG-UI LFS fixture bug, already fixed in main by 8b94471a67.

Follow-up, not a blocker: the fallback is `undefined`, so a non-browser process with a relative URL still throws. The precedent at `agent.ts:249` falls back to `"http://localhost"`.


BenTaylorDev:
Follow-up on my approval, now that #7041 arrived on the same defect and is
closed as a duplicate.

#7041 used a `"http://localhost"` fallback, which is the pattern already in
`packages/core/src/agent.ts:250-252` and
`packages/core/src/intelligence-agent.ts:731`. Please take it here, keeping
`href` as the base, because `href` resolves a `runtimeUrl` without a leading
slash against the current path:

```ts
const base =
  typeof window !== "undefined" && window.location
    ? window.location.href
    : "http://localhost";
const runtime = new URL(runtimeUrl, base);
```

Two other things before merge:

- The branch is behind `main`. Please update it yourself. I hold the
  approval, so a push from me would take the approver slot, and the update
  also picks up the AG-UI LFS fixture fix (8b94471a67) for the three red
  checks.
- Please move the `FAC-209` link from #7041 to this PR.

The update will dismiss my approval, so ping me or another reviewer for a
re-approval after you push.

marthakelly (approved):
Reviewed URL resolution and resource classification across root-relative, path-relative, absolute, and non-browser configurations. All eight focused runtime tests pass; chat and cross-origin requests remain unchanged. CI is green and the branch includes current main.

PR #7069, opened 2026-09-10
fix(core): honor single-route transport for Intelligence chat

tylerslaton:
## Problem

With the LangGraph TypeScript quickstart’s single-route handler, Intelligence chat and reconnect posted to `/api/copilotkit/agent/graph/{run,connect}`, which that mount does not serve.

## Why

`ProxiedCopilotRuntimeAgent` retained the selected transport for SSE but dropped it when creating its Intelligence delegate. The runtime already supports Intelligence run/connect envelopes; the client did not use them.

## Fix

Pass the selected transport to the delegate. For single-route mounts, post `{ method, params, body }` to the base URL. Keep REST as the default and preserve run input, replay cursor, headers, and credentials.

Reproduction: both new proxy tests failed before the change: expected `https://app.example.com/api/copilotkit`, received the REST `/agent/graph/run` or `/connect` URL. Both pass after the fix, including envelope and caller credential assertions.

Validation: all 849 core tests pass across 73 files using `nx run @copilotkit/core:test --excludeTaskDependencies -- --maxWorkers=2`; `nx run @copilotkit/core:check-types --excludeTaskDependencies` passes. The existing health-check mock now distinguishes `info` and `agent/run` envelopes at their shared URL, preserving the assertion of one join request and two info requests.

Join requests now reject redirects to prevent a 307/308 from forwarding chat bodies to another endpoint. Eight real HTTP regression cases (REST/single-route, run/connect, 307/308) reproduced body forwarding before the fix and pass afterward. Configure the runtime URL to point directly to its handler; redirected endpoints now fail.

This fixes the reported docs mismatch at its SDK source, allowing the documented handler to work. The separate relative-URL crash is addressed in #7064; these PRs are independent.

The broad pre-commit package gate is unavailable in this partial dependency checkout; targeted tests, typecheck, lint, and commit hooks were run.

The PR also includes the one-line AG-UI LFS checkout fix from #7073 so the Dojo jobs receive the actual image fixture. That independent fix has passed all three LangGraph suites in hosted CI.

## Summary by CodeRabbit

* **New Features**
  * Intelligence agents now support single-endpoint transport for run and connect operations.
  * Single-endpoint requests use the configured runtime endpoint and include the appropriate operation details.
  * Runtime endpoint paths, including trailing slashes, are preserved when required.

* **Bug Fixes**
  * Redirected intelligence requests are no longer followed, preventing request data from being forwarded elsewhere.
  * Runtime connection health checks now correctly handle single-endpoint agent joins.


BenTaylorDev (dismissed):
Verified the load-bearing claim that the runtime already accepts these envelopes. `single-route-helpers.ts:1-11` lists both method names, `fetch-handler.ts:404-410` unwraps `methodCall.body` for `agent/run` and `agent/connect`, and `handle-run.ts:89-90` routes to `handleIntelligenceRun` regardless of transport, so the join JSON still comes back.

I tried three ways to break it and failed:
- The delegate cannot receive `"auto"`. `agent-registry.ts:1294` resolves the transport before constructing agents at `:1331`.
- `clone()` passes the whole config object, so the new `transport` field survives (`intelligence-agent.ts:205`).
- `singleEndpointUrl` is always assigned from `runtimeUrl` (`agent.ts:150` and `:615`).

The bundled workflow hunk is a no-op now that 8b94471a67 is in main.

Follow-up, not a blocker: the new test configures `transport: "single"` explicitly. A case that reaches `single` through auto-resolution would lock in the behavior I had to verify by hand.


BenTaylorDev (commented):
Re-reviewing the current head, since the force-push at 04:00 dismissed my approval from the 11th. The mechanism still holds, and drift is clean: 34 commits behind, but the diff against current `main` is only these five files and none of its source files changed on `main` since the base.

One blocking finding, plus two notes.

## The delegate receives the slash-stripped URL

CodeRabbit is right on `agent.ts:688`, and it is worth stating why this one matters rather than treating it as a nit.

`createIntelligenceDelegate` passes `runtimeUrl: this.runtimeUrl`, and the delegate now POSTs that value verbatim as its single-route endpoint. But this file documents the opposite at `agent.ts:111-116`:

> `runtimeUrl` is the slash-stripped form used for path joins; the single-route endpoint (run, connect, stop, info envelopes) is this verbatim value, because a trailing slash can select a different proxy location.

Every other single-route POST target obeys that: `agent.ts:161`, `agent.ts:461`, `agent.ts:616`, and `agent-registry.ts:635`, which cites #7028. After this change the Intelligence delegate is the only one that does not.

I tried to talk myself out of this on the grounds that the registry probably normalizes the URL upstream, which would make the trailing slash unreachable. The opposite is true. Both construction sites pass the verbatim form deliberately — `agent-registry.ts:505` and `agent-registry.ts:1344`, both `this._runtimeEndpointUrl ?? …` — so the slash survives into `config.runtimeUrl` by design.

The failure case: a developer sets `runtimeUrl` to `https://app.example.com/api/copilotkit/` on a single-route Intelligence mount. Info and stop envelopes go to the slashed path, run and connect go to the stripped one. On a mount that serves only the slashed location, chat fails while the Inspector still reports connected — which is the exact symptom this PR exists to remove.

It also compounds with `redirect: "error"`: if the host answers the stripped path with a 308 to add the slash, what would previously have been a followed redirect is now a hard failure.

## `redirect: "error"` applies to both transports

This is hardening I agree with in principle, since a 307 or 308 preserves the body and would forward the conversation to another origin. I am not asking you to drop it. But it lands on the REST path too, where it converts a redirecting mount from working into hard-failing, and that is wider than the single-route bug this PR is fixing. A Next.js app configured with `trailingSlash: true` is the case I would want confirmed — I did not check whether App Router route handlers take part in that redirect, so I am flagging it as plausible rather than proven. My preference is to split it into its own commit or ticket so it can be judged on its own merits.

## What the tests do and do not prove

`intelligence-single-route.test.ts` mocks `fetch`, so it asserts the URL and the envelope shape and proves nothing about the server accepting them. I verified that half by hand against current `main`, and it holds end to end: `fetch-handler.ts:404-410` unwraps `methodCall.body` for `agent/run` and `agent/connect`, `resolveSingleRoute` at `fetch-handler.ts:773-789` reads exactly `params.agentId`, and `handle-run.ts:87` routes on `isIntelligenceRuntime` with no transport condition. So the fix is correct — but a rename of that param would pass CI. Worth a cross-package test at some point, not in this PR.

The redirect test uses real loopback servers, so that half is genuinely proven.

Also still open from my earlier pass: the test pins `transport: "single"` by hand, so nothing covers reaching `single` through auto-resolution, which is what makes the `"auto"` → `"rest"` mapping on the new line safe.

I am pushing the URL fix and an auto-resolution test to this branch now, so it is not blocked on a round-trip. Leaving the redirect decision to you.

BenTaylorDev (approved):
Approving. The blocking finding from my last pass is fixed, CI is now fully green (69 passing, 0 failing), and I verified the change locally rather than reading it.

## Verified locally

Clean worktree at the PR head, frozen install, `@copilotkit/core` built from source:

- full `packages/core` suite: **74 files / 863 tests green**
- the three PR test files: **21 green**

Both load-bearing mechanisms are mutation-checked, so the tests are not vacuous:

- Reverting the delegate back to `runtimeUrl: this.runtimeUrl` fails the two single-route URL tests, on the trailing slash exactly.
- Removing `redirect: "error"` fails all 8 redirect tests, across both transports and both modes. Those drive real loopback servers, not a fetch mock.

## Four attempts to break it, all failed

- **A stop path that missed the fix.** There is none. `abortRun` pushes `STOP_RUN_EVENT` over the Phoenix channel, not HTTP. Run and connect are the only HTTP calls in this agent, and both are covered.
- **`runtimeEndpointUrl ?? runtimeUrl` silently degrading to the stripped URL.** It cannot be undefined at that point: `createIntelligenceDelegate` throws unless `runtimeUrl` is truthy, and both fields derive from the same `config.runtimeUrl`. The `??` is defensive, not a hole.
- **A stale agent id in the single-route envelope.** The envelope uses `this.config.agentId`, but so does `buildRuntimeUrl` on the REST path. The two paths are symmetric, and the staleness predates this PR.
- **My own `redirect: "error"` objection.** It was wider than the facts support, and I withdraw it as a blocker. The change lands at exactly one site, `intelligence-agent.ts:399`, so the blast radius is Intelligence runtimes only rather than the REST path generally. `redirect: "error"` is already on `main` in the Intelligence platform client at `client.ts:883`, so this follows existing practice, and the new suite pins it deliberately on both transports.

## One thing for the release notes

The behaviour change is real, and I measured it on a live 308: a POST that returns 200 today throws `TypeError: fetch failed / unexpected redirect` under `redirect: "error"`. The realistic trigger is narrower than I first thought, because both paths now address the mount the caller configured, so a trailing-slash canonicaliser no longer fires. What remains is a scheme or host redirect, an `http://` runtime URL behind an HSTS proxy being the likely one. Failing loudly there beats forwarding chat data and credentials to another origin, so I would ship it and note it rather than change it.

PR #7070, opened 2026-09-10
fix(docs): document the ADK agent context state key

tylerslaton:
## Problem

The documentation does not clearly establish the Google ADK agent-side path for `useAgentContext`. The hook reference points generically to integration guides, while the existing ADK setup uses `state["copilotkit"]["context"]`, which is not where the adapter stores the AG-UI request context.

## Why

Reproduced using the actual adapter and ADK Runner with an instruction provider and a deterministic local model. Given `RunAgentInput.context=[{description: "Incident records", value: "[{\"id\":\"INC-1041\"}]"}]` and empty request state:

```text
state["copilotkit"]["context"]: []
state[CONTEXT_STATE_KEY] (_ag_ui_context):
  [{"description": "Incident records", "value": "[{\"id\":\"INC-1041\"}]"}]
```

## Fix

Document the public `CONTEXT_STATE_KEY` export in the hook reference and link directly to the ADK context guide. Replace that guide's setup with a standalone instruction-provider example using the correct state key. Preserve the Gemini final-text termination safeguard with a fully defined callback in the same example. Explain the string values, parsing structured values, callback/tool access, and the distinction between exposing frontend tools and including context in model instructions.

Validation: regenerated the setup-content bundle through Nx, extracted and executed its Python example, and drove the documented instruction provider through the real adapter/Runner. It delivered `INC-1041` into the model request's system instruction and completed with text events and `RUN_FINISHED`. This is a deterministic integration reproduction, not a live Gemini/browser test. `git diff --check` passes.

Targeted shell-docs checks: **26 passed across 3 files** (`setup-content`, `setup-concept`, and `frontend-tools-setup-coverage`), run through Nx after generating the normal docs prerequisites.

The context example also passes nine copied-code callback checks (registration, final text, partial/thinking output, tool calls, and private-context compatibility). The bundled context-to-model execution check and 26 targeted docs tests pass after this correction.

### DIFF
--- a/showcase/integrations/google-adk/docs/setup/agent-context-setup.mdx
+++ b/showcase/integrations/google-adk/docs/setup/agent-context-setup.mdx
@@ -10,12 +10,43 @@
   <Step>
     ### Read CopilotKit context before each model call
 
-    `useAgentContext` entries arrive in ADK session state under
-    `state["copilotkit"]["context"]`. Add `AGUIToolset()` to the agent and
-    use a `before_model_callback` to inject those read-only values into the
-    system instruction.
+    The adapter copies the AG-UI request's `context` list into ADK session
+    state at `CONTEXT_STATE_KEY` (the public constant for `"_ag_ui_context"`).
+    Read it from an instruction provider's `ctx.state`, a callback's
+    `callback_context.state`, or a tool's `tool_context.state`.
 
-    <DemoCode file="src/agents/readonly_state_agent_context_agent.py" region="agent-context-setup" />
+    Each entry is a dictionary with `description` and `value` strings.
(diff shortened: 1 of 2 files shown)


BenTaylorDev (approved):
Verified that the setup snippet renders through the generated bundle.

Two follow-ups, neither a blocker:
- `integrations/adk/agent-app-context.mdx` already documents `CONTEXT_STATE_KEY` and `_ag_ui_context` at lines 25, 74, 79, 92, 154 and 159. Worth reconciling with that page so there is one description rather than three.
- I could not confirm the link `/google-adk/shared-state/agent-readonly`. The only shared-state route I found in the repo is `/google-adk/shared-state/workflow-execution`, and the feature slug is `readonly-state-agent-context`.

PR #7073, opened 2026-09-10
fix(ci): fetch AG-UI LFS fixtures for Dojo tests

tylerslaton:
## Problem

The three LangGraph Dojo multimodal suites fail before sending the image because `ag-ui/apps/dojo/e2e/fixtures/test-image.png` is a Git LFS pointer rather than an image.

## Why

The workflow enables LFS for its CopilotKit checkout but not its separate AG-UI checkout. The fixture is stored in AG-UI through LFS.

## Fix

Set `lfs: true` on the AG-UI checkout. No test or application behavior changes.

Evidence: all three failed CI jobs report that the image is a Git LFS pointer. Locally, the Git object is a 127-byte pointer; applying LFS smudge produces a 70-byte file with a valid PNG signature. `actionlint` passes with the existing Depot runner label configured and shellcheck disabled.

This addresses a separate CI blocker seen on #7069. Full Dojo verification requires the hosted workflow; local validation covers fixture materialization and workflow syntax.

### DIFF
--- a/.github/workflows/test_e2e-dojo.yml
+++ b/.github/workflows/test_e2e-dojo.yml
@@ -187,6 +187,7 @@ jobs:
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
         with:
           repository: ag-ui-protocol/ag-ui
+          lfs: true
           path: ag-ui
           ref: main
           persist-credentials: false


BenTaylorDev:
Closing as already landed. The identical change is on `main` as 8b94471a67 ("fix(ci): fetch AG-UI LFS fixtures for dojo tests", 2026-09-10 18:02), so this branch is now a no-op:

```
$ git diff origin/main refs/prs/7073 -- .github/workflows/test_e2e-dojo.yml
(no output)
```

The diagnosis was right, and the CI evidence backs it. The failing run on #7064 logs `lfs: false` on the ag-ui checkout, then `Event trace fixture .../test-image.png is a Git LFS pointer, not the real file` across all three LangGraph suites. Those suites should go green once the affected branches pick up `main`.

Reopen if I have misread which commit carries it.

PR #7079, opened 2026-09-11
fix: carry AG-UI 1.0 tool result content parts (PNI-427)

mme:
Stacked on #6944. Validates CopilotKit against [ag-ui-protocol/ag-ui#2727](https://github.com/ag-ui-protocol/ag-ui/pull/2727) (PNI-427), which lets a tool result carry content parts: `ToolMessage.content` and `TOOL_CALL_RESULT.content` become `string | ContentPart[]`, and the parts are renamed (`InputContent` → `ContentPart`, `TextInputContent` → `TextPart`, …) with the old names kept as aliases.

## What changes

**Pins.** Every `@ag-ui/*` preview moves from `390d432a` to `1084c9b8` (the head of ag-ui#2727), in the root overrides, every package and example manifest, and the lockfile. The lockfile was updated the way the previous bump was: URLs and SHA-512 checksums rewritten from the published tarballs, then `pnpm install --frozen-lockfile` — a fresh non-frozen install still trips `blockExoticSubdeps` on the URL pins, as #6944 already notes.

**Tool results reaching a `result: string` contract.** Everywhere CopilotKit hands a tool result to a renderer or a provider that holds a string, the content is narrowed with `contentToText()` from `@ag-ui/core` — the text parts concatenated, media left on the message, no placeholder invented:

- `react-core` `useRenderToolCall` (the `result` prop of tool-call renderers)
- `vue` `copilotChatToolCallItemRender`
- `angular` `render-tool-calls`
- `react-native` `CopilotChat`'s `toolResultContent`, which already coerced unknown content; a parts array is now legitimate 1.0 content and is rendered as text without the "non-string content" warning
- `demo-agents` OpenAI Chat Completions adapter (tool messages take text there)

**Tool results reaching the model.** The runtime's AI SDK path maps parts faithfully instead of flattening: text parts become AI SDK text, inline media becomes `media` with its bytes and media type, and a URL-referenced part is passed as the URL it carries. A string result is unchanged.

**Tests.** `useRenderToolCall` gains tests for a parts-shaped and a string result reaching the renderer; the runtime's `finalize-events` test narrows the result it parses.

## Not changed

`packages/core`'s `normalizeToolResultContent` already handled arrays of text parts; the runtime keeps stringifying handler results before minting tool messages. Nothing in CopilotKit produces media parts yet — this PR makes it consume them without loss.

## Checks run locally

`pnpm install --frozen-lockfile`, whole-workspace `check-types`, and the test suites of the touched packages (`react-core`, `react-native`, `runtime`, `angular`, `vue`, `core`, `demo-agents`), and the lefthook pre-commit run of tests, publint and attw over the 26 affected packages.


mme:
Review pass (Codex, read-only, scoped to this PR's commits) found four things; three are fixed in 41b4a61f and the pins move to the fixed ag-ui head in 7e86bdcc.

1. **runtime, Gemini tool-response count** — a text-only parts result is one text output again, and adjacent text parts merge inside a content list, so a provider sees one response per tool call. The Google adapter emits one `functionResponse` per text entry of a content list, and two text parts had produced two. Inline media still rides as `media`.
2. **runtime, TanStack converter** — a parts-shaped tool result becomes its text instead of `null`, which had dropped the result.
3. **react-native** — only real content parts (a text part with a string, a media part with a typed source) take the parts path; any other typed array keeps the serialise-and-warn fallback it had.
4. **Left as is, on purpose: URL-referenced media passed to the AI SDK as text.** Codex is right that the Anthropic and OpenAI adapters accept typed image URLs. The AI SDK `media` entry wants bytes, URL support differs per adapter, and deciding how much the runtime fetches or forwards on the model's behalf is a product call that does not belong in this PR. The test now names the behaviour as the current fallback; a typed-URL mapping is a follow-up.

Pins: `1084c9b8` → `92d00fd1` (ag-ui#2727 after its own review fixes). Whole-workspace `check-types` green; the touched suites and the lefthook test/publint/attw gates ran on each commit.

mme:
Second Codex pass. TanStack confirmed fixed; two partials and one new point, closed in a5000daa, with the pins moving to ag-ui `d8d4b3c1` after that:

1. **runtime, one response per tool call** — all of a result's text is now collected into a single AI SDK text entry placed ahead of the media, whatever the order of the parts, so the Google adapter emits one `functionResponse`. A media-only result is media alone: a placeholder text would be content the tool never returned, and whether an adapter can place media inside a tool response is the adapter's to decide (the installed Google adapter cannot; a `functionResponse.parts` mapping there is upstream work).
2. **runtime, URL boundaries** — URL references contribute the URL on a line of its own, so two references no longer run into one string.
3. **react-native** — a media part whose inline source lacks its mime type is not a part and keeps the serialise-and-warn path.

Whole-workspace `check-types` green; runtime and react-native suites re-run; lefthook test/publint/attw gates ran on each commit.

PR #7084, opened 2026-09-11
fix(channels)!: make the Microsoft Agents SDK and express optional peers of channels-teams

BenTaylorDev:
Part of OSS-1177. Companion to #7081, which is merged (`08952450db`).

## What

Three of `@copilotkit/channels-teams`'s runtime dependencies become **optional peer dependencies**: `@microsoft/agents-hosting`, `@microsoft/agents-activity`, and `express`. `@copilotkit/channels` gets the same declaration, since its `./teams` subpath re-exports the adapter.

`express` is additionally moved to a **lazy import inside `createTeamsServer().start()`**, so it is not needed to import the package at all — only to use the built-in listener.

Also folded in: a one-line correctness fix to `channels-slack`'s `lookup_slack_user` schema from #7081's review, which is on `main` today.

## Why this shape, and not the runtime-level cut

OSS-1173 originally proposed fixing this one layer up — make `@copilotkit/channels-intelligence` optional on `@copilotkit/runtime`, so the whole channels subtree leaves a plain install. **That exact change already shipped and was reverted as a field incident.** #5963 made it an optional peer, package managers omitted it, and consumers on runtime `1.63.0` hit `Module not found`. #6012 restored the hard dependency 29 hours later and added `scripts/release/verify-runtime-package.ts` to stop it recurring.

It is easy to misread `channel-manager.ts` as evidence that the package is already optional — it *is* reached through a dynamic `import()` with a friendly module-not-found error. But the comment on that specifier says why the import is dynamic, and the reason is module format, not optionality: `channels-intelligence` is pure ESM and the runtime has a CJS build. The friendly error is a safety net for a broken install, not a supported mode. It did not save the 1.63.0 users either.

So the durable fix cuts the **leaf** edges. `runtime` → `channels-intelligence` → `channels-{slack,teams}` is load-bearing, because `delivery-adapter.ts` statically imports both `/render` subpaths. But those subpaths are light. The weight is the platform SDKs hanging off the adapter packages, and those are reached only from the **self-hosted** surface a user explicitly opts into.

`express` is folded in with the Microsoft move deliberately: shipping them separately would make self-hosted Teams users add packages, upgrade, and add packages again.

## Why it is safe

Both SDKs are reached only from the self-hosted surface — `adapter.ts` (Microsoft) and `listener.ts` (Microsoft + `express`, which serves `POST /api/messages`). `types.ts` imports Microsoft `import type` only (erased), and `download-files.ts` merely names it in a comment.

Managed Channels reach this package **solely** through `@copilotkit/channels-teams/render` (`channels-intelligence/src/delivery-adapter.ts:52`), and that graph touches none of them.

`src/render/managed-entrypoint-isolation.test.ts` pins the invariant. It now reads the forbidden set from the package's own `peerDependenciesMeta` rather than hardcoding `@microsoft/*`, so moving a further dependency to an optional peer extends the guard automatically instead of silently leaving it behind.

## Testing

**1. The Microsoft stack leaves a plain runtime install, and with it zod 3.25.75.** Real consumer install of published `@copilotkit/runtime@1.71.0` + `react-core@1.71.0`, with only `channels-teams` overridden to the packed build from this branch:

```
--- is the Microsoft stack installed? ---
  @microsoft/* : NOT INSTALLED
--- zod copies ---   (12 before, 10 now; version 3.25.75 is GONE)
  3.25.76  node_modules/@a2ui/web_core/node_modules/zod
  ... (8 x 3.25.76)
  4.6.2    node_modules/zod
```

**2. The constraint is now satisfiable — this is the result that matters.** Same install, plus the top-level `zod: ^3.25.76` pin the onboarding run already tried and which previously bottomed out at 3 copies:

```
added 1222 packages in 37s
  3.25.76  node_modules/zod          <- exactly one copy
```

**3. The packed manifest carries neither SDK as a dependency:**

```
packed: @copilotkit/channels-teams 0.9.2
dependencies      : ['@ag-ui/client', '@ag-ui/core', '@copilotkit/channels-core',
                     '@copilotkit/channels-ui', '@copilotkit/core', '@copilotkit/shared', 'rxjs']
peerDependencies  : {"@microsoft/agents-activity": "^1.5.3",
                     "@microsoft/agents-hosting": "^1.5.3",
                     "express": "^4.21.2 || ^5.0.0"}
optional peers    : ['@microsoft/agents-activity', '@microsoft/agents-hosting', 'express']
express in deps?  : False
@microsoft in deps?: False
```

**4. The isolation guard is not self-fulfilling** — mutation check on the widened version:

```
### M1: add an express import into the render graph
  AssertionError: expected [ Array(1) ] to deeply equal []
  Tests  1 failed | 2 passed (3)

### M2: delete peerDependenciesMeta (vacuity guard — an empty forbidden set must fail)
  Tests  1 failed | 2 passed (3)

### restored
  Tests  3 passed (3)
```

**5. The same isolation holds in the shipped artifact, not just source.** The build is unbundled, so I walked the built `dist` closures independently of the test, with a walker that also follows dynamic `import()`:

```
packages/channels-teams/dist/render/index.js:  9 modules | forbidden hits: NONE
packages/channels-slack/dist/render.js:       15 modules | forbidden hits: NONE
```

**6. The `express` peer range is honest.** `listener.ts` uses only `express()`, `express.json()`, `app.post` and `app.listen` — identical across v4 and v5 — and this repo already resolves express `5.2.1` through a root override, so `^4.21.2 || ^5.0.0` is what we actually support. `express` resolves from all three changed packages:

```
examples/teams               5.2.1
packages/channels-teams      5.2.1
packages/channels            5.2.1
```

**7. The Slack message fix is mutation-checked and matches Zod.** `lookup_slack_user` reported `Expected object, received object` for `null` and arrays, because `typeof` reports `"object"` for both and Zod names them separately. Reverting the fix fails the new assertions:

```
### revert to bare typeof
  Received: "(root): Expected object, received object"   x2
  Tests  2 failed | 8 passed (10)
### restored
  Tests  10 passed (10)
```

Parity confirmed against zod 3.25.76 directly:

```
null       -> Expected object, received null
"atai"     -> Expected object, received string
["atai"]   -> Expected object, received array
7          -> Expected object, received number
```

**8. Lockfile carries only this change.** A plain `pnpm install` on this tree today produces ~1300 lines of unrelated churn (`@swc/helpers` 0.5.18 → 0.5.23 rewrites peer hashes across hundreds of entries), so the lockfile was edited to record only the three affected importers — 9 insertions, 3 deletions. `pnpm install --frozen-lockfile` accepts it, which proves the manifests and lockfile agree.

**9. Affected projects pass:**

```
NX   Successfully ran targets publint, attw, test, check-types for 5 projects
     and 21 tasks they depend on
```

(`channels-teams`, `channels`, `channels-slack`, `channels-intelligence`, `teams-example`.)

`pnpm run check:public-api-manifest` → manifest is current.

> The `express` commit was made with `--no-verify`. The pre-commit hook fails on `runtime-dotnet:test` and `runtime-ruby:test`, which need toolchains absent from this machine (`dotnet: command not found`; ruby lacks the `websocket` gem) and which this change does not touch. They are pulled in only because the change touches the shared public-API manifest. Everything this change does touch was run directly, as above.

## Breaking change

Self-hosted Teams users must add the Microsoft SDK:

```sh
pnpm add @microsoft/agents-hosting @microsoft/agents-activity
```

`express` is an optional peer too, but it is loaded lazily inside `createTeamsServer().start()`, so it is required **only if you use the built-in listener**. A bot that serves `POST /api/messages` from its own HTTP server never installs it. `examples/teams` is exactly that case and declares no express dependency.

Optional peers are not auto-installed, so this is silent at install time and loud at import:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@microsoft/agents-hosting'
```

Both package READMEs carry the install line and the reason.

`examples/teams` now declares the two Microsoft packages (`c9eae5f`). It previously resolved them only by falling through to the devDependencies of `packages/channels-teams` — a path that exists in this workspace and nowhere else. So the example kept working in-repo while a copy of it outside the monorepo would install clean and then fail at import. No CI job exercises that example as a published consumer, so this could not have been caught automatically.

Managed Channels users are unaffected — nothing about that path changes.

## Not in scope

`@slack/bolt`, `@slack/web-api` and `@slack/types` on `channels-slack` are the same shape, and evidence 5 above shows the Slack managed entrypoint is already SDK-free. They are the remaining work on OSS-1177, left out here so the Slack and Teams breaks land in separate, separately-announceable releases.

---

## Addendum: behaviour verified against a real consumer install

Packed the workspace closure and installed it as an external consumer with npm, to check what actually happens rather than what the declarations imply.

**Managed path is genuinely unaffected.** With *zero* peers installed, `import("@copilotkit/channels-teams/render")` resolves and its exports work. The `/render` **type** closure is clean too — 5 `.d.ts` files, no peer references — so managed consumers' `tsc` is unaffected.

**Three findings from that pass, all now addressed or documented:**

1. **Install is silent.** Both `npm install` and `pnpm install` complete with no warning about the missing optional peers. That is what `optional: true` means, but it confirms there is no pre-runtime safety net. Documented in both READMEs.

2. **TypeScript gives no signal under `skipLibCheck: true`** — which is what most projects use. A consumer importing the self-hosted root entry with no peers typechecks **clean, exit 0**. Under `skipLibCheck: false` it surfaces as `TS2307` naming the two Microsoft packages, and never names express (no `.d.ts` imports it).

3. **It used to be whack-a-mole, and no longer is.** Node reports one missing package at a time, and `adapter.js` is imported before `listener.js`. So a user installed the Microsoft SDK, re-ran, and only then hit `Cannot find package 'express'` — two fix-and-rerun cycles. The lazy import removes the second one entirely.

**The lazy import, verified end to end:**

```
express installed? NO
=== import root (express absent) ===
OK, createTeamsServer: function

=== start() without express ===
MESSAGE: createTeamsServer() requires 'express', an optional peer dependency of
         @copilotkit/channels-teams. Install it (`npm install express`), or serve
         POST /api/messages from your own HTTP server and do not call createTeamsServer().
CAUSE CODE: ERR_MODULE_NOT_FOUND
```

**And the listener still really serves, on both majors** — a live request, not a typecheck:

```
express 4 | GET /healthz -> 200 ok
express 5 | GET /healthz -> 200 ok
```

**The built root entry no longer statically imports express:**

```
STATIC externals : @ag-ui/client, @copilotkit/channels-ui,
                   @microsoft/agents-activity, @microsoft/agents-hosting,
                   node:crypto, rxjs
DYNAMIC externals: express
```

`listener-lazy-express.test.ts` fails if a static express import returns to any source file, and names the offending file. Mutation-checked: reinstating the static import in `listener.ts` fails two assertions, and smuggling one into `adapter.ts` fails naming `adapter.ts`.

Worth stating plainly: nothing in this repo calls `createTeamsServer`. `examples/teams` builds its server with `node:http`. So before this change, every self-hosted user installed express to satisfy a static import for a function they likely never call.

## Summary by CodeRabbit

- **Documentation**
  - Clarified Teams setup and installation requirements, including optional Microsoft Agents SDK and Express packages.
  - Documented managed rendering versus self-hosted server usage, custom HTTP server behavior, and common missing-package errors.
  - Expanded guidance for copying the Teams example outside the monorepo.

- **Bug Fixes**
  - Teams server startup now loads Express only when needed and provides clearer errors when it is unavailable.
  - Slack tool validation now reports the actual received value type, including numbers, arrays, and null.

- **Tests**
  - Added coverage for validation messages, optional package loading, and managed Teams entrypoint isolation.

### DIFF
--- a/packages/channels-teams/README.md
+++ b/packages/channels-teams/README.md
@@ -36,6 +36,24 @@ This self-hosted adapter remains fully supported. Choose it when you want the pr
 
 ```sh
 pnpm add @copilotkit/channels @copilotkit/channels-ui @copilotkit/channels-teams
+pnpm add @microsoft/agents-hosting @microsoft/agents-activity
+```
+
+The Microsoft 365 Agents SDK packages are **optional peer dependencies**, so
+they are not installed for you. Add them as shown above — this adapter is the
+self-hosted path, and it needs them to run the Teams ingress.
+
+They are optional because the managed path does not need them. Managed Channels
+reach this package only through the `./render` subpath, which never touches the
+Microsoft SDK, so an app that installs `@copilotkit/runtime` for a plain chat UI
+no longer pulls the whole Microsoft Agents stack, or the exact `zod` pin that
+comes with it.
+
+If you import the adapter without them, Node fails at import time and names the
+package to install:
+
+```
+Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@microsoft/agents-hosting'
 ```
 
 ## Quickstart
--- a/packages/channels-teams/package.json
+++ b/packages/channels-teams/package.json
@@ -54,15 +54,27 @@
     "@copilotkit/channels-ui": "workspace:^",
     "@copilotkit/core": "workspace:^",
     "@copilotkit/shared": "workspace:^",
-    "@microsoft/agents-activity": "^1.5.3",
-    "@microsoft/agents-hosting": "^1.5.3",
     "express": "^4.21.2",
     "rxjs": "^7.8.1",
     "zod": "^3.25.76",
     "zod-to-json-schema": "^3.25.1"
   },
+  "peerDependencies": {
+    "@microsoft/agents-activity": "^1.5.3",
+    "@microsoft/agents-hosting": "^1.5.3"
+  },
+  "peerDependenciesMeta": {
+    "@microsoft/agents-activity": {
+      "optional": true
+    },
+    "@microsoft/agents-hosting": {
+      "optional": true
+    }
+  },
   "devDependencies": {
     "@copilotkit/typescript-config": "workspace:^",
+    "@microsoft/agents-activity": "^1.5.3",
+    "@microsoft/agents-hosting": "^1.5.3",
     "@types/express": "^4.17.21",
     "@types/node": "^22.10.0",
     "typescript": "^5.6.3",
--- a/packages/channels-teams/src/render/managed-entrypoint-isolation.test.ts
+++ b/packages/channels-teams/src/render/managed-entrypoint-isolation.test.ts
@@ -1,0 +1,93 @@
+import { describe, it, expect } from "vitest";
+import { readFileSync, existsSync } from "node:fs";
+import { dirname, join, normalize } from "node:path";
+import { fileURLToPath } from "node:url";
+
+/**
+ * `@microsoft/agents-hosting` and `@microsoft/agents-activity` are OPTIONAL
+ * peer dependencies of this package, not dependencies (OSS-1173). They are
+ * reached only from the self-hosted surface — `adapter.ts` and `listener.ts`.
+ *
+ * That is only safe while the `./render` subpath stays free of them. It is the
+ * entry point `@copilotkit/channels-intelligence` imports for managed Teams
+ * delivery, and managed hosts do not install the Microsoft SDK. If a Microsoft
+ * import ever reaches this graph, every managed host breaks at import time with
+ * a module-not-found — so pin the invariant here rather than discover it in
+ * production.
+ *
+ * The walk is over source, not `dist`, so it fails in the PR that introduces
+ * the import rather than at publish time.
+ */
+
+const HERE = dirname(fileURLToPath(import.meta.url));
+
+/** Resolve a relative ESM specifier (`./x.js`) to its TypeScript source file. */
+function resolveLocal(specifier: string, importer: string): string | null {
+  if (!specifier.startsWith(".")) return null;
+  const base = normalize(join(dirname(importer), specifier)).replace(
+    /\.js$/,
+    "",
+  );
+  for (const candidate of [
+    `${base}.ts`,
+    `${base}.tsx`,
+    join(base, "index.ts"),
+  ]) {
+    if (existsSync(candidate)) return candidate;
+  }
+  return null;
+}
+
+/**
+ * Walk every module reachable from `entry`, returning the bare (non-relative)
+ * specifiers the graph imports. Matches `import ... from "x"`, `export ... from
+ * "x"`, and `import("x")`, which is every form this package uses.
+ */
+function externalSpecifiersFrom(entry: string): Map<string, string[]> {
+  const seen = new Set<string>();
+  const external = new Map<string, string[]>();
+  const queue = [entry];
+
+  while (queue.length > 0) {
+    const file = queue.pop()!;
+    if (seen.has(file)) continue;
+    seen.add(file);
+
+    const source = readFileSync(file, "utf8");
+    for (const match of source.matchAll(
+      /(?:from|import)\s*\(?\s*["']([^"']+)["']/g,
+    )) {
+      const specifier = match[1]!;
+      const local = resolveLocal(specifier, file);
+      if (local) {
+        queue.push(local);
+      } else if (!specifier.startsWith(".")) {
+        external.set(specifier, [...(external.get(specifier) ?? []), file]);
+      }
+    }
+  }
+  return external;
+}
+
+describe("./render is importable without the Microsoft Agents SDK", () => {
+  const external = externalSpecifiersFrom(join(HERE, "index.ts"));
+
+  it("reaches no @microsoft/* package", () => {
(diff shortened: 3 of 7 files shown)


BenTaylorDev:
Holding until morning, per Ben. Ready on the merits — 60 checks green, mergeable, branch updated against main.

Two decisions attached to it:

1. **Version: `0.10.0`, not a patch.** This is `fix(channels)!` — making `@microsoft/agents-*` and `express` optional peers means anyone using the Teams adapter installs them, so a patch would understate it. On a `0.x` line a breaking change takes the minor.
2. **It also lands the umbrella.** Channels `0.9.3` published 8 of 9 packages; `@copilotkit/channels` timed out waiting for registry propagation (`not visible on npm yet`, 31 retries) and is still at `0.9.2`. The next release picks it up.

Context for why this matters now: `0.9.3` removed CopilotKit's own `zod ^3.25.76` range, which was only half the fix. A clean `@copilotkit/runtime` install still resolves two zods — `3.25.75` pinned exactly by `@microsoft/agents-*`, against the `^3.25.76` floor from `@ai-sdk/*`, `ai` and `langchain`. This PR removes the Microsoft stack from a plain install, which is what collapses it to one.

PR #7094, opened 2026-09-11
fix(web-inspector): unify Intelligence onboarding states

tylerslaton:
## Problem

The Inspector locked Threads state varied by license metadata even though the real availability gate is whether Runtime exposes the Threads list endpoint. That let users with Threads unavailable land in different legacy onboarding states.

## Why

License state and Threads capability are separate signals. Keying the pane on license metadata made Threads unavailable diverge from the license-disabled setup view, even though neither state can list saved Threads.

## Fix

- Gate the locked setup view on Runtime Threads capability.
- Use the intended Rich Threads product header, Copy setup prompt, engineer CTA, and video for every locked state.
- Keep the existing enabled-but-empty Threads preview when the list endpoint is available.
- Update Inspector docs to describe the capability-based state instead of removed license-specific variants.
- Simplify the static locked copy and regression coverage.
- Squash the PR into one semantic commit.

The earlier copied-prompt/intent commit was intentionally dropped during rebase because #7030 landed that work on main.

Validation:

- Inspector: 696 tests passed, 0 skipped.
- Inspector typecheck passed.
- Shell-docs typecheck passed.
- Shell-docs production build passed, including 227 static pages.
- Standalone `pro-disabled-zero` visual check passed.

## Summary by CodeRabbit

* **User Experience**
  * Unified the locked Threads view around Rich Threads setup guidance.
  * Added **Copy setup prompt** and **Talk to an Engineer** actions.
  * Threads availability is now determined by Runtime capability rather than license metadata.
  * Empty Threads views no longer display local example threads or example tours.

* **Documentation**
  * Updated setup and troubleshooting guidance to describe the capability-based Threads experience.
  * Clarified that Inspector uses action links supplied by Runtime and directs incomplete configurations to the Runtime route guide.

### DIFF
--- a/packages/web-inspector/dev/threads-state-lab.spec.ts
+++ b/packages/web-inspector/dev/threads-state-lab.spec.ts
@@ -395,26 +395,12 @@ function expectedOverviewCopy(
   scenario: ThreadsStateScenario,
 ): Readonly<{ heading: string; description: string }> | null {
   if (scenario.data === "error") return null;
-  if (scenario.runtimeInfo.licenseStatus === "none") {
+  if (scenario.capability !== "enabled") {
     return {
       heading:
         "Production-grade chat threads without the complexity. Self hostable.",
       description:
         "Chat threads that go beyond text with generative UI and multimodal inputs, built to replay missed events and stay in sync across tabs, sessions, and devices.",
-    };
-  }
-  if (scenario.runtimeInfo.licenseStatus === "expired") {
-    return {
-      heading: "Renew Intelligence to inspect Threads.",
-      description:
-        "Your Intelligence access has expired. Renew it to inspect saved thread history.",
-    };
-  }
-  if (scenario.capability !== "enabled") {
-    return {
-      heading: "Finish setting up Rich Threads",
-      description:
-        "Copy this prompt into your coding agent to finish the setup.",
     };
   }
   if (scenario.data === "existing") return null;
--- a/packages/web-inspector/src/__tests__/inspector-metadata.spec.ts
+++ b/packages/web-inspector/src/__tests__/inspector-metadata.spec.ts
@@ -294,22 +294,17 @@ test.each([
   },
 );
 
+const LOCKED_THREADS_HEADING =
+  "Production-grade chat threads without the complexity. Self hostable.";
+
 test.each([
-  ["valid", "manage_plan", "Finish setting up Rich Threads"],
-  [
-    "none",
-    "enable_intelligence",
-    "Production-grade chat threads without the complexity. Self hostable.",
-  ],
-  ["expired", "renew", "Renew Intelligence to inspect Threads."],
-  [
-    "unknown",
-    "manage_plan",
-    "Production-grade chat threads without the complexity. Self hostable.",
-  ],
+  ["valid", "manage_plan"],
+  ["none", "enable_intelligence"],
+  ["expired", "renew"],
+  ["unknown", "manage_plan"],
 ] as const)(
-  "locked Threads use %s license copy and the unified actions",
-  async (licenseState, actionKind, heading) => {
+  "locked Threads ignore %s license metadata and use the unified actions",
+  async (licenseState, actionKind) => {
     const context = await setup({
       metadata: fullMetadata(licenseState, actionKind),
       runtimeLicense: licenseState,
@@ -326,7 +321,7 @@ test.each([
       const talk = root.querySelector<HTMLAnchorElement>(
         '[data-inspector-locked-feature-talk="threads"]',
       );
-      expect(root.textContent).toContain(heading);
+      expect(root.textContent).toContain(LOCKED_THREADS_HEADING);
       expect(action).toBeNull();
       expect(talk?.textContent?.trim()).toBe("Talk to an Engineer");
       expect(
@@ -442,7 +437,7 @@ test("known license disagreement uses Runtime copy and hides the metadata action
 
     const root = context.inspector.shadowRoot!;
     expect(root.textContent).toContain(
-      "Renew Intelligence to inspect Threads.",
+      "Production-grade chat threads without the complexity. Self hostable.",
     );
     expect(
       root.querySelector('[data-inspector-action-placement="locked"]'),
--- a/packages/web-inspector/src/__tests__/threads-states.spec.ts
+++ b/packages/web-inspector/src/__tests__/threads-states.spec.ts
@@ -1094,8 +1094,10 @@ const lockedActionCases: ReadonlyArray<LockedActionCase> = [
     runtimeLicense: "valid",
     actionKind: "manage_plan",
     actionUrl: "https://cloud.copilotkit.ai/actions/manage",
-    heading: "Finish setting up Rich Threads",
-    description: "Copy this prompt into your coding agent to finish the setup.",
+    heading:
+      "Production-grade chat threads without the complexity. Self hostable.",
+    description:
+      "Chat threads that go beyond text with generative UI and multimodal inputs, built to replay missed events and stay in sync across tabs, sessions, and devices.",
   },
   {
     name: "none enable action",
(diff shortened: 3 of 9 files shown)


BenTaylorDev (changes requested):
Reviewed against current `main`. The branch is current with main, and every check is green, including all six unit shards and `dev/threads-state-lab.spec.ts` at 25 tests.

The code change itself is sound, and I verified the premise rather than taking it from the description. `areThreadEndpointsAvailable()` (`packages/web-inspector/src/index.ts:7017`) reads `threadEndpoints.list` only and takes no license input, so keying the locked splash on capability instead of on license metadata matches the real gate. The state-lab predicate swap is exact rather than a loosening: `license-none` and `license-expired` are both `capability: "disabled"`, and the only scenario with an absent `licenseStatus` (`oss-no-metadata-enabled-zero`) is `capability: "enabled"`, so the old and new predicates agree on every route. An expired plan keeps its renew path in `renderSidebarIntelligenceStatus` (`index.ts:12579`), and the splash never carried a metadata action anyway, because nothing emits `placement="locked"` on `main` either.

Two things block a merge.

## 1. The description no longer matches the diff

The force-pushes dropped `649bc0ddaf` ("simplify Intelligence onboarding", 13 files, -337 lines). The body still claims example-thread removal, the shortened CLI prompt, and `add-learning`. None of those are in this diff: the example-thread machinery is untouched, `onboarding-prompt.ts` is untouched, and `add-learning` already landed on `main`. The test evidence does not match either. The body reports 689 passed with 6 skipped, and this head runs 696 passed with 0 skipped. CodeRabbit's two actionable comments point at `thread-capability.spec.ts` and `thread-selection.spec.ts`, which this diff does not touch.

Please confirm whether dropping that commit was intentional. If it was, rewrite the body down to the one thing this PR does, so the next reader is not reviewing against a description of other work.

## 2. Three docs pages quote the copy this PR deletes

- `showcase/shell-docs/src/content/docs/backend/runtime-endpoints.mdx:223` uses "If the Inspector says **Finish setting up Rich Threads**" as the entry condition for the whole route-setup section.
- `showcase/shell-docs/src/content/docs/inspector.mdx:186` and `showcase/shell-docs/src/content/snippets/shared/intelligence/inspector.mdx:115` carry a four-row table keyed by license state whose Active row quotes the same string.

This is not an old stale reference. The copy and its docs landed together in b3e1586330 (2026-08-04), and the table was rewritten on 2026-09-08 in b00d1b0eeb, keeping the license-state model on purpose. After this change that table collapses to one row. The docs need to move with the code, and the docs owner should know.

## Nits, not blocking

- 611650cb6d adds "Enable Threads to inspect saved history." and d2b8baf275 deletes it. We merge with a merge commit, so both land on main. Worth squashing into one.
- `getThreadsLockedCopy()` now takes no input and reads no state. A module constant, or an inline literal at the single call site (`index.ts:18809`), says that plainly.
- The four rows of the `test.each` at `inspector-metadata.spec.ts:297` now assert an identical heading and an identical null action, so the license column no longer discriminates anything. Hoisting the string and renaming the test would stop it reading as coverage it no longer provides.

## The one thing I would most want confirmed

In the valid-license, routes-not-mounted state the developer is one command away, and the headline that named that command now reads as a product pitch. The "Copy setup prompt" button survives, so the action stays reachable. Is that copy change the intended product call, or a side effect of unifying on the not-enabled design?

tylerslaton:
Addressed the review in `14a983104f`:

- Confirmed the prompt/intent commit was intentionally dropped because #7030 now provides that work on `main`.
- Confirmed the Rich Threads product heading is intentional for every locked state; locked means Runtime does not expose the Threads capability, regardless of license metadata.
- Updated all three stale docs references and removed the obsolete license-specific state table.
- Replaced the state-free copy method with a module constant.
- Hoisted the shared test heading and renamed the parameterized test to describe license metadata being ignored.
- Squashed the PR to one semantic commit.
- Rewrote the PR body to match the actual post-rebase diff and current validation.

Local validation: Inspector 696/696 tests and typecheck passed; shell-docs typecheck and production build passed (227 static pages). The repository-wide pre-commit packaging sweep separately hit three `ENOENT` tarball cleanup races that Nx marked flaky; the pushed CI run exercises the full matrix.

BenTaylorDev (commented):
Re-reviewed at `14a983104f`. One commit, current with main, all 52 checks green, including `build-check (shell-docs)` and `doc-tests`.

Everything I raised is addressed, and I verified each item rather than taking the summary:

- `THREADS_LOCKED_COPY` at `index.ts:1027` with one call site.
- Test string hoisted, and "locked Threads ignore %s license metadata" is now an accurate test name.
- The #7030 claim checks out. #7030 merged at 20:39 today and its merge commit `f24760fdbd` is this branch's own base. Diffed against its first parent, it removed the coding-agent identifier and shortened both prompt templates. What the dropped commit held beyond that is a small copy refinement, not the work being claimed.
- No reference to either deleted heading survives anywhere in the tree.

Two things left, and then this is ready.

## 1. A fourth docs page still describes the old mechanism

`showcase/shell-docs/src/content/docs/troubleshooting/event-inspector.mdx:14-19` says the Inspector "handles a locked Threads tab from the Runtime's current license state", lists **Enable Intelligence**, **Renew**, and **Manage plan**, and links to "[Inspector](/inspector) for the state matrix". That is the claim this PR disproves, and the state matrix it points at is the table this PR deletes, so the cross-reference now dangles.

My first review named three pages and this is a fourth. I had grepped for the deleted headings, and this page describes the mechanism without quoting them. I re-ran the search structurally and this is the only page left.

## 2. Decide where the trusted-URL guarantee lives

Both inspector pages also lost the paragraph stating that the Inspector opens only Runtime-supplied action URLs, never a constructed project URL or a fixed signup fallback, and that a metadata-versus-runtime conflict lets the Runtime win and drops the incompatible action.

That behavior is live, not stale. `packages/web-inspector/src/lib/inspector-metadata.ts:120-138` computes `hasLicenseConflict` and skips `projectAction` entirely on conflict. The only surviving trace of the guarantee is the sentence on `event-inspector.mdx`, which is the page that still needs correcting, so fixing that page could remove the last statement of it. Please keep it somewhere deliberately instead of letting it fall out of the docs.

tylerslaton:
Addressed the follow-up review in `0db8de0784`:

- Updated `troubleshooting/event-inspector.mdx` to describe the actual capability gate and unified locked Threads actions.
- Removed the dangling state-matrix reference.
- Deliberately restored the trusted-action guarantee under **Understand project and usage information** in both the canonical Inspector page and its shared Intelligence snippet: Inspector only opens Runtime-supplied URLs, never constructs a project/signup URL, and hides incompatible actions when Runtime and metadata license states conflict.
- Kept the PR squashed to one semantic commit.

Validation: `git diff --check` passed and the shell-docs production build completed successfully, including all 227 static pages.

BenTaylorDev (approved):
Approving at `0db8de0784`. One commit, 52 of 52 checks green, including `build-check (shell-docs)` and `doc-tests`.

Verified this round:

- `troubleshooting/event-inspector.mdx` now describes the capability gate and the two real CTAs, and its link resolves: `#enable-or-repair-intelligence` is a real heading at `inspector.mdx:173`.
- The trusted-action guarantee is back in both the canonical page and the shared snippet, and the wording matches the code. `packages/web-inspector/src/lib/inspector-metadata.ts:120-138` computes `hasLicenseConflict`, lets the Runtime state win, and skips `projectAction` entirely, so both halves of that sentence are true.
- The source diff is unchanged from the version I already checked.

I also tried to find a reason not to approve, and did not find one. No locked state loses its only action, because the sidebar still renders "Intelligence plan expired" and "Renew to restore access". A structural sweep of shell-docs returns no remaining claim that the locked Threads tab follows license state. The deletion of the state table created no dangling anchor.

Two follow-ups, neither blocking this PR:

1. The branch is behind main and the ruleset is strict, so it needs an update from you before it can merge. Please update it yourself. If I update it, `require_last_push_approval` locks me out of re-approving. Ping me after the update and I will re-approve whatever the push dismisses.
2. Pre-existing and not caused by this change: `docs/intelligence/quickstart.mdx:192`, `:203`, and `:214` all link to `/inspector#project-context-and-usage`, which matches no heading on main either. The link text is "Inspector setup states", which is the subject you just rewrote, so it is worth fixing here or filing separately.

PR #7102, opened 2026-09-12
fix(examples): keep Next.js overrides within supported majors

Swpn0neel:
## What does this PR do?

The bare `next: ^16.0.10` override forces Next 14 and 15 examples onto 16.1.3, while examples that declare Next 16 resolve to 16.2.3. This removes the bare override and replaces the overlapping ranges with two major-scoped security floors:

- Next 15: `15.5.24`
- Next 16: `16.3.3`

These are the patched versions for [GHSA-2xp9-vwfh-vxw4](https://github.com/vercel/next.js/security/advisories/GHSA-2xp9-vwfh-vxw4) and [GHSA-p293-qw3h-jr36](https://github.com/vercel/next.js/security/advisories/GHSA-p293-qw3h-jr36). The rationale is recorded alongside the overrides.

Neither advisory has a Next 14 fix. State Machine, Travel, and the v1 Pages Router example therefore explicitly declare `15.5.24`. The other nine manifests retain their existing Next ranges. This avoids keeping those three apps on an unsupported major or silently moving them across majors through another override.

The updated lockfile covers all 12 Next importers and their compiler, image-processing, and peer dependencies. Unrelated package versions are preserved.

The v1 Pages Router example also needs two compatibility adjustments when it actually runs on Next 15: bundle the SDK's global KaTeX styles and use the [documented ESM interoperability setting](https://nextjs.org/docs/messages/import-esm-externals) for `react-syntax-highlighter`/`refractor`. Its API route now lets `OpenAIAdapter` create the default client, avoiding a duplicate-OpenAI-type error that also reproduces with the previous Next 16.1.3 resolution.

## Validation

Node 22.23.2 and pnpm 10.33.4:

- `pnpm install --frozen-lockfile` passes for all 70 workspace projects.
- Resolution audit: nine importers resolve to Next 15.5.24 and three to Next 16.3.3; every importer matches its declared major.
- Nx production builds pass for State Machine, Travel, the v1 Pages Router example, `examples/v2/next-pages-router` (which uses the App Router), and Chat With Your Data (Next 16.3.3).
- Changed source/configuration files pass formatting and lint checks.
- A clean merge with current main (`06b8901d4`) also passes dependency-reference checks and `pnpm install --frozen-lockfile --lockfile-only --ignore-scripts` across its 75 workspace projects.

Builds use a placeholder OpenAI key. Next reports the existing dynamic-import warning from the runtime's channel manager; the v1 Pages Router build also reports its warning about opting into `esmExternals: "loose"`.

## Related PRs and Issues

Addresses the Next.js half of #6423. The separate Pino fix is #7089. This PR does not change Pino or MCP dependencies; the MCP follow-up was handled separately in #7095.

## Checklist

- [x] I have read the contribution guide.
- [x] The regenerated lockfile is included and the frozen install passes.
- [x] Allow edits by maintainers.

## Summary by CodeRabbit

* **Improvements**
  * Updated the Pages Router, state machine, and travel examples for compatibility with newer Next.js releases.
  * Improved the Pages Router example’s handling of module compatibility and syntax highlighting dependencies.
  * Simplified the example API configuration by using the adapter’s standard setup.
  * Updated supported Next.js version ranges and security-floor settings across the project.

### DIFF
--- a/examples/v1/next-pages-router/next.config.mjs
+++ b/examples/v1/next-pages-router/next.config.mjs
@@ -1,6 +1,12 @@
 /** @type {import('next').NextConfig} */
 const nextConfig = {
   reactStrictMode: true,
+  // Bundle the SDK's global KaTeX CSS for the Pages Router.
+  transpilePackages: ["@copilotkit/react-core"],
+  experimental: {
+    // react-syntax-highlighter's CommonJS entry loads ESM refractor languages.
+    esmExternals: "loose",
+  },
 };
 
 export default nextConfig;
--- a/examples/v1/next-pages-router/package.json
+++ b/examples/v1/next-pages-router/package.json
@@ -18,7 +18,7 @@
     "@google/generative-ai": "^0.11.2",
     "@heroicons/react": "^2.0.18",
     "clsx": "^1.2.1",
-    "next": "14.2.35",
+    "next": "15.5.24",
     "openai": "^4.85.1",
     "react": "^18",
     "react-dom": "^18"
--- a/examples/v1/next-pages-router/pages/api/copilotkit.ts
+++ b/examples/v1/next-pages-router/pages/api/copilotkit.ts
@@ -3,11 +3,9 @@ import {
   OpenAIAdapter,
   copilotRuntimeNextJSPagesRouterEndpoint,
 } from "@copilotkit/runtime";
-import { NextApiRequest, NextApiResponse } from "next";
-import OpenAI from "openai";
+import type { NextApiRequest, NextApiResponse } from "next";
 
-const openai = new OpenAI();
-const serviceAdapter = new OpenAIAdapter({ openai });
+const serviceAdapter = new OpenAIAdapter();
 
 const runtime = new CopilotRuntime({
   actions: [
--- a/examples/v1/state-machine/package.json
+++ b/examples/v1/state-machine/package.json
@@ -15,7 +15,7 @@
     "@copilotkit/runtime-client-gql": "workspace:*",
     "clsx": "^1.2.1",
     "motion": "^11.18.1",
-    "next": "14.2.35",
+    "next": "15.5.24",
     "react": "^18.3.1",
     "react-dom": "^18.3.1",
     "reactflow": "^11.11.4",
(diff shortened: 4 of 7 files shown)


BenTaylorDev (approved):
Verified this against current `main` rather than the PR's merge-base, and ran the affected example live. Everything checks out.

**The security claim is exact.** Both cited advisories are real and critical, and their patched versions are `15.5.24` and `16.3.3` — precisely the floors set here. Neither has a Next 14 fix (`>=10.0.0 <15.5.24` and `>=13.4.0 <15.5.24`), so moving State Machine, Travel and the Pages Router example to 15 is the only way to get them onto a patched release. My earlier guidance on #6423 said to keep 14, 15 and 16 each on their declared major; that advice predates these advisories and is the thing that went stale, not this PR.

**Resolution verified.** In a real merge with current `main`, all 12 workspace importers resolve to a patched Next: 9 on 15.5.24, 3 on 16.3.3. Matches the description exactly.

**No collision with the pino change.** This branch predates #7089, so I checked whether merging would resurrect the `pino@<=10.1.1` override. It does not. The merge is clean, the pino removal survives, and `pnpm install --frozen-lockfile` passes across all 75 projects on the merge result.

**The example changes are load-bearing repairs, not scope creep.** I built and ran `examples/v1/next-pages-router` from `main` as a baseline, and it is thoroughly broken there: the production build fails on a duplicate `openai@4.104.0` type conflict (two copies differing only by peer `ws`), and in dev `GET /` returns 500 with "more than one copy of React". With this PR the build passes, `GET /` returns 200, the chat UI mounts, and the client's own POSTs to `/api/copilotkit` return 200. `transpilePackages` is what fixes the duplicate-React failure, and the adapter change fixes the type error — both exactly as described.

**`esmExternals: "loose"` verified at runtime, not just at build.** I rendered a fenced Python block through the real markdown path: 17 Prism token spans, 16 colour-styled. The refractor interop works.

Two notes, neither blocking and neither yours:

- `GET /api/copilotkit/info` 404s. `pages/api/copilotkit.ts` is a single Pages Router file and cannot serve a subpath; it needs a catch-all. Pre-existing, non-fatal, and untouched here.
- Five example apps still declare Next 14 and sit outside the pnpm workspace with their own lockfiles, so no root override has ever reached them: `_legacy/copilot-fully-custom`, `_legacy/copilot-anthropic-pinecone`, `showcases/todo`, `showcases/enterprise-brex`, `showcases/presentation`. Your reasoning applies to them too. I'll file that separately.

Thanks for splitting this out and for the thorough validation notes; the resolution audit made this much faster to check.

PR #7106, opened 2026-09-12
LangGraph Middleware: use interrupt() to wait frontend tool results

### LINKED ISSUE #4920: 🚀 Feature Request: (LangGraph) Use `interrupt()` for FrontEnd tools
### Problem or Motivation

When you have a LangGraph graph (`create_agent`), FrontEnd tool calls are temporarily removed from LLM model responses, and re-injected with responses once the graph run finished (see https://github.com/ag-ui-protocol/ag-ui/discussions/1505).

I understand this is avoids having a node that is triggered for a tool call.

But this creates issues (like https://github.com/CopilotKit/CopilotKit/issues/4759) and agents are not production ready as the history the agent sees in between tool calls and the same turn is hacked, ruining the natural tool calling sequence:

Agent calls a tool -> tool executed -> Tool message added -> Agent continues with **full awareness**.

Also this setup is not friendly for pop-up widgets like "ask the user multiple question" widget.
 

### 🎤 Tell us your idea

Standard Human-in-the-loop: LangGraph graphs interrupts for tool calls -> Front End runs tools -> Front End resumes graph from the interruption

1. FrontEnd tools are executed in a node where they are batched and go through a single `interrupt()`, before executing any backend tool. Once can modify `CopilotKitMiddleware` to catch FE tools and perform this, see below.

<details><summary>CopilotKitMiddleware example </summary>

```python
"""Batched-interrupt routing for client-side (frontend) tools.

Subclasses ``CopilotKitMiddleware`` and overrides ``after_model`` so that
when the model emits ``tool_calls`` whose names match frontend descriptors
forwarded by the client (``state["copilotkit"]["actions"]``), the graph
pauses on a single batched ``interrupt(...)`` carrying every frontend
call. ``ag_ui_langgraph`` surfaces that pause to the client as
``CustomEvent(name="on_interrupt", value={"frontend_tool_calls": [...]})``;
the client dispatches each call locally and POSTs
``forwardedProps.command.resume = {"tool_results": [{"toolCallId", "content"}, ...]}``
to re-enter the graph. On resume, this middleware appends one
``ToolMessage(tool_call_id=…, content=…)`` per resumed result and routes
back to the model (or, if backend tool_calls share the turn, lets the
standard conditional edge route through ``ToolNode`` first).

The frontend ``tool_calls`` remain on the AIMessage throughout — that
keeps each new ``ToolMessage`` paired with a matching ``tool_call_id``
so the parent class's ``_fix_messages_for_bedrock`` doesn't erase them
as orphans, and the persisted history reflects what the model actually
emitted. The built-in ``ToolNode`` filters tool_calls by its own
registry before running, so the frontend calls are skipped there.

**Standalone usability.** This middleware works with or without
backend tools. ``@hook_config(can_jump_to=["model"])`` on
``after_model`` (combined with ``{"jump_to": "model"}`` in the
returned state for the all-frontend case) installs the conditional
edge that re-enters the model after resume even when ``create_agent``
was called with ``tools=[]``. Mixed FE+BE turns inherently need a
``tool_node``, since the backend tools have to execute somewhere.
"""

import json
import logging
from typing import Any

from copilotkit.copilotkit_lg_middleware import StateSchema
from langchain.agents.middleware import hook_config
from langchain_core.messages import AIMessage, ToolMessage
from langgraph.runtime import Runtime
from langgraph.types import interrupt

class CopilotKitMiddlewareWithFrontendInterrupt(CopilotKitMiddleware):
    """``after_model`` hook that batch-interrupts on frontend tool calls."""

    @property
    def name(self) -> str:
        return "CopilotKitMiddlewareWithFrontendInterrupt"

    @hook_config(can_jump_to=["model"])
    def after_model(
        self,
        state: StateSchema,
        runtime: Runtime[Any],  # noqa: ARG002
    ) -> dict[str, Any] | None:
        frontend_tools = state.get("copilotkit", {}).get("actions", [])
        if not frontend_tools:

            return None

        frontend_tool_names = {t.get("function", {}).get("name") or t.get("name") for t in frontend_tools}

        messages = state.get("messages", [])
        if not messages:

            return None

        last_message = messages[-1]
        if not isinstance(last_message, AIMessage):

            return None

        tool_calls = getattr(last_message, "tool_calls", None) or []
        if not tool_calls:

            return None

        frontend_calls = [c for c in tool_calls if c.get("name") in frontend_tool_names]
        backend_calls = [c for c in tool_calls if c.get("name") not in frontend_tool_names]

        if not frontend_calls:
            return None

        # Pause the graph with every frontend call batched into one
        # interrupt. On resume, ``payload`` carries the per-call results
        # the client produced.
        interrupt_payload = {
            "frontend_tool_calls": [
                {"id": c.get("id"), "name": c.get("name"), "args": c.get("args") or {}} for c in frontend_calls
            ]
        }
        payload = interrupt(interrupt_payload)

        results = _parse_tool_results(payload)
        results_by_id = {r["toolCallId"]: r["content"] for r in results}

        # Build a ToolMessage for every frontend call. If the client
        # dropped a result (cancel, crash, etc.), synthesise a placeholder
        # so the AIMessage's tool_calls stay paired with ToolMessages and
        # Bedrock doesn't reject the conversation.
        new_tool_messages: list[ToolMessage] = []
        for call in frontend_calls:
            call_id = call.get("id")
            content = results_by_id.get(call_id)
            if content is None:
                content = json.dumps({"ok": False, "error": "missing_tool_result"})
            new_tool_messages.append(ToolMessage(tool_call_id=call_id, content=str(content)))

        # Keep the frontend tool_calls on the AIMessage so each newly
        # appended ToolMessage stays paired with a matching tool_call_id
        # — ``_fix_messages_for_bedrock`` (inherited from the parent
        # class) strips ToolMessages whose id isn't present in any
        # AIMessage's ``tool_calls``, which would otherwise erase the
        # frontend results we just received. The built-in ToolNode
        # filters by its own registry before running, so leaving the
        # frontend calls in place doesn't cause it to try executing them.
        #
        # Routing: ``_make_model_to_tools_edge`` (in langchain's
        # ``agents/factory.py``) is the conditional edge that runs after
        # this hook when ``tool_node`` exists. It honours
        # ``state["jump_to"]`` first, then falls back to checking
        # pending tool_calls.
        #
        # - All-frontend turn: no backend calls left to run. We need to
        #   re-enter the model so it can react to the FE results.
        #   Setting ``jump_to: "model"`` covers both the ``tool_node``
        #   exists case AND the no-tool_node test case (where the
        #   default edge would otherwise route to END, silently
        #   stopping the run after every resume).
        # - Mixed turn: backend calls share the AIMessage. We must NOT
        #   set ``jump_to`` here — the default edge correctly ``Send``s
        #   only the pending backend calls to ToolNode (FE ones now
        #   have matching ToolMessages, so they're filtered out).
        #   Skipping ToolNode here would leave the backend tool_calls
        #   unanswered → model re-emits them → loop.
        update: dict[str, Any] = {"messages": new_tool_messages}
        if not backend_calls:
            update["jump_to"] = "model"

        return update

    async def aafter_model(
        self,
        state: StateSchema,
        runtime: Runtime[Any],
    ) -> dict[str, Any] | None:
        return self.after_model(state, runtime)


def _parse_tool_results(payload: Any) -> list[dict]:
    """Normalise the resume payload into a list of ``{toolCallId, content}``.

    The wire shape the client POSTs is
    ``forwardedProps.command.resume = {"tool_results": [...]}``;
    ``ag_ui_langgraph`` JSON-decodes string payloads before re-entering
    the graph, so ``payload`` is usually a dict. Tolerate a bare list
    """
    if isinstance(payload, dict):
        results = payload.get("tool_results")
        if isinstance(results, list):
            return [_coerce_result(r) for r in results if isinstance(r, dict)]
    if isinstance(payload, list):
        return [_coerce_result(r) for r in payload if isinstance(r, dict)]
    return []



def _coerce_result(raw: dict) -> dict:
    call_id = raw.get("toolCallId") or raw.get("tool_call_id") or raw.get("id")
    content = raw.get("content")
    if content is None:
        content = raw.get("result")
    if not isinstance(content, str):
        content = json.dumps(content) if content is not None else ""
    return {"toolCallId": str(call_id) if call_id is not None else "", "content": content}

```

</details>

2. The FE:
  - executes FrontEnd tool, 
  - then issues a `Command(resume=<all tools responses>)` , whether tools require human-in-the-loop interaction or not

I managed to make part 2 t work with some AI support. I can try workout how general this fix can be for JS CopilotKit and post a minimal example. But I would like some opinion first, as maybe this approach has already been already discarded for some reason.

But to me it looks the natural/correct way to use a LangGraph to have a robust experience.

Also possibly allows to remove the Orphaned Tools logic from the LangGraph implementation.

Let me know 😃 

- [x] I’m open to collaborating but need guidance.
- [ ] 
### Pre-flight Checklist

ltoniazzi:
## What does this PR do?

Adds `CopilotKitMiddleware(interrupt_frontend_tools=True)`.
An opt-in path that awaits frontend tool results **in the same turn** via LangGraph's `interrupt()`, instead of the default strip-and-restore (which carries several complications).

[Python middleware only]

#### Default path

On the default path `after_model` strips frontend tool calls off the `AIMessage` and `after_agent`
restores them, so the result only reaches the model on the *next* run. 

```
Agent calls any tools → FE calls stripped → turn ends → FE tool runs on the client → FE ToolMessage arrives on the next run
```
In a mixed turn the graph does not even end: the backend call runs and the frontend call is restored
with a `{"status": "forwarded_to_frontend"}` placeholder, since an unpaired `tool_call` is invalid.

So human-in-the-loop on a frontend tool is not possible here, as the graph moves on before the user is
asked, and their answer lands a run after the decision it should have gated. It is fragile too as there is no clear interruption contract for the langgraph graph.

#### New path

Meanwhile, with the new flag `interrupt_frontend_tools=True`:

```
Agent calls any tools → FE tool runs on the client → FE ToolMessage appended → agent continues
```

So the model builds `AIMessage` → `ToolMessage` straight away, as if it called an MCP server for
those tools.

<details><summary>How it works in mode detail</summary>

Basically:
1. After model response, collect all FE tools a model might have requested in parallel, 
2. then `interrupt()` and wait for FE to `resume` with all the FE tool responses (No matter if it's a fire-and-forget tool, HITL or other.)
3. After resume, agent continues as before (either run backend tools or loops back to model).

In more detail:

- `after_model` batches a turn's unanswered frontend calls into one `interrupt()`, keyed
  `__copilotkit_frontend_tool_calls__` so a client can distinguish it from a human-facing interrupt on
  the same thread.
- The client resumes with `{"tool_results": [{"toolCallId": ..., "content": ...}, ...]}`. The middleware
  appends one `ToolMessage` per call and the model re-enters the same turn.
- The `AIMessage` is left untouched, so each new `ToolMessage` has a matching id. Unanswered calls still
  get one, because an unpaired `tool_call` is rejected by Bedrock.
- One interrupt per turn rather than one per call: `after_model` runs once per model call, so every
  frontend call has to be answered before the routing edge evaluates, and `interrupt()` indexes its
  answers by call order.
- The state update carries `messages` only. No `jump_to`, because the `model`→`tools` edge already
  routes correctly. No `copilotkit` key, because that channel has no reducer and writing it would wipe
  `actions`, which the hook re-reads on resume.

</details>

### Open question: how should react/core handle an interrupt-borne tool call?

The agent side works within today's protocol surface. But the client side has no standard route at the moment (details below).

In the docs I added a **react "bridge"** one could use today to leverage the new opt-in interrupt flag. But this needs discussion, because it should also point in a direction that makes eventually this feature work out-of-the-box. 

<details><summary>What the react client does today</summary>

- **The default executor always runs a frontend handler, interrupt or not.** `runAgent` computes
  `newMessages` as an id-diff after the stream ends, regardless of how the run ended, so an
  interrupt-terminated run still reports the assistant message carrying the frontend `tool_calls`.
  `processAgentResult` finds no `ToolMessage` for the call, runs the handler, splices its own
  `ToolMessage`, and requests a follow-up run (`followUp` defaults to on). A bridge that acks instead of
  dispatching relies on this to execute fire-and-forget tools; a bridge that dispatches to
  `getTool().handler()` runs them twice.
- **A tool needing user input cannot be driven from the interrupt's `handler`.** Calling a handler
  directly never emits `onToolExecutionStart`, so `executingToolCallIds` never gains the id, the
  renderer never reaches `ToolCallStatus.Executing`, and a `useHumanInTheLoop` tool never receives a
  live `respond`. The promise never settles and the pause never ends. Resolving from the interrupt's own
  `render` works, so that is what the docs use.
- **AG-UI already models this case.** `Interrupt` carries `reason` and `tool_call_id`, `ɵInterruptState`
  derives `toolResults` from `reason: "tool_call"`, and `useInterrupt` appends the `ToolMessage` and
  resumes per interrupt id. None of it is reachable from LangGraph, because `ag_ui_langgraph` emits
  interrupts as a legacy `on_interrupt` CustomEvent carrying neither field.
- **The bridge in the docs is ~130 lines of protocol plumbing**: resolve from `render`, parse a value
  that arrives JSON-encoded, guard against Strict Mode resolving twice, and keep `enabled`/`render`
  identities stable so in-progress widget state survives.

</details>

<details><summary>Questions</summary>

The bridge I converged to feels confusing to me, because it's compromising between the `useInterrupt` and the `useFrontEndTool/useHumanInTheLoop` paths ([docs](https://docs.copilotkit.ai/teams/langgraph-typescript/human-in-the-loop#two-patterns-for-hitl-in-copilotkit)). 

I think this needs a bit of thinking from someone familiar with the react side of CopilotKit. I add here some observations and questions that arise around my bridge.

1. Should an interrupt-borne tool call go through the normal tool-execution lifecycle (emitting
   `onToolExecutionStart`/`End`, so renderers and `useHumanInTheLoop` behave as they do on the default
   path), or stay a separate track the bridge owns?
2. Is upstream `reason: "tool_call"` emission in `ag_ui_langgraph` the intended destination? If so, this
   flag's sentinel key and resume shape are transitional and should be documented as such.
4. Should react/core ship the bridge, so apps write no interrupt plumbing? If so, does the middleware
   keep classifying frontend calls from `copilotkit.actions`?
5. Should `processAgentResult` skip calls with a pending interrupt? It closes the double execution, but
   stops fire-and-forget tools that currently depend on it, so it needs a migration path.

</details>

<details><summary>Docs added</summary>

- `integrations/langgraph/frontend-tools` — new "awaiting frontend tool results in the same turn"
  section: agent setup, and a client bridge that claims only interrupts carrying the sentinel key,
  answers the batch once, and resolves from `render` so a call can ask the user something. Tools with no
  registered widget are acknowledged. This client bridge needs to be discussed in this PR.
- `integrations/langgraph/human-in-the-loop/interrupt-flow` — points readers there.

</details>

#### Testing

- `test_frontend_tool_interrupt.py` — end-to-end through real `create_agent` graphs 
- `test_copilotkit_lg_middleware.py` — unit coverage for batching, placeholder handling, etc

## Related PRs and Issues

- Closes #4920 
- ag-ui-protocol/ag-ui#1505
- ag-ui-protocol/ag-ui#2178

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Added an opt-in mode that pauses agent execution while frontend tools run, then resumes the same turn with their results.
  - Supports batching multiple frontend tool calls and handling unanswered results.
  - Mixed frontend and backend tool calls are handled appropriately.

- **Documentation**
  - Added setup and usage guidance for frontend tool interrupts, client-side result handling, and resume behavior.
  - Updated interrupt-flow documentation with frontend tool support details.

- **Tests**
  - Added coverage for interrupt, resume, mixed-tool, sequential-turn, and default behavior scenarios.

### DIFF
--- a/sdk-python/copilotkit/copilotkit_lg_middleware.py
+++ b/sdk-python/copilotkit/copilotkit_lg_middleware.py
@@ -14,10 +14,8 @@ Example:
     )
 """
 
-import asyncio
 import json
 import re
-import sys
 from typing import Any, Callable, Awaitable, ClassVar, Iterable, Optional, Union
 
 from langchain_core.messages import AIMessage, SystemMessage, ToolMessage
@@ -147,14 +145,6 @@ def _parse_frontend_tool_results(payload: Any) -> Any:
             # Last wins, so a client retrying one call in the same batch is fine.
             results[tool_call_id] = content
     return results
-
-
-def _in_async_task() -> bool:
-    """Whether we are executing inside a running asyncio task."""
-    try:
-        return asyncio.current_task() is not None
-    except RuntimeError:
-        return False
 
 
 def _current_thread_id() -> "str | None":
@@ -355,17 +345,13 @@ class CopilotKitMiddleware(AgentMiddleware[StateSchema, Any]):
             appended, so the model finishes the turn with the results in hand,
             in natural ``AIMessage`` → ``ToolMessage`` order. ``True`` requires:
 
-            1. **Python 3.11+ when async**, which the runtime always is: below
-               3.11 the run config does not reach asyncio tasks, so *any* async
-               ``interrupt()`` cannot read it. Raises ``CopilotKitMisuseError``,
-               not ``Called get_config outside of a runnable context``.
-            2. **An explicit client resume** — the default frontend-tool loop
+            1. **An explicit client resume** — the default frontend-tool loop
                fires a follow-up run with no resume command, which an
                interrupted thread ignores. Mount a handler that reads
                ``__copilotkit_frontend_tool_calls__`` off the interrupt and
                resumes with ``{"tool_results": [{"toolCallId": ...,
                "content": ...}, ...]}``.
-            3. **One interrupt per turn** — a resume cannot address multiple
+            2. **One interrupt per turn** — a resume cannot address multiple
                pending interrupts without interrupt ids
                (ag-ui-protocol/ag-ui#2178), so batching is mandatory.
     """
@@ -1187,34 +1173,6 @@ class CopilotKitMiddleware(AgentMiddleware[StateSchema, Any]):
             t.get("function", {}).get("name") or t.get("name") for t in frontend_tools
         }
 
-    @staticmethod
-    def _check_interrupt_preconditions() -> None:
-        """Fail early, and actionably, when interrupt mode cannot work.
-
-        On Python 3.10 the run config does not propagate into asyncio tasks, so
-        ``interrupt()`` cannot read it and dies with LangGraph's opaque ``Called
-        get_config outside of a runnable context``. This applies to any
-        ``interrupt()`` from an async node, not just this flag.
-        """
-        try:
-            from langgraph.config import get_config
-
-            get_config()
-        except RuntimeError as error:
-            if sys.version_info < (3, 11) and _in_async_task():
-                raise CopilotKitMisuseError(
-                    "CopilotKitMiddleware(interrupt_frontend_tools=True) needs "
-                    "Python 3.11+ when the agent runs asynchronously: on 3.10 the "
-                    "run config does not propagate into asyncio tasks, so "
-                    "LangGraph's interrupt() cannot read it. Upgrade to 3.11+, or "
-                    "leave the flag off to use the default frontend-tool flow."
-                ) from error
-            # Not inside a runnable context at all (e.g. a direct unit-test
-            # call) — nothing to validate.
-            return
-        except Exception:  # noqa: BLE001 - never block a run on a probe failure
-            return
-
     # Intercept frontend tool calls after model returns, before ToolNode executes.
     #
     # NOTE: this hook must stay free of side effects. In interrupt mode
@@ -1294,8 +1252,6 @@ class CopilotKitMiddleware(AgentMiddleware[StateSchema, Any]):
         # this node finds nothing outstanding instead of interrupting again.
         if not frontend_calls:
             return None
-
-        self._check_interrupt_preconditions()
 
         # One interrupt for the whole batch. A resume carries a single value and
         # cannot address several pending interrupts without their ids
--- a/sdk-python/copilotkit/copilotkit_lg_middleware.py
+++ b/sdk-python/copilotkit/copilotkit_lg_middleware.py
@@ -320,27 +320,16 @@ class CopilotKitMiddleware(AgentMiddleware[StateSchema, Any]):
             model (the host cannot supply the live, header-hooked model), and
             folds the registered catalog id + component schema into the params
             unless the host already set them — so host values win.
-        interrupt_frontend_tools: Await frontend tool results *in the same
-            turn* via LangGraph's ``interrupt()``, instead of the default
-            strip-and-restore.
+        interrupt_frontend_tools: Await frontend tool results in the same turn
+            via LangGraph's ``interrupt()``, instead of the default
+            strip-and-restore that only delivers them on the next run.
 
-            ``False`` (default) strips the calls off the ``AIMessage`` in
-            ``after_model`` and restores them in ``after_agent``, so the result
-            only lands on the next run, never in-run. ``True`` batches a turn's
-            calls into one ``interrupt()``: the graph pauses, the client runs
-            the tools and resumes, and one real ``ToolMessage`` per call is
-            appended, so the model finishes the turn with the results in hand,
-            in natural ``AIMessage`` → ``ToolMessage`` order. ``True`` requires:
-
-            1. **An explicit client resume** — the default frontend-tool loop
-               fires a follow-up run with no resume command, which an
-               interrupted thread ignores. Mount a handler that reads
-               ``__copilotkit_frontend_tool_calls__`` off the interrupt and
-               resumes with ``{"tool_results": [{"toolCallId": ...,
-               "content": ...}, ...]}``.
-            2. **One interrupt per turn** — a resume cannot address multiple
-               pending interrupts without interrupt ids
-               (ag-ui-protocol/ag-ui#2178), so batching is mandatory.
+            Batches a turn's calls into one ``interrupt()`` keyed
+            ``__copilotkit_frontend_tool_calls__``; the client resumes with
+            ``{"tool_results": [{"toolCallId": ..., "content": ...}, ...]}`` and
+            one ``ToolMessage`` per call is appended. Needs an explicit resume —
+            the default frontend-tool loop fires a follow-up run that an
+            interrupted thread ignores.
     """
 
     state_schema = StateSchema
@@ -1160,9 +1149,6 @@ class CopilotKitMiddleware(AgentMiddleware[StateSchema, Any]):
             t.get("function", {}).get("name") or t.get("name") for t in frontend_tools
         }
 
-    # Intercept frontend tool calls after the model returns, before ToolNode runs.
-    # Must stay side-effect free: in interrupt mode ``interrupt()`` raises out of
-    # here, so the node re-runs from the top on resume.
     def after_model(
         self,
         state: StateSchema,
--- a/sdk-python/copilotkit/copilotkit_lg_middleware.py
+++ b/sdk-python/copilotkit/copilotkit_lg_middleware.py
@@ -14,8 +14,10 @@ Example:
     )
 """
 
+import asyncio
 import json
 import re
+import sys
 from typing import Any, Callable, Awaitable, ClassVar, Iterable, Optional, Union
 
 from langchain_core.messages import AIMessage, SystemMessage, ToolMessage
@@ -26,7 +28,9 @@ from langchain.agents.middleware import (
     ModelResponse,
 )
 from langgraph.runtime import Runtime
+from langgraph.types import interrupt
 
+from .exc import CopilotKitMisuseError
 from .header_propagation import install_httpx_hook, set_forwarded_headers
 from .langgraph import CopilotKitProperties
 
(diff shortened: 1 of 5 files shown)


ltoniazzi:
@BenTaylorDev this is the pr discussed. 

I think it needs some alignment on how copilotkit's react/core handles this opt-in FE tools mode (see PR description)

BenTaylorDev:
Thanks for putting the open question at the top rather than burying it — and for
doing the client-side archaeology before proposing anything. I checked your five
claims against the source rather than taking them, and four hold exactly as
written. The fifth is wrong in a way that changes what you should ask upstream
for, so I want to lead with that.

## The correction that matters

You write that `ɵInterruptState` derives `toolResults` from `reason: "tool_call"`,
and conclude that upstream emitting that `reason` is the destination.

It does not. The derivation keys on the **presence of `toolCallId`**, not on
`reason` at all:

```ts
// packages/core/src/interrupt-state.ts:147-155
const toolResults = mutableInterrupts.flatMap((interrupt) => {
  if (!interrupt.toolCallId) return [];
  return [{ toolCallId: interrupt.toolCallId,
            content: toolResultContent(this.#responses[interrupt.id]!) }];
});
```

`reason` appears **zero times** in `interrupt-state.ts` and zero times in
`use-interrupt.tsx`. It is only ever written, by our own runtime producer
(`packages/runtime/src/agent/index.ts:1423` and `:1634`, both defaulting to
`"tool_call"`). Nothing downstream reads it.

So the upstream ask is not "emit `reason: "tool_call"`". It is:

> emit a **standard** interrupt carrying `toolCallId`, and set `outcome` on
> `RUN_FINISHED` — instead of the legacy `on_interrupt` CustomEvent.

If `ag_ui_langgraph` shipped `reason` alone, you would still land in the legacy
branch and still get no `ToolMessage`. Worth correcting before you file it.

## What checked out

Your other four claims are accurate, and two are worth restating because they
are load-bearing.

**The default executor really does run the handler regardless of outcome.** The
id-diff lives in AG-UI's `AbstractAgent.runAgent`, which snapshots message ids
before the pipeline and diffs after, with no branch on how the run ended. So an
interrupt-terminated run still reports the assistant message carrying the
frontend `tool_calls`, and `processAgentResult` executes it
(`run-handler.ts:569-602`), splices a `ToolMessage` (`:870-876`) and requests a
follow-up (`:878-880`, `followUp !== false`).

One thing you did not mention, and it sharpens your question 5: **that follow-up
is already dead on arrival.** `@ag-ui/client@0.0.57` — the version
`@copilotkit/core` pins — throws from `onInitialize` when pending interrupts are
not addressed by a resume (`Thread has N pending interrupt(s) not addressed by
resume`), which we catch into `AGENT_RUN_FAILED` at `run-handler.ts:507-519`. So
today the double-execution does not merely waste a call, it produces a failed
run. (The guard is absent in 0.0.53, so this is version-dependent.)

**The lifecycle chain really is closed.** `onToolExecutionStart` fires only from
inside the executor (`run-handler.ts:752-762` and `:938-948`); the provider's
`executingToolCallIds` is populated only by that event
(`CopilotKitProvider.tsx:723-741`); `ToolCallStatus.Executing` derives only from
that set (`use-render-tool-call.tsx:168`); and `useHumanInTheLoop` passes a live
`respond` only in the `Executing` branch (`use-human-in-the-loop.tsx:77-85`). One
precision: `copilotkit.runTool()` *does* route through the executor and emit the
event — but it mints a new `toolCallId`, so it cannot drive the interrupt's own
call. Your claim is right for the call that matters.

## Answers

**1. Normal lifecycle, not a separate track.** The chain above is one mechanism,
and a separate track either re-implements it or leaves HITL broken. Your bridge
is the evidence: it resolves from `render` specifically because the `handler`
path never marks the call executing, which your own docs say at the bottom of the
new section.

**2. Upstream, with the corrected ask above.** Once a standard interrupt carries
`toolCallId`, `ɵInterruptState` already derives the tool result and
`useInterrupt` already appends the `ToolMessage` and resumes per interrupt id.
Your sentinel key and resume shape are genuinely transitional and should be
documented as such.

**4. We should not ship the bridge as-is.** Reading it, its substance is a
`JSON.parse` of a value that arrives encoded, a `resolved` ref guarding Strict
Mode's second pass, `useCallback`-with-refs to stop identity churn discarding
widget state, and resolving from `render` instead of `handler`. All four are
compensations for the legacy event and the lifecycle gap, not domain logic.
Shipping them would freeze the workaround into our public surface.

Which raises the thing I would push back on hardest: **the bridge is in the
docs.** Publishing it as the recommended pattern commits us nearly as hard as
shipping it as API, and docs outlive the workaround they document. I would rather
not merge that section in its current form.

**5. Yes, and there is precedent.** The executor has no notion of interrupts
today — one doc-comment mention in `run-handler.ts`, no read of
`pendingInterrupts` anywhere in `packages/core/src`. But `executeFrontendTools`
already exists as a skip flag (`run-handler.ts:536-548`), set `false` for
`connectAgent` at `:394`. Your migration concern is real and is the hard part:
fire-and-forget tools currently depend on the executor running them, so whichever
path owns the interrupt has to take them over rather than leave them unrun.

## Where that leaves this PR

The Python middleware I have not reviewed yet, and I would like to keep that
separate from this design thread. My suggestion is to split: land the agent-side
work on its own merits, and let the client story follow the upstream change
rather than encoding today's gap into our docs.

Does the corrected upstream ask match what you are seeing from the
`ag_ui_langgraph` side?

kvnloo:
Thanks @ltoniazzi and @BenTaylorDev for separating the middleware work from the client ownership question.

One upstream detail may narrow the next step: at AG-UI `e62b348`, the LangGraph adapter already has an opt-in structured outcome (`emit_interrupt_outcome=True`), and [its interrupt normalizer](https://github.com/ag-ui-protocol/ag-ui/blob/e62b348680c41f52a5ea0ed4eb714a6600066fa1/integrations/langgraph/python/ag_ui_langgraph/interrupts.py) preserves `toolCallId`/`tool_call_id`. The [README](https://github.com/ag-ui-protocol/ag-ui/blob/e62b348680c41f52a5ea0ed4eb714a6600066fa1/integrations/langgraph/python/README.md) also explains why enabling that outcome with a legacy-resume client can strand a run.

That does not automatically map this PR's one batched pause to several tool calls, or settle which client path executes them. But it suggests a version-pinned round-trip fixture before committing applications to the bridge:

`one ordinary frontend call + one HITL call → real user response → one result per original call ID → graph continues with those actual results`

The ordinary handler should run once, the HITL control should receive its normal execution lifecycle, and an unanswered decision must not become a successful placeholder. A bridge that prevents duplication by skipping all handlers must fail the ordinary-call control.

Would targeting the existing standard outcome/resume path be the preferred client follow-up, with the batching translation kept at the adapter boundary? That could keep app authors out of the execution/replay plumbing without expanding this middleware PR.

AI-assisted source review; I have not run that paired client/adapter fixture or verified a minimum published version.

ltoniazzi:
@kvnloo Thanks, having a look at the new ag-ui-langraph interrupt standard and trying to look for a more consistent approach (setting PR in draft for now, will ping when done)

Issue #6933, opened 2026-09-07
docs(ag2): AG2 integration docs are still on the pre-1.0 autogen API


vvlrff:
### What's out of date

The AG2 integration docs under `showcase/shell-docs/src/content/docs/integrations/ag2/` — and the showcase integration they are generated against — are still written for the pre-1.0 `autogen` API. AG2 1.0.0 was published on 2026-07-27 and renamed the top-level module `autogen` → `ag2`, dropping `autogen/` entirely; the current release is **1.0.4**.

Eight AG2 doc pages still contain `autogen` imports (`auth.mdx`, `readables.mdx`, `frontend-tools.mdx`, `human-in-the-loop.mdx`, `shared-state/read.mdx`, `shared-state/write.mdx`, `generative-ui/tool-rendering.mdx`, `generative-ui/state-rendering.mdx`), so anyone following them with a current `pip install ag2` gets `ModuleNotFoundError` on the first import.

This is known and deliberate as a stopgap: [#6176](https://github.com/CopilotKit/CopilotKit/pull/6176) pinned `showcase/integrations/ag2/requirements.txt` to `ag2[openai,ag-ui]>=0.9.0,<1.0.0` to unblock CI when 1.0.0 landed, and its own comment says:

> Lift this bound only as part of a real 1.0 migration that ports the ~20 `autogen` import sites and re-verifies every ag2 demo cell.

Four releases later (1.0.0 → 1.0.4) the pin is still there, so the docs describe an AG2 that `pip` no longer installs.

### The ask

Update the AG2 docs to the 1.0 API. That migration is already written and open as [#5938](https://github.com/CopilotKit/CopilotKit/pull/5938) (opened 2026-07-13, no review yet): it ports every `autogen` import site, lifts the `<1.0.0` bound, and adapts to the 1.0 behaviour the pin comment calls out — the deleted `AgentService`-based `AGUIStream`, `dispatch(context=…)` → `variables=`, and the dropped `event_interceptors` kwarg. Every claim in it was cross-checked against the `ag2==1.0.1` sources.

Opening this as an issue because the PR alone hasn't drawn a look in eight weeks, and the stale docs are a user-visible problem independent of whether that particular PR is the shape you want. If you'd rather split it — docs-only first, showcase harness changes second — say so and I'll cut it that way. I maintain AG2 and am happy to keep these pages current as the framework moves.

### References

[...]

PR #7109, opened 2026-09-12
docs(ag2): update integration docs and examples for AG2 1.0

vvlrff:
## Problem

AG2 1.0 reshaped the public API, but every AG2 code sample in the docs still targets 0.x. On `main` there are **127 references across 10 pages** to symbols that no longer exist — `autogen`, `ConversableAgent`, `LLMConfig`, `ContextVariables`, `register_for_llm`, `register_for_execution`.

Following the AG2 quickstart today fails with an `ImportError` on the first import.

## What changed

All AG2 documentation migrated to the 1.0 API:

| 0.x (current `main`) | 1.0 (this PR) |
| --- | --- |
| `from autogen import ...` | `from ag2 import ...` |
| `ConversableAgent(system_message=, llm_config=LLMConfig(...))` | `Agent(prompt=, config=OpenAIConfig(...))` |
| `register_for_llm` / `register_for_execution` | the `@tool` decorator |
| `ContextVariables` | conversation variables (`Context.variables`) + `Variable()` injection |
| `context["copilotkit"]["context"]` dict digging | `RunAgentInput.context` (typed items with `.description` / `.value`) |

New content:

- **quickstart** — a "Connect your own AG2 agent" section: `pip install "ag2[ag-ui,openai]"`, the manual `AGUIStream.dispatch` pattern, and the `AGUIStream.build_asgi()` shorthand.
- **generative-ui/a2ui** — AG2 1.0 ships a first-class `ag2.a2ui` module, so `A2UIServer` + `AgUiTransport` and `@a2ui_action` replace hand-written bridge code.
- **multi-agent/subagents** — new page covering delegation and how it surfaces in the chat.
- **auth** — identity is threaded through a server-side dependency instead of a context variable, so it is never streamed to the client as state.

## Two bugs fixed along the way

1. `prebuilt-components.mdx` and `interactive.mdx` never passed `framework` into the shared snippets, which fall back to `props.framework || "langgraph"` — so **the AG2 pages were embedding LangGraph examples**. They now pass `framework="ag2"`.
2. AG2 doc links pointed at `docs.ag2.ai/latest/docs/...`; the canonical path no longer carries the `/latest/` prefix. To be precise these URLs were stale rather than broken — `/latest/` still resolves, but serves the 0.9 `autogen` docs behind an upstream "newer version available" banner.

## Verification

- Every snippet checked against the AG2 1.0 user guide — `Agent`, `OpenAIConfig`, `AGUIStream.dispatch` / `build_asgi`, `Context.variables`, `Variable()`, `@tool`.
- All 26 Python snippets compile.
- Every `meta.json` entry resolves; MDX fences and JSX tags balance.

## Scope and context

Docs only — 22 files. Most are `.mdx` / `.json` under `showcase/shell-docs/src/content/`; review feedback added two paths outside it: `showcase/integrations/ag2/docs/setup/frontend-tools-setup.mdx` (a setup cell that renders into `/ag2/frontend-tools` and was still on 0.x) and the `auth_pattern` description in `showcase/shared/manifest.schema.json`. No executable code changes, so this can land independently of the integration work.

### The showcase itself still runs AG2 0.x

Worth saying out loud: `showcase/integrations/ag2/requirements.txt` still pins `ag2[openai,ag-ui]>=0.9.0,<1.0.0`, and 19 files under `showcase/integrations/ag2/src/` still import `autogen`. These pages therefore describe an API the showcase does not execute yet. Lifting that upper bound here would break it rather than help — the `autogen` module is gone in 1.0, and `_multimodal_normalize.py:330` still calls `super().dispatch(incoming, context=...)`, which 1.0 replaced with `variables=`, so an `ImportError` would simply become a `TypeError`. Porting those call sites is #5938's job, and that branch already moves the pin to `ag2[openai,ag-ui,a2ui]>=1.0.0`. The pin stays untouched here on purpose.

Reported in #6933. The documentation here is split out of #5938, which additionally migrates the AG2 showcase integration; that PR is larger and needs a closer review, while these doc pages are wrong on `main` today and can be fixed on their own.

## Summary by CodeRabbit

- **Documentation**
  - Updated AG2 integration guides for current authentication, tools, shared state, human-in-the-loop, readables, and frontend integration patterns.
  - Added guides for A2UI rendering, AG2 subagents, custom agents, and connecting AG2 endpoints to CopilotKit.
  - Expanded guidance on conversation variables, server-side dependencies, authorization, state snapshots, and intermediate streaming.
  - Refreshed examples, framework labels, component configuration, tool rendering, and AG2 documentation links.
  - Added new pages to AG2 documentation navigation and Generative UI navigation.


NathanTarbert:
Hey @vvlrff,

Thanks for splitting this out of #5938. The AG2 pages really are broken on main right now, and there's no reason they should have to wait on the showcase work to get fixed.

The 1.0 API surface all holds up. `Agent`, `OpenAIConfig`, `AGUIStream.dispatch` and `build_asgi`, `Inject` with `dependencies`, the `ag2.a2ui` module, `Agent.as_tool` — all of it matches what `ag2` 1.0.5 actually ships, down to the `task:<agent name>` step names on the subagents page.

A few things to sort out before this goes in.

The one I'd fix first is in `readables.mdx`. `get_readable` already decodes its JSON before returning, so when the endpoint calls `json.loads` on the result again a few lines down, anyone who copies that snippet gets `TypeError: the JSON object must be str, bytes or bytearray, not list` on their first request. Dropping the second decode and falling back to an empty list is all it needs.

Two of the `useRenderTool` calls are missing their `parameters` schema — the `get_weather` one on the tool rendering page, and the new one in `subagents.mdx`. The hook only accepts two shapes, either the wildcard `name: "*"` form or a named tool with a schema, so a named registration without one doesn't type-check and `parameters.location` has nothing behind it.

One page slipped through the migration. `showcase/integrations/ag2/docs/setup/frontend-tools-setup.mdx` is still importing `ConversableAgent` and `LLMConfig`, and it gets rendered into `/ag2/frontend-tools` through the `FrameworkSetup` component, so a reader would still hit a 0.x snippet right after this merges.

Worth saying in the description too: the showcase's own `requirements.txt` still pins `ag2` below 1.0, so until #5938 lands these pages describe an API the bundled showcase can't actually run. Nothing to change here, it's just better stated up front than discovered later.

A handful of smaller things, none of them blocking on their own:

- The new A2UI page won't appear in the sidebar, because `generative-ui/meta.json` doesn't list it and has no catch-all entry — so those 84 lines are currently unreachable. mastra's version of that file lists it explicitly.
- Further down that same page, the runtime block uses `copilotRuntimeNextJSAppRouterEndpoint`, while the shared snippet just above it points readers at `createCopilotRuntimeHandler` and calls the other one legacy. Both work, but the page ends up arguing with itself within a screen.
- In `shared-state/read.mdx` the `useAgent` link still goes to the v1 reference even though the snippet imports from v2, which is the same fix you already made over in `frontend-tools.mdx`.
- The auth example never grants anything, since `validate_your_token` doesn't return `allowed_accounts` and the permission check below it reads that key — so every call comes back forbidden.
- The manifest schema still describes `ag2-context-variables` as threading `ContextVariables` to tools, which is exactly what this PR moves away from.

One note on the link change: the `/latest/` URLs do still resolve. They just serve the old 0.9 `autogen` docs with an upgrade banner on top. Dropping the prefix is still the right call, the description might just want to say those pages are stale rather than broken.

And if you'd rather move the subagents page to `ag2/subagents.mdx` so it matches where the other integrations keep theirs, that's fine by me. Not something I'd hold this up for.

vvlrff:
Hello @NathanTarbert, I have made some edits, please take a look

NathanTarbert:
Hey @vvlrff, thanks, taking a look!

NathanTarbert:
Hey @vvlrff,

Everything from last round is sorted — readables, the setup page, both schemas, the runtime block, the sidebar entry, the auth stub, the manifest description. The `objective` and optional `context` arguments you documented are right as well. That's the signature every `as_tool()` delegation gets.

A few new things came up on this pass.

The important one is in `readables.mdx`, near line 229. When you pass `prompt=` to `dispatch`, it doesn't add to the agent's prompt. It takes over completely. The agent gets built with `"You are a helpful assistant."` a few lines earlier, and that instruction never reaches the model — it runs on the readables alone. What makes this one rough is that nothing fails. The agent still answers, just without the instructions you gave it, and there's no reason you'd think to look at the prompt. Building the list with the instruction first, like `prompt = ["You are a helpful assistant.", *readables]`, keeps both. One thing that explains the trap: ag2's other transports merge the agent's prompt back in for you. AG-UI's is the only one that doesn't, so the docs have to do it by hand.

The list itself is correct, by the way. `dispatch` wants an iterable of strings, and a plain string would get split into individual characters.

Next is `shared-state/read.mdx` and `write.mdx`, both at line 118. The snippet is careful about `agent.state` being empty early on — it writes `?? {}` twice in the lines just above. Then the next line reads `agent.state.language` straight, which crashes in that same early window. The `state` variable right above it is already safe, so reading `language` off that one clears it.

The auth page has a gap between what it says and what it shows. Line 137 tells the reader to register the tool with `tools=[get_account_data]`, but neither `Agent(...)` block on the page includes `tools=`. Follow the page start to finish and you end up with a tool the model can't call. The shared auth page does the same thing at line 319.

Adding `a2ui` to `generative-ui/meta.json` fixed that page's sidebar problem, but `mcp-apps` and `your-components` are missing from the same list, so those two are still invisible in the sidebar. agno's copy of that file lists all four, and you're already in there.

`manifest.schema.json` has a similar leftover. You updated the description to talk about dependencies and `Inject`, which is right, but the key it describes is still named `ag2-context-variables`. Renaming it would stop someone grepping for context variables later and drawing the wrong conclusion about what the flag does.

There's also a `type AgentState` near line 114 of `state-rendering.mdx` that nothing uses. `useAgent` is called without it, so it just sits there. Either hand it to the hook or drop it.

Last thing is a decision rather than a fix. `frontend-tools-setup.mdx` lives inside `showcase/integrations/ag2`, and that package still pins `ag2` below 1.0 — there's a note in its `requirements.txt` saying the rename would break around twenty imports. So someone who installs that package and follows the snippet you just migrated hits `ModuleNotFoundError: No module named 'ag2'`. The title on that snippet also points at `src/agents/chart_agent.py`, which isn't in the package. I don't think this should hold the docs back. They're accurate about AG2 and ahead of our showcase, which is the direction I'd rather be wrong in. But it's worth saying out loud whether these land now with an issue filed for the showcase bump, or wait for #5938. My vote is land them.

Thanks for your continued work on this.

vvlrff:
Hello @NathanTarbert, I’ve made the changes based on the review

PR #7113, opened 2026-09-13
docs(react-core): document useHumanInTheLoop's followUp option

### LINKED ISSUE #3206: 🚀 Feature Request: Respond to tool call without followUp

nuthalapativarun:
## What does this PR do?

#3206 asks for a way to record a `useHumanInTheLoop` response without immediately resuming the agent, so a rejected/edited tool result and a separate user follow-up message can reach the agent together in one turn.

That capability already exists: `ReactHumanInTheLoop<T>` is `Omit<FrontendTool<T>, "handler"> & {...}`, and `FrontendTool` already has a `followUp?: boolean` field. `respond()` resolves the same promise every tool handler execution awaits in `packages/core/src/core/run-handler.ts`, and the *same* `tool?.followUp !== false` check gates whether the run auto-continues after any tool handler resolves (`executeToolHandler`/`executeSpecificTool`). Setting `followUp: false` on a `useHumanInTheLoop` tool already means: `respond()` records the tool result in message history immediately, but does **not** trigger a new run — the caller can then add more messages (e.g. `agent.addMessage(...)`) and call `copilotkit.runAgent({ agent })` once, exactly matching the requester's desired flow. This is the same two-step pattern already documented elsewhere in these docs (e.g. `frontends/react-native.mdx`, `human-in-the-loop/headless.mdx`).

The gap was purely documentation: `followUp` wasn't listed in `reference/hooks/useHumanInTheLoop.mdx`'s Parameters (the Vue equivalent already documents it), and there was no worked example showing the combined-response-plus-follow-up-message pattern. This PR:

- Adds `followUp` to the Parameters section of `reference/hooks/useHumanInTheLoop.mdx`.
- Adds a new "Combining a Response with a Follow-up Message" usage example demonstrating the exact flow from the issue.

No code changes — the feature already works; it just wasn't discoverable.

## Related PRs and Issues

- Closes #3206

## Checklist

## Summary by CodeRabbit

- **Documentation**
  - Documented the new `followUp` option for human-in-the-loop tool parameters, including its default behavior.
  - Added guidance and an example for combining a recorded response with a follow-up message before resuming an agent run.


BenTaylorDev:
Thanks for documenting `followUp` here. One change before this can merge: please switch `Closes #3206` to `Refs #3206`. The `followUp: false` plus manual `runAgent` pattern covers the rejection flow, but it makes every response resume by hand, including approvals. #3206 asks for a per-call override so the default can stay on, and that work is still planned (design in the issue thread). As written, merging this would close the feature request.

PR #7118, opened 2026-09-14
refactor(react-native)!: replace the useRenderTool shim with react-core's hook

### LINKED ISSUE #6976: Remove the deprecated useRenderTool shim from @copilotkit/react-native
Tracks removal of the deprecated `useRenderTool` compatibility shim added to `@copilotkit/react-native` in #6533.

## Why the shim exists

#6533 converged React Native's render-tool hooks onto react-core. RN's `useRenderTool` was never react-core's `useRenderTool` — its body forwarded to a *different* core hook, `useFrontendTool`, while wearing the other one's name, which is why `name: "*"` registered a frontend tool literally named `*` instead of a wildcard renderer.

`@copilotkit/react-native` sits in the `monorepo` release scope — 16 packages versioned in lockstep — so a major bump is not available and the break ships in a minor. The `!` marker buys a release-note line and nothing more. That left the compiler as the only signal reaching consumers, and there are call shapes it cannot see (a hoisted config whose `render` ignores its props, old fields arriving via a spread, and untyped or `@ts-nocheck` call sites). The shim exists so those consumers get a runtime warning rather than a silent behaviour change.

## What to remove

- `packages/react-native/src/hooks/useRenderTool.ts` — the shim, its route table, and its warn-once dedup
- its tests under `packages/react-native/src/hooks/__tests__/`
- the delegation test and graph-wide registry deny-list in `packages/react-native/src/__tests__/headless-entry-surface.test.ts` should revert to the simpler identity assertion (`mod.useRenderTool === coreHeadless.useRenderTool`), which is what they replaced
- the `@deprecated` blocks and the "removal in the next minor" wording, which appears in 5 files
- the shim's route table and deprecation sections in `showcase/shell-docs/src/content/reference/react-native/hooks/useRenderTool.mdx` and `packages/react-native/USAGE.md`

After removal, `@copilotkit/react-native` re-exports react-core's `useRenderTool` and `useFrontendTool` directly and carries no render-tool implementation of its own.

## What must stay

- `FrontendToolRenderFunction<T>` — the opt-in type that narrows a renderer's return to `React.ReactElement | null`. It exists because `useFrontendTool`'s `render` is a `React.ComponentType` returning `ReactNode`, so a bare string compiles and then throws *Text strings must be rendered within a `<Text>` component* on device. Unrelated to the shim's lifetime.
- The migration table on the docs page. Consumers will still be arriving at it after the shim is gone.

## Before removing

The shim's renderer-only route warns unconditionally (see #6533's third round of review) precisely because untyped-JS callers omitting `description` were losing tool registration silently. Removing the shim removes that warning too, so it is worth checking whether anything still needs to reach that population — a note in the release notes for the removing version is probably the answer.

Requested by @BenTaylorDev during #6533's review: "removal in the next minor" appears 11 times across 5 files with no issue behind it, which is how a shim becomes permanent.

mxmzb:
Closes #6976 (Linear OSS-1148). Follow-up to #6533, which converged React Native's render-tool hooks onto react-core but kept the old call shapes working behind a deprecated routing shim. This removes the shim: `@copilotkit/react-native` now re-exports react-core's `useRenderTool` itself and carries **no render-tool implementation of its own**.

## What the shim was

#6533 deleted RN's local `useRenderTool` — whose whole body forwarded to core's *other* hook, `useFrontendTool`, so `name: "*"` registered a frontend tool literally called `*` and auto-answered every otherwise-unanswered tool call with an empty tool result. In its place it shipped a 501-line shim that routed by config shape:

| call shape | routed to |
| --- | --- |
| `name: "*"` | core's `useRenderTool` (unconditional — wildcard always wins) |
| carries `handler` or `description` | `useFrontendTool` (what the old hook did) |
| anything else | core's `useRenderTool` |

plus a dev-only, once-per-tool-name `console.warn` on every route.

It shipped in 1.68 and said "scheduled for removal in the next minor" in 11 places across 5 files. We are on 1.71.

## What this removes

- `packages/react-native/src/hooks/useRenderTool.ts` — the shim, its route table, its frozen-route ref, its drift warning, its warn-once dedup, and the `__resetRenderToolShimWarnings` test hatch
- the shim's routing and warning tests (two of the three routing rules, six warning tests)
- the shim's route table, warning transcripts, and deprecation sections from the docs

`src/headless.ts` goes back to a plain `export { useRenderTool } from "@copilotkit/react-core/v2/headless"`.

## What this keeps

- **`FrontendToolRenderFunction<T>`** — the opt-in type narrowing a `useFrontendTool` renderer's return to `ReactElement | null`. Unrelated to the shim's lifetime; a bare string still typechecks on `useFrontendTool`'s `ComponentType` render and still throws *Text strings must be rendered within a `<Text>` component* on device.
- **The migration table** on the reference page. Consumers coming from a pre-1.68 version still arrive at it. Its last column is rewritten from "where it stands today" to the compiler error each row now produces.
- **The graph-wide registry deny-list** in `headless-entry-surface.test.ts`. #6976 proposed reverting it along with the delegation test, but it is not the shim's guard — it catches a competing registry in a module that never touches either hook's name, which identity cannot see. RN shipped exactly that once (a local `Map` behind a `RenderToolProvider`), and the consequence was `useComponent` rendering nowhere on RN.

## The guard goes back to identity, and gains a source check

`headless-entry-surface.test.ts` was written to pick its instrument from the entry source: identity when the re-export comes from core, delegation-plus-deny-list when it comes from a local module. With the shim gone it takes the identity branch on its own — so the test went green without an edit, which is precisely the failure mode worth closing.

The rewritten test therefore asserts **both**: runtime identity (`mod.useRenderTool === coreHeadless.useRenderTool`) *and* that `src/headless.ts` re-exports the name straight from `@copilotkit/react-core/v2/headless`. A local module that re-exports core's binding unchanged today passes identity and could grow a body tomorrow; requiring the specifier means an RN-local module cannot be introduced under this name at all.

## Behaviour tests: what survived, and why

The shim's suite drove a real `CopilotKitCoreReact` rather than mocks, so most of it was never about the shim. Kept:

- the wildcard paints for an unmatched call **and** registers no tool named `*` (`getTool({ toolName: "*" })` undefined, `core.tools` clean) — the original defect, still pinned
- a render-only registration produces no callable tool and does not shadow a same-named server tool: it paints the call and advertises nothing on a real run
- `agentId` scoping still paints under a different agent — a **pinned limitation**, not a passing behaviour
- a named renderer lands in **core's** registry and advertises nothing — the behavioural half of "RN owns no registry"

Deleted: two of the three routing rules and the six warning tests — 8 of the file's 14. `292 passed / 23 files`, down 8 from 300.

## The population the warning reached

#6976 asked this be checked before removal. The shim's warning existed for three call shapes TypeScript's excess-property check cannot see — a hoisted config, old fields arriving via a spread, and untyped or `@ts-nocheck` call sites. In all three the call compiles clean, core's hook ignores `description`/`handler`, and because the renderer bridge spreads `{ ...props, parameters: props.args }`, an old `render: ({ args }) => …` keeps painting. The screen looks unchanged while the tool stops being advertised and the handler never runs.

There is no mechanism left that reaches them, and inventing one is out of scope for a removal. What this PR does instead is promote that section on the docs page from a footnote about why the shim exists to the section that tells a reader to **audit by hand**, with the three shapes spelled out as code. The `!` marker buys the release-note line.

## Testing

```
react-native vitest             292 passed / 23 files    ✅
public-api manifest --check     current, no regen needed ✅
react-native check-types        + 37 dependency tasks    ✅
react-native build              + 37 dependency tasks    ✅
react-native test:scripts       26 pass / 0 fail         ✅
oxlint packages/react-native    0 errors                 ✅
```

Same caveat as #6533, restated because it has not changed: **nothing here proves the package works on a real device.** The RN suite runs in jsdom against a 52-line `react-native` stub; no CI job runs Metro, a simulator, or Detox, and `examples/v2/react-native/demo` is still not built or typechecked in CI. What is verified is types, export surface, and behaviour against core's real registry.

## ⚠️ BREAKING

`useRenderTool` on `@copilotkit/react-native` no longer accepts `description` or `handler`, and a call carrying them no longer registers a tool. Rename it to `useFrontendTool` — same config object. The diagnostic is `TS2769: No overload matches this call` (verified against `tsc`), with the excess property named in the nested per-overload detail rather than on the first line. A render body reading `args` still needs the rename to `parameters`; the shim never covered that either.

This package sits in the `monorepo` release scope (16 packages in lockstep), so — as with #6533 — this lands in a minor and the `!` marker buys a release-note line rather than a major.

## Summary by CodeRabbit

* **Breaking Changes**
  * React Native’s `useRenderTool` now matches the core hook and no longer supports configurations containing `description` or `handler`.
  * Update those calls to `useFrontendTool` with the same configuration.
  * The wildcard name `*` no longer registers a frontend tool.

* **Documentation**
  * Updated React Native guides and reference materials with the new hook behavior, migration guidance, and type-checking changes.


BenTaylorDev (approved):
Approving. Rebased onto main, so this is current.

I went looking for a reason to block this and did not find one:

- no dangling reference to the deleted `hooks/useRenderTool`, and `index.ts` is untouched
- no in-repo consumer uses the `handler` or `description` shape
- no third doc surface. No bundled skill documents this hook, so the two MDX files here are the whole set
- every shim mention left in the tree is past tense
- the entry guard got stronger, not weaker. It now requires both runtime identity and that `headless.ts` re-exports straight from `@copilotkit/react-core/v2/headless`, and it keeps the graph-wide registry deny-list
- `package.json` is untouched, so the public-API manifest needs no regeneration

The framing in `headless.ts` is right, which matters because #6533 got it wrong. Core never advertises a tool named `*`, because `buildFrontendTools` filters the name. The real consequence is that `*` is core's catch-all handler name, so a display-only wildcard auto-answered every otherwise-unanswered tool call with an empty tool result. That matches what I measured on main.

Two notes:

1. The three `expect(agent.advertised).toEqual([[]])` assertions in the retained suite are vacuous. They pass identically against the old hook. The assertions that actually carry the defect are the ones beside them: `core.getTool({ toolName: "*" })` undefined and `core.tools` excluding `*`. Worth trimming or re-pointing next time you are in that file.
2. Your own caveat is the real decision here, and I agree with how you framed it. For a hoisted config, a spread, or an untyped call site, excess-property checking never fires, so those callers get a silent behavior change with the docs as the only mitigation. The shim shipped in 1.68 with a dev warning and we are on 1.71, so I am fine with that trade.

PR #7132, opened 2026-09-14
fix(docs): return true 301s for retired AG-UI URLs

tylerslaton:
## Problem

The retired `/ag-ui/*` mirror redirects were described and tested as HTTP 301s, but `permanent: true` makes Next.js emit HTTP 308.

## Why

These URLs carry accumulated search ranking, and the chosen migration behavior is an explicit 301 to the canonical `docs.ag-ui.com` pages.

## Fix

Use `statusCode: 301` for every AG-UI mirror redirect, including suffix-aware rules and the nine mapped exceptions. Update regression coverage to assert the explicit status and verify representative routes against a production build.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected retired AG-UI mirror redirects to explicitly return HTTP 301 responses.
  * Preserved existing redirect behavior for all other routes.

* **Tests**
  * Updated redirect validation to confirm HTTP 301 status codes and the absence of legacy redirect properties.


BenTaylorDev (approved):
Approving.

**The mechanism is right.** Next.js emits 308 for `permanent: true` and never 301, so `statusCode` is the only way to get the status these rules were documented and tested as returning.

**The conversion is complete.** I verified this structurally rather than by keyword search. Every `docs.ag-ui.com` destination in `next.config.ts` is constructed in exactly two places — `agUiMirrorRedirects()` (lines 330, 335, 340) and the three suffix-aware catch-alls (lines 373, 378, 383). All six are converted, so no retired mirror rule is left on 308.

**The scoping is correct.** Four `permanent: true` rules still mention `ag-ui`: `/unselected/ag-ui`, `/built-in-agent/ag-ui`, `/learn/ag-ui-protocol` and `/concepts/ag-ui-protocol`. Those are internal hops to in-site destinations, not the retired mirror, and they should stay 308. Good call leaving them alone.

**The regression guard is real.** `expect(exception).not.toHaveProperty("permanent")` fails if anyone reverts the shape, so this cannot silently regress to 308.

`build-check (shell-docs)` is green, which validates the config against a real Next build rather than only unit-testing the array.

One non-blocking note for the record. The stated rationale is accumulated search ranking, but Google treats 301 and 308 as equivalent permanent signals and passes ranking through both. The change is still correct — the real justification is that the code did not match its documented and tested intent. Worth knowing so nobody goes hunting for other 308s expecting an SEO win. If you did observe a specific client mishandling 308, that is worth capturing in an issue, because it would mean there are other places to check.

Merging with admin, since the branch is behind and `main` enforces strict up-to-date.

PR #7136, opened 2026-09-15
fix(examples): point strands-python's Next.js runtime at the 8123 sidecar

jpr5:
## Problem

The `strands-python` starter is red on the `@chat` rung in showcase staging with `RUN_ERROR` / ECONNREFUSED, while `@health` and `@agent` stay green.

`entrypoint.sh` starts the Python sidecar on **8123** (`AGENT_PORT=8123`), but `src/agent.ts` defaults to **8000**:

```ts
url: `${process.env.AGENT_URL || process.env.STRANDS_AGENT_URL || "http://localhost:8000"}/`,
```

Nothing in the image sets `AGENT_URL`, so the Next.js runtime dials a port nothing listens on.

**Why the other rungs stayed green:** neither `/` nor `/api/copilotkit/info` dials the sidecar, so both answer 200 regardless.

**Why CI never caught it:** `docker-compose.test.yml` builds a *separate two-container* stack (`docker/Dockerfile.agent` + `docker/Dockerfile.app`) and passes `AGENT_URL=http://agent:8000` explicitly. The production **single-image** path (root `Dockerfile` + `entrypoint.sh`) — the one staging deploys — is never exercised by the compose smoke test.

## Which side I changed, and why

I changed the **app side** (make it dial 8123), not the sidecar.

The repo has two established conventions, and `strands-python` belongs to the 8123 family:

| starter | sidecar port | how the app is pointed at it |
|---|---|---|
| `claude-sdk-python` | 8123 | `export AGENT_URL="${AGENT_URL:-http://localhost:8123}"` |
| `claude-sdk-typescript` | 8123 | `export AGENT_URL="${AGENT_URL:-http://localhost:8123}"` |
| `strands-typescript` | 8123 | `export AGENT_URL="${AGENT_URL:-http://localhost:8123}"` |
| **`strands-python`** | **8123** | **(missing)** ← the bug |
| `pydantic-ai`, `adk`, `agno`, … | 8000 | app default already 8000 |

`claude-sdk-python/entrypoint.sh` is a near-verbatim template of this file — same header, same structure, same comments — differing only by the three lines this PR restores. `git log` confirms the file was added whole and never had them. So this is a dropped line, not a design choice.

Moving the sidecar to 8000 instead would also work, but it would fight `main.py`'s deliberate `AGENT_PORT` parameterisation and split `strands-python` from its three closest siblings.

[...]

PR #7141, opened 2026-09-15
fix(examples): stop starters running dev-mode servers in prod, and shellcheck starter entrypoints

jpr5:
Two starter-hygiene problems, both found while chasing something else.

## 1. Production containers were running dev-mode servers

`examples/integrations/strands-python/entrypoint.sh` ran `uv run python main.py`, and `crewai-crews/entrypoint.sh` ran `python server.py`. Both of those `__main__` paths are the **dev** entry points and call uvicorn with `reload=True` — so the shipped production image started a StatReload file watcher. The sibling `claude-sdk-python` entrypoint already avoids this deliberately; both now match it by invoking uvicorn directly. The `__main__` blocks are untouched, so local dev keeps hot-reload.

### Full dev-settings sweep (all 15 starters that ship a Dockerfile)

| Starter | Agent launched by entrypoint | Dev settings in prod image? |
|---|---|---|
| **strands-python** | `uv run python main.py` | **YES — `reload=True` → StatReload. FIXED** |
| **crewai-crews** | `python server.py` | **YES — `reload=True` → StatReload. FIXED** |
| adk | `python main.py` | No — its `__main__` calls `uvicorn.run(app)` with no reload |
| agno | `python -m uvicorn main:app` | No — bypasses the `reload=True` `__main__` |
| llamaindex | `.venv/bin/python -m uvicorn main:app` | No — same |
| ms-agent-framework-python | `.venv/bin/python -m uvicorn src.main:app` | No — same |
| pydantic-ai | `.venv/bin/python -m uvicorn src.main:app` | No — same |
| langgraph-python / langgraph-fastapi | `python serve.py` | No — `serve.py` has no reload (the reload lives in `agent/main.py`, which prod never runs) |
| claude-sdk-python | `uv run uvicorn main:app` | No — this is the reference pattern |
| claude-sdk-typescript / strands-typescript | `npx tsx src/server.ts` | No — the `--watch` variant is only in `package.json`'s `dev` script |
| ms-agent-framework-dotnet | `./ProverbsAgent` | No |
| mastra | (no agent sidecar in entrypoint) | No |
| langgraph-js | `npx @langchain/langgraph-cli dev` | **Ambiguous — reported, not fixed.** `dev` is a hot-reload dev server, but the LangGraph JS CLI has no non-dev serve mode outside LangGraph Platform, so there is no drop-in fix. Flagging for a separate decision. |

`ENV NODE_ENV=production` is already present in all 15 Dockerfiles — no gap there.

The five reload-by-`__main__` starters marked "No" are safe only because the entrypoint happens to bypass `__main__`. That is one careless edit away from regressing, which is part of why item 2 exists.

## 2. Nothing linted starter entrypoints

The `Shell script tests (bats + shellcheck)` job shellchecked exactly three `showcase/scripts/*.sh` files. The file that is literally the production container's `CMD` was linted by nothing.

This PR extends that job with a **globbed** `examples/integrations/**/*.sh` pass at `--severity=warning` (same bar as the release-scripts job), plus a guard that fails if the glob matches nothing so it can't silently rot. `showcase_validate.yml` already triggers on `examples/integrations/**`, so no trigger change was needed. 51 scripts are now covered.

### What the gate actually catches — and what it does not

**Honest answer first: it does NOT catch the #7136 class.** The pre-fix `strands-python/entrypoint.sh` was missing `export AGENT_URL`, and shellcheck is silent on it, because a variable the script never references but a *child program* reads is not a shell defect. Verified on the unmodified file:

```
$ grep -c AGENT_URL examples/integrations/strands-python/entrypoint.sh
0
$ shellcheck --severity=warning examples/integrations/strands-python/entrypoint.sh
In .../strands-python/entrypoint.sh line 23:
HOSTNAME=0.0.0.0 PORT=${PORT:-3000} ./node_modules/.bin/next start -H 0.0.0.0 -p "${PORT:-3000}" 2>&1 &
                 ^--^ SC2097 (warning): This assignment is only seen by the forked process.
                                                                                  ^-----------^ SC2098 (warning): This expansion will not see the mentioned assignment.
```

No `AGENT_URL` finding. Buying the gate does not buy protection against a dropped `export`. Catching that needs a container smoke test, which is separate in-flight work.

**What it does catch** is the adjacent family: SC2097/SC2098 fired on ten of the fourteen entrypoints — an inline `PORT=...` prefix assignment is not visible to a `${PORT}` expansion on the same command line. Today those ten resolve to the same value either way, so this is latent rather than live; I am not claiming it was shipping a bug. It is exactly the "env var in an entrypoint is not what it looks like" shape, and it is now a hard failure instead of invisible.

All 12 pre-existing warnings are fixed in this PR: SC2097/SC2098 across the entrypoints (hoisted to a real assignment + `export`), SC1090 in `agentcore/docker/up.sh`, SC2155 in `agentcore/infra-terraform/scripts/deploy-frontend.sh`.

## Red / green

Images built locally with `docker buildx build --builder desktop-linux` from the real starter Dockerfiles, then run.

### RED — strands-python, pre-fix production image

```
$ docker logs strands-red
[entrypoint] Starting: strands-python starter
[entrypoint] OPENAI_API_KEY: set
[entrypoint] Starting agent on port 8123...
INFO:     Will watch for changes in these directories: ['/app/agent']
INFO:     Uvicorn running on http://0.0.0.0:8123 (Press CTRL+C to quit)
INFO:     Started reloader process [13] using StatReload
INFO:     Started server process [15]
```

### RED — crewai-crews, pre-fix production image

```
$ docker run --rm --entrypoint sh crewai-red:latest -c 'grep -n python /app/entrypoint.sh'
6:PORT=8000 python server.py &

$ docker logs crewai-red
▲ Next.js 16.2.7
- Local:         http://localhost:8000
INFO:     Will watch for changes in these directories: ['/app/agent']
ERROR:    [Errno 98] Address already in use
```

### GREEN — same builds, post-fix

```
$ docker logs strands-green
[entrypoint] Starting: strands-python starter
[entrypoint] OPENAI_API_KEY: set
[entrypoint] Starting agent on port 8123...
[entrypoint] Starting Next.js on port 3000...
[entrypoint] Agent=7 Next=14
▲ Next.js 16.3.5
- Local:         http://localhost:3000
✓ Ready in 131ms
INFO:     Started server process [13]
INFO:     Application startup complete.
INFO:     Uvicorn running on http://0.0.0.0:8123 (Press CTRL+C to quit)

$ docker logs crewai-green
▲ Next.js 16.2.7
INFO:     Started server process [6]
INFO:     Application startup complete.

$ docker logs strands-green | grep -ci "StatReload\|Will watch for changes"
0
$ docker logs crewai-green | grep -ci "StatReload\|Will watch for changes"
0
```

No reloader process in either image. strands-python also serves for real end-to-end: agent `http://localhost:8123/health` → **200** from inside the container, frontend `http://localhost:13002/` → **200** from the host.

### RED / GREEN — the shellcheck gate

RED, on `origin/main`, running the exact step body this PR adds (bash 5 container, since macOS bash 3.2 has no `globstar`):

```
Linting 51 starter shell script(s)
...
10 SC2097 (warning)
10 SC2098 (warning)
 1 SC1090 (warning)
 1 SC2155 (warning)
exit=1
```

GREEN, same command on this branch:

```
Linting 51 starter shell script(s)
CI-STEP-EXIT=0
```

`actionlint` on the modified workflow reports only the pre-existing `depot-ubuntu-24.04-4` unknown-runner-label warning on line 57 (the untouched `validate` job); nothing in the added step.

## Notes

- **Unrelated bug found, not fixed here (out of scope):** `crewai-crews/Dockerfile` sets `ENV PORT=8000`, so the frontend's `next start -p "${PORT:-3000}"` resolves to 8000 and collides with the agent — visible as `[Errno 98] address already in use` in both the red and green logs above. The starter's agent cannot bind in its own production image. That is a functional starter bug, separate from this concept.
- **`langgraph-js`** runs a hot-reload dev server in production (see table). Needs a product decision, not a mechanical fix.
- Overlaps #7136, which also edits `strands-python/entrypoint.sh` on a different line (`export AGENT_URL`). Whichever lands second may need a trivial rebase; neither change reverts the other.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved integration startup scripts by consistently handling frontend ports, including a default port of 3000 and reliable environment propagation.
  * Updated selected agent examples to run through Uvicorn, avoiding development file-watching behavior.
  * Improved shell compatibility and quoting across integration entrypoints.

* **Tests**
  * Added automated ShellCheck validation for all integration shell scripts, failing validation for warnings or higher.


jpr5:
Superseded by #7145, which consolidates this PR with the other starter-production fixes into one review. Every commit here is cherry-picked as-authored and the full red/green proof from this body is carried over verbatim. Branch `fix/starter-container-hygiene` is left in place.

PR #7140, opened 2026-09-15
fix(examples): repair the langgraph starters' agent layer

jpr5:
## Problem

Stacked on #7138 (`fix/langgraph-starters-docker-route`) — **that PR must merge first.** #7138 fixed the Docker *route* layer, so all three langgraph starters now build and serve `/api/copilotkit/info` = 200 JSON. But the `agent` and `chat` rungs still fail: every chat run dies. #7138's own scope note called out two follow-on defects; this PR fixes them, plus a third that only became visible once the first was cleared.

### Defect 1 — `langgraph-python` + `langgraph-fastapi`: the agent 500s on every request

`ag-ui-langgraph`'s fastapi endpoint clones the agent on **every** request. Since 0.0.42 the base `clone()` forwards three newer kwargs (`enable_legacy_on_interrupt_event`, `emit_interrupt_outcome`, `emit_raw_events`), and the **published** `copilotkit==0.1.96`'s `LangGraphAGUIAgent.__init__` accepts only `(name, graph, description, config)`.

This repo's own `sdk-python` source already fixes it (a `**kwargs` passthrough, plus `tests/test_agui_agent_clone.py`), but that fix is **not in any released copilotkit** — PyPI's latest is `0.1.96`, which predates it.

So the resolution is on the `ag-ui-langgraph` side. Measured directly against published `copilotkit==0.1.96`:

| ag-ui-langgraph | `clone()` |
|---|---|
| 0.0.42 | OK |
| **0.0.43** | **FAILS** |
| 0.0.44 | OK |
| 0.0.45 | OK |

`0.0.43` — the version both Dockerfiles pinned — is the **sole** broken release; its `clone()` forwards the flags unconditionally, where 0.0.44 forwards them conditionally. Pin bumped to **0.0.44**, which is also the newest version `sdk-python`'s own `clone()` CI matrix exercises.

### Defect 2 — same two starters: Dockerfile pins had drifted from `pyproject.toml`

Fixing defect 1 unmasked this. The Dockerfiles hand-list their Python deps rather than installing `agent/`, and that list had drifted:

| | Dockerfile (before) | `agent/pyproject.toml` + `uv.lock` |
|---|---|---|
| langchain | `1.0.1` | `1.2.15` |
| langgraph | `1.0.5` | `1.1.6` |
| langchain-openai | `>=1.1.0` | `1.1.9` |
| langsmith | `>=0.4.49` | `0.7.33` |
| openai | `>=1.68.2,<2.0.0` | `1.109.1` |

[...]

PR #7139, opened 2026-09-15
fix(examples/llamaindex): honour OPENAI_BASE_URL in the starter agent

jpr5:
## Problem

The `llamaindex` starter (`examples/integrations/llamaindex`) has been red on the `chat` cell in showcase staging since 2026-07-20 (~1345 consecutive failures) with `RUN_ERROR` — an OpenAI **401** — even though `docker-compose.test.yml` sets `OPENAI_BASE_URL=http://aimock:4010/v1`. Its env vars are byte-identical to working starters, so the variable was being *set* correctly but not *honoured* in-process.

## Root cause (proven, not inferred)

LlamaIndex's `OpenAI` LLM does not read `OPENAI_BASE_URL` at all. It resolves its base URL through `resolve_openai_credentials()`:

```python
api_base = get_from_param_or_env("api_base", api_base, "OPENAI_API_BASE", "")
...
final_api_base = api_base or openai.base_url or DEFAULT_OPENAI_API_BASE
```

It reads **`OPENAI_API_BASE`** only, then falls back to `DEFAULT_OPENAI_API_BASE = https://api.openai.com/v1`. The starter constructed `OpenAI(model="gpt-4.1")` with no `api_base`, so every request went to the real OpenAI API.

Observed in the running container with `OPENAI_BASE_URL` set:

```
ENV OPENAI_BASE_URL = http://aimock:4010/v1
ENV OPENAI_API_BASE = None
llm.api_base        = https://api.openai.com/v1
underlying sdk base = https://api.openai.com/v1/
```

**Control:** every agent in `showcase/integrations/llamaindex/src/agents/` already forwards it explicitly (`_openai_kwargs["api_base"] = os.environ["OPENAI_BASE_URL"]`) — 15/15 of them. The starter was the only LlamaIndex entry point missing that idiom. This fix applies the identical pattern.

## Red–green proof

Real containers, real outbound requests, packet-level evidence of the destination. Setup: `aimock` at `172.23.0.2:4010`, agent at `172.23.0.3`; `api.openai.com` resolves to `162.159.140.245` / `172.66.0.243` from inside the agent container. `tcpdump` runs inside the **agent's own network namespace** (`--net container:<agent>`), capturing TCP SYNs and DNS.

### RED (pre-fix image, `OPENAI_BASE_URL=http://aimock:4010/v1`)

`POST /run` with `{"messages":[{"role":"user","content":"Hello"}]}`:

```
[...]

PR #7138, opened 2026-09-15
fix(examples): restore /api/copilotkit subpath routing in langgraph starters

jpr5:
## Problem

`examples/integrations/langgraph-js`, `langgraph-python`, and `langgraph-fastapi` have been red on health + agent + chat in showcase staging since 2026-06-04 (~2440 consecutive failures). Two independent defects, both in the starter `Dockerfile` that `showcase_build.yml`'s `build-starters` job publishes to GHCR (`file: examples/integrations/<slug>/Dockerfile`) and that the Railway starter services run.

### Defect 1 — the Docker route override kills subpath matching (all three)

```
RUN rm -f ./src/app/api/copilotkit/\[\[...slug\]\]/route.ts
COPY docker-route-override.ts ./src/app/api/copilotkit/route.ts
```

This deletes the optional catch-all and installs the override as an **exact** App Router segment. An exact segment matches only `/api/copilotkit` — never `/api/copilotkit/info`, which is what the showcase health probe hits. Result: the Next.js HTML 404 handler answers `/info` while `/` serves 200.

### Defect 2 — `npm install @ag-ui/client` fails EOVERRIDE (js + python)

`d6ad0f8d31` added an `@ag-ui/client` override to the `langgraph-js` and `langgraph-python` `package.json`. npm then rejects adding the same package as a direct dependency with a non-matching spec. **Those two starters cannot build an image at all right now.** `langgraph-fastapi` has no override, so it still builds — and still 404s.

## Fix

Copy the override **over** the catch-all instead of deleting it. This keeps the single-route requirement the original comment guards against (Next.js rejects an exact `route.ts` and a `[[...slug]]/route.ts` side by side) while restoring subpath matching:

```
COPY docker-route-override.ts ./src/app/api/copilotkit/\[\[...slug\]\]/route.ts
```

And for js + python, pin the install to the version the existing `overrides` block already names, so the direct-dependency spec matches the override exactly:

```
RUN npm install --save-exact @ag-ui/client@0.0.59
```

`langgraph-fastapi` has no `@ag-ui/client` override, so its `RUN npm install @ag-ui/client` is left untouched. Diff is 3 Dockerfiles, +11 −12. No app source, agent code, or `package.json` changed.

## Red / green proof

Real `docker build` + real HTTP request against the running container, `--builder desktop-linux`, at `f72d34421` (main).

### RED — defect 2, `langgraph-js` (`langgraph-python` identical)

```
[...]

PR #7143, opened 2026-09-15
ci(starters): gate the production image Railway actually runs, and make @chat able to fail

jpr5:
> **Expect red `smoke-starter-production-image` slots on this PR, and that is the point.** This PR touches the workflow and `showcase/tests/**`, so the gate runs all 12 starters. It will be red for `strands-python` (fixed by #7136), `langgraph-python` / `langgraph-js` / `langgraph-fastapi` (#7138, and fastapi/python also hit the `ag-ui-langgraph` pin bug noted at the bottom) and `llamaindex` (#7139) — the starters that are broken in production **right now** while `smoke-starter` is green on all of them. The reds are the gate working.

## Problem

Three starter fixes landed today (#7136 strands, #7138 langgraph trio, #7139 llamaindex). All three found **green CI on genuinely broken code**, by three different mechanisms. Only the third is about the test; the first two are about the artifact.

**1 & 2 — CI never builds what production runs.**

`showcase_build.yml` → `build-starters` publishes:

```
context: examples/integrations/${{ matrix.starter.slug }}
file:    examples/integrations/${{ matrix.starter.slug }}/Dockerfile
tags:    ghcr.io/copilotkit/starter-<slug>:<sha>
```

That **root** `Dockerfile` — one container, its own `entrypoint.sh` starting the Next.js server and the agent sidecar, its own `docker-route-override.ts` swap — is what the Railway starter services pull.

The only pre-merge gate, `smoke-starter`, runs each starter's `docker-compose.test.yml`, which builds `docker/Dockerfile.app` + `docker/Dockerfile.agent` into a **two**-container stack and injects `AGENT_URL` / `LANGGRAPH_DEPLOYMENT_URL` from compose. **Nothing in CI ever built the root Dockerfile**, and `build-starters` itself only runs on `push: main`, so even the build was post-merge-only.

The masking is structural, not incidental — compose supplies the exact value whose absence was the strands bug, and builds the unmodified `[[...slug]]` route the langgraph Dockerfile deletes.

**3 — the `@chat` assertion cannot tell success from failure.** `sendChatMessage` swallowed its 60s `waitForFunction` timeout in a bare `catch {}`, and CopilotKit renders a `RUN_ERROR` as an assistant message — so `gotResponse && responseText.length > 0` is satisfied by a run that failed outright. llamaindex rode that through 1345 consecutive failed probe runs. The only tell was `@chat` taking 1.1 minutes instead of ~1 second.

## Divergences found (all 14 starter dirs inspected firsthand)

| | Divergence | Consequence |
|---|---|---|
| D1 | Different Dockerfiles: prod `<slug>/Dockerfile` vs CI `<slug>/docker/Dockerfile.app`. Different bases (`python:3.12.10-slim`+node vs `node:22-alpine`), different pins, different build flags (`next build --webpack` vs `next build`). | The whole prod build path is ungated. `langgraph-python`/`langgraph-js` currently cannot build their prod image at all (npm EOVERRIDE) with `smoke-starter` green. |
| D2 | Different topology: prod = 1 container + `entrypoint.sh`; CI = 2 containers on a compose network. | `entrypoint.sh` never runs in CI — the strands bug (#7136). |
| D3 | CI **injects** `AGENT_URL` / `LANGGRAPH_DEPLOYMENT_URL`; prod relies on defaults baked into the override/entrypoint. | CI supplies the value whose absence is the bug. |
| D4 | The route override exists only in the prod Dockerfile. | The langgraph 404 (#7138) is invisible to CI by construction. |
| D5 | `smoke-starter` runs 14 slugs, `build-starters` publishes 12. `crewai-flows` (no root Dockerfile) and `strands-typescript` are tested but never shipped. | Nothing to gate for those two. |
| D6 | `build-starters` is `push: main` only, not a PR check. | A prod-image build break merges freely. |
| D7 | Assertion blindness (above). | Right artifact, wrong verdict. |
| D8 | Latent, not fixed: `langgraph-fastapi`/`strands-python` `entrypoint.sh` both print `"Starting: langgraph-python starter"`, and their compose default is `STARTER=${STARTER:-langgraph-python}`. | Harmless in CI (the workflow sets `STARTER`), misleading locally. |

## Fix

**`examples/integrations/docker-compose.production-image.yml`** — one `$STARTER`-parameterised stack (aimock + the **root Dockerfile** image + the existing `starter-smoke` spec). One file, not 12, so it cannot drift per starter. Nothing is injected that the Railway service does not also get.

**`examples/integrations/production-image-gate.sh`** — the tests-container entrypoint (readiness poll, then playwright). In a file because Compose shell-word-splits a multi-line `command:` string into argv.

**`.github/workflows/test_smoke-starter.yml`** — new `detect-production-starters` + `smoke-starter-production-image` jobs. A first step greps `showcase_build.yml` for `file: examples/integrations/.*/Dockerfile` and fails loudly if that path ever moves, so the divergence cannot silently reopen.

Not build-only: the gate asserts `@health`, `@agent` (incl. `/api/copilotkit/info`), `@chat` (a real AG-UI round-trip against aimock) and `@interaction`. The langgraph bug builds clean and 404s at runtime.

**`showcase/tests/e2e/helpers.ts` + `starter-smoke.spec.ts`** — `sendChatMessage` now captures the `/api/copilotkit` SSE transcript off the wire and returns `agui: { runFinished, runError, sawTextDelta, timedOut, transcriptBytes }`. The timeout is recorded instead of swallowed; the drain is bounded at 15s so a mid-stream hang yields a real assertion failure. `@chat` now requires a terminal `RUN_FINISHED`, no `RUN_ERROR`, no timeout, and — for prose-answering starters; strands-typescript answers with an A2UI render — at least one non-empty `TEXT_MESSAGE_CONTENT` delta. `integration-smoke.spec.ts` reads only `gotResponse`/`responseText` and is unaffected.

## Scope — what is and is NOT gated

- **Gated on PRs:** only starters whose own `examples/integrations/<slug>/**` changed; all 12 when the workflow, compose file, gate script, or `showcase/tests/**` change. Each slot is a full Next.js + Python/.NET image build (~10-15 min), so all-12-per-PR would add hours to every starter PR.
- **NOT gated on PRs:** a starter broken from outside its own directory (`packages/**`, a floating transitive dep, an upstream release). The existing 6h `schedule` and post-release `workflow_run` triggers run **all 12** and cover exactly that.
- **NOT covered at all:** `crewai-flows`, `strands-typescript` — `build-starters` publishes no image for them.
- A reduced or empty matrix is printed with `::notice::`. Coverage is never silently dropped.

## Red / green

All runs: real `docker compose` on `origin/main` d9bec2047, `--builder desktop-linux`, aimock offline.

### A. Artifact divergence — same starter, same commit, two gates

`langgraph-fastapi`, unmodified main:

```
# EXISTING gate (docker/Dockerfile.app two-container stack)
$ STARTER=langgraph-fastapi docker compose -f docker-compose.test.yml up --build \
    --abort-on-container-exit --exit-code-from tests
  ✓  4 @agent langgraph-fastapi — agent endpoint is reachable (94ms)
  ✓  2 @health langgraph-fastapi — health endpoint responds (138ms)
  ✓  3 @interaction langgraph-fastapi — UI interactions work (3.3s)
  ✓  1 @chat langgraph-fastapi — chat round-trip via aimock (1.1m)
  4 passed (1.2m)
EXIT=0                      <-- GREEN, while production 404s

# NEW gate (root Dockerfile, same commit, same spec)
$ STARTER=langgraph-fastapi docker compose -f docker-compose.production-image.yml up --build \
    --abort-on-container-exit --exit-code-from tests
  ✓  2 @health ... (102ms)
  ✘  1 @agent langgraph-fastapi — agent endpoint is reachable (108ms)
  ✘  3 @interaction langgraph-fastapi — UI interactions work (3.0s)
  [console.error] [CopilotKit] Error (runtime_info_fetch_failed): Runtime info request failed with status 404
  2 failed / 2 passed
EXIT=1                      <-- RED on the real bug (red in prod since 2026-06-04)
```

With #7138's Dockerfile fix applied, `@agent` and `@interaction` flip ✓ — the gate tracks the fix, it is not red on everything.

### B. Assertion blindness — llamaindex, the full 2×2

Pre-fix `llamaindex` ignores `OPENAI_BASE_URL` (LlamaIndex reads `OPENAI_API_BASE`), dials api.openai.com with the aimock key and gets `openai.AuthenticationError: 401 Incorrect API key provided: test-key*******mock`. **The production image is built and exercised correctly in every cell below — only the assertion differs.**

```
# NEW gate, OLD assertions, pre-fix llamaindex
  ✓ @health (165ms)  ✓ @agent (352ms)  ✓ @interaction (969ms)
  ✓ @chat llamaindex — chat round-trip via aimock (1.1m)
  4 passed (1.2m)   EXIT=0      <-- GREEN on an agent that 401s

# NEW gate, HARDENED assertions, pre-fix llamaindex
  ✓ @health (72ms)  ✓ @agent (374ms)  ✓ @interaction (1.0s)
  ✘ @chat llamaindex — chat round-trip via aimock (1.1m)
  ✘ @chat llamaindex — chat round-trip via aimock (retry #1) (1.1m)
  1 failed / 3 passed (2.3m)   EXIT=1      <-- RED

# NEW gate, HARDENED assertions, + PR #7139
  ✓ @health (41ms)  ✓ @agent (336ms)  ✓ @interaction (981ms)
  ✓ @chat llamaindex — chat round-trip via aimock (1.3s)
  4 passed (7.8s)   EXIT=0      <-- GREEN
```

| llamaindex, production image | OLD assertions | HARDENED assertions |
|---|---|---|
| pre-fix (main) | 4 passed, exit 0, `@chat` **1.1m** | **1 failed, exit 1** |
| + PR #7139 | — | 4 passed, exit 0, `@chat` **1.3s** |

1.1 min → 1.3 s. The `1.1m` on a *passing* `@chat` is the tell-tale.

The same flip reproduces on `langgraph-fastapi` **with #7138 applied**, where the agent 500s on `LangGraphAGUIAgent.__init__() got an unexpected keyword argument 'enable_legacy_on_interrupt_event'`: old assertions `✓ @chat (1.1m)` → 4 passed / exit 0; hardened assertions `✘ @chat (1.1m)` → exit 1.

### C-live. Both gates, same commit, same CI run

| starter | `smoke-starter` (today's gate) | `smoke-starter-production-image` (this PR) |
|---|---|---|
| langgraph-python | **pass** | **FAIL** |
| langgraph-js | **pass** | **FAIL** |
| adk | pass | pass |
| agno | pass | pass |
| mastra | pass | pass |
| ms-agent-framework-python | pass | pass |
| pydantic-ai | pass | pass |

Five starters agree, two disagree — and the two that disagree are exactly the
ones whose published production image cannot be built. That is the whole thesis
of this PR, rendered as a check list.

### C-pre. The gate failing on a real GitHub runner

`smoke-starter-production-image (langgraph-python)` and `(langgraph-js)` on this PR:

```
#28 0.887 npm error code EOVERRIDE
#28 0.887 npm error Override for @ag-ui/client@* conflicts with direct dependency
Dockerfile:19
failed to solve: process "/bin/sh -c npm install @ag-ui/client" did not complete successfully: exit code: 1
```

The image `build-starters` publishes to GHCR **cannot build at all** (fixed by
#7138), while `smoke-starter (langgraph-python)` is green on the same commit
because it builds `docker/Dockerfile.app`. The divergence, observed in CI rather
than argued.

### C. No false positives

`mastra` production image, unmodified main, both assertion sets:

```
  ✓ @agent (51ms)  ✓ @health (70ms)  ✓ @interaction (807ms)  ✓ @chat (947ms)
  4 passed (7.7s)   EXIT=0
```

### D. Matrix logic

Table-tested in a real bash script (a jq scoping bug — `$changes | index(.)` rebinds `.` to `$changes`, silently matching everything — was caught and fixed this way):

```
pull_request / ["llamaindex"]            -> ["llamaindex"]
pull_request / ["adk","agno"]            -> ["adk","agno"]
pull_request / ["workflow_config","adk"] -> all 12
pull_request / ["crewai-flows"]          -> []        (not a deployed starter)
pull_request / []                        -> []
schedule     / <none>                    -> all 12
```

`actionlint` and `shellcheck` clean; `prettier --check` clean.

## What the hardened `@chat` exposed in the EXISTING gate

Making `@chat` able to fail turned three `smoke-starter` slots red that have always been green. Two are true positives it was hiding:

- **`smoke-starter (langgraph-fastapi)`** — `Agent execution failed: TypeError: fetch failed / code: 'ECONNREFUSED'`. The app never reaches its agent, so that chat has **never** worked. The old assertion passed it in 1.1m, in CI and locally.
- **`smoke-starter (llamaindex)`** — the 401-against-api.openai.com bug (#7139), present in the two-container stack too.
- **`smoke-starter (ms-agent-framework-dotnet)`** — `@chat` 1.1m ×2 with **no agent-side error in the log**, only the 60s DOM wait blowing its deadline.

Because the dotnet case could not be pinned to a real failure, the standalone `timedOut` gate was **demoted to a diagnostic**. Waiting on the `copilot-assistant-message` testid is a rendering detail, not the contract — the helper has always carried a fallback path precisely because some shells render the reply without it. The protocol assertions remain the gate, and a genuine hang still fails because a hung run never reaches `RUN_FINISHED`. All five `agui` values are printed in every failure message.

Verified the demotion does not make the gate decorative — pre-fix llamaindex on the production image still fails, now on the protocol rather than the DOM:

```
Error: AG-UI stream contained a RUN_ERROR
  [agui runFinished=false runError=true sawTextDelta=false timedOut=true transcriptBytes=1726]
1 failed / 3 passed (2.3m)
```

- `langgraph-fastapi` and `langgraph-python` production images still cannot complete a chat even with #7138: the root Dockerfile's `ag-ui-langgraph==0.0.43` + `copilotkit==0.1.96` pins raise `LangGraphAGUIAgent.__init__() got an unexpected keyword argument 'enable_legacy_on_interrupt_event'` → HTTP 500 on every run. **This gate will be red for those two until it is fixed — correctly so.**
- `smoke-starter (langgraph-fastapi)`'s two-container stack cannot reach its agent (`ECONNREFUSED`) — a separate bug from the production-image ones, newly visible now that `@chat` can fail.
- `crewai-flows` has no root Dockerfile and `strands-typescript` is not in `build-starters`, yet both are in the `smoke-starter` matrix.

https://claude.ai/code/session_01ErcmeS9SFqc9ggEz5Rjpwp

## Summary by CodeRabbit

* **Tests**
  * Added automated smoke testing for published starter production images.
  * Production builds are checked for successful startup, configuration, and basic browser functionality.
  * Chat smoke tests now verify protocol completion, streamed text, and absence of reported errors.
  * Test diagnostics capture protocol activity and container logs to simplify failure investigation.

* **Chores**
  * Smoke tests run automatically for relevant showcase changes and release-related workflows.
  * Scheduled and post-release test failures can trigger alerts.

---

> **Merge order: this PR must land LAST.** Its new `smoke-starter-production-image` job is legitimately red until **#7145** (the consolidated starter production fixes — formerly #7136 / #7138 / #7139 / #7140 / #7141 / #7144) is merged. That is the point: this gate is what would have caught those defects. Merge #7145 first, then re-run this PR's checks.


jpr5:
## Added: `sawTextDelta` matched by proximity, not by frame

Follow-on to 40d7de73f. That commit fixed *which event names* count; this one fixes *how* the delta is found.

`sawTextDelta` required `"delta"` within **200 characters** of the text event name. That holds for compact frames but not for the langgraph starters, which embed the full LangChain `rawEvent` blob **between** the two keys — measured at **1170 characters** on langgraph-fastapi. A completely healthy run was therefore called textless:

```
Error: AG-UI stream carried no non-empty TEXT_MESSAGE_CONTENT delta
[agui runFinished=true runError=false sawTextDelta=false timedOut=false
 transcriptBytes=633269]
```

while the same stack's raw SSE plainly carries text:

```
data: {"type":"TEXT_MESSAGE_CONTENT",<1170 chars of rawEvent>,"delta":"Hello! I'm your todo"}
```

**Fix.** No window is the right window. Split the transcript on SSE frames, `JSON.parse` each, and require the non-empty `delta` to sit on the **same frame** as a `TEXT_MESSAGE_CONTENT` / `TEXT_MESSAGE_CHUNK` type. That is strictly **stronger** than the regex it replaces, which matched any two nearby keys across frame boundaries. A regex fallback remains for a transport that is not line-delimited JSON, reachable only when zero frames parse, so it cannot loosen the framed check.

### Table test — 8 cases with negative controls

```
PASS  want=true  got=true   far-apart CONTENT+delta (the 1170-char bug)
PASS  want=true  got=true   CHUNK spelling
PASS  want=false got=false  empty delta (negative)
PASS  want=false got=false  START/END only (negative)
PASS  want=false got=false  tool-call delta, not text (negative)
PASS  want=false got=false  RUN_ERROR only (negative)
PASS  want=false got=false  cross-frame smear must NOT match (stricter than regex)
PASS  want=true  got=true   non-SSE concatenated fallback
```

### Red — real langgraph-fastapi two-container stack, agent reachable, run healthy

```
✘ @chat langgraph-fastapi — chat round-trip via aimock
  AG-UI stream carried no non-empty TEXT_MESSAGE_CONTENT delta
  [runFinished=true runError=false sawTextDelta=false timedOut=false]
1 failed / 3 passed (10.4s)
```

### Green — same stack

```
4 passed (10.1s)
```

**It still fails what should fail.** With the starter's agent URL reverted so the run errors, the same suite returns to `AG-UI stream contained a RUN_ERROR [... runError=true ...]`, 1 failed / 3 passed.

The connection-layer half of that starter's failure is fixed separately on `fix/starter-production-fixes` (#7145); both are needed for langgraph-fastapi to go green.

PR #7144, opened 2026-09-15
fix(showcase/harness): send a UUID threadId so the langgraph-js chat cell can go green

jpr5:
Stacked on **#7140** (`fix/langgraph-starters-agent-layer`), which is itself stacked on **#7138**. Both must merge first.

#7138 + #7140 made the three LangGraph starter *images* correct — all three complete a real AG-UI SSE chat round-trip. But the `langgraph-js` **`chat` cell would still render RED in the dashboard**, because the harness probe's request is wrong, not the image. This PR fixes that, plus a leftover pin #7140 deliberately deferred.

---

## Defect 1 — the harness sends a non-UUID `threadId`

`showcase/harness/src/probes/drivers/starter-smoke.ts` hardcoded:

```ts
const CHAT_RUN_BODY = JSON.stringify({
  threadId: "starter-smoke-thread",
  runId: "starter-smoke-run",
  ...
```

`langgraph-js` fronts a real **LangGraph Platform** server, whose `POST /threads` validates `thread_id` as a UUID and rejects that literal with HTTP 400. The runtime surfaces it to the probe as `RUN_ERROR ... Failed to create thread`, so the chat rung was red on **every** run regardless of image health.

Both ids are now **fresh UUIDs per invocation**.

**Why fresh, not a stable UUID per starter:** LangGraph Platform *persists* thread state. A constant id would accumulate the probe's `"Hello"` turns in one ever-growing thread, and every subsequent run would answer with that history in context — the rung would stop testing a cold single-turn round-trip, and the thread would grow without bound. Per-run debuggability is already served by the per-level `starter:<col>/<level>` rows and their keyed error classes.

**Downstream dependence on the literal — checked, none:**

- `git grep starter-smoke-thread|starter-smoke-run` across the whole tree: **the driver was the only hit** (both literals).
- Probe/row keys come from `key_template: "starter_smoke:${name}"` (`showcase/harness/config/probes/starter_smoke.yml`) and the starter → dashboard-column slug remap — never the request body.
- The aimock fixture is selected by the `X-AIMock-Context` header (the column slug), not by `threadId`.
- Run bookkeeping / dedup keys are keyed off the probe key, not the body.

[...]

PR #7145, opened 2026-09-15
fix(examples): repair the starters that are broken only on the production image path

jpr5:
## What this is

A consolidation of six single-root-cause PRs — **#7136, #7138, #7140, #7139, #7141, #7144** — into one reviewable change, plus **one new fix** (crewai-crews `ENV PORT=8000`) folded in rather than opened as yet another PR. Nothing was dropped: every commit is cherry-picked as-authored, and every red/green proof from the original PR bodies is reproduced verbatim below.

## What was broken

The `examples/integrations/*` starters are the images Railway actually deploys to showcase staging. Several of them were broken **only on that production path** — the single-image build from each starter's root `Dockerfile` + `entrypoint.sh`. CI never noticed, because `docker-compose.test.yml` builds a *different* artifact (a two-container `docker/Dockerfile.agent` + `docker/Dockerfile.app` stack) with different env wiring. So the gate was green while staging was red.

Six independent root causes:

| # | Starter(s) | Root cause | Symptom |
|---|---|---|---|
| #7136 | `strands-python` | `entrypoint.sh` starts the sidecar on 8123; `src/agent.ts` defaults to 8000; nothing exported `AGENT_URL` | `@chat` red, `RUN_ERROR` / ECONNREFUSED; `@health` + `@agent` green because neither dials the sidecar |
| #7138 | `langgraph-js`, `langgraph-python`, `langgraph-fastapi` | The Docker route override was **copied over** `[[...slug]]`, destroying subpath matching; `npm install @ag-ui/client` failed EOVERRIDE | `/api/copilotkit/*` 404s; two of three images failed to build at all |
| #7140 | `langgraph-python`, `langgraph-fastapi`, `langgraph-js` | `ag-ui-langgraph` 0.0.43 500s on every request; Dockerfile langchain/langgraph pins had drifted from `agent/pyproject.toml`; the js override spoke the wrong protocol | agent 500s on every request |
| #7139 | `llamaindex` | LlamaIndex reads `OPENAI_API_BASE`, never `OPENAI_BASE_URL` — the starter only forwarded the latter | agent ignores the configured base URL entirely |
| #7141 | `strands-python`, `crewai-crews` | Production entrypoints invoked dev entry points that enable `reload=True`, starting a StatReload file watcher inside the prod image. Nothing linted starter entrypoints at all. | dev-mode servers in production containers; 12 latent shellcheck defects across 14 scripts |
| #7144 | `langgraph-js` (+ `langgraph-python`, `langgraph-fastapi`) | The harness `starter_smoke` chat driver hard-coded `threadId: "starter-smoke-thread"`; LangGraph Platform rejects a non-UUID with `HTTP 400 ZodError {validation:"uuid", path:["thread_id"]}`. Separately, `agent/pyproject.toml` was still on the broken `ag-ui-langgraph` 0.0.43. | chat rung `red / smoke-failed / "chat stream emitted RUN_ERROR"` — **this is the one that keeps the langgraph-js chat cell red even after every other fix lands** |
| *(new here)* | `crewai-crews` | `Dockerfile` sets `ENV PORT=8000` — the port the agent binds — while `entrypoint.sh` starts Next.js on `${PORT:-3000}` | on any host that does not inject its own `PORT`, Next.js claims 8000 first, uvicorn dies `[Errno 98] Address already in use`, container `Exited (1)` and crash-loops |

## The fix

Per the table above, plus a new `shellcheck` gate over `examples/integrations/**/*.sh` (from #7141) so the entrypoint class of defect cannot recur silently.

**Overlap resolution (the only one between source PRs):** `examples/integrations/strands-python/entrypoint.sh` is touched by both #7136 (+4/−0, adds the `AGENT_URL` export) and #7141 (+10/−3, swaps `python main.py` → direct `uvicorn`, and fixes shellcheck's `PORT` warning). The two hunks are at different lines and compose cleanly; the consolidated file carries **both** (+14/−3). No other file is touched by more than one source PR.

The new crewai-crews commit additionally lands on `crewai-crews/entrypoint.sh` (already modified by #7141) — it only rewrites the stale leading comment, which claimed "Railway sets PORT=3000"; the actual staging value is 8080.

### Why #7144's defect is invisible to current CI, by construction

`smoke-starter` drives the **browser UI**, and the CopilotKit client already mints UUIDs. Only the harness hand-rolls the request body, so only the harness path ever sent the literal. No existing gate could see it.

## Diff equivalence vs. the six source branches

Per-file blob hashes on this branch are **byte-identical** to the corresponding source branch for all 19 non-overlapping files, and the one overlapping file carries both changes:

```
Consolidated cherry-picks only (before the new crewai-crews commit):
20 files changed, 134 insertions(+), 79 deletions(-)

sum of sources:
  #7136                  1 file   +4  -0
  #7138 + #7140 (stack)  5 files  +55 -62
  #7139                  1 file   +10 -1
  #7141                  14 files +65 -16
  ----------------------------------------
  21 entries / 20 distinct files (strands-python/entrypoint.sh counted twice)
  +134 / -79   ← exact match

  #7144                  6 files  +131 -29   (stacked on #7140; blob-identical here)

This branch as pushed, including #7144 and the new crewai-crews commit:
27 files changed, 266 insertions(+), 109 deletions(-)
```

Two apparent mismatches against `origin/fix/harness-uuid-threadid` were run down and are **not** content loss:
- `examples/integrations/langgraph-js/docker-route-override.ts` — deleted on **both** sides (it is removed by #7140, which #7144 is stacked on).
- `scripts/release/lib/channels-registry{,.test}.ts` — that branch predates `origin/main` commits `1752dc2264` / `02db3eb367`; this branch is cut from current `main`, so it simply carries the newer files.

## What this PR does NOT do

- It does **not** add the CI gate that would have caught this class of bug. That is **#7143**, held separately and intentionally merged *last* — its new `smoke-starter-production-image` job is legitimately red until this PR lands.
- The `#7140` body below states the langgraph-js chat cell will stay red pending a harness fix "NOT in this PR". That harness fix **is** in this PR now — it is #7144, folded in below.
- Scope notes and deliberately-unfixed observations from each source PR are preserved verbatim in the sections below.

---
---

# New in this PR — crewai-crews `ENV PORT=8000`

Not from any of the six source PRs. Red/green below is firsthand against two locally built images differing only in the one line this commit deletes, and was independently reproduced against the published artifact.

### RED — pre-fix image, no `PORT` supplied

Built from this branch's tree with the `ENV PORT=8000` line put back, so the only variable is the one line this commit deletes.

```
$ docker run -d --name crewai-red-run -e OPENAI_API_KEY=sk-test crewai-red:latest
$ sleep 30 && docker inspect -f 'status={{.State.Status}} exit={{.State.ExitCode}}' crewai-red-run
status=exited exit=1

$ docker logs crewai-red-run
▲ Next.js 16.2.7
- Local:         http://localhost:8000
- Network:       http://172.17.0.3:8000
✓ Ready in 49ms
INFO:     Started server process [7]
INFO:     Waiting for application startup.
INFO:     Application startup complete.
ERROR:    [Errno 98] error while attempting to bind on address ('0.0.0.0', 8000): address already in use
INFO:     Waiting for application shutdown.
INFO:     Application shutdown complete.
```

### GREEN — post-fix image, identical invocation

```
$ docker run -d --name crewai-green-run -p 13000:3000 -e OPENAI_API_KEY=sk-test crewai-crews-consolidate-check:latest
$ sleep 30 && docker inspect -f 'status={{.State.Status}} exit={{.State.ExitCode}}' crewai-green-run
status=running exit=0

$ docker logs crewai-green-run
▲ Next.js 16.2.7
- Local:         http://localhost:3000
✓ Ready in 53ms
INFO:     Uvicorn running on http://0.0.0.0:8000 (Press CTRL+C to quit)

$ curl -s -o /dev/null -w '%{http_code}' localhost:13000/api/copilotkit/info
200
$ curl -s -o /dev/null -w '%{http_code}' localhost:13000/
200
```

### GREEN control — post-fix image, host DOES supply `PORT` (no regression)

The host-injected case is the one that used to work; it still does.

```
$ docker run -d -p 18080:8080 -e PORT=8080 -e OPENAI_API_KEY=sk-test crewai-crews-consolidate-check:latest
- Local:         http://localhost:8080
INFO:     Uvicorn running on http://0.0.0.0:8000 (Press CTRL+C to quit)
$ curl -s -o /dev/null -w '%{http_code}' localhost:18080/api/copilotkit/info
200
```

### Independently reproduced against the published artifact

Same symptom, found first on `ghcr.io/copilotkit/starter-crewai-crews:latest` (amd64 `sha256:4c27c6d7…`), pulled and run — not a rebuild: no `PORT` → `ERROR: [Errno 98] Address already in use`, `Exited (1)`; `-e PORT=8080` → Next on 8080, `Uvicorn running on 0.0.0.0:8000`, `/api/copilotkit/info` 200.

### Why nothing caught it

Staging and CI both happen to supply a `PORT` — staging `PORT=8080` (confirmed via `railway ssh`), CI compose `PORT=3000`. So the image only crash-loops on a host that does *not* set `PORT`. A latent landmine, not a live outage.

`crewai-crews` is the **only** starter of the twelve carrying this `ENV`. The agent's port is already pinned explicitly in `entrypoint.sh`, so removing the `ENV` is a no-op for the agent and lets Next.js fall back to 3000 — which is what `EXPOSE 3000` already advertises.

---
---

# Carried over verbatim — #7136 · `fix(examples): point strands-python's Next.js runtime at the 8123 sidecar`

## Problem

The `strands-python` starter is red on the `@chat` rung in showcase staging with `RUN_ERROR` / ECONNREFUSED, while `@health` and `@agent` stay green.

`entrypoint.sh` starts the Python sidecar on **8123** (`AGENT_PORT=8123`), but `src/agent.ts` defaults to **8000**:

```ts
url: `${process.env.AGENT_URL || process.env.STRANDS_AGENT_URL || "http://localhost:8000"}/`,
```

Nothing in the image sets `AGENT_URL`, so the Next.js runtime dials a port nothing listens on.

**Why the other rungs stayed green:** neither `/` nor `/api/copilotkit/info` dials the sidecar, so both answer 200 regardless.

**Why CI never caught it:** `docker-compose.test.yml` builds a *separate two-container* stack (`docker/Dockerfile.agent` + `docker/Dockerfile.app`) and passes `AGENT_URL=http://agent:8000` explicitly. The production **single-image** path (root `Dockerfile` + `entrypoint.sh`) — the one staging deploys — is never exercised by the compose smoke test.

## Which side I changed, and why

I changed the **app side** (make it dial 8123), not the sidecar.

The repo has two established conventions, and `strands-python` belongs to the 8123 family:

| starter | sidecar port | how the app is pointed at it |
|---|---|---|
| `claude-sdk-python` | 8123 | `export AGENT_URL="${AGENT_URL:-http://localhost:8123}"` |
| `claude-sdk-typescript` | 8123 | `export AGENT_URL="${AGENT_URL:-http://localhost:8123}"` |
| `strands-typescript` | 8123 | `export AGENT_URL="${AGENT_URL:-http://localhost:8123}"` |
| **`strands-python`** | **8123** | **(missing)** ← the bug |
| `pydantic-ai`, `adk`, `agno`, … | 8000 | app default already 8000 |

`claude-sdk-python/entrypoint.sh` is a near-verbatim template of this file — same header, same structure, same comments — differing only by the three lines this PR restores. `git log` confirms the file was added whole and never had them. So this is a dropped line, not a design choice.

Moving the sidecar to 8000 instead would also work, but it would fight `main.py`'s deliberate `AGENT_PORT` parameterisation and split `strands-python` from its three closest siblings.

**`PORT` collision check:** not applicable here. `main.py` reads `AGENT_PORT`, never `PORT`, so the Railway `PORT=3000` trap that `pydantic-ai`/`adk`/`agno` guard against does not apply. No `PORT` override needed.

## Diff

4 added lines, one file.

```diff
+# Point the Next.js runtime at the agent's port (the agent runs on 8123 in this
+# single-container image; the route defaults to 8000 otherwise).
+export AGENT_URL="${AGENT_URL:-http://localhost:8123}"
+
 # Start the Strands agent via AG-UI protocol.
```

## Red / green proof

Real `docker buildx build` of the production root `Dockerfile`, real container, real AG-UI chat round-trip against aimock (`ghcr.io/copilotkit/aimock:latest` + this starter's own `fixtures/`), driven by:

```
POST /api/copilotkit/agent/default/run   (Accept: text/event-stream)
{"threadId":"probe-thread-1","runId":"probe-run-1",
 "messages":[{"id":"m1","role":"user","content":"Hello"}],
 "tools":[],"context":[],"state":{},"forwardedProps":{}}
```

### Pre-condition (identical in both runs) — reproduces the staging symptom exactly

```
$ docker exec <c> sh -c 'curl -s http://localhost:8123/health; curl -s http://localhost:8000/health'
{"status":"ok"} <- 8123
                <- 8000 exit=7          (connection refused)

$ curl -s http://localhost:3111/api/copilotkit/info
{"version":"1.71.1","agents":{"default":{...}},...}    # @agent rung green
```

### RED (before the fix, image `strands-red:latest`)

```
POST http://localhost:3111/api/copilotkit/agent/default/run
data: {"type":"RUN_ERROR","message":"fetch failed","code":"INCOMPLETE_STREAM"}
```

container log:

```
Agent execution failed: TypeError: fetch failed
    at ignore-listed frames {
  [cause]: Error: connect ECONNREFUSED 127.0.0.1:8000
      at <unknown> (Error: connect ECONNREFUSED 127.0.0.1:8000) {
    errno: -111,
    code: 'ECONNREFUSED',
    syscall: 'connect',
    address: '127.0.0.1',
    port: 8000
  }
}
```

The uvicorn sidecar log shows **no** `POST /` — the request never reached it.

### GREEN (after the fix, image `strands-green:latest`, same probe)

`next-server` now inherits the variable:

```
$ docker exec <c> ... /proc/*/environ
pid 16: next-server (v1 -> AGENT_URL=http://localhost:8123
```

```
data: {"type":"RUN_STARTED","threadId":"probe-thread-1","runId":"probe-run-1",...}
data: {"type":"STATE_SNAPSHOT","snapshot":{}}
data: {"type":"MESSAGES_SNAPSHOT","messages":[{"id":"m1","role":"user","content":"Hello"}]}
data: {"type":"TEXT_MESSAGE_START","messageId":"16827ebb-...","role":"assistant"}
data: {"type":"TEXT_MESSAGE_CONTENT","messageId":"16827ebb-...","delta":"Hello! I'm your todo"}
data: {"type":"TEXT_MESSAGE_CONTENT","messageId":"16827ebb-...","delta":" app assistant. I ca"}
data: {"type":"TEXT_MESSAGE_CONTENT","messageId":"16827ebb-...","delta":"n help you manage yo"}
...  (19 non-empty deltas total)
data: {"type":"TEXT_MESSAGE_CONTENT","messageId":"16827ebb-...","delta":"ke to do today?"}
data: {"type":"TEXT_MESSAGE_END","messageId":"16827ebb-..."}
data: {"type":"MESSAGES_SNAPSHOT","messages":[...,{"role":"assistant","content":"Hello! I'm your todo app assistant. ..."}]}
data: {"type":"STATE_SNAPSHOT","snapshot":{}}
data: {"type":"RUN_FINISHED","threadId":"probe-thread-1","runId":"probe-run-1"}
```

Terminal `RUN_FINISHED`, 19 non-empty text deltas, **no `RUN_ERROR`**.

The sidecar now confirms it served the run:

```
INFO:     127.0.0.1:54970 - "POST / HTTP/1.1" 200 OK
```

### Previously-green rungs undisturbed

```
GET /                       -> 200
GET /api/copilotkit/info    -> 200
```

## Notes — unrelated issues observed, deliberately NOT fixed here

1. `entrypoint.sh` runs the sidecar as `uv run python main.py`, and `main.py`'s `__main__` block calls `uvicorn.run(..., reload=True)`. The container therefore runs a StatReload watcher in production (`INFO: Will watch for changes in these directories: ['/app/agent']`). `claude-sdk-python` explicitly avoids this by invoking `uv run uvicorn main:app` instead, with a comment saying so.
2. `docker-compose.test.yml` only ever exercises the two-container `docker/Dockerfile.{agent,app}` stack with `AGENT_URL` supplied externally, so no test covers the single-image production path this bug lived in. That is the coverage gap that let this ship.
3. Pre-existing shellcheck SC2097/SC2098 warnings on the `next start` line (line 27), identical across the sibling starters.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved integration startup reliability with consistent port and host configuration.
  * Updated LangGraph examples to restore catch-all routing, correct agent communication, and honor configured service URLs.
  * LlamaIndex now respects custom OpenAI-compatible endpoints.
  * CrewAI and Strands agents now use production-ready server startup.
  * Dependency versions are aligned for reliable LangGraph agent requests.
  * Improved conversational responses when shared-state updates cannot be parsed.

* **Tests**
  * Added ShellCheck validation for integration shell scripts.
  * Smoke tests now generate fresh UUIDs for each thread and run.
  * Added regression coverage for request identifier uniqueness.

---
---

# New in this PR — `ms-agent-framework-dotnet` completed an EMPTY run and called it success

Not from any of the six source PRs. Found by the `smoke-starter-production-image` gate in #7143, which was right: the starter finished an AG-UI run carrying **no assistant text at all** — `RUN_STARTED`, then `RUN_FINISHED {outcome: success}`, 584 bytes, zero deltas, no `TEXT_MESSAGE_*` frame of any spelling.

An earlier diagnosis blamed the gate's `sawTextDelta` matcher. That was a real gate bug (fixed in #7143 as `40d7de73f`, plus `af4ab2502`), but it was **not** dotnet's cause. dotnet still failed on a run carrying both.

### Root cause — `agent/SharedStateAgent.cs`

`SharedStateAgent` is a `DelegatingAIAgent` that runs the model **twice** whenever the request carries a `state` object — which the browser always sends, because `page.tsx` seeds `{proverbs: [...]}` through `useAgent`:

1. **Pass 1** forces `ChatResponseFormat.ForJsonSchema<ProverbsStateSnapshot>` and deliberately **withholds every text-only update**, because on the happy path that text *is* the JSON state snapshot and must not be shown to the user.
2. It parses `response.Text` as that snapshot.
3. On `JsonException` the pre-fix code did a bare **`yield break`** — discarding the model's entire answer.
4. **Pass 2**, which produces the user-visible summary, is only reached when step 2 succeeded.

So when the model answers in prose instead of the requested JSON, the run yields nothing and the AG-UI server still terminates `RUN_FINISHED {outcome: success}`. An empty run that reports success is indistinguishable to the user from a dead agent — exactly the class of lie this whole effort exists to remove.

That path is reached on a plain conversational turn, on a refusal, or against any provider that does not honour `response_format: json_schema`. **aimock is the last of those**, so under the gate it is the *only* path taken.

### Isolating probe — pins it to the `state` branch, on the pre-fix image

Straight at the .NET agent on `:8000`, same container, same aimock:

```
with    "state":{...}  -> RUN_STARTED, RUN_FINISHED success, 0 deltas
without "state"        -> RUN_STARTED, TEXT_MESSAGE_START,
                          8 TEXT_MESSAGE_CONTENT deltas,
                          TEXT_MESSAGE_END, RUN_FINISHED success
```

Ruled out with evidence, not assumption:
- **aimock / base-URL wiring** — `curl` from *inside* the app container to `http://aimock:4010/v1/chat/completions` returns the fixture body, 200.
- **the AG-UI .NET adapter not translating model output** — the no-`state` probe streams `TEXT_MESSAGE_*` correctly on that same image.
- **version skew in the AG-UI .NET bindings** — same binding, same image, both probes.
- **an assertion artifact** — this is raw `curl`, no Playwright anywhere.

**Control:** `ms-agent-framework-python`, the green sibling, has no two-pass structured-output wrapper at all — a plain `Agent` with tools plus `state_schema` / `predict_state_config` — so it never suppresses model text. Only the dotnet starter hand-rolls the swallow.

### The fix — in-repo, one file, +20 lines

Buffer the text-only updates pass 1 withholds, and deliver them on the not-a-snapshot path instead of dropping them. Nothing is synthesised: the text emitted is the model's own answer, byte-for-byte the content that was being thrown away.

### RED — production root `Dockerfile`, real container, real AG-UI SSE round-trip against aimock

```
POST /api/copilotkit/agent/default/run   (Accept: text/event-stream)
{"threadId":"11111111-...","runId":"22222222-...",
 "messages":[{"id":"m1","role":"user","content":"Hello"}],
 "tools":[],"context":[],
 "state":{"proverbs":["CopilotKit may be new, but it's the best thing since sliced bread."]},
 "forwardedProps":{}}
```

```
bytes=584

data: {"type":"RUN_STARTED","threadId":"11111111-1111-4111-8111-111111111111","runId":"22222222-2222-4222-8222-222222222222","input":{...}}

data: {"type":"RUN_FINISHED","threadId":"11111111-1111-4111-8111-111111111111","runId":"22222222-2222-4222-8222-222222222222","outcome":{"type":"success"}}

deltas: 0    TEXT_MESSAGE frames: 0
```

And through the real gate — `STARTER=ms-agent-framework-dotnet docker compose -f examples/integrations/docker-compose.production-image.yml up --build --abort-on-container-exit --exit-code-from tests`, unpatched tree:

```
✘  4 e2e/starter-smoke.spec.ts:102:7 › @chat ms-agent-framework-dotnet — chat round-trip via aimock (1.1m)
   Error: AG-UI stream carried no non-empty TEXT_MESSAGE_CONTENT delta
   [agui runFinished=true runError=false sawTextDelta=false timedOut=true transcriptBytes=1316]
   3 passed     EXIT=1
```

### GREEN — same probe, same aimock, post-fix image

```
bytes=5304    (rawEvent elided for readability)

data: {"type":"RUN_STARTED","threadId":"11111111-...","runId":"22222222-..."}
data: {"type":"TEXT_MESSAGE_START","messageId":"chatcmpl-I0yV2-EoxnVG4h0o","role":"assistant","name":"ProverbsAgent"}
data: {"type":"TEXT_MESSAGE_CONTENT","messageId":"chatcmpl-...","delta":"Hello! I'm the Prove"}
data: {"type":"TEXT_MESSAGE_CONTENT","messageId":"chatcmpl-...","delta":"rbs Agent. I can hel"}
data: {"type":"TEXT_MESSAGE_CONTENT","messageId":"chatcmpl-...","delta":"p you manage a colle"}
data: {"type":"TEXT_MESSAGE_CONTENT","messageId":"chatcmpl-...","delta":"ction of proverbs, d"}
data: {"type":"TEXT_MESSAGE_CONTENT","messageId":"chatcmpl-...","delta":"iscuss their meaning"}
data: {"type":"TEXT_MESSAGE_CONTENT","messageId":"chatcmpl-...","delta":"s, and even check th"}
data: {"type":"TEXT_MESSAGE_CONTENT","messageId":"chatcmpl-...","delta":"e weather. What woul"}
data: {"type":"TEXT_MESSAGE_CONTENT","messageId":"chatcmpl-...","delta":"d you like to do?"}
data: {"type":"TEXT_MESSAGE_END","messageId":"chatcmpl-I0yV2-EoxnVG4h0o"}
data: {"type":"RUN_FINISHED","threadId":"11111111-...","runId":"22222222-...","outcome":{"type":"success"}}
```

8 non-empty text deltas, terminal `RUN_FINISHED`, **no `RUN_ERROR`**.

And through the real gate, patched tree, same command:

```
✓  1 @health ms-agent-framework-dotnet — health endpoint responds (47ms)
✓  2 @agent ms-agent-framework-dotnet — agent endpoint is reachable (355ms)
✓  3 @interaction ms-agent-framework-dotnet — UI interactions work (993ms)
✓  4 @chat ms-agent-framework-dotnet — chat round-trip via aimock (1.3s)
   4 passed (7.8s)     EXIT=0
```

`@chat` goes from **1.1m** (the DOM wait blowing its deadline on a reply that never came) to **1.3s**.

### Happy-path control — the fix changes nothing when the snapshot IS valid

Both images re-run against an aimock fixture that returns a valid `{"proverbs":[...]}` snapshot. The streams are **byte-identical, 2696 bytes each**: pass-1 text still withheld, no duplication, only pass 2's text emitted. The fallback only fires where the run was previously empty.

### Notes — observed, deliberately NOT fixed here

1. `SharedStateAgent` accepts *any* JSON value as the state snapshot. A pass-1 answer of `"hello"` (a quoted JSON string) or `42` parses fine, is emitted as a `DataContent` state snapshot, and pass 2 then runs on garbage. A `ValueKind == JsonValueKind.Object` guard would be correct, but it changes behaviour outside this red/green.
2. The `DataContent` state update the happy path emits does not surface as an AG-UI `STATE_SNAPSHOT` frame in either image — pre-existing, unchanged by this commit, and out of scope here.
3. The dotnet starter has no test project, so there is no unit-level seat for a regression test. #7143's production-image gate is the regression test.
4. aimock ignores `response_format: json_schema`. Worth its own issue — but the starter defect is real regardless of provider, since a refusal or a non-compliant completion produces the same silent empty run against real OpenAI.


jpr5:
## Added: langgraph-fastapi two-container chat has never worked

`smoke-starter (langgraph-fastapi)` goes red on `ECONNREFUSED` once `sendChatMessage` stops swallowing its timeout. This is distinct from the 500s already fixed here (the `ag-ui-langgraph` 0.0.43 `clone()` kwarg and the drifted langchain pins) — this is the two-container/compose path never connecting at all.

**Cause.** `docker-compose.test.yml` injected `LANGGRAPH_DEPLOYMENT_URL=http://agent:8123`, copied from langgraph-python, whose `createDefaultAgent()` falls back through **both** `AGENT_URL` and `LANGGRAPH_DEPLOYMENT_URL`. This starter's route builds a `LangGraphHttpAgent` and reads **only** `AGENT_URL`, so the app fell through to its own `http://localhost:8123` default. The agent container's access log is the tell — before the fix it logged only its own healthchecks, zero requests from `app`.

**Fix.** Set `AGENT_URL` in compose, and accept `LANGGRAPH_DEPLOYMENT_URL` in the route as a fallback so it matches its langgraph-python / langgraph-js siblings.

### Red — real two-container stack, aimock as the LLM

```
app-1  | Agent execution failed: TypeError: fetch failed
app-1  |   [cause]: AggregateError:
app-1  |     code: 'ECONNREFUSED'

✘ @chat langgraph-fastapi — chat round-trip via aimock (1.1m)
  Error: AG-UI stream contained a RUN_ERROR [agui runFinished=false runError=true
  sawTextDelta=false timedOut=true transcriptBytes=434]
1 failed / 3 passed (2.3m)
```

### Green — same stack

```
✓ @health langgraph-fastapi — health endpoint responds (45ms)
✓ @agent langgraph-fastapi — agent endpoint is reachable (35ms)
✓ @chat langgraph-fastapi — chat round-trip via aimock (1.5s)
✓ @interaction langgraph-fastapi — UI interactions work (2.9s)
4 passed (9.4s)
```

A direct AG-UI SSE round-trip against the running app confirms real streamed text rather than a rendered error:

```
data: {"type":"RUN_STARTED",...}
data: {"type":"TEXT_MESSAGE_START",...,"role":"assistant"}
data: {"type":"TEXT_MESSAGE_CONTENT",...,"delta":"Hello! I'm your todo"}
data: {"type":"TEXT_MESSAGE_CONTENT",...,"delta":" app assistant. I ca"}
```

**Negative control.** Reverting both changes returns the stack to `RUN_ERROR` at 2.3m, so the fix — not the assertion — is what flips it.

> Note: going fully green also needs `fix/ci-gate-starter-production-image` (#7143), which carries a `sawTextDelta` framing fix. With the connection repaired but that gate unpatched, this starter's healthy run (`runFinished=true runError=false timedOut=false`) still read as "no text emitted".

PR #7156, opened 2026-09-15
fix(types): allow plain FC for chatView slot (#7126)

### LINKED ISSUE #7126: 🐛 Bug: chatView slot is typed incorrectly
### ♻️ Reproduction Steps

1. Create a chat component using the appropriate type like `type CopilotChatViewProps = ComponentProps<typeof CopilotChat.View>`
2. slot it in `chatView={CustomChatView}`
...

### ✅ Expected Behavior

You should be able to slot the appropriately typed component into a slot.

### ❌ Actual Behavior

You get a type error, that forces you to either change what should be a simple render func/component into dynamically having to wrap it in something like `const InstrumentedChatView = Object.assign(CustomChatView, CopilotChat.View)` or to cast it.

### Screenshots

_No response_

### 𝌚 CopilotKit Version

```shell
"@copilotkit/react-core": "1.71.0",
"@copilotkit/react-ui": "1.71.0",
```

### 📄 Logs (Optional)

```shell

```

Zsanz3:
## What does this PR do?

`chatView` was typed as `SlotValue<typeof CopilotChatView>`, which required namespace static members. A normal `FC<CopilotChatViewProps>` failed to assign.

Type-only change: `chatView` is now `SlotValue<React.ComponentType<CopilotChatViewProps>>`.

## Related
Fixes #7126

## Summary by CodeRabbit

* **Improvements**
  * Custom chat views can now be provided using any compatible React component.
  * Components no longer need additional static properties or special wrapping to work as a chat view.
  * Existing built-in chat views remain supported.
* **Tests**
  * Added coverage confirming support for standard React components and compatible render functions, while rejecting incompatible values.


BenTaylorDev (approved):
Approving. I reviewed this against current `main` rather than the merge base, and verified the mechanism locally before CI was approved to run.

**The bug is real, and it bites our own code.** `CopilotChat.tsx` typed the slot as `SlotValue<typeof CopilotChatView>`, which demands the namespace statics. CopilotKit itself pays that tax twice: `CopilotPopup.tsx` does `Object.assign(PopupViewOverride, CopilotChatView) as typeof CopilotChatView`, and `CopilotSidebar.tsx` does `Object.assign(Component, CopilotChatView)`. Those statics are never read — `renderSlot` calls `React.createElement(slot, props)` — so the type demanded something the runtime never uses.

**Mutation check.** I reverted the one-line change and re-ran `tsc --noEmit`. Lines 18-19 of the new `.test-d.ts` then failed. The test is real, not self-fulfilling.

**The widening costs no type safety.** I probed the new type directly. Still rejected: `FC<{ mustHaveThis: number }>`, `FC<{ messages: number }>`, `{ notASlot: true }`. Still accepted: `typeof CopilotChatView`, `string`, `Partial<CopilotChatViewProps>`, plus `FC<{}>` and `ComponentClass<CopilotChatViewProps>`. Prop checking survives intact; only the statics requirement is gone.

**The test file is wired correctly.** The vitest include glob is `*.{test,spec}.{ts,tsx}`, so `.test-d.ts` is excluded — the comment in the file is accurate. `tsconfig` includes `src/**/*`, so `check-types` is the real consumer, and `static_quality.yml` runs it on every PR. That job is green here.

**On the narrowing.** The first commit fixed `SlotValue` itself; the second backed that out and narrowed to `chatView`. That retreat was correct, and I want it on the record so nobody asks for the general fix later. Adding `ComponentType<ComponentProps<C>>` to `SlotValue` puts `C` in a circular inference position and degrades `renderSlot`'s inference — I measured three new errors in `renderSlot.test.tsx` (lines 93, 533, 574). The narrow fix is the right scope.

**Follow-up, not a blocker.** The `input` slot (`typeof CopilotChatInput`) and the `header` slot (`typeof CopilotModalHeader`) still reject a plain FC for the same reason. I confirmed both with a type probe and will file that separately, since #7126 is scoped to `chatView`.

Thanks for the tightly-scoped change and for adding a type-level regression guard rather than just flipping the annotation.

PR #7161, opened 2026-09-15
feat(mcp-apps): consolidate the Vue and Angular hosts onto the shared package

### LINKED ISSUE #6823: Extract a shared MCP Apps host package to remove per-framework duplication
## Summary

The MCP Apps host is currently re-implemented once per frontend framework, each hand-rolling the app↔host protocol:

- React: `packages/react-core/src/v2/components/MCPAppsActivityRenderer.tsx`
- Vue: `packages/vue/src/v2/components/MCPAppsActivityRenderer.ts`
- Angular: `packages/angular/src/mcp-apps/lib/*`

This triplication is the root cause of recurring drift between hosts. #6707 migrated the **React** host onto `@modelcontextprotocol/ext-apps` (`AppBridge` + `PostMessageTransport`), which ties it to the spec at compile time — but Vue and Angular still hand-roll everything, so the same widget now behaves differently depending on the host framework.

## Proposal: a shared `mcp-apps-support` package

Extract a single, framework-agnostic package (working name `@copilotkit/mcp-apps-support`) that owns the whole host protocol layer on top of `@modelcontextprotocol/ext-apps`, and have each framework renderer become a thin adapter that only:

1. creates/attaches the sandbox iframe in its own render model, and
2. wires framework-specific glue (agent access, state updates).

The shared core would own everything that is currently copy-pasted and drifting:

- `AppBridge` + `PostMessageTransport` wiring and the sandbox-proxy relay (`buildSandboxHTML`)
- protocol-version negotiation (single `LATEST_PROTOCOL_VERSION` source)
- host-context seeding (theme/platform, and later display-mode / available-display-modes)
- the CopilotKit behaviors: `tools/call` proxy + request queue, `ui/message` role/followUp extensions via `_meta.copilotkit`, thread-capture semantics (#5819), `ui/open-link` scheme denylist, tool-input/result forwarding

## Why now (concrete drift this removes)

Protocol version is the clearest example. After #6707, React negotiates the **MCP Apps** protocol version `2026-01-26` (via ext-apps `LATEST_PROTOCOL_VERSION`), while Vue and Angular still hardcode `2025-06-18`:

- `packages/vue/src/v2/components/MCPAppsActivityRenderer.ts:9` — `const PROTOCOL_VERSION = "2025-06-18";`
- `packages/angular/src/mcp-apps/lib/mcp-apps-widget.ts:25` — `const PROTOCOL_VERSION = "2025-06-18";`

Note `2025-06-18` is a **base MCP protocol** version, independent from the **MCP Apps** protocol (`2026-01-26`) — so these hosts are not just behind, they are advertising a version from a different spec line. Historically the same duplication produced other drift too (e.g. `size-change` vs the spec's `size-changed`, CSP directive mapping, incomplete capabilities). A shared package fixes this class of bug by construction.

## Scope / sequencing

- React is the pilot (#6707); it already consumes ext-apps directly and can be refactored to consume the shared package once it exists.
- Vue and Angular migrate onto the shared package next, deleting their hand-rolled protocol code.
- Follow-up work (display-mode, update-model-context, theme host-context) then lands once in the shared package and is available to all three frameworks.

Refs: #6707

yannj-fr:
Closes #6823.

## What

Moves Vue and Angular onto the shared MCP Apps host (`@copilotkit/mcp-apps-renderer`), which React already consumed since #6884. All three frontends now speak one protocol implementation, and the hand-rolled JSON-RPC routers, sandbox builders, request queues and `PROTOCOL_VERSION` literals are deleted rather than re-implemented.

Net effect on the adapters:

| package | change |
| --- | --- |
| `packages/vue` | +537 / -712 (the renderer drops from 782 to ~260 lines) |
| `packages/angular` | +542 / -1123 (local queue + router deleted) |
| `packages/react-core` | +165 / -34 |
| `packages/mcp-apps-renderer` | +2502 / -116 (largely tests) |

## The part that was not just moving files

The host was extracted from react-core *before* the other frontends moved onto it, so it carried React's behaviour and none of the protections Vue and Angular enforced locally. Migrating them as-is would have dropped those silently. Those guarantees are now owned by the shared session and queue:

**Queue** (was: React's, with no ownership, no timeout, thread-keyed)
- Requests carry an owner, so teardown drops the work of one widget instead of everything sharing a thread id, and releases the idle wait immediately rather than holding its timer and agent subscription. (Angular had this; Vue cancelled by thread, which was too broad.)
- Keyed by agent AND thread: two agents sharing a thread id no longer block each other. (Angular had this.)
- A busy agent is waited on for at most 30s, configurable, instead of forever. (Vue had this.)
- The thread is captured at enqueue and re-checked before executing. A shared agent object can have its `threadId` mutated in place while work waits, which would otherwise run a thread-1 request against thread-2 and stream into it. This was missing from every host, including Angular, whose queue supported the flag but never enabled it on the proxy paths.

**Session**
- The resource is selected by matching the requested URI rather than taking `contents[0]`, and base64 bodies decode as UTF-8 instead of latin1. The previous `atob` mangled any non-ASCII widget HTML (`<p>Été ☀️</p>` came out as `<p>Ã‰tÃ© â˜€ï¸</p>`). Both were Angular guarantees; React and Vue had the bugs.
- Content is validated before anything is forwarded, so tool input can no longer reach the widget for an exchange whose result turns out to be invalid.
- A content rejection is recoverable: reported through a distinct `onContentError` hook and cleared when valid content resumes, rather than leaving a fatal-looking error on screen while the widget works again.
- Host identity, capabilities, context and the two timeouts are options merged over shared defaults, so a frontend overrides only what it exposes (absorbs Angular's `MCP_APPS_CONFIG`).
- The sandbox handshake timeout terminates the session, cancelling owned work and disconnecting the bridge, instead of only reporting an error while the bridge stays live until the adapter happens to unmount.

**Content schema** aligned with the MCP schema revision this host speaks: the `ContentBlock` union, `content` defaulted to `[]`, `structuredContent` as an object, and loose envelopes so `result._meta` and server-specific fields reach the widget instead of being stripped. A test compares the schema against the installed SDK rather than against another hand-written copy.

## Behaviour changes worth announcing

1. Vue and Angular move from the hardcoded `2025-06-18` to the ext-apps negotiated version; the host returns its own version rather than echoing the widget's.
2. `ui/initialize` is validated against the spec schema by the bridge, so a widget omitting required fields now fails initialize instead of receiving a lenient hand-rolled response.
3. `ui/open-link` follows **one** policy for all three frontends. Angular previously allowed http(s) only; that rule predated the product decision to allow app-defined deep links, so it is dropped. The shared policy keeps the scheme denylist and gains the two guarantees Angular had and React/Vue lacked: embedded credentials refused, and `/path` style relative links resolved.
4. `ui/message` gains the `_meta.copilotkit` extension channel in Vue and Angular (top-level `role`/`followUp` still honored), plus the #5819 thread-capture guard.
5. A result update no longer reloads the widget in Angular: session identity is separated from content, so the iframe survives. The pre-migration Angular widget re-fetched its resource on every result change.
6. Angular's `data-mcp-app-initialized` now follows the `initialized` notification rather than the `ui/initialize` request.

## Examples

Adds the MCP Apps demo the Angular example was missing, mirroring the React demo's scope: same tool list, same cards, same MCP servers. Two wiring differences are forced by Angular's DI and documented in the component: `provideCopilotKit` and the activity renderer both resolve from the root injector, so the page selects the MCP-enabled agent by name (`[agentId]`) instead of re-pointing the runtime URL per route, and `provideMCPApps()` is registered at the root.

## Verification

- `mcp-apps-renderer` 79, `@copilotkit/vue` 1117, `@copilotkit/angular` 354, `@copilotkit/react-core` 1689. `check-types` and builds green for the four packages, plus the Angular demo and demo server.
- Browser end to end against a real MCP server and a real agent, on all three demos: the widget mounts, `ui/initialize` negotiates `2026-01-26`, the tool result renders real data inside the sandboxed iframe, and `size-changed` resizes the host.
- Each behavioural fix has a test that fails without it (verified by reverting the fix and re-running).

## Known gaps

- `MCPAppsProxy.e2e` has no Vue counterpart: negotiation and proxy error codes are proven at the shared-package level and in Angular, not through the Vue adapter. Recorded as `unmapped` in `packages/vue/PARITY.md`.
- The recoverable-content DOM test exists for Vue only; React wires the same hook but has no counterpart yet.
- `ui/open-link` is covered by unit tests but was not exercised in the browser, since no widget in the test server emits one.
- `@copilotkit/vue` lint remains red with 168 pre-existing errors, unchanged by this branch.

## Summary by CodeRabbit

- **New Features**
  - Added an Angular MCP Apps demo with sidebar chat, tool examples, setup guidance, and dedicated navigation.
  - Added shared MCP Apps rendering support for Angular and Vue, with React integration updates.
  - Added activity message tracking, session synchronization, configurable host options, and recoverable content-error handling.

- **Bug Fixes**
  - Improved resource matching, content validation, link safety, timeout handling, cancellation, and thread-change protection.
  - Preserved tool-result metadata and normalized missing content consistently.

- **Documentation**
  - Documented MCP Apps integration status and migration details for Vue.

- **Tests**
  - Expanded coverage for protocol negotiation, rendering, resource handling, session lifecycle, errors, and request processing.


yannj-fr:
Thanks, both code findings were real and are fixed in 01eb19c. I reproduced each one before changing anything, and each fix has a test that fails without it.

**Whole-content validation.** Correct: `pushFromContent` validated only `result`, then forwarded `toolInput` through an unchecked cast. Content arriving through `syncContent` is as untrusted as the store's, so a `toolInput` that is not a record reached the widget whenever the result happened to parse. The whole content now goes through `MCPAppsActivityContentSchema` before anything is forwarded, and a rejection is reported on the recoverable channel rather than being dropped.

**Identity key collision.** Also correct, and worse than it looks: the guard exists precisely to stop the store pushing a replacement widget's data into the old iframe, and the delimiter made it fail at that job.

```
identityKeyOf({ resourceUri: "ui://app::v2", serverHash: "h" })   -> "ui://app::v2::h::"
identityKeyOf({ resourceUri: "ui://app",     serverHash: "v2::h" }) -> "ui://app::v2::h::"
```

A resourceUri is arbitrary text, so no separator is safe. It is now a serialised tuple, which also keeps an absent field distinct from an empty string.

**PARITY.md statuses.** Fixed. `unmapped` and `vue-only` are now defined next to the existing values, and the snapshot sentence no longer claims every mapped row is `matched` or `intentional-divergence`. The three exceptions are named there instead.

On the docstring-coverage warning: the functions it counts are mostly test helpers and small arrow callbacks inside the adapters. The exported surface of the shared package (`bindMcpApp`, its options and hooks, the queue, the policies) does carry docs, including why each guarantee exists and which host it came from. I would rather leave the helpers undocumented than pad them to reach a percentage, but happy to revisit if a maintainer prefers otherwise.

Verification after the fixes: `mcp-apps-renderer` 81, `@copilotkit/react-core` 1691, `@copilotkit/vue` 1117, `@copilotkit/angular` 354, check-types green for the four packages.

BenTaylorDev (changes requested):
Two blockers found while reviewing locally. CI has not run on this PR at all — the only check is CodeRabbit — so neither would have been caught.

## 1. The committed lockfile does not install

`pnpm install --frozen-lockfile`, under the pinned `pnpm@10.33.4`, fails at the root:

```
ERR_PNPM_LOCKFILE_MISSING_DEPENDENCY  Broken lockfile: no entry for
'tsdown@0.20.3(@arethetypeswrong/core@0.18.2)(@emnapi/core@1.11.2)(@emnapi/runtime@1.11.2)(oxc-resolver@11.24.2)(publint@0.3.17)(synckit@0.11.12)(typescript@5.8.2)' in pnpm-lock.yaml

This issue is probably caused by a badly resolved merge conflict.
```

The branch moved most `@emnapi/runtime` peer-hashes from `1.11.2` to `1.11.3` but left nine references behind:

| ref | `1.11.2` | `1.11.3` |
| --- | --- | --- |
| `main` | 63 | 2 |
| this branch | 9 | 57 |

pnpm's own diagnosis matches the history here, which carries several merges from `main`. Every workflow that installs frozen fails, including the react-19 leg of `test_unit.yml`. Regenerating the lockfile locally produced 120 insertions and 257 deletions against the committed file, so this is not a near-miss.

## 2. React and Vue now drop widgets whose result the new schema rejects, with no visible error

`MCPAppsActivityContentSchema` is a faithful port of the Angular contract. I diffed it against `main:packages/angular/src/mcp-apps/lib/mcp-apps-content.ts` and the logic is identical. But React and Vue arrive from a permissive schema (`content: z.array(z.any())`, `structuredContent: z.any()`). Parsing the same values against both:

```
PASS main | FAIL PR   unknown content block type (future/vendor ContentBlock)
PASS main | FAIL PR   structuredContent as a JSON array
PASS main | FAIL PR   image data as unpadded base64
PASS main | FAIL PR   image data as base64url (- and _)
PASS main | FAIL PR   embedded resource with unpadded blob
PASS main | PASS PR   plain text block (control)
```

Both providers register this schema with the activity registry (`CopilotKitProvider.tsx:450`, `CopilotKitProvider.vue:354`). A rejection therefore reaches `use-render-activity-message.tsx:53-59`, which returns `null`. The widget disappears and leaves one `console.warn`. The new `onContentError` channel cannot report it, because the session never binds.

The same tightening reaches the live proxy path. `session.ts:835` now runs `MCPAppsActivityContentSchema.shape.result.parse(...)` and throws, where `main` returned the result uncast (`main` `session.ts:324`). A `tools/call` that returns an unknown block type now comes back to the widget as a JSON-RPC error instead of its result.

If this narrowing is intended for React and Vue, it belongs under "Behaviour changes worth announcing" with a test, and the rejection needs to be visible rather than an empty gap in the transcript.

## What held up

Build, `check-types` and tests are green across all four packages. `@copilotkit/vue` lint is red at exactly 168 errors, as disclosed, and none of the 46 erroring files is touched here. The deleted Angular queue spec loses no coverage: all six cases have counterparts in the new shared suite. I mutation-checked two load-bearing guards — removing the identity guard fails two tests, removing the thread guard fails one — so the new tests are not self-fulfilling.

BenTaylorDev (commented):
Re-reviewed at `173143ab`. CI has finally run on this PR and it is green, which resolves the first blocker and gives the second one a much narrower shape.

## 1. The lockfile now installs. Resolved.

Every job that installs frozen passed, and I confirmed it directly: `pnpm install --frozen-lockfile` under the pinned `pnpm@10.33.4` reports `Lockfile is up to date, resolution step is skipped` and completes. The `@emnapi/runtime` hash split is gone — the branch now carries 63 x `1.11.2` and 2 x `1.11.3`, identical to `main`. The later merges fixed it.

## 2. The schema narrowing is half resolved.

Same six cases, re-run against this head:

| case | at my review | now |
| --- | --- | --- |
| plain text block (control) | PASS | PASS |
| image data, unpadded base64 | FAIL | **PASS** |
| embedded resource, unpadded blob | FAIL | **PASS** |
| image data, base64url (`-`, `_`) | FAIL | FAIL |
| unknown content block type | FAIL | FAIL |
| `structuredContent` as a JSON array | FAIL | FAIL |

I am dropping base64url. `base64Schema` now validates by decoding, which is the check the SDK itself makes, so rejecting a payload the SDK would also reject is the right call and the comment says so.

Two narrowings are still live, both React/Vue only — Angular was already strict on `main`, so nothing regresses there:

- **An unknown content block still deletes the whole widget.** `contentItemSchema` is a `discriminatedUnion` over five literals, and one unmatched block fails the entire `content` array. `ContentBlock` is an open extension point in MCP, and `main`'s `z.array(z.any())` passed these through.
- **`structuredContent` is `z.record(z.string(), z.unknown())`**, so a JSON array is rejected where `main` accepted it.

## 3. The rejection is still invisible — and I owe you a correction on the mechanism.

I wrote last time that "the session never binds" `onContentError`. That was wrong: all three hosts bind it (React `MCPAppsActivityRenderer.tsx:158`, Vue `:186`, Angular `mcp-apps-widget.ts:191`). The problem is where it sits. The callback lives inside the renderer, and the renderer only mounts once the registry schema has already accepted the content — `use-render-activity-message.tsx:53-59` returns `null` and logs one `console.warn` before that. React registers the schema at `CopilotKitProvider.tsx:450`, Vue at `CopilotKitProvider.vue:354`. So a rejection can never reach the channel built to report it, and the user sees an empty gap in the transcript.

Either of these closes it for me: let an unknown block through with a catch-all arm, or keep the narrowing and make it visible — render an error state instead of nothing, list it under the behaviour changes, and cover it with a test.

yannj-fr:
Pushed the F8 lot on top of the consolidation: the MCP Apps failure lifecycle, end to end in the three frontends, plus the core change it rests on. Branch merged with `main` first (no conflicts). The PR body has a new section describing it; this comment is the map of what changed since the last push and how it was checked.

## Commits, in order

1. `feat(core): track activity provenance and run closure`. The one change outside the MCP packages, and not an MCP change: `CopilotKitCore` now knows, for every `activityType`, whether an activity is still being produced, already final, or replayed from history (`getActivityExchangeState`), and notifies when the producing run closes (`onActivityRunSettled`). MCP Apps is the first consumer; A2UI, Open Generative UI and custom activity renderers can use the same state. Why the existing tables could not answer it is in `packages/core/README.md`.
2. `feat(mcp-apps): add the shared widget failure controller`. Two independent cycles (widget, report), bounded failure reasons, pure classification.
3. `feat(mcp-apps): wire the failure lifecycle into the session and React`. Transport-send barrier for the proxy path, `onRemoved`, exchange state from the core.
4. `fix(mcp-apps): decide widget failures before the load and across remounts`. Everything the first review rounds found in the shared package: subscriptions installed before the resource fetch, a bind-time observation so an activity already terminal is decided even when the fetch fails, `acquire(exchangeKey)` so a remount keeps its verdict while a new resource starts a new cycle, report and retirement ownership surviving remounts, `retiring` reserved to the proxy path, teardown mid-send concluded synchronously.
5. `feat(react-core)`, `feat(vue)`, `feat(angular)`: the three adapters wired the same way (exchange state, run-settled subscription, removal recorded per exchange identity, dispatch by renderer mark), each with chat-level or widget-level proofs of the six scenarios.
6. `chore(examples)`: the React demo agent sets `forwardDeveloperMessages: true`, which is what lets the model read the explanation request. The role stays `developer`; `user` was considered and rejected, since it would make the model believe the user is talking about errors.
7. `docs(mcp-apps)`: the lifecycle contract in the shared package README, the core contract in the core README, the integrator-facing behaviour and the `forwardDeveloperMessages` requirement in the docs site, and `packages/vue/PARITY.md` with the per-framework proof matrix.

## Two decisions worth a second look

- `isError: true` on the initial activity is now reported, not only removed. The earlier exemption assumed the model would mention a failure it had seen in its own run; driving the demo showed it saying nothing. A possibly redundant sentence is a smaller cost than a vanished widget with no explanation.
- A teardown while a `tools/call` response is still being sent is treated as a cancellation, like an abort: nothing was delivered, so nothing is explained. Preserving the reservation instead would explain a stale failure, and start a model turn, on the next history reload.

## What is not claimed

- Connect mode (`connectAgent`, one pipeline across server runs) was not exercised; the lifecycle was verified with per-run `runAgent` pipelines.
- Terminal markers live for the page session; tying their lifetime to conversation deletion is the next retention step.
- `MCPAppsProxy.e2e` still has no Vue counterpart file; only its `tools/call` failure case was ported, next to `MCPAppsUiMessage.e2e`.

## Checks on the pushed commit

After merging `main` and a frozen install: build, check-types, test, publint and attw green for core (962 tests), mcp-apps-renderer (145), react-core (1728), vue (1162) and angular (372, 1 skipped); compat-check green where the target exists (core, mcp-apps-renderer, react-core). `pnpm verify:angular-package` green including the packed consumer (installs, SSR-renders, hydrates, zoneless Angular 22). `check:public-api-manifest` and `check:plugin-skills` current. Formatting and lint clean on the changed files (12 pre-existing warnings). The full table, and the one Angular test that flaked under load and passes in isolation, are in the PR body.

BenTaylorDev (changes requested):
Re-reviewed at `eba012c8`. The fork workflows on this head are waiting for approval, so I ran the checks locally. After a frozen install, `build`, `check-types` and `test` pass for all five packages with the Nx cache skipped: core 962, mcp-apps-renderer 145, react-core 1728, vue 1162, angular 372 (1 skipped). Those are the same counts as in the PR body.

## 1. Schema narrowing: resolved, with a correction

In my last two reviews I said that the stricter schema drops results that `main` accepted. My examples were an unknown content block and `structuredContent` sent as an array. That was mostly wrong. Two of the three runtimes already validate every result with their MCP SDK, before any activity exists:

- TypeScript: `@ag-ui/mcp-apps-middleware` 0.1.1 uses `Client.callTool`, which parses against `CallToolResultSchema` by default (`sdk@1.29.0`, `dist/esm/client/index.js:490`).
- Python: `runtime-python/.../mcp_apps.py:126-133` sends the request with `types.CallToolResult`.

I ran both validators on the two cases, and both reject them. My claim that `ContentBlock` is an open extension point was also wrong, because the SDK union is closed.

The .NET runtime is the exception. `McpAppsAgent.cs:220` returns the raw `result`, so a server that is not spec-compliant can still send one of these values there. But the rejection is now visible: the widget is removed, and the user gets an explanation. That was one of the two options I offered last time, so I consider this point closed. Sorry for the extra rounds.

## 2. Blocker: a widget's own tool error now destroys the widget

In the proxy handler, `classifyProxyResult` treats `isError: true` as a reason to retire the widget (`session.ts:1023-1110`). The widget receives the error result first. Then the iframe goes away, the activity leaves `agent.messages`, and a developer message starts a new model turn.

`MCPAppsProxy.e2e.test.tsx:538` shows the effect. A checkout widget calls `pay`, gets back `"Payment declined."` with `isError: true`, and the test asserts that the widget is gone. In that case the user wants to try another card. MCP uses `isError` for ordinary tool failures that the caller is expected to handle, and for a widget-initiated call the caller is the widget. On `main` the widget receives the result and stays usable.

For me to approve, a proxied `isError` result must go to the widget and the widget must stay. If you want to keep retirement on the proxy path, limit it to a result that the host itself cannot parse.

The same question applies with less force to `isError` on the initial activity. An app can render its own error state for a failed tool, and the host now removes the widget before the app can show it.

## 3. Please split F8 into its own PR

The consolidation that this PR started with was close to approval once the lockfile was fixed. F8 adds a separate set of decisions, and each one needs its own review:

- New public core API: `getActivityExchangeState`, `isRunActive` and `isAgentInstanceObserved` on `CopilotKitCore`, the `onActivityRunSettled` subscriber event, and about 250 lines of run tracking in `state-manager.ts`.
- The framework now starts a paid model turn without user action, and it sends a fixed English `developer` prompt to do it.
- A new integrator requirement: `forwardDeveloperMessages: true`, or the user gets no explanation.
- The renderer now edits the conversation itself: it removes the activity with `setMessages` and adds a `developer` message with `addMessage`.

These decisions are product and core-API calls, not MCP Apps plumbing, and I do not want them to hold back the consolidation. Please move the F8 commits (`2e6556af` to `9eeca662`, without the `main` merge in between) to a follow-up PR on top of this branch. Then I can review the consolidation again on its own. When you open the F8 PR, please link where the auto-explain behaviour was agreed. #6823 covers the shared host package only.

Issue #7166, opened 2026-09-15
pnpm.overrides: 13 security floors no longer exclude the versions their advisories cover


sonukapoor:
Thirteen entries in the root `pnpm.overrides` block sit below the versions their advisories now require, so the floor no longer excludes anything vulnerable. Two are pinned exactly at the affected version.

Same block as #6423, different defect. That issue is about overrides contradicting package declarations. This one is about floors going stale over time. #7102 added the comment that states the intent:

> `"//"`: "Next.js overrides keep each supported major on its **security floor**"

Thirteen other entries no longer do that.

### The two clearest cases

```jsonc
"js-yaml": ">=4.1.1",          // 4.1.1 is the affected version; needs >=4.3.2
"socket.io-parser": ">=4.2.6", // 4.2.6 is the affected version; needs >=4.2.7
```

Both came in with `868f6b171` (2026-04-12, "fix: deep security vulnerability sweep — 155 → 3 remaining"). They were correct that day. Advisories published since moved the line and the floors did not follow.

### Full list

| override | current floor | installed | first non-vulnerable | severity |
|---|---|---|---|---|
| `tar` | `>=7.5.11` | 7.5.13 | 7.5.21 | **critical** |
| `js-yaml` | `>=4.1.1` | 4.1.1 | 4.3.2 | high |
| `socket.io-parser` | `>=4.2.6` | 4.2.6 | 4.2.7 | high |
| `hono` | `>=4.11.7` | 4.12.15 | 4.13.5 | high |
| `axios` | `>=1.15.0` | 1.15.2 | 1.18.0 | high |
| `langsmith` | `>=0.5.18` | 0.5.25 | 0.6.0 | high |
| `svgo` | `>=3.3.3` | 4.0.1 | 4.1.0 | high |
| `dompurify` | `>=3.3.2` | 3.4.1 | 3.4.13 | medium |
| `fast-xml-parser` | `>=4.5.2` | 5.5.8 | 5.7.0 | medium |
| `qs` | `>=6.14.2` | 6.15.1 | 6.16.0 | medium |
| `@hono/node-server` | `>=1.19.13` | 2.0.0 | 2.0.10 | medium |
| `body-parser` | `>=1.20.3` | 2.2.2 | 2.3.0 | low |
| `esbuild` | `>=0.25.4` | 0.27.3 | 0.28.1 | low |

Measured on `130e9d92e`, so after #7089 and #7102 landed.

### Effect of refreshing them

[...]

PR #7167, opened 2026-09-15
fix(deps): refresh stale security override floors

sonukapoor:
Refreshes thirteen `pnpm.overrides` entries whose floors had drifted below the versions their advisories now require. Details and the full table are in #7166.

The two that show the problem most plainly:

```jsonc
"js-yaml": ">=4.1.1",          // 4.1.1 is the affected version
"socket.io-parser": ">=4.2.6", // 4.2.6 is the affected version
```

Both arrived with `868f6b171` ("deep security vulnerability sweep") and were correct then. The advisories moved and the floors did not.

Each floor here is bounded to the installed major, so `">=4.3.2 <5"` rather than `">=4.3.2"`. Left unbounded, `js-yaml` resolves to 5.4.2 and the lockfile churn roughly doubles. This deviates slightly from the bare `>=` style used elsewhere in the block, so if that style is deliberate I am glad to drop the bounds and push again.

Measured on `6be20fd1c`:

| | before | after |
|---|---|---|
| findings | 116 | 99 |

17 resolved, none introduced. I am quoting totals rather than a per-severity split because severity assignment shifts a little between runs when advisory detail lookups vary; the resolved set is stable. Four of those (`follow-redirects`, `@ai-sdk/provider-utils`, `baseline-browser-mapping`, `fast-xml-builder`) resolve through the cascade rather than being targeted.

The lockfile diff is large because pnpm encodes peer resolutions into composite keys, so changing one resolved version rewrites every key that names it. For comparison, re-resolving the lockfile on `main` with no manifest change at all still moves 50 lines. The reviewable change is the thirteen lines in `package.json`.

No package versions or changelogs are touched, and there is no changeset file.

Refs #7166

*Found with [cve-lite-cli](https://github.com/OWASP/cve-lite-cli), an OWASP project I maintain. I can post the full before and after scans if that is useful.*

## Summary by CodeRabbit

* **Security**
  * Updated minimum versions for several transitive dependencies to incorporate security and stability fixes.
  * Added upper-version limits to help maintain compatibility with supported major releases.


BenTaylorDev:
Verified the refresh against the live advisory API and `pnpm-lock.yaml` on main. Twelve of the thirteen land on or above the first non-vulnerable version for their major. One does not:

```jsonc
"fast-xml-parser": ">=5.7.0 <6",
```

`>=5.7.0` still admits 5.9.3 through 5.10.0, which carry GHSA-8r6m-32jq-jx6q (high, first patched 5.10.1, published 2026-07-21, not withdrawn). 5.7.0 is the right target for the installed 5.5.8 — that is the moderate `< 5.7.0` advisory — but as a *floor* it leaves a high-severity window open above it.

Suggest `"fast-xml-parser": ">=5.10.1 <6"`.

Two things I checked that came back in your favour, noting them so nobody re-litigates them:

- `esbuild: >=0.28.1 <0.29` is right, and `low` in the issue table is right. The high range that also covers 0.27.3 (GHSA-gv7w-rqvm-qjhr) is withdrawn.
- For axios the live justification for 1.18.0 is GHSA-gcfj-64vw-6mp9. GHSA-68jp-44vc-2x5h from the same window is a withdrawn duplicate, so the target is unaffected either way.

On your "separately" note: confirmed, `@isaacs/brace-expansion` and `jsondiffpatch` are absent from the resolved tree, and only `@types/markdown-it@14.1.2` is installed. Agree with leaving them out of this PR.

sonukapoor:
You're right, fixed in 2f98cc7, floor is now `>=5.10.1 <6`.

One precision worth noting: the lockfile resolves 5.11.1, so nothing vulnerable is installed today. The defect is that the override didn't guarantee that, and a regeneration could legitimately have landed on 5.9.3 or 5.10.0. Still wrong, I just want to be clear about which kind of wrong.

I re-derived all thirteen after your comment, enumerating every published version each range admits and intersecting against the npm advisory set rather than checking the installed version. Yours was the only hit. langsmith looked like a second one until I noticed the `<0.8.0` range on GHSA-3644 is the pip entry; npm is `<0.6.0`, patched at exactly 0.6.0, so that floor holds.

Thanks for the esbuild and axios checks, that saves them coming up again.

Regenerating the lockfile also pulled in three unrelated registry drifts (form-data patch bump, an unreferenced undici entry dropped, one eslint peer key simplified). Happy to strip them if you'd rather the diff stayed minimal.

Worth saying plainly: the tool I used to find these stale floors doesn't catch this class. It answers "what fixes the version you have installed", which for 5.5.8 is 5.7.0 and is right, but a floor is a range and nothing was validating the window above it. [Added a rule for it](https://github.com/OWASP/cve-lite-cli/issues/1143).

BenTaylorDev:
On the `@hono/node-server` line specifically: the floor bump is right — [GHSA-9mqv-5hh9-4cgg](https://github.com/advisories/GHSA-9mqv-5hh9-4cgg) covers `>=2.0.0 <=2.0.9` and the lockfile resolves 2.0.0 today. That one row is held up on something outside this PR, though. `packages/runtime` declares `"@hono/node-server": "^1.13.5"` in `dependencies`, so the existing bare override already makes CI test a major that the published manifest forbids, and `>=2.0.10 <3` widens that gap. #7276 is moving `hono` and `@hono/node-server` to optional peers, which is where the declaration gets fixed. Details in [this comment](https://github.com/CopilotKit/CopilotKit/issues/7276#issuecomment-5750381782).

The other twelve entries I verified independently against the GitHub advisory database and `main`'s lockfile — installed versions and first-patched versions all match your table. Note that CI has not run here: a fork PR needs workflow approval, so the only check reporting is CodeRabbit.

sonukapoor:
Thanks for verifying those twelve independently, that's more than I'd expect from a reviewer.

Agreed on `@hono/node-server`, and your framing is right that the gap already exists. The bare `>=1.19.13` override was already making CI test against something `packages/runtime`'s `^1.13.5` forbids; `>=2.0.10 <3` just makes it explicit. So that row belongs with #7276, not here.

**I'll drop the `@hono/node-server` row from this PR** so the twelve you verified can land on their own, and pick it up again once the declaration moves to an optional peer. No reason for one blocked entry to hold the rest.

`hono` itself I'd keep: `>=4.13.5 <5` stays inside the declared major, so it doesn't have the same problem. Shout if you disagree.

Two housekeeping things. The branch has gone conflicting since I opened it, so I'll rebase. And CI hasn't run because fork PRs need workflow approval - if you can approve it, CodeRabbit is currently the only check reporting.

sonukapoor:
Rebased on main and dropped the `@hono/node-server` row, so this is back to the twelve entries you verified. One commit now.

While doing it I checked what dropping that row actually changes, and the answer is nothing, which I do not think is what you were expecting.

The existing override is `"@hono/node-server": ">=1.19.13"` with no upper bound, so it already admits 2.x, and `pnpm-lock.yaml` on main resolves `@hono/node-server@2.0.0` today. My row only made that explicit by writing `>=2.0.10 <3`. It did not cause it, and removing it does not undo it. `packages/runtime` is still on `^1.13.5`, ten manifests in the repo declare the package and every one of them is on the 1.x line, and CI is still testing 2.0.0.

Two advisories apply to 2.0.0 specifically:

- GHSA-9mqv-5hh9-4cgg, moderate, `>=2.0.0 <2.0.10`
- GHSA-frvp-7c67-39w9, moderate, `>=2.0.0 <2.0.5`

The current floor of 1.19.13 is not clean either, since GHSA-frvp also covers `<1.19.15`.

If the intent is that the monorepo stays on the 1.x line the manifests declare, the one-line version is:

```json
"@hono/node-server": ">=1.19.15 <2"
```

1.19.15 comes back clean from the advisory database, it sits inside `packages/runtime`'s `^1.13.5` and inside every other declaration in the repo, and it takes the tree off 2.0.0 without a major bump for anyone.

I have deliberately kept that out of this PR. You asked for the row gone and I would rather not hand you back a modified version of the thing you asked me to remove. Happy to open it separately, or just leave it with you.

BenTaylorDev:
Thanks for dropping the row, and your follow-up is right: main resolves `@hono/node-server@2.0.0` today, so removing it changed nothing. I checked the rebased commit against current main. It merges clean, `pnpm install --frozen-lockfile` passes on the merged tree, and every resolved version lands on or above its new floor.

Two things, and the first one is my mistake from Sep 15.

**esbuild.** I said `>=0.28.1 <0.29` was right. I checked the advisory, not the packages that depend on esbuild. The override forces 0.28 onto `tsx@4.21.0` (`~0.27.0`), `vite@7.3.2` (`^0.27.0`), and `wrangler@4.80.0` and `@angular/build@21.2.14`, which both pin `0.27.3` exactly. The only live advisory it fixes is GHSA-g7r4-m6w7-qqqr: low, the Windows dev server only, and no patched 0.27 exists. 0.28.0's breaking change is small (install-path integrity checks), but forcing a 0.x minor past four declared ranges for that advisory is not a good trade. Please restore the existing `"esbuild": ">=0.25.4"` row.

**`@hono/node-server`.** #7276 closed on Sep 21 via #7278, which moved only express. Nothing is tracking the hono half now, so your `">=1.19.15 <2"` suggestion is the right fix after all. Every declaration in the tree is on 1.x, including the transitive ones (`@langchain/langgraph-api` `^1.12.0`, `@modelcontextprotocol/sdk` `^1.19.9`), and it clears both GHSA-9mqv-5hh9-4cgg and GHSA-frvp-7c67-39w9. Please add it back here in that form.

The three registry drifts are fine to keep. I approved the workflow runs on this commit so CI reports now; the next push will need another approval, and I will do that.

sonukapoor:
Both done, and thanks for catching the esbuild one.

**esbuild** is back to `>=0.25.4`. Your read was right and if anything understated: I checked every declaration in the tree rather than the four, and **137 of 145 reject 0.28.1**, including the ones you named at `~0.27.0`, `^0.27.0` and an exact `0.27.3`. Forcing a 0.x minor past that many declared ranges for one low dev-server advisory with no patched 0.27 is not a trade worth making.

One thing worth saying plainly, because it isn't free: restoring the floor puts esbuild back on **0.27.3**, which is inside `>=0.27.3 <0.28.1`, so GHSA-g7r4-m6w7-qqqr is live again on this branch. I think that's the right call given the alternative, but it's a knowing acceptance rather than a clean restore, and I'd rather say so now than have it turn up in a later pass.

**@hono/node-server** is back as `>=1.19.15 <2`. Confirmed it clears both: `frvp` covers `<1.19.15` and `>=2.0.0 <2.0.5`, `9mqv` covers `>=2.0.0 <=2.0.9`, and the ceiling excludes both 2.x windows. It resolves to 1.19.17.

One implementation note. I regenerated the lockfile from `main` rather than carrying the branch's forward. Restoring the row alone left esbuild pinned at 0.28.2, because pnpm keeps a resolution that still satisfies a widened range, so the row would have looked restored while the behaviour wasn't. Starting from `main` puts it back on 0.27.3 as intended.

Every overridden package resolves within its declared range. Twelve rows changed, the three registry drifts untouched.

PR #7187, opened 2026-09-16
docs(shell-docs): delete 18 inert shadowed pages and ratchet the rest

BenTaylorDev:
## What

Under `docs_mode: generated` the root MDX wins for the page **body** — `src/app/[framework]/[[...slug]]/page.tsx:829` and the matching `candidateOrder` in `src/app/llms-mdx/[[...slug]]/route.ts:365`. A per-framework file at a slug the root also defines contributes no body to either surface.

It is **not** fully inert. `frameworkMetadata` (`page.tsx:255`) loads the framework-scoped doc first with no `docsMode` branch:

```js
const frameworkScopedDoc = loadDoc(`integrations/${docsFolder}/${slugPath}`);
const doc = frameworkScopedDoc ?? loadDoc(slugPath);
```

so such a file still supplies the page's `<title>`, meta description and OG text.

This PR therefore deletes **only the 18 files whose frontmatter is identical to their root twin in every field**. No URL changes what it reports to search engines.

The other **31 differ in at least one frontmatter field** and are recorded as a baseline in the test rather than deleted.

## Why a ratchet instead of a clean assertion

`expect(shadowed).toEqual(BASELINE)` — so **adding** a new shadowed file fails, and **deleting** a baseline entry also fails. The list cannot silently drift out of date in either direction.

## Testing

**Live check that found the problem.** `/langgraph-python/build-with-agents` (43,472 views / 90d):

```
$ curl -sL https://docs.copilotkit.ai/langgraph-python/build-with-agents | grep -oE '<meta name="description"[^>]*'
<meta name="description" content="Use our MCP server to connect your LangGraph agents to CopilotKit."
```

That string is in `integrations/langgraph/build-with-agents.mdx`, not the root page. Hence the narrowing.

**Frontmatter comparison across all 49 candidates:** 18 identical in every field, 31 differ.

**Every deleted slug still has a live root twin**, checked against current `origin/main`:

```
root twin missing: 0   global-path excluded: 0
PASS: every deleted slug still has a live root page
```

**Sitemap is unaffected.** `sitemap.ts:107` emits the full bare-slug x integration cross-product and adds each URL to `seenFrameworkUrls`; the override loop at line 127 skips URLs already seen. All 18 slugs are bare root slugs and none is a `channels/*` global path, so their `/<framework>/<slug>` URLs were already emitted by the cross-product.

**Ratchet mutation-checked in both directions:**

```
--- 1. baseline state ---
baseline=31 actual=31 -> PASS
--- 2. MUTATION: add a new shadowed file (should FAIL) ---
baseline=31 actual=32 -> FAIL   unexpected: [ 'adk/cli' ]
--- 3. MUTATION: remove a baseline file (should FAIL) ---
baseline=31 actual=30 -> FAIL   missing: [ 'langgraph/auth' ]
--- 4. restored ---
baseline=31 actual=31 -> PASS
```

**Prettier:** `All matched files use Prettier code style!`

**Not run locally:** the shell-docs vitest suite. `showcase/shell-docs` is not a pnpm workspace member and needs its per-entry `node_modules` setup first. The checks above replicate the test's logic in plain Node; CI runs the real thing.

1. **`frameworkMetadata` is missing the `docsMode` branch** that the body resolver has, despite a comment claiming it mirrors it. Generated-mode pages currently advertise metadata describing content the site does not render.
2. Deciding what those 31 URLs should say — keep the framework-specific descriptions by getting that content into the rendered page, or let them fall back to the root text.
3. #7106 edits `integrations/langgraph/frontend-tools.mdx`, which is in the baseline (not deleted here), so that PR is unaffected for now.

## Summary by CodeRabbit

- **Documentation**
  - Removed several integration documentation pages covering threads, Inspector, telemetry, self-hosting, and contribution workflows across supported integrations.
  - Related guidance for package linking, documentation contributions, and code contributions is no longer available in the documentation.

- **Tests**
  - Added regression coverage to verify integration documentation page mappings remain consistent and detect unexpected additions or removals.


BenTaylorDev:
## Holding this — the premise is partly wrong

The PR body claims these files are never served. That is true of the page **body** but **not of the page metadata**, and I verified it against the live site rather than only by reading the router.

`/langgraph-python/build-with-agents` (43,472 views / 90d) serves:

```
<title>Build with agents</title>
<meta name="description" content="Use our MCP server to connect your LangGraph agents to CopilotKit."
```

That description comes from `integrations/langgraph/build-with-agents.mdx`, not from the root page.

**Cause.** `frameworkMetadata` resolves the framework file first, unconditionally:

```js
// src/app/[framework]/[[...slug]]/page.tsx:255-257
const frameworkScopedDoc = loadDoc(`integrations/${docsFolder}/${slugPath}`);
const doc = frameworkScopedDoc ?? loadDoc(slugPath);
```

It does not branch on `docsMode`, although the body resolver at line 829 does — despite the comment above it claiming it mirrors that order. So on generated-mode frameworks the body comes from the root file while the title, meta description and OG image text come from the framework file.

**Impact.** Of the 49 files here, **26 carry a title or description that differs from the root page**, on live URLs with real traffic. Deleting them changes what those URLs report to search engines. The other 23 are byte-identical in both fields and can go with no visible change.

This also means the underlying defect is larger than this PR: those pages currently advertise metadata describing content the site does not render.

Next step is a decision, not a rebase — see the discussion. Not merging until then.

PR #7194, opened 2026-09-16
fix(web-inspector): add Intelligence access and preserve Learning shortcuts

marthakelly:
The Inspector Learning results view now has an **Open Intelligence** button in its header. The button opens the Runtime-provided app origin, including custom deployments, without requiring new Threads or pending Skill reviews.

Existing Learning shortcuts keep their scoped destinations. Skill review opens the selected container's Skills page. Analysis and Learning links continue to use the URLs from `snapshot.links`. Links retain new-tab behavior, security attributes, and click telemetry.

The workbench fixtures now include the organization and project paths. [Intelligence PR #1289](https://github.com/CopilotKit/Intelligence/pull/1289) corrects the outdated Learning and analysis URLs at their source. The Skills-review URL stays unchanged.

Sam's reported HTTP 409 remains undiagnosed. This PR does not claim to fix it. We requested more diagnostic information in [the Slack thread](https://copilotkit.slack.com/archives/C0BR0V2P4QJ/p1789666318592249?thread_ts=1789510835.789859&cid=C0BR0V2P4QJ).

Refs https://linear.app/copilotkit/issue/PE-54

Marketing backlog: https://linear.app/copilotkit/issue/EGG-313

## Validation

- All 711 Inspector unit tests and eight Learning workbench browser tests pass.
- Link checks cover Skills review, analysis, empty results, container selection, custom origins, new tabs, and telemetry.
- Package build, TypeScript checks, and commit hooks pass. Hooks cover tests, package lint, and export/type checks for seven affected packages.
- Updated both Inspector documentation pages to describe the general button and the scoped shortcuts.

QA uses the standalone Inspector workbench and local fixtures. It does not cover a live Intelligence sign-in session. One browser run overlapped a CSS rebuild and lost its scenario navigation. All eight tests passed after the rebuild finished.

## UI screenshots

These screenshots show the current standalone Inspector workbench with local fixtures.

### Results and Skill review

The general header button and the scoped Skill-review shortcut remain separate.

Results and Skill review

### Threads ready for analysis

The analysis action opens the selected container's Analysis results page.

Threads ready for analysis

### Empty results

The quiet link opens the selected Learning container.

Empty results


BenTaylorDev:
Thanks for the detailed write-up and the screenshots. The branch is clean and CI is green. Before we land it I want to check the premise, because I read the Intelligence routes differently.

**Question first:** which link did Sam click, and do we have the network trace for the 409? "Open in web app" is the label on the `runs` and `learning` links. The candidates link reads "N Skills for review in web app". That distinction drives everything below.

### The candidates deep link works today

Intelligence registers three Learning route shapes under `/o/:org/:project/*` (`apps/app-frontend/react-shell/src/app.tsx:2558-2589`):

```
/learning
/learning/:containerId
/learning/:containerId/:tab
```

The comment there reads: "The Container and its selected view are path segments, so a shared Learning URL reopens exactly what the sender was looking at." `skills` is a valid tab (`learning-routes.ts:8-12`).

The snapshot's `candidates` link is `/o/<org>/<project>/learning/<containerId>/skills` (`apps/app-api/src/learning/inspector-learning-store.ts:75-78`). That matches the third shape exactly and opens the right container on the right tab. Routing it to the bare origin is a regression: the reviewer loses the container and the tab and has to find both by hand.

### The learning and runs links are not unrouted, but their query string is dead

Those two links carry `?container=<id>` and `&tab=runs` (`inspector-learning-store.ts:69-77`). `LearningWorkspaceRoute` reads path parameters only:

```ts
const containerId = params['containerId'] ?? null;
const requestedTab = params['tab'];
```
(`app.tsx:2203-2205`)

I searched `apps/app-frontend/react-shell/src/learning` for `useSearchParams`, `location.search`, and `URLSearchParams`. The only hit is the demo harness. So both links open Learning, then fall back to the first container (`learning-page.tsx:176`) and the default `insights` tab.

That is a real bug, and I think it is the one worth fixing. It lives in Intelligence's `projectLinks`, which builds a query string where the app wants `learningContainerRoute` path segments.

### The 409 is still unexplained

A resolved route does not return 409. `apps/app-api/src/errors/registry.ts` defines 19 codes at that status. The two auth-shaped ones are `MANAGED_AUTH_ACTIVE_ORG_MISMATCH` (line 461) and `MANAGED_ORGANIZATION_PLAN_SELECTION_REQUIRED` (line 478). I could not attribute the failure from source, so I am not asserting either. Sam's network trace would settle it. If it turns out to be an active-organization or plan handshake, the fix belongs in the Intelligence bootstrap, and this change hides it rather than resolving it.

### Where I think the diagnosis came from

`packages/web-inspector/dev/learning-state-fixtures.ts:350-354` builds:

```ts
`${WEB_APP_ORIGIN}/learning?container=checkout-assistant-default`
```

There is no `/o/<org>/<project>` prefix, and no top-level `/learning` route exists, so that URL really is broken. Production never emits it. Line 357 of the same file already uses the correct production shape for `candidates`. The PR body says QA ran against the standalone workbench with these fixtures, which fits.

The comment at `learning-view.ts:969`, "Intelligence does not support the snapshot's deep routes yet", does not hold as written.

### Suggested split

1. Fix `learning-state-fixtures.ts` to emit production link shapes.
2. Change `projectLinks` in Intelligence to build path segments for `learning` and `runs`.
3. Keep the candidates deep link.
4. Track the 409 separately once we have the trace.

If the trace shows the handshake fails on every deep route, then the origin-entry workaround is justified. In that case, could the comment record the auth reason rather than route support, so the next reader does not conclude the routes are missing?

### Smaller notes

- `snapshot.links` is no longer rendered anywhere, but it stays fatal. `packages/web-inspector/src/index.ts:8033` still throws "Learning snapshot is missing a required web-app link", and `packages/shared/src/utils/inspector-learning.ts:377` still drops the whole snapshot. This is latent today, because unconfigured snapshots carry zero counts, so it is a coherence point rather than a defect.
- The new case at `learning-view.spec.ts:322` builds `${origin}/learning?container=container-1`, which copies the incorrect fixture shape.
- In the `results` state the header link, the results CTA, and the review link now resolve to the same URL. Three identical destinations on one screen.
- The docs edits are accurate for the change, but they do document the manual container search as the workflow.

marthakelly:
@BenTaylorDev Updated based on your review:

- Preserved the scoped Skills-review shortcut and the other snapshot links.
- Kept the new general **Open Intelligence** button in the results header.
- Corrected the workbench URLs, regression tests, and docs. Removed the incorrect statement that Intelligence does not support these routes.
- Opened [Intelligence PR #1289](https://github.com/CopilotKit/Intelligence/pull/1289) to replace the outdated Learning and analysis query parameters with supported paths.

We asked Sam whether the 409 happened again and requested more diagnostic information in [the Slack thread](https://copilotkit.slack.com/archives/C0BR0V2P4QJ/p1789666318592249?thread_ts=1789510835.789859&cid=C0BR0V2P4QJ). The cause remains unconfirmed, and neither PR claims to fix it.

All 711 Inspector unit tests and eight Learning browser tests pass. The companion change passes 12 focused API tests and two existing frontend routing tests. Fresh screenshots are in the PR description. CI is running on the new commits.

BenTaylorDev (approved):
Approving. Round two addresses the round-one regression, and I verified the restoration rather than taking it on faith.

## What I verified

- **Branch is current with main.** The merge-base is the live tip, so the forge diff is the real diff.
- **The scoped shortcuts are genuinely back.** The net change to `learning-view.ts` is the nine added lines. All four existing `externalLink` call sites are byte-identical to `main`.
- **`snapshot!.webAppOrigin` is safe.** `parseInspectorLearningSnapshotV1` rejects any snapshot whose origin is not a URL with pathname `/` (`packages/shared/src/utils/inspector-learning.ts:356-361`), and `deriveLearningViewState` reaches `results` only with a non-null snapshot (`learning-view.ts:19-51`).
- **The bare origin lands somewhere useful.** Intelligence routes `/` to the org home, falls back to `/select-org`, and redirects a signed-out user to `/login?redirectTo=` (`apps/app-frontend/react-shell/src/app.tsx:1866-1901`).
- **The new fixture URL shape matches the real app**, checked against Intelligence `main` and not against the PR description: `learning-routes.spec.ts:29` expects `/o/acme/support/learning/support-learning/analysis-results`.
- **Tests pass locally** at `264b7c80`: `learning-view.spec.ts` 19/19, `learning-state-lab-server.spec.ts` 3/3.
- **The new tests are not self-fulfilling.** I broke the gate so the button never renders; exactly the two new assertions failed. The `querySelectorAll("a")` length check also pins each state's full anchor set, so a stray link cannot slip in.

I could not run the full unit suite or the Playwright workbench locally. My installed dependencies are behind `main`, which is a local harness limit, not a problem with this branch. CI covers both.

## Non-blocking follow-ups

1. **One pane, two product names.** The pane shows "Open in web app" four times (`learning-view.ts:1171,1471,1532,1588`) and "Open Intelligence ↗" once. `inspector.mdx:206` still says "the web app" two paragraphs below the new "Open Intelligence" sentence. Please pick one name.
2. **The docs no longer name the controls.** The text says "The analysis and Skill-review links", but no label in the pane matches that wording, so a reader cannot map the sentence to a button.
3. **Telemetry cannot separate the new button.** It emits `learning-web-link` with the default category `learning`, the same value the quiet link and the selection-required link emit. Measuring the new button needs a new member on the `category` union at `learning-view.ts:968`.
4. **Results-only placement.** The `error` state offers only Retry, which is arguably where a general escape hatch helps most. Intentional, or worth extending?
5. **Merge order.** The new docs sentence is true only after Intelligence #1289 ships. The fixtures are dev-only, so this PR is safe either way, but land #1289 first or at the same time.

Land it once the remaining checks finish.

PR #7230, opened 2026-09-17
Add an opt-in conversation reader for saved threads

ataibarkai:
Opening a saved thread currently starts with an event-derived diagnostic timeline, which can bunch user messages together and bury the conversation among run markers. Add an opt-in read-only Conversation view that follows canonical message order, keeps every role and rich payload accessible, and preserves Timeline, AG-UI Events, State, and the existing developer-inspector default.

Tool calls and results use keyboard-accessible disclosures without inferring success from arguments. Recorded run errors, missing message access, loading failures, and stale refreshes remain explicit. Regression coverage also fixes focus targeting a hidden duplicate after a tab switch a superseded initial fetch leaving the loading state stuck, and canceled old-provider failures leaking into a new provider with the same thread ID.

Validation: 719 unit tests across 39 files; package build and types; 1280px/390px browser tests for layout, native disclosures, keyboard tabs, and preserved state; lint has no errors (three existing monolith warnings). New behavior tests failed on the unchanged baseline first. A standalone fixture at `/conversation-reader.html` uses the production component and illustrative data for reproducible screenshots.

Release dependency: this PR does not publish a package. Intelligence currently pins 1.62.2 and needs a later verified published release containing `conversationView` before its adapter change can take effect. No package version bump, deployment, billing change, or automatic merge.

Independent review identified non-JSON provider payloads that could break rendering. The follow-up preserves the surrounding conversation, marks BigInt/circular/unsupported values explicitly, and displays a clear fallback if a provider value cannot be serialized. Two provider-level regressions failed before the fix and now pass. Final independent source/failure/type/comment/test/packaging review is clean.


samjulien:
Closing as superseded by the shipped conversational Inspector implementation and the reliability fixes merged in https://github.com/CopilotKit/CopilotKit/pull/7361. That PR incorporates the selected error-handling, accessibility, refresh, and caching improvements into the existing reader. This does not mean every idea in this PR was adopted; remaining ideas can be evaluated separately without maintaining a second conversation reader.

PR #7234, opened 2026-09-17
feat(showcase): auto-open a docs prod pin PR

AlemTuzlak:
Merge this, then docs prod ships when a maintainer merges the bot PR `release/docs/prod`. You do not run **Showcase: Promote** for docs-only changes.

After a docs merge to `main`, staging still updates first. If the docs staging probe is green, devops-bot opens one pin PR and leaves that candidate unchanged while it is under review. The digest must match before and after the staging probe. Merge of that PR pins `docs.copilotkit.ai` to the digest in `showcase/pins/docs-prod.json`.

## What does this PR do?

- Adds pin parse/write logic and a CLI that writes `showcase/pins/docs-prod.json`.
- Opens `release/docs/prod` after docs staging is green (`docs_open_release_pr.yml`).
- Pins prod on merge of that PR (`docs_promote.yml`), isolated from the `showcase-promote` queue.
- Drops standalone `docs` from fleet `service=all`. The promote dropdown still has `shell-docs` for an emergency.

## Related PRs and Issues

None.

## Checklist

### Commands run

1. `pnpm --dir showcase/scripts exec vitest run docs-prod-pin.test.ts prepare-docs-release-pin.test.ts docs-release-workflows.test.ts reconcile-staging.test.ts railway-envs.test.ts` — pass (151 tests)
2. `npx --yes bats showcase/scripts/__tests__/resolve-promote-targets.bats` — pass (17/17)
3. YAML parse of both new workflow files — `ok`

I did not run root `pnpm test` or `pnpm build`. This change is showcase scripts and GitHub Actions only. Root `oxfmt` is not installed in this worktree.

### Manual test

1. Merge this PR to `main`.
2. Merge a docs-only change, or run **Docs: open prod release PR** by hand.
3. Open the bot PR `release/docs/prod`. Make sure that the digest and staging URL look right.
4. Merge that PR.
5. Open `https://docs.copilotkit.ai/` and make sure that `.shell-docs-commit-label` matches the pin git SHA.

### How this PR makes testing easy

Unit tests cover pin parse, skip vs write, the CLI write path, and workflow contracts (`release/docs/prod`, `docs-promote` concurrency, `parseDocsProdPin` on promote). Bats covers `all` omitting standalone docs.

## Risk / rollback

A bad pin merge can pin prod docs to a bad digest. Rollback: `showcase/bin/railway promote docs --digest <prior-image> --yes`, or revert this PR to restore the old manual promote path. Fleet `service=all` no longer includes docs. Use `shell-docs` in the promote dropdown if you need an emergency pin.

## Review fixes (September 17)

- Unwrap the Railway token and test the real CLI live-fetch path with only HTTP mocked; the regression reproduced HTTP 401 before the fix.
- Gate both docs workflows with Showcase Validate, and check out the exact merged commit for promotion.
- Preserve open release candidates instead of refreshing away approvals. The runbook explains updating from main before approval, replacing a candidate, and current review requirements. GitHub's extra-approval rule concerns unattributed Copilot PRs, not arbitrary GitHub App commits.
- Document explicit-digest precheck tradeoffs, workflow chaining limits, and the distinction between verification-trigger SHA and image revision.
- Bind the candidate to the digest observed before the staging probe; skip if staging changes or disappears afterward.
- Merge current main without rewriting the existing PR commits.

Validation: 157 focused tests via Nx, 17/17 promotion-target bats tests, actionlint on both new workflows, and commit hooks passed. Standards review found no blocking violations (one optional metadata-name clarity note); spec review has no remaining findings. No live production promotion was run.


BenTaylorDev (commented):
Reviewed at head `ad4c393`. One blocking defect, six smaller findings, and a note on how close this gets us to continuous docs deployment.

## Blocking: the live pin path cannot authenticate

`showcase/scripts/prepare-docs-release-pin.ts:33` passes the wrong type.

```ts
const token = resolveRailwayToken();          // returns { token, source }
return liveFetchDeployedDigest(token, ...);   // parameter is token: string
```

Every other caller in the repo unwraps it: `provision-starter-fleet.ts:736` and `verify-autoupdates.ts:396` both use `resolveRailwayToken().token`.

Verified two ways in a worktree at this PR head:

```
prepare-docs-release-pin.ts(35,5): error TS2345:
  Argument of type 'RailwayTokenResolution' is not assignable to parameter of type 'string'.
```

```
# the same CLI on the flagless (production) path, with fetch intercepted:
Authorization header: Bearer [object Object]
Railway deployments query HTTP 401: unauthorized
EXIT=1
```

`liveFetchDeployedDigest` throws on a non-ok response, so the CLI exits 1 and the "Prepare pin file" step fails on every real run. The result is that the `release/docs/prod` PR never opens. The fix is one line:

```ts
const { token } = resolveRailwayToken();
```

Please also add coverage for the path that production actually uses. All three CLI tests pass `--staging-digest` and `--prod-digest`, so the live digest fetch is never exercised, and no CI job runs `tsc` over `showcase/scripts`. A fetch seam or an injectable digest fetcher closes the gap.

## Should fix

1. **The workflow contract test does not gate its own workflows.** `showcase_validate.yml` path filters list `showcase_build.yml` and `showcase_deploy.yml`, but not `docs_open_release_pr.yml` or `docs_promote.yml`. A PR that edits only the two new workflow files never runs `docs-release-workflows.test.ts`. This is the hole that the comment above the `showcase_build.yml` path entry already warns about. Add both paths.

2. **`docs_promote.yml` checks out the PR merge ref of a closed PR.** The job has no `ref:`, so `actions/checkout` resolves `refs/pull/N/merge`, which stops updating once the PR closes and is not guaranteed to exist. Pass `ref: ${{ github.event.pull_request.merge_commit_sha }}`, or trigger on `push` to `main` filtered to `showcase/pins/docs-prod.json` and drop the `head.ref` condition.

3. **The pin PR can lose its approval before it merges.** The main ruleset sets `dismiss_stale_reviews_on_push: true` and `require_last_push_approval: true`. `create-pull-request` re-pushes the branch on every Verify Deploy run on `main`, and the `shell_docs` build slot fires on `showcase/scripts/**` as well as on docs content. An approval given at 10:00 is gone after the next showcase build. Please state the intended landing procedure, for example auto-merge on the pin PR.

4. **Confirm the bot commit does not need a second approval.** The same ruleset carries `require_extra_approval_for_unattributed_changes: true`. If a devops-bot commit counts as unattributed, every docs release needs two humans instead of one.

## Notes

5. `--digest` deliberately skips P2's in-flight-race check and the staging-drift signal (`showcase/bin/railway:1890` and `:2012`). P1 (digest exists in GHCR) and P3 (staging live-green) still run, so the risk is bounded, but a stale pin merge ships an older digest with no warning. This repo documents every other bypass, so one comment in `docs_promote.yml` keeps that habit.

6. The chain is now push to Build and Push to Verify Deploy to Docs open PR. That is three levels of `workflow_run`, which is GitHub's documented chaining limit. A comment saves the next person from trying to add a fourth.

7. `git_sha` in the pin records the Verify Deploy head SHA. For a `showcase/scripts`-only commit that is not a docs-content commit, so the PR title reads "release docs" for a change that touched no docs.

## Verification run locally

- 16 new tests pass. Full `showcase/scripts` suite: 2557 pass, 8 fail. The same 8 fail on `origin/main`, so they are environment-only (oxfmt binary and registry fixtures).
- bats: 17 of 17 pass.
- Mutation-checked three mechanisms. Reverting the `standalone != true` jq filter fails bats 17. Renaming `release/docs/prod` and flipping `cancel-in-progress` fail both workflow-contract tests. The tests are not vacuous.
- `docs` is the only `standalone: true` service, and `standalone` is emitted on `.services[]`, not only on closure entries. So the `all` filter drops docs and nothing else: 41 prod-eligible becomes 40.
- The `railway` environment carries no protection rules, so the PR merge is the only human gate.

## How close this gets us

Docs staging was already decoupled. `shell_docs` is its own build slot with its own path filter, and `docs` is already `standalone: true`, so a `shell-docs` promote never dragged in the Tier-1 set.

The coupling that actually blocked continuous docs was the prod promote: one dispatch workflow on a single global `showcase-promote` concurrency group, plus `service=all` sweeping docs along with the fleet. This PR breaks both, and that is the real win.

What remains is one human merge per release, plus the review friction in findings 3 and 4. So this lands continuous delivery with a reviewable digest diff, not continuous deployment. The last step is either dropping the PR gate and promoting straight off the green staging probe, or enabling auto-merge on the pin PR once finding 4 is answered.

BenTaylorDev (commented):
The `Static analysis (zizmor)` failure is this PR's, not pre-existing.

```
error[dangerous-triggers]: use of fundamentally insecure workflow trigger
  --> .github/workflows/docs_open_release_pr.yml:3:1
      workflow_run is almost always used insecurely      severity: high
```

`docs_open_release_pr.yml` is new here and is the only new `workflow_run` trigger — `docs_promote.yml` triggers on `pull_request`, so it is not flagged. The repo's three existing `workflow_run` workflows (`showcase_deploy.yml`, `showcase_capture-previews.yml`, `test_smoke-starter.yml`) are all listed in `.github/zizmor.yml` under `dangerous-triggers.ignore`, each with a justification. This one is not, so the audit exits 14.

The suppression is warranted, and by a stronger argument than its siblings carry: the trigger is the in-repo "Showcase: Verify Deploy" workflow filtered to `branches: [main]`, the `pending-release` job guards again on `github.event.workflow_run.head_branch == 'main'`, and the checkout is `ref: main` with `persist-credentials: false`, so no fork-controlled ref is ever fetched. Workflow-level permission is `contents: read`.

The config asks that every new entry carry a comment explaining why the workflow is safe, so something like:

```yaml
      # docs_open_release_pr.yml: workflow_run triggered by the in-repo
      # "Showcase: Verify Deploy" workflow, filtered to main. The job
      # re-checks head_branch == 'main' and checks out `ref: main` with
      # persist-credentials: false, so no fork-controlled ref is fetched.
      - docs_open_release_pr.yml
```

Everything else on the PR is green, and the blocking defect from my last pass is fixed — I mutation-checked the new `authenticates both live digest queries without override flags` test and restoring the old call reproduces the `401 unauthorized` exactly. Once zizmor is green I am happy to approve.

BenTaylorDev (approved):
Approving at `caafb7af`. zizmor is green and all seven findings from my earlier passes are fixed: the token unwrap (`prepare-docs-release-pin.ts:35`), both workflow paths in `showcase_validate.yml`, the durable checkout ref, the `pending-release` guard that protects an approval, and the missing comments on the `--digest` bypass, the chain limit and `git_sha`.

I tried to break three things and could not. A fork cannot reach the `workflow_run` trigger, because the chain roots at a `push` to this repo's `main`. `--digest` wants the full `image@sha256:…` ref, which is exactly what the pin schema enforces. Against the real generated SSOT the `standalone` filter drops one name from `all`: `docs`.

Two notes, neither blocking. Nothing here can run until it is on `main`, so please dispatch the open-PR workflow by hand and use `docs_promote.yml`'s `dry_run` once before the first real release. And `require_last_push_approval` means whoever presses "Update branch" cannot approve the result, which README step 2 reads as though one person can do.

PR #7254, opened 2026-09-17
feat(runtime): add automatic skill delivery to BuiltInAgent

maxkorp:
BuiltInAgent can now consume published Learning skills through an optional `learnedSkills` configuration, without a wrapper or separate adapter package. Classic mode adds the catalog and executable skill tools automatically. Factory callbacks receive `{ catalog, tools }`; both are empty when delivery is unconfigured or a verified snapshot contains no skills.

Each invocation, including a resume, captures one verified snapshot. Clones share the refresh cache, while tool closures retain their invocation's snapshot. Existing revision pins, ETags, stale-cache recovery, and fail-closed delivery denials are preserved. The shared registry now lives inside Runtime, with the private adapter package retaining its exports as a compatibility facade. Classic mode defaults to 10 steps when skills are available so the model can use loaded guidance; explicit `maxSteps` still wins.

Validation:
- Runtime: 2,500 tests passed; shared registry: 89; LangGraph: 40; Mastra: 21. All four packages built and passed type checks.
- Real AI SDK execution over the canonical HTTP client covers both skill tools, follow-up model context, empty snapshots, and refreshed resumed runs. Real interrupt tools also pause correctly before and after skill loading with the default multi-step limit. Unit coverage also includes cloning, cancellation, pinned revisions, denial behavior, and tool-name collisions.
- A clean external consumer installed packed Runtime artifacts, imported ESM/CJS, exercised both modes, and compiled both documentation examples under strict TypeScript/NodeNext.
- Minimum and latest supported versions of both TypeScript adapters passed standalone installation, ESM/CJS, type inference, and native execution suites.
- Independent code review found a cancellation race; the fix and regression test are included.

The usage tests use a scripted model and local HTTP snapshot server. Live Learning generation and browser UI testing were not performed for this change.

Review notes:
- The real AI SDK pauses at interrupt tools without results, even with a multi-step limit. Tests cover interrupts both before and after loading a skill; the old single-step-only comment was corrected.
- The suggested HTTPS-only policy would change existing registry and canonical-client behavior, including private HTTP deployments. This PR preserves the existing URL policy; transport-policy changes need a separate SDK-wide decision.

Martha review follow-up:
- Export `BuiltInAgentFactoryContext` without changing the existing request-scoped `AgentFactoryContext`, and type the frozen skill tools as `Readonly<ToolSet>`. Source type checks and an installed NodeNext consumer cover both.
- Stop MCP setup after cancellation at every asynchronous boundary, and close a managed client that finishes connecting after unsubscribe. Regression tests cover pending connection and tool lookup.
- Cover TanStack delivery and resumed refresh, an explicit classic `maxSteps` override, and classic delivery errors. Correct empty-snapshot docs and the factory context name.

Final CI: 78 checks passed, 3 skipped on `99bb9e4b1ece6fac584eecb9302323b1a5177b0f`. The docs UI timeout passed on rerun. Branch includes freshly fetched `origin/main`.


marthakelly (commented):
The remaining CodeRabbit feedback about cancellation during MCP setup and the empty-snapshot docs looks valid to me.

**The exported `AgentFactoryContext` has no `learnedSkills`**

The docs point users to `AgentFactoryContext` from `@copilotkit/runtime/v2`, but that export only contains `request`.

The BuiltInAgent context is hidden by the name collision. Can we export it as `BuiltInAgentFactoryContext`?

**Mutating `learnedSkills.tools` compiles and then throws**

The runtime freezes this map, but its public type is mutable `ToolSet`.

Can we use `Readonly<ToolSet>` so TypeScript rejects mutation?

**Three learned-skill paths can break without failing CI**

The tests do not cover TanStack delivery, an explicit `maxSteps` override, or classic-mode delivery errors.

Can we add these cases?

showcase/shell-docs/src/content/docs/backend/custom-agent.mdx
```diff
@@ -160,10 +160,11 @@ Factory mode accepts a config with two fields:
 - **`type`:** which backend you're using: `"aisdk"`, `"tanstack"`, or `"custom"`
 - **`factory`:** a function that receives the raw request and returns a backend-native stream
 
-The factory receives an `AgentFactoryContext` (from `@copilotkit/runtime/v2`):
+The factory receives an `BuiltInAgentFactoryContext` (from `@copilotkit/runtime/v2`):
```
marthakelly, line 163: **The factory guide makes `learnedSkills` look unavailable**

Factory mode now accepts `learnedSkills`, but this section still says the config has two fields. It lists only `type` and `factory`.

Can we add `learnedSkills` to this list and the shared version?

packages/runtime/src/agent/__tests__/learned-skills.test.ts
```diff
@@ -0,0 +1,388 @@
...
+      contexts[0].learnedSkills.catalog,
+    );
+    fetch.mockRejectedValue(new LearnedSkillsError("DELIVERY_DISABLED", false));
+    await expect(
+      collectEvents(agent.run(createDefaultInput())),
+    ).rejects.toMatchObject({ code: "DELIVERY_DISABLED" });
+    expect(contexts).toHaveLength(2);
+  });
+
+  it("rejects a client tool that shadows a learned skill tool before calling the model", async () => {
```
marthakelly, line 368: **Reserved skill tools can be shadowed without a test failure**

The docs say client, config, and MCP tools cannot use the reserved skill-tool names. The current test covers only client tools.

Can we add cases for `config.tools`, `mcpClients`, and `mcpServers`?

marthakelly (approved):

PR #7262, opened 2026-09-17
fix(web-inspector): verify evidence content before highlighting

marthakelly:
## What does this PR do?

If a message changes after Learning loads, its evidence link can highlight text that Learning never analyzed. The message ID stays the same, so navigation by ID cannot detect the edit.

Inspector now checks the expected message hash against the current conversation and the row it will highlight. A mismatch, removed message, missing hash, or failed verification shows an unavailable notice. The Thread remains open for inspection. Changes during verification cannot apply an old result, and a later content update clears an existing highlight.

Shared preserves optional `messageHashes` through runtime and browser parsing. Each entry is SHA-256 of `JSON.stringify([role, content])`, aligned with `messageIds`. Both parsers reject malformed hashes and mismatched counts.

## Related PRs and Issues

- API producer: https://github.com/CopilotKit/Intelligence/pull/1268
- Refs https://linear.app/copilotkit/issue/PE-53

Deploy the API producer and update the runtime and Inspector to enable the full check. Older runtimes discard hashes. With missing hashes or unavailable Web Crypto, the new Inspector shows the notice and does not highlight evidence. Web Crypto requires HTTPS or localhost. Ordinary Thread navigation is unchanged. Existing snapshot-only assistant and tool-result focus gaps still show unavailable evidence.

## Validation

- `NX_DAEMON=false pnpm nx run @copilotkit/web-inspector:test --excludeTaskDependencies` — 717 tests pass, including 11 new navigation cases.
- `NX_DAEMON=false pnpm nx run @copilotkit/shared:test --excludeTaskDependencies -- src/utils/inspector-learning` — 13 tests pass.
- `NX_DAEMON=false pnpm nx run-many -t check-types,build -p @copilotkit/shared,@copilotkit/web-inspector` — passes, including test types.
- `NX_DAEMON=false pnpm nx run @copilotkit/web-inspector:test:browser` — 11 browser tests pass.
- Oxlint on changed TypeScript files through `nx exec` — zero errors, 9 warnings. Oxfmt and `git diff --check` pass.

The new navigation tests use the actual Learning controls, response parser, Thread store, and detail renderer. Before the fix, seven cases failed because Inspector did not reject unverified evidence. Unchanged user and assistant messages remain positive controls. Additional cases cover edits during verification and after an initial highlight.

A manual browser check used persisted API output in the standalone workbench. Unchanged evidence opened the original refund text. Editing the same message before the click showed the unavailable notice without a highlight. This used local HTTP fixtures, not a deployed Intelligence service. Production latency was not measured.

The full pre-commit package gate also passes: tests, publint, and attw across all 28 affected packages, with their build dependencies. The initial worktree dependency and SQLite native-binding errors were resolved before commit.

## Documentation

Updated Shared and Web Inspector READMEs with the hash format and mixed-version behavior. The API PR documents the producer contract and rollout order.

## Summary by CodeRabbit

* **New Features**
  * Verifies learning evidence before highlighting referenced messages.
  * Shows verification status when evidence is changed, removed, unavailable, or cannot be verified.
  * Preserves optional message hashes and supports older evidence data without hashes.
  * Clears highlights when message content changes during verification.

* **Documentation**
  * Documented evidence hash formats and verification behavior.

* **Tests**
  * Added coverage for valid, invalid, missing, and unavailable hashes and navigation scenarios.


tylerslaton (approved):

tylerslaton (approved):
Reviewed at 6c2d8988. No actionable findings in the standards or behavioral review, including hash compatibility with the linked producer, verification races, and failure handling. Tests were not rerun locally because this worktree has no installed dependencies.

marthakelly (commented):
**A failed message refresh can leave stale evidence marked as verified**

When live messages change, Inspector starts a silent `/messages` refresh. While it waits, verification still uses the old conversation. If the refresh fails, the old evidence remains accepted even though Inspector could not check the current message.

Can we mark the evidence as unverified when `liveMessageVersion` changes, then verify it again after the refresh succeeds? Please add a regression test where the message changes and the next `/messages` request fails.

**The browser suite can pass without exercising evidence verification**

The new navigation tests replace `crypto` with Node Web Crypto. None of the Playwright tests clicks an evidence link or checks the browser `crypto.subtle` path. The browser suite can stay green if the Learning-to-Threads wiring or browser hash check breaks.

Can we add a Playwright test for a valid hash and an invalid or missing hash?

PR #7270, opened 2026-09-18
feat: AG-UI 1.0 for CopilotKit

mme:
Moves CopilotKit onto the released AG-UI 1.0 packages, with the source changes 1.0 requires.

Supersedes #6944, which validated the same work against `pkg.pr.new` preview tarballs while 1.0 was still unreleased. AG-UI 1.0.0 is now published on npm, so the previews and everything built to consume them are gone.

## Dependencies

`@ag-ui/core`, `client`, `proto` and `encoder` move to `1.0.0` across 23 manifests. The integration packages keep the versions already in use, because they have not been released for 1.0: `langgraph` 0.0.43 (0.0.42 in sdk-js), `mcp-middleware` 0.0.2, `mcp-apps-middleware` ^0.1.1, `a2ui-middleware` 0.0.10. They resolve their own older `@ag-ui/client` transitively, exactly as on main today.

The scaffolding that only existed to consume URL pins is removed: the twelve `pnpm.overrides` entries that forced every transitive dependency onto a preview tarball, the root AG-UI devDependencies and the `yaml` helper they needed, the explicit client dependencies added to `sdk-js` and the Angular demo server for exotic-dependency re-resolution, and `scripts/release/lib/ag-ui-previews.ts` with its test. The Angular, Runtime and Channels packed-consumer verifiers are back to their base form, since there is no preview tarball to materialize.

Two dependencies stay, for the reason they were added. The Vue demo's server code imports `@ag-ui/client` directly. And `@copilotkit/core`'s raw-event test imports `@ag-ui/core/schemas` — that had only ever resolved through the overrides, so removing them surfaced an undeclared dependency, now a devDependency of that package.

## Source changes for 1.0

Runtime and Channels emit completion reasons in `metadata.finishReason`, preserving other metadata and usage aggregation. Runtime schema imports use `@ag-ui/core/schemas`; private legacy binary adapter types retain image/PDF conversion. Test agents supply the required lifecycle identifiers and error messages while preserving their behaviour assertions.

CopilotKit's run/connect request parsers apply a private local copy of AG-UI's seven legacy request-null conversion rules before strict validation. Direct request parsing is the documented exception to AG-UI's internal compatibility boundary, so this needs no new public AG-UI helper. Historically accepted whole optional nulls become absent, required and nested application nulls survive, and fields already invalid on main remain invalid. Regression tests cover parsing and actual HTTP/BuiltInAgent execution with the public helper absent.

Runtime's existing OpenAI peer is required by removing only its optional marker, with the generated public API manifest updated accordingly. Dependency ranges and runtime source are unchanged, preserving caller-provided OpenAI instances.

## Summary by CodeRabbit

* **New Features**
  * Added support for AG-UI 1.0 content formats across chat, tool rendering, and agent integrations.
  * Tool results now convert text and media content into readable output.
  * Run completion metadata, including finish reasons and usage details, is preserved across nested runs.

* **Bug Fixes**
  * Improved handling of legacy null request fields while preserving valid application data.
  * Improved state and event validation and recovery for malformed updates.
  * Cancelled runs now transition correctly to the final agent node.


nathan-tj-xie:
waiting for release

AlemTuzlak:
## Review of the AG-UI 1.0 move: what we found and what we changed

We reviewed this PR for three things: breaking changes for existing users, compliance with the AG-UI 1.0 spec, and Intelligence. We also checked whether CopilotKit can accept AG-UI 0.x and 1.0 at the same time, so that this can ship in a minor release. We pushed the fixes to this branch and merged `main` three times. The last merge is `71bbd2d42e`.

### What we found

**Blockers**

1. **The runtime rejected legacy `binary` parts with HTTP 400.** The request parser uses the 1.0 `RunAgentInputSchema`, which has no `binary` part. 0.0.59 accepted it. Older frontends, stored history, and app code that adds `binary` parts (for example, the showcase multimodal shim) failed.
2. **Thread replay skipped the 1.0 compatibility boundary.** `ɵconnectWithoutEventVerification` is a copy of the 0.0.57 `connectAgent` pipeline. Every reconnect uses it, both self-hosted and Intelligence. Stored `THINKING_*` events and `binary` parts in `MESSAGES_SNAPSHOT` were not translated, so reasoning history was lost and attachments did not render.
3. **Intelligence run streams could flood the console.** The Intelligence runner stamps `thread_id` and `run_id` on every event. It also adds `threadId` and `runId` to events that do not declare them. The 1.0 client strips unknown fields and warns once per field, per event.
4. **`instanceof HttpAgent` failed for an `HttpAgent` from another `@ag-ui/client` copy.** This happens when an app pins its own 0.x client. CopilotKit then skipped core headers (auth, license) and abort wiring, and it did not report an error.

**Spec gaps (1.0 MUST rules)**

- A stopped run finished with no outcome, which means success. The spec requires `outcome: { type: "cancelled" }`.
- The runner finalizer did not close reasoning. In 1.0 the client fails a run that ends with reasoning still open, so a stop during reasoning failed the run. The synthetic `RUN_FINISHED` also had no `threadId` or `runId`.
- `BuiltInAgent` did not send `protocolVersion` on `RUN_STARTED`.
- A provider file handle (`source.type: "file"`) was sent to the model as base64 media. `getSourceUrl` turned it into `data:undefined;base64,<handle>`.

**Breaking changes and misses**

- The react-core and vue v2 entries re-export `@ag-ui/client`, which no longer exports the zod validators. About 110 names that users import today would have disappeared.
- The `mcp-apps-renderer` peer was pinned to exactly `@ag-ui/client` 0.0.59, so every React install reported an unmet peer.
- The runtime imports `@ag-ui/core/schemas`, which loads `zod/v4`, but it accepted `zod` ^3.23.3. `zod/v4` needs 3.25.18 or later.
- The first merge of `main` failed to type-check. `shared`'s `>=0.0.48` peer resolved to `@ag-ui/core` 0.0.57 inside the workspace.
- The PR description said that the integration packages use an older client of their own. They use the 1.0 client through their `>=0.0.x` peer. We corrected the description.
- There was no upgrade note for users.

### What we changed

| Commit | Change |
| --- | --- |
| `1b8a6e7c3f` | Merged `main`. `shared` now dev-depends on `@ag-ui/core` 1.0.0, and the lockfile is re-resolved from main's version. |
| `8731c50e11`, `fa4794a78d` | `mcp-apps-renderer` accepts `@ag-ui/client >=0.0.59`. The public API manifest records it. |
| `ed070800c3` | The react-core and vue v2 entries also re-export `@ag-ui/core/schemas`. Only names that 1.0 removed are gone (`THINKING_*` types, `BinaryInputContent`, `SubAgentInfo`, `BackwardCompatibility_0_0_47`). |
| `d47ffa3b31` | The request parser upgrades `binary` parts to media parts, as the client does. It drops a mirror of a modern part, and drops an id-only part with a warning. |
| `79cda350d4` | Replay runs through `CompatibilityBoundary`. It still skips `verifyEvents` and also skips `enforceEvents`, so one malformed stored event cannot block hydration. The gateway replays `RUN_STARTED` without `runId`. |
| `43ef41b753` | `stripIntelligenceRoutingFields` removes the routing fields on the run stream only. The gateway still receives them. |
| `53a034d7e8` | `ɵisHttpAgent` replaces `instanceof HttpAgent` in core, react-core, vue and angular. |
| `6d927178e3` | A stop finishes as `cancelled`, but only when the client declared `protocolVersion`, because a 0.x client cannot parse that outcome. The finalizer closes reasoning and stamps `threadId` and `runId`. `BuiltInAgent` sends `protocolVersion`. |
| `832df205f9` | File-handle parts are dropped with a warning. A tool result whose parts were all dropped is `""`. `getSourceUrl` returns `""` for a handle. React Native accepts the part. |
| `b7fdfb54b9` | The runtime requires `zod` ^3.25.18. |
| `8ec452a6a4` | The new helper uses string literals instead of the `EventType` value. That one import made the bundler put the client and zod v4 into the initial chunk: the Angular showcase went from 1.49 MB to 1.95 MB and failed its budget. It is back to 1.49 MB. |
| `60b8bb0c51` | A new docs page: `docs/migrate/ag-ui-1.0`. |

### Can 0.x and 1.0 run together?

- **On the wire: yes.** A 1.0 frontend reads 0.x streams through the client's compatibility boundary. A 0.0.59 frontend reads the 1.0 runtime's output, because the additions are optional and `cancelled` is sent only to 1.0 clients. The runtime accepts 0.x requests, including `binary` parts and legacy nulls.
- **At the TypeScript level: not fully.** An app with its own `@ag-ui/*` 0.x can see type errors, because `AbstractAgent` and `Message` come from two copies. Such apps must upgrade `@ag-ui/*` to 1.0 together with CopilotKit. The upgrade guide says this. Full type-level support would need a structural agent type in CopilotKit's public API. We did not add one.

### How we verified it

- `check-types` passes for all 32 affected projects.
- Full test suites pass for these packages (number of tests):

  | Package | Tests | Package | Tests |
  | --- | --- | --- | --- |
  | runtime | 2635 | react-core | 1732 |
  | vue | 1146 | core | 954 |
  | web-inspector | 728 | shared | 502 |
  | channels-slack | 461 | angular | 369 |
  | channels-core | 313 | react-native | 299 |
  | channels-intelligence | 219 | channels-telegram | 173 |
  | runtime-client-gql | 133 | sdk-js | 118 |
  | intelligence-delivery-core | 89 | react-ui | 75 |
  | channels-whatsapp | 74 | intelligence-langgraph | 40 |
  | intelligence-mastra | 21 | mcp-apps-renderer | 16 |
  | sqlite-runner | 15 | agentcore-runner | 5 |

- During the full runs, four tests timed out under machine load. All four pass when run alone.
- CI was fully green on `8ec452a6a4`. CI runs again on `71bbd2d42e` after the latest merge of `main`.
- One note: we skipped the pre-commit hook's `test-packages` step, because it needs Ruby and .NET toolchains that the machine did not have. We ran the tests above instead.

### Still open before merge

- **Test with the real Intelligence gateway.** The Intelligence fixes rest on the gateway shapes that the test fixtures encode. A manual run is still necessary: live streaming, reconnecting a thread, and stopping a run.

PR #7271, opened 2026-09-18
fix: restore pending frontend HITL tools after replay

ranst91:
Reloading an Intelligence thread restored frontend HITL tool calls but left them without an active response handler, so the user could see the approval UI without being able to answer it.

This adds an internal frontend-tool classification and tags HITL registrations in React, Vue, and Angular. Shared core restores unanswered HITL handlers after replay and continues the agent with the user's result. Ordinary frontend tools remain passive; native interrupt handling is unchanged. React Native and legacy React wrappers inherit the React registration.

Restoration finishes independently of the connection lifecycle, preserves sequential prompts across reconnects, ignores saved or remotely supplied answers, and cancels superseded interactions when switching threads. The classification is not sent in the AG-UI tool schema.

Validation: formatting and lint pass; explicit type checks, full test suites, and builds pass for core, React, Vue, and Angular. Coverage includes replay → answer → continuation, saved and remote answers, repeated reconnects, sequential prompts, thread switching, and keeping tool classification off the wire. Commit hooks also passed tests and package-export checks across 24 affected projects. Live Intelligence browser validation remains a follow-up before marking this draft ready.

PNI-449

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved human-in-the-loop tool behavior after reconnecting or switching conversations.
  * Pending approval requests now resume correctly, while completed or remotely answered requests are not duplicated.
  * Responses remain sequential for multi-step interactions, and stale responses are safely ignored.
  * Ordinary tools continue to operate normally during standard agent runs.
* **Improvements**
  * Human-in-the-loop tools are now explicitly distinguished from regular tools for more reliable processing across supported Angular, React, and Vue integrations.


packages/core/src/core/run-handler.ts
```diff
@@ -714,73 +765,124 @@ export class RunHandler {
...
-            let wildcardTool: FrontendTool<any> | undefined;
-            const getWildcardTool = () => {
-              if (tool || wildcardTool) {
-                return wildcardTool;
-              }
-              wildcardTool = this.getTool({
-                toolName: WILDCARD_TOOL_NAME,
-                agentId: agent.agentId,
-              });
+    for (const message of newMessages) {
```
tylerslaton, line 810: [P2] Reconcile already-known pending calls after a remote answer

If the initial replay contains unanswered HITL calls A and B, the first restoration pass waits on A before reaching B. When another client answers A, a same-thread reconnect excludes B from `newMessages` because its message ID was already present (see `connect-replay.ts`). The barrier above releases, but this loop never visits B. Meanwhile, the original pass still awaits A's local promise, and A's completed UI no longer exposes `respond`, so B cannot progress. The new remote-answer test misses this because B first appears during the second replay. Please reconcile unanswered calls from current `agent.messages`, or settle/cancel the superseded local handler, and cover the case where both A and B exist before the remote answer.

packages/core/src/core/run-handler.ts
```diff
@@ -233,6 +244,9 @@ export class RunHandler {
    */
   abortCurrentRun(): void {
     this._runAbortController?.abort();
+    for (const replay of this._replayAbortControllers.values()) {
```
tylerslaton, line 247: [P2] Scope replay cancellation to the agent being stopped

`CopilotKitCore.stopAgent({ agent: A })` calls this method, but the new loop aborts replay controllers for every agent. If B has a restored pending approval, stopping A also rejects B's HITL handler and clears its response resolver. Because replay aborts discard the result, B is left unanswered without a working handler until reconnect. Please pass the target agent into cancellation and abort only its replay controllers. A regression test should restore B's approval, stop A, and verify B can still answer and continue.

tylerslaton (changes requested):
Two replay lifecycle issues need addressing before merge. I confirmed both control-flow cases using isolated reproductions of the changed methods; I did not run the full test suites.

packages/core/src/core/run-handler.ts
```diff
@@ -714,73 +770,157 @@ export class RunHandler {
...
+              this._interactionAbortControllers.set(agent, interactions);
+              signal?.addEventListener("abort", abortInteraction, {
+                once: true,
+              });
+              if (signal?.aborted) abortInteraction();
+            }
+            const executionSignal = interactionController?.signal ?? signal;
+            const discardOnAbort =
+              toolExecutionMode === "human-in-the-loop" ||
+              !!interactionController;
```
tylerslaton, line 952: [P2] `stopAgent` during a live HITL no longer records a tool result

With `discardOnAbort` true for every live tool tagged `human-in-the-loop`, a stop skips both the result insert in `executeSpecificTool` and the `TOOL_HANDLER_FAILED` error. I confirmed with a core probe: the tagged path leaves zero tool results and no error; an untagged tool on the same stop (main's behavior, and the provider-prop path today) inserts `Error: Human-in-the-loop interaction aborted` and emits `TOOL_HANDLER_FAILED`.

For agents without `/connect` (plain `HttpAgent`, every non-Intelligence app) nothing restores the call afterwards: the card drops to `InProgress` with no `respond` (`onToolExecutionEnd` clears `executingToolCallIds`, and no tool message exists), and the next user message ships an assistant tool call with no result. `ɵrepairToolCallPairs` only runs when a `messageFilter` is set. Reachability is programmatic (`stopAgent` / `abortRun`), since the chat's Stop button is gated on `agent.isRunning`, which is already false while a HITL waits.

The "restores a stopped live interaction even if its handler ignores cancellation" test asserts the new behavior, so I take this as deliberate. Please either keep the error result when the agent cannot be restored (connect not implemented / no reconnect will follow), or state the new contract explicitly and update the comments in `use-human-in-the-loop.tsx` and both providers that still say the abort exists "so core records an explicit error tool result".

packages/core/src/core/run-handler.ts
```diff
@@ -224,15 +224,34 @@ export class RunHandler {
   private _anonymousAgentIds = new WeakMap<AbstractAgent, string>();
   private _nextAnonymousAgentId = 0;
 
+  /** Prevent reconnects from starting an interaction already executing locally. */
+  private _executingToolCalls = new WeakMap<AbstractAgent, Set<string>>();
+
+  private _interactionAbortControllers = new Map<
```
tylerslaton: [P3] Make this a `WeakMap`

Entries live until every controller settles, and a handler orphaned by a remote answer never settles until an abort. A thread-scoped `useAgent({ runtimeAgentId, threadId })` proxy unmounted with a pending HITL is retained forever. Nothing iterates this map, so it can be a `WeakMap` like `_executingToolCalls`.

tylerslaton (changes requested):
Re-review at 6825cbe. Both earlier P2s are resolved and covered by tests (remote-answer reconciliation via the full `agent.messages` walk plus the "already known: true" case; per-agent `abortCurrentRun`). Two new P2s below, one inline and one outside the diff.

**[P2] Provider-prop HITL registrations are not tagged (outside the diff)**

The React `humanInTheLoop` prop path (`packages/react-core/src/v2/providers/CopilotKitProvider.tsx`, `processedHumanInTheLoopTools`) and the Vue equivalent (`packages/vue/src/v2/providers/CopilotKitProvider.vue`, `processedHumanInTheLoop`) build their `FrontendTool` without `type: "human-in-the-loop"`. Those tools are still not restored after replay, are not cancelled on thread switch, and keep the old stop behavior, so the two registration paths now diverge. Angular is fine because both its paths go through `#bindHumanInTheLoopTool`. Please add `type` in both providers with a test.

**Verified locally**

- `type` never reaches the wire; ordinary tools stay passive on connect; wildcard HITL restores.
- Extra probes: a live HITL plus two same-thread churn reconnects yields one handler, one result, one follow-up; a user run while a restored HITL is pending still lets the answer land and continue.
- Suites at 6825cbe: core 941 passed, react-core 1714 passed, vue HITL 26 passed, angular 354 passed.

tylerslaton (approved):

